PR #146, opened 2018-03-16
MH-12775 Add JavaScript source map generation

JulianKniephoff:
This is my preliminary work for [MH-12775](https://opencast.jira.com/browse/MH-12775). It currently adds source maps for the JavaScript files only, so in particular no CSS source files. I am working on adding those as well but I encountered a problem which I will likely not solve before the weekend and I wanted to have a PR out for people to potentially discuss over the weekend. Or we could already merge this (after a review, of course) and add CSS source maps in a later PR, I don't care either way.

The generation is pretty roundabout, unfortunately, which is mainly due to the many (nearly) unmaintained build tools we use. However, the result source maps currently at least allow for browsing the original JS code in both, Chrome and Firefox. Breakpoints work, too, although stepping through the code behaves a bit strange sometimes. The name resolution is also a bit wonky still. When hovering over variables in debug mode, you should be able to see their values, but the scope section (in Chrome) still shows minified names and the console also does not understand the original names, unfortunately.

I unfortunately don't know how much of this is my fault, the tools' fault or the browsers fault, but I will try to research further. In the long run, we should probably update some of these old tools anyway, though.

But hey, it's a start! :smiley: 

Oh and I also couldn't help myself but to clean up the Gruntfile and its dependencies a bit, so if you look at the diff here on GitHub, remember that [`w=1`-trick](https://github.com/opencast/opencast/pull/117#issuecomment-370500980) I learned from @smarquard, to ignore pure whitespace changes. :wink:

PR #216, opened 2018-05-04
MH-12323: Add `package-lock.json`

JulianKniephoff:
[As discussed on list](https://groups.google.com/a/opencast.org/forum/#!topic/dev/jCqzhkIUCZU), here is a PR adding `package-lock.json`.

This implements [MH-12323](https://opencast.jira.com/browse/MH-12323) which could be understood as a feature, but it also fixes [MH-12865](https://opencast.jira.com/browse/MH-12865) making this kind of a bug fix; so I hope it's okay for me to file this against 3.x, but I can change that of course if the review gods decide that. :)

One important thing to note here is: Some care must be taken into forward merging this pull request. For example: #146, merged into 5.x, introduced a few updates to some npm dependencies. This will change `package-lock.json` on build, which is fine (necessary even), but these changes then need to be committed as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,9 +39,6 @@ docs/guides/*/site
 Makefile
 /modules/matterhorn-admin-ui-ng/bower_components/
 
-# Ignore package-lock.json
-/modules/matterhorn-admin-ui-ng/package-lock.json
-
 # Ignore OS artifacts
 .DS_Store
 
--- a/modules/matterhorn-admin-ui-ng/package-lock.json
+++ b/modules/matterhorn-admin-ui-ng/package-lock.json
@@ -1,0 +1,7175 @@
+{
+  "name": "adminui",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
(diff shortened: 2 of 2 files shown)


gregorydlogan:
Is this safe to merge into 4/4.x as is?  I'll do the full merge from 3.x, to 4.x, to 5.x, but it's also a new file in 4.x.  I'm assuming the versions specified in the lock are correct for 4.x at least?

lkiesow:
While merging this, we would at least need to check.
If I would do the merge, I would simply remove it for each version and rebuild the admin module to check if there are any changes.

gregorydlogan:
As discussed on IRC, @lkiesow will merge this forward and do the appropriate checks.

JulianKniephoff:
In my own tests, building the admin UI on 4.x without a `package-lock.json` resulted in the same one committed here. 5.x produced a different one, and there were also differences between building 5.x without a `package-lock.json` and building it with this one. You see, npm **can** update the `package-lock.json` if necessary, but I think it tries to minimize the changes, which is sensible, IMO. So it should not actually be necessary to do a clean rebuild for every branch.

PR #217, opened 2018-05-07
MH-12870: allow series acl update to complete even if mediapackage exceptions are raised at the mediapackage level

slampunk:
Suppose a series has a mediapackage which, for some reason, cannot be snapshotted. Lets call this mediapackage _corrupted_.

If an ACL update is applied to the series (e.g. to share series content with another series), the update consequently fails at the corrupted mediapackage as it cannot be snapshotted. All those sibling mediapackages which have been snapshotted prior to the attempt against the corrupted mediapackage will have the update applied to them; on the other hand, all mediapackages subsequent to the corrupted mediapackage remain unchanged.

We now have mediapackages in a series where ACLs are not applied consistently.

This PR puts forward a best attempt approach, whereby those mediapackges which can be updated _are_ updated (exceptions at the mediapackage level do not stop the update); those that can not are logged for administrators to deal with.

### DIFF
--- a/modules/matterhorn-authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/impl/AclServiceImpl.java
+++ b/modules/matterhorn-authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/impl/AclServiceImpl.java
@@ -243,8 +243,12 @@ public final class AclServiceImpl implements AclService {
 
         for (MediaPackage mp : mediaPackages) {
           // remove episode xacml and update in archive service
-          if (assetManager != null)
-            assetManager.takeSnapshot(authorizationService.removeAcl(mp, AclScope.Episode));
+          try {
+            if (assetManager != null)
+              assetManager.takeSnapshot(authorizationService.removeAcl(mp, AclScope.Episode));
+          } catch (Exception e) {
+            logger.info("Error applying series ACL to mediapackage {}", mp.getIdentifier().toString());
+          }
         }
       }
       // update in series service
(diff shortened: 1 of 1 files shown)


modules/matterhorn-authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/impl/AclServiceImpl.java
```diff
@@ -243,8 +243,12 @@ public boolean applyAclToSeries(String seriesId, AccessControlList acl, boolean
 
         for (MediaPackage mp : mediaPackages) {
           // remove episode xacml and update in archive service
-          if (assetManager != null)
-            assetManager.takeSnapshot(authorizationService.removeAcl(mp, AclScope.Episode));
+          try {
+            if (assetManager != null)
+              assetManager.takeSnapshot(authorizationService.removeAcl(mp, AclScope.Episode));
```
staubesv, line 248: Hi @slampunk 

Not related to your change but... have you got an idea why the Episode ACL is removed here? (instead of the Series ACL)

slampunk: @staubesv I have no idea tbh. I had assumed it was a design choice to keep mediapackage ACLs in lockstep with that of their container series'.

lkiesow: I would assume the same. I guess the idea is that if someone edits a series' acl all episodes belonging to that series should get that acl.

PR #219, opened 2018-05-07
MH-12869 Remove superfluous playback tool

staubesv:
See discussion and proposal on https://groups.google.com/a/opencast.org/forum/#!topicsearchin/dev/remove$20playback/dev/U-Io91-xA_U

Now looks like:

screen shot 2018-05-07 at 17 05 22

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -476,7 +476,6 @@
            "LOCATION": "Filter for this location",
            "DETAILS": "Open event details",
            "DELETE": "Delete event",
-           "PREVIEW": "Open preview player",
            "EDITOR": "Open video Editor",
            "COMMENTS": "View comments",
            "PAUSED_WORKFLOW": "View paused workflow"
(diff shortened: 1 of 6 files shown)


CGreweling:
@staubesv so the proposal has passed. I will have a look and merge it in.

CGreweling:
@staubesv  Works. 
But now we see th Preview symbol as an eye when the event ist published. This should than be an cut icon for the editor. As it is now its a little confusing.
preview
Maybe this schould be done in another PR ?

staubesv:
@CGreweling Good point. The cut icon is currently used to indicate that the event has an open comment "Requires Cutting". If we would just always show the cut icon, that additional information would get lost.

I definitely missed to adapt the tooltip, though.

What would you suggest considering the "eye" vs "cut" icons?

staubesv:
Hmm... What would you think of always using the cut icon but in different shades (depending on whether there is an open comment)?

staubesv:
Hi @CGreweling I've played around with the icon and ended up with the following:

screen shot 2018-05-17 at 15 17 11

Sorry, I'm quite "un-knowledgable" if it comes to CSS so that is about the best I could achieve ;-)

The red dot indicates the the event "needs cutting". The tooltip does explain that.

PR #220, opened 2018-05-07
MH-12871 Ability to use user names in to/cc/bcc fields in send-email woh

rute-santos:
Send-email WOH will look for a user in the user directory and use the user's email address if cc/to/bcc don't have a valid email address.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/send-email-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/send-email-woh.md
@@ -7,12 +7,14 @@ Description
 The EmailWorkflowOperationHandler invokes the SMTP Service to send an email with the parameters provided. It is useful
 to send email notifications when some operation(s) have been completed or some error(s) have occurred in a workflow.
 
-The email body, if not specified by body or body-template-file, will consist of a single line of the form: `<Recording
-Title> (<Mediapackage ID>)`.
+The email body, if not specified by body or body-template-file, will consist of a single line of the form:
+`<Recording Title> (<Mediapackage ID>)`.
 
 Freemarker templates can be used in the following fields to allow replacement with values obtained from the workflow or
-media package: to, subject, and body. If body-template-file is specified, the operation will use a Freemarker template
+media package: to, cc, bcc, subject, and body. If body-template-file is specified, the operation will use a Freemarker template
 file located in `<config_dir>/etc/email` to generate the email body.
+
+User names can be provided in `to`, `cc`, or `bcc` in lieu of email addresses so that the user directory is searched and that user's email address is used (see Example 5).
 
 
 Parameter Table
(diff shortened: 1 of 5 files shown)


modules/notification-workflowoperation/src/main/resources/OSGI-INF/operations/email.xml
```diff
@@ -11,4 +11,6 @@
   <reference name="EmailTemplateService" interface="org.opencastproject.email.template.api.EmailTemplateService" policy="static" cardinality="1..1" bind="setEmailTemplateService" />
   <reference name="ServiceRegistry" cardinality="1..1" interface="org.opencastproject.serviceregistry.api.ServiceRegistry"
     policy="static" bind="setServiceRegistry" />
+  <reference name="userDIrectoryService" cardinality="1..1" interface="org.opencastproject.security.api.UserDirectoryService"
```
staubesv: userD**I**rectoryService

modules/notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/notification/EmailWorkflowOperationHandler.java
```diff
@@ -123,6 +131,36 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+    String templateApplied = applyTemplateIfNecessary(workflowInstance, operation, configName);
+    if (templateApplied == null)
+      return null;
+
+    // Are these valid email addresses?
+    StringBuffer result = new StringBuffer();
+    for (String part : templateApplied.split(",|\\s")) {
+      result.append(result.length() > 0 ? "," : "");
+      try {
+        InternetAddress emailAddr = new InternetAddress(part);
```
staubesv: This code assumes that usernames don't have the format of email addresses while a lot of services suggest users to use their email addresses as usernames. Just tested that Opencast works with usernames that are valid email addresses - it does.

I suggest to just omit this whole try block, i.e. try to look up the user. If a user is found, use its email address if available. If no user is found, just use the plain string.

If a user is found but has no email address, it might be better to log an error (instead of taking the raw string).

What do you think?

staubesv (changes requested):
Hi @rute-santos 

I've done a first round of review and have the following findings:

- Typo (see comment)
- Usernames vs. email addresses (see comment)
- You should specify the dependency to the user directory in the pom too, shouldn't you?
- Please resolve the CI errors (see https://travis-ci.org/opencast/opencast/jobs/376081500)

After that, I will continue with the final tests.

modules/notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/notification/EmailWorkflowOperationHandler.java
rute-santos: Good point! I will make the changes.
In the case the user name is an email address and that user doesn't have an 'email' registered, I think we should return the raw string.

rute-santos:
Hi @staubesv,
I made all the changes you requested. Thanks for your comments!
Rute

staubesv (approved):

docs/guides/admin/docs/workflowoperationhandlers/send-email-woh.md
```diff
@@ -277,3 +280,38 @@ Logged incident of the error looks like this:
...
+
+```xml
+   <operation
+      id="send-email"
+      fail-on-error="false"
+      description="Notify user associated to this recording that it is ready to be trimmed">
+      <configurations>
+        <configuration key="to">${(catalogs['episode']['contributor'])}</configuration>
+        <configuration key="subject">Recording is ready for EDIT</configuration>
+        <configuration key="body-template-file">eventDetails</configuration>
```
staubesv, line 297: Hi @rute-santos In case you have a good 'eventDetails' template at hands, it would be great if you could contribute that, too (as a separate pull request)

PR #221, opened 2018-05-09
MH-12873, Speed up test builds

lkiesow:
This patch changes the `site` build by:

- Skipping tests since they are run in parallel anyway and they are
  unnecessary for generating JavaDocs.
- Skip PMD and CPD since we are not publishing these reports at the
  moment anyway. It can easily be turned on again once we decide to use
  those reports. For details, see
  https://maven.apache.org/plugins/maven-pmd-plugin/faq.html#Is_there_any_way_to_skip_the_PMD_or_CPD_reports_temporarily
- Aggregate site in preparation for publishing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,13 +82,20 @@ jobs:
         - mvn clean install -Pnone
 
     # Make a simple build to generate all Opencast artifacts while excluding as many checks as possible since we do a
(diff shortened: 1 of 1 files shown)


gregorydlogan:
We're not publishing the reports *yet*.  We will be shortly, but I can always turn that flag on in Buildbot.

ts23 (approved):

PR #224, opened 2018-05-11
MH-12878 Don't verify NPM cache to speed up build process

staubesv:


lkiesow:
Can you say a few words about this patches background? The [commit message](https://github.com/opencast/opencast/pull/224/commits/9bb895159f3c85cb2182f62a4a8dbc58d486b5d1) isn't very informative ;-P

staubesv:
Hi @lkiesow, see https://opencast.jira.com/browse/MH-12878 for more details.

lkiesow:
It's more like 7s instead of the claimed minute:

```bash
# without patch
% grep 'cache verify' pom.xml
                  <arguments>cache verify</arguments>
% repeat 5 mvn clean install -Pdev 2>&1 | grep 'Total time:'
[INFO] Total time: 01:46 min
[INFO] Total time: 01:45 min
[INFO] Total time: 01:46 min
[INFO] Total time: 01:46 min
[INFO] Total time: 01:45 min

# with patch
% grep 'cache verify' pom.xml
% repeat 5 mvn clean install -Pdev 2>&1 | grep 'Total time:'
[INFO] Total time: 01:38 min
[INFO] Total time: 01:39 min
[INFO] Total time: 01:38 min
[INFO] Total time: 01:39 min
[INFO] Total time: 01:38 min
```

Anyway, I've merged this. But in the future, please add a description to your commits. Especially to ones like this where one might ask what the reason for disabling this might be. The commit's title does not answer that question and there is no description at all. Having the explanation in a 3rd party system like Jira is better than having no explanation at all, but it's quite impractical to look up ticket descriptions for every git commit.

Note that you also benefit from adding a proper commit description since Github will automatically pick that up as pull request description for one-commit pull requests, sparing you from having to add another description here.

PR #225, opened 2018-05-11
MH-12864 Don't try to parse 'undefined' as JSON

staubesv:


staubesv:
Hi @ts23 While testing another pull requests, I've found a bug I have introduced with PR #215: If no license has been selected for the event, the Admin UI throws hundreds of exceptions when Event-Details->Metadata is open in read-only mode (seen in the browser console). As you reviewed PR #215, would you mind taking this review?

modules/admin-ui/src/main/webapp/scripts/shared/directives/editableDirective.js
```diff
@@ -53,8 +53,12 @@ angular.module('adminNg.directives')
                            result = value;
                        }
                     });
-                    if (scope.params.type === 'ordered_text') {
-                        result = JSON.parse(result)['label'];
+                    if ((scope.params.type === 'ordered_text') && angular.isDefined(result)) {
+                        try {
+                            result = JSON.parse(result)['label'];
+                        } catch (e) {
+                            console.warn("Could not find displayable value for " + result);
```
lkiesow, line 60: Instead of just catching all errors, would it not make sense to actually handle the `undefined` value instead?
This patch may hide other errors and would logs warnings for perfectly fine metadata values, wouldn't it?

staubesv: The bug this PR addresses is that `undefined` is a legal and displayable value but attempting to `JSON.parse` it causes an exception which breaks the app.

The additional `angular.isDefined()` check fixes this error (leaving `result` `undefined`).

In case that `result` is defined and the exception occurs, that would indicate either an application error or configuration error (`ordered_text` must be JSON and comes from a configurable list provider).
The try cause just prevents the string "{{ getDisplayableValue() }}" from being displayed in the Admin UI in case the method `getDisplayableValue()` throws an exception but instead logs a warnings in the browser console.

No, the patch does not hide other errors or log warnings for correct metadata (as `ordered_text` implies the value being `undefined` or valid JSON with field `label`).

gregorydlogan:
Shouldn't this go into r/5.x?

staubesv:
@gregorydlogan Yep, you're right. I will close this PR and create a new one against r/5.x

staubesv:
Hi @lkiesow I'm closing this PR in favour of PR #282. The later removes the try-catch clause as it is somewhat on the too defensive side. Feel free to grab the review of #282.

PR #226, opened 2018-05-11
MH-12879, Default location of paella configuration

lkiesow:
The paella player lacks any default configuration for the location of
its main configuration files and is thus failing if this key is not
explicitly set, even if the default configuration file is provided.

This patch adds a default to `${karaf.etc}/paella` which is already
being used for the default configuration file.

### DIFF
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -146,7 +146,10 @@ org.opencastproject.db.ddl.generation=true
 
 ######### Paella player configuration folder #########
 
-org.opencastproject.engage.paella.config.folder=${karaf.etc}/paella
+# Path to the Paella player configuration files. Note that the configuration file for each tenant needs to be in a
+# subfolder with the tenants identifier, e.g. …/etc/mh_default_org/config.json
+# Default: ${karaf.etc}/paella
+#org.opencastproject.engage.paella.config.folder=${karaf.etc}/paella
 
 
 ######### Workspace Cleanup #########
--- a/modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java
+++ b/modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java
@@ -24,13 +24,14 @@ package org.opencastproject.engage.paella;
 import org.opencastproject.rest.RestConstants;
 import org.opencastproject.security.api.Organization;
 import org.opencastproject.security.api.SecurityService;
+import org.opencastproject.util.ConfigurationException;
 import org.opencastproject.util.PathSupport;
 import org.opencastproject.util.doc.rest.RestQuery;
 import org.opencastproject.util.doc.rest.RestResponse;
 import org.opencastproject.util.doc.rest.RestService;
 
-
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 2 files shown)


etc/custom.properties
```diff
@@ -146,7 +146,10 @@ org.opencastproject.db.ddl.generation=true
 
 ######### Paella player configuration folder #########
 
-org.opencastproject.engage.paella.config.folder=${karaf.etc}/paella
+# Path to the Paella player configuration files. Note that the configuration file for each tenant needs to be in a
+# subfolder with the tenants identifier, e.g. …/etc/mh_default_org/config.json
```
staubesv, line 150: Hmm... is that multi-tenant stuff really true? I can't see the logic that locates the file in a subfolder <tenant identifier> of 'etc' ....

staubesv (changes requested):

modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java
```diff
@@ -79,9 +83,16 @@ protected void setSecurityService(SecurityService securityService) {
...
-
     paellaConfigFolder = cc.getBundleContext().getProperty(PAELLA_CONFIG_FOLDER_PROPERTY);
+
+    // Fall back to default location if necessary
+    if (StringUtils.isBlank(paellaConfigFolder)) {
+      paellaConfigFolder = cc.getBundleContext().getProperty("karaf.etc");
+      if (StringUtils.isBlank(paellaConfigFolder)) {
+        throw new ConfigurationException("Paella configuration not set and unable to fall back to default location");
+      }
+      paellaConfigFolder += "/paella";
```
staubesv: Hmm... Using a platform-independent method that uses the correct path separator would be preferable, wouldn't it?

etc/custom.properties
lkiesow: [I can](https://github.com/lkiesow/opencast/blob/46c6194fc84e423cb22dfe15f054e42593f9e295/modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java#L113) ;-P

modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java
lkiesow: This supports all platforms Opencast supports. But I can change the concatenation method if you want.

etc/custom.properties
staubesv: Oh, should have had a closer look to the code... thanks ;-)

modules/engage-paella-player/src/main/java/org/opencastproject/engage/paella/PaellaConfigRest.java
staubesv: Let's change it for sake of cleaner code ;-)

lkiesow: Interesting. For me, this is the simpler and cleaner variant. The alternative does make the code more complex. But will push a patch in a minute.

staubesv (approved):

PR #227, opened 2018-05-11
MH-12880, Remove redundant OSGI declarations

lkiesow:
This patch removes some redundant OSGI declarations and unused OSGI
methods from the Paella player module.

### DIFF
--- a/modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
+++ b/modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0"
- name="org.opencastproject.engage.paella.PaellaConfigRest" immediate="true" activate="activate" deactivate="deactivate">
+               name="org.opencastproject.engage.paella.PaellaConfigRest" immediate="true">
   <implementation class="org.opencastproject.engage.paella.PaellaConfigRest" />
   <property name="service.description" value="Paella config REST Endpoint" />
 
@@ -11,5 +11,5 @@
     <provide interface="org.opencastproject.engage.paella.PaellaConfigRest" />
   </service>
   <reference name="security-service" interface="org.opencastproject.security.api.SecurityService"
-    cardinality="1..1" policy="static" bind="setSecurityService" />
+             bind="setSecurityService" />
 </scr:component>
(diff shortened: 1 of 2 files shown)


modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0"
- name="org.opencastproject.engage.paella.PaellaConfigRest" immediate="true" activate="activate" deactivate="deactivate">
```
staubesv, line 3: But PaellaConfigRest does have an activate method... why have you removed "activate"?

modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
```diff
@@ -11,5 +11,5 @@
     <provide interface="org.opencastproject.engage.paella.PaellaConfigRest" />
   </service>
   <reference name="security-service" interface="org.opencastproject.security.api.SecurityService"
-    cardinality="1..1" policy="static" bind="setSecurityService" />
```
staubesv, line 14: Just wondering: Do we start to use the OSGI defaults (1..1 & static)? Just because those defaults seem to be explicitly set in many, many places (and it would then make more sense to have a single large PR removing all those explicitly set defaults).

staubesv (changes requested):

modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
lkiesow: Because it's `activate` and if there is an `activate` method you do not need to name it here explicitly. It will be picked up automatically.

modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
lkiesow: We have them set explicitly in many places that's true. But we also haven't set them explicitly in many cases ;-P

I've adjusted this since I was working on this anyway.

staubesv: Fine for me.

modules/engage-paella-player/src/main/resources/OSGI-INF/paella-config-file.xml
staubesv: Ok, thanks for the clarifications.

staubesv (approved):

PR #231, opened 2018-05-14
MH-12877: Add new modal to edit multiple scheduled events at once

pmiddend:
This PR adds a new modal dialog, accessible via Actions->"Start Task" in
the events table. It only works for events that are scheduled, and
allows editing of such events using a simple wizard. It’s possible to
modify some of the metadata attributes as well as the scheduling
information.

This addresses the following requirement:

As a organisation administrator, I want to be able to edit multiple scheduled events at once

Note that the “start date” is not editable, because it’s unclear what
that would do to multiple events. Also, you can only edit the start
time and the end time, not the duration. Again, this hopefully reduces
confusion.

Here are some screenshots:

A new action "Edit Scheduled Events" has been added to the Actions dropdown on the Events page:

screen shot 2018-05-15 at 10 13 56

The dialog can only be used to edit scheduled events. The first page of the wizard will check this and enables users to unselect archived events:

screen shot 2018-05-15 at 09 39 44

The second page of the wizard allows users to edit most of the scheduling and the most important metadata fields:

screen shot 2018-05-15 at 09 41 09

Finally, the first page provides an overview about what would be changed:

screen shot 2018-05-15 at 09 41 32

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -40,12 +40,15 @@ import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
 import static org.opencastproject.index.service.util.RestUtils.conflictJson;
 import static org.opencastproject.index.service.util.RestUtils.notFound;
+import static org.opencastproject.index.service.util.RestUtils.notFoundJson;
 import static org.opencastproject.index.service.util.RestUtils.okJson;
 import static org.opencastproject.index.service.util.RestUtils.okJsonList;
+import static org.opencastproject.index.service.util.RestUtils.serverErrorJson;
 import static org.opencastproject.util.DateTimeSupport.toUTC;
 import static org.opencastproject.util.RestUtil.R.badRequest;
 import static org.opencastproject.util.RestUtil.R.conflict;
 import static org.opencastproject.util.RestUtil.R.forbidden;
+import static org.opencastproject.util.RestUtil.R.noContent;
 import static org.opencastproject.util.RestUtil.R.notFound;
 import static org.opencastproject.util.RestUtil.R.ok;
 import static org.opencastproject.util.RestUtil.R.serverError;
@@ -56,6 +59,7 @@ import static org.opencastproject.util.doc.rest.RestParameter.Type.TEXT;
 import org.opencastproject.adminui.exception.JobEndpointException;
 import org.opencastproject.adminui.impl.AdminUIConfiguration;
 import org.opencastproject.adminui.impl.index.AdminUISearchIndex;
+import org.opencastproject.adminui.util.BulkUpdateUtil;
 import org.opencastproject.adminui.util.QueryPreprocessor;
 import org.opencastproject.authorization.xacml.manager.api.AclService;
 import org.opencastproject.authorization.xacml.manager.api.AclServiceException;
@@ -164,6 +168,7 @@ import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.text.ParseException;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
@@ -171,8 +176,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -212,9 +219,9 @@ public abstract class AbstractEventEndpoint {
   /**
    * Scheduling JSON keys
    */
-  private static final String SCHEDULING_AGENT_ID_KEY = "agentId";
-  private static final String SCHEDULING_START_KEY = "start";
-  private static final String SCHEDULING_END_KEY = "end";
+  public static final String SCHEDULING_AGENT_ID_KEY = "agentId";
+  public static final String SCHEDULING_START_KEY = "start";
+  public static final String SCHEDULING_END_KEY = "end";
   private static final String SCHEDULING_AGENT_CONFIGURATION_KEY = "agentConfiguration";
   private static final String SCHEDULING_OPT_OUT_KEY = "optOut";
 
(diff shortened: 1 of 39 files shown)


lkiesow:
I did not test it so far but a few things of this  seem weird:

- An action for one event should *never* modify multiple events. That's something you could add e.g. for series.
- Editing the schedule is already present in the event details. Having two places to go to is definitely a usability issue, especially if they do slightly different things.
- The way you edit scheduling data should preferably be the same across Opencast. In “Add Event” you can edit both duration and end time. here you explicitly block that. That's inconsistent and seems weird.

staubesv:
As SWITCH asked Plapadoo to implement this, here some comments from my side:

I've adapted the description and added some screenshot to make it obvious that we talk about Actions->"Edit scheduled events" (not about the Action button in the Events table).

The "Editing the schedule is already present in the event details" is true, but our large volume customers are quite annoyed and have complained about the number of clicks they need in case, for example, a room changes for a series of lectures. They have asked us to provide a more efficient solution, i.e. a way to edit multiple scheduled events at once...

... which turned out to be more difficult then we expected.

For example, when creating multiple events, the rrule is not made persistent and there is no way to recognise that multiple scheduled events were created based on a rrule (we would have liked to edit that rrule). Also, allowing to change the start date or end date would imply that editing scheduled events results in a different number of scheduled events (in case of multiple events). 

In theory, it would also be desirable to have the same functionality to edit metadata for archived events. But allowing the user to select mixed sets (scheduled & archived) of events would be quite hard if the user should understand what exactly he is actually doing - hence the limitation to scheduled events.

Anyway, this pull request is an attempt to get most out of our available time budget and therefore focuses quite a lot on the actual requirements reported by our users.

staubesv:
@pmiddend When doing the screenshots, I've realised that the tables on "Edit scheduled events"->Edit come with a non-standard layout. While I actually do like its look, please make those tables look like all other tables (no white border within the table) for sake of design consistency.

staubesv:
I've created [MH-12884 As a user, I expect a consistent representation of the event status](https://opencast.jira.com/browse/MH-12884) as the same problem (different title and values for event status) also exists in the Actions->"Stark Task" wizard.

pmiddend:
Note that I added two commits to this PR to fix MH-12884, too. It would be a bit of a hassle to make this a separate PR, since I added a new CSS class, so there would be a dependency between the PRs. Hope that's okay.

CGreweling:
@lkiesow  I think this functionality is at the right place. The confusion comes with the usability of the admin ui, not with this feature i think.
This has nothing todo with series, because there is not necessary a Series for events. 
I` have nothing to say against this feature, love to see it back implemented as it was there in MH 1.6.
Will check the new standard layout you just pushed.

CGreweling:
@pmiddend  found some things during testing. Will look at the code tomorrow.

Editing-TAB: Editing times is disabled during check for Schedule conflicts =>but not the next-button.

I always get the Fail message: "event could not be saved". But all changes are correctly done.

No translation for "Edit Scheduled events".
No translation for "TABS: general ,edit, summary "

If you edit the title it would be nice to add a counter to the title.
Otherwise it is not possible to identify the different events.

staubesv:
@CGreweling I disagree on the "add a counter to the title": That way, the user would need to edit the individual titles for each event again as it is not necessarily the goal to have the titles numbered. This modal should, however, prevent exactly that: Having to edit events individually.
Finally, the events are distinguishable by the start and end date.

pmiddend:
I added German translations now. I hope that's enough languages for now, my Spanish is a little rusty. ;-)

I also disabled the "Next" button while checking conflicts - good catch! Sadly, I couldn't reproduce the issue with the error notification as of yet.

staubesv:
@pmiddend I've found some other issues:

1. Editing does not work

Reproduction:
1. Select a few scheduled events and an uploaded event in the events table
2. Open Actions->"Edit scheduled events"
3. Unselect the uploaded event in the tab "General" to be able to proceed
4. Change start time in tab "Edit"

Problem: In the Summary page, there is now a message saying "No changes detected"

Another small issue:

Please change the title of the column indicating the event status on tab "General" from "Progress" to "Status" (idea: Same as in events table)

staubesv:
@pmiddend And another issue: The start time entered on tab "Edit" is not the same as the start time entered displayed in the tab "Edit". Note that the user always expects to enter the time in local time (the summary page should also display the local time).

staubesv:
@CGreweling The reason you see the translation keys in languages other than enUS is that those keys are completely missing (not just untranslated) in the other languages. That will be solved once en-US has been uploaded to Crowdin and the translations are downloaded again to the Opencast repo: Crowdin will add the translation keys to all tanslations even if the keys have not been translated.

CGreweling:
- Adding counters to the title was only my personal point of view, i have no strong feelings about that. It is oky as it is. I should have expressed this more clearly.
- Thanks for the translations, an example is enough, just wanted to be sure that it can be translated.
Code looks good to me, besides some minor hints:
 - there isn't any javadoc would be nice to have some
 - is it possible to get a test ?

And something for future development.
The bulk edit functionality is implemented in the admin ui, if i get that correctly.
It would have been a better idea to implement this in the scheduler itself. with a rest-end point.
So this could be easier used from other parts of opencast in the future. No it is only a feature of the admin ui...
Anyway; good work, i know the whole date time thing is complicated ;)

pmiddend:
Okay, so we added Javadoc and tests for the backend. I also included some tests for the frontend. The issues @staubesv reported are also fixed now.

CGreweling:
@pmiddend are you working on the issues  @staubesv found? (editing and Local time) or already fixed them? ;)

pmiddend:
@CGreweling Those should be fixed now.

lkiesow:
Note that in the section “Weekday**s**” you can select only one day and it's highly confusing that you cannot select a day but you just select a random Weekday and you have no idea to which one the event will be changed to.

staubesv:
The "weekday" stuff cannot be solved in general.

This already starts with the idea of having one rrule to scheduled a lecture and allowing multiple weekdays to be set: Too bad if the lectures do not start at the same time on the different weekdays ;-)

Besides it is not really that useful to select multiple weekdays from that perspective, the rrule that was used to generate the events is not persistent, i.e. we don't know how the events where created. Therefore, I don't see what scheduled multiple events would do if a user selected multiple weekdays.

Even if we had the rrule, what would we do if the user selected 10 events that were generated by a rrule but later had edited one of those 10 events individually (e.g. changed weekday)?

... also, the user can selected events that have been generated by different rrules.

What the weekday selection does: 

- If all selected events start on the same weekday, it will display that weekday. In this case, we indeed can change the weekday of all events. That is a valid use case.
- If not all selected events start on the same weekday, no weekday is pre-selected. If the user selects a weekday, all events will be shifted to that weekday within the respective week. That is not a valid use case ;-)

Maybe we could just disable the weekday selection in case not all selected events start at the same day as changing the day in this situation seems not really useful.

What do you think?

Update: It might be an option to display the selected events grouped by weekdays and then allow the user to edit the events per weekday. But we can't change this right now from our side.

lkiesow:
My first point was that the label is not correct. It should be `weekday` not `weekdays` ;-P

<strike>Anyway, your description does not seem to be correct. Testing the feature, the event I selected was shifted to the week after the currently selected one and was not shifted within the week which would not have been possible since that would have been in the past.</strike> In any case, that behavior is just not intuitive. Basically, you cannot use this without reading how it works (or figuring that out by yourself) in advance.

There is also an additional bug where the tool in some cases [refuses to edit scheduled recordings
](https://opencast.jira.com/browse/MH-13004)

*EDIT:* I'll take back what I said earlier. It seems consistent now. Maybe I did something wrong. The week always seems to stay the same. But the current behavior sets scheduled dates to the past which can then not be editted anymore. So you cannot take back your incorrect changes.

PR #233, opened 2018-05-15
MH-12828 re-enable Scheduler service conflicts json REST endpoint

karendolan:
This pull reverts the removal of the Scheduler Service REST getConflictingEventsJson(). 
It is a companion to Scheduler Service REST's getConflictingEventsXml().

Note: This pull does not address the known TimeZone issues, so junit tests still require the local system to be in UTC or -UTC.

### DIFF
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
@@ -1091,7 +1091,7 @@ public class SchedulerRestService {
                   @RestResponse(responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR, description = "A detailed stack track of the internal issue.")})
   public Response getConflictingEventsJson(@QueryParam("agent") String device, @QueryParam("rrule") String rrule,
           @QueryParam("start") Long startDate, @QueryParam("end") Long endDate, @QueryParam("duration") Long duration,
-          @QueryParam("timezone") String timezone) {
+          @QueryParam("timezone") String timezone) throws UnauthorizedException {
     try {
       List<MediaPackage> events = getConflictingEvents(device, rrule, startDate, endDate, duration, timezone);
       if (!events.isEmpty()) {
@@ -1103,7 +1103,7 @@ public class SchedulerRestService {
     } catch (IllegalArgumentException e) {
       return Response.status(Status.BAD_REQUEST).build();
     } catch (UnauthorizedException e) {
-      return Response.status(Status.UNAUTHORIZED).build();
+      throw e;
     } catch (Exception e) {
       logger.error("Unable to find conflicting events for {}, {}, {}, {}, {}: {}",
               device, rrule, startDate, endDate, duration, getStackTrace(e));
@@ -1129,7 +1129,7 @@ public class SchedulerRestService {
 
   public Response getConflictingEventsXml(@QueryParam("agent") String device, @QueryParam("rrule") String rrule,
           @QueryParam("start") Long startDate, @QueryParam("end") Long endDate, @QueryParam("duration") Long duration,
-          @QueryParam("timezone") String timezone) {
+          @QueryParam("timezone") String timezone) throws UnauthorizedException {
     try {
       List<MediaPackage> events = getConflictingEvents(device, rrule, startDate, endDate, duration, timezone);
       if (!events.isEmpty()) {
@@ -1140,7 +1140,7 @@ public class SchedulerRestService {
     } catch (IllegalArgumentException e) {
       return Response.status(Status.BAD_REQUEST).build();
     } catch (UnauthorizedException e) {
-      return Response.status(Status.UNAUTHORIZED).build();
+      throw e;
     } catch (Exception e) {
       logger.error("Unable to find conflicting events for {}, {}, {}, {}, {}: {}",
               device, rrule, startDate, endDate, duration, getStackTrace(e));
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
@@ -100,6 +100,9 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
 import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1071,6 +1075,40 @@ public Response getEventsAsXml(@QueryParam("agent") String device, @QueryParam("
       throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
     }
   }
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("conflicts.json")
+  @RestQuery(name = "conflictingrecordingsasxml", description = "Searches for conflicting recordings based on parameters", returnDescription = "Returns NO CONTENT if no recordings are in conflict within specified period or list of conflicting recordings in XML", restParameters = {
```
staubesv: Too many occurrences of the string **xml** ;-)

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1743,6 +1748,64 @@ public Response prolongCapture(@PathParam("agent") String agentId) throws NotFou
     }
   }
 
+  /**
+   * Utility for getConflictingEventsJson and getConflictingEventsXml
```
staubesv: Both the description of the method and its parameters are not providing any useful information. Please either provide useful content here or just remove the whole comment. Since it is a private method, I recommend the latter.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1793,4 +1852,62 @@ private Properties parseProperties(String serializedProperties) throws IOExcepti
...
+   *        "attachments": "",
+   *        "id": "mp1",
+   *        "media": "",
+   *        "publications": ""
+   *     },
+   *    {
+   *        "metadata": "",
+   *        "attachments": "",
+   *        "id": "mp2",
+   *         "media": "",
```
staubesv: indentation

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1793,4 +1852,62 @@ private Properties parseProperties(String serializedProperties) throws IOExcepti
...
+   *         "publications": ""
+   *     }
+   * ],
+   *"totalCount": "2"
```
staubesv: indentation

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1793,4 +1852,62 @@ private Properties parseProperties(String serializedProperties) throws IOExcepti
...
+   *}
+   *
+   * @return serialized array as json array string
+   * @throws SchedulerException 
+   * @throws ParseException
+   */
+  @SuppressWarnings("unchecked")
```
staubesv: What warning needs to be suppressed?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1793,4 +1852,62 @@ private Properties parseProperties(String serializedProperties) throws IOExcepti
...
+   *        "metadata": "",
+   *        "attachments": "",
+   *        "id": "mp2",
+   *         "media": "",
+   *         "publications": ""
+   *     }
+   * ],
+   *"totalCount": "2"
+   *}
+   *
```
staubesv: @param mpList is missing

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1793,4 +1852,62 @@ private Properties parseProperties(String serializedProperties) throws IOExcepti
...
+    jsonObj.put("events", jsonArray);
+    return jsonObj.toJSONString();
+  }
+
+  /**
+   * Generates sample Dublin core.
+   *
+   * @return sample Dublin core
+   */
+  public String getSampleDublinCore() {
```
staubesv: Can you please provide a reason why you want this method (in its comment) or remove it?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1071,6 +1075,40 @@ public Response getEventsAsXml(@QueryParam("agent") String device, @QueryParam("
...
+          @QueryParam("timezone") String timezone) throws UnauthorizedException {
+    try {
+      List<MediaPackage> events = getConflictingEvents(device, rrule, startDate, endDate, duration, timezone);
+      if (!events.isEmpty()) {
+        String eventsJsonString = getEventListAsJsonString(events);
+        return Response.ok(eventsJsonString).build();
+      } else {
+        return Response.noContent().build();
+      }
+    } catch (IllegalArgumentException iae) {
```
staubesv: Suggestion: As you are not using the variable `iae` anyway, `e` would be a more common choice.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1088,53 +1126,20 @@ public Response getEventsAsXml(@QueryParam("agent") String device, @QueryParam("
...
-      } else {
-        events = service.findConflictingEvents(device, start, end);
-      }
+      List<MediaPackage> events = getConflictingEvents(device, rrule, startDate, endDate, duration, timezone);
       if (!events.isEmpty()) {
         return Response.ok(MediaPackageParser.getArrayAsXml(events)).build();
       } else {
         return Response.noContent().build();
       }
+    } catch (IllegalArgumentException iae) {
```
staubesv: Suggestion: As you are not used the variable `iae` anyway, `e` would be a more common choice.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1088,53 +1126,20 @@ public Response getEventsAsXml(@QueryParam("agent") String device, @QueryParam("
...
+      return Response.status(Status.BAD_REQUEST).build();
     } catch (UnauthorizedException e) {
-      throw e;
+      return Response.status(Status.UNAUTHORIZED).build();
```
staubesv: Don't get me wrong: I think responding with Status.UNAUTHORIZED is preferable, but...

... this is a change of the interface, isn't it?

Also: Would this method then still need to declare that it might raise an UnauthorizedException on line 1128?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -32,6 +32,7 @@
 import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
 import static org.opencastproject.capture.CaptureParameters.AGENT_REGISTRATION_TYPE;
 import static org.opencastproject.capture.CaptureParameters.AGENT_REGISTRATION_TYPE_ADHOC;
+import static org.opencastproject.mediapackage.MediaPackageElement.Type.Catalog;
```
staubesv: Your additions seem not to require this import and line 50 already imports a type of the same name (` import org.opencastproject.mediapackage.Catalog;`)

Why have you added this import?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1760,14 +1823,10 @@ private MediaPackage addCatalog(Workspace workspace, InputStream in, String file
...
     // Update comments catalog
-    try {
-      URI catalogUrl = workspace.put(mediaPackage.getIdentifier().compact(), c.getIdentifier(), fileName, in);
-      c.setURI(catalogUrl);
-      // setting the URI to a new source so the checksum will most like be invalid
-      c.setChecksum(null);
-    } finally {
-      IOUtils.closeQuietly(in);
-    }
+    URI catalogUrl = workspace.put(mediaPackage.getIdentifier().compact(), c.getIdentifier(), fileName, in);
```
staubesv: Not familiar enough with Java, but while I would expect the garbage collector to close the InputStream `in` at some point of time, it might by preferable to do it once it is not required anymore (although the try block already misses the in.close()).

Or does workspace.put close the stream?

If you have time for a small improvement: There are two callers of this private method. Both open the InputStream without having a reference to it themselves:

`   mediaPackage = addCatalog(workspace, IOUtils.toInputStream(eventCatalog.toXmlString(), "UTF-8"),
                "dublincore.xml", MediaPackageElements.EPISODE, mediaPackage);`

It would be nicer if this method took the DublinCoreCatalog `eventCatalog` as argument. In that case, the caller could do:

`   mediaPackage = addCatalog(workspace, eventCatalog, mediaPackage);`

Idea: Open and close InputStream in same method (the arguments "dublincore.xml" and  "MediaPackageElements.EPISODE" are not relevant for the caller and could be hard-coded in this method instead).

What do you think?

staubesv (changes requested):
Hi @karendolan 

I've just finished the first round of the review. Besides the build fails because of mvn site, I have some comments of mostly cosmetic nature.

The only thing I'm not sure about is the change of the interface (although I prefer your version).

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
lkiesow: Adding to Sven's comment here: If we really need this, can we not load the dublin core catalog as resource?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: Wow, sorry about that.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: The org.json.simple.JSONObject and org.json.simple.JSONOArray have unchecked warnings with put and add because they don't support generics. I removed the @SuppressWarnings("unchecked") so that it is obvious where the warnings are.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: The comment is removed from this private method to avoid clutter.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: I removed the jSON example all together from the comment to remove confusion and clutter.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: That method is an artifact, it is not needed and is removed.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: changed to "e".

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: I agree with your hesitation on changing this API. I notice that some of the other endpoints do a "throw e". So, maybe it would be better not to change it.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
karendolan: That import must be an artifact but I'm surprised it didn't throw an "unused import" checkstyle error. I've removed that import.

staubesv:
Hi @karendolan, thanks for resolving the comments - the PR is much more concise now. I'm continuing the review...

staubesv:
@lkiesow & @gregorydlogan I've ran mvn site locally to verify that the Travis CI tests will work.

staubesv (approved):

PR #236, opened 2018-05-16
MH-12889 Intuitive Merging of Video Segments

krinnewitz:
This patch changes the behavior when removing segments in the video
editor. Previously, the free space resulting from removing a segment was
taken up by extending the previous segment. Now, the subsequent segment
is extended instead, which leads to a more intuitive behavior. This is
especially the case when editing a video from left to right (from start
to end), which should be the most common use case.

For the last segment, the behavior is different: it has to be merged
with the previous segment since a subsequent one doesn't exist.

This very subtle change is based on two assumptions:

- Western culture people tend to read from left to right
- The icon 'X' for segment deletion is very close to the right boundary of the segment. It seems more expected that that boundary is removed (instead the left boundary of the segment which is not close to the 'X' button)

So that means:

`|  segment 1 ---- X1 | segment 2 ---- X2 | segment 3 ----  X3|`

Pressing the 'X2' on segment 2 previously resulted in...

`|  segment 1 ------------------------X1 | segment 3 ----  X3|`

So hitting 'X2' removed the segment boundary just next to 'X1'.

This now results in ...

`|  segment 1 ---- X1 | segment 3 --------------------------  X3|`

Update: June 15th, 2018
Just tested this on our staging environment. It indeed seems more intuitive ;-)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/video/editController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/video/editController.js
@@ -57,17 +57,17 @@ angular.module('adminNg.controllers')
         };
 
         $scope.clearSelectedSegment = function () {
-            
+
             angular.forEach($scope.video.segments, function (segment) {
                 if (segment.selected) {
-                    
+
                     var index = $scope.video.segments.indexOf(segment);
 
-                    if ($scope.video.segments[index - 1]) {
-                        $scope.video.segments[index - 1].end = segment.end;
-                        $scope.video.segments.splice(index, 1);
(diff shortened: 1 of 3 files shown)


lkiesow:
@slampunk, @rrolf, @CGreweling, @corne-oosthuizen you have worked on this and/or originally implemented this differently. I'm fine with this general change, are you? (I also do not have to work a lot with the editor)

lkiesow (approved):
Works as expected. No complaints. If no one else has in the next couple of days, I will merge it like this.

TurRil:
Hi,

I'm fine with this change - makes more sense :)

Thanks,
CO

lkiesow:
No one is complaining → merged

PR #218, opened 2018-05-07
MH-12829 Switch and rename event details tabs

KatrinIhler:
This PR switches the metadata and general tabs in the event details modal so that the metadata tab is in front. This accomplishes two things: 1. It's consistent with the series details modal, and 2. it makes more sense in my opinion to view the metadata first.

Additionally the general tab is renamed to publications tab since it contains publications and nothing else. For consistency a number of files and identifiers are also renamed. Because of naming conflicts (see discussion on list here) I thought it best to rename the resources for the assets tab to contain 'Asset' in their name, thus avoiding conflicts concerning 'eventPublicationsResource'. (I hope I didn't miss anything during my renaming spree, if I did, just comment and I'll fix it.)

See the screenshot for more details: 

screenshot from 2018-05-07 16-46-19

_This work is sponsored by ETH._

PR #237, opened 2018-05-16
MH-12829 Fix broken sub tabs of Event Details->Assets

staubesv:
The sub tabs of Event Details->Assets don't work anymore because the sub tab names were not adjusted by MH-12829.

Reproduction:
1. Open Event details->Assets
2. Try to open any sub tab (Attachments, Catalogs, ...)

-> Does not work anymore.

Seems I was a bit sloppy when reviewing #218... sorry for that!


KatrinIhler:
Whoops, that somehow slipped through my tests... but this should fix it!

staubesv:
Hi @wsmirnow This bug is very annoying when testing other pull requests where it would be helpful to see what assets are available. Could you please take a look at it soon?

wsmirnow:
Works as expected. Thank you @staubesv.

PR #238, opened 2018-05-16
MH-12892 Added ability to configure the job load for the transcription service.

rute-santos:


etc/org.opencastproject.transcription.ibmwatson.IBMWatsonTranscriptionService.cfg
```diff
@@ -37,3 +37,6 @@ ibm.watson.password=<SERVICE_PSW>
 # Email to send notifications of errors. If not entered, the value from
 # org.opencastproject.admin.email in custom.properties will be used.
 #notification.email=
+
+# Start transcription job load
+job.load.start.transcription=0.1
```
lkiesow: Since this is set to the default, can we state the default and comment out the setting. That way it's clear that the default value is used. A lot of other configuration keys use this format already. It helps quickly identify values a user has modified:

```
# Start transcription job load
# Default: 0.1
￼#job.load.start.transcription=0.1
```

lkiesow (commented):
Looks good. I've added one small suggestion. I'll try to actually test it tomorrow.

etc/org.opencastproject.transcription.ibmwatson.IBMWatsonTranscriptionService.cfg
rute-santos: Done! Also for MH-12893.

lkiesow:
Thanks for the change. Merged.

PR #239, opened 2018-05-16
MH-12893 Added ability to configure the job load for the aws s3 distribution service.

rute-santos:


modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -108,6 +109,29 @@
   // config.properties
   public static final String OPENCAST_DOWNLOAD_URL = "org.opencastproject.download.url";
 
+  /** The load on the system introduced by creating a distribute job */
+  public static final float DEFAULT_DISTRIBUTE_JOB_LOAD = 1.0f;
```
lkiesow: Should the default distribution job load actually be that high. 1 is the same as an enocde job iirc.

etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
```diff
@@ -24,3 +24,7 @@ org.opencastproject.distribution.aws.s3.distribution.enable=false
 # The AWS secret access key to use for connecting to S3.
 #org.opencastproject.distribution.aws.s3.secret.key=
 
+# Job loads
+job.load.aws.s3.distribute=1.0
```
lkiesow: Can we put these in the same form I suggested for your other pull request` For example:

```
# Description of what job this load is for.
# Default: 1.0
#job.load.aws.s3.distribute=1.0
```

lkiesow (commented):
Overall, this looks good.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
rute-santos: The download distribution service has 1.0 as the distribute job load in its configuration  (which in my opinion is too much so maybe a bug?).
The S3 distribution is different from the download distribution because it actually copies files to S3 as opposed to linking them in the file system. I picked 1.0 based on that, but agree that it may be too much. Suggestions? It's just a default :)

lkiesow: The download distribution job loads do not even match up in code and configuration. I created pull request #244 to fix that.

What about 0.4 for this? Given that 1 means full usage of one CPU core, I would expect a single threaded machine to handle a bit more than two of these jobs in parallel. Of course, this is picked by me with no actual testing done. Hence, feel free to go higher or lower ;-)

rute-santos: Hi @lkiesow, I did some tests yesterday and the job didn't produce as much load as I expected so I think it's ok to match to the download distribution values. I am going to set the default to 0.2 and 0.1.

lkiesow:
Thanks for testing the job loads. Merged.

PR #244, opened 2018-05-17
MH-12898, Fix download distribution job load defaults

lkiesow:
This patch fixes the default values for the download distribution job
loads. It increases the default value for distribution jobs since they
can be slightly more expensive due to file transfers though it does not
go so far as to raise it all the way up to one like the configuration
file did.

The retract job load is lowered to 0.1 since it is a very simple job
after all and a system should be able to handle multiple retract jobs in
parallel.


staubesv (approved):

rute-santos:
Looks ok to me!

PR #246, opened 2018-05-17
MH-12900, Fix search service job loads

lkiesow:
This patch fixes the search service job load defaults setting them to
the values previously defined in the configuration file.


staubesv (approved):

rute-santos:
Looks ok to me!

PR #247, opened 2018-05-17
MH-12901, Fix YouTube publication job loads

lkiesow:
This patch adjusts the YouTube publication job load defaults. Retracting
should be very inexpensive as already defined in the configuration.
Publishing should be more expensive but still less expensive than tasks
like encoding.


staubesv (approved):

staubesv:
Hi @lkiesow All the job load PRs are fine. I just wait with merging so that others have a chance to comment the new job loads.

PR #248, opened 2018-05-18
MH-12902, closing videoeditor should continue in events list

doofy:
This patch adds a resume feature to admin ui events module

Before this patch every way you could leave the video editor and go to the events list got you back to the beginngn of the list. With this patch the close button brings to the page of the events list where you left off from.

This work was sponsored by SWITCH


KatrinIhler (approved):
Seems to do exactly what it's supposed to do and sounds reasonable to me.

lkiesow:
Merged (trusting it @KatrinIhler's review)

PR #254, opened 2018-05-22
MH-12910: Put temporary grunt files into the target folder

JulianKniephoff:
This moves some temporary files grunt, or more specifically the usemin plugin, creates during the admin UI build into the `target` folder, which is ignored in `.gitignore`.

This way, `git` should no longer show untracked files after switching between branches with different module naming schemes (`matterhorn-*` vs. no prefix), at least as soon as #216 is merged and propagated through all the releases. :wink: 

Note that previously, `.tmp` and `coverage` were ignored in a directory local `.gitignore`. This is not sufficient, because that `.gitignore` vanishes, too, in a branch switch as described above. The alternative solution would be to ignore them in the (or a more) global `.gitignore`, but I preferred to collect all (or most, I guess; I didn't check thoroughly. Let's say "more" :wink:) of grunts artifacts in one location.

---

**Note** for those confused about getting notified about this PR twice: I got the branch name wrong and had to reraise the PR. :sweat_smile: 

I'm sorry.

### DIFF
--- a/modules/matterhorn-admin-ui-ng/.gitignore
+++ b/modules/matterhorn-admin-ui-ng/.gitignore
@@ -1,6 +1,5 @@
 /node_modules
 /dist
-/.tmp
 /.sass-cache
 www
 .jshintrc
@@ -8,7 +7,6 @@ deploy.sh
 proxy.sh
 *.swo
 bin
-coverage
 src/test/resources/*.jar
 src/test/resources/chromedriver
 .agignore
--- a/modules/matterhorn-admin-ui-ng/Gruntfile.js
+++ b/modules/matterhorn-admin-ui-ng/Gruntfile.js
@@ -23,7 +23,8 @@ module.exports = function (grunt) {
   // Configurable paths for the application
   var appConfig = {
     app: require('./bower.json').appPath,
-    dist: 'target/grunt/webapp'
+    dist: 'target/grunt/webapp',
+    staging: 'target/grunt/.tmp'
   };
 
   // Define the configuration for all the tasks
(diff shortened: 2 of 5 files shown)


staubesv:
Hi @JulianKniephoff and @gregorydlogan Just lost quite some time to find out that this patch now has rippled up to 5.x & develop and breaks the Admin UI development environment. Could you please either a) revert this patch or b) provide a follow-up patch that makes the Admin UI dev environment work again (at least in 5.x / develop)?

BTW: This patch doesn't fix anything related to operations. I therefore recommend to add this kind of patches to develop in the future to avoid unnecessary work.

JulianKniephoff:
@staubesv can you provide more detail on the breakage? The only problems I could imagine could be due to left over files in the old locations, due to switching from/to a branch that doesn't have the change, yet. Everything should work fine with a clean environment. It does for me at least. :sweat_smile:

staubesv:
cd modules/admin-ui
grunt proxy --proxy.host=http://localhost:8080 --proxy.username=opencast_system_account --proxy.password=CHANGE_ME

... does not work anymore.

JulianKniephoff:
And I presume `git status` shows you no unexpected untracked files and you build everything cleanly from scratch? :thinking:

staubesv:
Yep

JulianKniephoff:
Could you try clearing out your browser data for the site? Mainly cache and localStorage would suffice, I guess.

staubesv:
@JulianKniephoff It doesn't work anymore because it cannot load the CSS as you have changed its location.

JulianKniephoff:
Well, technically I changed the location of a temporary file, and I thought I also changed all places that would look for this file for further processing, but I just found one that I overlooked.

Mysteriously it works for me without that change, but maybe that's some sort of caching effect on **my** site; it **is** a mistake nonetheless and a follow-up PR will arrive shortly! Maybe @staubesv could review and merge it? :wink:

Sorry for the inconvenience! :cold_sweat:

PR #255, opened 2018-05-22
MH-12868: Make frame-by-frame skipping function in the editor use the "actual" framerate

JulianKniephoff:
This makes the admin UI tools endpoint return the framerate of the preview videos used for editing and previewing. This is then in turn used by the frontend to make the left and right key actually skip the correct amount of time (= 1 / framerate).

The assumption here is, of course, that the preview video has the same framerate as the end product. Otherwise, a skip of `1 / preview_fps` is meaningless.

Note also that seeking `1 / fps` seconds is kind of a hack, but there is currently no standard (also also not really any non-standard, barring [this](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/seekToNextFrame)) way of controlling a video element in a frame-precise manner.


lkiesow:
A patch like this shouldn't go into 3.x anymore, should it?

gregorydlogan:
I could see the argument that the scrubber is inaccurate, therefore this is a bug fix.  But I'd feel more comfortable punting this into 5.x or develop so that more eyes see it.  4.x if we have to, but 5.x is where QA is going to be focused.

JulianKniephoff:
@lkiesow I wasn't sure, but my thinking was along the lines of what @gregorydlogan said. I have no strong opinion on this, though; will retarget today. :smiley:

JulianKniephoff:
Done, btw. Just in case GitHub does not notify in the case of a redirect and/or push.

I can never remember what GitHub sends notifications for. :sweat_smile:

gregorydlogan:
As is this dumps files in admin-ui/.tmp and admin-ui/coverage.  I also can't see to get the admin UI to work right, but then that's true regardless of whether this patch is included or not.

gregorydlogan (changes requested):
Works as expected, aside from the junk left in the module directory.  Fix that and we're ready to go.

JulianKniephoff:
@gregorydlogan not sure what you mean. It builds cleanly for me, with no untracked files left behind. Maybe your's were leftovers from some previous build on a different branch, where they were ignored or something? (Which somewhat ironically is exactly the scenario I wanted to fix with #253.)

I just rebased the branch on the current `r/5.x` and pushed that, maybe that helps somehow? :thinking:

gregorydlogan:
Fair enough, double checked and no longer see the issue.  Merged!

PR #256, opened 2018-05-22
Please ignore

phoebem:
…e, process-smil

Lossless concat for multipart chunked video source(same codec)
Demux to deal with multipart files (epiphan)
Multiencode to encode multiple files by encoding profile in one operation
Process-smiltrack to concatenate trimmed clips and encode to multiple delivery formats by encoding profile in one operation


phoebem:
Only used for commparison

PR #208, opened 2018-04-27
MH-12814 Manually Select And Upload Thumbnails

krinnewitz:
This patch adds a new tab "Thumbnail" to the video editor which provides
the following new functionality:

- Upload a new thumbnail
- Go back to the default thumbnail (extracted at fixed position)
- Extract a thumbnail from the video at the current playback position
- Show a preview of the thumbnail

All involved flavors and tags are configurable. The External API and
OAI-PMH channels are automatically updated when the thumbnail changes.
To update other channels, you have to use the 'publish' workflow.
Dual-stream videos are supported as well. However, the default thumbnail
is only created for one of the tracks.

To update the external API and OAI-PMH on-the-fly, new code was added to
replace media package elements without retracting and republishing the
whole media package.

This first version is a bit slow since it waits for all channels to be
updated. This may not be necessary and could probably be optimized further.

There already was an "Asset Upload" mechanism implemented in Opencast,
which already allowed to upload a new thumbnail. This was left untouched
and should still work.

Here are some screenshots of the new "Thumbnail" tab in the video editor:

Default Thumbnail in use:

thumbnails_default

Extracted thumbnail (single track):

thumbnails_from_single_track

Extracted thumbnail (dual track)

thumbnails_from_dual_track

Please note: This PR was rebased to include PR #193 since it is needed to access some thumbnail-related settings when running workflows. Please discuss any code regarding PR #193 by commenting on PR #193  instead of this one.

Also, due to a bug in develop, you won't be able to test publishing to engage, unless you manually merge-in PR #195.

This work is sponsored by SWITCH.

PR #259, opened 2018-05-23
MH-12916, Admin Interface Configuration Defaults

lkiesow:
This patch fixes the default configuration of the administrative user
interface.

### DIFF
--- a/etc/org.opencastproject.adminui.cfg
+++ b/etc/org.opencastproject.adminui.cfg
@@ -8,7 +8,8 @@
 # like 'presenter/preview', therefore 'preview' is the
 # recommended default value.
 #
-preview.subtype=preview
+# Default: preview
+#preview.subtype=preview
 
 # Flavor subtype that identifies waveform elements.
 #
@@ -16,7 +17,8 @@ preview.subtype=preview
 # like 'presenter/waveform', therefore 'waveform' is the
 # recommended default value.
 #
-waveform.subtype=waveform
+# Default: waveform
+#waveform.subtype=waveform
 
 # Flavor that identifies the existence of a smil cutting file.
 # Note that only recordings with such an element shows the
@@ -28,16 +30,17 @@ waveform.subtype=waveform
 # Has to be the same as <configuration key="target-smil-flavor">smil/cutting</configuration>
 # within a Workflow editor operation.
 #
-smil.catalog.flavor=smil/cutting
+# Default: smil/cutting
+#smil.catalog.flavor=smil/cutting
 
 # Tags which are set to the generated smil cutting catalog
-# by the video editor.
+# by the video editor. The value is a comma separated list.
 #
 # The default workflows will archive artifacts with the tag
-# "archive", therefore "archive" is the recommended
-# default value.
+# "archive", therefore "archive" is recommended.
 #
-smil.catalog.tags=archive
+# Default: archive
+#smil.catalog.tags=archive
 
 # Flavor that identifies the existence of silence catalogs. Note that
 # only recordings with such elements shows silence detection segments
@@ -47,4 +50,5 @@ smil.catalog.tags=archive
 # like 'presenter/silence', therefore '*/silence' is the
 # recommended default value.
 #
-smil.silence.flavor=*/silence
+# Default: */silence
+#smil.silence.flavor=*/silence
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/impl/AdminUIConfiguration.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/impl/AdminUIConfiguration.java
@@ -46,7 +46,7 @@ public class AdminUIConfiguration implements ManagedService {
 
   private String previewSubtype = "preview";
   private String waveformSubtype = "waveform";
-  private Set<String> smilCatalogTagSet = new HashSet<String>();
+  private Set<String> smilCatalogTagSet = new HashSet<>();
   private MediaPackageElementFlavor smilCatalogFlavor = new MediaPackageElementFlavor("smil", "cutting");
   private MediaPackageElementFlavor smilSilenceFlavor = new MediaPackageElementFlavor("*", "silence");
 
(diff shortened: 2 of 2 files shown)


staubesv (approved):

staubesv:
Hi @lkiesow 

I was just about to adapt PR #208 to this change and found the following problem:
When changing the configuration at runtime, changing the log level to DEBUG (as this PR does) means that Opencast will not provide any feedback about the configuration change in the log...

.... so the admin doing the configuration change has to hope that Opencast actually did somehow react as he does not get any confirmation about the configuration being successfully being reloaded.

Also, it is not entirely clear what the values of the configuration will be as once a value has been explicitly set, unsetting it will not return to the default value but rather to whatever value was set last.

Seem from that operational point of view, it does not seem a good idea to change the log level to debug as for an operational system, it will be unclear what values are set (unless you restart Opencast which brings it in an defined state - if the default values are still correctly documented) and whether the configuration was reloaded successfully.

What do you think about this?


Update: Furthermore, a configuration change is an event relevant for operations. Think of analysis a problem based on logs - it is relevant to know that the configuration has changed.

lkiesow:
I would count as a bug that the configuration is not set to the default value. Though it is an old one and not introduced with this pull request. I can take a look at that anyway.

Your other argument does not make a lot of sense though:

- Running with your argument, you would need to actually log each any every change in Opencast. After all, how can you ever be certain that something is the state it should be if it's not logged ;-P That would be very confusing. This does also not only apply to the configuration. E.g., you would want to log all FFmpeg output (which is at debug log level right now). Otherwise, you wouldn't be able to identify that FFmpeg does not come with an encoder you want to use (just one example).
- If you suspect something is wrong, you can always set the log level to debug to get more information, but in a normal state these logs are unimportant.
- If you want to check your loaded configuration, you can check all of it using the Karaf console or the webconsole right now already. Hence, if you want to make sure the correct values are applied: You already can.
- Finally, you should be able to trust your configuration. If you can't because Opencast does something wrong, that's a bug. But we cannot log everything just because there could be a bug somewhere. Or, if you want to, you can set the global log level to trace :-)

staubesv:
Please don't forget that we are talking about the log level INFO which is usually quite verbose and presents daily business information rather than indicating errors.

My arguments for having configuration changes logged at log level INFOs would be:

- Configuration changes are relevant (actually, more relevant than most other INFO logs)
- Configuration changes happen rarely and therefore won't clutter the log anyway

As reducing the log level of configuration changes doesn't help much during Opencast is running (as configuration changes are rare), I assume you want a more compact log for the startup?

There is potential for more compact logging by just log one INFO message per component rather than one per configuration property which is indeed overkill.

@gregorydlogan We might want to add some guidelines to our dev guide considering the use of log levels to ensure that people use them consistently. Not a priority, but would be useful.

I'll bring the topic "log level" up in one of the next technical meetings.

lkiesow:
I don't want to have any configuration logged by default. It's just not helpful at all. I know my configuration. I can look it up in my configuration files which I should be able to trust. If not, that's definitely a bug.

If I suspect a bug related to Opencast not dealing with configuration properly, I can always check my runtime configuration. Either by setting the log level to debug or by looking at the actually loaded configuration e.g. in the web console which gives me *all* of the configuration for a managed OSGI service–which also means that no one can make a mistake in incorrect values, …

Honestly, if all of Opencast's configuration would be logged, no one would completely read through it to check that everything is correct. People may be using it for checking specific values if there is a problem which, as I mentioned before, is something you can already do.

@gregorydlogan, are you now also working for SWITCH? (Asking since @staubesv assigned work to you)

staubesv:
Thought some more about this: I think the information I would expect to see at log level INFO is the fact that a component has loaded its configuration (rather than the actual configuration values).

Considering pinging @gregorydlogan I thought it would be worth mentioning the idea of having a guide about log levels (in his role as QA manager)... simply forgot that @lkiesow is QA manager, too.

So the correct ping would be:

@gregorydlogan and @lkiesow: I think it would be helpful to have some guidelines considering the use of log levels. The motivation is quite similar like for code style guidelines: Ensuring consistent use of log levels and avoiding "toggling" of log levels as different people might adapt them to different values over time.

That is neither a priority nor an assignment of work - it is just a suggestion about something that I think might be useful for the Opencast development process.

lkiesow:
I was just confused, that's all :-D

So, your suggestion would be that all 145 (r/5.x) ManagedService implementations log an updated message? That's at least not the ~1k+ messages you would log if you log every configuration key. I'm still not convinced but I have a less strong opinion on that. Let's talk about it on Tuesday.

gregorydlogan:
@staubesv, what you're looking for is defined [here](https://docs.opencast.org/develop/developer/code-style/), although in many (most? all?) cases the logging was written without consulting those rules.  They got lost in the shuffle between the wiki and markdown, and I suspect most committers are not aware of them.  I'm fine with raising awareness of this though - it's definitely a technical debt/code smell.

I agree with @lkiesow that dropping the level makes sense.  I don't particularly like it - I prefer from an ops/user standpoint that Opencast make it clear when it changes/reloads from config, but in the grand scheme of things that doesn't matter much.  We almost need a level between DEBUG and INFO - OPS, where things like this can be logged without dumping the whole of ffmpeg's output :)  Maybe push the ffmpeg output down into TRACE?  Anyway, I'm fine with this as is.

staubesv:
> @staubesv, what you're looking for is defined here

Damn, forgot about that...

> although in many (most? all?) cases the logging was written without consulting those rules. 

Well, these rules are just not precise enough to really be helpful. Example of INFO:

> Information relevant to server administrators. Creating, updating, and deleting files should be logged here.

It very much depends on the point of view whether an information is relevant for a server administrator. For example, we don't seem to agree whether configuration changes are relevant for a server admin or not.
On the other hand, if we really would log all changes to all files, that might be quite a lot of log messages ;-)

So what we could do is to make this guide more precise on ad hoc basis.
For example, we could add explicit information for common cases as "what should be logged if a component reloads its configuration".

Let's bring this up on Tuesday and see what others say. If we agree on what should be logged in the "configuration reload" case, I can raise a PR to document this.

PR #262, opened 2018-05-24
MH-12919, REST Docs Dependencies

lkiesow:
This patch removes the inclusion of a lot of unnecessary resources
originating from a time where this module did hold more than the REST
documentation. All of the few necessary resources are now included.

This patch also includes an update for the now unsupported version 1.x
of jQuery, updated to v3.3.1.

Finally, this separates JavaScript and CSS to make using some security
features like Content-Security-Policy headers possible.


modules/runtime-info-ui-ng/src/main/resources/ui/js/jquery/jquery.js
```diff
@@ -0,0 +1,2 @@
+/*! jQuery v3.3.1 | (c) JS Foundation and other contributors | jquery.org/license */
```
staubesv, line 1: Should be included in NOTICES, right?

staubesv (changes requested):
Hi @lkiesow Besides jQuery probably should be mentioned in NOTICES, the PR is fine. Can you add the missing lines to NOTICES?

lkiesow:
Makes sense. I fixed that.
I also fixed the runtime-info-ui package which had parts of the necessary resources included…

staubesv (approved):

PR #193, opened 2018-04-18
MH-12826: Make workflow processing settings persistent

pmiddend:
This PR adds “workflow properties” to events, as described here:

https://opencast.jira.com/browse/MH-12826

The idea is to store the last workflow instance’s properties using
asset manager properties and load them when the user invokes “Actions
-> Start Task”. This, of course, has to work when multiple events are
selected, which makes the code a lot more difficult.

For example, if two events are selected, both have the (boolean,
checkbox) workflow property “publish to X”, but with different
values. Then the gui now displays an indeterminate checkbox. In a
similar fashion, radio buttons are all deselected for ambiguous
values, and text fields are blanked.

The “Schedule Task” wizard also has a table at the end listing all the
selected properties. This table was changed to display an asterisk for
properties that are different for each event. When hitting the submit
button, the corresponding workflow properties are resolved for each
event separately.

There’s a new endpoint to retrieve these properties in bulk for
multiple events (for performance reasons). To accomplish that, the
`AssetManager` query language had to be extended.

This work is sponsored by SWITCH.

PR #264, opened 2018-05-25
MH-12607, Process-Smil - edit and encode to multiple delivery formats

phoebem:
This is the process-smil part of 12607, updated to new composer code.

PR #265, opened 2018-05-26
MH-12813: Add audio and video track selection to video editor

pmiddend:
This PR adds a new tab “Tracks” to the video editor that allows the user to select which audio and video tracks will be published. In this new tab, the user simply clicks on the video track (represented by a
thumbnail) or the audio track (represented by its waveform) to hide it. Sanity checks are included, so its not possible to hide _every_ video and audio track there is, for example.

40481539-0407fbda-5f52-11e8-872c-704e588a36d0

The information about which tracks are hidden is stored in asset manager properties, which are then injected into the running workflow(s). Thus, this PR depends on [193](#193) (and is rebased on it).

This _theoretically_ works for an arbitrary number of tracks in the media package, but it currently is constrained to one and two tracks (the most common use case). The allowed flavors flavors are configurable.

To make this possible, a specialized workflow operation handler `SelectTracksWorkflowOperationHandler` had to be written, which is capable of removing audio tracks from multiple input tracks, as well as muxing a video track and an audio track from two different input
tracks (so it’s very similar to the `PrepareAV` WOH).

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/index.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/index.md
@@ -77,6 +77,7 @@ The following table contains the workflow operations that are available in an ou
 |retract-youtube     |Retracts media from YouTube                                    |[Documentation](retract-youtube-woh.md)|
 |segment-video       |Extracting segments from presentation                          |[Documentation](segmentvideo-woh.md)|
 |segmentpreviews     |Extract segment images from a video using FFmpeg               |[Documentation](segmentpreviews-woh.md)|
+|select-streams       |Select streams for further processing                           |[Documentation](select-streams-woh.md)|
 |send-email          |Sends email notifications at any part of a workflow            |[Documentation](send-email-woh.md)|
 |series              |Apply series to the mediapackage                               |[Documentation](series-woh.md)|
 |silence             |Silence detection on audio of the mediapackage                 |[Documentation](silence-woh.md)|
--- a/docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
@@ -1,0 +1,79 @@
+SelectStreamsWorkflowOperationHandler
+====================================
+
+
+Description
+-----------
+
+The SelectStreamsWorkflowOperationHandler can be used in case not all source streams should be processed. For example,
+given a recording with a presenter and a presentation track, the final recording to be published should only include the
+video stream of the presenter track and the audio stream of the presentation track.
+
+The workflow operation will use workflow properties set by the Opencast video editor to determine which streams should be
+selected for further processing and add them to the media package based on `target-flavor` and `target-tags`.
+
+
+Parameter Table
+---------------
+
+Configuration Key | Example   | Description
+:-----------------|:----------|:-----------
+source-flavor\*   | */source  | The flavor of the track(s) to use as a source input
+target-flavor\*   | */work    | The flavor of the target track(s)
+target-tags       | download  | The tags applied to all target tracks
+audio-muxing      | force     | Move single-audio media packages to a specific track (see below)
+force-target      | presenter     | Target track for the `force` setting for `audio-muxing`
+
+\* mandatory configuration key
+
+Audio Muxing
+-----------------
+
+The optional `audio-muxing` parameter has three possible values: `none` (same as omitting the option), `force` and
+`duplicate`.
+
+### `none` ###
+
+If `none` is specified or the option is omitted, the audio stream is taken from the specified `source-flavor` track and is
+edited according to the selections in video editor’s “Tracks” panel. The resulting tracks are stored in the
+corresponding `target-flavor` and `target-tags` are applied.
+
+Note: If your editing results in a single video and single audio (track/stream) they will be muxed together even if
+this option is set to `none`.
+
+### `force` ###
+
+The parameter value `force` only applies to media packages that have exactly one non-hidden audio stream. For media
+packages without an audio stream or with more than one audio stream, the behavior is the same as if the parameter were
+omitted. The same applies to media packages for which there is only one audio stream, and it already belongs to the
+track with flavor type given by `force-target` (or `presenter` if that parameter is omitted).
+
+If, however, there is only one non-hidden audio stream and it does *not* belong to the track given by `force-target`,
+then the WOH will “move” the audio stream to this target track. Specifically, it will mux the video stream of
+`force-target` with the audio stream it found. Then, it removes the audio stream from the original track.
+
+For example, consider a media package with two tracks, *presenter* and *presentation*. Both of these tracks have
+audio components, however the *presenter* audio stream is hidden. This WOH will mux *presentations*'s audio stream
+with *presenter*'s video, and remove the audio track from *presentation*'s video.
+
+### `duplicate` ###
+
+The parameter value `duplicate` only applies to media packages that have exactly one non-hidden audio stream and no
+hidden video streams. For these media packages, the WOH will mux the audio stream it found to all video streams in
+the media package. For media packages without an audio stream or with more than one audio stream, the behavior is
+the same as if the parameter were omitted.
+
+Operation Example
+-----------------
+
+    <operation
+      id="select-tracks"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error"
+      description="Select tracks for further processing">
+      <configurations>
+        <configuration key="source-flavor">*/source</configuration>
+        <configuration key="target-flavor">*/work</configuration>
+        <configuration key="audio-muxing">force</configuration>
+      </configurations>
+    </operation>
--- a/docs/guides/admin/mkdocs.yml
+++ b/docs/guides/admin/mkdocs.yml
@@ -143,6 +143,7 @@ pages:
    - Retract YouTube: 'workflowoperationhandlers/retract-youtube-woh.md'
    - Segment Previews: 'workflowoperationhandlers/segmentpreviews-woh.md'
    - Segment Video: 'workflowoperationhandlers/segmentvideo-woh.md'
+   - Select Streams: 'workflowoperationhandlers/select-streams-woh.md'
    - Send Email: 'workflowoperationhandlers/send-email-woh.md'
    - Series: 'workflowoperationhandlers/series-woh.md'
    - Silence: 'workflowoperationhandlers/silence-woh.md'
--- a/etc/encoding/opencast-movies.properties
+++ b/etc/encoding/opencast-movies.properties
@@ -48,7 +48,7 @@ profile.mux-av.work.name = mux audio and video
 profile.mux-av.work.input = stream
 profile.mux-av.work.output = visual
 profile.mux-av.work.suffix = -work.#{in.video.suffix}
-profile.mux-av.work.ffmpeg.command = -i #{in.video.path} -i #{in.audio.path} -shortest -c copy #{out.dir}/#{out.name}#{out.suffix}
+profile.mux-av.work.ffmpeg.command = -i #{in.video.path} -i #{in.audio.path} -shortest -c copy -map 0:v:0 -map 1:a:0 #{out.dir}/#{out.name}#{out.suffix}
 
 # Trim a stream
 #   This command will trim and input stream. Trimming will be fast, as no
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
@@ -125,6 +125,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Dictionary;
@@ -182,6 +183,9 @@ public class ToolsEndpoint implements ManagedService {
 
   /** The Json key for the default thumbnail position. */
   private static final String DEFAULT_THUMBNAIL_POSITION_KEY = "defaultThumbnailPosition";
+
+  /** The Json key for the source_tracks array. */
+  private static final String SOURCE_TRACKS_KEY = "source_tracks";
 
   /** Tag that marks workflow for being used from the editor tool */
   private static final String EDITOR_WORKFLOW_TAG = "editor";
@@ -434,29 +438,65 @@ public class ToolsEndpoint implements ManagedService {
       throw new WebApplicationException(e, SC_INTERNAL_SERVER_ERROR);
     }
 
-    final MediaPackageElementFlavor sourceTracksFlavor = new MediaPackageElementFlavor("*",
-      adminUIConfiguration.getThumbnailSourceFlavorSubtype());
+    final Map<String, String> latestWfProperties = WorkflowPropertiesUtil
+      .getLatestWorkflowProperties(assetManager, mediaPackageId);
+    // The properties have the format "hide_flavor_audio" or "hide_flavor_video", where flavor is preconfigured.
+    // We filter all the properties that have this format, and then those which have values "true".
+    final Collection<Tuple<String, String>> hiddens = latestWfProperties.entrySet()
+      .stream()
+      .map(p -> Tuple.tuple(p.getKey().split("_"), p.getValue()))
+      .filter(p -> p.getA().length == 3)
+      .filter(p -> p.getA()[0].equals("hide"))
+      .filter(p -> p.getB().equals("true"))
+      .map(p -> Tuple.tuple(p.getA()[1], p.getA()[2]))
+      .collect(Collectors.toSet());
 
-    final List<JValue> sourceTracks = Arrays.stream(mp.getTracks(sourceTracksFlavor))
-          .map(MediaPackageElement::getFlavor)
-          .map(e -> {
-            String side;
-            if (e.equals(this.adminUIConfiguration.getSourceTrackLeftFlavor())) {
-              side = "left";
-            } else if (e.equals(this.adminUIConfiguration.getSourceTrackRightFlavor())) {
-              side = "right";
-            } else {
-              side = "unknown";
-            }
-            return obj(f("flavor", obj(f("type" ,e.getType()), f("subtype", e.getSubtype()))), f("side", side));
-          })
-          .collect(Collectors.toList());
+    final Collection<MediaPackageElementFlavor> acceptedFlavors = Arrays
+      .asList(this.adminUIConfiguration.getSourceTrackLeftFlavor(),
+        this.adminUIConfiguration.getSourceTrackRightFlavor());
+
+    // We already know the internal publication exists, so just "get" it here.
+    final Publication internalPub = getInternalPublication(mp).get();
+
+    final List<JValue> sourceTracks = Arrays.stream(mp.getElements())
+      .filter(e -> e.getElementType().equals(Type.Track))
+      .map(e -> (Track)e)
+      .filter(e -> acceptedFlavors.contains(e.getFlavor()))
+      .map(e -> {
+        String side = null;
+        if (e.getFlavor().equals(this.adminUIConfiguration.getSourceTrackLeftFlavor())) {
+          side = "left";
+        } else if (e.getFlavor().equals(this.adminUIConfiguration.getSourceTrackRightFlavor())) {
+          side = "right";
+        }
+        final boolean audioHidden = hiddens.contains(Tuple.tuple(e.getFlavor().getType(), "audio"));
+        final String audioPreview = Arrays.stream(internalPub.getAttachments())
+          .filter(a -> a.getFlavor().getType().equals(e.getFlavor().getType()))
+          .filter(a -> a.getFlavor().getSubtype().equals(this.adminUIConfiguration.getPreviewAudioSubtype()))
+          .map(MediaPackageElement::getURI).map(URI::toString)
+          .findAny()
+          .orElse(null);
+        final SourceTrackSubInfo audio = new SourceTrackSubInfo(e.hasAudio(), audioPreview,
+          audioHidden);
+        final boolean videoHidden = hiddens.contains(Tuple.tuple(e.getFlavor().getType(), "video"));
+        final String videoPreview = Arrays.stream(internalPub.getAttachments())
+          .filter(a -> a.getFlavor().getType().equals(e.getFlavor().getType()))
+          .filter(a -> a.getFlavor().getSubtype().equals(this.adminUIConfiguration.getPreviewVideoSubtype()))
+          .map(MediaPackageElement::getURI).map(URI::toString)
+          .findAny()
+          .orElse(null);
+        final SourceTrackSubInfo video = new SourceTrackSubInfo(e.hasVideo(), videoPreview,
+          videoHidden);
+        return new SourceTrackInfo(e.getFlavor().getType(), e.getFlavor().getSubtype(), audio, video, side);
+      })
+      .map(SourceTrackInfo::toJson)
+      .collect(Collectors.toList());
 
     return RestUtils.okJson(obj(f("title", v(mp.getTitle(), Jsons.BLANK)),
             f("date", v(event.getRecordingStartDate(), Jsons.BLANK)),
             f("series", obj(f("id", v(event.getSeriesId(), Jsons.BLANK)), f("title", v(event.getSeriesName(), Jsons.BLANK)))),
             f("presenters", arr($(event.getPresenters()).map(Functions.stringToJValue))),
-            f("source_tracks", arr(sourceTracks)),
+            f(SOURCE_TRACKS_KEY, arr(sourceTracks)),
             f("previews", arr(jPreviews)), f(TRACKS_KEY, arr(jTracks)),
             f("thumbnail", obj(thumbnailFields)),
             f("duration", v(mp.getDuration())), f(SEGMENTS_KEY, arr(jSegments)), f("workflows", arr(jWorkflows))));
@@ -586,6 +626,20 @@ public class ToolsEndpoint implements ManagedService {
         logger.warn("Unable to create a SMIL cutting catalog ({}): {}", details, getStackTrace(e));
         return R.badRequest("Unable to create SMIL cutting catalog");
       }
+
+      final Map<String, String> workflowProperties = java.util.stream.Stream
+        .of(this.adminUIConfiguration.getSourceTrackLeftFlavor(), this.adminUIConfiguration.getSourceTrackRightFlavor())
+        .flatMap(flavor -> {
+          final java.util.stream.Stream.Builder<Tuple<String, String>> r = java.util.stream.Stream.builder();
+          final Optional<SourceTrackInfo> track = editingInfo.sourceTracks.stream().filter(s -> s.getFlavor().equals(flavor)).findAny();
+          final boolean audioHidden = track.map(e -> e.audio.hidden).orElse(false);
+          r.accept(Tuple.tuple("hide_" + flavor.getType() + "_audio", Boolean.toString(audioHidden)));
+          final boolean videoHidden = track.map(e -> e.video.hidden).orElse(false);
+          r.accept(Tuple.tuple("hide_" + flavor.getType() + "_video", Boolean.toString(videoHidden)));
+          return r.build();
+        }).collect(Collectors.toMap(Tuple::getA, Tuple::getB));
+
+      WorkflowPropertiesUtil.storeProperties(assetManager, mediaPackage, workflowProperties);
 
       try {
         addSmilToArchive(mediaPackage, smil);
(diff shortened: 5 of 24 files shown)


JulianKniephoff:
I am looking at this in context of another possibly related issue I am working on for the ETH and I noticed the following:

image

This happened when selecting only one video stream, but two audio streams, obviously; the problem is that this message does not disappear when I then do "the right thing" and successfully publish the event.

Also I find it kind of confusing that this message appears only after I hit the "do the thing button", but the error message for deselecting for example all audio streams appears instantly (and in a different position).

pmiddend:
@JulianKniephoff You're absolutely right. I fixed the issues you mentioned.

staubesv:
Hi @pmiddend Please note that this PR currently has merge conflicts. Could you please resolve them?

pmiddend:
@staubesv Thanks for noticing. Fixed.

staubesv:
Hi @pmiddend, please note that this PR has merge conflicts again.

krinnewitz:
Hi @staubesv, Conflicts should now be fixed (again).

staubesv:
Thanks ;-)

etc/encoding/opencast-movies.properties
```diff
@@ -48,7 +48,7 @@ profile.mux-av.work.name = mux audio and video
 profile.mux-av.work.input = stream
 profile.mux-av.work.output = visual
 profile.mux-av.work.suffix = -work.#{in.video.suffix}
-profile.mux-av.work.ffmpeg.command = -i #{in.video.path} -i #{in.audio.path} -shortest -c copy #{out.dir}/#{out.name}#{out.suffix}
+profile.mux-av.work.ffmpeg.command = -i #{in.video.path} -i #{in.audio.path} -shortest -c copy -map 0:v:0 -map 1:a:0 #{out.dir}/#{out.name}#{out.suffix}
```
staubesv, line 51: As this encoding profile is hard-coded in WOH prepare-av, you might want to introduce another encoding profile for WOH tracks-select. Or is this change supposed to be useful for WOH prepare-av, too?

pmiddend: Ah yes. So when the prepare-av workflow receives a mediapackage with two tracks, and it's supposed to mux these together (so take the audio subtrack of track A and the video subtrack of track B), it actually strips the video subtrack from A and then merges the two streams together. This is unnecessary, since ffmpeg can mux the two streams with a single command. I asked why this isn't being done in the profile above, and Greg replied that it might be because the code is ancient (meaning from times when ffmpeg wasn't able to mux as intelligently as it can, now).
To be extremely sure, I could introduce another profile.

staubesv: Thanks for the info! In that case, just modifying the existing encoding profile should be fine.

staubesv:
Hi @pmiddend I've just merged #264 which caused merged conflicts in this PR. Could you please resolve them?

krinnewitz:
Hi @staubesv The merge conflict is resolved. Travis build seems to fail due to random DNS lookup error.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
```diff
@@ -0,0 +1,500 @@
...
+  private enum AudioMuxing {
+    NONE("none"), FORCE("force"), DUPLICATE("duplicate");
+
+    private final String value;
+
+    AudioMuxing(final String value) {
+      this.value = value;
+    }
+
+    static AudioMuxing fromConfigurationString(final String s) {
```
lkiesow: Isn't this basically what the `enum` does anyway and you could e.g. just use `AudioMuxing.valueOf(…)` instead of this custom code?

docs/guides/admin/docs/workflowoperationhandlers/select-tracks-woh.md
```diff
@@ -0,0 +1,69 @@
...
+target-tags       | download  | The tags applied to all target tracks
+audio-muxing      | force     | Move single-audio media packages to a specific track (see below)
+force-target      | presenter     | Target track for the `force` setting for `audio-muxing`
+
+\* mandatory configuration key
+
+Audio Muxing
+-----------------
+
+The optional `audio-muxing` parameter has three possible values: `none` (same as omitting the option), `force` and
```
lkiesow: There is no description for `none`? What happens if `none` is specified (or the configuration is omitted)?

docs/guides/admin/docs/workflowoperationhandlers/select-tracks-woh.md
```diff
@@ -0,0 +1,69 @@
...
+`duplicate`.
+
+### `force` ###
+
+The parameter value `force` only applies to media packages that have exactly one non-hidden audio track. For media packages
```
lkiesow: The usage of tracks in this section is highly confusing. You are confusing tracks with streams.

For example:

> …a media package with two tracks, “presenter” and “presentation”, and the audio track of “presenter” is hidden.

actually means

> …a media package with two tracks, “presenter” and “presentation”, and the audio *stream* of “presenter” is hidden.

because if you hide a track's track, you hide itself, don't you? Note that, of course, it's possible to have two tracks flavored the same (e.g. two presenter tracks) but I don't think you mean that.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
```diff
@@ -0,0 +1,500 @@
...
+
+    // First case: We have only tracks with non-hidden video streams. So we keep them all and possibly cut away audio.
+    if (allNonHidden(augmentedTracks, SubTrack.VIDEO)) {
+      final Optional<AudioMuxing> audioMuxing = getConfiguration(workflowInstance, CONFIG_AUDIO_MUXING)
+              .map(AudioMuxing::fromConfigurationString);
+      // For both special options, we need to find out if we have exactly one audio track present
+      final Optional<AugmentedTrack> singleAudioTrackOpt = findSingleAudioTrack(augmentedTracks);
+      final boolean multipleVideo = augmentedTracks.size() > 1;
+      if (multipleVideo && audioMuxing.map(m -> m == AudioMuxing.DUPLICATE).orElse(Boolean.FALSE) && singleAudioTrackOpt
+              .isPresent()) {
```
lkiesow: Maybe use
```java
AudioMuxing.DUPLICATE.equals(audioMuxing.orElse(null))
```

instead of

```java
audioMuxing.map(m -> m == AudioMuxing.DUPLICATE).orElse(Boolean.FALSE)
```

to be less obscure or, even better, not use an Optional at all since the default value is basically `AudioMuxing.EMPTY` anyway, making this not really optional.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
```diff
@@ -0,0 +1,500 @@
...
+    } else {
+      // Second case: we have exactly one video track that is not hidden (hopefully, because for all other cases there
+      // were no requirements given).
+      final MuxResult muxResult = muxSingleVideoTrack(mediaPackage, augmentedTracks);
+      result.add(muxResult);
+    }
+
+    final MediaPackageElementFlavor targetTrackFlavor = MediaPackageElementFlavor.parseFlavor(StringUtils.trimToNull(
+            getConfiguration(workflowInstance, "target-flavor")
+                    .orElseThrow(() -> new IllegalStateException("Target flavor not specified"))));
```
lkiesow: Fail fast. No need to handle all the media file operations if we finally fail with a configuration error anyway.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
```diff
@@ -0,0 +1,500 @@
...
+
+        if (singleAudioTrack.track != forceTargetTrack.track) {
+          // Copy it over...
+          final TrackJobResult muxResult = mux(forceTargetTrack.track, singleAudioTrack.track, mediaPackage);
+          result.add(muxResult);
+
+          // ...and remove the original
+          final TrackJobResult hideAudioResult = hideAudio(singleAudioTrack.track, mediaPackage);
+          result.add(hideAudioResult);
+        }
```
lkiesow: You need to copy tracks you do not mux as well. Otherwise, the output of this operation is impossible to predict.

Example:

- You have two video files `presenter/source` and `presentation/source`. The first has video only, the second has video and audio.
- You set this operation to force audio in `presentation` and set the target flavor to `*/selected`
- The operation will run, claim success but since the audio was already in the right place, there is no track with the flavor `*/selected`
- Result: The next operation will likely fail.

here is a simple patch which fixes the problem. I'm not sure if I covered all edge cases:

```diff
diff --git a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
index dd135ddbd..385687bdd 100644
--- a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
+++ b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectTracksWorkflowOperationHandler.java
@@ -39,11 +39,13 @@ import org.opencastproject.workflow.api.WorkflowOperationResult;
 import org.opencastproject.workflow.api.WorkflowOperationTagUtil;
 import org.opencastproject.workspace.api.Workspace;
 
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -51,6 +53,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -273,6 +276,8 @@ public class SelectTracksWorkflowOperationHandler extends AbstractWorkflowOperat
           if (t.track != singleAudioTrack.track) {
             final TrackJobResult jobResult = mux(t.track, singleAudioTrack.track, mediaPackage);
             result.add(jobResult);
+          } else {
+            result.add(copyTrack(t.track));
           }
         }
       } else if (multipleVideo && audioMuxing.map(m -> m == AudioMuxing.FORCE).orElse(Boolean.FALSE)
@@ -302,6 +307,15 @@ public class SelectTracksWorkflowOperationHandler extends AbstractWorkflowOperat
           // ...and remove the original
           final TrackJobResult hideAudioResult = hideAudio(singleAudioTrack.track, mediaPackage);
           result.add(hideAudioResult);
+        } else {
+          result.add(copyTrack(singleAudioTrack.track));
+        }
+
+        // Just copy the rest of the tracks to ensure they got the correct output flavor
+        for (final AugmentedTrack augmentedTrack : augmentedTracks) {
+          if (augmentedTrack.track != singleAudioTrack.track && augmentedTrack.track != forceTargetTrack.track) {
+            result.add(copyTrack(augmentedTrack.track));
+          }
         }
       } else {
         // No special options selected, or conditions for special options don't match.
@@ -497,4 +511,27 @@ public class SelectTracksWorkflowOperationHandler extends AbstractWorkflowOperat
       return new AugmentedTrack(t, hideAudio, hideVideo);
     }).collect(Collectors.toList());
   }
+
+  private TrackJobResult copyTrack(Track track) throws WorkflowOperationException {
+    Track copiedTrack = (Track) track.clone();
+    copiedTrack.setIdentifier(UUID.randomUUID().toString());
+    try {
+      // Generate a new filename
+      String targetFilename = copiedTrack.getIdentifier();
+      String extension = FilenameUtils.getExtension(track.getURI().getPath());
+      if (extension.length() > 0) {
+        targetFilename += "." + extension;
+      }
+
+      // Copy the files on dis and put them into the working file repository
+      logger.debug("Start copying element {}.", track.getURI());
+      URI newUri = workspace.put(track.getMediaPackage().getIdentifier().toString(), copiedTrack.getIdentifier(),
+              targetFilename, workspace.read(track.getURI()));
+      copiedTrack.setURI(newUri);
+    } catch (IOException | NotFoundException e) {
+      throw new WorkflowOperationException(String.format("Error while copying track %s", track.getIdentifier()), e);
+    }
+
+    return new TrackJobResult(copiedTrack, 0);
+  }
 }
```

lkiesow (changes requested):
This is mainly a review of the select-tracks woh which has some broken edge cases. Note that the operations name is highly confusing since it does not select tracks but video and audio streams.

pmiddend:
Thanks @lkiesow for the excellent review. I've pushed the following changes:

  - I fixed the markdown docs, after banging my head for not realizing "stream" was the term I was looking for all along. I also added a section about the default behavior of `none` for `audio-muxing`. 
  - I applied your patch containing `copyTrack` and tested it. Looked good so far, and I think you got all the edge cases.
  - I removed `value` from AudioMuxing and am now relying on the correspondence between the enum names and their WOH values.
  - I removed the `Optional` for audio-muxing, opting for `orElse(NONE)` instead.
  - The code fails faster now. :)

By the way, I'm very open to suggestions regarding the WOH name. I think neither SWITCH nor plapadoo were excited about it. Maybe we use `select-streams` instead?

docs/guides/admin/docs/workflowoperationhandlers/select-tracks-woh.md
```diff
@@ -0,0 +1,79 @@
+SelectTracksWorkflowOperationHandler
```
KatrinIhler: I also think that _SelectStreamsWorkflowOperationHandler_ is clearer.

docs/guides/admin/docs/workflowoperationhandlers/select-tracks-woh.md
```diff
@@ -0,0 +1,79 @@
...
+====================================
+
+
+Description
+-----------
+
+The SelectTracksWorkflowOperationHandler can be used in case not all source tracks should be processed. For example,
+given a recording with a presenter and presentation track, the final recording to be publish may should only include the
```
KatrinIhler: ...and _a_ presentation track, ...to be _published_  _may/should_ ...

docs/guides/admin/docs/workflowoperationhandlers/select-tracks-woh.md
```diff
@@ -0,0 +1,79 @@
...
+
+For example, if there is a media package with two tracks, “presenter” and “presentation”, and the audio stream of
+“presenter” is hidden. Then the WOH will mux presenter’s video stream and presentations audio stream and store it in
+presenter’s place. It will also remove the audio stream from “presentation”.
+
+### `duplicate` ###
+
+The parameter value `duplicate` only applies to media packages that have exactly one non-hidden audio stream. For media
+packages without an audio stream or with more than one audio stream, the behavior is the same as if the parameter were
+omitted. For these media packages, the WOH will mux the audio stream it found to all video streams in the media package.
```
KatrinIhler: Will the audio stream also be muxed into the hidden video streams?

KatrinIhler (changes requested):
I just tested this and found the following things:
1. If I deselect all audio streams, there is an error message, but I can still hit save. and it's computed in the backend. The save button should be disabled.
2. When I deselect all video streams, there's also an error message and I can also still hit save, but then another error message appears. Here the save button should also be disabled.
3. Additionally, the error _You have selected no video tracks to process_ doesn't disappear even if I open a different video in the video editor.

pmiddend:
@KatrinIhler Thanks for the thorough review! I've fixed all of the issues you described.

docs/guides/admin/docs/workflowoperationhandlers/index.md
```diff
@@ -77,6 +77,7 @@ The following table contains the workflow operations that are available in an ou
 |retract-youtube     |Retracts media from YouTube                                    |[Documentation](retract-youtube-woh.md)|
 |segment-video       |Extracting segments from presentation                          |[Documentation](segmentvideo-woh.md)|
 |segmentpreviews     |Extract segment images from a video using FFmpeg               |[Documentation](segmentpreviews-woh.md)|
+|select-tracks       |Select tracks for further processing                           |[Documentation](select-tracks-woh.md)|
```
gregorydlogan: select-*tracks*-woh vs select-*streams*-woh.  Should match the rest of the documentation as well (which is probably using both :D)

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
+SelectStreamsWorkflowOperationHandler
+====================================
+
+
+Description
+-----------
+
+The SelectStreamsWorkflowOperationHandler can be used in case not all source tracks should be processed. For example,
```
gregorydlogan: track vs stream in this paragraph as well.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+
+Audio Muxing
+-----------------
+
+The optional `audio-muxing` parameter has three possible values: `none` (same as omitting the option), `force` and
+`duplicate`.
+
+### `none` ###
+
+If `none` is specified or the option is omitted, tracks are taken from the specified `source-flavor` and are edited
```
gregorydlogan: *audio* tracks I'm assuming?  Maybe rephrase this as "the audio track is taken..."?

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+
+\* mandatory configuration key
+
+Audio Muxing
+-----------------
+
+The optional `audio-muxing` parameter has three possible values: `none` (same as omitting the option), `force` and
+`duplicate`.
+
+### `none` ###
```
gregorydlogan, line 35: I'm not a huge fan of this formatting when rendered, but I don't have a better suggestion.  Everything I can find makes things either looks weird or like it's a separate section entirely.

gregorydlogan:
Still working on my review, but I ran into this in testing

```
2018-09-13 11:49:48,360 | ERROR | (WorkflowOperationWorker:180) - Workflow operation 'operation:'image', position:11, state:'FAILED'' failed                                                                                                                                              
java.lang.RuntimeException: Parse failure                                                                                                                                                                                                                                                 
        at com.entwinemedia.fn.parser.Result.getRest(Result.java:50)[25:functional:1.4.2]                                                                                                                                                                                                 
        at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler.parsePositions(ImageWorkflowOperationHandler.java:595)[63:opencast-composer-workflowoperation:6.0.0.SNAPSHOT]                                                                                      
        at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler.configure(ImageWorkflowOperationHandler.java:512)[63:opencast-composer-workflowoperation:6.0.0.SNAPSHOT]                                                                                           
        at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler.start(ImageWorkflowOperationHandler.java:165)[63:opencast-composer-workflowoperation:6.0.0.SNAPSHOT]                                                                                               
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.start(WorkflowOperationWorker.java:233)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                           
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.execute(WorkflowOperationWorker.java:155)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                         
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.runWorkflowOperation(WorkflowServiceImpl.java:892)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                    
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.process(WorkflowServiceImpl.java:1918)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                                
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2348)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                         
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2314)[193:opencast-workflow-service-impl:6.0.0.SNAPSHOT]                                                                                                                         
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)[:1.8.0_181]                                                                                                                                                                                                           
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)[:1.8.0_181]                                                                                                                                                                                    
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)[:1.8.0_181]                                                                                                                                                                                    
        at java.lang.Thread.run(Thread.java:748)[:1.8.0_181]                                    
```

All I did was ingest a presenter video+audio file, a presentation video file, and then attempt to process to a single presentation+audio output.  I'm using ffmpeg 3.4-static.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+
+If `none` is specified or the option is omitted, tracks are taken from the specified `source-flavor` and are edited
+according to the selections in video editor’s “Tracks” panel. The resulting tracks are stored in the corresponding `
+target-flavor` and `target-tags` are applied.
+
+Editing in the “Tracks” panel usually means removing video or audio streams from tracks. However, if there are multiple
```
gregorydlogan: I get what you're trying to say, but this is confusing.  What about something like:

Note: If your editing results in a single video and single audio (track/stream) they will be muxed together even if this option is set to `none`.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+The parameter value `force` only applies to media packages that have exactly one non-hidden audio stream. For media
+packages without an audio stream or with more than one audio stream, the behavior is the same as if the parameter were
+omitted. The same applies to media packages for which there is only one audio stream, and it already belongs to the
+track with flavor type given by `force-target` (or `presenter` if that parameter is omitted).
+
+If, however, there is only one non-hidden audio stream and it does *not* belong to the track given by `force-target`,
+then the WOH will “move” the audio stream to this target track. Specifically, it will mux the video stream of
+`force-target` with the audio stream it found. Then, it removes the audio stream from the original track.
+
+For example, if there is a media package with two tracks, “presenter” and “presentation”, and the audio stream of
```
gregorydlogan: For example, consider a mediapackage with two tracks, __presenter__ and __presentation__.  Both of these tracks have audio components, however the __presenter__ audio stream is hidden.  This WOH will mux __presentations__'s audio stream with __presenter__'s video, and remove the audio track from __presentation__'s video. ?

The quotes are used inconsistently, and adding both single and double quotes sucks.  Also reworded slightly.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+“presenter” is hidden. Then the WOH will mux presenter’s video stream and presentations audio stream and store it in
+presenter’s place. It will also remove the audio stream from “presentation”.
+
+### `duplicate` ###
+
+The parameter value `duplicate` only applies to media packages that have exactly one non-hidden audio stream and no
```
gregorydlogan: The parameter value `duplicate` only applies to media packages that have exactly one non-hidden audio stream and no hidden video streams.  For these media packages, the WOH will mux the audio stream it found to all video streams in the media package.  For media packages without an audio stream or with more than one audio stream, the behavior is the same as if the parameter were omitted.

Reworded for clarity.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+hidden video streams. For media packages without an audio stream or with more than one audio stream, the behavior
```
gregorydlogan: Why does this only work for MPs with non-hidden video streams?  If it's just duplicating from source to N outputs, why can't it handle N-1 outputs?

docs/guides/admin/mkdocs.yml
```diff
@@ -143,6 +143,7 @@ pages:
    - Retract YouTube: 'workflowoperationhandlers/retract-youtube-woh.md'
    - Segment Previews: 'workflowoperationhandlers/segmentpreviews-woh.md'
    - Segment Video: 'workflowoperationhandlers/segmentvideo-woh.md'
+   - Select Tracks: 'workflowoperationhandlers/select-tracks-woh.md'
```
gregorydlogan: Tracks vs streams again :)

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
```diff
@@ -0,0 +1,80 @@
...
+-----------------
+
+    <operation
+      id="select-tracks"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error"
+      description="Select tracks for further processing">
+      <configurations>
+        <configuration key="source-flavor">*/source</configuration>
+        <configuration key="target-flavor">*/source</configuration>
```
gregorydlogan: Do we want to target source flavours here?  Or something more intermediate?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
```diff
@@ -437,26 +441,65 @@ public Response getVideoEditor(@PathParam("mediapackageid") final String mediaPa
...
-            return obj(f("flavor", obj(f("type" ,e.getType()), f("subtype", e.getSubtype()))), f("side", side));
-          })
-          .collect(Collectors.toList());
+    final Map<String, String> latestWfProperties = WorkflowPropertiesUtil
+      .getLatestWorkflowProperties(assetManager, mediaPackageId);
+    // The properties have the format "hide_flavor_audio" or "hide_flavor_video", where flavor is preconfigured.
+    // We filter all the properties that have this format, and then those which have values "true".
+    final Collection<Tuple<String, String>> hiddens = latestWfProperties.entrySet()
+      .stream()
+      .map(p -> Tuple.tuple(p.getKey().split("_"), p.getValue()))
```
gregorydlogan, line 447: Wouldn't it make more sense to do a string comparison for the `hide_flavor` prefix?

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectStreamsWorkflowOperationHandler.java
```diff
@@ -0,0 +1,529 @@
...
+   *          an instance of the workspace
+   */
+  public void setWorkspace(final Workspace workspace) {
+    this.workspace = workspace;
+  }
+
+  private EncodingProfile getProfile(final String identifier) {
+    final EncodingProfile profile = this.composerService.getProfile(identifier);
+    if (profile == null) {
+      throw new IllegalStateException(String.format("couldn't find encoding profile \"%s\"", identifier));
```
gregorydlogan: IIRC there's an exception subclass floating around somewhere for this.

gregorydlogan (changes requested):
Looks good, minor nitpicks.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SelectStreamsWorkflowOperationHandler.java
pmiddend: Ah yes, the other WOH use `WorkflowOperationException`, so I'll adapt this.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
pmiddend: Well, it'll be a little shorter, but I don't think it makes _more_ sense. You'd still have to split the string to get `audio`/`video` and stuff. Or hard-code all the tests altogether.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
pmiddend: Honest reason: Because the code is really complicated as it is. ;) And it's not really properly prepared for videos with more than two tracks (yeah, "tracks" is correct this time :D), which I have stated in the long commentary before the if/else barrage. The plan was to improve on this WOH if somebody has N>2 as a requirement.
If you think it's wrong to merge it in this state, I'll of course rewrite it.

pmiddend:
@gregorydlogan Thanks for the input on the documentation. Incredible how many mistakes I've missed! I've corrected those and used your wording.

docs/guides/admin/docs/workflowoperationhandlers/select-streams-woh.md
staubesv: @pmiddend & @gregorydlogan Please note that we do explicitly not have the requirement that this code can deal with more than two tracks and it was never our plan to add support for N tracks. The feature is, however, implemented in a way that would allow other Adopters to continue to work on this in case they would want to implement support for N tracks.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
gregorydlogan: Ok, fair enough.  On first read I had some minor issue figuring out both the syntax and the goal.  Now that I get the syntax it's easier to read :)

gregorydlogan:
Hi @pmiddend, changes look good.  Still blocked on the processing failure.  Can reproduce on all my development machines.

pmiddend:
@gregorydlogan Okay, I reproduced and fixed it. 

I also fixed the error message you got. Turns out that during exception handling, an exception occured so the string that couldn't be parsed wasn't printed. :D

PR #266, opened 2018-05-28
MH-12911: Hotkey cheat sheet

JulianKniephoff:
Instead of cooking up my own solution to MH-12911, I thought I might try to use the cheat sheet already build into `angular-hotkeys`. After some frustrating debugging session with the tests, I got it working!

Pressing `?` anywhere now brings up a list of currently active keyboard shortcuts.

~~Some open questions, maybe:~~

* ~~Should we put a hint about this somewhere?~~
  * ~~Where?~~
* ~~Do we need to translate the keys themselves, too?~~


modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1922,13 +1922,50 @@
...
+   "HOTKEYS": {
+       "CHEAT_SHEET": {
+           "TITLE": "Keyboard shortcuts"
+       },
+       "KEYS": {
+           "CHEAT_SHEET": "Show/hide this help menu",
+           "GENERAL": {
+               "SELECT_NEXT_DASHBOARD_FILTER": "Select next dashboard filter",
+               "SELECT_PREVIOUS_DASHBOARD_FILTER":  "Select previous dashboard filter",
+               "REMOVE_FILTERS": "Remove filters"
```
staubesv: Comma missing

JulianKniephoff:
So much for "I don't need to test this tiny last minute addition to the PR." :sweat_smile:

lkiesow:
If you want a button in the UI, why not use the Help menu?

screenshot from 2018-05-29 12-03-26

JulianKniephoff:
That's a good idea, I haven't thought about that. I will add it shortly! :slightly_smiling_face:

JulianKniephoff:
Aaand done! :slightly_smiling_face:

modules/admin-ui/Gruntfile.js
```diff
@@ -239,8 +239,7 @@ module.exports = function (grunt) {
         }
       },
       sass: {
-        src: ['<%= yeoman.app %>/styles/{,*/}*.{scss,sass}'],
-        ignorePath: /(\.\.\/){1,2}bower_components\//
+        src: ['<%= yeoman.app %>/styles/{,*/}*.{scss,sass}']
```
staubesv: Seems not related to this PR

modules/admin-ui/lib/proxyMiddleware.js
```diff
@@ -25,7 +25,7 @@ module.exports = function (grunt, appPath) {
       ),
       connect().use(
         '/styles',
-        serveStatic('./.tmp/styles')
+        serveStatic('./target/grunt/.tmp/styles')
```
staubesv: Seems not related to this PR.

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
```diff
@@ -48,7 +48,7 @@ angular.module('adminNg.controllers')
         $scope.openTab = function (tab) {
             $scope.tab = tab;
             if ($scope.tab === "editor") {
-              $scope.area   = "segments";
+              $scope.area = "segments";
```
staubesv, line 51: Seems not related to this PR

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1922,13 +1923,50 @@
...
+   "HOTKEYS": {
+       "CHEAT_SHEET": {
+           "TITLE": "Keyboard shortcuts"
+       },
+       "KEYS": {
+           "CHEAT_SHEET": "Show/hide this help menu",
+           "GENERAL": {
+               "SELECT_NEXT_DASHBOARD_FILTER": "Select next dashboard filter",
+               "SELECT_PREVIOUS_DASHBOARD_FILTER":  "Select previous dashboard filter",
+               "REMOVE_FILTERS": "Remove filters",
```
staubesv: The key MAIN_MENU seems missing here (at least the cheat cheat misses it)

modules/admin-ui/src/main/webapp/scripts/app.js
```diff
@@ -106,9 +106,14 @@ angular.module('adminNg', [
...
     }]);
 }])
-.config(function(hotkeysProvider) {
-    hotkeysProvider.includeCheatSheet = false;
-})
+.config(['hotkeysProvider', function (hotkeysProvider) {
+    hotkeysProvider.includeCheatSheet = true;
+    hotkeysProvider.templateTitle = 'HOTKEYS.CHEAT_SHEET.TITLE';
+    hotkeysProvider.cheatSheetDescription = 'HOTKEYS.KEYS.CHEAT_SHEET';
+    hotkeysProvider.template = hotkeysProvider.template
```
staubesv: It might be worth to think about whether it wouldn't make more sense to provide a custom template here. That would avoid the "replace" that depend on implementation details of a library, the need of providing a vendor CSS and would enable the cheatsheet to fit within the Admin UI design.

modules/admin-ui/src/test/resources/test/unit/modules/events/subresources/controllers/toolsControllerSpec.js
```diff
@@ -1,5 +1,5 @@
 describe('Tools Edit controller', function () {
-    var $scope, $controller, $parentScope, $location, $window, $httpBackend, Notifications;
+    var $scope, $controller, $parentScope, $location, $httpBackend, Notifications;
```
staubesv, line 2: The changes to this file don't seem relevant to this PR at all.

staubesv (changes requested):

staubesv:
Hi @JulianKniephoff 

I've finished a first round of review and found to issues:

- Please try to avoid including changes that are not related to the actual PR in a PR
- The graphical design is currently not in any relationship with the Admin UI. It would be desirable to just re-use existing Admin UI styles, e.g. the modal window with its tables etc. That probably could be doable by providing a custom template. The "less-effort-but-less-consistent" approach would be to at least use font sizes, styles and colors the are also used in the Admin UI.

Besides that: Cool and helpful feature!

screen shot 2018-05-30 at 10 54 19

Note that the Admin UI uses neither the font sizes used here nor the font color, background color or close button.

modules/admin-ui/src/main/webapp/styles/main.scss
```diff
@@ -19,7 +19,8 @@
  *
  */
 
-// bower:scss
+// bower:css
+@import "../../../../bower_components/angular-hotkeys/build/hotkeys.css";
```
staubesv: Hmm... shouldn't this be included from index.html?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1922,13 +1923,50 @@
...
+               "STEP_FORWARD": "Jump 10s forward",
+               "PREVIOUS_SEGMENT": "Previous segment",
+               "NEXT_SEGMENT": "Next segment",
+               "VOLUME_UP": "Volume up",
+               "VOLUME_DOWN": "Volume down",
+               "MUTE": "Mute/unmute"
+           },
+           "EDITOR": {
+               "SPLIT_AT_CURRENT_TIME": "Split the video at current time",
+               "CUT_SELECTED_SEGMENT": "Remote current segment",
```
staubesv: Remote?

modules/admin-ui/Gruntfile.js
JulianKniephoff: It is, though. This excludes stylesheets from `bower_components` in the sass compilation. I need one of those stylesheets, though. :wink:

modules/admin-ui/lib/proxyMiddleware.js
JulianKniephoff: True, that was an oversight.

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
JulianKniephoff: True, I guess, and I admittedly am not yet that acquainted with the exact netiquette of the community, but I think small stylistic fixes should be okay to go with other changes. Kind of like in the boy scout rule: Always leave the code better than you found it.

Otherwise where are those changes going to go? "MH-4242: More style fixes"? :wink: 

At least give me credit for putting stuff like this in its own commit. :stuck_out_tongue:

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
JulianKniephoff: Yip, missed that.

modules/admin-ui/src/main/webapp/scripts/app.js
JulianKniephoff: I used `replace` here because I did not want to copy the template which we would then have to maintain and keep in sync with the original one. However, if you say we want a custom one anyway, that is a moot point. And I would even agree with that; I just wanted to see first whether this is something the community wants at all, before diving into extensive design work, which I am admittedly not the best at. :wink:

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
staubesv: I'm more in favour of "Never touch a running system".

It would be appropriate to create a separate issue and PR for style changes if they don't related to the ticket.
If that seems overkill, consider it as an indication that the change is simply not important enough to be individually addressed. Maybe it would then be better to go through more code and fix some similar style issue vs. having the fixes distributed over a large number of actually unrelated pull requests (but still having not fixed them everywhere).

Anyway, that is not a request to revert that change, but a suggestion to try to focus pull requests on the actual problem rather than including changes on whatever was near the code that needed to be changed.

modules/admin-ui/src/test/resources/test/unit/modules/events/subresources/controllers/toolsControllerSpec.js
JulianKniephoff: They are, though. The tests failed in an indescribably unhelpful manner because this test overrides the `$window` service with a very crippled version. It took me hours to find this. Trust me, this is very related. :expressionless: 

Also again: I think the "boy scout rule" should apply to this as well. :innocent:

staubesv: Ok, in that case the change is related to the PR then ;-)

modules/admin-ui/src/main/webapp/scripts/app.js
staubesv: Actually you do not need to dive into design work at all: The Admin UI comes with all the required styles - just re-use them.

modules/admin-ui/src/main/webapp/styles/main.scss
JulianKniephoff: I will have to think about this again; maybe I did something wrong. But for me, that stylesheet was not included in the production build, while it worked fine in "proxy mode".

I will get back to you. :wink:

modules/admin-ui/src/main/webapp/scripts/app.js
JulianKniephoff: Deciding which styles to use for what, though, is still a kind of design. :wink:

Anyway, I will give it a go soon.

JulianKniephoff:
@staubesv first of all thanks for the review! :slightly_smiling_face:

~~I will address the remaining two issues (design of the cheat sheet and the stylesheet inclusion issue) soon, it might have to wait until next week, though.~~

* [x] Fix the vendor stylesheet inclusion
* [x] Properly style the cheatsheet

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
JulianKniephoff: Sure, I see your point and will heed your advice in the future. However, I disagree with the notion that more or less artificial and arbitrary (not meant in a bad way, but in the literal sense) bureaucratic hurdles are an indication of the worth of a change. Sure, **this** particular change does not matter at all, but in general and in the aggregate code quality improvements should be encouraged, not encumbered by heavy weight processes. Plus I don't really see the added value of isolating such changes to their own pull request versus to their own commit.

On the other hand I have to admit that I **do** care about git history cleanliness to a probably unhealthy degree. So, in the annotation tool, I collect changes like this in a `nitpicks` branch to keep them out of the feature branches. But there I also have the freedom to just merge that branch every now and then, without going through any kind of process. Inb4: yes that f*cked me over on occasion, already. :sweat_smile:

JulianKniephoff:
image

This is how it looks now. :smile:

Since I now had to create a completely custom template anyway, I also took the time to group the shortcuts somewhat sensibly and to internationalize the keys itself (ctrl, etc.), as I alluded to in the original comment.

This should be good to go now. At least for another review round, @staubesv. :wink:

staubesv:
@JulianKniephoff Wow!!! Will continue the review tomorrow - need to have this ;-)

modules/admin-ui/src/main/webapp/scripts/shared/partials/hotkeyCheatSheet.html
```diff
@@ -0,0 +1,28 @@
+<div ng-class="{'modal-animation modal-overlay': helpVisible}"></div>
+<div ng-controller="HotkeyCheatSheetCtrl" ng-show="helpVisible" class="modal modal-animation">
+  <header>
+    <a class="fa fa-times close-modal" ng-click="toggleCheatSheet()"></a>
+    <h2 translate="HOTKEYS.CHEAT_SHEET.TITLE"><!-- Keyboard Shortcuts --></h2>
+  </header>
+  <div class="modal-content active"><div class="modal-body"><div class="full-col">
```
staubesv: Any special reason to have both div on the same line?

modules/admin-ui/src/main/webapp/scripts/shared/partials/hotkeyCheatSheet.html
```diff
@@ -0,0 +1,28 @@
...
+                  {{ $last ? '' : '+' }}
+                </span>
+              </span>
+            </p>
+          </td>
+          <td translate="{{ 'HOTKEYS.DESCRIPTIONS.' + hotkey.description | uppercase }}"></td>
+        </tr>
+      </tbody></table>
+    </div>
+  </div></div>
```
staubesv: That looks weird (both div on same line)

modules/admin-ui/src/main/webapp/scripts/shared/services/hotkeysService.js
```diff
@@ -52,31 +52,27 @@ angular.module('adminNg.services')
...
-
-      this.activateUniversalHotkey = function (keyIdentifier, description, callback) {
-        me.loading.then(function () {
-          var key = me.keyBindings[keyIdentifier];
-          if (key !== undefined) {
+          if (key) {
             hotkeys.add({
               combo: key,
-              description: description,
+              description: keyIdentifier,
```
staubesv: That reads somewhat weird... is keyIdentifier now an identifier or a description?

modules/admin-ui/src/main/webapp/styles/views/modals/_hotkey-cheat-sheet.scss
```diff
@@ -0,0 +1,19 @@
+td.hotkey {
+    text-align: center;
+}
+
+// TODO This needs to be more specific!
```
staubesv: What do you mean by "needs to be more specific"? I suggest to either a) just remove the TODO comment, b) make it more specific and remove the TODO comment or c) create a JIRA issue and remove the TODO comment.

You see, I don´t like TODO comments ;-)

modules/admin-ui/src/main/webapp/scripts/shared/partials/hotkeyCheatSheet.html
JulianKniephoff: Just a stylistic choice. I don't like deep nesting, especially if the "layers" don't contribute much to the understanding of main purpose of the piece of code (see also our discussion about single line `if`-s :wink:). This line just contains all the necessary `div`-s and `class`-es to make the whole thing "modal-content-like". I can split them if you want, though; my opinion on this one is less rigid than the one on the `if` issue. x)

modules/admin-ui/src/main/webapp/scripts/shared/partials/hotkeyCheatSheet.html
JulianKniephoff: At least it is consistent with the above. :wink: 

By the way, what about the `<table><tbody>` in one line, is that okay? ^_^ \*is shoveling his own grave, basically\*

modules/admin-ui/src/main/webapp/scripts/shared/services/hotkeysService.js
JulianKniephoff: Weeell, from our point of view it is an identifier, which we use to look up translations and also arrange for the grouping. For `angular-hotkeys` it is a description, but having overridden the default cheat sheet template that interpretation is meaningless. `angular-hotkeys` only uses the description in it's template. It is however the only place to put custom information to identify a hotkey later on in our own template and controller.

So yes, it looks strange, but calling the `HotkeyService` parameter `description` would be a lie, too, and unfortunately I can't (reasonably) change the parameter name that `angular-hotkeys` uses. :unamused:

JulianKniephoff: I could add a comment, if that helps. :slightly_smiling_face:

modules/admin-ui/src/main/webapp/styles/views/modals/_hotkey-cheat-sheet.scss
JulianKniephoff: I don't either (well, at least in finished and published commits) and this was totally an oversight probably due to prolonged exposure to high temperatures. :wink:

staubesv:
@gregorydlogan & @lkiesow Please note that this build fails because of

>The command "sudo apt-get install -y bzip2 gzip hunspell hunspell-de-de procps sox tar tesseract-ocr tesseract-ocr-deu unzip" failed and exited with 100 during .

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1922,13 +1923,82 @@
...
+       },
+       "DESCRIPTIONS": {
+           "GENERAL": {
+               "MAIN_MENU": "Main menu",
+               "SELECT_NEXT_DASHBOARD_FILTER": "Select next dashboard filter",
+               "SELECT_PREVIOUS_DASHBOARD_FILTER":  "Select previous dashboard filter",
+               "REMOVE_FILTERS": "Remove filters",
+               "EVENT_VIEW": "Open events table",
+               "SERIES_VIEW": "Open series table",
+               "NEW_EVENT": "Create new event",
```
staubesv: "Add event" and "Add series" would be more consistent with the labels of the buttons that do the same think.

What do you think?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1922,13 +1923,82 @@
...
+               "NEXT_SEGMENT": "Next segment",
+               "VOLUME_UP": "Volume up",
+               "VOLUME_DOWN": "Volume down",
+               "MUTE": "Mute/unmute"
+           },
+           "EDITOR": {
+               "SPLIT_AT_CURRENT_TIME": "Split the video at current time",
+               "CUT_SELECTED_SEGMENT": "Remove current segment",
+               "PLAY_CURRENT_SEGMENT": "Play current segment",
+               "CLEAR_LIST": "Clear segment list",
```
staubesv: The respective button label is "Clear segments".

staubesv (changes requested):
Hi @JulianKniephoff 

I've done with the next round with the review. I've found three issue:

1. The cheat sheet always displays capital letter which is problematic as sometimes, both lowercase and uppercase are defined as keyboard shortcut (but are indistinguishable in the cheat sheet)
2. If you move to the video editor and then back to the events table, the keyboard shortcut "Create new event" is lost
3. If the cheat sheet is displayed, the user can, for example, open the Add Series Wizard using a hotkey. The wizard works but is behind the cheat sheet. You should close the cheat sheet here.

Additionally, I have some suggestions. Implementing those is not a prerequisite for this review to pass, but I think they are worth mentioning:

1. Names of keyboard shortcuts should match button labels.

Example: On the series table, there is a button "Add series". There is also a keyboard shortcut for this function, but its name is "Create new series". That seems suboptimal.

2. Table row height
For sake of consistency and vertical space usage, it might be desirable to have the same row height as, for example, the events table. The current height is also a bit "much" from an graphical point of view.

3. For a newbie, it might not be obvious that the cheat sheet displays different keyboard shortcuts depending on where the cheat sheet is opened. Maybe add a text like "This page lists the currently available keyboard shortcuts. Note that the available keyboard shortcuts depend on the current page".
This would also help from a graphical point of view to move a little away from the very strict "box-in-box-in-table-in-box..." layout the Admin UI tends to have.

5. Now we are getting picky: The background color of the "keys" is somewhat dark. You could use the same color as the modal title does (a little lighter) to avoid the usage of many, many just slightly different grays in the Admin UI

6. The key symbol for the space key seems not clear to me: I had to look up the configuration to recognise that it means "SPACE". Maybe it would be better to just say "SPACE" here.

JulianKniephoff:
@staubesv point 2 of your first three points is actually not a bug of the cheat sheet. The cheat sheet actually **reveals** a bug that was already there in this case, which I quite enjoy, to be frank. :wink: The "New event" shortcut is registered by the `ApplicationCtrl` only once at the start of the application. When the editor overrides the shortcut for this command, the original one is lost until the next reload, because going back to the events table does not execute the application controller again. So whether or not the cheat sheet displays it does not really matter here. after the sequence of steps you described, the "new event" shortcut won't work anymore; the cheat sheet actually tells the truth.

I think this is a separate issue, which I will gladly create in JIRA and provide a PR for. However, I'm not quite sure what the fix actually **is**. The "new event" key is registered as a global hotkey, meaning that it should not only work on the events table, but also for example in the editor, actually. So step 0 would probably be to make the shortcuts for the two conflicting commands here unique. Beyond that one might think about whether that global behavior is actually what we want. If not, the fix is trivial. If so, I'd have to think about it a bit harder.

What do you think?

staubesv:
If the problem is not related to this PR, just ignore it in this PR ;-)

staubesv:
... not sure what would be best, but I would tend to say that "Add event" and "Add series" should not necessarily be global shortcuts - in our case, we even have explicitly disabled them as we had users that accidentally pressed such a shortcuts while editing the video which resulted in all work being lost (meanwhile, it would maybe work as the video editor warns about unsaved changes).

But I suggest to not address this in this PR as it is a very different topic....

JulianKniephoff:
Alright, I'll create a JIRA issue for further discussion and then ignore that point for this review.

**EDIT**: Here it is: [MH-12947](https://opencast.jira.com/browse/MH-12947).

staubesv:
Hi @JulianKniephoff I have to admit that I'm not sure anymore about "issue 3": At the end of the day, the user can also get into this situation when opening the cheat sheet while having a modal window open... maybe it's best to just ignore that for now and see whether it results in weird situations.

The only "issue" left then is issue 1 (lowercase & uppercase).

JulianKniephoff:
Regarding point 1, we spoke briefly about the possibility of rendering `N` as `shift + n`, but I now decided to just leave it at `N` and render `n` as ... well `n`. :wink: The reason is things like `?`. On a German keyboard, `?` would be `shift+ß`, but on a US keyboard it would be `shift+/` but finding out which one it has to be is hard to impossible in JavaScript. While this might be clear for the letters (although I don't know that for sure for all possible keyboard layouts ...), I don't like the distinction.

That said, the rendering of the cheat sheet is not the only piece of code playing into this decision. If a hotkey is **configured** as `shift+n`, it will also be rendered as such! This is fine though as long as we keep the configuration consistent, I think. Also note that there is a subtle difference to configuring a shortcut to be something like `?` vs. `shift+/`.

I will push that change with a few of the nitpicky things in a few moments.

JulianKniephoff:
Alright, I think I got everything. This is how it looks now.

image

Note that I configured some shortcuts differently for testing purposes and that this is running in the proxy, which has less shortcuts anyway. The "add event" shortcut shows a feature of Moustrap that we don't currently use, but it would totally work: Multiple shortcuts and even key combinations (vs. key chords) per combination. That means (in this case), that ehiter pressing `g` and then `n` **or** first `ctrl+k` and then `v` would open the new event modal.

I think this should be good to go now @staubesv. :smile:

staubesv:
Seems reasonable to me to just use the upper-/lowercase letters to avoid the need to deal with keyboard layouts. I'm continuing the review...

staubesv (approved):

staubesv:
@JulianKniephoff Thanks for patiently implementing a lot of suggestions - cool feature! As promised, you get a swift review in turn ;-)

JulianKniephoff:
And thank you @staubesv for the detailed review, I really appreciate that! :smile:

PR #270, opened 2018-05-29
MH-12928, Mitigation for KARAF-5526

lkiesow:
This patch introduces a mitigation for KARAF-5526 which causes the
interactive Karaf shell to fail for some terminals. This patch can be
removed once we update Karaf again but it seems like Karaf 4.0.11 will
not be released and updating to a new major version is no bugfix.

Note that for a major Karaf update, we likely need to update our
assembly step anyway since they made somemodifications to their scripts
and assembly plugin meaning this is unlikely to stay in our code
forever.

[KARAF-5526] https://issues.apache.org/jira/browse/KARAF-5526

### DIFF
--- a/assemblies/pom.xml
+++ b/assemblies/pom.xml
@@ -131,6 +131,8 @@
                   <replaceregexp file="target/assembly/bin/karaf" match="-Dcom.sun.management.jmxremote" replace="" byline="true"/>
                   <!-- Correct the start script name -->
                   <replaceregexp file="target/assembly/bin/karaf" match="^KARAF_SCRIPT=.*$" replace="KARAF_SCRIPT='start-opencast'" byline="true"/>
+                  <!-- Mitigation for https://issues.apache.org/jira/browse/KARAF-5526 -->
+                  <replaceregexp file="target/assembly/bin/karaf" match=" init$" replace=" [[ &quot;$TERM&quot; == *-256color ]] &amp;&amp; export TERM=xterm-color&#x0A;    init" byline="true"/>
                   <!-- Make sure bundle cache gets cleared by default -->
                   <replaceregexp file="target/assembly/etc/system.properties" match="^karaf.clean.cache .*=.*$" replace="karaf.clean.cache = true" byline="true"/>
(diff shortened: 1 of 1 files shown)


wsmirnow:
On my workstation the `TERM` is set to `xterm-256color` per default. Changing the value to `xterm-color` works for me.

lkiesow:
@wsmirnow, I adapted the patch to catch all termspecs ending with `-256color`.

Note for testers: Try setting your term to `xterm-256color` and start Opencast. Without this patch, you will get an error and will be unable to use the interactive shell:

    export TERM=xterm-256color
    ./bin/start-opencast

wsmirnow (approved):

gregorydlogan:
I'm not seeing the error either with or without the patch.  It also doesn't break anything though, so I'm going to just merge it as is.

gregorydlogan:
I noticed this morning that inside my tmux session I'm seeing this:
```
./build/opencast-dist-develop-6-SNAPSHOT/bin/start-opencast: 560: ./build/opencast-dist-develop-6-SNAPSHOT/bin/start-opencast: [[: not found
```
at startup.  TERM is set to ```screen``` and it doesn't seem to break anything, but I guess I missed this in my initial testing...

lkiesow:
Hi Greg,
`[[` is bashism and you are probably using dash then. I guess I should
not have tried to fix Waldemar's problem as well and introduce that.
I'll fix that until Monday though it should have no negative side
effects apart from looking ugly…
Regards,
Lars

gregorydlogan:
That's what surprised me, I'm definitely using Bash, unless Debian is pulling an Ubuntu and renaming things...

```
greg@np9150:~$ echo $SHELL
/bin/bash
greg@np9150:~$ /bin/bash --help
GNU bash, version 4.4.12(1)-release-(x86_64-pc-linux-gnu)
```

Doesn't look like it, unless they're *really* screwing around with things.

lkiesow:
You use bash but the script does not necessarily use bash :)

lkiesow:
Fixed by pull request #288

PR #271, opened 2018-05-30
MH-12929 Change paella URL to /paella/ui

miesgre:
Paella player external bundle URL was in /paella/ui/watch.html.

When bundle moved inside opencast, URL was changed to /engage/paella/ui/watch.html
This is causing problems to people using paella in opencast 4.x when migrating to 5.0, because the URL change (Old mediapackages has URL /paella/ui/watch.html)

This PR change the paella URL to /paella/ui/watch.html to be compatible with OC 4.x and external paella bundle.

### DIFF
--- a/docs/guides/admin/docs/modules/paella.player.md
+++ b/docs/guides/admin/docs/modules/paella.player.md
@@ -10,5 +10,5 @@ Select the Paella Player
 
 To activate the paella player set:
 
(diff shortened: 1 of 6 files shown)


rrolf:
You missed the reference in etc/org.opencastproject.organization-mh_default_org.cfg

miesgre:
I forgot to add these files to the PR. Sorry :(

PR #272, opened 2018-05-30
MH-12922 Job load fixes

rute-santos:
Fixes:
* Fix comparison by host name, which was causing the current host load to be always 0. 
* Make sure jobs are not removed/counted twice from local load cache by first checking their presence in it. 
* Queued jobs are not considered when calculating the current load on node. The previous version would count them sometimes (when job children e.g. inspect), sometimes not (when parent job e.g. compose, etc).
* Removed job from cache when deleted or not found in db. Unexpected problems could cause the job to stay in the local cache and be counted in the current load forever.

Additional Notes:
- isReadyToAccept() should be "synchronized" across all services i.e. the decision that the job can be accepted by a node and putting it into the node's load cache should be atomic.
Since the dispatcher only runs on admin and is sequential, isReadyToAccept() will not be called simultaneously by two separate threads so this is currently fine.
This will change if we distribute the dispatcher code or if it's configured to run on > 1 node.

- When a parent jobs goes from WAITING->RUNNING after its children are finished, it doesn't go through the isReadyToAccept() logic and is put back directly into the node's load cache so, at a point in time, the node load may be > max load.
I considered this OK because normally the parent job is just finishing up after all its children are done.

- For testing, I suggest turning the DEBUG logging on to see what's happening. Please keep in mind that the node load, being a float number, will not show as exact as you expect :)

### DIFF
--- a/docs/guides/admin/docs/configuration/load.md
+++ b/docs/guides/admin/docs/configuration/load.md
@@ -50,32 +50,10 @@ system.  If you wish to change this, set the `org.opencastproject.server.maxload
 maximum load you want this node to accept.  Keep in mind that exceeding the number of CPU cores present in the system is
 not recommended.
 
-The load values for the non-encoding jobs are set in the etc/services files.  Look for files containing the prefix
-`job.load`.  These configuration keys control the load for each job type.  For example, the
-`job.load.download.distribute` configuration key controls the load placed on the system when a download distribution job
-is running.  The current files with relevant configuration keys are:
-
-| File                                                                                     | Controls                           |
-|------------------------------------------------------------------------------------------|------------------------------------|
-| org.opencastproject.animate.impl.AnimateServiceImpl.cfg                                  | Animate service                    |
-| org.opencastproject.caption.impl.CaptionServiceImpl.cfg                                  | Caption convertion services        |
-| org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg                 | AWS S3 distribution service        |
-| org.opencastproject.distribution.download.DownloadDistributionServiceImpl.cfg            | Download distribution              |
-| org.opencastproject.distribution.distribution.streaming.<br/>StreamingDistributionService.cfg | Streaming distribution             |
-| org.opencastproject.distribution.streaming.wowza.<br/>WowzaAdaptiveStreamingDistributionService.cfg | Adaptive streaming distribution |
-| org.opencastproject.execute.impl.ExecuteServiceImpl.cfg                                  | Execute service                    |
-| org.opencastproject.ingest.impl.IngestServiceImpl.cfg                                    | Ingest services                    |
-| org.opencastproject.inspection.ffmpeg.MediaInspectionServiceImpl.cfg                     | Media inspection using ffmpeg      |
-| org.opencastproject.publication.youtube.YouTubeV3PublicationServiceImpl.cfg              | Youtube distribution               |
-| org.opencastproject.search.impl.SearchServiceImpl.cfg                                    | Opencast engage index jobs         |
-| org.opencastproject.silencedetection.impl.SilenceDetectionServiceImpl.cfg                | Silence detection                  |
-| org.opencastproject.sox.impl.SoxServiceImpl.cfg                                          | Sox service                        |
-| org.opencastproject.textanalyzer.impl.TextAnalyzerServiceImpl.cfg                        | Text analysis, including slide OCR |
-| org.opencastproject.timelinepreviews.ffmpeg.TimelinePreviewsServiceImpl.cfg              | Timeline previews service          |
-| org.opencastproject.transcription.ibmwatson.IBMWatsonTranscriptionService.cfg            | IBM Watson start transcription job |
-| org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg                          | Video editor                       |
-| org.opencastproject.videosegmenter.ffmpeg.VideoSegmenterServiceImpl.cfg                  | Video segmentation                 |
-| org.opencastproject.waveform.ffmpeg.WaveformServiceImpl.cfg                              | Waveform generation for video editing |
+The load values for the non-encoding jobs are set in the configuration files in the `etc` directory.  Search the
+this directory for files that contain the string `job.load` to find the relevant configuration keys.  These
+configuration keys control the load for each job type.  For example, the `job.load.download.distribute` configuration
+key controls the load placed on the system when a download distribution job is running.
 
 Note: Ingest jobs are a special case in Opencast.  Because of their immediate nature there is no way to limit the number
 of running jobs.  However, these jobs will block other jobs from running on the ingest/admin nodes if enough ingests
--- a/docs/guides/admin/docs/configuration/load.md
+++ b/docs/guides/admin/docs/configuration/load.md
@@ -50,8 +50,8 @@ system.  If you wish to change this, set the `org.opencastproject.server.maxload
 maximum load you want this node to accept.  Keep in mind that exceeding the number of CPU cores present in the system is
 not recommended.
 
-The load values for the non-encoding jobs are set in the configuration files in the `etc` directory.  Search the
-this directory for files that contain the string `job.load` to find the relevant configuration keys.  These
+The load values for the non-encoding jobs are set in the configuration files in the `etc` directory.  Search this
+directory for files that contain the string `job.load` to find the relevant configuration keys.  These
 configuration keys control the load for each job type.  For example, the `job.load.download.distribute` configuration
 key controls the load placed on the system when a download distribution job is running.
 
--- a/docs/guides/admin/docs/configuration/load.md
+++ b/docs/guides/admin/docs/configuration/load.md
@@ -37,7 +37,7 @@ job.
 
 Note: These job loads are specific for each *node* in the cluster.  This means that for any given job, each node can
 have a different load value associated.  For instance, if worker A has no job load specified for its encoding profiles,
-and worker B has job loads specified then any encoding jobs created by A will have the default load (0.6), and jobs
+and worker B has job loads specified then any encoding jobs created by A will have the default load (0.8), and jobs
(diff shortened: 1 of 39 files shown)


modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
```diff
@@ -89,7 +89,7 @@
   protected HashMap<String,String> suffixes = new HashMap<String, String>();
 
   @XmlElement(name = "jobLoad")
-  protected Float jobLoad = 1.0f;
+  protected Float jobLoad = 0.6f;
```
gregorydlogan: Is the choice of 0.6f equally as magical as the choice of 1.0f?

gregorydlogan (approved):

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
rute-santos: Yes, it is :) For a default, I think it's ok.

gregorydlogan: To be clear, this change changes *all* encoding to default to 0.6f.  So an 8 core machine with N workflows running now will attempt to process ~13 workflows at once if they're all encoding at the same time.  This seems high, which is why I chose 1.0f initially.  I fully support dropping down the default for the editing step, but I'm curious why we are dropping the cost of all encoding?

docs/guides/admin/docs/configuration/load.md
```diff
@@ -57,21 +57,25 @@ is running.  The current files with relevant configuration keys are:
...
 | org.opencastproject.search.impl.SearchServiceImpl.cfg                                    | Opencast engage index jobs         |
 | org.opencastproject.silencedetection.impl.SilenceDetectionServiceImpl.cfg                | Silence detection                  |
+| org.opencastproject.sox.impl.SoxServiceImpl.cfg                                          | Sox service                        |
 | org.opencastproject.textanalyzer.impl.TextAnalyzerServiceImpl.cfg                        | Text analysis, including slide OCR |
+| org.opencastproject.timelinepreviews.ffmpeg.TimelinePreviewsServiceImpl.cfg              | Timeline previews service          |
+| org.opencastproject.transcription.ibmwatson.IBMWatsonTranscriptionService.cfg            | IBM Watson start transcription job |
 | org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg                          | Video editor                       |
 | org.opencastproject.videosegmenter.ffmpeg.VideoSegmenterServiceImpl.cfg                  | Video segmentation                 |
+| org.opencastproject.waveform.ffmpeg.WaveformServiceImpl.cfg                              | Waveform generation for video editing |
```
lkiesow: I'm wondering if we should just tell people to grab for `job.load`. I suspect this list may constantly be outdated.

etc/org.opencastproject.animate.impl.AnimateServiceImpl.cfg
```diff
@@ -3,5 +3,5 @@
 #synfig.path=synfig
 
 # Job load for an animation job
-# Default: 1.0
-#job.load.animate=1.0
+# Default: 0.1
```
lkiesow: animate internally uses FFmpeg to encode a video file. 0.1 seems a bit low for that, even though it's just a single stream encoding.

modules/common/src/main/java/org/opencastproject/job/api/AbstractJobProducer.java
```diff
@@ -173,22 +173,32 @@ public boolean isReadyToAccept(Job job) throws ServiceRegistryException, Undispa
...
 
     /* Note that this first clause looks at the *job's*, the other two look at the *node's* load
      * We're assuming that if this case is true, then we're also the most powerful node in the system for this service,
      * per the current job dispatching code in ServiceRegistryJpaImpl */
     if (job.getJobLoad() > maxload.getLoadFactor() && acceptJobLoadsExeedingMaxLoad) {
-      logger.warn("Accepting job {} of type {} with load {} even though load of {} is above this node's limit of {}.",
-              new Object[] { job.getId(), job.getJobType(), df.format(job.getJobLoad()), df.format(currentLoad), df.format(maxload.getLoadFactor()) });
+      logger.warn(
+              "{} Accepting job {} of type {} with load {} even though load of {} is above this node's limit of {}.",
+              new Object[] { Thread.currentThread().getId(), job.getId(), job.getJobType(), df.format(job.getJobLoad()),
```
lkiesow: Don't use an array here. Just remove the `new Object[] { }` and it will still work.

modules/common/src/main/java/org/opencastproject/job/api/AbstractJobProducer.java
```diff
@@ -173,22 +173,32 @@ public boolean isReadyToAccept(Job job) throws ServiceRegistryException, Undispa
     // Note: We are not adding the job load in the next line because it is already accounted for in the load values we
     // get back from the service registry.
     float currentLoad = getServiceRegistry().getOwnLoad();
+    logger.debug("{} Current load on this host: {}, job's load: {}, job's status: {}, max load: {}", new Object[] {
```
lkiesow: Don't use an array here. Just remove the `new Object[] { }` and it will still work.

modules/common/src/main/java/org/opencastproject/job/api/AbstractJobProducer.java
```diff
@@ -173,22 +173,32 @@ public boolean isReadyToAccept(Job job) throws ServiceRegistryException, Undispa
...
+              new Object[] { Thread.currentThread().getId(), job.getId(), job.getJobType(), df.format(job.getJobLoad()),
+                      df.format(currentLoad), df.format(maxload.getLoadFactor()) });
       logger.warn("This is a configuration issue that you should resolve in a production system!");
       return true;
     } else if (currentLoad > maxload.getLoadFactor()) {
-      logger.debug("Declining job {} of type {} with load {} because load of {} would exceed this node's limit of {}.",
-              new Object[] { job.getId(), job.getJobType(), df.format(job.getJobLoad()), df.format(currentLoad), df.format(maxload.getLoadFactor()) });
+      logger.debug(
+              "{} Declining job {} of type {} with load {} because load of {} would exceed this node's limit of {}.",
+              new Object[] { Thread.currentThread().getId(), job.getId(), job.getJobType(), df.format(job.getJobLoad()),
```
lkiesow: Don't use an array here. Just remove the `new Object[] { }` and it will still work.

modules/common/src/main/java/org/opencastproject/job/api/AbstractJobProducer.java
```diff
@@ -173,22 +173,32 @@ public boolean isReadyToAccept(Job job) throws ServiceRegistryException, Undispa
...
+                      df.format(currentLoad), df.format(maxload.getLoadFactor()) });
       return false;
     } else  {
-      logger.debug("Accepting job {} of type {} with load {} because load of {} is within this node's limit of {}.",
-            new Object[] { job.getId(), job.getJobType(), df.format(job.getJobLoad()), df.format(currentLoad), df.format(maxload.getLoadFactor()) });
+      logger.debug("{} Accepting job {} of type {} with load {} because load of {} is within this node's limit of {}.",
```
lkiesow, line 199: …same here

lkiesow (commented):
Looks good. I only found a few minor things.

docs/guides/admin/docs/configuration/load.md
gregorydlogan: "Search the contents of the configuration directory (generally /etc/opencast) for `job.load` to find the relevant configuration keys"?

Seems reasonable to me.  We can already see how outdated this is...

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
rute-santos: Hi @gregorydlogan, I just changed because of Lars request to change all the default job loads to < 1 and I tried to follow the relative load among different services (e.g. video editor > encode > inspect, etc). I am fine with changing it to whatever you think is best. For us, we are going to use 4.0 and 8.0 (for multi-profile operations). We have 32-core systems.

docs/guides/admin/docs/configuration/load.md
rute-santos: I agree and will make the change.

etc/org.opencastproject.animate.impl.AnimateServiceImpl.cfg
rute-santos: Suggestion?

lkiesow: Whatever you and Greg settle for as default for encoding ;-)

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
gregorydlogan: Fair enough, @lkiesow I can't find the thread where you made that request, what was the reasoning again?

etc/org.opencastproject.animate.impl.AnimateServiceImpl.cfg
rute-santos: @gregorydlogan, which load value do you suggest?

gregorydlogan: I would suggest 1.0, but I'm fine with 0.6.  We could think about turning off the oversize job acceptance, but that conversation should probably just be rolled into the admin refactoring.

rute-santos: I will update to 0.8 then. Still under 1 but not too low.

PR #273, opened 2018-05-31
NOJIRA: Update the release process docs

gregorydlogan:
Updates the release process documentation, removing a defunct script and making it more clear how we should be creating the changelogs.

### DIFF
--- a/docs/guides/developer/docs/release-manager.md
+++ b/docs/guides/developer/docs/release-manager.md
@@ -315,54 +315,47 @@ assume the final release should be based on `3.0-rc2`.
 
         git checkout -b r/3.0
 
-4. Ensure the fix versions of the relevant tickets are correct.  Search GitHub's pull request list with a query like
-   this, modifying the dates and base to match the start date of the current release, and the current date
-
-        is:pr base:r/3.x is:closed merged:2018-03-29..2018-05-31
-
-   For each of these tickets, make sure the fix version is set to the new release in JIRA, then use the
-   `create-changelog` [helper script](https://github.com/opencast/helper-scripts/tree/master/create-changelog) to
-   easily format the list.
-
-5. Add release notes, and update the changelog, then commit them:
+4. Add release notes, and update the changelog using the `create-changelog`
+   [helper script](https://github.com/opencast/helper-scripts/tree/master/create-changelog) to
+   easily format the list, the commit:
 
         vim docs/guides/admin/docs/releasenotes.md docs/guides/admin/docs/changelog.md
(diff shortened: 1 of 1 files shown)


docs/guides/developer/docs/release-manager.md
```diff
@@ -315,83 +315,56 @@ assume the final release should be based on `3.0-rc2`.
 
         git checkout -b r/3.0
 
-4. Add release notes, and update the changelog, then commit them:
+4. Ensure the fix versions of the relevant tickets are correct.  Search GitHub's pull request list with a query like
+   this, modifying the dates and base to match the start date of the current release, and the current date
+
+        is:pr base:r/3.x is:closed merged:2018-03-29..2018-05-31
+
+   For each of these tickets, make sure the fix version is set to the new release in JIRA, then use the
```
lkiesow: I'm -1 for making release managers responsible for the accuracy of Jira tickets' metadata. For major versions, people would possibily need to go trough hundreds of tickets.

lkiesow (changes requested):

docs/guides/developer/docs/release-manager.md
lkiesow: If you really want this to happen, you could automate this and write a script which acts upon merged pull requests and updates the metadata in Jira.

gregorydlogan: It seems to me like there's a lot of pushback in the community for keeping JIRA, so I'm of the opinion that if we're going to keep it we should make the effort to make sure it's *correct* as well.  I'll try and get that script written in time for 5.0  In the meantime let's hold this I guess.

lkiesow: My point is that those who want to keep it should do the work and not those who already spend time on community tasks by volunteering as release managers. If it gets automated that's good. I'm not against correct metadata ;-P

As an alternative to the merge hook, you could also have the script run on the CI machines, automatically executing on builds for release branches.

gregorydlogan: One of those is definitely what we'll do.  When I'll have the time to write the script is a good question though :)

lkiesow:
@gregorydlogan, whatever the final automation will look like, this documentation change will be obsolete and we could close the issue, couldn't we?

Also, note that the current changelog scripts are not based on `fixVersion` anymore but on pull requests. Hence we could, based on that we know that `fixVersion` is mostly broken, also drop the Jira ticket field if we want to. But that's a discussion separate from this pull request.

gregorydlogan:
There's the bit about the Github vs BitBucket artifact upload at the bottom.  I've rolled the fixversion bits out, so this should be good to go now I think?

PR #274, opened 2018-06-01
MH-12930: Fill creator metadata field with actual us…

jchssystems:
Added a new createdby field to the adminui. The metadata field creator exists already, so no additional configuration is required. See similar implementation for series.

The field holds the logged in user in new event wizard.


modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventUtils.java
```diff
@@ -44,175 +45,190 @@
...
-   */
-  public static final Fn<Publication, Boolean> internalChannelFilter = new Fn<Publication, Boolean>() {
-    @Override
-    public Boolean apply(Publication a) {
-      if (InternalPublicationChannel.CHANNEL_ID.equals(a.getChannel()))
-        return false;
-      return true;
-    }
-  };
+    /** The logging facility */
```
staubesv, line 48: Opencast uses 2 spaces for indentation in Java code (see https://docs.opencast.org/develop/developer/code-style/). Please revert this back to 2 spaces.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -239,6 +240,8 @@
   /** The default time before a piece of signed content expires. 2 Hours. */
   protected static final long DEFAULT_URL_SIGNING_EXPIRE_DURATION = 2 * 60 * 60;
 
+  private static final int CREATED_BY_UI_ORDER = 16;
```
staubesv, line 243: The set of metadata fields and their display order is fully configurable. Therefore, hard-coding the display order of a specific field seems quite smelly. If you really need an explicit display order for this field, it might worth consider to configure the field "createdBy" in the metadata catalog configuration (as readonly) and introducing a property type that indicates that the value won't be included in Dublincore catalog itself.

staubesv (changes requested):

staubesv:
Hi @jchssystems 

I've finished a first round of review. Here my comments:

Most importantly, this PR does not work because the "Admin UI only field" has no notion of persistency: Its value is only stored in the Admin UI search index, but if you rebuilt that search index, the information about the creator is lost (the same holds for the existing series creator, btw).
As search index rebuilds are mandatory sometimes, that means that the creator is not persistent.

Besides that major comment, here some smaller issues:

- Opencast does use 2 spaces for indentation (please don't change to 4 spaces)
- Mixing up a hard-coded display order vs. fully configurable display order should be avoided
- You don't need to include the language translations here as they will be overwritten next time we download the translations from Crowdin. Only en-US is required.
- This PR currently does not build because some Admin UI unit tests need to be adapted to your changes.
- Why do you display the field in the New Event Wizard? Doesn't seem to be relevant there (and would not be consistent to the New Series Wizard)

That's all for now. 

How do you proceed?

jchssystems:
Thanks or reviewing.

>>    Opencast does use 2 spaces for indentation (please don't change to 4 spaces)
>>    This PR currently does not build because some Admin UI unit tests need to be adapted to your changes.
Yes, Sorry. I forgot to execute maven with tests and checkstyle.

>>    Mixing up a hard-coded display order vs. fully configurable display order should be avoided
Yes, I think so too. It is best to have the field in the configuration but we have really a metadata field, "creator", that is persistent (or I think so). My new UI field only copies the value from "creator" to "createdBy". I think it works (it is persistent), but you are right, it is better IMHO that the field "creator" be in configuration:
# Creator
property.creator.inputID=creator
property.creator.label=EVENTS.EVENTS.DETAILS.METADATA.CREATED_BY
property.creator.type=text
property.creator.readOnly=true
property.creator.required=false
property.creator.order=16

But I have problems because "presenters" use it (I don't understand why), and besides as "mixed_text". I have made tests with creator in the configurations file, without presenters, and goes well.

>>    You don't need to include the language translations here as they will be overwritten next time we download the translations from Crowdin. Only en-US is required.
Ok, good, I didn't knew that.

>>    Why do you display the field in the New Event Wizard? Doesn't seem to be relevant there (and would not be consistent to the New Series Wizard)
You are right. It is not a problem.


I think that a new field to store the creator of the event is not a good solution (we have already the "creator"). And if I define the "creator" in the configuration, we must see if "presenters" use it well. And the third solution is to allow the UI field as series (if it works, as I think).

jchssystems:
Thanks or reviewing.

> Opencast does use 2 spaces for indentation (please don't change to 4 spaces)
>  This PR currently does not build because some Admin UI unit tests need to be adapted to your changes.

Yes, Sorry. I forgot to execute maven with tests and checkstyle.

> Mixing up a hard-coded display order vs. fully configurable display order should be avoided

Yes, I think so too. It is best to have the field in the configuration but we have really a metadata field, "creator", that is persistent (or I think so). My new UI field only copies the value from "creator" to "createdBy". I think it works (it is persistent), but you are right, it is better IMHO that the field "creator" be in configuration:

#Creator
property.creator.inputID=creator
property.creator.label=EVENTS.EVENTS.DETAILS.METADATA.CREATED_BY
property.creator.type=text
property.creator.readOnly=true
property.creator.required=false
property.creator.order=16

But I have problems because "presenters" use it (I don't understand why), and besides as "mixed_text". I have made tests with creator in the configurations file, without presenters, and works well.

> You don't need to include the language translations here as they will be overwritten next time we download the translations from Crowdin. Only en-US is required.

Ok, good, I didn't knew that.

>  Why do you display the field in the New Event Wizard? Doesn't seem to be relevant there (and would not be consistent to the New Series Wizard)

You are right. It is not a problem.

I think that a new field to store the creator of the event is not a good solution (we have already the "creator"). And if I define the "creator" in the configuration, we must see if "presenters" use it well. And the third solution is to allow the UI field as series (if it works, as I think).

staubesv:
Just had a look at the JIRA issue:

> When I create a new event, I would like to save the user, who has created it. This user must be the one, who is logged in.
> 
> This user is already in the creator dublincore metadata but could be shown in the wizards Create event and Event details.

No, this is not correct: The Dublincore field "creator" is used to store the "Presenters". The reason of the name mismatch is that Opencast wanted to have the "Presenters" metadata field and needed to map that (and all other metadata fields) to Dublincore. Dublincore "creator" was likely (I was not there at this time) the most appropriate field.
The technical creator (person who created event in Opencast rather than the person that created the actual content that event represents) is currently not made persistent as far as I know.

Anyway, the "does not work" can be reproduced like this:

1. Create a new event
2. Open Event Details->Metadata for the event created in step 1
3. Verify that "Created by" displays the user who has created the event in step 1
4. Rebuild the Admin UI search index
5. Open Event Details->Metadata for the event created in step 1

Actual result:
The "Created by" field is blank

Expected result:
The "Created by" field should still show the "technical" creator of the event, i.e. the user who was logged in in step 1.

Note that Series Details->Metadata->Created by has the same problem: It does not work as expected.

The problem is that DublinCore->Creator does not contain the technical creator but rather the comma-separated list of "presenters" and if you have not set any presenters, DublinCore->Creator is indeed empty.

With other words: Event Details->Metadata->Presenters does already display the value of DublinCore->Creator. But that is not the technical creator of the event.

jchssystems:
It is definitely not persistent.

And what do you think is the best solution?
to define a new field "_created by_" in the configuration and make it persistent?
Could I use it as if it were a Dublincore field? I mean, adding it to the DublinCore.java class, and so on.

Thanks in advance.

staubesv:
> to define a new field "created by" in the configuration and make it persistent?

DublinCore is a standard that defines a set of metadata fields and their semantics. Therefore, you cannot just add "created by" as that field simply is not part of the standard.
It is also questionable whether this kind of information ("who has pressed Add Event in Opencast") should be included in the metadata catalogs as those catalogs are distributed to other systems.

> And what do you think is the best solution?

Probably, this information should be kept in the database. Maybe the asset manager could store this as property which would be most easy to implement as the asset manager already supports this kind of persistency (of course, the code rebuilding the search index would need to consider that).

-> I recommend to bring up this topic on the Opencast Developer list - maybe somebody has a good idea or better knowledge of relevant parts of Opencast.

BTW: Sorry for the bad finding during the review... but probably better than having it not working later...

karendolan:
Hi @jchssystems, it is less confusing if standard dublincore fields are classified as dublincore fields in Opencast. Our site uses the episode dublincore "contributor" field in a similar way to what you appear to need. The use case of auto-assigning that field to the user uploading new content for series that do not already pre-assign a contributor (in our case) is interesting for self-service users. 

@staubesv do you see issues with a flexible use of the episode contributor field?

staubesv:
@jchssystems

Just wondering: What is the use case behind "created_by"? 

@karendolan 

I don't see a general issue behind a flexible use of the catalog fields. Indeed, they are configurable, so each Adopter can adapt to whatever makes sense. The only concern would be to change/add/remove fields that are specific to Adopters in a way which is more "hard-coded" rather than "configurable".

jchssystems:
@staubesv 
> Just wondering: What is the use case behind "created_by"?

The field is important in searches. To find out what events a person has created. Without the search option, the field is not very useful.

On the other hand I have found that the Dublincore "publisher" field is not used for events. The field definition on dublincore.org is:

"An entity responsible for making the resource available. Examples of a Publisher include a person, an organization, or a service".

I think this field is going well for this purpose. I have made tests and it looks like it is working well, including the searches. And it is persistent.

Could that be an option?

PR #275, opened 2018-06-04
MH-12932, Kernel Build Failure

lkiesow:
The kernel tests try to reach out to ok.com. If the domain name cannot
be resolved or cannot be reached via HTTP, the tests will fail:

```
[ERROR] testResourceClosing(org.opencastproject.kernel.security.TrustedHttpClientResourceClosingTest)  Time elapsed: 0.196 s  <<< ERROR!
org.opencastproject.security.api.TrustedHttpClientException: java.net.UnknownHostException: ok.com: Name or service not known
        at org.opencastproject.kernel.security.TrustedHttpClientResourceClosingTest.testResourceClosing(TrustedHttpClientResourceClosingTest.java:78)
Caused by: java.net.UnknownHostException: ok.com: Name or service not known
        at org.opencastproject.kernel.security.TrustedHttpClientResourceClosingTest.testResourceClosing(TrustedHttpClientResourceClosingTest.java:78)
```

This patch fixes that by connecting to the local server instead which
was started already anyway.


lkiesow:
@staubesv, note that this currently breaks all builds on all systems. A swift review would hence be nice ;-P

staubesv (approved):

PR #281, opened 2018-06-06
MH-12938, Fix NullPointerException if no flavor is set

lkiesow:
If the flavor of a track was not set, the admin ui facade caused a null
pointer exception while generating the JSON for the frontend.

*Work sponsored by SWITCH*


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -2270,7 +2270,9 @@ private URI signUrl(URI url) {
   private List<Field> getEventMediaPackageElementFields(MediaPackageElement element) {
     List<Field> fields = new ArrayList<>();
     fields.add(f("id", v(element.getIdentifier(), BLANK)));
-    fields.add(f("type", v(element.getFlavor().toString(), BLANK)));
+    if (element.getFlavor() != null) {
+      fields.add(f("type", v(element.getFlavor().toString(), BLANK)));
```
gregorydlogan: This completely drops the type field if the flavour is unset.  Shouldn't we add an else clause that sets the value to BLANK?  The second argument to v() is what to set in the case the value is null...

PR #284, opened 2018-06-07
MH-12941, Gracefully handle empty flavors

lkiesow:
While a media package element's flavor *should* always be set in the
scheduler, when it is not, the ingest for that event will fail. If the
capture agent which did record that event dutifully tries to upload the
event again later, this might force the ingest service repeatedly into
an error state.

This patch fixes the issue by not trying to merge elements with unset
flavors which makes sense in any case since the merge is based on
flavors and you cannot merge an element based on something which does
not exist.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -1335,18 +1335,19 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
       mergeMediaPackageMetadata(mp, scheduledMp);
       return mp;
     } catch (NotFoundException e) {
-      logger.debug("No scheduler mediapackage found with id {}, skip merging", mp.getIdentifier().compact());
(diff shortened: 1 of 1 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1335,18 +1335,19 @@ private MediaPackage mergeScheduledMediaPackage(MediaPackage mp) throws IngestEx
       mergeMediaPackageMetadata(mp, scheduledMp);
       return mp;
     } catch (NotFoundException e) {
-      logger.debug("No scheduler mediapackage found with id {}, skip merging", mp.getIdentifier().compact());
+      logger.debug("No scheduler mediapackage found with id {}, skip merging", mp.getIdentifier());
       return mp;
     } catch (Exception e) {
-      logger.error("Unable to get event mediapackage from scheduler event {}", mp.getIdentifier().compact(), e);
```
lkiesow, line 1341: We do not need to log the error. By throwing a new one, we get the log anyway.

gregorydlogan:
Looks good, restarting travis just to be sure.

PR #287, opened 2018-06-08
MH-12905 TEMPORARY Karaf config assembly workaround (KARAF-5693)

karendolan:
As discussed in last Tues devOps meeting and proposed on list earlier, this is the workaround for missing libs in some of the assembled config.properties after a build.

### DIFF
--- a/etc/config.properties
+++ b/etc/config.properties
@@ -1,0 +1,412 @@
+################################################################################
+#
+#    Licensed to the Apache Software Foundation (ASF) under one or more
+#    contributor license agreements.  See the NOTICE file distributed with
+#    this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


karendolan:
@lkiesow The CI checks are failing because the static workaround config.properties contains the "$" symbol, as in "${includes}" and " ${jre-${java.specification.version}}". Is there a way to omit this file in that etc dir check? They exist in the usual assembled etc/config.properties.

gregorydlogan:
@karendolan that check is looking for spaces at the end of lines.  Line 390, and 392 both have spaces at the end ;)

karendolan:
@gregorydlogan Thanks! I removed those 2 spaces (and squashed pushed).

PR #289, opened 2018-06-11
MH-12905 Opencast does not startup anymore

staubesv:
The original commit against r/5.x had broken branch develop. This is a newly generated config.properties file based on branch develop.

This file has been generated the following way:
1. rm etc/config.properties
2. mvn clean install -Pdev -DskipTests
3. Copied the custom comment into the newly generated config.properties


lkiesow:
Can you fix the trailing spaces in the configuration file?

staubesv:
Oh, haven't realised that the build broke because of those (automatically generated) trailing spaces... done.

PR #290, opened 2018-06-11
MH-12946 Fix summary of add-event-dialog

KatrinIhler:
Fix the summary of the add-event-dialog. In detail:
* show input settings for chosen capture agent in summary
* show start date for 'upload'
* don't show weekdays for 'upload' and 'schedule single'
* don't show end date for 'schedule single' and 'upload'
* fix presentation of end date
* make width of capture agent dropdown for scheduled events in 'source' tab consistent
* improve presentation of input settings for chosen ca in 'source' tab
* show end time for scheduled events
* fix typos in method names

Screenshots of source and summary tab each for 'upload', 'schedule single' and 'schedule multiple':
source_schedule_multiple
source_schedule_single
summary_schedule_multiple
summary_schedule_single
summary_upload

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
@@ -255,7 +255,7 @@
                             <td>{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.LOCATION' | translate }} <i class="required">*</i></td>
                             <td>
                             <select chosen
-                                data-width="'200px'"
+                                data-width="'203px'"
                                 tabindex="11"
                                 ng-disabled="wizard.step.checkingConflicts"
                                 ng-change="wizard.step.roomChanged(); wizard.step.checkConflicts()"
@@ -410,7 +410,7 @@
                             <td>
                             <select chosen
                                 ng-disabled="wizard.step.checkingConflicts"
-                                data-width="'100%'"
+                                data-width="'203px'"
                                 tabindex="19"
                                 ng-change="wizard.step.roomChanged(); wizard.step.checkConflicts()"
                                 data-disable-search-threshold="8"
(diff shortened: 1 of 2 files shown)


modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -427,7 +427,10 @@
            "ACL": "Access policy"
          },
          "SUMMARY": {
-           "CAPTION": "Summary"
+           "CAPTION": "Summary",
```
gregorydlogan: Bit late for changing the translations innit?  I don't think this can go into 5.x as is...

gregorydlogan (changes requested):

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
gregorydlogan: Otherwise tests look good.  My sole complaint is this translation change.

staubesv: @KatrinIhler It is indeed true that a proposal was accepted that forbids translation changes into release branches. Can you factor out the translation change and do that in a separate PR against develop? You could just do it without i18n support here - at the end of the day, it is too late to re-translate 5.x anyway...

KatrinIhler: You're both right, and I had a feeling you would complain about that. ;) Just wanted to know what you would suggest... I've decided to simply leave out the 'Input'-prefix in the summary completely instead of having it be untranslated here since that's pretty ugly and obvious in my opinion and it's still understandable what these table entries mean without it, and I'm gonna reintroduce it in a separated PR targeted at develop.

without_input_prefix

PR #291, opened 2018-06-11
MH-12948, Directly read XACML files

lkiesow:
The XACML authorization service passed configuration files to a JBoss
library to determine if a specific action is allowed for a user given an
XACML file. The configuration itself contained file locations specifying
where the library could load the XACML file. This caused Opencast to
repeatedly load the XACML attachments into the workspace, cleaning them
up again afterwards.

This patch now allows Opencast to directly read those files from the
working file repository by using Opencast's internal parser for rules
and actions instead.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
+++ b/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
@@ -37,6 +37,7 @@ import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.mediapackage.MediaPackageElementBuilderFactory;
 import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.MediaPackageException;
+import org.opencastproject.security.api.AccessControlEntry;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AclScope;
 import org.opencastproject.security.api.AuthorizationService;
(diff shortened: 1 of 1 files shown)


smarquard:
This looks like a good improvement, +1 especially for getting rid of the oddly-specific use of JBoss libraries.

lkiesow:
> +1 especially for getting rid of the oddly-specific use of JBoss libraries.

It's still in other parts of this module though… :-(

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -428,88 +418,37 @@ public boolean hasPolicy(MediaPackage mp) {
...
+      return false;
+    }
+    Attachment attachment = xacml.get();
+    AccessControlList acl;
+    try {
+      acl = XACMLUtils.parseXacml(workspace.read(attachment.getURI()));
+    } catch (XACMLParsingException | NotFoundException | IOException e) {
+      return false;
+    }
+    logger.error("acl: {}", acl);
```
lkiesow: this development log needs to be removed.

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -428,88 +418,36 @@ public boolean hasPolicy(MediaPackage mp) {
...
+    Option<Attachment> xacml = getXacmlAttachment(mp);
+    if (xacml.isNone()) {
+      return false;
+    }
+    Attachment attachment = xacml.get();
+    AccessControlList acl;
+    try {
+      acl = XACMLUtils.parseXacml(workspace.read(attachment.getURI()));
+    } catch (XACMLParsingException | NotFoundException | IOException e) {
+      return false;
```
gregorydlogan: Do we want to log this error?

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -428,88 +418,36 @@ public boolean hasPolicy(MediaPackage mp) {
...
-              return false;
-            }
-
-            PolicyDecisionPoint pdp = getPolicyDecisionPoint(xacmlPolicyFile);
-            FileUtils.deleteQuietly(xacmlPolicyFile);
-
-            return pdp.evaluate(requestCtx).getDecision() == XACMLConstants.DECISION_PERMIT;
+    Option<Attachment> xacml = getXacmlAttachment(mp);
+    if (xacml.isNone()) {
+      return false;
```
gregorydlogan: I'd be tempted to pop a debug or trace log here.

gregorydlogan (approved):
Approved, but I would add a couple of logging lines to make life easier.

lkiesow:
I've added additional logs to be able to trace the evaluation of XACML files in Opencast. I'm not entirely sure how the warn level log will turn out on a production system. It might spam the logs (or not ;-) but we can evaluate that over the next month. It will be pretty obvious if it happens and will do no harm.

PR #296, opened 2018-06-14
MH-12953, stop loading editor.json twice

JamesUoM:
The tools.html is specifying the ng-controller even though this is already configured by the routeProvider in app.js. This lead to there being two ToolsCtrls


staubesv (approved):

JamesUoM:
Good grief that was quick @staubesv !

staubesv:
Happy to get this fix, too ;-)

PR #297, opened 2018-06-14
t/MH 12954 wrong date format in coverimage file

uzk-rrzk:
Fix the wrong format that creates the date for a cover image

### DIFF
--- a/etc/branding/coverimage.xsl
+++ b/etc/branding/coverimage.xsl
@@ -98,7 +98,7 @@
           <xsl:value-of select="opencast:XsltHelper.split(metadata/description, 50, 3, true)" />
         </tspan>
         <tspan class="presentationdate" dy="12%" x="50%">
-          <xsl:value-of select="date:format-date(metadata/start, 'MMMMMMMMMM dd, YYYY, HH:mm')" />
+          <xsl:value-of select="date:format-date(metadata/date, 'MMMMMMMMMM dd, YYYY, HH:mm:ss')" />
         </tspan>
         <tspan class="license" dy="10%" x="50%">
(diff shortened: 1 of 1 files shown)


etc/branding/coverimage.xsl
```diff
@@ -98,7 +98,7 @@
           <xsl:value-of select="opencast:XsltHelper.split(metadata/description, 50, 3, true)" />
         </tspan>
         <tspan class="presentationdate" dy="12%" x="50%">
-          <xsl:value-of select="date:format-date(metadata/start, 'MMMMMMMMMM dd, YYYY, HH:mm')" />
+          <xsl:value-of select="date:format-date(metadata/date, 'MMMMMMMMMM dd, YYYY, HH:mm:ss')" />
```
staubesv: Hi @uzk-rrzk Are you sure that you want to display the seconds here? Seems somewhat over-accurate....

uzk-rrzk: This is exact the line that I found in the documentation :-)

staubesv: I'm not saying that it would be wrong, but yes: You can indeed chose - omitting :ss will work.

So the question is not whether the doc provides an example but rather whether you would like to see the seconds or not ;-)

I tend to say the seconds are quite useless, but I don't have a strong opinion on that ;-)

uzk-rrzk: I agree that seconds make no sense in this case.

lkiesow (approved):

gregorydlogan:
Grabbing this from @staubesv since it's ready to go and we're talking about cutting a 5.1 shortly.

PR #302, opened 2018-06-18
MH-12829 Make admin-ui statistics configurable

KatrinIhler:
This PR moves the definition of the event statistics displayed at the top-right of the admin UI to a listprovider configuration file so statistics can be removed, added or reordered.

For defining relative date spans like _today_ or _this month_ SugarJS is used, which can translate strings like 'the beginning of today' or 'the end of this month' to actual dates in the user's timezone.

screenshot from 2018-06-18 17-33-15

This PR also adds tests and documentation where deemed necessary.

_This work is sponsored by ETH._


lkiesow:
Seems like you forgot to add some components

staubesv:
Hi @KatrinIhler The reason the build fails is that we currently include the components installed using bower in the codebase. When you run bower install locally, they're in the bower_components folder which is excluded via .gitignore.

staubesv (changes requested):
Hi @KatrinIhler 

I've done a first round of review and found the following:

Why are you adding an endpoint to access a list provider?
The Admin UI already does fully support list providers using the ResourcesListResource resource.

KatrinIhler:
@staubesv Added components and removed my endpoint, but now I broke stuff, gonna fix tomorrow...

KatrinIhler:
@staubesv Fixed! :)

lkiesow:
Looking at the screenshot, this seems to break the UI since the numbers are not centered over the labels anymore (e.g. look at the `0` above `Finished`)

lkiesow:
Do we really want to add more filters by default? All of them are constantly making HTTP requests in the background. My guess would be that the current filters should suffice for most people and if they want more, they can now configure them. But then, I'm no sysop for opencast.

PR #303, opened 2018-06-18
MH-12778, use media package locks to prevent or notify of concurrent …

JamesUoM:
Use media package locks to prevent or notify of concurrent edits. Also notify if trying to edit without available previews.

This replaces the PR previous made by Tobias, with code clean up and fixes

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/api/MediaPackageLockService.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/api/MediaPackageLockService.java
@@ -1,0 +1,85 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.adminui.api;
+
+import org.opencastproject.index.service.impl.index.event.Event;
+import org.opencastproject.security.api.User;
+
+/**
+ *
+ * @author Tobias M Schiebeck
+ */
+public interface MediaPackageLockService {
+  /**
+   * Minimum duration the media package is locked for.
+   */
+  long MIN_LOCK_DURATION = 1800000;
+
+  /**
+   * Try to acquire a lock for the event's media package, if already held
+   * by the user refresh the lock.
+   *
+   * @param event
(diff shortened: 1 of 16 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/api/MediaPackageLockService.java
```diff
@@ -0,0 +1,85 @@
...
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.adminui.api;
```
karendolan, line 21: Hi @JamesUoM and @ts23 , Did you ever consider having the mediapackge lock as an Opencast service versus an Admin UI service? If so, why did you feel that it would be better for this service to be exclusively for use in the Admin UI?

JamesUoM: I think @ts23 just implemented this as a soln to the problem of combative editing. But you are right, making this a pure "mediapackage" service rather than an "event/mediapackage" would be better. Where should the REST API lie though as at the moment it extends admin-ng/event? I can only think that it should be added to assets?

The real issue is I don't have a lot of time to refactor this, but will try.

JamesUoM: ... for us the location may be due to us calling the lock from our "edit" node ie not admin and we are forwarding all event API requests to admin, as that is all the "edit" nodes needs.

karendolan: @JamesUoM Thank you for the responses. I would only suggest refactoring as much as it makes sense to identify a path for iteration 2 of the mediapackage lock service. Since you are most knowledgeable about this service, you are best positioned to drop some vision how it could be extended to be used in more use case. For example, concurrent metadata editing or workflow start race conditions on same mp.

Event if mediapackage lock service was its own module, the admin UI can offer a REST api for it, as an adapter, like it does for other services. The mp lock service could also offer a REST api, or not, just its OSGI wiring.

staubesv: @karendolan and @JamesUoM 

Please be careful to not introduce the notion of locking in an ad-hoc way as locking can be quite problematic. Also, a generic media package lock service would likely have other requirements than the locking for the edit node does. Think of things as recursive locks, read/write locks, starvation avoidance, deadlock detection and avoidance, etc. which all are quite unproblematic for what James does here but more problematic if used in the backend.

Seen from that point of view, it might be a better idea to not make this simple locking mechanism a module which will then be used carelessly by others for just anything.

staubesv:
Hi @JamesUoM Please note that https://github.com/opencast/opencast/pull/314 has caused conflicts in this PR.

staubesv:
Hi @JamesUoM, are you aiming at 6.0 for this PR?

KatrinIhler:
@JamesUoM What's the plan for this PR?

KatrinIhler:
@JamesUoM I propose to decline this for now since it hasn't gotten any updates in quite a few months. This doesn't mean that this can't get into the codebase at a later time, in that case just refile the PR, but this way we only have open PRs that are actually available for review, keeping down the total number.

PR #304, opened 2018-06-18
NOJIRA: Skip install of Crowdin if it is already installed

gregorydlogan:
This patch checks to see if crowdin is installed prior to installing it.  This preserves existing behaviour for Travis, while also using the pre-installed crowdin binary built into the Buildbot docker images.  This also means that the docker worker process doesn't need to have sudo, which is handy.


lkiesow:
Seams reasonable. Merged.

PR #305, opened 2018-06-18
MH-12956, Incorrect permission check when requesting indexed workflows

lkiesow:
Opencast checks if users have write permissions on the workflow index
when users try to read data from there. It should obviously check for
read permissions.

Note that the security implications on this should be limited since in
practice users with write access should always have read access as well
and users with write access could control the contents of the index.


rute-santos:
Looks good to me!

rute-santos (approved):

PR #307, opened 2018-06-19
MH-12960 Remove unused JavaScript library backbone.js from Admin UI

staubesv:


staubesv:
Note that this PR requires #306 to be merged as visualsearch.js requires backbone.js.

PR #309, opened 2018-06-20
MH-12950 Fix for workflow with no acl in solr index

rute-santos:
This is not ideal, but the best solution we could come up with at this point. When a workflow finishes, the acl is gotten from asset manager because security attachments may have already been deleted from the workspace/wfr.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
@@ -93,10 +93,7 @@ public class WorkflowMessageReceiverImpl extends BaseMessageReceiverImpl<Workflo
           event.setWorkflowState(wf.getState());
           WorkflowInstance.WorkflowState state = wf.getState();
 
-          if (!(WorkflowInstance.WorkflowState.SUCCEEDED.equals(state)
-                  || WorkflowInstance.WorkflowState.FAILED.equals(state)
-                  || WorkflowInstance.WorkflowState.STOPPED.equals(state))) {
-
+          if (!state.isTerminated()) {
             Tuple<AccessControlList, AclScope> activeAcl = authorizationService.getActiveAcl(mp);
             List<ManagedAcl> acls = aclServiceFactory.serviceFor(getSecurityService().getOrganization()).getAcls();
             Option<ManagedAcl> managedAcl = AccessInformationUtil.matchAcls(acls, activeAcl.getA());
@@ -111,7 +108,8 @@ public class WorkflowMessageReceiverImpl extends BaseMessageReceiverImpl<Workflo
               if (loadedDC.isSome())
                 updateEvent(event, loadedDC.get());
             } catch (Throwable t) {
-              logger.warn("Unable to load dublincore catalog for the workflow {}", wf.getId(), t);
+              logger.warn(String.format("Unable to load dublincore catalog for the workflow %d, mp %s", wf.getId(),
+                      mp.getIdentifier().compact()), t);
             }
 
           }
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
@@ -108,8 +108,8 @@ public class WorkflowMessageReceiverImpl extends BaseMessageReceiverImpl<Workflo
               if (loadedDC.isSome())
                 updateEvent(event, loadedDC.get());
             } catch (Throwable t) {
-              logger.warn(String.format("Unable to load dublincore catalog for the workflow %d, mp %s", wf.getId(),
-                      mp.getIdentifier().compact()), t);
+              logger.warn("Unable to load dublincore catalog for the workflow {}, mp {}", wf.getId(),
+                      mp.getIdentifier().compact(), t);
             }
 
           }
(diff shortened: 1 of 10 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
```diff
@@ -111,7 +108,8 @@ protected void execute(WorkflowItem workflowItem) {
               if (loadedDC.isSome())
                 updateEvent(event, loadedDC.get());
             } catch (Throwable t) {
-              logger.warn("Unable to load dublincore catalog for the workflow {}", wf.getId(), t);
+              logger.warn(String.format("Unable to load dublincore catalog for the workflow %d, mp %s", wf.getId(),
+                      mp.getIdentifier().compact()), t);
```
lkiesow: No need for `String.format` when using the logger:

```Java
logger.warn("Unable to load dublincore catalog from media package {} for workflow {}", mp.getIdentifier(), wf.getId(), t);
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -36,7 +36,18 @@
 @XmlJavaTypeAdapter(WorkflowInstanceImpl.Adapter.class)
 public interface WorkflowInstance extends Configurable {
   enum WorkflowState {
-    INSTANTIATED, RUNNING, STOPPED, PAUSED, SUCCEEDED, FAILED, FAILING
+    INSTANTIATED, RUNNING, STOPPED, PAUSED, SUCCEEDED, FAILED, FAILING;
+
+    public boolean isTerminated() {
```
lkiesow, line 41: 👍

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
```diff
@@ -491,19 +496,27 @@ protected SolrInputDocument createDocument(WorkflowInstance instance) throws Exc
...
-      AccessControlList acl;
-      try {
-        acl = authorizationService.getActiveAcl(mp).getA();
-      } catch (Error e) {
-        logger.error("No security xacml found on media package {}", mp);
-        throw new WorkflowException(e);
+    // If workflow has ended, get the latest acl from the asset manager because there may be no security
+    // attachments in the workspace anymore
+    WorkflowInstance.WorkflowState state = instance.getState();
+    if (state.isTerminated()) {
```
lkiesow: No strong feelings but since we only use the state once in the following line, maybe:

```java
if (instance.getState().isTerminated()) {
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
```diff
@@ -491,19 +496,27 @@ protected SolrInputDocument createDocument(WorkflowInstance instance) throws Exc
...
+      AQueryBuilder query = assetManager.createQuery();
+      AResult result = query.select(query.snapshot())
+              .where(query.mediaPackageId(mp.getIdentifier().compact()).and(query.version().isLatest())).run();
+      if (result.getSize() > 0 && result.getRecords().head().isSome()) {
```
lkiesow: If `result.getSize() == 0`, shouldn't `result.getRecords().head()` return `None` anyway which would make `result.getRecords().head().isSome()` return `false`? Or in other words: Do we need the size check?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
```diff
@@ -491,19 +496,27 @@ protected SolrInputDocument createDocument(WorkflowInstance instance) throws Exc
...
+        aclMp = result.getRecords().head().get().getSnapshot().get().getMediaPackage();
       }
+    }
+
+    try {
+      AccessControlList acl = authorizationService.getActiveAcl(aclMp).getA();
       addAuthorization(doc, acl);
+    } catch (Error e) {
```
lkiesow: Do we need to catch a generic error here?

lkiesow (commented):
This looks good. I've noted a few minor things but nothing critical. We also might want to discuss if we want to bring this into `r/5.x`. Not sure how critical the bug is for people… I'll bring it up at Tuesday's technical meeting.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
rute-santos: Will do!

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
rute-santos: Probably not. I will remove.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceSolrIndex.java
rute-santos: I don't think so (I think this was already there...)

lkiesow:
Arrgh, completely forgot about this on Tuesday. Sorry! @rute-santos, what do you think, should we bring this into 5.1?

rute-santos:
Yes, I think so. It's a bug fix and someone else may have the same problem.

lkiesow:
I agree. Can you redirect this pull request to `r/5.x`?

rute-santos:
Done!

lkiesow (approved):

PR #310, opened 2018-06-21
MH-12963 Localize dates/times in add-event summary

KatrinIhler:
Localize the start and end dates and times in the summary of the
add-event dialog when creating a scheduled event. Also don't show the
seconds for the duration since they can't be set, so are always 0.


slampunk:
I will /review/ this

staubesv:
Hi @slampunk 

Thanks for taking the review! I've assigned this to @rrolf as he is your "mentoring committer".

slampunk:
LGTM, can be merged as is.

PR #312, opened 2018-06-22
MH-12966 Do not pre-select-from option in metadata property sheets

staubesv:
Open the Event Details modal causes the creation of a new snapshop under specific circumstances: In case that exactly one option is available, the pre-select-from directive will automatically select this option (to avoid the user to have to select the only available option manually).
In the case of the metadata property sheets, this behaviour seems neither correct nor appropriate: It is incorrect as for non-mandatory fields, there actually are two options: No option and the single available option. Inappropriate because most metadata fields can be configured and you would not configure a select for a single value, would you? I

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleSelect.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleSelect.html
@@ -6,7 +6,7 @@
 
   <div class="editable-select" ng-show="editMode">
 
-    <select chosen pre-select-from="collection"
+    <select chosen
             ng-model="params.value"
             name="{{ params.name }}"
             ng-change="submit()"


karendolan:
Hi @staubesv this is a good find and fix. 

I'm concerned that "pre-select-from" took all of the collection's single entry (key and value in JSON) instead of just the value string of the single key. This pull removes the pre-select-from altogether from the series selection, which removes the concern here. But, I wonder if that issue will spring up in other areas.

staubesv:
@karendolan Actually, it removes it all together from all property sheet elements that use a select (not just series).

You are right about it doesn't address the error that manifests itself in

`"value":{"label":"test+series","value":"08604172-a746-4120-83eb-29f334ed652c"}`

I think that pre-select-from originally did work for quite a while and has been broken in 4.x or 5.x.

-> Let's keep this PR onhold for now as it somewhat workarounds the core of the problem, but I won't have time for further investigations at this moment.

staubesv:
@karendolan See https://github.com/opencast/opencast/pull/334. Pre-select-from does work on all other places as far as I can tell.

KatrinIhler (approved):
This solves my problem that led me to create PR #365 and also works with setting a default workflow! :)

staubesv:
I removed the "onhold" as @KatrinIhler needs this PR merged and #334 has addressed the concerns mentioned by @karendolan

PR #313, opened 2018-06-26
NOJIRA-live-schedule-fix-issue-in-documatation

uzk-rrzk:
The documentation for configuring a live schedule contains a wrong pattern for naming the live stream.
The slash for the flavor (i.e. presenter/delivery) is substituted in the module by  '-' not '_' .


lkiesow (approved):
Checked this in the code: The change makes sense → merging now

PR #315, opened 2018-06-27
MH-12970, Senseless XACML parsing

lkiesow:
The method `getAclFromInputStream(…)` should parse an XACML file from a
given input stream and return the ACL object. The current implementation
internally parses the input stream, catching possible exceptions and
packing the result in an `Option` only to unpack it again right
afterwards which throws an exception in case the exception had been
caught before. Then the result is packed in a `Tuple` with a statically
set `AclScope` which, of course, is immediately unpacked again with the
scope just being discarded.

Finally, the one usage of this would break if there was any problem with
the ACL instead of handling this gracefully, falling back to a strict
non-permissive access rule.

This patch replaces all this mess with a straightforward implementation
of parsing the ACL from the given XACML input stream.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -53,12 +53,18 @@ import org.opencastproject.security.api.User;
 import com.entwinemedia.fn.Fn2;
 import com.entwinemedia.fn.data.Opt;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
 import java.io.InputStream;
 
 /**
  * Security layer.
  */
 public class AssetManagerWithSecurity extends AssetManagerDecorator {
+  private static final Logger logger = LoggerFactory.getLogger(AssetManagerWithSecurity.class);
+
   public static final String WRITE_ACTION = "write";
   public static final String READ_ACTION = "read";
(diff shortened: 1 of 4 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -90,7 +91,12 @@ private boolean isAuthorizedByAcl(Version version, String mpId, String action) {
     Opt<Asset> secAsset = super.getAsset(version, mpId, "security-policy-episode");
     if (secAsset.isSome()) {
       InputStream in = secAsset.get().getInputStream();
-      final AccessControlList acl = authSvc.getAclFromInputStream(in).getA();
+      final AccessControlList acl;
+      try {
+        acl = authSvc.getAclFromInputStream(in);
+      } catch (IOException e) {
+        return false;
```
staubesv: Hmm... there seem to be several problems that could cause this exception:

- Access to "in" works but data is corrupt ("cannot be parsed")
- Access to "in" does not work, e.g. unauthorised access

Maybe handle those cases separately and log an error in case the data is corrupt?

EDIT: Oh, the IOException likely means that data is corrupt. But still: Shouldn't this be logged?

lkiesow: Logging makes sense. Done

lkiesow:
Fixed typo in log statement. Should now be ready to go.

PR #314, opened 2018-06-26
MH-12969, Ensure formatting of OSGI configuration

lkiesow:
This ensures a homogeneous style for all OSGI configuration files based
on the Intellij IDEA code formatter.

PR #317, opened 2018-06-28
MH-12969 Eclipse IDE import Opencast XML style preferences

karendolan:
This addresses some of the XML formatting that @lkiesow proposes to standardize in pull #314 
These settings are imported into Eclipse differently than the other eclipse_settings files, using Eclipse File->Import->General->Preferences
This pull is in reponse to @lkiesow comments in https://github.com/opencast/opencast/pull/314


lkiesow (commented):
I am not using Eclipse but the settings look reasonable. The only thing I am wondering is if people will find how to import the settings. Would it make sense to add a very short readme [like this](https://github.com/harvard-dce/opencast-contrib/blob/64d99d2a7caf31155e44a6c1040d6f20f150030c/docs/intellij_settings/readme) for all of the Eclipse settings?

lkiesow:
Merging it like this. We can always improve things…

PR #318, opened 2018-06-28
MH-12974, Access denial to event for unprivileged user

lkiesow:
Due to the incorrect usage of ``getAcl()``, unprivileged users may not
have access to events despite the ACL explicitly allowing them access.
Instead, `getActiveAcl()` should be used to evaluate the correct access
permissions for a given event.

Note that there are further, less critical issues with `getAcl()` as
mentioned on list [1] which will be dealt with separately with a patch
against `r/6.x`.

[1] https://groups.google.com/a/opencast.org/forum/#!topic/dev/CqGR1K3LNXE

*Work sponsored by SWITCH*

### DIFF
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
@@ -106,7 +106,6 @@ import org.opencastproject.scheduler.api.TechnicalMetadataImpl;
 import org.opencastproject.scheduler.api.Util;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AccessControlUtil;
-import org.opencastproject.security.api.AclScope;
 import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.DefaultOrganization;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
How does one test this?  Process a workflow with the Authenticated ACL template, confirm that the anonymous user can't access the video, then add the anonymous user with read-only to the ACL?

lkiesow:
@JulianKniephoff, you mainly worked on this and tested this. Maybe, you can better describe this. If I remember correctly, you would create a series ACL and allow a user access through that acl and then not upload an episode acl. The series acl would not be effective since `authorizationService.getAcl` in contrast to `authorizationService.getActiveAcl` would not use the usual chain of fallbacks from an episode ACL to a series ACL to a global ACL. That's why this patch switches to `getActiveAcl`

JulianKniephoff:
@lkiesow no I think you got it. ;) Fortunately the SWITCH code base just does this when you create an event, i.e. they don't add an episode ACL by default. To test this with the community version, I did some manual `curl`-ing IIRC, to exclude the episode ACL, because the admin UI always sends (a potentially empty) one along.

This was before my vacation, though, so take it with the appropriately sized pinch of salt. :wink: But of course I can try to answer any further questions, @gregorydlogan. :smile:

PR #319, opened 2018-06-28
MH-12738: Tiered Storage for the Asset Manager

gregorydlogan:
This pull request implements the concept of tiered storage within the Asset Manager.  The codebase already had the concept in a few places, but it was only roughed in.  This pull request implements the functionality using modular stores, and the existing Opencast job architecture.

Major changes:
 * Assets and snapshots now have the concept of their storage location.  This requires additional DB fields.
   * Assets in non-hot storage are still present within the system, and are treated as such.
 * The Asset Manager itself now has another (argh) layer, which keeps changes to the existing layers to a minimum.
   * The REST endpoints live within the /assets namespace, so use those REST docs for testing
 * Triggering a storage migration creates a job, which then does the work.  There is no synchronous transfer mode accessible by design, since large files may take a long time to move.  The design goal is that a store backed by manual tape backup could be implemented, so s

Known, or suspected issues:
 * There are no UIs built for this yet.  This is strictly back-end functionality
 * The asset manager currently expects all files to be accessible at all times.  While (in theory) non-live files should work, this has not been extensively tested.  This will be discussed further in the AWS S3 PR.
 * The class names...  `TieredStorageAssetManagerJobProducer` anyone?  Ideas on renaming are welcome.

Review suggestion 1: Look at each of the commits separately.  This is a large changeset, but I think I've broken the major groups of changes out into their own commits.
Review suggestion 2: For testing whether this actually works as expected, use [this](https://github.com/gregorydlogan/opencast/tree/f/MH-12738-test) branch, which is this PR plus a fake test store.

*Work sponsored by the University of Stuttgart*


staubesv:
Wow, quite a PR... could you please add some basic documentation of this to https://docs.opencast.org/develop/developer/asset-manager/?

gregorydlogan:
Not today, but I'll add that in the first round of review comments!

docs/upgrade/4_to_5/mysql5.sql
```diff
@@ -231,6 +231,7 @@ ALTER TABLE oc_assets_snapshot ADD CONSTRAINT UNQ_oc_assets_snapshot  UNIQUE (me
 ALTER TABLE oc_assets_snapshot DROP INDEX IF EXISTS UNQ_mh_assets_snapshot;
 ALTER TABLE oc_assets_snapshot ADD CONSTRAINT FK_oc_assets_snapshot_organization FOREIGN KEY (organization_id) REFERENCES oc_organization (id);
 ALTER TABLE oc_assets_snapshot DROP FOREIGN KEY FK_mh_assets_snapshot_organization;
+ALTER TABLE oc_assets_snapshot ADD COLUMN storage_id VARCHAR(256) NOT NULL;
```
rute-santos: ADD COLUMN for the  oc_assets_asset table is missing from upgrade script.
I believe you also need to populate the values with the 'local storage' because it doesn't accept null?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+            chuck(e);
+          }
+          getDb().setStorageLocation(s, targetStoreId);
+          deleteAssetsFromStore(s, currentStore);
+        } else {
+          //Else, the content is *not* local and is going to a *different* remote
+          String intermediateStore = getLocalAssetStore().getStoreType();
+          logger.debug(format("Moving {} from {} to {}, then to {}",
+                  s.toString(), currentStoreId, intermediateStore, targetStoreId));
+          Snapshot snap = record.getSnapshot().get();
```
rute-santos: The variable 's' already has the snapshot.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+  private boolean isLocalAssetStoreId(String storeId) {
+    return getLocalAssetStore().getStoreType().equals(storeId);
+  }
+
+  /** Returns true if the store id is not equal to the local asset store's id */
+  private boolean isRemoteAssetStoreId(String storeId) {
+    return !isLocalAssetStoreId(storeId);
+  }
+
+  @Override public Opt<Asset> getAsset(Version version, final String mpId, final String mpeId) {
```
rute-santos: Formatting looks off?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
```diff
@@ -197,6 +238,22 @@ public void setAssetStore(AssetStore assetStore) {
     this.assetStore = assetStore;
   }
 
+  public void addRemoteAssetStore(RemoteAssetStore assetStore) {
```
rute-santos: Since the remote asset store dependency has to be dynamic, I believe you need to synchronize this method, the removeRemoteAssetStore(), and the part of activate() where it is creating the delegate and emptying 'remotes'. Checking a null delegate and adding the remote store should be atomic. For instance, if the code is executed in this order: line 242, then lines 155-159, then line 243, the remote asset store is not added to the delegate.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TieredStorageAssetManager.java
```diff
@@ -0,0 +1,160 @@
+/**
```
rute-santos, line 1: Formatting looks off here too.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TieredStorageAssetManagerJobProducer.java
```diff
@@ -0,0 +1,381 @@
...
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class TieredStorageAssetManagerJobProducer extends AbstractJobProducer {
+
+  /** The logging facility */
+  private static final Logger logger = LoggerFactory.getLogger(TieredStorageAssetManagerJobProducer.class);
+
+  public static final String JOB_TYPE = "org.opencastproject.assetmanager";
+  public static final Float JOB_LOAD = 1.0f;
```
rute-santos: Per previous conversations, please change job load to < 1 value.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TieredStorageAssetManagerJobProducer.java
```diff
@@ -0,0 +1,381 @@
...
+    Operation op = null;
+    String operation = job.getOperation();
+    List<String> arguments = job.getArguments();
+    String id;
+    String targetStore = arguments.get(0);
+    VersionImpl version;
+    Date start;
+    Date end;
+    boolean bypass;
+    Set<Job> subjobs;
```
rute-santos: Unused variables: bypass, subjobs.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TimedMediaArchiver.java
```diff
@@ -0,0 +1,190 @@
...
+    /*if (!archive.isStoreAvailable(storeType)) {
+      throw new RuntimeException("Store type " + storeType + " is not available to the archive");
+    }
+
+    try {
+      archive.archiveByInterval(null, maxAge, true, storeType);
+    } catch (ArchiveException e) {
+      throw new RuntimeException("Unable to offload archive data", e);
+    }*/
+  }
```
rute-santos: Was this commented out by accident?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
+/**
```
rute-santos, line 1: Formatting looks off here too.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+                          description = "The job created to move the snapshots.",
+                          responseCode = SC_OK),
+                  @RestResponse(
+                          description = "Invalid parameters, and the job was not created",
+                          responseCode = SC_BAD_REQUEST),
+                  @RestResponse(
+                          description = "There has been an internal error, and the job was not created",
+                          responseCode = SC_INTERNAL_SERVER_ERROR)},
+          returnDescription = "The Job created")
+  public Response moveByIdAndVersion(@FormParam("start") final Integer startInt, @FormParam("end") final Integer endInt, @FormParam("target") final String target) {
```
rute-santos: This should be moveByDate, right?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+                          description = "The job created to move the snapshots.",
+                          responseCode = SC_OK),
+                  @RestResponse(
+                          description = "Invalid parameters, and the job was not created",
+                          responseCode = SC_BAD_REQUEST),
+                  @RestResponse(
+                          description = "There has been an internal error, and the job was not created",
+                          responseCode = SC_INTERNAL_SERVER_ERROR)},
+          returnDescription = "The Job created")
+  public Response moveByIdAndVersion(@FormParam("id") final String id, @FormParam("start") final Integer startInt, @FormParam("end") final Integer endInt, @FormParam("target") final String target) {
```
rute-santos: moveByIdAndDate?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+  @Produces(MediaType.TEXT_XML)
+  @RestQuery(name = "moveByDate", description = "Move all snapshots between two dates.",
+          restParameters = {
+                  @RestParameter(
+                          name = "start",
+                          isRequired = true,
+                          type = RestParameter.Type.INTEGER,
+                          defaultValue = "",
+                          description = "The mediapackage ID to move."),
+                  @RestParameter(
```
rute-santos: Description incorrect. Please specify the expected date format in the description (unix format?). I would actually prefer a human-readable format.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+                          name = "end",
+                          isRequired = true,
+                          type = RestParameter.Type.INTEGER,
+                          defaultValue = "",
+                          description = "The version to move."),
+                  @RestParameter(
```
rute-santos: Same here.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+                          type = RestParameter.Type.TEXT,
+                          defaultValue = "",
+                          description = "The mediapackage ID to move."),
+                  @RestParameter(
+                          name = "start",
+                          isRequired = true,
+                          type = RestParameter.Type.INTEGER,
+                          defaultValue = "",
+                          description = "The mediapackage ID to move."),
+                  @RestParameter(
```
rute-santos: Description incorrect. Please specify the expected date format in the description.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,316 @@
...
+                          name = "end",
+                          isRequired = true,
+                          type = RestParameter.Type.INTEGER,
+                          defaultValue = "",
+                          description = "The version to move."),
```
rute-santos: Description incorrect. Please specify the expected date format in the description.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
```diff
@@ -159,6 +160,41 @@ public boolean saveProperty(final Property property) {
     });
   }
 
+  /**
+   * Save a property to the database. This is either an insert or an update operation.
+   */
+  public boolean saveAssetProperty(final Property property) {
```
rute-santos: Necessary? I didn't find where this method was called.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/query/AQueryBuilderImpl.java
```diff
@@ -190,6 +198,14 @@ public AQueryBuilderImpl(AbstractAssetManager am) {
     };
   }
 
+  @Override public Field<String> storage() {
+    return new AbstractSnapshotField<String, String>(Q_SNAPSHOT.availability) {
```
rute-santos: Q_SNAPSHOT.storageId?

modules/asset-manager-impl/src/main/resources/OSGI-INF/tiered-storage-timed-media-archiver.xml
```diff
@@ -0,0 +1,18 @@
...
+  <implementation class="org.opencastproject.assetmanager.impl.TimedMediaArchiver" />
+  <property name="service.description" value="Timed media archiver Service" />
+
+  <service>
+    <provide interface="org.osgi.service.cm.ManagedService" />
+  </service>
+
+  <reference cardinality="1..1" interface="org.opencastproject.assetmanager.impl.TieredStorageAssetManager" name="AssetManager" policy="static" bind="setAssetManager"/>
+  <reference cardinality="1..1" interface="org.opencastproject.security.api.OrganizationDirectoryService" name="OrganizationDirectoryService" policy="static" bind="bindOrganizationDirectoryService"/>
+  <reference cardinality="1..1" interface="org.opencastproject.security.api.SecurityService" name="SecurityService" policy="static" bind="bindSecurityService"/>
```
rute-santos: AbstractScanner also needs to bind the ServiceRegistry.

rute-santos (commented):
Hi Greg,
 
I have added my initial comments inline. I haven't tested yet. My plan is to take a look at the S3 code first and then test both together.
 
I like the way you implemented the transfer between stores as jobs. Much cleaner than our dce 1.x code :) I am just worried that executing transfers in parallel will possibly cause the same object to be copied more than once to the remote storage instead of 'linked' - think different versions of the same object being transferred at the same time. I haven't looked at the S3 code yet so I can't tell if that is true or not,
 but it would happen with the dce 1.x code if executed in parallel.
 
The code formatting overall seems off. Did you by any chance forget to add the ide settings? It would be good to re-format all the new files at least.

And yes, I forgot to ask you to add documentation too :)

docs/upgrade/4_to_5/mysql5.sql
gregorydlogan: This also needs to be in 5_to_6 now that I look at it...

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
gregorydlogan: This format is used in a few places in the AM already.  Personally I don't like it but was going for code that fits the style of the relevant codebase.  I'm going to fix it here regardless, since it helps with readability imo.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
gregorydlogan: Since we can (in theory) count on OSGI enforcing the dependency graphs between the various services, we know that the delegate already has to exist for this method to even be called, correct?  Karaf won't be attempting to bind new stores before the activate() method has returned.  Thus, we don't need to synchronize this, especially if we rip out the delegate protection that's present in this method.

At least, that's my read of the spec...

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TimedMediaArchiver.java
gregorydlogan: Hah, yes.  Will address.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
gregorydlogan: Yep!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
gregorydlogan: Good catch, this was the old way and I missed removing this method.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/TieredStorageAssetManager.java
gregorydlogan: Formatting rejiggered, hopefully better?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
gregorydlogan: Formatting rejiggered, hopefully better?

modules/asset-manager-impl/src/main/resources/OSGI-INF/tiered-storage-timed-media-archiver.xml
gregorydlogan: Done.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
rute-santos: I would agree if the remote store dependency were 'static'. But it is 'dynamic' (as it should be) and so the 'bind' method can be called after the class has been instantiated, but before activate(). With this change, it's possible to get a null pointer.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -51,6 +54,7 @@
 
 public abstract class AbstractTieredStorageAssetManagerRestEndpoint extends AbstractAssetManagerRestEndpoint {
 
+  private DateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
```
rute-santos: DateFormat is not thread safe so it should be a local variable.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
gregorydlogan: Huh, that's a dumb bug in the spec IMO.  Regardless, how would synchronizing the method help?  It would prevent multiple threads from adding assets stores at the same time, but it wouldn't stop the null pointer.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
gregorydlogan: Ah, right.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
rute-santos: The previous version of the code had an intermediate set (remotes) where the remote stores were added if the delegate was not ready yet.

gregorydlogan: Ah, I get it now.  Ok, added synchronization in relevant places!

staubesv:
Hi @gregorydlogan Please note that https://github.com/opencast/opencast/pull/314 has caused conflicts in this PR.

gregorydlogan:
Fixed, and formatting rules applied.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,338 @@
...
+
+  @POST
+  @Path("moveById")
+  @Produces(MediaType.TEXT_XML)
+  @RestQuery(name = "moveById", description = "Move a mediapackage based on its ID.",
+          restParameters = {
+                  @RestParameter(
+                          name = "id",
+                          isRequired = true,
+                          type = RestParameter.Type.TEXT,
```
rute-santos: RestParameter.Type.STRING instead of TEXT?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,338 @@
...
+                          defaultValue = "",
+                          description = "The mediapackage ID to move."),
+                  @RestParameter(
+                          name = "target",
+                          isRequired = true,
+                          type = RestParameter.Type.TEXT,
```
rute-santos: RestParameter.Type.STRING instead of TEXT?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractTieredStorageAssetManagerRestEndpoint.java
```diff
@@ -0,0 +1,338 @@
...
+
+  @POST
+  @Path("moveByIdAndVersion")
+  @Produces(MediaType.TEXT_XML)
+  @RestQuery(name = "moveByIdAndVersion", description = "Move a mediapackage based on its ID and version.",
+          restParameters = {
+                  @RestParameter(
+                          name = "id",
+                          isRequired = true,
+                          type = RestParameter.Type.TEXT,
```
rute-santos: Same: RestParameter.Type.STRING instead of TEXT? (and in all places that have id, version, target?)

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+    final String prettyMpId = mpId + "@v" + version;
+    logger.debug(format("Moving assets for snapshot %s to store %s", prettyMpId, store.getStoreType()));
+    for (final MediaPackageElement e : snap.getMediaPackage().getElements()) {
+      if (!MOVABLE_TYPES.contains(e.getElementType())) {
+        logger.debug(format("Skipping %s because type is %s", e.getIdentifier(), e.getElementType()));
+        continue;
+      }
+      logger.debug(format("Moving %s to store %s", e.getIdentifier(), store.getStoreType()));
+      final StoragePath storagePath = StoragePath.mk(orgId, mpId, version, e.getIdentifier());
+      final Opt<StoragePath> existingAssetOpt = findAssetInVersionsAndStores(e.getChecksum().toString(), store.getStoreType());
```
rute-santos: In my tests, findAssetInVersionsAndStores does not find the asset previously archived to the aws-s3 store when there's another version of the asset that is "local" because the "local" asset is always returned. I tracked it to the findAssetByChecksum that uses the "singleResult" clause so the first one found is returned.
This causes several copies of the asset in aws-s3 instead of only one.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
```diff
@@ -197,6 +238,22 @@ public void setAssetStore(AssetStore assetStore) {
...
+    } else {
+      delegate.addRemoteAssetStore(assetStore);
+    }
+  }
+
+  public synchronized void removeRemoteAssetStore(RemoteAssetStore assetStore) {
+    if (null != delegate) {
+      delegate.removeRemoteAssetStore(assetStore);
+    } else {
+      logger.warn("Unable to remove remote store of type {} because delegate is null!", assetStore.getStoreType());
```
rute-santos, line 253: You could have the asset store removed before activate() was called. Very unlikely, but possible :)

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerStorageMoveOperationHandler.java
```diff
@@ -0,0 +1,106 @@
...
+  @Override
+  public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobContext context)
+          throws WorkflowOperationException {
+    MediaPackage mp = workflowInstance.getMediaPackage();
+    WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
+
+    logger.debug("Working on mediapackage {}", mp.getIdentifier().toString());
+
+    String targetStorage = StringUtils.trimToNull(operation.getConfiguration("target-storage"));
+    if (tsamjp.datastoreExists(targetStorage)) {
```
rute-santos: Reverse test: if (!tsamjp.datastoreExists(targetStorage)).

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerStorageMoveOperationHandler.java
```diff
@@ -0,0 +1,106 @@
...
+      } catch (NumberFormatException e) {
+        throw new WorkflowOperationException("Invalid version number", e);
+      }
+    }
+
+    logger.debug("Target version set to {}", version);
+
+    logger.debug("Beginning moving process");
+    //Note that a null version implies *all* versions
+    Job fakejob = new JobImpl(workflowInstance.getId());
```
rute-santos: fakeJob not used?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+    final String mpId = snap.getMediaPackage().getIdentifier().toString();
+    final String orgId = snap.getOrganizationId();
+    final Version version = snap.getVersion();
+    final String prettyMpId = mpId + "@v" + version;
+    logger.debug(format("Moving assets for snapshot %s to store %s", prettyMpId, store.getStoreType()));
+    for (final MediaPackageElement e : snap.getMediaPackage().getElements()) {
+      if (!MOVABLE_TYPES.contains(e.getElementType())) {
+        logger.debug(format("Skipping %s because type is %s", e.getIdentifier(), e.getElementType()));
+        continue;
+      }
```
rute-santos, line 346: We need to check if the asset has already been moved. For instance, if there was an exception in a previous run and some mp assets were moved, some were not. Otherwise, an asset may not exist in the origin anymore.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+                existingAsset.getMediaPackageElementId(), e.getChecksum(), store.getStoreType());
+        if (!store.copy(existingAsset, storagePath)) {
+          throw new AssetManagerException(
+                  format("An asset with checksum %s has already been archived but trying to copy or link asset %s to it failed",
+                          e.getChecksum(), existingAsset));
+        }
+      } else {
+        final Opt<Long> size = e.getSize() > 0 ? Opt.some(e.getSize()) : Opt.<Long>none();
+        store.put(storagePath, Source.mk(e.getURI(), size, Opt.nul(e.getMimeType())));
+      }
```
rute-santos, line 368: The asset storeId in the database is only updated after all assets in the media package are moved. This will also cause assets to take more space in the remote store than they should. For instance, if a media package has 2 references to the same file, such as presenter/source and presenter/work, 2 files will be put in the remote store instead of having a logic hard-link between them.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -0,0 +1,412 @@
...
+    }
```
rute-santos: Why are you not moving the manifest.xml too? Is the manifest.xml an exact copy of what is in the database?

KatrinIhler:
@gregorydlogan Has conflicts (+ failed checks)!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
gregorydlogan: I feel like that's so unlikely that we should leave this here.  If it becomes an issue then we need to explore how to properly log this :)

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerStorageMoveOperationHandler.java
gregorydlogan: Typo repaired!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
gregorydlogan: This was addressed in your PR I believe?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
gregorydlogan: Line 369 (getDb().setAssetStorageLocation) is at the bottom of the for loop, which is iterating over the list of media package elements.  Barring two versions of a given snapshot being moved at the same time, the asset's storage ID should already be set correctly if it was just moved, wouldn't it?

Regarding hard-linking, that's the store's job IMO.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
gregorydlogan: The manifest isn't included in the MP afaik.  It also does not get updated beyond the initial ingest, so it would be out of date regardless.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
rute-santos: Yes!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
rute-santos: Well, it's this piece of code that decides between a 'copy' or a 'move' so it is this code's responsibility to tell the store if that element is a new one or not. The store's particular implementation may do linking or not.
Anyway, I think my PR addressed that too?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
rute-santos: The manifest.xml is a copy of the media package for that version. When a snapshot is taken, it's created in the /archive file system. It seems to be an exact copy of the mediapackage_xml column in the oc_assets_snapshot table. So it doesn't need to be copied? Can you double-check if this is true? I am just afraid of deleting something that may be needed later...

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
gregorydlogan: I think so.  We'll see in further testing :D

gregorydlogan:
@KatrinIhler ty!

@rute-santos fixes applied, ready for further review.

rute-santos (approved):

PR #320, opened 2018-06-28
MH-12975, Inconsistent access control handling

lkiesow:
Opencast's authorization service has two different methods of evaluating
XACML (access control) attachments with should work similarly but do
work differently resulting in an unwanted denial of access in some cases.

In the end, both paths should work the same way and could even share most
of the code making the whole process less confusing and easier to
understand.

*This contains and extends pull request #318 and addresses [this post on list](https://groups.google.com/a/opencast.org/forum/#!topic/dev/CqGR1K3LNXE)*

*Work sponsored by SWITCH*


gregorydlogan:
Merge conflict here @lkiesow

PR #322, opened 2018-07-03
MH-12977 Fix data placeholders in edit scheduled events

staubesv:
Before:

bug

After:

bug_fixed

Note that fixing this for the series select would require to adjust the generic metadata edit fields which is by intention not done in this PR as that field is used in many places.


staubesv:
@CGreweling As you did the review for #231, would you mind take this related review, too?

PR #325, opened 2018-07-04
MH-12979, Automatically test ddl scripts

lkiesow:
Make Travis CI automatically test the DDL scripts to ensure they work
with both MariaDB and MySQL and continue to do so.

Note that for now, this does not test the upgrade scripts.


gregorydlogan:
Oh, now that's a good idea!  Do you want to add the upgrade scripts to this PR, or file a second?

lkiesow:
@gregorydlogan I'll file a second one. This already works. Otherwise, what would have been the point in filing this pull request ;-)

PR #327, opened 2018-07-04
[MH-12930] Fill creator metadata field with actual user when new event

jchssystems:
When I create a new event, I would like to save the user, who has created it.
- This user must be the one, who is logged in, in the new event wizard.
- And the user who has created the event in the details window.
- The publisher dublincore metadata is used.
- The field is persistent.

### DIFF
--- a/etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
+++ b/etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
@@ -31,7 +31,7 @@ property.uid.label=EVENTS.EVENTS.DETAILS.METADATA.ID
 property.uid.type=text
 property.uid.readOnly=true
 property.uid.required=false
-property.uid.order=16
+property.uid.order=17
 
 # Presenters
 property.presenters.inputID=creator
@@ -144,3 +144,22 @@ property.license.readOnly=false
 property.license.required=false
 property.license.listprovider=LICENSES
 property.license.order=5
+
+# Publisher
+# By default, the publisher is the logged in user in the creation event
+# and it is read only.
+# You can set up a list provider, changing the properties:
+#   property.publisher.readOnly=false
+#   property.publisher.listprovider=YOUR_LIST_PROVIDER
+# If you want to use the events publishers as list provider, you must
+# set up the provider in this way:
+#   property.publisher.listprovider=EVENTS.PUBLISHER
+# If you use a list provider, the property will have the users of the list plus
+# the logged in user in the event creation wizard and the users of the list
+# in the edit properties.
+property.publisher.inputID=publisher
+property.publisher.label=EVENTS.EVENTS.DETAILS.METADATA.PUBLISHER
+property.publisher.type=text
+property.publisher.readOnly=true
+property.publisher.required=true
+property.publisher.order=16
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -105,6 +105,7 @@ import org.opencastproject.mediapackage.track.VideoStreamImpl;
 import org.opencastproject.metadata.dublincore.DublinCore;
 import org.opencastproject.metadata.dublincore.EventCatalogUIAdapter;
 import org.opencastproject.metadata.dublincore.MetadataCollection;
+import org.opencastproject.metadata.dublincore.MetadataField;
 import org.opencastproject.rest.BulkOperationResult;
 import org.opencastproject.rest.RestConstants;
 import org.opencastproject.scheduler.api.Recording;
@@ -2018,6 +2019,18 @@ public abstract class AbstractEventEndpoint {
         collection.removeField(collection.getOutputFields().get("startTime"));
       if (collection.getOutputFields().containsKey("location"))
         collection.removeField(collection.getOutputFields().get("location"));
+      if (collection.getOutputFields().containsKey(DublinCore.PROPERTY_PUBLISHER.getLocalName())) {
+        MetadataField<String> publisher = (MetadataField<String>) collection.getOutputFields().get(DublinCore.PROPERTY_PUBLISHER.getLocalName());
+        Map<String, String> users = new HashMap<String, String>();
+        if (!publisher.getCollection().isNone()) {
+          users = publisher.getCollection().get();
+        }
+        String loggedInUser = getSecurityService().getUser().getName();
+        if (!users.containsKey(loggedInUser)) {
+          users.put(loggedInUser, loggedInUser);
+        }
+        publisher.setValue(loggedInUser);
+      }
       metadataList.add(getIndexService().getCommonEventCatalogUIAdapter(), collection);
     }
     return okJson(metadataList.toJSON());
@@ -2252,6 +2265,8 @@ public abstract class AbstractEventEndpoint {
         query.withOptedOut(Boolean.parseBoolean(filters.get(name)));
       if (EventListQuery.FILTER_REVIEW_STATUS_NAME.equals(name))
         query.withReviewStatus(filters.get(name));
+      if (EventListQuery.FILTER_PUBLISHER_NAME.equals(name))
+        query.withPublisher(filters.get(name));
       if (EventListQuery.FILTER_COMMENTS_NAME.equals(name)) {
         switch (Comments.valueOf(filters.get(name))) {
           case NONE:
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventUtils.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventUtils.java
@@ -202,6 +202,14 @@ public final class EventUtils {
       metadata.addField(newUID);
     }
 
+    if (metadata.getOutputFields().containsKey(DublinCore.PROPERTY_PUBLISHER.getLocalName())) {
+      MetadataField<?> publisher = metadata.getOutputFields().get(DublinCore.PROPERTY_PUBLISHER.getLocalName());
+      metadata.removeField(publisher);
+      MetadataField<String> newPublisher = MetadataUtils.copyMetadataField(publisher);
+      newPublisher.setValue(event.getPublisher());
+      metadata.addField(newPublisher);
+    }
+
     return metadata;
   }
 
--- a/etc/index/adminui/event-mapping.json
+++ b/etc/index/adminui/event-mapping.json
@@ -51,6 +51,8 @@
 
             "creator": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
 
+            "publisher": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
+
             "license": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
 
             "rights": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
--- a/etc/index/externalapi/event-mapping.json
+++ b/etc/index/externalapi/event-mapping.json
@@ -51,6 +51,8 @@
 
             "creator": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
 
+            "publisher": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
+
             "license": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
 
             "rights": { "type" : "string", "index" : "not_analyzed", "store" : "no" },
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/impl/index/AdminUISearchIndex.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/impl/index/AdminUISearchIndex.java
@@ -154,4 +154,14 @@ public class AdminUISearchIndex extends AbstractSearchIndex implements EventInde
     return getTermsForField(ThemeIndexSchema.NAME, Option.some(new String[] { Theme.DOCUMENT_TYPE }));
   }
 
+  /**
+   * Returns all the known events' publishers
+   *
+   * @return a list of events' publishers
+   */
+  @Override
+  public List<String> getEventPublishers() {
+    return getTermsForField(EventIndexSchema.PUBLISHER, Option.some(new String[] { Event.DOCUMENT_TYPE }));
+  }
+
 }
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -833,6 +833,7 @@
            "SOURCE":         "Source",
            "CREATED":        "Created",
            "CREATED_BY":     "Created by",
+           "PUBLISHER":      "Publisher",
            "LICENSE":        "License",
            "CONTRIBUTORS":   "Contributor(s)",
            "RIGHTS":         "Rights",
@@ -1862,6 +1863,9 @@
        },
        "CREATOR": {
          "LABEL": "Creator"
+       },
+       "PUBLISHER": {
+         "LABEL": "Publisher"
        },
        "HOSTNAME": {
          "LABEL": "Host name"
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/api/EventIndex.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/api/EventIndex.java
@@ -67,4 +67,11 @@ public interface EventIndex {
    */
   List<String> getThemeNames();
 
+  /**
+   * Returns all the known events publishers' usernames
+   *
+   * @return a list of events publishers' usernames
+   */
+  List<String> getEventPublishers();
+
 }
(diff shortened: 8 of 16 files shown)


jchssystems:
I get the same errors that in the pull request of the branch _f/MH-12782-improve-adminui-usability-show-only-events-and-series-with-write-access_

But these goals run well:
mvn site runs well, errors in these tests but the build is succesfull at the end.
mvn javadoc:javadoc javadoc:aggregate runs well too, with no errors.

lkiesow:
If you look at [the build logs](https://api.travis-ci.org/v3/job/400134363/log.txt), you will see that the complaint is actually quite valid since you have used [nonexistent parameters](https://github.com/opencast/opencast/pull/327/files#diff-85187d2a10b915f3d4cc2cb4663f0519R482) in your JavaDocs

jchssystems:
Hi @lkiesow,

yes, I have _publishers_, instead _publisher_, but the rest of the errors are in the package
org/opencastproject/mediapackage
and I have done nothing there. It is the same problem that I have with the branch
f/MH-12782-improve-adminui-usability-show-only-events-and-series-with-write-access

If you want, I can change my error in the @param, but site will fail again.

karendolan:
Hi @jchssystems, This looks like a great feature that will be useful for our site, too. I see that the code checks if the publisher field is already filled before making the query to getPublisher. My question is if that scenario has been tested or if you are willing to create a test case for that scenario. Our site pre-populates new episode metadata with defaults from the series. The publisher field is prepopulated with the point person (Opencast user) responsible for that series. For example, when some one is uploading an event, he or she does not have to be the point person if there is already a default for that series. Occasionally, an event is created that does not have a default point person. It would be great to auto-assign the current user as producer for the event, then, allow the user to re-assign the publisher if some other point person is more appropriate.

lkiesow:
> yes, I have publishers, instead publisher, but the rest of the errors are in the package
org/opencastproject/mediapackage

@jchssystems, I cannot sport any more errors in the build log. Do you get more locally? That would be strange. Can you provide a log?

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
```diff
@@ -144,3 +144,11 @@ property.license.readOnly=false
 property.license.required=false
 property.license.listprovider=LICENSES
 property.license.order=5
+
+# Created By
+property.createdBy.inputID=publisher
+property.createdBy.label=EVENTS.EVENTS.DETAILS.METADATA.CREATED_BY
+property.createdBy.type=text
```
karendolan: Since this pull focuses on "publisher" as the person initiating the publication of the event, it would avoid confusion with "creator" if the METADATA label and text was PUBLISHER (vs CREATED_BY). i.e avoid duplicating the craziness of the "#Presenters" section in this file.

jchssystems:
Hi Lars, Karen,

sorry for the long absence. I came back from a long vacation yesterday. I had an incorrect version of the ffmpeg library. That was the greatest problem.

Karen, I changed the name, as you suggested. I called it _createdBy_ as in series. Both do the same but maybe you're right. It's very confusing.

Lars, I don't understand. I had problems in Javadoc in /mediapackage/ package, but I fixed my error in the javadoc parameter @publishers and all run well.

KatrinIhler:
@jchssystems This branch now has conflicts in the EventsListProvider.java!

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
```diff
@@ -144,3 +144,11 @@ property.license.readOnly=false
 property.license.required=false
 property.license.listprovider=LICENSES
 property.license.order=5
+
+# Publisher
+property.publisher.inputID=publisher
+property.publisher.label=EVENTS.EVENTS.DETAILS.METADATA.PUBLISHER
+property.publisher.type=text
+property.publisher.readOnly=true
```
KatrinIhler: Why read only?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -2018,6 +2019,10 @@ public Response getNewMetadata() {
         collection.removeField(collection.getOutputFields().get("startTime"));
       if (collection.getOutputFields().containsKey("location"))
         collection.removeField(collection.getOutputFields().get("location"));
+      if (collection.getOutputFields().containsKey(DublinCore.PROPERTY_PUBLISHER.getLocalName())) {
+        MetadataField<String> publisher = (MetadataField<String>) collection.getOutputFields().get(DublinCore.PROPERTY_PUBLISHER.getLocalName());
+        publisher.setValue(getSecurityService().getUser().getName());
```
KatrinIhler: This would overwrite an existing value, right? Maybe the field should be checked if it has a (default) value first?

KatrinIhler (commented):
Commented on a couple minor things, but when I tested it everything looked good.

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
jchssystems: The publisher is the one who creates the new event, the one that is logged in. Once created, it would not make sense to change it.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
jchssystems: The event is new, so the field is still empty. The logged user is assigned directly.

jchssystems: When the event is already created and you see its properties, then you can read the publisher because it already has a value. That is in the EventUtils.java class. Thanks for the review.

KatrinIhler: Theoretically there could be a default value configured, though. This currently only works for metadata fields that are fed by a listprovider (e.g. _license_) but I haven't given up hope that one day a default can be configured for all metadata fields...then it would be overwritten here.

jchssystems: Publisher is also a list provider, but not static as licenses. You need a list provider because you can search events by publisher (there is a publisher filter). Only users who have created events are in the list. But as event property of the wizard, a list has no sense in this case.

It would be possible to take a default value from a properties configuration file, for example, org.opencastproject.adminui.endpoint.OsgiEventEndpoint.cfg, and programatically fill the property first time with this value.

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
karendolan: @jchssystems Our site's use case is that the publisher is the person responsible for the publishing of the event, not necessarily the person performing a manual schedule or a manual upload. Our site requires the person performing the upload, or scheduling the event, to re-assign the publisher to be the correct point person, if the current user is not that person. I have tested your pull and realize that the strategy for auto-filling the publisher to the current user negates being able to use a listprovider for that field. This means the current user does not get the drop down list of other publishers to enable easily changing the assignment. I'm trying to find a work around that can auto-select or insert the current user with a listprovided set of publishers. Any help you can provide now to make it easier is grateful appreciated!

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
karendolan: +1, Our site's use case is that the person responsible for publishing all events in a series is assigned as the publisher value in the series. The value gets first priority when assigned to an event. If there is no publisher in the series, the current user is assigned as the responsible publisher. The current user can re-assign the publisher value if another person should responsible for publishing as specific event. 

Our use case is because the uploading/scheduling user is not necessarily the one responsible for triming or performing the final publish task.

The user case may be different for sites that use only one workflow or do not have a coorperative team approach to lecture processing and publishing in Opencast.

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
karendolan: @jchssystems @KatrinIhler, I like this pull,  but I am -1 for making the publisher field unediable by default prior to event publication. I can see that it might be useful to freeze the value of this field after publication.

jchssystems: @karendolan Maybe something could be done.

With:
_property.license.readOnly=true_
could work as it does now.

With:
_property.license.readOnly=false_
_property.license.listprovider=PUBLISHER_
could work as you need if I change the code of "new event". You can have a list of all the editors plus the user who is logged in for example.

It would be also possible to have any list at first as provider.
_property.license.listprovider=CONTRIBUTORS_

I have not tested anything yet. Let me take a look and I will tell you something tomorrow.

jchssystems: Hi @karendolan, @KatrinIhler We are lucky. I have tested with no change in this branch:

property.publisher.readOnly=false
property.publisher.listprovider=CONTRIBUTORS

Yo can change only the properties, with no compilation and does work. You can see the list of contributors to choose one. The selected user of the list is the one is logged in, but you can change it. Once the event is created, you could also change it in the properties of the event. The filter shows, of course, only the users who have created events.

But:
property.publisher.listprovider=PUBLISHER
does not work yet. For now, it provides the filter, but not the property. I will try to change this.

Test that if it is fit for you.

jchssystems: Hi again @karendolan, @KatrinIhler, the code works also with the publisher list provider, but the name is EVENTS.PUBLISHER:

property.publisher.readOnly=false
property.publisher.listprovider=EVENTS.PUBLISHER

With this provider:
1. If there are no publishers yet, the list is empty (?).
2. If there are publishers and the logged in user is already a publisher, the user is selected.
3. If there are publishers and the logged in is not a publisher yet, no-option is selected.

With this provider, the question is, how can an user be a publisher first time? I suggest that, if publisher property is a list, this list be the users of the list provider plus the logged in user. So first time, if an user wants to be a publisher, he must log in and creates an event. Next times, another user can create events and assign him as publisher. What do you think about this solution? Have you another better solution?

jchssystems: Hi @karendolan, I have upload a new version of the branch. I think that it is what you need.

By default, the publisher is the logged in user and you cannot change it.

But if you change the file configuration
_etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg_,
for example with the lines:

property.publisher.readOnly=false
property.publisher.listprovider=PROVIDER

you can choose the publisher between the users of the list. Once the event has been created, this field is read only.

The new event wizard shows all the users of the provider list, plus the logged in user. The logged in user is selected by default.

If you want the list of publishers as provider, you must set:
property.publisher.listprovider=EVENTS.PUBLISHER

The filter of publishers in the events page always shows the users who have published at least one event.

If the list has only one item, no one is selected. There is a bug that I have already reported:
MH-12971-The list of series in the Event Details window does not work properly
I think that it occurs in all lists, when there is only one item.

Please, check the branch if you can and tell me if that is fit to you.

karendolan: Hi @jchssystems That sounds perfect,  thank you for all the work! I'm trying it all out in the next couple hours.

Q: "how can an user be a publisher first time?". 
A: "this list be the users of the list provider plus the logged in user" +++1!

A slight twist for our site is that we are not expecting to use the EVENTS.PUBLISHER listprovider, but a SERIES.PUBLISHER listprovider that I locally created (coded) to prepopulate the event publisher option list with default publishers for all the usual events. We require all "real" events to be associated to a series. All our series are assigned default publishers responsible for publishing the events of that series. Since our publishers know how to upload and troubleshoot in Opencast, they frequently cover for each other to do manual uploads when a CA fails the upload.

With your and our scenario combined, it is a good for our needs that user who is not already on the SERIES.PUBLISHER list will only show up there when they are actively creating a new event. 

This is what I am testing for in the next couple hours, using this pull as the solid core, and a few unintrusive local additions:

1. If there are no publishers yet, the list contains the currently logged in user and auto-selected to that user.
2. If there are publishers and the logged in user is or is not already a publisher, the user is selected.
3. Upon assigning a series to the new-event, the publisher associated to that series is auto-selected from the SERIES.PUBLISHER provided list (I'm also writing a auto-populate-from-series directive for our site today)
4. The user can override the auto-assigned publisher, at any time, to select themself or another publisher from the list.
5. If the series selection is change again, the field will also be changed again with that series publisher. The user can then override the new change by choosing themself or another from the list.

jchssystems: @karendolan,

1. _"If there are no publishers yet, the list contains the currently logged in user and auto-selected to that user."_
Be careful with the bug. If the list has only one item, publisher.setValue(...) does not work. I think It occurs in all lists.

2. _"If the series selection is change again, the field will also be changed again with that series publisher. The user can then override the new change by choosing themself or another from the list."_
But if this occurs before the event is created. After that, the field is read only. If you want to make it editable, you must delete:
newPublisher.setReadOnly(true);
in EventUtils.java
The list in events properties is really there. Only making newPublisher true, the field is read only and you do not see the list.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventUtils.java
```diff
@@ -202,6 +202,15 @@ public static MetadataCollection getEventMetadata(Event event, EventCatalogUIAda
       metadata.addField(newUID);
     }
 
+    if (metadata.getOutputFields().containsKey(DublinCore.PROPERTY_PUBLISHER.getLocalName())) {
+      MetadataField<?> publisher = metadata.getOutputFields().get(DublinCore.PROPERTY_PUBLISHER.getLocalName());
+      metadata.removeField(publisher);
+      MetadataField<String> newPublisher = MetadataUtils.copyMetadataField(publisher);
+      newPublisher.setValue(event.getPublisher());
+      newPublisher.setReadOnly(true);
```
KatrinIhler: Why are you setting read-only to true here instead of using the read-only value configured? Is that because you want it to be editable on event-creation, but read-only later? Then I feel like it would be better to introduce a second configuration value for this since it's rather unintuitive that the read-only value can be configured, but is then not obeyed as expected. The metada fields are messy enough as it is...

jchssystems: @KatrinIhler,

1. _"Why are you setting read-only to true here instead of using the read-only value configured? Is that because you want it to be editable on event-creation, but read-only later?"_
Yes.

2. _"Then I feel like it would be better to introduce a second configuration value for this since it's rather unintuitive that the read-only value can be configured, but is then not obeyed as expected"_
Much better, but in my opinion this is an important change that I should not decide it. All fields use the same configuration properties: _inputID_, _outputID_, _label_, _type_, _readOnly_, _required_ and _order_.
I have not seen the impact of a new property, but I am sure, it is not trivial, and only for the publisher...
Besides, if it were editable, the list would not show the logged in user. It is not easy. The code is in an utility class and we have no access to the service _SecurityService_. In _Series_ is diferent. _Events_ and _Series_ do not resolve the same problems in the same way.

I see two solutions:
1. The easiest solution would be to remove that line:
newPublisher.setReadOnly(true);
and the property would be also editable (without the logged in user).
2. The other, also very easy, to add the line:
publisher.setReadOnly(false);
when we create a new event, and delete the property in the file configuration. But I do not like this.

KatrinIhler: @jchssystems Yeah, there's probably no perfect solution for this right now. :/ I would prefer solution 1 because while it's not exactly what we want, it doesn't make the metadata code more complicated and obscure than it already is.

jchssystems:
Hi @KatrinIhler,
the change was made. Check it and I hope, that it may be useful.

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
karendolan: @jchssystems & @KatrinIhler I finally finished tested everything locally and it works great for our use case! Thank you so much for offering this pull.

gregorydlogan:
If @karendolan and @KatrinIhler are happy with this, I'm ok with it.  Code looks good to me.  My sole complaint is that there is a distinct lack of documentation on how to switch the read-onlyness of the publisher.  @jchssystems, can you address that in time for feature freeze?  If not we can probably merge this in just under the wire, and you can file a documentation PR afterwards.

jchssystems:
Thanks @gregorydlogan,

I have written a few documentation lines for the configuration of the event's properties in the same configuration file. See _org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg_.

I have not seen a good place in the (admin) documentation to write it. The  _Metadata_ page is very general. Tell me if this is enough. Otherwise, merge the pull request and give me a hint where is the best place to document this in a documentation pr.

1. As comment in configuration.
2. Anywhere in the (admin) docs.
3. Elsewhere.

And I do it later.

gregorydlogan:
Hi @jchssystems , I would put it in `docs/guides/admin/docs/configuration/ui.md`.  I've just fixed an issue where that page was not being linked into the documentation correctly, so when you file the new PR please make sure to base it off of the latest 6.x :)

PR #329, opened 2018-07-06
MH-12803 Fix for mp 'start' when event is created (affects live scheduler service)

rute-santos:
Problem:
The media package was created with 'start' set to date/time the event was created.
According to MH-12250, media package 'created' and 'start' should match and come from the temporal dc property.
The 'createEvent' method was checking for a PROPERTY_TEMPORAL in the metadata, but the 'temporal' at that point was stored with output id 'startDate' and had no value so it was never found.
After the capture agent ingested, 'start' was correctly set to the same value as temporal so everything else worked fine.
Solution:
The change was to get the 'temporal' from the 'source' metadata if it (as 'startDate') doesn't come in the 'event' metadata and set the 'created' and 'start' to the same value, complying to MH-12250.

### DIFF
--- a/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
+++ b/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
@@ -454,11 +454,12 @@ public class LiveScheduleServiceImpl implements LiveScheduleService {
       }
 
       // Capture agent did not pass any CA_PROPERTY_RESOLUTION_URL_PREFIX property when registering
-      // so use the service configuration, including stream overrides (temporary)
-      if (liveStreamingUrl == null)
-        throw new LiveScheduleException(
-                "Cannot build live tracks because '" + LIVE_STREAMING_URL + "' configuration was not set.");
+      // so use the service configuration
       if (mp.getTracks().length == 0) {
+        if (liveStreamingUrl == null)
+          throw new LiveScheduleException(
+                  "Cannot build live tracks because '" + LIVE_STREAMING_URL + "' configuration was not set.");
+
         for (MediaPackageElementFlavor flavor : liveFlavors) {
           for (int i = 0; i < streamResolution.length; i++) {
             String uri = replaceVariables(mpId, caName, UrlSupport.concat(liveStreamingUrl.toString(), streamName),
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/events/CommonEventCatalogUIAdapter.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/events/CommonEventCatalogUIAdapter.java
@@ -38,8 +38,9 @@ import org.osgi.service.cm.ManagedService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.text.ParseException;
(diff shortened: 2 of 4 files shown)


lkiesow:
@JulianKniephoff, this seems to be related to your recent changes. Can you take the review?

JulianKniephoff:
I'll take a look. :slightly_smiling_face:

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -454,11 +454,12 @@ void addLiveTracks(MediaPackage mp, String caName) throws LiveScheduleException
...
-      // so use the service configuration, including stream overrides (temporary)
-      if (liveStreamingUrl == null)
-        throw new LiveScheduleException(
-                "Cannot build live tracks because '" + LIVE_STREAMING_URL + "' configuration was not set.");
+      // so use the service configuration
       if (mp.getTracks().length == 0) {
+        if (liveStreamingUrl == null)
+          throw new LiveScheduleException(
+                  "Cannot build live tracks because '" + LIVE_STREAMING_URL + "' configuration was not set.");
+
```
JulianKniephoff, line 462: This change seems to be unrelated to the PR, but I might of course be wrong. @rute-santos can you comment on this?

I'm not against unrelated changes to style or other code quality related issues (I'm actually strongly **for** it), but this seems to change behavior as well, so I'm a bit more reserved.

Oh and if we touch this code here, anyway, we might as well add the `{}`, like we discussed on the mailing list some time ago, right? :sweat_smile:

JulianKniephoff (approved):
Despite my one nitpicky comment I approve this. The code looks fine. After a few concentrated readings I even understand what it does (but the fact that the code surrounding metadata fields is a mess is not this PR's fault). :sweat_smile: It works and does what it advertises as well, so I guess everything is fine.

However, I have to give the following caveat: Despite @lkiesow's trust in my ability to judge this code because I worked on a similar issue before, I have to admit that I'm still pretty shaky when it comes to this whole topic of metadata handling in general and the start date discussion in particular. To the point where I don't even know where I should look to see whether these changes don't break any assumptions elsewhere. I thought about it a bit with the background I **do** have now and it seems fine to me.

But yeah, another pair of eyes with more Opencast experience would probably not be a bad thing. :wink:

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
rute-santos: The original ticket was created because of a problem in the live scheduler service. Later, I realized that the problem was caused by the metadata. That's why I thought it was OK to fix the bug in the same PR.

gregorydlogan:
Is this at all relevant to #352?  If I merge that one, then it gets brought forward into develop, are we going to step on eachother's toes?

JulianKniephoff:
I don't think so. What #352 does makes certain decisions in the code **independent** of any start date, which is the only thing that this PR affects AFAICT.

~~This is just a hunch of course; I can try to test it later if you want.~~ I tested this now and it seems fine. I merged `r/5.x` (with #352 now merged) with this branch, scheduled a live event and it showed up correctly in the media module. I changed the start date into the past and back into the future and metadata updates still worked fine, as did the display in the media module. I think there is no conflict here.

PR #330, opened 2018-07-06
MH-12982, 3.0 database upgrade error

lkiesow:
The database upgrade script from 2.3 to 3 tries to create a database
table called `mh_oaipmh_harvesting` which already exists in previous
versions. Hence, the upgrade will fail if users do not simply ignore the
error.


lkiesow:
I can change this to `r74.x` if you want me to but it might make sense to have a working upgrade script in the release branch and merging this back to `r/4.x` should be no problem at all.

lkiesow:
As discussed at today's technical meeting I've changed the documentation to link the online version (which should soon include this fix) in the upgrade documentation.

PR #332, opened 2018-07-09
MH-12985 Fix incorrect warnings in event modals

krinnewitz:
Using Actions->Start Task and Actions->Edit scheduled events you get a
table displaying, among other information, the event status. Under
certain circumstances, the user should be kept from continuing within
the wizard. 

However, to check if the user can proceed, the event source
is used in contrary to the event_status being displayed in the table.
This leads to weird situations where the user doesn't understand why he
can't continue. Fixed by checking for event_status instead of source
when generating the messages.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
@@ -338,7 +338,7 @@ function ($scope, Table, Notifications, EventBulkEditResource, SeriesResource, C
     };
 
     $scope.nonScheduleSelected = function() {
-        return JsHelper.filter($scope.getSelected(),function (r) { return r.event_status_raw !== 'EVENTS.EVENTS.STATUS.SCHEDULED'; }).length > 0;
+        return JsHelper.filter($scope.getSelected(), $scope.nonSchedule).length > 0;
     };
 
     $scope.rowsValid = function() {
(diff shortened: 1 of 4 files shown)


lkiesow:
Forcing this review on @wsmirnow since he touched this before

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
```diff
@@ -334,11 +334,11 @@ function ($scope, Table, Notifications, EventBulkEditResource, SeriesResource, C
     };
 
     $scope.nonSchedule = function(row) {
-        return row.source !== 'SCHEDULE';
+        return row.event_status_raw !== 'EVENTS.EVENTS.STATUS.SCHEDULED';
     };
 
     $scope.nonScheduleSelected = function() {
-        return JsHelper.filter($scope.getSelected(),function (r) { return r.source !== 'SCHEDULE'; }).length > 0;
+        return JsHelper.filter($scope.getSelected(),function (r) { return r.event_status_raw !== 'EVENTS.EVENTS.STATUS.SCHEDULED'; }).length > 0;
```
KatrinIhler: Wouldn't it make sense to call the nonSchedule function here instead, so the logic isn't duplicated?

KatrinIhler (approved):
I finally managed to reproduce the bug and can confirm that this PR fixes it. Didn't find much to complain about either... ;)

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
krinnewitz: Yes, makes sense. I changed it. Thank you for your review :+1:

KatrinIhler:
Looks good, lets merge this!

PR #334, opened 2018-07-09
MH-12966: Fix pre-select-from for metadata fields

pmiddend:
This adds a new attribute called `pre-select-from-value` to select inputs which use `pre-select-from`. With this enabled, the single value isn’t taken directly from the list given, but from the `.value` attribute of the selected list element.

We need this for the metadata fields, for example in the “Event Details” modal. This is currently broken, as it tries to set the value of the series to an object containing a label and a value, instead of just the value.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/preSelectFromDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/preSelectFromDirective.js
@@ -44,7 +44,14 @@ function (_) {
                     }
 
                     if (_.size(options) === 1 && _.size(ngModel) > 0) { // supports objects and arrays
-                        ngModel[0].$setViewValue(options[_.keys(options)[0]], 'myevent');
+                        var valueToSet = options[_.keys(options)[0]];
+                        // There's an additional attribute you can set: "pre-select-from-value", which
+                        // sets the value not to the "v" in the array itself, but to "v.value". This
+                        // is currently used in the metadata fields for certain dialogs.
+                        if (!angular.isUndefined($attr.preSelectFromValue)) {
+                            valueToSet = valueToSet.value;
+                        }
+                        ngModel[0].$setViewValue(valueToSet, 'myevent');
                         unregister();
                     }
                 }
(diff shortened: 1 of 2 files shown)


pmiddend:
See also:

https://github.com/opencast/opencast/pull/312

KatrinIhler:
I guess this might no longer be necessary if PR #312 gets merged, or is it?

lkiesow:
@KatrinIhler, do you want to take the review for this? You seem to be involved already.

KatrinIhler (changes requested):
So this fixes the preselectfrom on metadata fields without breaking it in other places, which is good. I don't agree with the way it works on the metadata fields though (and which was one reason the preselection was removed completely from metadata fields in PR #312), and even though this PR only recreates the old behaviour, I think it could be fixed here:

In my opinion, the preselection of a value for a metadata field should only happen if the field is required, because if it is not, there are really two valid options: The value, or no value at all. Otherwise the user might have to set the fields back to blank that should not be filled, so the preselection doesn't save any work in that case...

staubesv:
Hi @KatrinIhler 

This PR was basically created because of concerns mentioned in #312.

It is true that the pre-selection would need to consider whether a field is mandatory or not if used in the metadata property sheets. But as #312 has removed this directly from the metadata property sheet, the problem does not exist anymore.

I see two options here:
a) we just decline the PR as it is not needed anymore
b) we merge the PR in case it would be needed in the future

No strong opinion from my side - what do you prefer?

KatrinIhler:
@staubesv Well there is a third option: Add to this PR so that preselection will only happen on metadata fields that are required. Not sure how long that would take though. If that option is not acceptable, I guess I would prefer declining this PR and maybe holding on to that code somewhere else in case someone wants to go for option 3, because there's already a lot of code in opencast that is there "just in case" and as a developer I find that more confusing than helpful.

staubesv:
@KatrinIhler We don't need the pre-select-from in the metadata property sheets for us and therefore won't work on this anymore.

-> I'm fine with this PR being declined

KatrinIhler:
@staubesv I don't have the rights to decline a PR (not a commiter yet), so can you do it?

staubesv:
I decline this PR as it has become obsolete.

PR #335, opened 2018-07-09
MH-12987 Prohibit changing a scheduled event to be in the past

krinnewitz:
While it is not possible to create a scheduled event which ends in the
past, changing an existing one to end in the past is still possible
using Event Details -> Scheduling.

Fixed by adding an error message and not saving the data.

Please note: The problem also exists for Actions -> Edit scheduled events. Also, the backend should check accordingly and refuse to save scheduling information when the event ends in the past. We did not fix this, yet, since we didn't want to put too much effort into this right now.

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -152,6 +152,7 @@
      "CONFLICT_BULK_DETECTED":   "Scheduling conflict: changing the events results in conflicts. Please change the location, dates or times.",
      "CONFLICT_ALREADY_ENDED":   "Scheduling error: The event has already ended.",
      "CONFLICT_END_BEFORE_START":   "Scheduling error: Schedule end has to be later than the start.",
+     "CONFLICT_IN_THE_PAST": "The schedule could not be updated: You cannot schedule an event to be in the past.",
      "INVALID_ACL_RULES":   "Rules have to contain a valid role and read or/and write right(s).",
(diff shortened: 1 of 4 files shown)


karendolan:
Hi @krinnewitz, "As an administrator... I would like to stop a recording for a class that has ended early withouth having to figure out the current time and then add a few minutes to the current time to put in the schedule end time in order to stop the event prior to it's scheduled end time". In other words, how hard would it be to add an "end now" button that adds the magic minimum now+buffer time to the schedule end time and submits the change? The button prevents the user from *wanting* to add a time in the past to the schedule end.

krinnewitz:
Hi @karendolan, I think adding such a button would take about one or two hours. Buffer time should be configurable, I guess.

However, SWITCH currently has other priorities and doesn't want to put additional effort into this issue.

lkiesow:
@karendolan, note that CAs would need to support updating the schedule while they are already recording.

Btb, do you want to review and merge this?

gregorydlogan:
@krinnewitz conflicts here

krinnewitz:
Thanks @gregorydlogan. I have fixed the conflict.

lkiesow (changes requested):
@krinnewitz, looks good and is ready to be merged except for the notification which reads `Scheduling error: The event has already ended.` But since it's scheduled and the date could not be changed to something in the past, it has not yet ended :-)

My suggestion would be to use a notification like this:

>  The schedule could not be updated: You cannot schedule an event to be in the past.

krinnewitz:
@lkiesow good point. I updated the PR accordingly, based on your proposal.

lkiesow:
Thanks for the change → merged

PR #336, opened 2018-07-09
Update version info

pascalseeland:


staubesv (changes requested):
That is not correct as 6.x has not yet been released.

Probably, it would be better to have an example that does not refer to the current supported versions at all as this requires us to update this page again and again.

-> But at least, the docs need to be correct: We currently support Opencast 5.x and Opencast 4.x.

pascalseeland:
I did put 6.x, since I thought these docs would be released together with 6.x so it would fit. A more generic text would probably would be helpful to not update the page on every release. But I can also create more pull requests for the other branches.

lkiesow:
Reading this, it seems to be OK since it is stated to be an example from the time of writing. Hence we can leave it as it is or apply this patch and it should be fine.

staubesv:
Ok, makes sense.

@pascalseeland Thanks for the patch!

staubesv (approved):

PR #337, opened 2018-07-09
MH-12988-delete-scheduled-live Fix for scheduled live event not deleted

rute-santos:
Problem: the Live Scheduler service listens to messages from services (Scheduler, Asset Manager), including delete messages, to update/retract the live media package from the search index, but the admin UI blocks the deletion if the event has a publication channel.

This was fixed by changing the admin UI to allow the scheduled event be deleted if it has only a live publication channel.

To test:
* Configure the live scheduler service
* Register a capture agent 
* Schedule a capture in the future
* Before the capture starts, delete the event
* The event should be deleted without having to run a workflow to retract it before
* The live event should be retracted from the search index (use rest endpoint to verify)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
@@ -33,7 +33,9 @@ angular.module('adminNg.controllers')
             publicationCount++;
         });
 
-        if (publicationCount > 0) {
+        if (publicationCount == 1 && currentEvent.publications[0].id == "engage-live") {
+        	return false;
+        } else if (publicationCount > 0) {
             return true;
         } else {
             return false;
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
@@ -33,13 +33,7 @@ angular.module('adminNg.controllers')
             publicationCount++;
         });
 
-        if (publicationCount == 1 && currentEvent.publications[0].id == "engage-live") {
-        	return false;
-        } else if (publicationCount > 0) {
-            return true;
-        } else {
-            return false;
-        }
+        return publicationCount > 0 && currentEvent.publications[0].id != "engage-live";
     },
     tableByPublicationStatus = function(isPublished) {
         var result = {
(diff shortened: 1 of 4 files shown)


modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/message/SchedulerUpdateHandler.java
```diff
@@ -79,8 +79,7 @@ else if (BooleanUtils.toBoolean(publishLive)) {
           break;
         case Delete:
         case DeleteRecordingStatus:
-          if (isLive(mpId))
-            liveScheduleService.deleteLiveEvent(mpId);
+          liveScheduleService.deleteLiveEvent(mpId);
```
staubesv: It seems that omitting the isLive check here will cause the live schedule service to issue a search query to find the media package mpId and then does another test (based on tracks) to find out whether it is live.

Can you please provide some background information about this change?

Maybe a short comment that explains why at this particular code location, isLive is not used?

staubesv (changes requested):
@rute-santos Generally looks good to me. Just have a question about one specific change.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/message/SchedulerUpdateHandler.java
rute-santos: When a delete notification comes in, the workflow configuration, which is checked by "isLive" to determine if this is a live event, does not exist anymore. This is actually a bug fix :)

staubesv: Thanks for clarifying this.

staubesv:
@rute-santos No strong opinion, but it would be an option to redirect this against 5.x...

What do you think?

rute-santos:
Hi @staubesv, yes, I will redirect to 5.x.

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
```diff
@@ -33,7 +33,9 @@ angular.module('adminNg.controllers')
             publicationCount++;
         });
 
-        if (publicationCount > 0) {
+        if (publicationCount == 1 && currentEvent.publications[0].id == "engage-live") {
+        	return false;
```
lkiesow: indentation :)

But this seems rather complex and I'm wondering if this couldn't as well be solved by:

```js
return publicationCount > 0 && currentEvent.publications[0].id != "engage-live"
```

It's not exactly the same but it feels like that's what you want.

lkiesow (commented):
I'll take over since @staubesv is still on vacation. I'm basically ok with merging it like this though I have one sugestion for a change. Let me know if you want to modify the code or if you want me to merge it like this.

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/bulkDeleteController.js
rute-santos: Yes, it will do the trick. I will change.

PR #338, opened 2018-07-10
MH-12989 Add missing roles for actions->edit scheduled

krinnewitz:
For Actions -> Edit Scheduled Events, a POST request to
event/scheduling.json is sent. Even though a POST request is used, this
is a read-only action. So an _VIEW role is used.
    
The same applies for event/bulk/conflicts and event/bulk/update
(needs write-access).

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -111,6 +111,7 @@
     <sec:intercept-url pattern="/admin-ng/event/*/comment/*/*" method="PUT" access="ROLE_ADMIN, ROLE_UI_EVENTS_DETAILS_COMMENTS_EDIT" />
     <sec:intercept-url pattern="/admin-ng/event/*/comment/*" method="PUT" access="ROLE_ADMIN, ROLE_UI_EVENTS_DETAILS_COMMENTS_EDIT"/>
     <sec:intercept-url pattern="/admin-ng/event/*/optout/*" method="PUT" access="ROLE_ADMIN, ROLE_UI_EVENTS_STATUS_EDIT" />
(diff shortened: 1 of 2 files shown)


etc/security/mh_default_org.xml
```diff
@@ -125,10 +126,12 @@
     <sec:intercept-url pattern="/services/sanitize" method="POST" access="ROLE_ADMIN, ROLE_UI_SERVICES_STATUS_EDIT" />
     <sec:intercept-url pattern="/staticfiles" method="POST" access="ROLE_ADMIN, ROLE_UI_THEMES_CREATE, ROLE_UI_THEMES_EDIT" />
     <sec:intercept-url pattern="/admin-ng/acl" method="POST" access="ROLE_ADMIN, ROLE_UI_ACLS_CREATE" />
+    <sec:intercept-url pattern="/admin-ng/event/bulk/conflicts" method="POST" access="ROLE_ADMIN, ROLE_UI_EVENTS_DETAILS_SCHEDULING_VIEW" />
```
lkiesow: Where would you need a conflict check when you only have read access to that dialog. Should this maybe be `ROLE_UI_EVENTS_DETAILS_SCHEDULING_EDIT`?

staubesv: It is a read request that uses HTTP POST because a parameter is a potentially large list of event IDs that could cause problems then using a GET request because of length limitations of query parameters.

I agree that this requires somebody looking at this line to know about the implementation which should not be necessary...

Should I add a comment? I could also just change it to a *_EDIT role.

What do you think?

staubesv: @krinnewitz  Could you please change to ROLE_UI_EVENTS_DETAILS_SCHEDULING_EDIT here? @lkiesow seems right as conflict check will only be needed if you can edit scheduled events.

krinnewitz:
@staubesv @lkiesow done

lkiesow:
Thanks for fixing this → merged

PR #340, opened 2018-07-10
MH-12992: Trigger conflict check in “Edit scheduled” on “Next”

pmiddend:
Conflict checking can be fairly slow on large systems with lots of
scheduled events. Currently, when you select a few of those events and
open the "Edit scheduled events" tab, changing a single item, such as
the start time, will already trigger a conflict check. If this takes
very long then, it's bad for usability.

This commit changes the behavior, so conflict checking is done when
you press the “Next” button in the wizard.

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -243,6 +243,7 @@
       },
       "EDIT_EVENTS": {
           "CAPTION": "Edit scheduled events",
+          "CONFLICT_CHECK_RUNNING": "Conflict check is running…",
           "GENERAL": {
               "CAPTION": "General",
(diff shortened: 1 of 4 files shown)


modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -243,6 +243,7 @@
       },
       "EDIT_EVENTS": {
           "CAPTION": "Edit scheduled events",
+          "CONFLICT_CHECK_RUNNING": "Conflict check running…",
```
lkiesow: “Conflict check *is* running…” or “Checking for conflicts…”?

pmiddend: Yep, sounds better, I'll change it,

lkiesow:
Seems to work → merged.
Thanks for the patch.

PR #342, opened 2018-07-10
MH-12993 REST docs for Admin UI Event endpoint broken

staubesv:
See https://opencast.jira.com/browse/MH-12993


rute-santos (approved):
Looks good and works.

PR #344, opened 2018-07-10
MH-12995 Fix conflict check not detecting some conflicts

krinnewitz:
Using Actions -> Edit scheduled events, you can re-schedule multiple
events at once. While conflict checks were performed against the
existing events, a check for conflicts between the events to edit was
missing.

Now, the conflict check also checks for conflicts between the selected
events.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1198,7 +1198,7 @@ public abstract class AbstractEventEndpoint {
               // different agent -> no conflict
               return;
             }
-            if (!start.after(otherEnd) && !end.before(otherStart)) {
+            if (start.before(otherEnd) && end.after(otherStart)) {
               // conflict
               currentConflicts.add(convertEventToConflictingObject(DateTimeSupport.toUTC(otherStart.getTime()),
                 DateTimeSupport.toUTC(otherEnd.getTime()), otherEvent.getTitle()));
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1170,7 +1170,7 @@ public abstract class AbstractEventEndpoint {
       return notFoundJson(JSONUtils.setToJSON(notFoundIds));
     }
 
-    final Map<String, List<MediaPackage>> conflicts = new HashMap<>();
+    final Map<String, List<JValue>> conflicts = new HashMap<>();
     events.values().forEach(e -> e.ifPresent(event -> {
       try {
         if (instructions.getScheduling() != null) {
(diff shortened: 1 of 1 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1180,27 +1180,51 @@ public Response getBulkConflicts(@FormParam("update") final String updateJson) t
...
+            final JSONObject otherScheduling = BulkUpdateUtil.addSchedulingDates(otherEvent, instructions.getScheduling());
+            final Date otherStart = Date.from(Instant.parse((String) otherScheduling.get(SCHEDULING_START_KEY)));
+            final Date otherEnd = Date.from(Instant.parse((String) otherScheduling.get(SCHEDULING_END_KEY)));
+            final String otherAgentId = Optional.ofNullable((String) otherScheduling.get(SCHEDULING_AGENT_ID_KEY))
+              .orElse(otherEvent.getAgentId());
+            if (!otherAgentId.equals(agentId)) {
+              // different agent -> no conflict
+              return;
+            }
+            if (!start.after(otherEnd) && !end.before(otherStart)) {
```
lkiesow: Why not

```
if (start.before(otherEnd) && end.after(otherStart)) {
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1180,27 +1180,51 @@ public Response getBulkConflicts(@FormParam("update") final String updateJson) t
...
+              // conflict
+              currentConflicts.add(convertEventToConflictingObject(DateTimeSupport.toUTC(otherStart.getTime()),
+                DateTimeSupport.toUTC(otherEnd.getTime()), otherEvent.getTitle()));
+            }
+          }));
+
+          // Check for conflicts with other events from the database
+          final List<MediaPackage> conflicting = getSchedulerService().findConflictingEvents(agentId, start, end)
```
lkiesow, line 1209: Do we still need to look for conflicts if we have already detected an event has conflicts in the previous step?

lkiesow: and maybe `.forEach` → `if (!events.keySet().contains(mp.getIdentifier().toString()))` → `currentConflicts.add` is easier to read￼

lkiesow (commented):
Just a few comments. No strong feelings about the suggested changes.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
krinnewitz: That's not the same. start.before(otherEnd) is false for start == otherEnd, while !start.after(otherEnd) is true for start == otherEnd. Same for end.after(otherStart) and !end.before(otherStart)

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
krinnewitz: We could skip checking for conflicts with other events from the Database in case we already found conflicts between the events to edit. That would mean, that the response won't necessarily include _all_ conflicts. It would be faster, but usability would suffer a bit. I think it's a trade-off. @staubesv any preference?

staubesv: @krinnewitz & @lkiesow No strong preference here. I would tend to always do the full conflict check as I would consider it unlikely that the scheduled events being edited conflict themselves (of course they can, but that is likely if the user accidentally enters some incorrect data and therefore not the main use case here).

Seen from that point, it is mainly the schedule conflict detection that counts and the usability would be better.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
lkiesow: Hm… then, this code would mark two events as conflicting if I schedule one from 1 pm to 2 pm and the other one from 2 pm to 3 pm, wouldn't it? That's no conflict. I'm pretty sure any Opencast user would agree.

lkiesow (changes requested):

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
lkiesow: fine with me.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
krinnewitz: You are right. I fixed the check accordingly.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
krinnewitz: @lkiesow I tried to refactor the code to use `.forEach → if (!events.keySet().contains(mp.getIdentifier().toString())) → currentConflicts.add`

However, things became ugly due to checked exceptions. So I suggest to keep the code as currently is.

krinnewitz:
@lkiesow Thanks for your input. I have addressed the issues you came up with. Feel free to continue the review :-)

PR #189, opened 2018-04-12
MH-12177 TimeZone non-threadsafe issue in scheduler-api, scheduler-impl

karendolan:
NOTE: this fix is to resolve the recurrence ical4j issues from the 4x bug-fixes merge back to develop/5x. This is *NOT* a fix for the 4x branch were, presumably, everything is working OK. (i.e. the bug fixes from 4x conflicted with new features added in develop)

This fixes the scheduler-impl unit test TimeZone issues with the Util's create periods method.
This includes a  first-Sunday-after-DST-time-change-within-a-period fix that is partially described in the link in the code comment. 
@gregorydlogan feel free to take and improve!

PR #346, opened 2018-07-11
MH-12998: Clear conflicts when closing “Edit Scheduled Events” modal

pmiddend:
When you have conflicts displayed in the “Edit scheduled events”
modal, then close the modal an reopen it again (possibly with
different events and without changing anything), the previous error
message resides.

This commit clears existing conflicts before the modal is closed.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/editEventsController.js
@@ -22,8 +22,8 @@
 
 // Controller for the "edit scheduled events" wizard
 angular.module('adminNg.controllers')
-    .controller('EditEventsCtrl', ['$scope', 'Table', 'Notifications', 'EventBulkEditResource', 'SeriesResource', 'CaptureAgentsResource', 'EventsSchedulingResource', 'JsHelper', 'SchedulingHelperService', 'WizardHandler', 'Language', '$translate', 'decorateWithTableRowSelection','$timeout',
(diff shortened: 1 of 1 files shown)


pmiddend:
Travis reports failing tests, but I didn't even change any .java files.

doofy:
I think greg was still working on the travis build at the moment. lets wait a moment and see what happens

gregorydlogan:
@pmiddend @doofy just ignore Buildbot for now.  If Travis is happy then we're happy.  Buildbot is experiencing #189, which has not been solved yet.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/edit-events-modal.html
```diff
@@ -1,6 +1,6 @@
 <section ng-form="editEventsForm" ng-show="open" ng-keyup="keyUp($event)" tabindex="1" class="modal wizard modal-animation ng-hide" ng-controller="EditEventsCtrl">
     <header>
-        <a class="fa fa-times close-modal" ng-click="close()"></a>
+        <a class="fa fa-times close-modal" ng-click="clearAndClose()"></a>
```
lkiesow: This would not fix the problem if you hit escape to close the modal, would it?

pmiddend: Oh, you might be right, good catch!

lkiesow:
Any update on this?

pmiddend:
Oh, thanks for the reminder, will look into it today.

pmiddend:
@lkiesow Ok, I've pushed a solution that's better in my opinion. I tested it both with the X button and with escape. Both work now, and no HTML has to be touched.

lkiesow:
Reviewing now…

lkiesow:
Seems to work fine now. Thanks for the patch → merged

PR #348, opened 2018-07-13
MH-12829 Make admin-ui statistics configurable

KatrinIhler:
This PR moves the definition of the event statistics displayed at the top-right of the admin UI to a listprovider configuration file so statistics can be removed, added or reordered.

This PR is the new version of PR #302 and uses Moment.js instead of SugarJS for configuring relative time spans so we don't introduce another big library.

screenshot from 2018-06-18 17-33-15

(The statistics are not properly aligned in this screenshot as mentioned my @lkiesow at #302, but this seems to be a problem with my system and not a bug introduced by this work.)

This PR also adds tests and documentation where deemed necessary.

Whether the proposed default configuration (introducing more filters for things like yesterday and tomorrow) makes sense is kinda debatable, I could also comment them out. Opinions?

_This work is sponsored by ETH._


docs/guides/admin/docs/configuration/ui.md
```diff
@@ -0,0 +1,72 @@
...
+instead of a string. This object has to contain a `relativeDateSpan` property which itself contains the fields `from`,
+`to` and `unit`. The `unit` defines the unit of time that is being considered, e.g. _hour_, _day_, _week_, _month_ or
+_year_, while `from` and `to` specify the beginning and end of the time span by defining an integer offset relative to
+the current hour, day, ... depending on the unit. So if the `unit` is defined as _day_, **0** is the current day while
+**-1** is yesterday and **1** is tomorrow. If the unit is _week_ instead, **0** is the current week while **-1** is the
+last and **1** the next week, and so on.
+
+Every date/time unit below the one defined by `unit` depends on whether the offset is defined by `to` or `from`. So if
+the unit is _day_, **from: -1** would be the beginning of yesterday (so the time is 00:00:00 in the user's timezone)
+while **to: -1** would be the end of yesterday (23:59:59). If the unit is *week*, **from: -2** is the beginning of last
```
staubesv: from -2 would be the beginning of the week before the last week, wouldn't it?

etc/listproviders/adminui.stats.properties
```diff
@@ -0,0 +1,65 @@
...
+#  "description": "DATES.THIS_WEEK",\
+#  "order":3}
+#THIS_MONTH=\
+#  {"filters": [{"name": "startDate", "filter":"FILTERS.EVENTS.START_DATE", \
+#                "value": {"relativeDateSpan": {"from": "0", "to": "0", "unit": "month"}}}],\
+#  "description": "DATES.THIS_MONTH",\
+#  "order":4}
+#THIS_YEAR=\
+#  {"filters": [{"name": "startDate", "filter":"FILTERS.EVENTS.START_DATE", \
+#                "value": {"relativeDateSpan": {"from": "1", "to": "1", "unit": "year"}}}],\
```
staubesv: Shouldn't this be from: 0 and to:0 ?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
```diff
@@ -111,6 +111,11 @@ public String apply(Date date) {
    * @throws JsonCreationException
    */
   public static <T> JSONObject generateJSONObject(Map<String, T> list) throws JsonCreationException {
+
+    if (list == null) {
```
staubesv, line 115: Just wondering: Why have you changed this? Returning an empty JSON object for an empty or non-existing list seems actually fine...

staubesv (changes requested):

docs/guides/admin/docs/configuration/ui.md
KatrinIhler: Yes, you're right, gonna fix that.

etc/listproviders/adminui.stats.properties
KatrinIhler: Also correct, I changed the configuration around too much...

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
KatrinIhler: The list being null is not okay here because then list.entrySet() will throw a NullpointerException. But you're right, one could return an empty json object instead of throwing the JsonCreationException. Not sure what makes more sense...

KatrinIhler:
@staubesv Fixed the first two points, unsure about the third.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
staubesv: Oh, I haven't looked at the code close enough: If it throw an exception before anyway, your change is an improvement. Let's just keep it.

staubesv:
Hi @KatrinIhler 

I found a last minor issue: Please add a mockup for STATS.json to /modules/admin-ui/src/test/resources/app/GET/admin-ng/resources so that the Admin-UI mockup ('cd modules/admin-ui' followed by 'grunt') shows the counters.

After that, this PR should be fine to go.

KatrinIhler:
@staubesv Done!

staubesv (approved):

PR #349, opened 2018-07-16
MH-13003, Stop changing technical start date after event started

JulianKniephoff:
This should fix the problem described in [MH-13003](https://opencast.jira.com/browse/MH-13003), however, I have to say that I am not that sure about the changes. I don't know whether the lines I removed weren't there for a reason; comment above them subtly suggests to me that they were. But I wouldn't know why you would want to override the technical start date with the bibliographical one when the event has started (or is finished) recording.

# The problem

What follows is a lengthy description of my findings regarding MH-13003, so that other people can reproduce the error and track it through the code.

The first part of the problem is that the `IndexServiceImpl#updateEventMetadata` method which the `/event/:id/metadata` endpoint delegates more or less immediately to treats the event differently based on what it calls the "event source", which---for all of our intents and purposes---just means whether the events technical start date lies in the past. (Workflows also play into this, but let's ignore this here; for details: I'm referring to the `switch`-statement in the aforementioned method and the `getEventSource` method in the same class.)

If it does, the event is handled by the "`ARCHIVE`" branch and a snapshot is taken via the asset manager, and everything is find from our point of view. If the technical start date does **not** lie in the past, though, the event is treated as "`SCHEDULED`" and handed to the scheduler service.
The scheduler service in turn calls `SchedulerServiceImpl#verifyActive`, which throws the exception we see in the log when the current time is after the end date of the event.

Now you might ask: *"How can that be? The index service things the event starts in the future, gives it to the scheduler and that complains that the event is in the past?!"* Well, behold the power of caches/indexes/other redundant storage systems! :smile:

The crux is that the scheduler asks the asset manager for the start date, which still has the original scheduled date (or the one that you set manually in the Scheduling tab after the fact, as described in MH-13003). The index service on the other hand has this information stored itself, too, and of course uses **that**. This is a problem once these two pieces of information get out of sync.

One place where that might happen is the one I changed. `EventIndexUtils#updateTechnicalDate` is called from `updateEvent` in the same class as a response to both, the asset manager and scheduler service messages generated by `IndexServiceImpl#updateEventMetadata`, and what it does is override the technical start date in the admin UI index with the bibliographical one **if** the former has already passed as reported, again, by the index service itself.

So step 2 in the description of MH-13003 generates a `takeSnapshot` message, in which the technical start date of the event **in the index service** is set to the future, which makes further metadata updates as in step 3 be handled by the scheduler, which still thinks that the event is in the past and thus refuses to update the metadata, which makes sense in my opinion. (Well, you might of course want to change meta data of a passed event, but keep in mind that the treatment of the event in question as an event that is scheduled as opposed to archived is already a faulty consequence of the above behavior. At least I think it is. :wink:)

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventIndexUtils.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventIndexUtils.java
@@ -479,17 +479,10 @@ public final class EventIndexUtils {
   }
 
   public static Event updateTechnicalDate(Event event) {
-    if (event.hasRecordingStarted()) {
-      // Override technical dates from recording if already started
+    if (StringUtils.isBlank(event.getTechnicalStartTime()))
       event.setTechnicalStartTime(event.getRecordingStartDate());
+    if (StringUtils.isBlank(event.getTechnicalEndTime()))
       event.setTechnicalEndTime(event.getRecordingEndDate());
-    } else {
-      // If this is an upload where the start time is not set, set the start time to same as dublin core
-      if (StringUtils.isBlank(event.getTechnicalStartTime()))
-        event.setTechnicalStartTime(event.getRecordingStartDate());
-      if (StringUtils.isBlank(event.getTechnicalEndTime()))
-        event.setTechnicalEndTime(event.getRecordingEndDate());
-    }
     return event;
   }
 


JulianKniephoff:
Interesting. The tests passed on the branch I originally made these changes on. But that was admittedly not quite `r/5.x`. I'm looking into it.

**EDIT**: No, in fact on my machine the tests still run successfully in this very branch this PR contains. Maybe this was a spurious fail? Can one of the committers maybe restart the build? Buildbot tells me I can't do that. :sweat_smile: 

**EDIT 2**: Okay, @lkiesow just informed me that I can ignore the buildbot failure. :smile:

karendolan:
@JulianKniephoff I could dig into the code to verify, but do you know if it is Ok for a technical start date to be null, or nullified, for an event after it starts recording, after ingest, or during an upload? 

The motivation for the question is to check if setting tecnical start date to null is an quick way to remove the event from the scheduler relm and ensure it will be processed in the usual way.

JulianKniephoff:
@karendolan no, sorry, I would have to do the same digging I'm afraid. Why do you ask?

karendolan:
@JulianKniephoff I ask because a person monitoring the processing of the recording in the Admin UI needs to know the scheduled technical start time up until the point of the end of the schedule event. But after the ingest, upload, or republish (in our case), the manager is concerned about the bibliographical start date that will be used for searching and displaying the event to consumers. The technical start date is useful for debugging after the fact, but no longer the primary date of concern when processing the metadata and materials for publication.

If the techncial start date should not be synchronized with the bibliographical start date after ingest, how can the user search by bibliographic start dates from the admin Index/UI/

JulianKniephoff:
Hmm I'm not sure I understand you correctly. Is the bibliographical start date in the search indices only stored as this copy of the technical start date? But then setting the latter to nothing wouldn't help either, right? The bibliographical start date would be lost either way. Also if that were really the case---if there is really no explicit storage space for the real bibliographical start date in the indices---that would indicate a pretty big design oversight to me, given your use case, which should be fixed a little closer to its root, imo.

But then again: This is exactly what I was hoping for, because I have probably too little knowledge still about many of the Opencast internals involved in this bug. Can you maybe provide a specific scenario that I can reconstruct locally, which breaks with this change?

JulianKniephoff:
Looking at `org.opencastproject.index.service.impl.index.event.Event`, there are two relevant fields: `technicalStartDate` and `recordingStartDate`. So there is definitely room to store both. Now of course if all the queries relevant to the display, filtering, sorting, ... in the UI use the former, I would see a need for these to be synced, but I think that would be a design smell in my opinion. Why not introduce a new field that reflects the use case you just gave: It would contain the technical start date before the recording started and the bibliographical one afterwards. This way the original fields would stay intact and can properly be used to make decisions about the actual state of the event.

**Or** the index service has to ask the scheduler service to make these decisions, which still contains the correct start date as I said before. But I guess that would be too slow? Which is why this information is stored in an index in the first place? :thinking:

So yeah, I'm a bit out of my depth here; any further feedback is greatly appreciated. :sweat_smile:

karendolan:
@JulianKniephoff  From the discussion, to keep the technicalStartDate isolated and not synched to secondary date sources, the updateTechnicalDate() method should be removed altogether. Uploaded events with SPACIAL and TEMPORAL location and start and stop times must have a NULL technicalStartDate to prevent design smell and represent that this value is not set and should not be  included in technical metadata searches. The Index endpoint should provide technical search services if that is what is needed by clients such as the scheduler service.

To go halfway for this pull, if it is not technically reasonable to avoid setting a technicalStartDate for uploads at this time, it would be more clear in the admin UI if the technicalStartDate for uploads is something like Jan 1, 1970, to represent an unknown to technical start time, to avoid confusion and design smell by mixing it with the bibliographic TEMPORAL start time of the event.

JulianKniephoff:
Okay so after some internal discussion here in Osnabrück, we decided to not touch the date semantics in the index service for this. Let the ElasticSearch update and index service rewrite/refactor deal with this as it turns out rather hairy issue. :wink:

We can fix this a step before already: Why does the index service even treat a completed event as scheduled only based on the date field? It should rather check whether it is done already somehow. This should be possible since this is exactly what the scheduler does as well and which eventually leads to the offending exception.

I will close this PR for now and create a new one once I cooked up a solution that does this. Thanks to @karendolan for the feedback, pushing me to explore the complexities of this topic a bit further. :smile:

PR #352, opened 2018-07-18
MH-13003: Implement detection of already recorded (as opposed to yet to be recorded, scheduled) events by the index service

JulianKniephoff:
This is my second attempt to fix [MH-13003](https://opencast.jira.com/browse/MH-13003). For a detailed description of the problem please refer to my first pull request: #349. I hope this does not brake any other use cases, but if it does not, then this was simpler than I thought. :smile:

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/Event.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/Event.java
@@ -36,7 +36,6 @@ import org.codehaus.jettison.mapped.Configuration;
 import org.codehaus.jettison.mapped.MappedNamespaceConvention;
 import org.codehaus.jettison.mapped.MappedXMLStreamReader;
 import org.codehaus.jettison.mapped.MappedXMLStreamWriter;
-import org.joda.time.DateTime;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -44,7 +43,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringWriter;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
Hi @JulianKniephoff, I'm thinking something is still wrong, or I'm not understanding what I'm doing.  I have a demo pyca hooked up to my localhost for testing.  After waiting for the capture to process...

STR:
- Open the metadata tab
- Change the start date to a week from today
- Change the description

Results:
- Start date change is persisted in the metadata page, and the overall events page after a refresh
- Description change is not persisted
- Further changes to the metadata, including the start date are *not* persisted
- The event's creation date is now equal to its start date (previously it was the correct, current date)

Stack Trace generated when failing to modify things:

`Caused by: org.opencastproject.index.service.exception.IndexServiceException: Unable to update scheduled event b47c3a75-b199-48b3-86b4-dccf1bbb1d0a
        at org.opencastproject.index.service.impl.IndexServiceImpl.updateEventMetadata(IndexServiceImpl.java:1284)
        at org.opencastproject.index.service.impl.IndexServiceImpl.updateAllEventMetadata(IndexServiceImpl.java:1166)
        at org.opencastproject.adminui.endpoint.AbstractEventEndpoint.updateEventMetadata(AbstractEventEndpoint.java:1044)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)[:1.8.0_171]`

gregorydlogan:
Bah, it helps to have the correct build running when testing.  This does work as expected as far as I'm concerned, with the sole caveat that this is also change the event creation date which is likely not what we want.

JulianKniephoff:
Ah, right, I agree that this is surprising, at least. However, I stumbled upon many a surprise about date handling in Opencast while working on this. We also had a few internal discussions here at ELAN about this issue and apparently, what you describe actually **is** what we, or I should say some people, want: See [MH-12250](https://opencast.jira.com/browse/MH-12250). :confounded: 

Agreeing to disagree on this apparently hairy topic for now, we decided not to touch the date handling logic in the course of the work on this issue, especially in light of the potentially large changes that might come to the index service due to the crowd funding. With this in mind I tried to make the minimal change to fix MH-13003 and I think any date handling weirdness that we now see (that we **couldn't** even see before, mind you) is an independent issue, that probably has to be discussed first anyway, before "fixing" anything. :thinking: 

And while I do have an opinion on the matter (that in the case you brought up above aligns with yours, I think) I am not the one to lead this discussion as I still lack the Opencast experience and knowledge of all the use cases. :disappointed_relieved:

gregorydlogan:
As discussed on list, my complaints are more or less expected behaviour so I'm merging this as is.

PR #353, opened 2018-07-19
MH-13006 Waveform operation cleanup creates problem with asynchronous NFS

KatrinIhler:
Currently the WaveformOperationHandler deletes the whole folder for the mediapackage from the workspace after the operation is done. This can create problems when using an asynchronous NFS: Sometimes the deleted folder is still visible for the  worker node (after deleting it from the admin node) at the beginning of the next operation (an image operation in this case), but when FFMPEG tries to open the output file it fails because the directory that's supposed to contain it no longer exists.

The (sort of hacky and not really satisfactory, but working) solution is to make the WaveformOperationHandler clean up only the files from the workspace but leave the folder structure, thus avoiding this problem. The rest of it should be cleaned up at the end of the workflow...


gregorydlogan:
@KatrinIhler Conflicts detected!

KatrinIhler:
@gregorydlogan Thanks, fixed!

PR #359, opened 2018-07-25
MH-13013, Unused code in scheduler

lkiesow:
The scheduler's calendar code contains some unnecessary code which will
always produce the same outcome and which could simply be removed with
no change in behavior.

### DIFF
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
@@ -1622,19 +1622,6 @@ public class SchedulerServiceImpl extends AbstractIndexProducer implements Sched
 
       CalendarGenerator cal = new CalendarGenerator(seriesService);
       for (ARecord record : records) {
-        boolean blacklisted;
-        // isBlacklisted() methods are not implemented in the persistence layer and return always false
-//        try {
-//          //blacklisted = isBlacklisted(record.getMediaPackageId());
-//        } catch (NotFoundException e) {
-//          continue;
-//        }
-        blacklisted = false;
-
-        // Skip blacklisted events
-        if (blacklisted)
-          continue;
-
         Opt<MediaPackage> optMp = record.getSnapshot().map(episodeToMp);
 
         // If the event media package is empty, skip the event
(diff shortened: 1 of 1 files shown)


modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1622,19 +1622,6 @@ public String getCalendar(Opt<String> captureAgentId, Opt<String> seriesId, Opt<
...
-        // isBlacklisted() methods are not implemented in the persistence layer and return always false
-//        try {
-//          //blacklisted = isBlacklisted(record.getMediaPackageId());
-//        } catch (NotFoundException e) {
-//          continue;
-//        }
-        blacklisted = false;
-
-        // Skip blacklisted events
-        if (blacklisted)
```
lkiesow, line 1635: `blacklisted` is always `false` and this will never be executed.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1667,9 +1654,7 @@ public String getCalendar(Opt<String> captureAgentId, Opt<String> seriesId, Opt<
         try {
           cal.addEvent(optMp.get(), catalogOpt.get(), agentId, start, end, lastModified, toPropertyString(caMetadata));
         } catch (Exception e) {
-          logger.warn("Error adding event '{}' to calendar, event is not recorded: {}", record.getMediaPackageId(),
-                  getStackTrace(e));
-          continue;
+          logger.warn("Error adding event '{}' to calendar, event is not recorded", record.getMediaPackageId(), e);
```
lkiesow, line 1657: The logger can handle stack traces by itself and continue is unnecessary since it's the end of the loop anyway.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1685,9 +1670,6 @@ public String getCalendar(Opt<String> captureAgentId, Opt<String> seriesId, Opt<
       return cal.getCalendar().toString();
 
     } catch (Exception e) {
-      if (e instanceof SchedulerException)
-        throw e;
-      logger.error("Failed getting calendar: {}", getStackTrace(e));
```
lkiesow, line 1690: - `SchedulerException` is never thrown inside this block. Hence, the condition is always false.
- The exception does not need to be logged since it's thrown again anyway. Logging statements like this will just cause the same problem to be logged multiple times.

PR #361, opened 2018-07-25
MH-13015, 5.x database upgrade scripts

lkiesow:
This patch adds additional uphrade scripts which only use statements
supported by MySQL and older MariaDB versions. The previous script still
exists but has been moved to a separate `mariadb.sql` file.
    
Documentation about these two scripts and the difference between them
has also been added so users can choose what their need during an
upgrade.

This pull request also adds additional Travis CI tests to ensure a proper upgrade
path from Opencast 2.2 onwards.

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -22,7 +22,9 @@ As part of removing mentions of the old name Matterhorn, all database table name
 with `oc_`.  This requires an database schema update. As with all database migrations, we recommend to make a database
 backup before attempting the upgrade.
 
-You can find the database upgrade script at `docs/upgrade/4_to_5/mysql5.sql`.
+You can find the database upgrade script in `docs/upgrade/4_to_5/`. There are two scripts which are essentially doing
+the same thing with the MariaDB one running a few more checks on the database for safety which cannot be run like this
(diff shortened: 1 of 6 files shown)


docs/guides/admin/docs/upgrade.md
```diff
@@ -22,7 +22,9 @@ As part of removing mentions of the old name Matterhorn, all database table name
 with `oc_`.  This requires an database schema update. As with all database migrations, we recommend to make a database
 backup before attempting the upgrade.
 
-You can find the database upgrade script at `docs/upgrade/4_to_5/mysql5.sql`.
+You can find the database upgrade script in `docs/upgrade/4_to_5/`. There are two scripts which are essentially doing
```
gregorydlogan, line 25: If this is going into develop, shouldn't this be 5_to_6?  Or maybe we should reword this to fit a more generic format?  For example:

You can find the database upgrade scripts in `docs/upgrade/X_to_Y/`.  As an example, to upgrade from Opencast 4 to Opencast 5 you would use the scripts in `4_to_5`.

lkiesow: Good catch. It's fixing a 5.x bug and should go into `r/5.x`. I'll change the target branch…

PR #362, opened 2018-07-25
MH-13016, Workflow display order not working in editor screen

rute-santos:


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
```diff
@@ -366,7 +366,8 @@ public Response getVideoEditor(@PathParam("mediapackageid") final String mediaPa
     // Get workflows
     List<JValue> jWorkflows = new ArrayList<>();
     for (WorkflowDefinition workflow : getEditingWorkflows()) {
-      jWorkflows.add(obj(f("id", v(workflow.getId())), f("name", v(workflow.getTitle(), Jsons.BLANK))));
+      jWorkflows.add(obj(f("id", v(workflow.getId())), f("name", v(workflow.getTitle(), Jsons.BLANK)),
+              f("displayOrder", v(workflow.getDisplayOrder())));
```
karendolan: @rute-santos does this need a "Jsons.BLANK" as the third f param?

rute-santos: No. The default display order is zero.

lkiesow:
Seems to work as expected → merged

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
karendolan: My comment wasn't actually about the return value, I thought this had been throwing an error because of missing third param. Apparently not!, sorry.

PR #365, opened 2018-07-30
MH-13019 Remove broken PreSelectFromDirective

KatrinIhler:
Remove the PreSelectFromDirective since it currently doesn't work and additionally breaks the functionality of setting a default value for a listprovider.

_This work is sponsored by the ETH Zürich._


staubesv:
Hi @KatrinIhler I disagree with this change: In our case, we do need the pre-select-from directive to avoid users being asked to select the only available options and it does work as far as I know.

There are two other open PRs that might be of your interest: https://github.com/opencast/opencast/pull/334 fixes the pre-select-from directive for series and https://github.com/opencast/opencast/pull/312 removes it from the metadata property sheets where it doesn't make too much sense.

staubesv (changes requested):
Hi @KatrinIhler 

As for us, we do need this directive at the following two places:

1. Add Event->Processing: Select the processing workflow
2. Delete Event: Select the retraction workflow

The reason is that we only offer one workflow and we really don't want to ask the user to select the only available option each time.

The directive does work correctly at those places.

The directive does not work correctly in the metadata property sheets (it is not aware of the option to select "no option" for non-mandatory fields) and I also think it doesn't make too much sense to use it there.

In case you only have problems with the list provider default values used in the metadata property sheet, #312 would remove the pre-select-from directive there and should solve the problem.

What do you think?

KatrinIhler:
@staubesv So that explains why I thought it didn't work, it really doesn't where the metadata fields are concerned, but it does work for the workflows! There it also makes sense since a workflow has to be selected, while most metadata fields can also be left empty...
I'm going to check whether a default workflow also causes problems with the preselection. Otherwise your PR would be exactly what I'm looking for!

PR #366, opened 2018-07-30
MH-13020 Read listproviders as UTF-8

KatrinIhler:
Load the configuration files for the listproviders with encoding set to UTF-8 instead of the default Latin-1 because otherwise characters like ü aren't displayed correctly.

_This work is sponsored by the ETH Zürich._

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/impl/ListProvidersScanner.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/impl/ListProvidersScanner.java
@@ -27,7 +27,6 @@ import org.opencastproject.index.service.resources.list.api.ResourceListProvider
 import org.opencastproject.index.service.resources.list.api.ResourceListQuery;
 import org.opencastproject.security.api.Organization;
 
-import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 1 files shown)


smarquard:
Just about "the default Latin-1": the default isn't necessarily Latin-1; it's a platform or JVM default. So these types of problems can show up on some systems but not others depending on local configuration. It's always best to specify explicitly as this change does.

lkiesow (approved):

KatrinIhler:
@smarquard Actually, the Java properties seem to expect Latin-1 as default independently of the platform: http://www.baeldung.com/java-properties

smarquard:
@KatrinIhler, thanks for clarification. I did not look closely enough, and was thinking of InputStreamReader which defaults to the system character set (or -Dfile.encoding=... value).

So the lesson from here is whenever we read from files containing text we should assume they may contain UTF-8 and use InputStreamReader which reads characters, allowing the character set to be explicitly set, rather than FileInputStream which reads bytes, and leaves the choice of encoding to code lower down which could apply an unwanted character set, as Properties seems to do when using load(InputStream ...) rather than load(Reader ...).

staubesv (approved):

PR #369, opened 2018-08-07
Documentation: message-broker: binding localhost

pawohl:
Default is not accepting connections from all hosts; only localhost. This changed with 31e88e23982c61c859ebd33879086b3930890538


gregorydlogan:
Looks good, thanks for the contribution @pawohl!

PR #370, opened 2018-08-08
MH-13023 Let default value fulfill requirement

KatrinIhler:
This fixes a problem where a required metadata field is not considered filled even though there is a default value set.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
@@ -29,6 +29,14 @@ angular.module('adminNg.services')
 
         this.requiredMetadata = {};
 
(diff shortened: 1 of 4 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-group/metadata.js
```diff
@@ -1,3 +1,4 @@
+
```
staubesv: Please remove that newline (or is there a specific reason to add it?)

staubesv (changes requested):
I've just found a minor cosmetic issue (dubious newline). Once that fixed, this PR is ready to go.

KatrinIhler:
@staubesv Fixed, thanks for noticing!

staubesv (approved):

PR #371, opened 2018-08-08
MH-13024 Video editor does not display information when being opened while an event is being processed

staubesv:
See https://opencast.jira.com/browse/MH-13024

Before:

without_patch

After:

with_patch

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
@@ -284,8 +284,6 @@ public class ToolsEndpoint implements ManagedService {
                   @RestResponse(description = "Media package not found", responseCode = SC_NOT_FOUND) })
   public Response getVideoEditor(@PathParam("mediapackageid") final String mediaPackageId)
           throws IndexServiceException, NotFoundException {
-    if (!isEditorAvailable(mediaPackageId))
-      return R.notFound();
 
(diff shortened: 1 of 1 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
```diff
@@ -284,8 +284,6 @@ private Boolean elementHasPreviewFlavor(MediaPackageElement element) {
                   @RestResponse(description = "Media package not found", responseCode = SC_NOT_FOUND) })
   public Response getVideoEditor(@PathParam("mediapackageid") final String mediaPackageId)
           throws IndexServiceException, NotFoundException {
-    if (!isEditorAvailable(mediaPackageId))
```
staubesv, line 287: This checks whether event.source is "ARCHIVE". While being processed, event.source is "WORKFLOW" though.

I don't think that it hurts to return editor.json while the event is being processed: This is a read-only request and the user already has opened the video editor anyway. Why shouldn't we allow that while an event is being processed?

slampunk:
Looks good, works as advertised. Can be merged.

The feature makes sense, and doesn't impact backend operations as POSTing new edits for events under transaction results in a 500 response.

I suppose with #377 being brought forward, there'd be a notification at the top of the page as well, notifying users of the event's state (i.e. informing of the inability to save edits for events currently currently undergoing transactions).

lkiesow:
@slampunk, thanks for the review → merged

PR #372, opened 2018-08-08
MH-13025: Add workflow API to external API

mtneug:
This extends the external API with endpoints for workflow instances and definitions.

One question I had during the development is, whether it makes sense to add create, update, and delete endpoints for workflow definitions assuming those endpoints only operate in-memory and don't actually persist workflows as XML in the configuration directory. What do you think?

For the full documentation see the [`docs/guides/developer/docs/api/workflow-api.md`](https://github.com/mtneug/opencast/blob/f/mh-13025-workflow-api/docs/guides/developer/docs/api/workflow-api.md) guide.

**tl;dr:** Extension with the following endpoints:
```
GET    /api/workflows
POST   /api/workflows
GET    /api/workflows/{id}
PUT    /api/workflows/{id}
DELETE /api/workflows/{id}

GET    /api/workflow-definitions
GET    /api/workflow-definitions/{id}
```


mtneug:
I updated the PR with create, update, and delete endpoints for workflow instances. I also removed the preparatory code for write endpoints of workflow definitions. Next, I will work on adding tests.

docs/guides/developer/docs/api/workflow-api.md
```diff
@@ -0,0 +1,587 @@
...
+
+### POST /api/workflows
+
+Creates a workflow instance.
+
+Form Parameters                  | Required |Type                             | Description
+:--------------------------------|:---------|:--------------------------------|:-----------
+`event_identifier`               | yes      | [`string`](types.md#basic)      | The event identifier this workflow should run against
+`workflow_definition_identifier` | yes      | [`string`](types.md#basic)      | The identifier of the workflow definition to use
+`parent_identifier`              | no       | [`integer`](types.md#basic)     | The optional parent workflow identifier
```
staubesv: This (the "parent_identifier") seems quite "internal" to me. Is this required in the External API?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,629 @@
...
+      logger.error("The workflow service was not able to get the workflow instance: {}", getStackTrace(e));
+      return ApiResponses.serverError("Could not retrieve workflow instance, reason: '%s'", getMessage(e));
+    }
+
+    return ApiResponses.Json.ok(acceptHeader, workflowInstanceToJSON(wi, withOperations, withConfiguration));
+  }
+
+  @PUT
+  @Path("{workflowInstanceId}")
+  @RestQuery(name = "createworkflowinstance", description = "Creates a workflow instance.", returnDescription = "", pathParameters = {
```
staubesv: create?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,629 @@
...
+        } catch (ParseException e) {
+          return RestUtil.R.badRequest("Passed parameter 'configuration' is invalid JSON.");
+        }
+      }
+
+      // Start workflow
+      WorkflowInstance wi = workflowService.start(wd, mp, parentIdentifier, properties);
+      return ApiResponses.Json.created(acceptHeader, URI.create(getWorkflowUrl(wi.getId())),
+              workflowInstanceToJSON(wi, withOperations, withConfiguration));
+    } catch (UnauthorizedException e) {
```
staubesv: You might want to add...

```
    } catch (IllegalStateException e) {
      final ApiVersion requestedVersion = ApiMediaType.parse(acceptHeader).getVersion();
      return ApiResponses.Json.conflict(requestedVersion, obj(f("message", v(getMessage(e), BLANK))));
    }
```

Otherwise, clients receive a HTTP 500 Server Error in case the workflow could not be started because a workflow is already running for the event.

Example response for the code above in that situation:

`{"message":"IllegalStateException: Can't start workflow 'Fast Testing Workflow' for media package '8d3c8cf0-99e9-4821-9f4b-dec95598c7e2' because another workflow is currently active."}`

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,629 @@
...
+  /** OSGi DI */
+  public void setIndexService(IndexService indexService) {
+    this.indexService = indexService;
+  }
+
+  /**
+   * OSGi activation method
+   */
+  void activate(ComponentContext cc) {
+    this.serverUrl = "http://localhost:8080";
```
staubesv: Maybe add a constant for this value as you are using it more than once...

etc/security/mh_default_org.xml
```diff
@@ -190,6 +190,10 @@
     <sec:intercept-url pattern="/api/series/*/properties" method="GET" access="ROLE_ADMIN, ROLE_API_SERIES_PROPERTIES_VIEW"/>
     <sec:intercept-url pattern="/api/version" method="GET" access="ROLE_ADMIN, ROLE_API"/>
     <sec:intercept-url pattern="/api/version/*" method="GET" access="ROLE_ADMIN, ROLE_API"/>
+    <sec:intercept-url pattern="/api/workflow" method="GET" access="ROLE_ADMIN, ROLE_API_WORKFLOW_INSTANCE_VIEW"/>
```
staubesv: Everywhere in this file: It is /api/workflow**s** (not /api/workflow)

docs/guides/developer/docs/api/workflow-api.md
mtneug: When I implemented this, I checked whether the parent ID is used at all and the only place where it was set is in [WorkflowRestService.java](https://github.com/opencast/opencast/blob/1871a85d8a1b3cc51fea714c1fecc49aabbf377a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java#L566-L619). What is a parent/child workflow anyway? So I figured that maybe some adopters are using this or it is dead code. Should I remove it then?

staubesv: I would recommend to not include it in the External API. If you have the time, it would make sense to remove it completely from the entire codebase (if unused), but that is not in the scope of this PR.

docs/guides/developer/docs/api/workflow-api.md
```diff
@@ -0,0 +1,577 @@
...
+
+Filter Name                      | Description
+:--------------------------------|:-----------
+`state`                          | Workflow instances that are in this state. Can be included multiple times (inclusive or).
+`state_not`                      | Workflow instances that are not in this state. Can be included multiple times (inclusive or).
+`current_operation`              | Workflow instances that are currently executing this operation. Can be included multiple times (inclusive or).
+`current_operation_not`          | Workflow instances that are currently not executing this operation. Can be included multiple times (inclusive or).
+`workflow_definition_identifier` | Workflow instances that use this workflow definition
+`event_identifier`               | Workflow instances where the identifier of the processed event matches
+`event_title`                    | Workflow instances where the title of the processed event matches
```
staubesv, line 31: Just wondering... does the workflow need to be that tightly coupled to the event and series metadata? (separation of concerns)

Maybe consider to omit all filter & sort option that relate to events and series.

mtneug: I have no strong feelings about this. I just used all the available filters and sorting keys of workflow queries. Given that this is implemented in the workflow service, maybe there are use cases where this is necessary? Although, I cannot think of any right now.

mtneug:
The functionality and tests are now implemented. There are only a couple of open questions from my side (see inlined `TODOs` and `FIXMEs`).

docs/guides/developer/docs/api/workflow-api.md
staubesv: I think that is a legacy of Matterhorn 1.x: The old admin UI displayed workflow instances (instead of events). Those filters and sort options made it possible to filter/sort workflow instances in an "event-like manner".

If you don't need this for whatever you are working on this PR, I suggest to omit it. I also suggest to bring this up at the technical meeting tomorrow to see what others think...

staubesv:
@mtneug Note that I am currently on standby mode. Please let me know once I should continue the review.

mtneug:
@staubesv have we come to a conclusion about event/series metadata? Apart from that, please continue the review.

Looking at your PR #387, I have updated the `event_created` filter accordingly.

Also, could you comment on wether camelCase, snake_case, or just concatenation should be used for the various API fields (see e.g. `GET /api/events`: `textFilter` vs. `start_date` vs. `withmetadata`)?

staubesv:
Hi @mtneug 

Let's "decide" the events/series metadata stuff tomorrow in the technical meeting.

The naming stuff is unclear and inconsistent... I would suggest you tried to "fit into" existing names ;-)

docs/guides/developer/docs/api/types.md
```diff
@@ -353,3 +353,44 @@ __Example__
...
+`description`                | [`string`](#basic)           | The description of this workflow operation definition
+`configuration`              | [`property`](#property)      | The configuration for this workflow operation definition
+`if`                         | [`string`](#basic)           | The condition of this workflow operation definition that is required to be true for this operation to be executed
+`unless`                     | [`string`](#basic)           | The condition of this workflow operation definition that is required to be false for this operation to be executed
+`fail_workflow_on_error`     | [`boolean`](#basic)          | Whether an error during this workflow operation definition fails the workflow
+`exception_handler_workflow` | [`string`](#basic)           | The identifier of the exception handler workflow of this workflow operation definition
+`retry_strategy`             | [`string`](#basic)           | The retry strategy of this workflow operation definition in case of an error ("none": no retry, "retry": restart the operation, "hold": keep the operation in hold state)
```
staubesv: Suggestion: The retry_strategy is an enumeration whose values are documented at multiple locations. Could you introduce a "data type" for this so that the values (none, retry, hold) are just defined once?

docs/guides/developer/docs/api/types.md
```diff
@@ -353,3 +353,44 @@ __Example__
...
+
+The workflow operation_instance entries are represented as JSON objects with the following fields:
+
+Field                        | Type                                 | Description
+:----------------------------|:-------------------------------------|:-----------
+`identifier`                 | [`integer`](#basic)                  | The unique identifier of this workflow operation instance (equals the job identifier)
+`operation`                  | [`string`](#basic)                   | The operation of this workflow operation instance
+`description`                | [`string`](#basic)                   | The description of this workflow operation instance
+`configuration`              | [`property`](#property)              | The configuration for this workflow operation instance
+`state`                      | [`string`](#basic)                   | The state of this workflow operation instance ("instantiated", "running", "paused", "succeeded", "failed", "skipped", "retry")
```
staubesv: Suggestion: The field state is an enumeration whose values are currently listed at multiple places in the documentation. Could you introduce a "data type" for it to avoid having the values (instantiated, running,...) listed at multiple locations?

docs/guides/developer/docs/api/workflow-api.md
```diff
@@ -0,0 +1,585 @@
...
+`404 (NOT FOUND)`: The workflow instance could not be found.
+
+### DELETE /api/workflows/{workflow_instance_id}
+
+Deletes a workflow instance.
+
+__Response__
+
+`204 (NO CONTENT)`: The workflow instance has been deleted.<br/>
+`400 (BAD REQUEST)`: The workflow instance cannot be deleted in this state.<br/>
```
staubesv: That should not be a 400 as this is not a client error. Maybe 409 (conflicts)? See https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

docs/guides/developer/docs/api/workflow-api.md
```diff
@@ -0,0 +1,585 @@
...
+Field                 | Type                                                       | Description
+:---------------------|:-----------------------------------------------------------|:-----------
+`identifier`          | [`string`](types.md#basic)                                 | The unique identifier of this workflow instance
+`title`               | [`string`](types.md#basic)                                 | The title of this workflow instance
+`description`         | [`string`](types.md#basic)                                 | The description of this workflow instance
+`tags`                | [`array[string]`](types.md#array)                          | The (potentially empty) list of workflow tags of this workflow instance
+`configuration_panel` | [`string`](types.md#basic)                                 | The configuration panel of this workflow instance
+`operations`          | [`array[operation_instance]`](types.md#operation_instance) | The list of operations of this workflow instance
+
+`400 (BAD REQUEST)`: The request is invalid or inconsistent.<br/>
```
staubesv, line 388: Probably should also return a dedicated status code in case an invalid state transition is attempted as the client cannot know the current state of the workflow on the server (it might have changed since the last time asked)...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,608 @@
...
+              break;
+            case STOPPED:
+              workflowService.stop(wi.getId());
+              break;
+            case RUNNING:
+              if (currentState == WorkflowInstance.WorkflowState.INSTANTIATED
+                      || currentState == WorkflowInstance.WorkflowState.PAUSED) {
+                workflowService.resume(wi.getId());
+              } else {
+                return RestUtil.R.badRequest(
```
staubesv: Probably a 409 as the client cannot know that current status of a workflow on the server (might have changed since last time updated)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,608 @@
...
+                        String.format("Cannot resume from workflow state '%s'", currentState.toString().toLowerCase()));
+              }
+              break;
+            default:
+              return RestUtil.R.badRequest(
```
staubesv: Same here...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowsEndpoint.java
```diff
@@ -0,0 +1,608 @@
...
+          @RestResponse(description = "The workflow instance has been deleted.", responseCode = HttpServletResponse.SC_NO_CONTENT),
+          @RestResponse(description = "The workflow instance cannot be deleted in this state.", responseCode = HttpServletResponse.SC_BAD_REQUEST),
+          @RestResponse(description = "The user doesn't have the rights to make this request.", responseCode = HttpServletResponse.SC_FORBIDDEN),
+          @RestResponse(description = "The specified workflow instance does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
+  public Response deleteWorkflowInstance(@HeaderParam("Accept") String acceptHeader,
+          @PathParam("workflowInstanceId") Long id) {
+    try {
+      workflowService.remove(id);
+    } catch (WorkflowStateException e) {
+      return RestUtil.R.badRequest("Cannot delete workflow instance in this workflow state");
```
staubesv: Probably 409...

staubesv (changes requested):
I've done a first round of review and found some minor issue.

docs/guides/developer/docs/api/workflow-api.md
```diff
@@ -0,0 +1,585 @@
...
+`event_title`                    | Workflow instances where the title of the processed event matches
+`event_created`                  | Workflow instances where the processed event was created between two dates
+`event_creator`                  | Workflow instances where the creator of the processed event matches
+`event_contributor`              | Workflow instances where the contributor of the processed event matches
+`event_language`                 | Workflow instances where the language of the processed event matches
+`event_license`                  | Workflow instances where the license of the processed event matches
+`event_subject`                  | Workflow instances where the subject of the processed event matches
+`series_identifier`              | Workflow instances where the identifier of series of the processed event matches
+`series_title`                   | Workflow instances where the title of series of the processed event matches
+`textFilter`                     | Workflow instances where any part of the Workflow definition's metadata fields match this value
```
staubesv, line 40: Hi @karendolan 

I don't have a strong opinion about which exact fields workflows can be filtered or sorted by, but I still think that  mixing up different kinds of objects (workflows vs. events vs. series) is undesirable from an architectural point of view.
Think of our database that will be never able to deliver answers to such filters/sort options (unless we duplicate the metadata in the workflow database table which we hopefully don't) and the need to have another data source ("search index") just to accumulate data of different services to be able to filter/sort across service boundaries. Or why do we have a search index for workflows at all?

I'm talking about the following fields...

- event_title
- event_created
- event_creator
- event_contributor
- event_language
- event_license
- event_subject
- series_identifier
- series_title

... which are not fields of a workflow instance to my understanding  (The event identifier in means of media package identifier is, as the workflow instance is bound to a media package).

I'm fine if this added to the External API if somebody actually needs this, but could we at least try to keep it up to an minimum?

I see that things like event_title, series_identifier and series_title can be useful. But do we really need all the other filters & sort options like event_language, for example?

Can you please have a look at this?

What do you think? Include all those fields? Can some be "safely" omitted?

You might guess that having all those copies of the same metadata values at various locations across the entire system and (trying to) keep them in sync is not exactly what I love most about Opencast... ;-)

docs/guides/developer/docs/api/workflow-api.md
mtneug: All 4xx codes are client errors 😉 Yes, 409 would be a better option.

docs/guides/developer/docs/api/workflow-api.md
mtneug: I added `409 (CONFLICT)` as possible response. Note that `400` is also returned when the configuration could not be parsed. This is clearly a bad request.

docs/guides/developer/docs/api/workflow-api.md
mtneug: Just a quick note: As noted in the `FIXME`, sorting by `event_identifier` currently leads to an exception. Solr complains that it `can not sort on undefined field: mediapackageid_sort`. I have created [MH-13066](https://opencast.jira.com/browse/MH-13066) to track the issue. Fixing this is outside the scope of this pull request and I also don't know what the difference between this and the other fields is.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowDefinitionsEndpoint.java
```diff
@@ -0,0 +1,272 @@
...
+      return ApiResponses.serverError("Could not retrieve workflow definitions, reason: '%s'", getMessage(e));
+    }
+
+    // Apply filter
+    if (StringUtils.isNotBlank(filter)) {
+      for (String f : filter.split(",")) {
+        int sepIdx = f.indexOf(':');
+        if (sepIdx < 0 || sepIdx == f.length() - 1) {
+          logger.info("No value for filter {} in filters list: {}", f, filter);
+          continue;
```
staubesv, line 144: Hmm... shouldn't we return a 400 bad request here?

Also: Can we filter for "no tag"?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowDefinitionsEndpoint.java
```diff
@@ -0,0 +1,272 @@
...
+        String name = f.substring(0, sepIdx);
+        String value = f.substring(sepIdx + 1);
+
+        if ("tag".equals(name))
+          workflowDefinitions = workflowDefinitions.filter(wd -> ArrayUtils.contains(wd.getTags(), value));
+      }
+    }
+
+    // Apply sort
+    // TODO: this seems to not function as intended
```
staubesv, line 155: Why? Just did some basic tests and it seems to work as expected?

staubesv (approved):

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowDefinitionsEndpoint.java
mtneug: This is inline with the other APIs (e.g. [events](https://github.com/opencast/opencast/blob/15b367378cf89b236e116aeb7ad031f340cf70d7/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java#L609-L612) and [series](https://github.com/opencast/opencast/blob/15b367378cf89b236e116aeb7ad031f340cf70d7/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java#L199-L202)).

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowDefinitionsEndpoint.java
mtneug: I only did some quick manuell tests. Maybe I did something wrong. Glad it worked for you.

mtneug:
@staubesv thank you for the review!

modules/external-api/src/main/java/org/opencastproject/external/endpoint/WorkflowDefinitionsEndpoint.java
staubesv: @mtneug Note that I've already merged this PR...

PR #373, opened 2018-08-09
MH-13026 Update Mac OS X 'Install from source' documentation

jaysuz:
https://opencast.jira.com/browse/MH-13026


lkiesow:
While I'm no Mac user, the update looks good and seems correct → merged

PR #377, opened 2018-08-10
MH-13031 Active transaction notification on top

KatrinIhler:
Show the warning notification that the event shown in the videoeditor is currently being processed also on top like it used to be in 4.x.


gregorydlogan:
So, I think you've succeeded.  But perhaps too well...

As an example, I processed a workflow, then used the video editor to cut out a couple of sections, then jumped back into the editor while it was processing.  I see this on load

screenshot

I get it, there's a transaction in progress :)  The editor UI has top and bottom as expected, but I'm thinking perhaps we can trim down the warnings on the playback tab?

KatrinIhler:
@gregorydlogan I removed the event-access notifications from the metadata and comments tabs, what do you think?

gregorydlogan:
Looks good to me!

PR #378, opened 2018-08-10
MH-13022: fixed LTI highly trusted keys being discarded

t-schroeder:


t-schroeder:
This fixes a bug in the way the constructor with three arguments is called. That bug was introduced with https://github.com/opencast/opencast/commit/ef7bc8e94936b426bec5cb52f846d81c8c903657.

staubesv (approved):

PR #379, opened 2018-08-10
MH-12016: Scrolling role fetch

gregorydlogan:
Previously the list of the roles in the user creation/modification modal would load additional roles when scrolled to  the bottom.  This behaviour is currently broken, and this patch resolves the issue.

To test, modify modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js, set the limit to 20 in the UserRolesResource.query, and then rebuild (or better, use the proxy).  Scrolling to the bottom should pull up the *rest* of the roles from the system.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
@@ -24,8 +24,8 @@ angular.module('adminNg.controllers')
 .controller('UserCtrl', ['$scope', 'Table', 'UserRolesResource', 'UserResource', 'UsersResource', 'JsHelper', 'Notifications', 'Modal', 'AuthService', 'underscore',
     function ($scope, Table, UserRolesResource, UserResource, UsersResource, JsHelper, Notifications, Modal, AuthService, _) {
         $scope.manageable = true;
-        var roleSlice = 100;
-        var roleOffset = roleSlice; //Note that the initial offset is the same size as the initial slice so that the *next* slice starts at the right place
+        $scope.roleSlice = 100;
+        $scope.roleOffset = $scope.roleSlice; //Note that the initial offset is the same size as the initial slice so that the *next* slice starts at the right place
         var loading = false;
         var showExternalRoles = false; // Should the External Roles tab be visible
 
@@ -40,7 +40,7 @@ angular.module('adminNg.controllers')
         });
 
         $scope.role = {
-            available: UserRolesResource.query({limit: $scope.roleSlice, offset: 0, filter: 'role_target:USER'}), // Load all the internal roles
+            available: UserRolesResource.query({limit: $scope.roleSlice, offset: 0, filter: 'role_target:USER'}), // Load the first rolesSlice internal roles
             external: [],
             selected:  [],
             derived: [],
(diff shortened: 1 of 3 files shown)


staubesv:
Hi @gregorydlogan 

I have not managed to test this. I've changed "var roleSlice = 20;" in userController.js and changed "return $resource('/admin-ng/resources/ROLES.json?limit=20', {}, {" in userRolesResource.js.

I will see 20 roles then in Organization->Users->User Details->Roles.

How exactly should I test this?

gregorydlogan:
Hi @staubesv, scroll the list of roles to the bottom, more should show up.  There is an issue where if the initial set of roles is less than the height of the scrollbox it won't trigger, but that's been present since the initial inclusion of this code and is unlikely considering our default roles are much, much larger than that scrollbox usually is.

If 20 is too few for you, experiment with larger numbers.  As long as you're initially loading less than the full set of roles things should work.

staubesv:
Hi @gregorydlogan 

I have a r/5.x with the following modifications:

```
diff --git a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
index 45c3c2ae6..449478950 100644
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
@@ -24,7 +24,7 @@ angular.module('adminNg.controllers')
 .controller('UserCtrl', ['$scope', 'Table', 'UserRolesResource', 'UserResource', 'UsersResource', 'JsHelper', 'Notifications', 'Modal', 'AuthService', 'underscore',
     function ($scope, Table, UserRolesResource, UserResource, UsersResource, JsHelper, Notifications, Modal, AuthService, _) {
         $scope.manageable = true;
-        var roleSlice = 100;
+        var roleSlice = 20;
         var roleOffset = roleSlice; //Note that the initial offset is the same size as the initial slice so that the *next* slice starts at the right place
         var loading = false;
         var showExternalRoles = false; // Should the External Roles tab be visible
diff --git a/modules/admin-ui/src/main/webapp/scripts/shared/directives/loadingScrollerDirective.js b/modules/admin-ui/src/main/webapp/scripts/shared/directives/loadingScrollerDirective.js
index 09e5b7202..a3f395d65 100644
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/loadingScrollerDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/loadingScrollerDirective.js
@@ -38,7 +38,7 @@ angular.module('adminNg.directives')
 
             element.bind('scroll', function () {
                 if (raw.scrollTop + raw.offsetHeight > raw.scrollHeight) {
-                    scope.$apply(attrs.adminNgLoadingScroller);
+                    scope.$apply(attrs.loadingScroller);
                 }
             });
         }
diff --git a/modules/admin-ui/src/main/webapp/scripts/shared/resources/userRolesResource.js b/modules/admin-ui/src/main/webapp/scripts/shared/resources/userRolesResource.js
index fe920a0ed..8864adfe6 100644
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/userRolesResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/userRolesResource.js
@@ -22,7 +22,7 @@
 
 angular.module('adminNg.resources')
 .factory('UserRolesResource', ['$resource', function ($resource) {
-    return $resource('/admin-ng/resources/ROLES.json', {}, {
+    return $resource('/admin-ng/resources/ROLES.json?limit=20', {}, {
         query: {
           method: 'GET',
           isArray: true,
```

In the Add user wizard and the user detail modal, I get the following result:

screen shot 2018-08-14 at 13 38 39

When I scroll down, the Admin UI issues the following request:

`http://localhost:8080/admin-ng/resources/ROLES.json?limit=20&filter=role_target:USER`

... which returns the first 20 roles you see in the screenshot.  There is a larger number of roles missing (think of all the ROLE_UI_*).

What am I testing wrong?

gregorydlogan:
Bah, good catch @staubesv, I forgot to commit a couple of things.  This should be working now.

staubesv (approved):

PR #382, opened 2018-08-14
MH-12761: Fixed event to listen to "plugin.events.captionsFound".

TurRil:


lkiesow:
@TurRil, can you give a short comment about the problem, its effects and how to reproduce it?

TurRil:
Problem:
An event might have a caption attachment but on loading in engage theodul player the caption is loaded correctly but not enabled/shown because the caption button is not active. It is set to a disabled state.

The fix was to just enable the the engage plugin to listen to the correct event and then set the caption buttons state correctly.

PR #387, opened 2018-08-17
MH-12810 External API 1.1.0 - Add filters for new fields

staubesv:
This PR adds more filters to GET /api/events and GET /api/series for version 1.1.0 of the External API.

The changes are backwards-compatible and therefore included in the not yet released minor version 1.1.0.

Note that I did by intention not add the filter "creator" to GET /api/events as I found that it is not working and I unfortunately don't have time to dig into that.

Since the new date range filters require the character ":" in the filter values, I've stumbled over  MH-13038 (":" not working in filter values) which I have fixed for version 1.1.0. Fixing that for older Opencast version could be done by introducing version 1.0.1 (to avoid different "versions" of version 1.0.0) but it would require some workaround for the not correctly working versioning in < 1.1.0. As I never heard of anybody running into this so far, there seems to be a chance the problem solves itself by time (as 1.1.0 is backwards compatible, people could move to 1.1.0 instead of 1.0.1). In case that changes, there is still the option to add 1.0.1.
The same problem exists with "," characters that cannot be used in values currently, but that is another issue.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -97,6 +97,7 @@ import org.opencastproject.util.RestUtil;
 import org.opencastproject.util.RestUtil.R;
 import org.opencastproject.util.UrlSupport;
 import org.opencastproject.util.data.Option;
+import org.opencastproject.util.data.Tuple;
 import org.opencastproject.util.doc.rest.RestParameter;
 import org.opencastproject.util.doc.rest.RestParameter.Type;
 import org.opencastproject.util.doc.rest.RestQuery;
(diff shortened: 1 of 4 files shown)


mtneug:
I don't want to put the discussion too off track. Take this just as a side note.

We run into this problem some time ago (see <https://groups.google.com/a/opencast.org/forum/#!topic/dev/CW1sSZigS0c>), but I had no time to look at it further. Splitting on the first colon is a good fix in my opinion. As for commas: the linked thread contains some suggestions, however, maybe it would be better to redesign and use multiple query arguments for multiple filters, e.g. `filter[0]=name:my name&filter[1]=title:a, b and c`.

staubesv:
@mtneug Sorry, wasn't aware of this. We could solve this in 1.1.0 by introducing a second query parameter "filters" in addition to "filter". 
Changing the semantics of "filter" would be a major change that requires the introduction of 2.0.0 as this breaks existing clients.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -617,20 +618,80 @@ public Response getEvents(@HeaderParam("Accept") String acceptHeader, @QueryPara
...
+        } else if ("textFilter".equals(name)) {
           query.withText("*" + value + "*");
-        if ("series".equals(name))
+        } else if ("series".equals(name)) {
           query.withSeriesId(value);
-        if ("subject".equals(name))
+        } else if ("subject".equals(name)) {
           query.withSubject(value);
+        } else if (!requestedVersion.isSmallerThan(ApiVersion.VERSION_1_1_0)) {
+          // add filters only available with Version 1.1.0 or higher
```
CGreweling: add = additional?

staubesv: add = add filters to query

Anyway, I've changed this to "additional"

Fixed.

PR #388, opened 2018-08-17
MH-13040, Make options fit “Actions” drop-down

lkiesow:
This patch adjusts the width of the “Actions” menu to ensure the new
option to bulk edit scheduled events fits properly. This is obviously
not guaranteed for arbitrary translation which may still have a line
break but at least for the default language.

screenshot from 2018-08-18 01-18-37

### DIFF
--- a/modules/admin-ui/src/main/webapp/styles/components/_dropdowns.scss
+++ b/modules/admin-ui/src/main/webapp/styles/components/_dropdowns.scss
@@ -21,7 +21,7 @@
 
 
 .drop-down-container {
-    width: 140px;
+    width: 160px;
     height: 40px;
     line-height: 40px;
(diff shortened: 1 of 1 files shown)


staubesv:
Hi @lkiesow 

Please note that this won't be width enough for languages like German...
Note sure, though, whether it is desirable to make it that width.

I don't have a strong opinion on this: Shall I accept the PR as is or do you want to change something more?

lkiesow:
I intentionally made this fit the default languages since all other can change all the time. If we wanted to ensure they fit in a single line as well, we would need to make the width dynamical with a set maximum width. But that's outside of the scope of this pull request. Hence, please go ahead and merge this.

staubesv (approved):

PR #390, opened 2018-08-20
MH-13044, Organization server configuration documentation

lkiesow:
In the previous configuration it was not exactly clear how this should
be configured in a non-multi-tenant system and neither where the values
it should get if multi-tenancy was used.

### DIFF
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -32,16 +32,25 @@ id=mh_default_org
 #
 name=Opencast Project
 
-# Server name that identifies the organization (tenant) on this Opencast installation.
+# Server names that belong to this organization (tenant)
 #
-# Use the public url that is assigned to this tenant. In out-of-the-box installations
-# with only one tenant, this value corresponds to the server.url in etc/config.properties.
-# Make sure to adjust these two keys accordingly:
+# Use the public domain names which are assigned to this tenant. Multiple values can be separated by comma. The final
+# set of domain names identifying this organization within Opencast will be the union of all domains defined on all
+# nodes.
 #
-#   - prop.org.opencastproject.admin.ui.url
-#   - prop.org.opencastproject.engage.ui.url
+# The usage of this is to map a request to a specific organization. For example, if admin-a.opencast.org is configured
+# as part of this organizations server list and an HTTP request uses this domain name, Opencast knows that the requests
+# belongs to this organization.
 #
-# Value: <hostname without schema and port>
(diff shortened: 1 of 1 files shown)


lkiesow:
@JulianKniephoff, can you check if this is correct?

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -32,16 +32,22 @@ id=mh_default_org
 #
 name=Opencast Project
 
-# Server name that identifies the organization (tenant) on this Opencast installation.
+# Server names that identifies the organization (tenant)
```
JulianKniephoff: Typo: "[...] names that **identify** [...]"

JulianKniephoff: I'm also not sure whether "identify" is really the best choice of word here. The organization can have more hosts than are listed here, as you correctly state below, namely those potentially listed by other nodes.

Maybe something like:

> Server names that **belong** to the organization (tenant)

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -32,16 +32,22 @@ id=mh_default_org
...
-# Use the public url that is assigned to this tenant. In out-of-the-box installations
-# with only one tenant, this value corresponds to the server.url in etc/config.properties.
-# Make sure to adjust these two keys accordingly:
+# Use the public domain names which are assigned to this tenant. Multiple values can be separated by comma. The final
+# set of domain names identifying this organization within Opencast will be the union of all domains defined on all
+# nodes.
 #
-#   - prop.org.opencastproject.admin.ui.url
-#   - prop.org.opencastproject.engage.ui.url
+# The usage of this is to map a request to a specific organization. For example:
```
JulianKniephoff: I'm not sure this example is entirely clear to me. Maybe expand it a bit? Trying it myself, though, makes this comment super long which is also not ideal so maybe it's fine. :thinking:

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -32,16 +32,22 @@ id=mh_default_org
...
+# - admin-a.opencast.org, presentation-a.opencast.org  →  org_a (defined in configuration file a)
+# - admin-b.opencast.org, presentation-b.opencast.org  →  org_b (defined in configuration file b)
+# Even though admin-a and admin-b are the same node.
```
JulianKniephoff: Maybe "might be" instead of "are".

JulianKniephoff (changes requested):
Regarding the content this looks right. I added a few nitpicks about the wording, though. :wink:

JulianKniephoff (approved):

PR #392, opened 2018-08-22
MH-12688: Add translations for comment filter values

jaysuz:
https://opencast.jira.com/browse/MH-12688

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-de_DE.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-de_DE.json
@@ -769,6 +769,8 @@
                "CAPTION": "Kommentare",
                "PLACEHOLDER": "Kommentieren…",
                "SELECTPLACEHOLDER": "Gründe wählen",
+               "NONE": "Keine",
+               "OPEN": "Offen",
                "RESOLVED": "Gelöst",
                "REPLY": "Antworten",
                "REPLY_TO": "Antworten",
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-de_DE.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-de_DE.json
@@ -769,8 +769,6 @@
                "CAPTION": "Kommentare",
                "PLACEHOLDER": "Kommentieren…",
                "SELECTPLACEHOLDER": "Gründe wählen",
-               "NONE": "Keine",
-               "OPEN": "Offen",
                "RESOLVED": "Gelöst",
                "REPLY": "Antworten",
                "REPLY_TO": "Antworten",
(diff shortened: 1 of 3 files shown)


lkiesow:
Nothing you have to change but please note that your German translation will automatically be dropped with the next Crowdin sync. Hence, it's usually preferable to not include translations in pull requests but only provide the source keys (`en_US`). It's also less work for you since you do not have to create the translation files ;-)

staubesv:
@lkiesow @jaysuz does know about this. The reason to include one translation (that will be later overwritten by the Crowdin translations) is to make it easy for a reviewer to test the change.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-de_DE.json
```diff
@@ -769,6 +769,8 @@
                "CAPTION": "Kommentare",
                "PLACEHOLDER": "Kommentieren…",
                "SELECTPLACEHOLDER": "Gründe wählen",
+               "NONE": "Keine",
+               "OPEN": "Offen",
```
KatrinIhler: I feel like these are actually unnecessary since they are never used in the event details, unless I'm missing something?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -785,6 +785,8 @@
            "CAPTION":     "Comments",
            "PLACEHOLDER": "Comment…",
            "SELECTPLACEHOLDER": "Select reasons",
+           "NONE": "None",
+           "OPEN": "Open",
```
KatrinIhler: See above.

modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/EventsListProvider.java
```diff
@@ -117,15 +117,15 @@ public void setIndex(EventIndex index) {
       }
     } else if (COMMENTS.equals(listName)) {
       for (Comments comments : Comments.values())
-        list.put(comments.toString(), comments.toString());
+        list.put(comments.toString(), "EVENTS.EVENTS.DETAILS.COMMENTS." + comments.toString());
```
KatrinIhler: Wouldn't it make more sense to use the FILTERS.EVENTS.COMMENTS translation keys here, since that's what this list provider is providing for?

KatrinIhler (changes requested):
Does what it's supposed to do, but I have a few questions:

modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/EventsListProvider.java
jaysuz: Wow, I must have lost orientation in the language file. I'll address this shortly.

jaysuz:
@KatrinIhler I forgot to ping you, but I have addressed the issues you had pointed out. Feel free to check if everything is fine now :)

KatrinIhler (approved):
Looks good!

PR #393, opened 2018-08-22
MH-13047, Document using Nginx for HTTPS

lkiesow:
Our current documentation is not very helpful when it comes to
explaining the use of Nginx (or any other reverse proxy) with Opencast to
provide a proper HTTPS set-up despite this being the de-facto standard
for most Opencast installations.

This patch adds that documentation, while improving the general
structure of the HTTPS documentation and making it more important in the
general table of contents.

Note that the Nginx example included here is a minimal set-up, not
focused on security. A few general security hints will probably follow.


docs/guides/admin/docs/configuration/https/migration.md
```diff
@@ -0,0 +1,45 @@
+Migrating old content to HTTPS
```
lkiesow, line 1: Note that this is an old guide. I just moved this to a separate documentation page.

docs/guides/admin/docs/configuration/https/opencast.only.md
```diff
@@ -1,24 +1,5 @@
...
-
-Your site visitor's privacy. While ISPs and public HotSpot operators and those
-who sniff unencrypted WiFi traffic might be able to see your visitors
-connecting to your site, they can't see, which video is watched when HTTPS is
-enabled.
-
-
-Enable HTTPS directly in Opencast – without proxy
--------------------------------------------------
+Enable HTTPS directly in Opencast
```
lkiesow, line 1: The guide for enabling HTTPS inside of Opencast now got its own separate documentation page.

docs/guides/admin/docs/configuration/https/nginx.md
```diff
@@ -0,0 +1,123 @@
...
+
+    # HTTP set-up
+    server {
+        listen 80;
+        listen [::]:80;
+        server_name example.opencast.org;
+
+        # Enforce HTTPS by redirecting requests
+        location / {
+            return 301 https://stable.opencast.org$request_uri;
```
gregorydlogan: IMO this should be `example.opencast.org` rather than `stable`

docs/guides/admin/docs/configuration/https/nginx.md
```diff
@@ -0,0 +1,123 @@
...
+    # HTTPS set-up
+    server {
+        listen      443 ssl http2;
+        listen [::]:443 ssl http2;
+        server_name example.opencast.org;
+
+        # Path to the TLS certificate and private key. In almost all cases, you
+        # need to provide intermediate certificates as well to ensure browsers
+        get the whole certificate chain.
+        ssl_certificate_key /var/lib/acme/private/stable.opencast.org.key;
```
gregorydlogan: We should have links to acme.sh, or certbot, or something earlier in the docs.  A native user who just wants that magic lock but doesn't understand what else you need to do might be confused where these files come from otherwise.

These should also be `example.opencast.org.*` rather than `stable` :)

gregorydlogan (changes requested):
Looks good, minor nitpicks

docs/guides/admin/docs/configuration/https/nginx.md
lkiesow: It should → fixed

docs/guides/admin/docs/configuration/https/nginx.md
lkiesow: Changed this to be a more generic path and mentioned Let's Encrypt in the recommendations section of the overview page.

PR #395, opened 2018-08-23
MH-13058 Remove unused font libraries

staubesv:
Note that font-awesome is included in other places.


lkiesow:
No problem without those files → merged

PR #150, opened 2018-03-20
MH-12091: Implement per-tenant digest user for capture agents

pmiddend:
This PR adds a new in-memory per-tenant user and a new role `ROLE_CAPTURE_AGENT` for capture agents. It builds upon work by @smarquard and @lkiesow, see the [bitbucket PR](https://bitbucket.org/opencast-community/opencast/pull-requests/1542).

The new user is able to digest-auth in order to ingest recordings and thus has to be an in-memory user with clear-text password. Restricting the user's rights turned out to be rather tricky because of a "somewhat circular" dependency between `AssetManagerWithMessaging` and `AssetManagerWithSecurity`.

I tried to preserve the older commits in order to see who introduced which feature.

This work is sponsored by SWITCH.

PR #396, opened 2018-08-23
MH-13055: Stop making events with no ACL public on ingest

JulianKniephoff:
See also the [JIRA issue](https://opencast.jira.com/browse/MH-13055).

This PR stops giving events a default, "everybody can read" ACL when it is ingested without one. This happens for example when you use the [ingest scripts](https://github.com/opencast/helper-scripts/tree/master/ingest), which I used to test this, but as evidenced by [this length thread](https://groups.google.com/a/opencast.org/forum/#!topic/anwender/qj94Xqr2Luw) on the German users list it also happens in real life. :wink: It does not happen when going through the admin UI.

#150 (merged in `develop`) introduced the `GLOBAL_CAPTURE_AGENT_ROLE` that was also part of this default ACL. I don't know whether it is actually necessary; tests with the different ingest scripts worked but I do not have multi tenant system here, so maybe it breaks there? (This is just a hunch based on the title of that PR.) Maybe the (co-)authors of that PR can say something about that? @lkiesow, @smarquard, @pmiddend

If it is still needed we might want to incorporate it into the fallback (basically empty) ACL that other parts of the system return when no ACL is found. Either way, putting together a custom ACL in this part of the code, when we already have fallbacks in place elsewhere, is wrong IMO.

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -58,10 +58,8 @@ import org.opencastproject.scheduler.api.SchedulerException;
 import org.opencastproject.scheduler.api.SchedulerService;
 import org.opencastproject.security.api.AccessControlEntry;
 import org.opencastproject.security.api.AccessControlList;
-import org.opencastproject.security.api.AclScope;
 import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.OrganizationDirectoryService;
-import org.opencastproject.security.api.Permissions;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.TrustedHttpClient;
 import org.opencastproject.security.api.UnauthorizedException;
@@ -1134,9 +1132,6 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
       // Merge scheduled mediapackage with ingested
       mp = mergeScheduledMediaPackage(mp);
 
-      // Set public ACL if empty
-      setPublicAclIfEmpty(mp);
-
       ingestStatistics.successful();
       if (workflowDef != null) {
         logger.info("Starting new workflow with ingested mediapackage '{}' using the specified template '{}'",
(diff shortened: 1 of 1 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1277,17 +1271,6 @@ private MediaPackage checkForLegacyMediaPackageId(MediaPackage mp, Map<String, S
     }
   }
 
-  private void setPublicAclIfEmpty(MediaPackage mp) {
-    AccessControlList activeAcl = authorizationService.getActiveAcl(mp).getA();
-    if (activeAcl.getEntries().size() == 0) {
-      String anonymousRole = securityService.getOrganization().getAnonymousRole();
-      activeAcl = new AccessControlList(
-              new AccessControlEntry(anonymousRole, Permissions.Action.READ.toString(), true),
-              new AccessControlEntry(GLOBAL_CAPTURE_AGENT_ROLE, Permissions.Action.WRITE.toString(), true));
```
staubesv: @JulianKniephoff Have you tested this using an restricted access capture agent user? That one does not have ROLE_ADMIN and therefore needs explicit permissions on the event because the workflow runs as that user. My guess is that processing would fail because of no access permission then (write permission is required by WOH snapshot).

Can you please verify?

JulianKniephoff: See, this is what I expected. x) No I haven't, I will try to put together a test later today or tomorrow. Of course if someone beats me to it, that would be appreciated. xP

staubesv: As the restricted access capture agent user is quite new: It will have the role GLOBAL_CAPTURE_AGENT ROLE and can be enabled / configured in https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.organization-mh_default_org.cfg#L71

Just do an ingest using that user and make sure a workflow is triggered (by ingest) that involves a WOH snapshot.

staubesv: (anyway: Absolutely +1 for removing the anonymous user's role here)

JulianKniephoff: The interesting thing, too, is that if you provide an ACL, the global capture agent role won't get access either, so things were probably already partially broken before, right? :thinking:

staubesv: Good catch!!! I've filed https://opencast.jira.com/browse/MH-13059 for this.

JulianKniephoff: Okay, so after an internal discussion with @lkiesow we decided to rebase this on `r/5.x` for now, so that it still gets into 5.1. This makes this discussion irrelevant for now (sorry @staubesv). @lkiesow promised to look into this issue further, though, and potentially forward port it into `develop`.

lkiesow:
@staubesv, we will take a look at 6.x which introduces the CA user though I vaguely remember that AssetManagerWithSecurity was modified so maybe we do not need this there anymore.

Anyway, this should now nw ready for a 5.1 security fix. There we should definitely be able to drop this.

staubesv:
Makes sense.

Note that the restricted access user is not specific to multi-tenant setups, i.e. if it works in a single-tenant setup, it will also work on a multi-tenant setup.

PR #398, opened 2018-08-27
MH-13000: Group “Edit scheduled” events by weekday

pmiddend:
As described in the associated JIRA ticket, this PR reworks the “Edit
scheduled events” modal to group all selected events by day of the
week, as can be seen here:

screenshot_4

It also fixes a minor annoyance by making the column widths in the
last wizard step (the summary) equal width.


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1164,29 +1164,32 @@ public Response getBulkConflicts(@FormParam("update") final String updateJson) t
     final BulkUpdateUtil.BulkUpdateInstructions instructions;
     try {
       instructions = new BulkUpdateUtil.BulkUpdateInstructions(updateJson);
-    } catch (IllegalArgumentException e) {
+    } catch (final IllegalArgumentException e) {
```
KatrinIhler: Here the final feels like overkill to me, even more so as we're not using e at all... It's also not consistent with pretty much every other catch-block in Opencast.

KatrinIhler (changes requested):
The grouping by weekday seems a bit arbitrary for me? So if I edit two events that are scheduled for different days, I can edit them individually, but two events scheduled for the same day can only be edited together (unless I edit them one at a time obviously). Is there a specific reason for this, especially since one wants to avoid overlap in the scheduling of events for the same day (same start date & room = conflict)? Additionally, once a conflict has been caused by changing the scheduling settings of events for the same day it also can't be resolved within this tab since the settings can't be changed individually for each event... :/

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
staubesv: @KatrinIhler We expect lectures to take place on regular intervals. For example, on each Monday at 10am to 12am during the semester. You also see this in the Add Event wizard that allows you to create events on rrules (in particular, by setting a week day).

It is true that this might seem arbitrary if you select an arbitrary set of events. But in real world use cases, selecting arbitrary events to change their scheduling won't be of any use anyway: Unrelated events will not have anything in common you would want to change at once (neither room, week day, start time or any other value you can edit with this function).

Example:
The lecture Mathematics was scheduled to take place starting in the first week of the semester until end of semester (specified in date) on Mondays, 10am-11am, Room A and Thursdays, 1pm-3pm, Room B.

The reason why people would want to edit scheduled events (as far as I understood) are changes like: 

Later, the schedule has changed: The lecture Mathematics takes place in Room X on Mondays, and Room B at 2pm-4pm on Fridays.

That kind of changes can now easily be done because of the grouping by week days.

If not grouped by weekdays, the user would need to select all events of the lecture that take place on Thursdays (that is error prone) to be able to change the week day.

This way, the user can filter for a series and just select all events of the series. This works quite well and is the expected use case for this feature.

KatrinIhler:
@staubesv Okay, I can see the motivation now. But I think that the dialog can (and thus probably will) be used to edit arbitrary events simply because it's faster editing them in one overview instead of one after another, and one can set different values for start- and enddate, room and so on for each event if they're not on the same weekday. 
Because of this, I propose to only group events that not only have the same weekday, but also the same start- and enddate & room (which should always be true for the use case you described). This would mean that one cannot stumble into an unresolvable conflict anymore because events that happen on the same day in the same room but on different times could be edited separately. This also fixes things for a use case in which a course happens twice in one day, e.g. once from 10 to 12 and once from 14-16 - these events should be grouped into two groups (by weekday and time), not into one (by weekday only).

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
pmiddend: @KatrinIhler You're right, it does seem unusual. Do you know if there's a guide on where to use final in Opencast? I think the most consistent code style would be to avoid final altogether (bug-catching and readability issues aside).

KatrinIhler: @pmiddend I don't think we have a guide for that. In general I try to be consistent with the existing code unless it makes no sense at all (though that is sometimes not consistent with itself, then I choose the option I like better...) I would use final to inform the reader that a variable cannot and will not be reassigned, but since e is only visible in this block and no sane person would reassign it I would leave it out here.

pmiddend: Okay, fixed that.

staubesv:
@KatrinIhler 

I don't really agree on the usefulness of editing "unrelated events": While it is true that it would be more efficient if one wanted to edit "unrelated events", I have a hard time to believe that this is a relevant use case. At the end of the day, I would expect the motivation to actually change something to be an incoming service request asking for the change. For what I have seen at our institutions, those requests are related to individual lecture series and not just to arbitrary sets of events.

Agree that the inability to edit multiple events taking place at the same days is indeed suboptimal, but this is currently not a priority for us.

I would be happy if you could accept this change as is in means of that it does improve the situation while admittedly not solve (but also not introduce) the "multiple events on the same day problem".

gregorydlogan:
Issue identified in brief testing:

- Start/end time checks are not done on the client.  I tried setting one of my events to have a duration of -1 hours, and it didn't show any kind of error until *after* the error check had been performed on the admin node
- The title in the edit page is empty initially.  It's unclear what would happen upon a successful update, would it erase the title, or is the content of this input appended?
- The update does not seem to function for me.  I initially scheduled a repeating capture on a completely blank system from 10-11 on Tuesday, Thursday, and Sunday, on a pyCA.  I select all of the events, edit them, and attempt to change from Sunday to Friday.  The error says the changed event (capture 10) is conflicting with event 12, which surprises me since they should be at least 24 hours apart!

Edit:
- I feel like the box displaying the conflicting events needs some kind of further highlight.  It blends into an already lengthy modal, and I missed it initially.  Does extending the highlight to that box make sense?

staubesv:
@gregorydlogan 

- The title field has the same behaviour as all other fields: If no input is shown, the existing titles will not be changed. This behaviour is described in the text. Additionally, the summary provides a full list of changes that will applied.

- In theory, the checks that the Admin UI does in the client could be performed first to avoid the conflict check in the backed. From a usability point, though, all checks should always be performed to always get the full list of conflicts. Otherwise, the user gets conflicts reported checked by either the frond- or backend and after "fixing" the conflicts just runs into the next conflicts reported by the other part

The update function is supposed to work, though ;-)

Will try to reproduce the problem locally.

KatrinIhler:
@pmiddend Merge conflicts!

pmiddend:
@KatrinIhler Thanks for the reminder. Fixed.

gregorydlogan:
@staubesv, did you ever get this reproduced?

staubesv:
@pmiddend Please note that this PR has merge conflicts.

@gregorydlogan Will try to reproduce one the merge conflicts have been solved.

pmiddend:
I've fixed the conflicts and tested this briefly.

staubesv:
> Start/end time checks are not done on the client. I tried setting one of my events to have a duration of -1 hours, and it didn't show any kind of error until after the error check had been performed on the admin node

@gregorydlogan I could reproduce this locally, thanks for reporting this bug!

Reproduction:
1. Select at least one scheduled event
2. Open 'Edit scheduled events'
3. Set the "End time" to a time before the "Start time"
4. Press 'Next' 
5. Look at Summary Page ;-)

So the problem is that the user can select invalid pairs <start time, end time>, for example a start time that is later than the end time:

screen shot 2018-10-29 at 14 01 55

... and then just press next that leads to the summary even displaying the incorrect values:

screen shot 2018-10-29 at 14 02 02

Expected behavior:
If the pair <start time, end time> does not make any sense (e.g. start time later than end time or start time equals the end time), the "Next" button should be inactive and some kind of error notification should tell the user that the value is not correct.

@pmiddend Could you please fix this?

pmiddend:
@staubesv @gregorydlogan Alright, I fixed it in exactly the way Sven described and tested it.

gregorydlogan:
New edge cases :)
- Edit one event to overlap with another event -> not detected until update is attempted
- Events are allowed to start and end in the same minute, where normally a buffer zone of 1 minute is enforced
- Editing the duration of events carries over to the default duration for scheduling a new event

pmiddend:
Yay, edge cases! I'm on it.

@gregorydlogan 
  - I fixed the "one minute buffer" behavior now, by moving the "check for overlap" function from `SchedulerServiceImpl` to `Util` inside the scheduler package.
  - What do you mean by "duration of events"? The "Edit scheduled" dialog explicitly disallows setting the duration of events - just start and end.
  - What do you mean "until update is attempted"? If you're referring to the fact that you can edit the events and not _immediately_ get a conflict popping up - that's intentional. You have to press the "Next" button to trigger the check.

gregorydlogan:
- You can set the duration of events in the *new* event dialog, which is what I was talking about.  It's carrying state between dialog modals for some reason.
- Re: until update attempted, ok.  This is suboptimal, but I can deal with it :)

pmiddend:
@gregorydlogan Okay, I think I know what you mean. I tried changing the duration of an already scheduled event from, say, 1h to 2h and I assume that if I now *add* a new event, it'll be 2h long by default (in the new event modal)? Doesn't happen for me, though. Are there other circumstances in your tests that might not be present in mine?

gregorydlogan:
Hum, interesting, I can't seem to reproduce that anymore.  Let's disregard that one :)

pmiddend:
@gregorydlogan Then I suppose this can be merged now? :)

gregorydlogan:
Yep

PR #401, opened 2018-08-29
MH-13062, Added credentials for the Ingest Service.

CGreweling:
Add Track and Add Attachment can download itself content to Workspace direct form a given url.
This PR enhances the ingest Service with credentials.

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -194,20 +194,29 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
   /** The key to look for in the service configuration file to override the {@link DEFAULT_INGEST_ZIP_JOB_LOAD} */
   public static final String ZIP_JOB_LOAD_KEY = "job.load.ingest.zip";
 
+  /** The source to download from  */
+  public static final String DOWNLOAD_SOURCE = "org.opencastproject.download.source";
+
+  /** The user for download from external sources */
+  public static final String DOWNLOAD_USER = "org.opencastproject.download.user";
+
+  /** The password for download from external sources */
+  public static final String DOWNLOAD_PASSWORD = "org.opencastproject.download.password";
+
   /** The approximate load placed on the system by ingesting a file */
   private float ingestFileJobLoad = DEFAULT_INGEST_FILE_JOB_LOAD;
 
   /** The approximate load placed on the system by ingesting a zip file */
   private float ingestZipJobLoad = DEFAULT_INGEST_ZIP_JOB_LOAD;
 
-  /** The User for Download from external sources */
-  private static String downloadUser = "opencast_system_account";
+  /** The user for download from external sources */
+  private static String downloadUser = DOWNLOAD_USER;
 
-  /** The Password for Download from external sources */
-  private static String downloadPassword = "CHANGE_ME";
+  /** The password for download from external sources */
+  private static String downloadPassword = DOWNLOAD_PASSWORD;
 
   /** The external source dns name */
-  private static String downloadSource = "localhost";
+  private static String downloadSource = DOWNLOAD_SOURCE;
 
   /** The JMX business object for ingest statistics */
   private IngestStatistics ingestStatistics = new IngestStatistics();
@@ -308,9 +317,9 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
       return;
     }
 
-    downloadPassword = StringUtils.trimToNull((String) properties.get("download.password"));
-    downloadUser = StringUtils.trimToNull(((String) properties.get("download.user")));
-    downloadSource = StringUtils.trimToNull(((String) properties.get("download.source")));
+    downloadPassword = StringUtils.trimToEmpty((String)properties.get(DOWNLOAD_PASSWORD));
+    downloadUser = StringUtils.trimToEmpty(((String) properties.get(DOWNLOAD_USER)));
+    downloadSource = StringUtils.trimToEmpty(((String) properties.get(DOWNLOAD_SOURCE)));
 
     ingestFileJobLoad = LoadUtil.getConfiguredLoadValue(properties, FILE_JOB_LOAD_KEY, DEFAULT_INGEST_FILE_JOB_LOAD,
             serviceRegistry);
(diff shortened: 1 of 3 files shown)


lkiesow:
This patch will make Opencast send the configured authentication credentials to arbitrary servers when requested (the request URL just needs to contain the server token). Given that the server then controls part of the authentication if HTTP Digest is used, the password is afaik only protected via md5 has which is not very strong. This essentially means that you expose the configured credentials to everyone who is able to use this endpoint which is problematic. I only played this through in my head though, so feel free to contradict me if you think I made a mistake.

Another question is why we store the credentials at all and do not leave this stateless and just pass the credentials to ingest as part of the URL?

CGreweling:
@lkiesow my be i am not familiar enough with this whole digest auth thing. 
But i need the diest-auth credentials of the admin to use this endpoint. Why would i be interested in getting em then? And then you will get the credentials for the domain your configured the user/password.
So where is the problem?
I was told, that i might be possible, to ingest a whole mediapackage, without adding tracks separately. But haven't tried that so far.

KatrinIhler:
@CGreweling @lkiesow What's the status on this? Is this on hold?

CGreweling:
This Pr is waiting for a Reviewer. :)

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1481,13 +1495,30 @@ public void discardMediaPackage(MediaPackage mp) throws IOException {
...
   protected URI addContentToRepo(MediaPackage mp, String elementId, URI uri) throws IOException {
     InputStream in = null;
     HttpResponse response = null;
+    TrustedHttpClient httpClientStandAlone = httpClient;
     try {
       if (uri.toString().startsWith("http")) {
         HttpGet get = new HttpGet(uri);
-        response = httpClient.execute(get);
+
+        if (uri.getHost().contains(this.downloadSource)) {
```
lkiesow: This check means that if credentials are configured for `develop.opencast.org` and an attacker uses a link like `develop.opencast.org.lkiesow.io`, Opencast will send the credentials to that server. Instead `equals()` or a regular expression (if you want to configure multiple servers) should be used.

lkiesow (dismissed):

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
CGreweling: @lkiesow regexp or equals would not do more security. It's still left to the user, for more security the user should wirte source=https://develop.opencast.org/ ....
But i like the idea of adding multiple source with regexp, so done with regexp -> matches

KatrinIhler:
@lkiesow @CGreweling Are there still security concerns for this PR or was this resolved?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1481,13 +1495,30 @@ public void discardMediaPackage(MediaPackage mp) throws IOException {
...
   protected URI addContentToRepo(MediaPackage mp, String elementId, URI uri) throws IOException {
     InputStream in = null;
     HttpResponse response = null;
+    TrustedHttpClient httpClientStandAlone = httpClient;
     try {
       if (uri.toString().startsWith("http")) {
         HttpGet get = new HttpGet(uri);
-        response = httpClient.execute(get);
+
+        if (uri.getHost().matches(this.downloadSource)) {
```
gregorydlogan, line 1526: We `trimToNull` downloadSource, and the other configurables above.  Does it make sense to throw some kind of warning or error here, rather than silently failing?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1481,13 +1495,30 @@ public void discardMediaPackage(MediaPackage mp) throws IOException {
...
 
+  /**
+   * Creates a StandAloneTrustedHttpClientImpl
+   *
+   * @param user
+   * @param password
+   * @return
+   */
+  protected TrustedHttpClient createStandaloneHttpClient(String user, String password) {
+    return new StandAloneTrustedHttpClientImpl(this.downloadUser, this.downloadPassword, none(), none(), none());
```
gregorydlogan, line 1515: If the `user` parameter is null here there will be an IllegalArgumentException thrown.  I'm not sure what would happen with a null password, but I'd guess nothing good :)

gregorydlogan (changes requested):
I would suggest using `trimToEmpty` to work around null related issues.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
CGreweling: Yep right. But im am not sure where to put this in our opencast code world. From my perspective this null should be handled during variable allocation at the beginning? Or check all credentials user,pw,source right here and throw an error. What do you think?

gregorydlogan: Personally I would do the sanity check when they're first parsed in `updated()`, and throw an (IIRC) `ConfigurationException` if they're invalid.  This fails the startup of the service, which prevents having to handle the case later :)

gregorydlogan:
@lkiesow are you concerns here resolved?  I think this looks ok, although I definitely see your point re: statelessness.

gregorydlogan (approved):

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -193,12 +194,30 @@
...
+  public static final String DOWNLOAD_PASSWORD = "org.opencastproject.download.password";
+
   /** The approximate load placed on the system by ingesting a file */
   private float ingestFileJobLoad = DEFAULT_INGEST_FILE_JOB_LOAD;
 
   /** The approximate load placed on the system by ingesting a zip file */
   private float ingestZipJobLoad = DEFAULT_INGEST_ZIP_JOB_LOAD;
 
+  /** The user for download from external sources */
+  private static String downloadUser = DOWNLOAD_USER;
```
staubesv, line 213: @CGreweling Looks like you wanted to set a default value for downloadUser, downloadPassword and downloadSource but accidentally set the names of the respective configuration properties ;-)

PR #404, opened 2018-08-30
MH-13067, Configuration panel does not work for default workflow

lkiesow:
If a default workflow is configured in `custopm.properties` and hence
automatically selected in the “New Event” wizard, all user input from
the configuration panel is discarded and the default settings are used
instead.

This can be easily tested by configuring the default workflow:

    org.opencastproject.workflow.default.definition=schedule-and-upload

…and selecting the first two options in the configuration panel of the
“New Event” wizard. The summary will already show these values to be set
to false (despite of them being selected earlier).

This patch ensures the configuration panel is looked up when updating
the configuration to ensure the used configuration is always up-to-date.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/processing.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/processing.js
@@ -34,8 +34,7 @@ angular.module('adminNg.services')
             isWorkflowSet = function () {
                 return angular.isDefined(me.ud.workflow) && angular.isDefined(me.ud.workflow.id);
             },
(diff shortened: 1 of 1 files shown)


staubesv (changes requested):
Hi @lkiesow 

I think this PR makes line 38 and 90 obsolete. Can you please check and remove those lines?

staubesv:
Please also note that the configuration panel has another bug: If you select "onhold", the workflow will still publish. But that is another issue...

lkiesow:
@staubesv, updated the pull request. had to slightly modify the line I changed before as well.

The workflow definition (onhold issue) should be completely unrelated to this (workflow issue vs admin interface issue).

staubesv (approved):

PR #410, opened 2018-09-03
MH-13074, Clean up asset manager REST endpoints

lkiesow:
The asset manager REST endpoints contains a few pieces of weird or
duplicated code:

- Remove Entwine's functional library for exception handling
- Do not duplicate identical methods
- Remove broken “If-None-Match” header parameter

PR #413, opened 2018-09-05
MH-13079, Introduce REST Interface for AssetManager Properties

lkiesow:
Asset manager properties allow for an event based key-value store under
a given namespace. This is currently used, for example, by the workflow
service which stores workflow variables passed to workflows as user
input to ensure they are available for a new workflow as well.

One problem of these properties is that they are never exposed, making
them extremely hard to access for things like a migration from
file-based workflow variable storage to asset manager based storage.

This patch aims to provide a basic API to read and write properties. It
is especially focused on workflow variables.

- *This is based on pull request #410*
- *Work sponsored by SWITCH*

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
@@ -20,12 +20,15 @@
  */
 package org.opencastproject.assetmanager.impl.endpoint;
 
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_CREATED;
 import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
 import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
 import static org.opencastproject.assetmanager.api.AssetManager.DEFAULT_OWNER;
+import static org.opencastproject.systems.OpencastConstants.WORKFLOW_PROPERTIES_NAMESPACE;
 import static org.opencastproject.util.MimeTypeUtil.Fns.suffix;
 import static org.opencastproject.util.RestUtil.R.badRequest;
 import static org.opencastproject.util.RestUtil.R.forbidden;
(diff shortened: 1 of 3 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -286,6 +304,156 @@ public Response getAsset(@PathParam("mediaPackageID") final String mediaPackageI
...
+      return ok(gson.toJson(properties));
+    } catch (Exception e) {
+      return handleException(e);
+    }
+  }
+
+
+  @POST
+  @Path("{mediaPackageID}/workflowProperties")
+  @RestQuery(name = "detWorkflowProperties",
```
staubesv: Typo: **d**etWorkflowProperties

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -286,6 +304,156 @@ public Response getAsset(@PathParam("mediaPackageID") final String mediaPackageI
...
+    try {
+      properties = gson.fromJson(propertiesJSON, stringMapType);
+    } catch (Exception e) {
+      return badRequest();
+    }
+    for (final Map.Entry<String, String> entry : properties.entrySet()) {
+      final PropertyId propertyId = PropertyId.mk(mediaPackageId, WORKFLOW_PROPERTIES_NAMESPACE, entry.getKey());
+      final Property property = Property.mk(propertyId, Value.mk(entry.getValue()));
+      if (!getAssetManager().setProperty(property)) {
+        return notFound();
```
staubesv, line 451: Why "not found"? Shouldn't the property be created here?

lkiesow: No. `setProperty()` will return `false` if the event identified by the media package id does not exist in the asset manager. Well… that's the idea (and documentation) at least. There are a bunch of issues in the asset manager implementation related to this… but at least it's very likely the case ;-P

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
lkiesow: fixed

gregorydlogan:
Can you give an example of what the input JSON object looks like?  I'm taking the output of the GET method, and feeding it, both formatted and unformatted, to the POST and getting a 400...

lkiesow:
Sure. The method should accept a JSON abject containing simple key-value pairs. E.g.:

```JSON
{"a":"b"}
```

or

```JSON
{"a":"b", "c":"d"}
```

gregorydlogan:
Ah, I was misinterpreting what was expected.  This works!

PR #166, opened 2018-03-29
MH-12799 Add ability to configure which LTI parameter/value to use instead of the default LTI context id

paulgration:
_In planning a move to Canvas as our VLE and to use LTI as the integration method, however we have discovered that the Context IDs that Canvas passes as one of the LTI parameters isn't meaningful (random 40 char hex string) or searchable within Canvas so it's not possible to know them in order to create matching Series IDs in Opencast.
Default LTI integration between a VLE and Opencast appears to match the passed LTI Context ID with a Series ID to create user roles/grant access to view the media packages, so we would need to be able to select a different/optional (override) parameter to use as the Context ID instead of the default value.
For us, Canvas is the driver for this but it may be useful for other VLEs as well._

This change adds the ability for a VLE to specify a `custom_context_id_param` LTI parameter which Opencast can use/treat as the context ID to match with a series ID.
The value of the parameter referred to in the `custom_context_id_param` will be used, e.g.

```
context_label=Foo
custom_context_id_param=context_label
```

Opencast would treat the context ID from the VLE as `Foo` and create the appropriate LTI roles based on this.
Arbitrary values can be used by specifying another custom parameter in the VLE, e.g.

```
custom_foo=Bar
custom_context_id_param=custom_foo
```

PR #414, opened 2018-09-06
MH-13082: Removed LTI security vulnerability again

t-schroeder:
After https://github.com/opencast/opencast/commit/ef7bc8e94936b426bec5cb52f846d81c8c903657 the 3 parameter constructor silently discarded the second and third parameter. I found this behavior confusing so I "fixed" it by reverting that constructor back to its old behavior with #378 not realizing that that brought back the security vulnerability that @mtneug originally fixed with https://github.com/opencast/opencast/commit/ef7bc8e94936b426bec5cb52f846d81c8c903657.

To resolve this problem I simply removed the 3 parameter constructor. That way if you e.g. upgrade from Opencast 4.3 and forget to add the fourth parameter in your config file the `highlyTrustedKeys` you pass to your constructor won't just be silently discarded. Instead the constructor call itself will fail which (I hope) means people will see more quickly that something stopped working and what stopped working.

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
@@ -108,6 +108,18 @@ public class LtiLaunchAuthenticationHandler
   }
 
   /**
+   * Constructor for a LTI authentication handler that includes a list of highly trusted keys
+   *
+   * @param userDetailsService
+   * @param highlyTrustedkeys
+   */
+  public LtiLaunchAuthenticationHandler(UserDetailsService userDetailsService, SecurityService securityService,
+          List<String> highlyTrustedkeys) {
+    this(userDetailsService, securityService, highlyTrustedkeys, "^(admin|opencast_system_account)$");
+    logger.warn("No untrusted users pattern configured - using default \"" + this.untrustedUsersPattern + "\"");
+  }
+
+  /**
    * Full constructor for a LTI authentication handler that includes a list of highly trusted keys with exceptions.
    *
    * @param userDetailsService
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
@@ -104,18 +104,7 @@ public class LtiLaunchAuthenticationHandler
    *          the user details service used to map user identifiers to more detailed information
    */
   public LtiLaunchAuthenticationHandler(UserDetailsService userDetailsService) {
-    this(userDetailsService, null, new ArrayList<String>());
-  }
-
-  /**
-   * Constructor for a LTI authentication handler that includes a list of highly trusted keys
-   *
-   * @param userDetailsService
-   * @param highlyTrustedkeys
-   */
-  public LtiLaunchAuthenticationHandler(UserDetailsService userDetailsService, SecurityService securityService,
-          List<String> highlyTrustedkeys) {
-    this(userDetailsService, securityService, highlyTrustedkeys, null);
+    this(userDetailsService, null, new ArrayList<String>(), null);
   }
 
   /**


gregorydlogan:
@mtneug, isn't this basically a revert of what you just fixed?

lkiesow:
@do-you-even-curl, can you please add a proper description to this pull request, explaining what this is about? Also, if it's a security vulnerability, please consider sending such reports and patches to security@opencast.org

mtneug:
I will take the review.

@lkiesow there is a pretty good description in the Jira issue. @do-you-even-curl it would still be good to summarize the problem in the pull request.

So, @gregorydlogan #403 and #378 fixed the three argument constructor such that arguments were really past on. However, not using the four argument constructor leads to a configuration where the `untrustedUsersPattern` is `null` and thus no check is performed for sensitive user names. Now, in the current default configuration this is not a problem, since my original patch (ef7bc8e94936b426bec5cb52f846d81c8c903657) also updated the `mh_default_org.xml` to use the four argument constructor. But when you update from an older system and forget to update `mh_default_org.xml`, the old constructor is used and LTI is therefore not secured. This PR argues that it is better to remove the older constructor and let the startup of Opencast fail than to have systems with an unpatched `mh_default_org.xml`.

Can't we then remove the one argument constructor with the same argumentation @do-you-even-curl? As far as I can tell, there are no usages in the Opencast code.

mtneug:
Mhh, checked again: the default is to use the one argument constructor, which should be safe as there are no `highlyTrustedKeys` and thus no user name is trusted.

t-schroeder:
@lkiesow @mtneug I described the problem at https://opencast.jira.com/browse/MH-13082 but I added a summary here.

@lkiesow Sorry, I forgot there's a separate workflow for security related issues. As I understand it https://opencast.jira.com/browse/MH-13022, i.e. #378, (which re-introduced the security vulnerability) has not yet made it into any releases. If this fix is included in the releases 3.7, 4.5 and 5.2 there shouldn't be a problem, right? Do you think it's possible to get this fix into the next releases?

smarquard:
A better way of having a secure default is to put the default in the code, so if a constructor is used which does not include this parameter, there is a default deny list. This is a much more common pattern, and doesn't break startup for people who don't update their configuration.

t-schroeder:
@smarquard Good suggestion. That's a lot more user friendly. I added it to this PR.

mtneug:
I agree, this is better than failing, but maybe we should still log a warning. Ultimately we want users to update their configuration and think about special users in their systems.

modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
```diff
@@ -126,7 +115,7 @@ public LtiLaunchAuthenticationHandler(UserDetailsService userDetailsService, Sec
    * @param untrustedUsersPattern
    */
   public LtiLaunchAuthenticationHandler(UserDetailsService userDetailsService, SecurityService securityService,
-          List<String> highlyTrustedkeys, String untrustedUsersPattern) {
+          List<String> highlyTrustedkeys, String untrustedUsersPattern = "^(admin|opencast_system_account)$") {
```
mtneug: This is not possible in Java. You have to update the three argument constructor, which calls this one and pass `"^(admin|opencast_system_account)$"` instead of `null` as the fourth argument.

mtneug (changes requested):

modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
smarquard: Also these names are configurable, so this default regexp should probably be constructed from the property values of org.opencastproject.security.admin.user and org.opencastproject.security.digest.user

t-schroeder: @smarquard I saw that you can get these properties from a `ComponentContext` object via `getProperty` or `OsgiUtil.getComponentContextProperty` but how do I get the `ComponentContext` object? I noticed it is passed to some objects via the `activate` method which is apparently an OSGi callback. But the `LtiLaunchAuthenticationHandler` class doesn't have such a method. Can you tell me how I can add it? Or am I on the wrong track?

t-schroeder:
@mtneug I added the warning log message.

modules/kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
mtneug: Is this possible since the `LtiLaunchAuthenticationHandler` is instantiated by Spring and not OSGi?

Changing the default user name is a good point @smarquard. There probably should be a comment in `etc/custom.properties` that reminds users to also update the regex if they use LTI.  I would further suggest to add a note/section to the LTI documentation describing the purpose of the regex. I will create a separat PR for this.

gregorydlogan:
Ok, is this ready for merge then?  Does it need to go into the older versions as well?  If so can we repoint this PR against the oldest version and then I'll merge it forwards again.

mtneug:
@gregorydlogan I thought about the remarks @staubesv gave during the last developer meeting and I agree that his approach leads to a better solution. As I understood it, he proposed to convert `LtiLaunchAuthenticationHandler` to an OSGi service with the consequences that:

* `mh_default_org.xml` only contains a reference to the `LtiLaunchAuthenticationHandler` object and is no longer used to configure LTI.
* Instead, LTI is configured via a dedicated configuration file for this OSGi service.
* The configuration file contains a blacklist of users who are not allowed to authenticate via LTI.
* Maybe can have also a dedicated configuration key for excluding the admin and system user, so that name changes do not have to be reflected in the blacklist.
* The LTI documentation has to be updated.

The benefit of an OSGi service is, that the configuration can be changed during runtime, e.g. a new user is created and then inserted into the blacklist.

@gregorydlogan is this something you would be comfortable with merging in 3.x?

@do-you-even-curl would you be willing to implement these changes?

smarquard:
I like the idea of having a separate configuration file for LTI, in part because of the long-neglected PR #166 which I'm the reviewer for. It seems inelegant (and error-prone) to have more and more constructor arguments for extra parameters.

gregorydlogan:
@mtneug it's not ideal, but if it resolves the issue in a clean way we can probably do that.  This should probably be brought up with the 4.x and 5.x RMs as well at this point - if I merge it into 3.x it'll end up everywhere else too.  A final point: 3.x is EOL at this point, and after 6.0's release won't be formally maintained anymore.  A better target might be 4.x, which will be in 3.x's position after 6.0 releases.

t-schroeder:
@mtneug Unfortunately I'm not familiar enough with the code nor do I have the time right now to familiarize myself with it much more, refactor an entire class, setup a build environment to test it etc.

mtneug:
@do-you-even-curl no worries. You don't mind then, if I would submit such a patch?

t-schroeder:
@mtneug Of course not. Go ahead. :+1:

mtneug:
Closed in favor of #449.

PR #418, opened 2018-09-09
MH-13075: make ACL entries unique prior to running ACL comparisons

slampunk:
The ACL comparator method assumes that the entries of the ACLs to check are unique. However, it is possible to construct an ACL where entries are duplicated.

This patch alters one of the ACLs to have distinct entries prior to running the comparison check. The included unit test is an example of a false positive match.

### DIFF
--- a/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
+++ b/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
@@ -52,7 +52,7 @@ public final class EqualsUtil {
     return bothNotNull(a, b) && a.getClass().equals(b.getClass());
   }
 
-  /** Compare the elements of two lists for equality treating the lists as sets. */
+  /** Compare the (distinct) elements of two lists for equality treating the lists as sets. */
   public static boolean eqListUnsorted(List<?> as, List<?> bs) {
     if (as == null || bs == null) {
       return eqObj(as, bs);
--- a/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
+++ b/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
@@ -25,6 +25,7 @@ package org.opencastproject.util;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /** Utility function helping to implement equality. */
 public final class EqualsUtil {
@@ -53,7 +54,12 @@ public final class EqualsUtil {
 
   /** Compare the elements of two lists for equality treating the lists as sets. */
   public static boolean eqListUnsorted(List<?> as, List<?> bs) {
-    if (as != null && bs != null && as.size() == bs.size()) {
+    if (as != null && bs != null) {
+      as = as.stream().distinct().collect(Collectors.toList());
+
+      if (as.size() != bs.size()) {
+        return false;
+      }
       for (Object a : as) {
         if (!bs.contains(a))
           return false;
--- a/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
+++ b/modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
@@ -52,7 +52,16 @@ public final class EqualsUtil {
     return bothNotNull(a, b) && a.getClass().equals(b.getClass());
   }
 
-  /** Compare the (distinct) elements of two lists for equality treating the lists as sets. */
+  /**
+   * Compare the (distinct) elements of two lists for equality treating the lists as sets.
+   * <p>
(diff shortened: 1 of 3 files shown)


KatrinIhler:
This is really a question of definition, isn't it? Is a list that contains an entry twice really equal to a list that only contains it once? When considering ACLs that probably makes sense since the access rights are effectively the same whether an entry exists once or three times, but I wonder if that equals method is also used for other things were that doesn't make sense?

KatrinIhler (commented):
This is really a question of definition, isn't it? Is a list that contains an entry twice really equal to a list that only contains it once? When considering ACLs that probably makes sense since the access rights are effectively the same whether an entry exists once or three times, but I wonder if that equals method is also used for other things were that doesn't make sense?

modules/matterhorn-common/src/main/java/org/opencastproject/util/EqualsUtil.java
```diff
@@ -53,7 +54,12 @@ public static boolean eqClasses(Object a, Object b) {
 
   /** Compare the elements of two lists for equality treating the lists as sets. */
   public static boolean eqListUnsorted(List<?> as, List<?> bs) {
-    if (as != null && bs != null && as.size() == bs.size()) {
+    if (as != null && bs != null) {
```
lkiesow: By reducing this, you could just say:

```java
if (as == null || bs == null) {
  return as == bs;
}
```

lkiesow (approved):
Looks good and comparison makes sense for ACLs.
Maybe simplify the condition but that's an optional improvement ;-P

slampunk:
@KatrinIhler spot on, the definition of equality it depends entirely on the items subject to the comparison. FWIW, grep'ing suggests that this comparison method is only ever used for ACLs.
I'll update the comment to reflect that roles will be checked uniquely.

@lkiesow eh, looks like my set theory's a bit rusty. I'll use that :)

slampunk:
Ah, I see a bug.
`as = [role_a, role_a, role_b]`
`bs = [role_a, role_a, role_b]`

will fail. Both `as` and `bs` will have to be made to have unique entries. Will update the PR asap.

staubesv:
@slampunk Please don't tell me that Opencast allows multiple (possible contradicting) access control entries in an access control list... does it really do this?

slampunk:
@staubesv contracting? conflicting?

I was able to apply this ACL to an event:
{"acl":{"ace":[
      {"action":"read","allow":true,"role":"ROLE_USER_ADMIN"},
      {"action":"write","allow":true,"role":"ROLE_USER_ADMIN"},
      {"action":"read","allow":true,"role":"ROLE_USER_TESTUSER"},
      {"action":"write","allow":true,"role":"ROLE_USER_TESTUSER"},
      {"action":"read","allow":true,"role":"ROLE_USER_TESTUSER"},
      {"action":"read","allow":false,"role":"ROLE_USER_TESTUSER"}
]}}
where after this is the response when querying /api/events/{id}/acl:
selection_114

slampunk:
Hi @staubesv, I've updated the PR.

The binary operation in question is now reflexive and symmetric as far as ACLs are concerned, i.e. 

- `eqListUnsorted(ACL1, ACL1)` resolves to `true` and
- `eqListUnsorted(ACL1, ACL2) = eqListUnsorted(ACL2, ACL1)` respectively.

staubesv:
@slampunk Updated old comment to "contradicting" ;-)

Also created https://opencast.jira.com/browse/MH-13089 as it seems superfluos and over-complicated to handle multiple ACEs per role to me - no ambitions to work on that, though.

Will do the review soon.

staubesv:
@slampunk Since it seems easy to add test cases, could you maybe add some more test cases that further nail the behaviour down? 

E.g.

as -> [ROLE_USER read=true, ROLE_USER read=true]
bs -> [ROLE_USER read=true, ROLE_USER read=true]

--> as == bs

or more crazy things like:

as -> [ROLE_USER read=true, ROLE_USER read=false]
bs -> [ROLE_USER read=true]

-->  as ?? bs 

Maybe enhance the comment of the method you change by a definition of what is supposed to be equal and what is not.

slampunk:
@staubesv ok will do

slampunk:
Hi @staubesv 

I've added additional comments and unit tests.

To cover the case of equality for `null=null`, I've opted to place that unit test into matterhorn-common (`AccessControlUtil.equals` could potentially have a need for returning `false` when either of the ACLs are `null`).

Let me know if you'd prefer any additional changes.

smarquard:
@staubesv, this PR fixes a specific production issue that arose for us with a duplicate ACE in an ACL. It's not our intention to fix all possible ACL/ACE issues here.

staubesv:
Hi @smarquard That's why I have created https://opencast.jira.com/browse/MH-13089. I haven't assumed that this PR would address that issue.

staubesv (approved):

PR #419, opened 2018-09-11
MH-12699: Remove opencast-paella binding dependency on Admin server 

miesgre:
JIRA ticket: https://opencast.jira.com/browse/MH-12699

The Paella Player's opencast-paella binding uses the Opencast Series service's REST endpoint to get series metadata. The default configuration is for series-remote to be on the Engage server which calls the series-impl on admin to retrieve series information. This makes an unwanted dependency between Paella player and the Opencast admin node.

The admin node dependency, and issues introduced by it, can be avoided by using the search/series endpoint on the Search servie's REST endpoint to retrieve series information. The Search service is native to the Opencast engage server, and it is already used by Paella's opencast-paella binding to get the mediapackage. So, it does not introduce additional unwanted service dependencies.

### DIFF
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
@@ -79,20 +79,30 @@ class Opencast {
     return this.getEpisode()
     .then(function(episode) {
       return new Promise((resolve, reject)=>{
-        var serie = episode.mediapackage.series;
-        if (serie != undefined) {
-          base.ajax.get({url:'/series/'+serie+'.json'},
-            function(data,contentType,code) {
-              self._series = data;
-              resolve(self._series);
-            },
-            function(data, contentType, code) {
-              reject();
-            }
-          );
-        }
-        else {
-          reject();
+        if (self._series) {
+	  resolve(self.series);
+	}
+	else {
+          var serie = episode.mediapackage.series;
+          if (serie != undefined) {
+            base.ajax.get({url:'/search/series.json', params:{'id': serie}},
+              function(data, contentType, code) {
+                if (data['search-results'].result) {
+                  self._series = data['search-results'].result;
+                  resolve(self._series);
+                }
+                else {
+                  reject();
+                }
+              },
+              function(data, contentType, code) {
+                reject();
+              }
+            );
+          }
+          else {
+            reject();
+          }
         }
       });
     });
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
@@ -80,9 +80,9 @@ class Opencast {
     .then(function(episode) {
       return new Promise((resolve, reject)=>{
         if (self._series) {
-	  resolve(self.series);
-	}
-	else {
+	        resolve(self.series);
+	      }
+	      else {
           var serie = episode.mediapackage.series;
           if (serie != undefined) {
             base.ajax.get({url:'/search/series.json', params:{'id': serie}},
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
@@ -80,9 +80,9 @@ class Opencast {
     .then(function(episode) {
       return new Promise((resolve, reject)=>{
         if (self._series) {
-	        resolve(self.series);
-	      }
-	      else {
+          resolve(self.series);
+        }
+        else {
           var serie = episode.mediapackage.series;
           if (serie != undefined) {
             base.ajax.get({url:'/search/series.json', params:{'id': serie}},


modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
```diff
@@ -79,20 +79,30 @@ class Opencast {
...
+        if (self._series) {
+	  resolve(self.series);
+	}
+	else {
+          var serie = episode.mediapackage.series;
+          if (serie != undefined) {
+            base.ajax.get({url:'/search/series.json', params:{'id': serie}},
+              function(data, contentType, code) {
+                if (data['search-results'].result) {
+                  self._series = data['search-results'].result;
```
lkiesow, line 91: Note that result can either be an array or a single element. You probably want to ensure it's always an array.
E.g. something along the lines of:

```
data = data['search-results'].result;
self._series = Array.isArray(data.result) ? data.result : [data.result];
```

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
```diff
@@ -79,20 +79,30 @@ class Opencast {
...
-            },
-            function(data, contentType, code) {
-              reject();
-            }
-          );
-        }
-        else {
-          reject();
+        if (self._series) {
+	  resolve(self.series);
```
lkiesow: Broken indentation.

lkiesow (changes requested):

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
lkiesow: Generally makes sense. I think there might be a problem though if you only have one series.

Also, could you please add a proper description to this pull request (e.g. reasoning for the change)? Copying the text over from Jira should be fine but it's really annoying as a reviewer to try finding information about pull requests in other sources. Be it Jira, be it the mailing list, … ;-P

miesgre: As the call is specifying the id of the series (?id=<serieid>) the call returns always a single element, so there is no need to convert it to an array. The function returns the series element of the episode.

karendolan: Hi @lkiesow , I can add my 2¢ on the benefit of avoiding the series service from the Player. Traditionally, for distributed Opencast systems that do not have an independent Solr server, the Series Impl is on the Admin node and the Series Remote is on the Engage node. When a call comes in from the Player to the series remote endpoint on the Engage node, the remote service uses the ServiceRegistry to locate a series Impl service that has availability to handle a series service inquiry.

Two issues
1. When admin is down the series remote service is out of service. This makes the player dependent on both the Engage and the Admin node.
2. The following ServiceRegistry method does a large scope search for services and the current load of services and the current availability of services running the Opencast system. This method takes multiple seconds to gather data, and even longer in a busy system. Once the series Impl is finally parsed out of all the gathered service data, the actual call from the remote to the impl takes only milliseconds. 

Not sure how to solve the expense of the getServiceRegistrationsByLoad() method. In the meantime, it's simplier and easlier for new implementers if the Player quickly retrieves locally accessible series metadata from the search endpoint. The search endpoint has all required series metadata from any mediapackage that has been previously published with a series association.

https://github.com/opencast/opencast/blob/ceb51e5a42d21997f478b0d84ef674ab4fc310ee/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java#L2166-L2187

lkiesow: > As the call is specifying the id of the series (?id=) the call returns always a single element, 

makes sense

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/01_prerequisites.js
```diff
@@ -79,20 +79,30 @@ class Opencast {
...
-            }
-          );
-        }
-        else {
-          reject();
+        if (self._series) {
+	        resolve(self.series);
+	      }
+	      else {
+          var serie = episode.mediapackage.series;
```
lkiesow: Still broken indentation.

lkiesow (changes requested):

lkiesow:
I tried testing this but one I clicked on a series, I got:

screenshot from 2018-09-12 23-06-52

What I did:

- Go to an existing episode
- Click on “Extend Tab Adapter”
- Click on series link

The same series can be found in the media module.
Looking at the network logs, I see a 404 for:

```
http://localhost:8080/paella/repository/noid/data.json?_=1536786813119
```

Is this related? If not, how can I test this?

lkiesow:
Note that a similar thing happens when I click on the presenter link.

miesgre:
@lkiesow This is not related. This is another issue. These links are pointing to an old HTML page.
We should change them to point to /engage/ui/. Please fill an issue and I will fix it.

To test this patch you need to open the browser developer and go to the javascript console and test the `paella.opencast.getSeries()` function. The function returns the series information of the current video (if the video is in a series)

lkiesow:
Ok, will test this tomorrow.
I've also added a ticket for the issue.

PR #420, opened 2018-09-11
MH-13090: Added support for blacklisting languages from the admin UI

gregorydlogan:
Usecase is that the Taiwanese flag is banned within mainland China, per the discussion on the mailing list.

* Blacklisting a user's current language causes them to use their fallback.
* Blacklisting *all* languages locks the user into either English, or whatever their default would be.
    * This causes a very minor UI bug in that the list is present, but blank, however I feel like this is highly unlikely to happen in a production system, and does not actually cause any issues anyway.

### DIFF
--- a/etc/org.opencastproject.adminui.endpoint.LanguageServiceEndpoint.cfg
+++ b/etc/org.opencastproject.adminui.endpoint.LanguageServiceEndpoint.cfg
@@ -1,0 +1,3 @@
+# Optional param to exclude languages from the admin-UI
+# Add languages to exclude using xx_XX notation, comma separated values
+org.opencastproject.adminui.languages.exclude=
--- a/etc/org.opencastproject.adminui.endpoint.LanguageServiceEndpoint.cfg
+++ b/etc/org.opencastproject.adminui.endpoint.LanguageServiceEndpoint.cfg
@@ -1,3 +1,4 @@
 # Optional param to exclude languages from the admin-UI
 # Add languages to exclude using xx_XX notation, comma separated values
-org.opencastproject.adminui.languages.exclude=
+# Default: <empty>
+#org.opencastproject.adminui.languages.exclude=
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
@@ -79,13 +79,14 @@ public class LanguageServiceEndpoint implements ManagedService {
   /** OSGi callback if properties file is present */
   @Override
   public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+    excludedLocales = new HashSet<>();
     if (properties == null) {
       logger.info("No configuration available, using defaults");
       return;
     }
 
     String excludes = StringUtils.trimToEmpty((String) properties.get(EXCLUDE_CONFIG_KEY));
-    excludedLocales = new HashSet<>(Arrays.asList(excludes.split(",")));
+    excludedLocales.addAll(Arrays.asList(StringUtils.split(excludes, ", ")));
   }
 
   @GET
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
@@ -24,17 +24,24 @@ package org.opencastproject.adminui.endpoint;
 import org.opencastproject.adminui.api.LanguageService;
 import org.opencastproject.adminui.util.Language;
 import org.opencastproject.adminui.util.LocaleFormattingStringProvider;
+import org.opencastproject.util.ConfigurationException;
 import org.opencastproject.util.doc.rest.RestQuery;
 import org.opencastproject.util.doc.rest.RestResponse;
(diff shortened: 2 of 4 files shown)


gregorydlogan:
The relevant config key to blacklist everything looks like this

```
org.opencastproject.adminui.languages.exclude=de_DE,el_GR,en_GB,en_US,es_ES,fil_PH,fr_FR,gl_ES,he_IL,nl_NL,pl_PL,sl_SI,sv_SE,tl_PH,tr_TR,zh_CN,zh_TW
```

etc/org.opencastproject.adminui.endpoint.LanguageServiceEndpoint.cfg
```diff
@@ -0,0 +1,3 @@
+# Optional param to exclude languages from the admin-UI
+# Add languages to exclude using xx_XX notation, comma separated values
+org.opencastproject.adminui.languages.exclude=
```
lkiesow: Can we mention the default and comment out the option if it's set to the default like this:

```
# Optional param to exclude languages from the admin-UI
# Add languages to exclude using xx_XX notation, comma separated values
# Default: <empty>
#org.opencastproject.adminui.languages.exclude=
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
```diff
@@ -65,6 +72,22 @@ void setLanguageService(LanguageService languageSrv) {
...
 
+  private static Set<String> excludedLocales;
+
+  public static final String EXCLUDE_CONFIG_KEY = "org.opencastproject.adminui.languages.exclude";
+
+  /** OSGi callback if properties file is present */
+  @Override
+  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+    if (properties == null) {
+      logger.info("No configuration available, using defaults");
```
lkiesow: This is not correct. Deleting the configuration file at runtime would cause `properties == null` but `excludedLocales` would still be set to the old value. Initializing the set right at the beginning of the method would solve the issue:

```java
excludedLocales = new HashSet<>();
if (properties == null) {
  …
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
```diff
@@ -65,6 +72,22 @@ void setLanguageService(LanguageService languageSrv) {
...
+      return;
+    }
+
+    String excludes = StringUtils.trimToEmpty((String) properties.get(EXCLUDE_CONFIG_KEY));
+    excludedLocales = new HashSet<>(Arrays.asList(excludes.split(",")));
```
lkiesow: This would cause problems when the value contains whitespaces like this: `prop = de, en, fr`.
Maybe use something like:

```
excludedLocales.addAll(StringUtils.split(properties.get(EXCLUDE_CONFIG_KEY).toString(), ", "));
```

This would also split `en fr` but that should not cause a problem.

lkiesow (changes requested):

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/LanguageServiceEndpoint.java
gregorydlogan: IMO there are going to be few places in the world where this is going to be a useful feature, so I'm not going to worry terribly about edge cases like this.

KatrinIhler (approved):
I tested this and it works as expected with one exception: If German is my current language and I blacklist it, it is still chosen afterwards. If I switch to another language, German does not appear in the list of available languages however.

Maybe this happens because German is my fallback? I cannot reproduce this behavior with other languages - if American English is chosen and I blacklist it, it falls back to German for example - even if German is blacklisted, so apparently the fallback ignores if a language is excluded... 

Two other nitpicks:
First, maybe the configuration to exclude everything could be added as an example to the configuration file comments? Then it would also be clearer that comma-separated means "en_US, de_DE", not "en_US,de_DE".

Second, when updating other configuration files, a message like "INFO Updating list provider /home/katrin/Code/opencast/build/opencast-dist-develop-6-SNAPSHOT/etc/listproviders/licenses.properties" appears in the log. I think that's quite helpful, maybe something like this could be added when updated() is called?

lkiesow:
> First, maybe the configuration to exclude everything could be added as an example to the configuration file comments? Then it would also be clearer that comma-separated means "en_US, de_DE", not "en_US,de_DE".

`StringUtils.split(…, separatorChars)` will split both cases:

```java
for (String x: StringUtils.split("asdf, asdf,ert", ", "))
  System.out.println(">>" + x + "<<");
```

produces:

```sh
% java Test                                                                                                                                            [0]
>>asdf<<
>>asdf<<
>>ert<<
```

lkiesow:
> Second, when updating other configuration files, a message like "INFO Updating list provider […]/licenses.properties" appears in the log. I think that's quite helpful, maybe something like this could be added when updated() is called?

I disagree. The expected case is that an admin sets a configuration and it's what the system uses. There is no need to state that it is indeed the case. If you would only log every configuration file loaded, you would already produce around 190 log messages right now. If we add something like this, I think it should be a debug level log entry. And, of course, we should log if something is wrong but that should already be the case as we log exceptions.

lkiesow (approved):

PR #423, opened 2018-09-12
MH-13083: Unify modal navigation

jaysuz:
(Copied from [Jira](https://opencast.jira.com/browse/MH-13083))
The detail modals in the Admin UI are not consistent considering...

- Title of modal window
- Navigation between individual objects

The goal of this story is to make the detail modals consistent considering the user experience.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
@@ -128,7 +128,7 @@ angular.module('adminNg.controllers')
         $scope.$on('change', function (event, id) {
           $scope.$broadcast('clear');
           $scope.group = GroupResource.get({ id: id }, function () {
-            reloadSelectedRoles();
+            reloadRoles();
             reloadSelectedUsers();
           });
         });
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
@@ -126,10 +126,11 @@ angular.module('adminNg.controllers')
         };
 
         $scope.$on('change', function (event, id) {
+          $scope.$broadcast('clear');
           $scope.group = GroupResource.get({ id: id }, function () {
             reloadSelectedRoles();
             reloadSelectedUsers();
-          })
+          });
         });
 
         $scope.getSubmitButtonState = function () {
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
@@ -86,13 +86,10 @@ angular.module('adminNg.controllers')
         };
 
         if ($scope.action === 'edit') {
-            $scope.caption = 'USERS.GROUPS.DETAILS.EDITCAPTION';
             $scope.group = GroupResource.get({ id: $scope.resourceId }, function () {
                 reloadSelectedRoles();
                 reloadSelectedUsers();
             });
-        } else {
-            $scope.caption = 'USERS.GROUPS.DETAILS.NEWCAPTION';
         }
 
         $scope.submit = function () {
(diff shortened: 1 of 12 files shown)


modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
```diff
@@ -128,6 +125,14 @@ angular.module('adminNg.controllers')
           }
         };
 
+        $scope.$on('change', function (event, id) {
+          $scope.$broadcast('clear');
+          $scope.group = GroupResource.get({ id: id }, function () {
+            reloadSelectedRoles();
```
KatrinIhler: Switching between groups in the event details window throws a n.role.available.$promise is undefined error and the roles aren't updated.

modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
```diff
@@ -167,6 +135,62 @@ angular.module('adminNg.controllers')
             }
         };
 
+      /**
+       * This private function updates the scope with the data of a given user.
+       *
+       * @param id the id of the user
+       */
+      function fetchChildResources(id) {
+        console.log($scope.searchField);
```
KatrinIhler: This should be removed.

modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
```diff
@@ -167,6 +135,62 @@ angular.module('adminNg.controllers')
...
+          Notifications.removeAll('user-form');
+
+          fetchChildResources(id);
+        });
+
+        $scope.$on('clear', function () {
+          $scope.searchFieldExternal = '';
+          $scope.searchFieldEffective = '';
+        });
+
```
KatrinIhler, line 194: When I create a new user by simply filling out all required fields (so no roles for example), I repeatedly get the error "Error: n.userForm is undefined" for as long as that user exists. I can't reproduce this on develop.opencast.org so this probably has something to do with your changes? No idea with what exactly, though.

KatrinIhler (changes requested):
I tested this and found a couple of things:

modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
jaysuz: Done.

modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/groupController.js
jaysuz: Addressed with 590a6b83291a6ca8074747e5492e3328dc4c7ce0.

modules/admin-ui/src/main/webapp/scripts/modules/users/controllers/userController.js
jaysuz: For me, this happens on the develop branch as well, so I don't think this issue is related to the changes I've made. However, I noticed that the "getSubmitButtonState" method is invoked constantly, thus possibly causing this error. When using the ngClass directive as in 48a49bfb6f2d3d07096b69e17175b3214d5f9464, the error does not occur.

staubesv:
H  @gregorydlogan & @lkiesow 

It seems that Travis is in trouble:

> [ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.5:install-node-and-npm (install node and npm) on project opencast-admin-ui: Could not extract the Node archive: Could not extract archive: '/home/travis/.m2/repository/com/github/eirslett/node/8.9.4/node-8.9.4-linux-x64.tar.gz': EOFException -> [Help 1]

I've manually restarted the build but the same error occurred again.

Could you please have a look at why Travis fails?

staubesv:
@KatrinIhler & @lkiesow Seems the ball is currently in your hands... Is there anything left that needs to be addressed?

lkiesow:
@staubesv, no idea until I take another look at this…

lkiesow (approved):

PR #425, opened 2018-09-13
MH-13096: Set workflow variables with duplicated media package IDs

mtneug:
This PR changes the `duplicate-event` WOH such that the IDs of the new media packages are exposed as new workflow properties.

Use case: We want to automatically start a new workflow for the duplicated event (expect a new PR for starting new workflows soon 😄).


KatrinIhler (approved):
I tested this and it works as expected. I would prefer that the workflow configuration values presented in the admin UI would be alphabetically sorted since it looks kinda weird when one duplicated media package id is displayed at the top and others farther below, but that's not really what this pullrequest is about. Looks good!

PR #426, opened 2018-09-13
MH-13099, Warn when default credentials are being used

lkiesow:
Using default credentials in production is bad and users should never do
that but we have still seen that happening.

This patch will cause Opencast to print a big warning if the passwords
are not changed. Everything will run as usual (important for testing)
but it's now more likely that users will notice they have a problem.

### DIFF
--- a/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/AdminUserAndGroupLoader.java
+++ b/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/AdminUserAndGroupLoader.java
@@ -124,6 +124,20 @@ public class AdminUserAndGroupLoader implements OrganizationDirectoryListener {
     adminEmail = StringUtils.trimToNull(bundleCtx.getProperty(OPT_ADMIN_EMAIL));
     adminRoles = StringUtils.trimToNull(bundleCtx.getProperty(OPT_ADMIN_ROLES));
 
+    if ("opencast".equals(adminPassword)) {
+    logger.warn("\n"
+            + "######################################################\n"
+            + "#                                                    #\n"
+            + "# WARNING: Opencast still uses the default admin     #\n"
+            + "#          credentials. Never do this in production. #\n"
+            + "#                                                    #\n"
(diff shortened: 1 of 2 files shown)


modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
```diff
@@ -99,11 +99,26 @@
...
-    if (digestUsername == null)
+    if (digestUsername == null) {
       logger.warn("Digest username has not been configured ({})", DIGEST_USER_KEY);
+    }
+
     digestUserPass = StringUtils.trimToNull(cc.getBundleContext().getProperty(DIGEST_PASSWORD_KEY));
-    if (digestUsername == null)
+    if (digestUserPass == null) {
       logger.warn("Digest password has not been configured ({})", DIGEST_PASSWORD_KEY);
+    } else if ("CHANGE_ME".equals(digestUserPass)) {
```
gregorydlogan: Can we make this a constant?  It's lame, but IMO better form.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/AdminUserAndGroupLoader.java
```diff
@@ -124,6 +124,20 @@ public void activate(ComponentContext cc) throws Exception {
     adminEmail = StringUtils.trimToNull(bundleCtx.getProperty(OPT_ADMIN_EMAIL));
     adminRoles = StringUtils.trimToNull(bundleCtx.getProperty(OPT_ADMIN_ROLES));
 
+    if ("opencast".equals(adminPassword)) {
```
gregorydlogan: Likewise, constant?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
lkiesow: Done.

PR #427, opened 2018-09-14
MH-13100 fix series view in Paella

miesgre:
Jira Ticket: https://opencast.jira.com/browse/MH-13100

The series view liken from episodes in the paella player will just show
a big flashy warning about the identifier not existing despite it being
correct.

To reproduce:

- Go to an existing episode
- Click on “Extend Tab Adapter”
- Click on series link

The same series can be found in the media module.

Looking at the network logs, I see a 404 for:

  http://localhost:8080/paella/repository/noid/data.json?_=1536786813119

A similar problem happens when clicking on the presenter link.


staubesv (approved):
Thanks for fixing this!

PR #430, opened 2018-09-14
MH-13098: Add start-workflow WOH

mtneug:
This PR adds the new WOH `start-workflow` which starts a new workflow for given media package, workflow definition, and workflow properties.

Use case: In combination with #425 we want to automatically run a workflow on a duplicated event.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
@@ -1,0 +1,28 @@
+# StartWorkflowWorkflowOperationHandler
+
+## Description
+
+The StartWorkflowWorkflowOperationHandler can be used to start a new workflow for given media package and workflow definition.
+
+## Parameter Table
+
+|Configuration Key         |Example                              |Description                                     |
+|--------------------------|-------------------------------------|------------------------------------------------|
+|media-package\*           |e72f2265-472a-49ae-bc04-8301d94b4b1a |The ID of the media package that should be used |
+|workflow-definition\*     |fast                                 |The workflow definition that should be used     |
+|*configProperty*          |abc / false                          |Workflow configuration property                 |
+
+\* mandatory configuration key
+
+## Operation Example
+
+```xml
+<operation id="start-workflow">
+  <configurations>
+    <configuration key="workflow-definition">fast</configuration>
+    <configuration key="media-package">e72f2265-472a-49ae-bc04-8301d94b4b1a</configuration>
+    <configuration key="key">value</configuration>
+    <configuration key="publish">true</configuration>
+  </configurations>
+</operation>
+```
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
@@ -1,0 +1,128 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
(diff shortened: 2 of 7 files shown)


docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
```diff
@@ -0,0 +1,28 @@
...
+
+|Configuration Key         |Example                              |Description                                     |
+|--------------------------|-------------------------------------|------------------------------------------------|
+|media-package\*           |e72f2265-472a-49ae-bc04-8301d94b4b1a |The ID of the media package that should be used |
+|workflow-definition\*     |fast                                 |The workflow definition that should be used     |
+|*configProperty*          |abc / false                          |Workflow configuration property                 |
+
+\* mandatory configuration key
+
+## Operation Example
```
lkiesow, line 17: Maybe use 

```
```xml
<operation id="start-workflow">
  <configurations>
    <configuration key="workflow-definition">fast</configuration>
    <configuration key="media-package">e72f2265-472a-49ae-bc04-8301d94b4b1a</configuration>
    <configuration key="key">value</configuration>
    <configuration key="publish">true</configuration>
  </configurations>
</operation>
` ``
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
```diff
@@ -0,0 +1,129 @@
...
+    final WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
+    final String configuredMediaPackageID = trimToEmpty(operation.getConfiguration(MEDIA_PACKAGE_ID));
+    final String configuredWorkflowDefinition = trimToEmpty(operation.getConfiguration(WORKFLOW_DEFINITION));
+
+    // Get media package
+    final AQueryBuilder q = assetManager.createQuery();
+    final AResult r = q.select(q.snapshot())
+                       .where(q.mediaPackageId(configuredMediaPackageID).and(q.version().isLatest()))
+                       .run();
+    if (r.getSize() == 0) {
```
lkiesow: While this shouldn't happen, this could theoretically return multiple results.
Maybe use:

```java
if (r.getSize() != 1) {
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
```diff
@@ -0,0 +1,129 @@
...
+      throw new WorkflowOperationException(format("Media package %s not found", configuredMediaPackageID));
+    }
+    final MediaPackage mp = Enrichments.enrich(r).getSnapshots().head2().getMediaPackage();
```
lkiesow: No strong opinion here (I actually think both ways are a bit ugly ;-) but I would probably use for the sake of a less obscure code path:

```java
final MediaPackage mp = r.getRecords().head().get().getSnapshot().get().getMediaPackage();
```

lkiesow (commented):
Generally looks good. The comments are just could-be changes none of them should be required.

The only thing I'm wondering, *and I'm absolutely not sure of this*, is if you could use this to start workflows on media packages belonging to other tenants. Do you know if that's possible?

mtneug:
Thanks for the review, I applied the changes.

Whether this works across tenants, is a good question. Honestly, I don't really know as I have never dealt with multiple tenants. I had a quick look at the asset manager and found `AssetManagerWithSecurity`. From there it seems to depend on the user that is running the workflow. If it is a global admin then yes, otherwise the Asset Manager query is restricted to the current organization and the media package should not be found.

lkiesow:
I've tested this in a multi-tenancy setup and this allows you to import any recordings which are present in another tenant if you have `ROLE_ADMIN` which has several bad side effects. But I can accept that as AM bug :D 

If you are not a global admin, things work as expected. Hence, I'll merge this

PR #432, opened 2018-09-16
MH-13103, Runtime UI NG JavaScript Dependencies

lkiesow:
This patch harmonizes the runtime-ui-ng module with other modules using
JavaScript by using npm as package manager for dependencies and ESLint
as linter.


staubesv (changes requested):
This PR breaks the Runtime UI:

Before PR:

screen shot 2018-09-17 at 18 47 00

After PR:

screen shot 2018-09-17 at 19 01 04

Tested with Chrome Version 69.0.3497.92 (Official Build) (64-bit)

Seems a jquery related issue:

Uncaught ReferenceError: $ is not defined
    at index.html:10

The browser reports a 404 on  http://localhost:8080/js/jquery/jquery.js

lkiesow:
Interesting. Seems like that page is using the jQuery delivered by the REST docs…
Will fix that ina moment.

modules/runtime-info-ui/pom.xml
```diff
@@ -136,6 +136,25 @@
...
+            <goals>
+              <goal>copy-resources</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${basedir}/target/classes/ui/js/</outputDirectory>
+              <resources>
+                <resource>
+                  <directory>node_modules/</directory>
+                  <includes>
+                    <include>jquery/dist/jquery.min.js</include>
```
staubesv, line 151: Probably that means you need to adapt the index page of runtime-info-ui (not the rest_docs page)

(this is runtime-info-ui, not runtime-info-ui-ng)

modules/runtime-info-ui/src/main/resources/ui/index.html
```diff
@@ -5,104 +5,8 @@
...
-
-                    if (buildNumbers.length > 0) {
-                        $buildNumber.before("Build number" + (buildNumbers.length > 1 ? "s: " : ": "));
-                        checkAndDisplay(buildNumbers, $buildNumber);
-                    }
-                }
-            });
-        });
-    </script>
+    <script type="text/javascript" src="js/jquery/dist/jquery.min.js"></script>
```
staubesv, line 8: No strong opinion, but do we want the "dist" in the URL? "js/jquery.min.js" would be nicer ;-)

What do you think?

lkiesow: We want. That makes the deployment much easier in case there are multiple libraries at one point. And really no one will care about this path ;-P

staubesv:
Hi @lkiesow Just realised you added further rules in another. Please let me know when I shall continue the review.

lkiesow:
I'm done

staubesv (approved):

PR #438, opened 2018-09-18
MH-13109: Update Paella Player to 6.0.x

miesgre:


modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -1,20 +1,66 @@
 class OpencastToPaellaConverter {
 
-  constructor() {
+  constructor() {
```
KatrinIhler: Adding unnecessary spaces was probably a mistake, no? There are also some farther below...

staubesv (approved):

lkiesow:
@miesgre, @staubesv: Are you sure the new Paella version works in Opencast? It does not at all work for me locally and seems to be broken on [develop.opencast.org](https://develop.opencast.org) as well:

1. Go to https://develop.opencast.org
2. Select a recording and switch to the Theodul player to verify the recording works
3. Switch to https://develop.opencast.org/paella/ui/watch.html?id=…
4. Get the error `Error loading video! No video traks found`

miesgre:
@lkiesow The problem is that paella is only using */deploy to display the videos, and the videos in develop.opencast.org are */preview, so paella is ignoring those videos.

PR #440, opened 2018-09-20
MH-13082: Fix LTI security vulnerability and refactor LTI and OAuth classes

mtneug:
This refactors LTI and OAuth classes to OSGi services and implements the LTI user blacklist as discussed in #414 and during the developer meeting.

There is a new functionality in that this changes allows to configure multiple OAuth consumers instead of just one, so this might be seen as feature instead of a bug fix. Although, this might be small enough?

@gregorydlogan note that I have a branch ready for 3.x, however since `SecurityUtil` did not have that method back then, it is marginally different. Should I file separately?

### DIFF
--- a/docs/guides/admin/docs/modules/ltimodule.md
+++ b/docs/guides/admin/docs/modules/ltimodule.md
@@ -14,48 +14,52 @@ More information about the LTI specifications is available at
 [IMS Learning Tools Interoperability](http://www.imsglobal.org/activity/learning-tools-interoperability).
 
 Configure Opencast
-------------------------
+------------------
 
-To enable LTI authentication in Opencast, edit `OPENCAST/etc/security/mh_default_org.xml`
+### Configure OAuth authentication
 
-* In the Authentication Filters section, uncomment the oAuthProtectedResourceFilter:
-````
+LTI uses OAuth to authenticate users. To enable OAuth in Opencast, edit `OPENCAST/etc/security/mh_default_org.xml` and
+uncomment the oAuthProtectedResourceFilter in the Authentication Filters section:
+
+```xml
     <!-- 2-legged OAuth is used by trusted 3rd party applications, including LTI. -->
     <!-- Uncomment the line below to support LTI or other OAuth clients.          -->
     <ref bean="oauthProtectedResourceFilter" />
-````
+```
 
-* Replace CONSUMER_KEY and CONSUMER_SECRET with LTI the key and secret values that you will use in your LMS:
-````
-    <!-- ####################### -->
-    <!-- # OAuth (LTI) Support # -->
-    <!-- ####################### -->
+To configure OAuth consumers (e.g. a LMS), edit
+`OPENCAST/etc/org.opencastproject.kernel.security.OAuthConsumerDetailsService.cfg` and replace CONSUMERNAME,
+CONSUMERKEY, and CONSUMERSECRET with the values you will use in your LMS:
 
-    <!-- This is required for LTI. If you are using LTI and have enabled the oauthProtectedResourceFilter  -->
-    <!-- in the list of authenticationFilters above, set custom values for CONSUMERKEY and CONSUMERSECRET. -->
+```properties
+oauth.consumer.name.1=CONSUMERNAME
+oauth.consumer.key.1=CONSUMERKEY
+oauth.consumer.secret.1=CONSUMERSECRET
+```
 
-    <bean name="oAuthConsumerDetailsService" class="org.opencastproject.kernel.security.OAuthSingleConsumerDetailsService">
-    <constructor-arg index="0" ref="userDetailsService" />
-    <constructor-arg index="1" value="CONSUMERKEY" />
-    <constructor-arg index="2" value="CONSUMERSECRET" />
-    <constructor-arg index="3" value="constructorName" />
-    </bean>
-````
+### Configure LTI (optional)
 
-* To give LMS users the same username in Opencast as the LMS username, uncomment the constructor arguments
-below and update CONSUMERKEY to the same key used above:
+To give LMS users the same username in Opencast as the LMS username, edit
+`etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg` and add the configured OAuth consumer key
+to the list of highly trusted keys.
 
-````
-    <!-- Uncomment to trust usernames from the LTI consumer identified by CONSUMERKEY.           -->
-    <!-- Users from untrusted systems will be prefixed with "lti:" and the consumer domain name. -->
+```properties
+lti.oauth.highly_trusted_consumer_key.1=CONSUMERKEY
+```
 
-    <constructor-arg index="1" ref="securityService" />
-    <constructor-arg index="2">
-      <list>
-        <value>CONSUMERKEY</value>
-      </list>
-    </constructor-arg>
-````
+Use can exempt specific users even if a highly trusted consumer is used by configuring a blacklist. Additionally, there
+are settings for excluding the system administrator as well as the digest user (enabled by default).
+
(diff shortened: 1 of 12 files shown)


gregorydlogan:
@mtneug That depends, what is in that branch? :)  I'm not sure we should be accepting features into older versions, but if the fix from #414 (or your other branch) works well enough for versions < 6.0 then we can use that.

mtneug:
@gregorydlogan I will prepare a new PR for Opencast 3, without the ability for multiple OAuth consumers. As discussed in #414, this solution has some shortcomings. I will close #414 in favor of the new PR.

As for this PR, still hope we find a reviewer for 6.x 😬.

modules/kernel/pom.xml
```diff
@@ -300,6 +300,8 @@
               OSGI-INF/pingback.xml,
               OSGI-INF/rest.xml,
               OSGI-INF/security-filter.xml,
+              OSGI-INF/security-lit.xml,
```
gregorydlogan: This is the right filename, but the wrong spelling.  Fix in a follow up PR.

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -0,0 +1,42 @@
+# OAuth consumer keys with should be highly trusted.
+#
+# By default OAuth consumer are regarded as untrusted and a user authenticating via such
+# systems receives a rewritten username in the form of "lit:{ltiConsumerGUID}:{ltiUserID}".
```
gregorydlogan: is it `lit:{...` or `lti:{...`?

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -0,0 +1,42 @@
...
+# Allow the Opencast system administrator user to authenticate as such via LTI.
+#
+# Note that this user may still authenticate via LTI, but the username will be rewritten,
+# even if a trusted OAuth consumer key is used.
+#
+# Note that this option does not apply to custom users having the ROLE_ADMIN. Use the
+# blacklist below instead.
+#
+# Default: false
+#lti.allow_system_administrator=true
```
gregorydlogan: Default and comment-out value should match.

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -0,0 +1,42 @@
...
+
+# Allow the Opencast digest user to authenticate as such via LTI.
+#
+# Note that this user may still authenticate via LTI, but the username will be rewritten,
+# even if a trusted OAuth consumer key is used.
+#
+# Default: false
+#lti.allow_digest_user=true
```
gregorydlogan: Default and commented-out value should match

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -0,0 +1,42 @@
...
+
+# A blacklist of users not allowed to authenticate via LTI as themselves.
+#
+# Note that these users may still authenticate via LTI, but their username will be rewritten,
+# even if a trusted OAuth consumer key is used.
+#
+# Multiple users can be configured, by incrementing the counter. The list is read sequentially
+# incrementing the counter. If you miss any numbers it will stop looking for further users.
+#lti.blacklist.user.1=myAdminUser
```
gregorydlogan, line 42: Should include the default (ie, no user is blacklisted).

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -0,0 +1,42 @@
...
+# Note that this user may still authenticate via LTI, but the username will be rewritten,
+# even if a trusted OAuth consumer key is used.
+#
+# Default: false
+#lti.allow_digest_user=true
+
+# A blacklist of users not allowed to authenticate via LTI as themselves.
+#
+# Note that these users may still authenticate via LTI, but their username will be rewritten,
+# even if a trusted OAuth consumer key is used.
```
gregorydlogan, line 38: I feel like I'm not 100% sure what this paragraph means.  They can't authenticate as themselves via LTI, but their username is rewritten, as it is in the other cases as well.  Maybe I'm just not familiar enough with how LTI works to get this, but it seems like this description needs to be fleshed out more.

docs/guides/admin/docs/modules/ltimodule.md
```diff
@@ -14,48 +14,51 @@ More information about the LTI specifications is available at
...
+are settings for excluding the system administrator as well as the digest user (enabled by default).
+
+```properties
+lti.allow_system_administrator=false
+lti.allow_digest_user=false
+lti.blacklist.user.1=myAdminUser
+```
+
+> **Notice:** Marking a consumer key as highly trusted can be a security risk! If the usernames of sensitive Opencast
+> users are not blacklisted, the LMS administrator could create LMS users with the same username and use LTI to grant
```
gregorydlogan, line 60: Can we list these usernames?  I'd think it would only be the two, and those are not likely to change, correct?

gregorydlogan (approved):
I don't have the backend systems to test this with, but it looks right.  Minor documentation and filename bits, but I would merge this into develop prior to 6.x cut and we'll test it more during the release phase.

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
mtneug: How would that look like? Just `#lti.blacklist.user.1=`?

docs/guides/admin/docs/modules/ltimodule.md
mtneug: By default yes, but there might be many more (e.g. custom admin or API users). However, I can list the default usernames.

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
mtneug: Well their username is rewritten (e.g. `lti:lmsng.school.edu:1234` instead of `mtneug`). Opencast's role provider will see the new username and therefore do not load roles associated with `mtneug`. So more accurately: they can't authenticate as themselves because their username is rewritten.

docs/guides/admin/docs/modules/ltimodule.md
mtneug: Done

mtneug:
I have had some problems yesterday in our test infrastructure, which I think should now be fixed. Will report back.

mtneug:
Is fixed now and working as intended. Originally I missed to include `org.osgi.service.cm.ManagedService` into the OSGi service description.

mtneug:
@gregorydlogan I created #449 for older releases.

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
gregorydlogan: Yes, I think so.  And something like `Default: no blacklisted users`?

etc/org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg
gregorydlogan: Ok, so this is down to my not having worked with LTI.  We're good here.

PR #449, opened 2018-09-25
MH-13082: Fix LTI security vulnerability and refactor LTI and OAuth classes

mtneug:
Same as #440, but for 3.x to 5.x. The only functional difference is the missing support for configuring multiple OAuth consumers.

### DIFF
--- a/modules/matterhorn-kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
+++ b/modules/matterhorn-kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
@@ -21,9 +21,11 @@
 
 package org.opencastproject.kernel.security;
 
-import org.opencastproject.security.api.SecurityService;
-
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.osgi.service.cm.ConfigurationException;
+import org.osgi.service.cm.ManagedService;
+import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.security.core.Authentication;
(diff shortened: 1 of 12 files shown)


lkiesow:
I've got some people reporting that their production cluster broke after a recent update with some LTI related complaints. Could this patch cause the issue if the configuration is not updated during that update? Note that in a lot of systems, auto-updates without configuration changes may happen on the same branch (e.g. 5.0 → 5.1 → 5.2). If so, is it possible to adapt the default configuration in a way that the old configuration does not break anything?

mtneug:
I don't thinks this is possible. The LTI objects are now created as OSGi services and no longer by Spring security. You would therefore have to adapt `mh_default_org.xml` and add two new configuration files (`org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg` and `org.opencastproject.kernel.security.OAuthConsumerDetailsService.cfg`). I would recommend reviewing the configuration in any case as the blacklist of users which should not use LTI varies between systems. The breaking change is unfortunate, but I don't see a way to use the original configuration.

modules/matterhorn-kernel/src/main/java/org/opencastproject/kernel/security/LtiLaunchAuthenticationHandler.java
```diff
@@ -35,109 +37,139 @@
...
+  private static final String HIGHLY_TRUSTED_CONSUMER_KEY_PREFIX = "lti.oauth.highly_trusted_consumer_key.";
+
+  /** The prefix of the key to look up a blacklisted user. */
+  private static final String BLACKLIST_USER_PREFIX = "lti.blacklist.user.";
+
+  /** The key to look up whether the admin user should be able to authenticate via LTI **/
+  private static final String ALLOW_SYSTEM_ADMINISTRATOR_KEY = "lti.allow_system_administrator";
+
+  /** The key to look up whether the digest user should be able to authenticate via LTI **/
+  private static final String ALLOW_DIGIST_USER_KEY = "lti.allow_digest_user";
```
aliyevTU, line 103: I noticed we have a typo in here (DIGIST - DIGEST). Would that be a problem, or was it intentionally left so?

lkiesow: It's likely not intentionally but just a typo.
It should not be a problem since it's just the name of an internal constant.

Nevertheless, it would be nice if it's being fixed.
Can you submit a pull request for this against `develop`?

mtneug: This was not intentional. Fixed in #605.

PR #450, opened 2018-09-25
MH-13061: Display responsible person for workflows

jaysuz:
(Copied from [Jira](https://opencast.jira.com/browse/MH-13061))
The Admin UI is currently missing to display an important piece of information: The name of the user that has started workflows.

The goal of this story is to add this information at the following places:

- Events->Events->Event Details->Workflows (just the name of user)
- Events->Events->Event Details->Workflows->Details (more details about the user here)

For full details and reasoning see [MH-13061](https://opencast.jira.com/browse/MH-13061).

Here are some screenshots:

**[OLD] Events->Events->Event Details->Workflows**

old_workflows

**[NEW]  Events->Events->Event Details->Workflows**

new_workflows

**[OLD] Events->Events->Event Details->Workflows->Details**

old_workflow_details

**[NEW/ADMIN]  Events->Events->Event Details->Workflows->Details**

new_workflow_details_privileged

**[NEW/UNPRIVILEGED]  Events->Events->Event Details->Workflows->Details**

new_workflow_details_unprivileged


lkiesow:
@jaysuz, note that this pull request has merge conflicts.

It would also be great if you could rebase your commits and squah them into one commit with a proper commit message. Start the message with the ticket number (`MH-13061, …`) and follow the suggestions of [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/).

In general, it's nice to clean up the commits directly before creating a pull request.

jaysuz:
@lkiesow, sorry about that, I'm still adapting to the conventions used in this particular project.

jaysuz:
Please be advised that in the meantime, we have found a few inconsistencies that need addressing.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
             "TITLE": "Title",
-            "STATE": "State",
-            "STARTED": "Started",
+            "STATUS": "Status",
+            "STARTED": "Started At",
             "SERIES": "Series",
-            "SUBMITTED": "Submitted",
+            "SUBMITTED_AT": "Submitted At",
+            "SUBMITTED_BY": "Submitted By",
+            "DURATION": "Duration",
```
staubesv: @jaysuz Please rename key and value to "EXECUTION_TIME": "Execution time"

staubesv:
@lkiesow Please note that the PR is ready for review

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
           "WORKFLOWS": {
             "ACTIONS": "Actions",
             "DETAILS": "Details",
-            "DESCRIPTION": "",
-            "MORE_INFO": "More Information & Actions",
-            "ID": "ID",
+            "DESCRIPTION": "Description",
+            "MORE_INFO": "More Information",
+            "ID": "WIID",
```
lkiesow: No one will know what this abbreviation means. Either leave it at `ID` because it is the identifier of this workflow after all and you explicitely removed the mention of instances further below, or expand it to something self-explanatory like `workflow identifier`

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
-            "STATE": "State",
-            "STARTED": "Started",
+            "STATUS": "Status",
+            "STARTED": "Started At",
             "SERIES": "Series",
-            "SUBMITTED": "Submitted",
+            "SUBMITTED_AT": "Submitted At",
+            "SUBMITTED_BY": "Submitted By",
+            "EXECUTION_TIME": "Execution Time",
+            "WDID": "WDID",
```
lkiesow: No one will know what this means, except maybe for developers. Since this identified the workflow definition/template, maybe name it `Workflow definition`?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
-            "ID": "ID",
+            "DESCRIPTION": "Description",
+            "MORE_INFO": "More Information",
+            "ID": "WIID",
             "TYPE": "Type",
             "TITLE": "Title",
-            "STATE": "State",
-            "STARTED": "Started",
+            "STATUS": "Status",
+            "STARTED": "Started At",
```
lkiesow: To be consistent with other tables, I suggest to not capitalize every word. Additionally, given only this string, you could not say what this field is for. E.g. it could also mean “Stared at location”. A better description (if you want to extend this) would be `Start date`

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
             "SERIES": "Series",
-            "SUBMITTED": "Submitted",
+            "SUBMITTED_AT": "Submitted At",
```
lkiesow: Same here. Suggestion: `Submission date`

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
+            "SUBMITTED_BY": "Submitted By",
```
lkiesow: `Suggestion: `Submitter`

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -872,15 +872,18 @@
...
+            "EXECUTION_TIME": "Execution Time",
```
lkiesow: `Execution time`

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -890,7 +893,8 @@
               "STOP": "Stop execution workflow instance",
```
lkiesow: Not your PR directly but can you maybe fix this as well? Either change it to `Stop execution of workflow instance` or to `Stop executing workflow instance`. Furthermore, you started removing the word instance at some places. If you want to remove it consistently, these keys could get be without as well: `Stop executing workflow` or even just `Stop workflow`

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -416,55 +416,58 @@ <h2 translate="EVENTS.EVENTS.DETAILS.HEADER" translate-values="{{titleParams}}">
...
-                      <td>{{ item.id }}</td>
-                      <td>{{ item.workflow }}</td>
+                      <td>{{ item.title }}</td>
+                      <td>{{ item.submittedBy }}</td>
+                      <td>{{ item.submittedAt | localizeDate : 'dateTime' : 'medium' }}</td>
                       <td>{{ item.status | translate }}</td>
-                      <td>{{ item.submitted | localizeDate : 'dateTime' : 'medium' }}</td>
-                      <td ng-if="$root.userIs('ROLE_UI_EVENTS_DETAILS_WORKFLOWS_EDIT')" >
+                      <td>
+                        <a class="more" title="{{ 'EVENTS.EVENTS.DETAILS.WORKFLOWS.TOOLTIP.DETAILS' | translate }}"
```
lkiesow: This is now quite inconsistent with all other tabs since it's the only one using an icon in an action column to get to a detailed view. Everywhere else a `Details` link is used.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -1483,31 +1486,46 @@ <h3>{{ item.name }}</h3>
             <header translate="EVENTS.EVENTS.DETAILS.WORKFLOW_DETAILS.TITLE"><!-- Workflow Details --></header>
             <div class="obj-container">
               <table class="main-tbl">
-                <tr>
-                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.ID"><!-- ID --></td>
```
lkiesow, line 1472: I would argue that this is more or less the most useful information in this view. Everything else, you have either seen before or is not really interesting at all. Hence I would prefer to leave this at the top.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -1483,31 +1486,46 @@ <h3>{{ item.name }}</h3>
...
+                  <td>{{ subNavData.submittedAt | localizeDate : 'dateTime' : 'medium' }}</td>
                 </tr>
                 <tr>
-                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.STARTED"><!-- Started --></td>
-                  <td>{{ subNavData.start | localizeDate : 'dateTime' : 'medium' }}</td>
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.STATUS"><!-- Status --></td>
+                  <td>{{ subNavData.status | translate }}</td>
+                </tr>
+                <tr ng-if="subNavData.status !== 'EVENTS.EVENTS.DETAILS.WORKFLOWS.OPERATION_STATUS.RUNNING'">
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.DURATION"><!-- Duration --></td>
```
lkiesow: This key is not defined

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -1483,31 +1486,46 @@ <h3>{{ item.name }}</h3>
...
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.SUBMITTED_AT"><!-- Submitted at --></td>
+                  <td>{{ subNavData.submittedAt | localizeDate : 'dateTime' : 'medium' }}</td>
                 </tr>
                 <tr>
-                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.STARTED"><!-- Started --></td>
-                  <td>{{ subNavData.start | localizeDate : 'dateTime' : 'medium' }}</td>
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.STATUS"><!-- Status --></td>
+                  <td>{{ subNavData.status | translate }}</td>
+                </tr>
+                <tr ng-if="subNavData.status !== 'EVENTS.EVENTS.DETAILS.WORKFLOWS.OPERATION_STATUS.RUNNING'">
```
lkiesow, line 1499: Why show this e.g. for paused workflows but not for running ones?

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -1483,31 +1486,46 @@ <h3>{{ item.name }}</h3>
...
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.DURATION"><!-- Duration --></td>
+                  <td>{{ subNavData.executionTime | humanDuration }}</td>
+                </tr>
+                <tr with-role="ROLE_ADMIN">
```
lkiesow, line 1503: Why show this only for ROLE_ADMIN?

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -1483,31 +1486,46 @@ <h3>{{ item.name }}</h3>
...
+                  <td translate="EVENTS.EVENTS.DETAILS.WORKFLOWS.WDID"><!-- WDID --></td>
+                  <td>{{ subNavData.wdid }}</td>
                 </tr>
               </table>
             </div>
           </div>
 
-          <div class="obj tbl-details">
+          <div class="obj tbl-details" with-role="ROLE_ADMIN">
```
lkiesow, line 1515: Again, why only ROLE_ADMIN (not even an org admin)?

lkiesow (dismissed):
Apart from the review comments:

- The code does not really do what this pull request claims it would do. Can you briefly update title and description? Don't spend too much time on it, it's just that if you look at pull requests later, it should be easy to tell what they are doing. (And it also helps to know what you are actually reviewing ;-P)
- The introduction of multi-word keys in the workflow details table often causes these to break into multiple lines (mainly due to the introduction to the usually very long workflow definition description). This looks rather odd. One idea for solving this would be to make the lines non-breakable. Another idea would be to introduce a minimal width for that column.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: I think it is not used anymore. Will check and remove the key if not used.

Note that there will be a date displayed as value so it is obvious that a date is meant.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: I prefer "Submitted at". Easier to read / not as clumsy.

Note that "Submitted At" is displayed just next to its value which is a date - therefore, its meaning is obvious (see screenshots).

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: I prefer "Submitted by". Very symmetric to "Submitted at". Reads easy and looks good (see screenshots).

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: What exactly is useful about the workflow identifier? The only thing I could think of is using the identifier for REST calls, but that is developer/system administration stuff... For any other uses, the identifier is meaningless.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Good catch, thanks!

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: The workflow definition makes only sense if you have access to the workflow definitions in means of configuration of Opencast. To my understanding, only developers and system administrators can change workflow definitions on disk. The assumption here is that those users have unrestricted access to Opencast which implies ROLE_ADMIN.

For all other users, the information seems not relevant.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Same as above: Only ROLE_ADMIN has access to the respective REST APIs (you could configure this differently, but seems common).

staubesv:
> The code does not really do what this pull request claims it would do. Can you briefly update title and description? 

This pull request is about the introduction of the "Submitted by" field. It needed some "space" in the table which caused some re-ordering.
But yes: We will update the title and description.

> The introduction of multi-word keys in the workflow details table often causes these to break into multiple lines (mainly due to the introduction to the usually very long workflow definition description). This looks rather odd. One idea for solving this would be to make the lines non-breakable. Another idea would be to introduce a minimal width for that column.

We don't introduce multi-word keys here: The workflow title is already "Process up on upload or schedule", isn't it? But it is true that the description spans multiple lines which looks really bad.
Minimal width won't work in all languages, but we will address the issue in a separate pull request that addresses https://opencast.jira.com/browse/MH-13126. Solution should be available within the next few weeks.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: "Workflow identifier" seems reasonable.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: "Workflow definition" seems reasonable.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: "Stop workflow" seems reasonable.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Agree, but in this table we have an "Actions" column with other actions (which we don't find in the other places that use "Details"). The use of the icon is consistent with all other tables that have an "Actions" column.

lkiesow: This would maybe ba consistent if it would also behave consistently in that it would open a modal dialog like the icon in the main table does. But here, it looks like the icon in the main table and behaves like the “Details” links used consistently in all of the dialogs ;-)

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
lkiesow: Consistency argument:

- It's “description” not “described as”
- It's “presenter” not “presented by”
- it's “contributor” not “contributed by”
- It's “start date” in the metadata tab, not “started at”
- …

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
lkiesow: <strike>I could imagine that e.g. an org admin would want to check which workflow was used for something and this is the only thing which identified the workflow definition.</strike>

lkiesow: Ups, workflow id. I meant the wdid below.

For the wiid:
Indeed, this is mostly useful for REST APIs (which do not all require ROLE_ADMIN) in terms of using that to launch something. Well, that or for reporting a problem with a specific workflow ;-)

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
lkiesow: I could imagine that e.g. an org admin would want to check which workflow was used for something and this is the only thing which identified the workflow definition. and more than ROLE_ADMIN has access to some API endpoints.

I agree that this is most useful for administrators, but that may also be org admins or again users reporting problems with a specific workflow.

lkiesow:
Seeing the table as key-value pairs, by *keys* I meant the left column. You cannot really do anything about the contents of the right one.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: Note that in the case of workflows, for example, we have two levels of metadata: Metadata describing the actual object itself (e.g. title, description, ...) and data about who has created the object at what point of time (submitted by, submitted at).

Currently, the "... by" and "... at" are used at several locations in the UI when it comes to the meta-metadata, e.g. Event: "Erstellt am"/but admittedly "Created", Series: "Erstellt von"/"Created By".

Personally - but that is a matter of taste - I would prefer going for the "... by" and "... at" variants across the UI instead of replacing the "... by" by "..." and "... at" by "... date".

Mostly because "Submitted by"/"Übermittelt von" seems more friendly to me as "Submitter"/"Übermittler". Same for "Submitted at"/"Übermittelt am" vs "Submission date"/"Übermittlungsdatum".

It might even help to distinguish between metadata vs meta-metadata.

Anyway, will bring up the point on the Today's technical meeting to see what others think.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Yep, but normally I would not expect any non-system-administrator to interact with REST APIs. And if there is something that cannot be done in the UI, we should rather implement the missing function in the UI instead of asking whatever users to use REST interfaces to achieve whatever they want.

Problems are - at least in our case - reported on event level, not workflow level: "Processing failed for event X".

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Again: Users - at least in our case - report problems on event level. They had the workflow instance ID for years now but there was no single case any kind of user would ever have contacted us referring to a workflow instance ID. I would be surprised if other Adopters would handle this different.

karendolan: @staubesv I could be reading this discussion incorrectly, but we run several types of workflows on events and the users/responsible non-admin personel definately want to know which workflows were run on an events, and which were not. They know to look at the details of the workflow so they can report on which operation caused an error. We do a hold-for-error-recovery on workflow errors, so it's important to get as much details from the users on the cause of the error as possible.

staubesv: @karendolan Unprivileged users can still see which workflows ran/run on an event as the workflow title is displayed. The users also still can look at the individual operations although I doubt that they will understand the error messages.
So they still can report which operation has failed - although this does not really provide more information as the event ID would have as it is easy to lookup the failed operation in the Admin UI anyway...

What unprivileged users (including tenant administrators) cannot do with this patch: They don't see the workflow instance ID, the workflow definition ID and the workflow configuration.

The motivation to hide this information are that...

- ... the workflow instance ID only makes sense if you - for whatever reason - would want to call REST endpoints for a given workflow instance ID
- ... the workflow definition ID and workflow configuration require access to / understanding of the workflow definitions (etc/workflows/*.xml). Without that, you present information to users which doesn't help them much

Generally, Opencast tends to be very technical which to me seems the main reason  that people are afraid of adopting Opencast. This could be improved by not displaying technical details that require quite some Opencast knowledge to just every user just everywhere. That is the main motivation to hide this fields from non-admin users.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
lkiesow: I don't think your "meta-metadata" declaration makes sense. Both are metadata about the job…

But in any case, I'm generally fine with changing these keys/labels. I don't really care for one variant or the other. But then let's do that consistently across the UI.

And please also consider that doing that has the downside of maybe invalidating a somewhat large (no idea how many keys we would need to touch) portion of our translation for arguably relatively little gain.

Again, we can do that. I have no intention on stopping that. But modifying just these keys looks weird.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
lkiesow: > Unprivileged users can still see which workflows ran/run on an event as the workflow title is displayed.

No, they can't. They can only tell that *a* workflow ran on the event. They cannot necessarily say which one because that's identified by the workflow instance id only and they cannot necessarily tell which type of workflow was used because that's only identified by the workflow definition id. These are identifiers after all. Descriptions or titles can help but they are not unique.

I agree that all of this workflow stuff is technical but remember that all of this is accessible only for users which you explicitly provided with access to this technical information. If you are looking at users which only report that something went wrong on an event level, there is no reason to provide that priviledge in the first place, is there?

gregorydlogan:
Why did we get rid of the workflow ID displays?  That's useful information that's now hidden...

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
gregorydlogan: If we're looking for shorter we could just use `Submitted`, which is clear enough with the date and time next to it.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
gregorydlogan: If we're looking for shorter we could just use `Started`, which is clear enough with the date and time next to it.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
gregorydlogan: Right, but there's no other easy way to get that information from the UI...

rute-santos:
Same here. Please put the workflow instance id back in the UI. It's very useful information.

staubesv:
As discussed in the today's technical meeting, I'll cross-post on the developer and users list to get more opinions about the controversial aspects of this pull request.

I've set it "on_hold" - please do not merge.

staubesv:
@gregorydlogan & @rute-santos Users with ROLE_ADMIN still see the workflow instance ID and workflow definition ID in the workflow details view.

Just wondering: Why would you want to information in the UI at all?

To my understanding it is useful if you want to do REST calls that require the workflow instance ID. But then again: What calls do you do that force you to use an API rather than just having a button in the UI?

Admittedly, the ID is in fact useful when the functionality is not offered by the UI or does not work when being performed through the UI. On the other hand, we maybe should then fix the broken functionality or add the missing functionality in the UI... 

Another argument would be the ability to uniquely refer to a workflow instance in case of processing failure, for example. But then again: Referring uniquely to an event does almost the same as it is very easy to look up the processing failure in the UI if the event ID is known.

In overall, I would prefer to remove the technical identifiers for non-admin users.

gregorydlogan:
@staubesv I can see the use case where a non-admin user has experienced some kind of issue, whether it's a bug in the UI, or some kind of weird processing error, and needs to refer to a specific workflow in an external ticketing system.  Being able to say workflow/recording 1234 has an error is much easier than workflow/recording 'classes about things, lecture 4'.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: Reverted to "ID"

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: Now uses "Workflow definition"

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: Changed to "Submitted"

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: Changed to "Submitter" and "Submitted"

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Re-introduced this column.

staubesv:
@lkiesow, @gregorydlogan, @rute-santos: We have reverted the following controversial changes:

- Reverted name "WIID" to "ID"
- Use name "Workflow definition" instead of "WDID"
- Ensure that columns with multi-word strings are formatted well
- Use "Submitter" and "Submitted" instead of "Submitted by" and "Submitted at"
- Re-introduce column "ID" on Event Details->Workflows

lkiesow:
I did another review on this and it looks mostly fine now.
The one thing I disagree with though is the usage of **☰** for the workflow details.

Every single other “details” view does use a link with the text “Details >” instead. Even within the same workflow dialog you are still presented with this text instead of the icon at some places:



But not only that. The icon is used in the admin interface to open new modal dialogs. A behavior which is distinctly different from the behavior of this new button since it modified the content of the already opened dialog instead.

Please use the usual “Details >” link here as well.
If you want to switch to **☰**, I can agree to that, but than please change it consistently within all of the dialogs and maybe(?) use a different color to distinguish the behavior. If you do, I also suggest to create a separate behavior for that since this pull request claims to be about “Displaying the responsible person for workflows” which has noting to do with the workflow details links.

staubesv:
@lkiesow @jaysuz has updated to pull request and re-introduced the "Details" link.

lkiesow:
Looks good now → merged

PR #451, opened 2018-09-25
MH-13125, Remove unused configuration keys

lkiesow:
This patch removes a number of configuration keys from the organization
configuration file since they are not used by Opencast any longer.


etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -202,55 +195,6 @@ prop.adminui.user.external_role_display=false
 # Default: USERS.INVERSE.WITH.USERNAME
 #prop.adminui.user.listname=USERS.INVERSE.WITH.USERNAME
 
-# Defines whether to display the default login and username on the login page
```
gregorydlogan, line 205: These are still present in some test resources, does it make sense to get rid of them there too?

- modules/admin-ui/src/test/resources/app/GET/groups/groups.json
- modules/admin-ui/src/test/resources/app/GET/groups/it_team.json
- modules/admin-ui/src/test/resources/app/GET/roles/roles.json
- modules/admin-ui/src/test/resources/app/GET/roles/roles.json

lkiesow: I noticed but these test resources are just responses representing user configuration and having more keys in there should not hurt. That said, I can remove them if you want me to. Do you want me to?

gregorydlogan: I'm in favour of having our testing resources match the current data that the system expects/produces, but I don't care that much.  If you don't get to it before I look at this again on Monday I'll just merge it as is.

lkiesow: I can do that later today.

lkiesow: Done

PR #454, opened 2018-09-27
MH-12727, Augument the suspendWaiterJob catch with RollbackException

karendolan:
The suspendWaiterJob currently has an OptimisticLock catch for the job complete race condition described in MH-12727. The RollbackException appears to be wrapping the OptimisticLockException  in cases at our site. This pull is to include the RollbackException to the existing catch to cover that case. 

This was logged moments after the inspect job had completed, implying that the OptimisticLockException was not caught because it was wrapped in the RollbackException (from a tx.commit())
``
JobBarrier.suspendWaiterJob - Unable to put Some(240757) into a waiting state, this may cause a deadlock: javax.persistence.RollbackException: javax.persistence.OptimisticLockException: Exception [EclipseLink-5006] (Eclipse Persistence Services - 2.6.4.v20160829-44060b6): org.eclipse.persistence.exceptions.OptimisticLockException#012Exception Description: The object [Job {id:240759, operation:Inspect, status:QUEUED}] cannot be updated because it has changed or been deleted since it was last read. #012Class> org.opencastproject.job.jpa.JpaJob Primary Key
``

### DIFF
--- a/modules/matterhorn-common/src/main/java/org/opencastproject/job/api/JobBarrier.java
+++ b/modules/matterhorn-common/src/main/java/org/opencastproject/job/api/JobBarrier.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 
 import javax.persistence.OptimisticLockException;
+import javax.persistence.RollbackException;
 
 /**
  * This class is a utility implementation that will wait for all given jobs to change their status to either one of:
@@ -159,7 +160,9 @@ public final class JobBarrier {
             if (setBlockerJob(j, waiter)) {
               blockedForJobs.add(j.getId());
             }
-          } catch (OptimisticLockException e) {
+          } catch (OptimisticLockException | RollbackException e) {
+            // Catch javax.persistence.RollbackException: javax.persistence.OptimisticLockException
+            // from tx.commit()
             // Try again, this happens if the job finishes before we get here
             // If the same exception happens again then we're in a very weird state
             if (setBlockerJob(j, waiter)) {


karendolan:
FYI -  I'm doing a proof-of-concept verification for this issue/workaournd and will post result...

gregorydlogan:
@karendolan I haven't looked at this yet, but will wait for the POC.

gregorydlogan:
@karendolan This looks good to me, it's ready to go, right?

karendolan:
Hi @gregorydlogan Yes! Sorry for the delay, it's good to go. It's already in our local site release.

karendolan:
@gregorydlogan 
Sorry for the modifications to this comment. In short, there may be other problems not addressed by this pull. But, this pull does support the case of the Optimistic lock being wrapped in a Rollback exception at that the nested try-catch.

The larger try-catch has an issue described by @rute-santos in MH-13147, which is thrown in this part of the suspendWaiterJob after the catch from this pull:
        this.serviceRegistry.updateJob(waiter);
Where waiter is updated by the dispatcher in the admin, after the the worker gets the waiter object and before it updates the job in the outer try-catch.

I other words, I think this pull is useful to support the inner try-catch, and a different more complex pull may be needed for the outer try-catch.

rute-santos:
Hi @gregorydlogan, @karendolan,
As Karen said, this change isn't enough. I am currently working on the fix.
Thanks!

gregorydlogan:
@rute-santos ok, fair enough.  Please ping me when this is ready to go!

gregorydlogan:
Note: I'm going to be offline from somewhat soon until January.  If this gets finished during that timeframe please find another reviewer, otherwise I'm happy to look at it next year :)

karendolan:
@gregorydlogan I'm closing this pull because it does not solve the underlying issue. @rute-santos has developed a better solution for a future pull.

PR #456, opened 2018-09-30
MH-13131, Fix Feed Service REST Docs

lkiesow:
Opening the REST documentation for the feed service's REST documentation
page results in internal server errors due to its invalid and
mismatching REST docs annotations.

This patch fixes the issue by adjusting the service path as well as
improving the validity check for path definitions.

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java
@@ -56,7 +56,7 @@ import javax.ws.rs.core.Variant;
 
 @Path("/")
 @RestService(name = "feedservice", title = "Feed Service",
-    abstractText = "This class is responsible of creating RSS and Atom feeds.", notes = "")
+    abstractText = "This class is responsible of creating RSS and Atom feeds.", notes = {})
 /**
  * This class is responsible of creating RSS and Atom feeds.
  * <p>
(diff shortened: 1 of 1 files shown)


modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/DocRestData.java
```diff
@@ -193,21 +193,16 @@ public void setAbstract(String abstractText) {
    * @return true if this path is valid, false otherwise
    */
   public static boolean isValidPath(String path) {
-    boolean valid = true;
```
staubesv: Maybe just

`
return (!isBlank(path) && path.matches("^/[\\w/{}.]*$") && !path.endsWith("/"));
`

BTW, shouldn't curly braces and the dot be escaped?

That would result in

`
return (!isBlank(path) && path.matches("^/[\\w/\\{\\}\\.]*$") && !path.endsWith("/"));
`

staubesv (changes requested):
Avoid multiple returns by simplifying the condition

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/DocRestData.java
lkiesow: No, the characters do not need to be escaped. Why would they need that?

And your return statement is not equivalent to the current functionality. That would be:

```java
return isNotBlank(path) 
  && (SLASH.equals(path)
      || (path.matches("^/[\\w/{}.]*$")
          && !path.endsWith("/")));
```

And I agreed to the previous author that not packing that in one line is easier to read.

lkiesow: Though you could pack the last two conditions in one regexp:

```java
path.matches("^/[\\w/{}.]*[\\w}.]$");
```

staubesv: > No, the characters do not need to be escaped. Why would they need that?

The escaping idea was because ".", "{" and "}" are special characters and I've seen examples regex where they were escaped. Absolutely possible that this is not required. The online regex evaluator I've tried did also not require that, although it did not work with "'\\\w" but did with "\w" ;-)
-> If that works, fine for me.

What about 

`return isNotBlank() && path.matches("^/$|^[\\w/{}.]+[^/]$")`

That one does match "/", requires start with "/" and also does not match anything that ends with "/" (except "/").

Anyway, I'm of course also fine with the if-statement approach although I'm not a big fan of decomposing expressions into multiple if-statements. That again is personal taste, though.

Should I accept as is or have I convinced you?

;-)

lkiesow: I think this discussion is becomming a good argument for avoiding complex regular expressions. `path.matches("^/$|^[\\w/{}.]+[^/]$")` would now accept probably unwanted characters as last character. `path.matches("^/$|^/[\\w/{}.]*[\\w}.]$");` would probably be the way to go then ;-P

I'll change it to this tomorrow:

```java
return isNotBlank() && path.matches("^/$|^/[\\w/{}.]*[\\w}.]$");
```

For the escaping: Java strings itself need escaping of `\`. Hence internally, `\\w` will become `\w`. Regular characters which have no special meaning don't need to be escaped. For example, in between `[` and `]` a `.` has no special meaning. It's just the character `.`.

lkiesow: Done

staubesv: The regex used is way to weak anyway and would accept all kinds of things like //{fds///{{{}{{fsdf///}, but that is not a problem introduced by this PR ;-)

Thanks for the simplification! Will continue the review.

lkiesow:
@staubesv, I noticed that the validity check we were looking at is actually part of a large chunk of dead and/or deprecated code. Since it's not strictly necessary for this fix, I've moved that into a larger maintenance patch which now is pull request #466 against `develop`. I think that makes more sense. I also had to extend the regexp a bit more to catch some other used paths. This now contains only the endpoint fix itself.

staubesv (approved):

PR #457, opened 2018-09-30
MH-13132, Fix REST Docs Overview Rendering

lkiesow:
The REST documentation overview page interprets special HTML characters
and syntax and does not render correctly on mobile devices.

This patch ensures loaded strings are rendered as Text and the UI is
displayed properly on mobile devices.

### DIFF
--- a/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js
+++ b/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js
@@ -33,13 +33,18 @@ $(document).ready(function($) {
   $('input').keyup(search);
 
   $.getJSON('/info/components.json', function(data) {
+    var docs = $('#docs'),
+        tpl = $('#template').html();
     $.each(data, function(section) {
       if ('rest' == section) {
         data.rest.sort((a,b) => a.path > b.path ? 1 : -1);
         $.each(data.rest, function(i) {
-          $('#docs').append('<li><a href="/docs.html?path=' + data.rest[i].path + '">'
-              + '<span>' + data.rest[i].path + '</span>'
-              + data.rest[i].description + '</a></li>');
+          let path = data.rest[i].path,
+              service = $(tpl);
+          $('a', service).attr('href', '/docs.html?path=' + path);
+          $('.path', service).text(path);
+          $('.desc', service).text(data.rest[i].description);
+          docs.append(service);
         });
         return;
       }
--- a/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.css
+++ b/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.css
@@ -13,7 +13,7 @@ h1 {
   background-color: #2075b1;
   margin: 0px;
   padding: 15px;
-  font-size: larger;
+  font-size: 16px;
   color: white;
 }
 
@@ -44,7 +44,7 @@ li {
 }
 
 ul {
-  font-size: smaller;
+  font-size: 12px;
   color: #333;
   padding: 20px 0;
 }
@@ -59,8 +59,14 @@ a:hover {
   color: gray;
 }
 
-li a span {
+li .path {
   display: inline-block;
   width: 200px;
-    font-family: monospace;
+  font-family: monospace;
+}
+
+li .desc {
+  display: inline-block;
+  min-width: 200px;
+  padding-left: 20px;
 }
--- a/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.html
+++ b/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.html
@@ -4,6 +4,7 @@
 <html lang=en>
 <head>
 <meta http-equiv=content-type content="text/html; charset=utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1">
 <link rel="shortcut icon" href="img/favicon.ico" />
 <title>Opencast REST Service Documentation</title>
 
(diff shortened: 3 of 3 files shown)


modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js
```diff
@@ -33,13 +33,18 @@ $(document).ready(function($) {
...
+    var docs = $('#docs'),
+        tpl = $('#template').html();
     $.each(data, function(section) {
       if ('rest' == section) {
         data.rest.sort((a,b) => a.path > b.path ? 1 : -1);
         $.each(data.rest, function(i) {
-          $('#docs').append('<li><a href="/docs.html?path=' + data.rest[i].path + '">'
-              + '<span>' + data.rest[i].path + '</span>'
-              + data.rest[i].description + '</a></li>');
+          let path = data.rest[i].path,
```
staubesv, line 42: According to https://caniuse.com/#search=let, this will not work on Opera mini (2.5% market share). No strong opinion on that, but seems that ECMA2015 is still not fully supported on all browsers out there.

Maybe just re-use data.rest[i].path two times to avoid variable hoisting and about service=$(tpl)? 

As said, no strong opinion here.

lkiesow: Other UI's of Opencast already require ES2015, including the Paella player. And I don't espect many Opera Mini users to complain about this UI ;-P

But in general, this might be something we want to bring up in the community.

staubesv:
The Paella player uses the traceur transpiler to ensure ES5-compatibility (it does not require ECMA 2015 therefore), Opencast does not.

modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js
lkiesow: Some example of ED2015 usage:
- [REST docs](https://github.com/opencast/opencast/blob/develop/modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js#L38)
- [Paella](https://github.com/opencast/opencast/blob/develop/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/04_datadelegates.js#L201)

staubesv: As said, the Paella player transpiles this into ES5 using traceur.

staubesv: And the REST docs example was introduced by you 16 days ago ;-)

lkiesow:
I'm talking about the bindings and they don't. They require ES2015

modules/runtime-info-ui-ng/src/main/resources/ui/rest_docs.js
lkiesow: > As said, the Paella player transpiles this into ES5 using traceur.

 If you insist: Please show me the code that calls the transpiler.

staubesv: Here you are: https://github.com/polimediaupv/paella/blob/develop/gulpfile.js#L88

Admittedly, their build task generates both a ES5 and ES6 variants of the player. Seems Opencast uses the ES5, at least assuming that our watch.html is correct (see https://github.com/opencast/opencast/blob/develop/modules/engage-paella-player/src/main/paella-opencast/ui/watch.html#L8).

Don't get me wrong: I've stumbled over the let statement because I fixed an error in the Paella build pipeline a while ago which was about replacing their uglifier library through a ES6-capable uglifier. The old one failed because of the let statement ;-)

In general, ES6 would be desirable. Just not clear to me whether that will cause problems or not. And it seems that you have silently introduced ES6 in Opencast with some of your recent pull requests. Not necessarily a bad thing, but not clear to me whether we can "just" do that.

lkiesow: As mentioned before:

> I'm talking about the bindings and they don't. They require ES2015

That's also the code I linked. It's in Opencast. Hence, please show me the code in Opencast that transpiles *our* code to ES5.

staubesv: Oh, now I got the point. In that case: @rrolf Have you been aware that Paella puts quite some effort to be compatible to ES5 when you updated the bindings? Or is this an oversight?

@miesgre What is actually your motivation of still having ES5 support in the Paella player in 2018? Is this still required? Is it because of the evil IE that still seems to have 10%+ market share?

lkiesow: @staubesv, let us not start a discussion here. If you are concerned about this, please bring this to list and maybe the technical meeting.

Back to topic: Do you now have a problem with this pull request containing `let`?

staubesv: No, the PR is fine for me. Just got confused ;-)

staubesv: @lkiesow The Opencast bindings also go through Traceur, so the official player is ES5.

https://github.com/opencast/opencast/blob/develop/modules/engage-paella-player/gulpfile.js#L11 copies the files into the folder where the external Paella standalone is copied and https://github.com/opencast/opencast/blob/develop/modules/engage-paella-player/gulpfile.js#L35 uses the build task of that external Paella player which will generate paella_player.js (or similar name) that is ES5 (including the Opencast bindings).

Note that Valencia still requires ES5 because some of their users rely on Internet Explorer support being working.

I can confirm that we also have institutions that still rely on Internet Explorer support in the player.

But anyway: Opencast Paella does support Internet Explorer (at least inofficially).

I don't see a problem for this PR as the REST docs are targeting developers which can be expected to not use IE.

staubesv (approved):

PR #458, opened 2018-09-30
MH-13092: Fix failing scheduling for non-english browsers

mtneug:
This should fix the scheduling problem in non-english browsers due to the recent usage of a moment.js version that includes locals.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/relativeDatesService.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/relativeDatesService.js
@@ -25,10 +25,9 @@ angular.module('adminNg.services')
 
   var RelativeDatesService = function () {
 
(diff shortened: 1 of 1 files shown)


lkiesow:
Note that this seems to break an admin-ui test. See Travis for details.

KatrinIhler (changes requested):
I've confirmed that changing the locale of moment locally instead of globally fixes the scheduling bug, however this change breaks the behavior of the counters (see below) as well as some tests - I don't yet know why though.

screenshot from 2018-10-04 11-33-27

mtneug:
I found that instantiating `localeMoment` within the function body fixes the failing test, although I have not the slightest idea why.

KatrinIhler (approved):
I tested it again and now the test and the stats work. I'm not entirely sure why, but it works.

PR #459, opened 2018-09-30
MH-13133, TypeError: Cannot read property 'results' of null

lkiesow:
Especially on slow networks, the admin interface throws exceptions since
it does not get all necessary resources but tries to access them anyway.

This patch checks if the received data is valid. It also prevents
Opencast from hiding errors in case there really are problems.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/resourceHelperService.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/resourceHelperService.js
@@ -44,26 +44,23 @@ angular.module('adminNg.services')
 
             var effectiveMapper = mapper || defaultMapper;
 
-            var result = [], data = {};
+            var result = [],
+                data = {};
 
-            try {
-              data = JSON.parse(responseBody);
-
-              // Distinguish between single and multiple values
-              if ($.isArray(data.results)) {
-                angular.forEach(data.results, function (event) {
-                  result.push(effectiveMapper(event));
-                });
-              } else {
-                result.push(effectiveMapper(data.results));
-              }
-            } catch (e) {
-              //console.warn(e);
-            }
+            data = JSON.parse(responseBody);
 
             // guard against empty dataset
-            if (angular.isUndefined(data.results)) {
+            if (data === null || angular.isUndefined(data.results)) {
               return;
+            }
+
+            // Distinguish between single and multiple values
+            if ($.isArray(data.results)) {
+              angular.forEach(data.results, function (event) {
+                result.push(effectiveMapper(event));
+              });
+            } else {
+              result.push(effectiveMapper(data.results));
             }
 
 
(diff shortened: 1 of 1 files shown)


lkiesow:
Btb. the error was easy to reproduce when throttling the network speed in Chrome (Developer Tools → Network →Throttling → Slow 3G) but you also get it from time to time on a fast network.

modules/admin-ui/src/main/webapp/scripts/shared/services/resourceHelperService.js
```diff
@@ -44,28 +44,25 @@ angular.module('adminNg.services')
...
-                result.push(effectiveMapper(data.results));
-              }
-            } catch (e) {
-              //console.warn(e);
-            }
+            data = JSON.parse(responseBody);
 
             // guard against empty dataset
-            if (angular.isUndefined(data.results)) {
+            if (data === null || angular.isUndefined(data.results)) {
```
staubesv: @lkiesow As this behaviour seems weird to me, I digged somewhat into what is causing JSON.parse to result in _null_. The cause is that responseBody is already _null_ and _JSON.parse(null)_ results in _null_.
This happens, for example, if a request is cancelled (or when ever the server responds with an empty body).

Maybe it would be better to check _(responseBody === null_) at the beginning of the method to make the code easier to understand.

What do you think?

staubesv (changes requested):

modules/admin-ui/src/main/webapp/scripts/shared/services/resourceHelperService.js
lkiesow: That would only catch half of the cases: If the returned JSON would be null (Important: JSON=null not JS-Value=null), you would still run into the same error. I can add a second check before parsing the response if you want me to. Though that's then more or less redundant.

Though in any case, `angular.isObject` may be a better chice than a check for `null`. I'll update this pull request.

Please let me know if you want me to also include the additional `null` check before parsing the JSON.

lkiesow: Thinking about this again, `angular.isObject` might hide some errors and it probably makes more sense to keep the `null` check. Still, let me know if you want another one before the `JSON.parse`.

lkiesow:
I've tried to simplify this file a bit more. I've also included the additional `null` check.

staubesv:
Hmm... the whole code is actually quite messy (talking about the original code, not your changes): If the method would consider the HTTP status code and the backend would always deliver the data correctly formatted, there would be no need for any of those checks like "responseBody = null , data == null, data.results undefined, data.results is array or not)...

Anyway, this PR certainly does improve the code and fixes the undesirable console output.

staubesv (approved):

lkiesow:
I agree. The backend not returning JSON in an error case while claiming it's JSON is already problematic. But unfortunately, that's a problem with nearly all admin-ng endpoints :-(

staubesv:
Well, the ResourceHelper is not supposed to parse anything if the backend does not return HTTP 200 as it expects a specific data format that only makes sense if the request succeeds...

Anyway, let's agree that this PR improves the situation and fixing "the Admin UI" might be out of scope ;-)

PR #460, opened 2018-09-30
MH-13134, Concurrency issue when loading admin interface statistics

lkiesow:
Opencast 6 makes the statistics counter in the admin interface
configurable. This can cause problems when the configuration is not yet
fully applied the first time, the counters are loaded as the refresh
interval is then never set and the counter values fail to load.

This patch simply adds the usual refresh if the configuration is not yet
properly set, ensuring that the values are loaded eventually.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/statsService.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/statsService.js
@@ -42,10 +42,11 @@ angular.module('adminNg.services')
       };
 
       /**
-         * Retrieve data from the defined API with the given filter values.
-         */
+       * Retrieve data from the defined API with the given filter values.
+       */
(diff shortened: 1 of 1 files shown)


KatrinIhler (approved):
I reproduced the bug and this fixes it! :)

staubesv:
@lkiesow & @KatrinIhler The problem is that the stats directive will immediately fetch the stats. At this point of time, the stats service is not yet configured. Just returning will cause the polling not to be continued. 

While this PR would solve this, I have written an alternative solution #469 which avoids fetching when the component is not yet initialized. I've found some other weird things that PR #469 resolves. The stats is still in quite bad shape, but don't have more time to work on this...

KatrinIhler:
I took a look at #469 and it seems to be the more elegant solution...

PR #467, opened 2018-10-03
MH-13135 Pending requests are not cancelled as expected

staubesv:
See https://opencast.jira.com/browse/MH-13135:

Steps to reproduce: 
1. Go to Events->Series 
2. Enter characters in the search fields 
 
The requests are not cancelled and the following exception is displayed in the browser console: 

angular.js:14199 TypeError: i.lastRequest.$cancelRequest is not a function 
    at Object.fetch (tableService.js:275) 
    at tableDirective.js:78 
    at t.$broadcast (angular.js:18320) 
    at Object.put (storageService.js:70) 
    at i.o.addFilterToStorage (tableFilterDirective.js:193) 
    at i.<anonymous> (tableFilterDirective.js:110) 
    at later (underscore-1.5.2.js:705)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/aclsResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/aclsResource.js
@@ -27,6 +27,7 @@ angular.module('adminNg.resources')
       params: { ext: 'acls.json' },
       method: 'GET',
       isArray: false,
+      cancellable: true,
       transformResponse: function (data) {
         return ResourceHelper.parseResponse(data, function (r) {
           var row = {};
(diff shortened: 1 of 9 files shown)


staubesv:
Hi @KatrinIhler  This is a follow-up of #360. Would you mind taking the review?

lkiesow:
Did you try this with a specific browser? I cannot reproduce the issue with either Chrome or Firefox.

staubesv:
I've tested with Chrome. Maybe you need to throttle the speed to slow 3G to reproduce the issue without too much tries. Chrome will indicate the cancelled requests in Inspector->Network. The issue only appears if requests are cancelled for all tables except the events table.

lkiesow:
Found a way to reproduce this.
Makes sense from the documentation.
Merged.

PR #468, opened 2018-10-03
MH-13142, Error parsing non-existent schedule

lkiesow:
In some situations (e.g. opening the event details), the admin interface
tries to load and parse scheduling information, even if none exist,
causing the JSON parser to throw an exception since the endpoint does
not return JSON.

This patch stops the data handling if the response code indicates that
there are no information.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/eventSchedulingResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/eventSchedulingResource.js
@@ -22,7 +22,13 @@
 
 angular.module('adminNg.resources')
 .factory('EventSchedulingResource', ['$resource', 'JsHelper', function ($resource, JsHelper) {
-  var transformResponse = function (data) {
+  var transformResponse = function (data, headers, status) {
+
+        // an event may have no schedule
+        if (status == 404) {
+          return;
+        }
+
         var parsedData,
             startDate,
             endDate,


modules/admin-ui/src/main/webapp/scripts/shared/resources/eventSchedulingResource.js
```diff
@@ -22,7 +22,13 @@
 
 angular.module('adminNg.resources')
 .factory('EventSchedulingResource', ['$resource', 'JsHelper', function ($resource, JsHelper) {
-  var transformResponse = function (data) {
+  var transformResponse = function (data, headers, status) {
+
+        // an event may have no schedule
+        if (status == 404) {
```
staubesv, line 28: Maybe status != 200 instead? That would also cover other cases where there is nothing to be parsed, e.g. 403

staubesv (changes requested):

modules/admin-ui/src/main/webapp/scripts/shared/resources/eventSchedulingResource.js
lkiesow: I've chosen 404 intentional since the service returning anything else than 200 or 404 indicates an error in Opencast and assuming that something is wrong it should then cause a JS error. Hiding that would make finding errors hard.

staubesv:
@lkiesow Note that Chrome and Safari will log the 404 anyway - but Firefox doesn't.

I'm fine with this PR, although it would take more effort to address this properly, e.g. the request should not return a 404 but rather a 200 indicating that no schedule information is available for the existing event.

staubesv (approved):

PR #469, opened 2018-10-04
MH-13141 Correctly initialize stats service

staubesv:
Steps to reproduce: 
1. Open the Admin UI in your browser 
2. Maybe reload the page (forcing full reload to prevent caching) 
  
Actual Results: 
Sometimes, the Admin UI Counters on the Events page don't work. 
  
Expected Results: 
The counters should work correctly 
  
Workaround (if any): 
Reload page until it works 

Analysis: 
The problem is the result of incorrect component initialization: The statsService is configured asynchronously which causes it to be used before it has been initialised.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/statsDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/statsDirective.js
@@ -58,7 +58,6 @@ angular.module('adminNg.directives')
         stats: '='
       },
       link: function (scope) {
-        scope.stats.fetch();
         scope.statsFilterNumber = -1;
 
         scope.showStatsFilter = function (index) {
@@ -107,6 +106,10 @@ angular.module('adminNg.directives')
           event.preventDefault();
           Storage.remove('filter');
           scope.statsFilterNumber = -1;
+        });
+        
+        scope.$on('$destroy', function() {
+          scope.stats.refreshScheduler.cancel();
         });
       }
     };
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/statsDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/statsDirective.js
@@ -107,7 +107,7 @@ angular.module('adminNg.directives')
           Storage.remove('filter');
           scope.statsFilterNumber = -1;
         });
-        
+
         scope.$on('$destroy', function() {
           scope.stats.refreshScheduler.cancel();
         });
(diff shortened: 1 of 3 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/directives/statsDirective.js
```diff
@@ -58,7 +58,6 @@ angular.module('adminNg.directives')
         stats: '='
       },
       link: function (scope) {
-        scope.stats.fetch();
```
staubesv, line 61: Fetching at this point of time doesn't reliable work as the stats service might not yet been configured

modules/admin-ui/src/main/webapp/scripts/shared/services/statsService.js
```diff
@@ -117,13 +111,6 @@ angular.module('adminNg.services')
         }
       };
 
-      // Reload the stats if the language is changed
```
KatrinIhler, line 120: I'm confused - why don't we want to do this anymore?

KatrinIhler (commented):
This seems to fix the bug and I agree that it's more elegant than #460. I have one question though:

modules/admin-ui/src/main/webapp/scripts/shared/services/statsService.js
staubesv: This file seems to be a copy-paste from the table service. The table service requires this because some localization work is done in the model ("resource"). 

For the stats service this was never required as the stats service has not localization logic in its model as the model is simply numbers.

-> We never needed this.

lkiesow (approved):

modules/admin-ui/src/main/webapp/scripts/shared/services/statsService.js
KatrinIhler: Ah, I see!

KatrinIhler (approved):

KatrinIhler:
@staubesv Before I merge this: Can you add the Jira Ticket number to your commits and maybe squash minor commits into one?

staubesv:
@KatrinIhler I have intentionally done this with several commits so that it is documented what has been changed. I don't think that it is important to have the JIRA ticket numbers as the pull request already has this.

Generally I highly recommend to not change GIT commits - this seems to be both tedious and a bad practise to me.

KatrinIhler:
@staubesv Fine with me, gonna merge as it is.

PR #324, opened 2018-07-03
MH-12978 Fix data-placeholder in add event wizard

staubesv:
See screenshots on https://opencast.jira.com/browse/MH-12978

PR #471, opened 2018-10-04
MH-13051: Fix dropdown placeholders

jaysuz:
This should resolve [MH-13051](https://opencast.jira.com/browse/MH-13051) by using the same workaround seen [here](https://github.com/opencast/opencast/pull/324/files).

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleValue.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleValue.html
@@ -36,7 +36,7 @@
             ng-disabled="!editMode && params.type === 'time'"
             ng-required="params.required"
             name="hours"
-            data-placeholder="{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.HOUR' | translate}}"
+            placeholder-text-single="'{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.HOUR' | translate}}'"
             ng-options="h.index as h.value for h in hours"
             />
       <select chosen
@@ -50,7 +50,7 @@
               ng-disabled="!editMode && params.type === 'time'"
               ng-required="params.required"
               name="minutes"
-              data-placeholder="{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.MINUTE' | translate}}"
+              placeholder-text-single="'{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.MINUTE' | translate}}'"
               ng-options="m.index as m.value for m in minutes"
               />
 </div>
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
@@ -260,7 +260,7 @@
                           data-disable-search-threshold="0"
                           ng-model="source.start.hour"
                           ng-change="onTemporalValueChange('start')"
-                          data-placeholder="{{ 'EVENTS.EVENTS.DETAILS.SOURCE.PLACEHOLDER.HOUR' | translate}}"
+                          placeholder-text-single="'{{ 'EVENTS.EVENTS.DETAILS.SOURCE.PLACEHOLDER.HOUR' | translate}}'"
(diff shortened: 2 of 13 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleValue.html
```diff
@@ -50,7 +50,7 @@
               ng-disabled="!editMode && params.type === 'time'"
               ng-required="params.required"
               name="minutes"
-              data-placeholder="{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.MINUTE' | translate}}"
+              placeholder-text-single="'{{ 'EVENTS.EVENTS.NEW.SOURCE.PLACEHOLDER.MINUTE' | translate}}'"
```
KatrinIhler, line 53: Any idea where this is visible in the Admin-UI?

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
```diff
@@ -260,7 +260,7 @@ <h2 translate="EVENTS.EVENTS.DETAILS.HEADER" translate-values="{{titleParams}}">
                           data-disable-search-threshold="0"
                           ng-model="source.start.hour"
                           ng-change="onTemporalValueChange('start')"
-                          data-placeholder="{{ 'EVENTS.EVENTS.DETAILS.SOURCE.PLACEHOLDER.HOUR' | translate}}"
+                          placeholder-text-single="'{{ 'EVENTS.EVENTS.DETAILS.SOURCE.PLACEHOLDER.HOUR' | translate}}'"
```
KatrinIhler, line 263: These placeholders are actually never visible, are they? Since there's always already an hour/minute chosen... (same is true for a couple others). But I'm not against leaving it in there in case that behavior changes.

modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/bulk-message.html
```diff
@@ -50,7 +50,7 @@
                          ng-change="wizard.step.applyTemplate()"
                          ng-model="wizard.step.ud.email_template"
                          ng-options="template as template.name for template in wizard.step.templates"
-                         data-placeholder="{{ 'EVENTS.SEND_MESSAGE.MESSAGE.EMAIL_TEMPLATE' | translate }}..."
+                         placeholder-text-single="'{{ 'EVENTS.SEND_MESSAGE.MESSAGE.EMAIL_TEMPLATE' | translate }}...'"
```
KatrinIhler, line 53: Do you know where this is visible in the Admin-UI?

KatrinIhler (commented):
Looks good so far, I didn't find anything to complain about, but for a couple of these I don't know where they're actually visible in the UI, maybe you do?

modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/bulk-message.html
staubesv: @KatrinIhler That one is not visible as the bulk-message implementation was never finished.

modules/admin-ui/src/main/webapp/scripts/shared/partials/editableSingleValue.html
staubesv: Just wanted to find an example to find out that this code might be dead: Originally, there was a metadata type start_time that would have used this, but seems like this was remove when switching to the datetime picker.

Anyway, not directly related to this PR.

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/event-details.html
staubesv: Would prefer to keep them - we already have a first request that asks us to not preselect the weekday in this wizard. Who know what happens next ;-)

KatrinIhler: Yeah, better leave it in then.

KatrinIhler (approved):
In this case I think this is ready for merging, unless @staubesv has found something that should be changed?

staubesv:
@KatrinIhler Haven't found anything here that needs to be changed.

PR #475, opened 2018-10-05
MH-13151, Update to Apache Karaf 4.1.6

lkiesow:
This updates Opencast's Karaf dependency to 4.1.6 (latest 4.1.x release)
as first part of the Karaf update.

Notably changes:

- log4j2 is now used at runtime which requires a different logging
  configuration.
- Pax web and its web component dependencies got a major update which
  also requires an update of cxf.
- Guava got a major update to prevent dependency issues caused by the
  use of javax.annotate. This also required a few minor code changes.
- A few javax.annotate annotations have been removed from commons to
  prevent dependency issues. We do not use them for automatic checks
  anyway.
- config.properties is now auto-generated again since the hack of
  hard-coding its content should not be necessary anymore with the
  latest two changes.
- A new start script binary called `inc` was added. The file is used by
  the updated start script.
- Opencast's Karaf features now have a version. This would theoretically
  allow us to install any specific Opencast version from a Karaf
  console.


lkiesow:
An update to Karaf 4.2.1 will follow as separate pull request.

etc/org.ops4j.pax.logging.cfg
```diff
@@ -1,30 +1,61 @@
...
+# Colors for log level rendering
+color.fatal = bright red
+color.error = bright red
+color.warn = bright yellow
+color.info = bright green
+color.debug = cyan
+color.trace = cyan
+
+# Common pattern layout for appenders
+log4j2.pattern = %d{ISO8601} | %-5p | %-16t | %-32c{1} | %X{bundle.id} - %X{bundle.name} - %X{bundle.version} | %m%n
```
gregorydlogan: This pattern generates insanely long lines.  With a quick test run I generated lines of 732 characters.  Any chance we can remove the `%-16t` and `%X{bundle.id} - %X{bundle.name} - %X{bundle.version}` fields?  This tells us (relatively) useless information like `fileinstall-/home/greg/opencast/upstream/build/opencast-dist-develop-7-SNAPSHOT/etc/encoding` and `64 - opencast-composer-ffmpeg - 7.0.0.SNAPSHOT`

lkiesow: I agree. I've replaced this with the currently used pattern.

staubesv (approved):
Did some basic tests. Seems to work fine.

PR #476, opened 2018-10-08
MH-13152, Reduce Workflow Messages

lkiesow:
The high amount of messages sent while workflows are processed can
easily cause performance issues and reduce parallel throughput.

Workflow messages contain metadata about the event and they are used to
update information in Opencast's administrative user interface and in
the external API. These information may change within workflow
operations and need to be updated after each operation as well as on
general workflow state changes (e.g. starting a workflow).

Updates are unnecessary, however, when new operations are launched since
they have been updated right before that operation already and noting
has changed since. Not sending those messages will cut the total amount
of workflow messages almost in half.

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -1397,8 +1397,15 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
       // Update both workflow and workflow job
       try {
         job = serviceRegistry.updateJob(job);
-        messageSender.sendObjectMessage(WorkflowItem.WORKFLOW_QUEUE, MessageSender.DestinationType.Queue,
-                WorkflowItem.updateInstance(workflowInstance));
+
+        WorkflowOperationInstance op = workflowInstance.getCurrentOperation();
+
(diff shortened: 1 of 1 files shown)


smarquard (approved):
I tested this on our 4.x dev system with some additional logging, and it reduces the number of workflow messages per operation from 2 to 1.

This seems a good candidate for cherry-picking into 3.x, 4.x and 5.x (or adopters handling high volumes may want to do this in local builds).

smarquard:
Is there any issue with the last operation in a workflow here? That is, if the last operation updates mediapackage info, is there a subsequent end-of-workflow operation that will update the index? I'm not seeing any issues with this, just wondering in theory.

lkiesow:
The `op == null` condition should catch the end of workflows. The `op.getState() != RUNNING` instead of something like `state == INSTANTIATED` is a deliberate choice to not block special cases like `FAILING`. Mostly, we probably wouldn't need to send messages then either but there may be special cases and these messages are in any case very rare. So, better safe than sorry ;-P

About the base branch, I have no problem redirecting this to `r/6.x` and maybe `r/5.x`. I wouldn't go before that. Any preferences?

smarquard:
r/5.x seems a good choice.

JamesUoM:
I would be interested in backporting this to r/3.x

karendolan:
+1

lkiesow:
Feel free to do that.
I'll leave this at 5.x where it is now.

PR #478, opened 2018-10-09
MH-13110: Fix display of metadata in event creation summary

jaysuz:
This will ensure that the summary page of the 'new-event' wizard displays all fields that have an assigned value, also handling cases where a data object key might be of the type 'ordered_text' or desired object values are passed as object keys.

Resolves: [MH-13110](https://opencast.jira.com/browse/MH-13110)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
@@ -97,7 +97,16 @@ angular.module('adminNg.services')
 
           params.presentableValue = presentableValue;
         } else {
-          params.presentableValue = params.collection[value];
+          if (params.collection.hasOwnProperty(value)) {
+            params.presentableValue = params.collection[value];
+          } else {
+            // this should work in older browsers, albeit looking clumsy
+            var matchingKey = Object.keys(params.collection)
+              .filter(function(key) {return params.collection[key] === value;})[0];
+            params.presentableValue = params.type === 'ordered_text'
+              ? JSON.parse(matchingKey)['label']
+              : matchingKey;
+          }
         }
       } else {
         params.presentableValue = value;
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
@@ -47,14 +47,20 @@ angular.module('adminNg.services')
       me.ud[mainMetadataName] = mainData;
       if (mainData && mainData.fields) {
         for (i = 0; i < mainData.fields.length; i++) {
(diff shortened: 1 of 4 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
```diff
@@ -97,7 +97,16 @@ angular.module('adminNg.services')
...
+          if (params.collection.hasOwnProperty(value)) {
+            params.presentableValue = params.collection[value];
+          } else {
+            // this should work in older browsers, albeit looking clumsy
+            var matchingKey = Object.keys(params.collection)
+              .filter(function(key) {return params.collection[key] === value;})[0];
+            params.presentableValue = params.type === 'ordered_text'
+              ? JSON.parse(matchingKey)['label']
+              : matchingKey;
+          }
```
KatrinIhler, line 108: I don't quite understand why this is necessary, could you elaborate?

KatrinIhler (commented):
I took a look at this and this definitely improves the situation! A couple of things I noticed:
* default values are currently not shown in the summary (try configuring a default license for example)
* the read-only publisher field is also not shown in the summary, although that might make sense since it can't be changed anyway...but would probably also be the cause if readOnly is set to False (didn't test this)
* the start date for the Upload option is still shown as "Today at..."

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
jaysuz: Previously, it was assumed that if the metadata object had a collection property and the value property did not have an array assigned to it, the value property would have a value that is named like a property of said collection. (I realise that this is worded confusingly, but that happens when a property is named *value*.) As you can see in the following screenshot, this would hold true for certain metadata fields having this collection property, but not for all of them. Those fields that did not make it to the summary had a **value** named like the value of the value property, not a property. To make the mess complete, collection elements with the type *ordered_text* are passed with a JSON string literal as property name, so to get to the desired *presentable value* such a property name would have to be parsed first, allowing to extract the wanted value from the parsed object.

metadata

Obviously, this is far from a perfect solution, but these inconsistencies are apparently rooted so deep that I dare not jump down into this rabbit hole.

jaysuz:
@KatrinIhler, this PR should be ready to be reviewed again as well.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
KatrinIhler: God, reading this could cause a stroke. But yeah, that's not really the topic of this PR.

KatrinIhler:
@jaysuz Sorry for the wait - I just tested this again and it looks good. Lets merge this!

PR #479, opened 2018-10-10
MH-13111: Fix display of metadata in series creation summary

jaysuz:
This will ensure that the summary page of the 'new-series' wizard displays all fields that have an assigned value, also handling cases where a data object key might be of the type 'ordered_text' or desired object values are passed as object keys.

Resolves: [MH-13111](https://opencast.jira.com/browse/MH-13111)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-series/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-series/metadata.js
@@ -44,14 +44,20 @@ angular.module('adminNg.services')
       me.ud[mainMetadataName] = mainData;
       if (mainData && mainData.fields) {
         for (i = 0; i < mainData.fields.length; i++) {
+          var field = mainData.fields[i];
+          // preserve default value, if set
+          if (field.hasOwnProperty('value') && field.value) {
+            field.presentableValue = me.extractPresentableValue(field);
+            me.ud[mainMetadataName].fields[field.id] = field;
+          }
           // just hooking the tab index up here, as this is already running through all elements
-          mainData.fields[i].tabindex = i + 1;
-          if (mainData.fields[i].required) {
-            me.updateRequiredMetadata(mainData.fields[i].id, mainData.fields[i].value);
-            if (mainData.fields[i].type === 'boolean') {
+          field.tabindex = i + 1;
+          if (field.required) {
+            me.updateRequiredMetadata(field.id, field.value);
+            if (field.type === 'boolean') {
               // set all boolean fields to false by default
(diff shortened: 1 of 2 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-series/metadata.js
```diff
@@ -92,7 +92,16 @@ angular.module('adminNg.services')
...
+          if (params.collection.hasOwnProperty(value)) {
+            params.presentableValue = params.collection[value];
+          } else {
+            // this should work in older browsers, albeit looking clumsy
+            var matchingKey = Object.keys(params.collection)
+              .filter(function(key) {return params.collection[key] === value;})[0];
+            params.presentableValue = params.type === 'ordered_text'
+              ? JSON.parse(matchingKey)['label']
+              : matchingKey;
+          }
```
KatrinIhler, line 103: Which older browsers is this supposed to support? (As far as I know, the Admin UI doesn't work for Internet Explorer anyway...) Would there be a more elegant solution otherwise? Also if I understand this correctly you assume that the entries of an ordered_text metadata field will always be in JSON, and I'm not sure that has to be that way...

KatrinIhler (commented):
Pretty much the same as in #478: Default values aren't being shown in the summary and I'm kind of wary of that one code snippet, but it sure looks better than before.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-series/metadata.js
jaysuz: I was not aware that the Admin UI does not support Internet Explorer. I sometimes forget that this project has been in existence for a long time, not realising that old looking code probably **is** old.

This would be more elegant (and more performant, for that matter): 
```Object.keys(params.collection).find(key => params.collection[key] === value);```

As to the possibility of the *ordered_text* type not being formatted as a JSON string literal, I think this will (in our case) never happen because for an element to be ordered, there needs to be information about a positional index, as well as a value, which almost requires a JS object. Or am I missing something?

jaysuz:
@KatrinIhler, I have implemented your feedback.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-series/metadata.js
KatrinIhler: Yeah, you're right. Let's leave it like this for now.

KatrinIhler:
@jaysuz Looks good, let's merge this! :)

PR #480, opened 2018-10-10
MH-13158, Use default functional interface for SecurityUtil#runAs

lkiesow:
Instead of using the custom internal functional code, this patch allows
the `runAs` method to accept a simple Lambda function which makes using
the method easier.


karendolan:
Dumb question, but this pull makes me wonder how active ingests, scheduling, and running workflows are impacted by running index rebuilds. In other words, the index appears to be juggling data updates from multiple activies and sources for the same mediapackages.

staubesv:
Hi @lkiesow 

Would you be willing to also adopt this for SecurityContext.runInContext()?

staubesv (approved):

PR #481, opened 2018-10-10
MH-13159: Fix mattermost notification operation issues

mtneug:
This fixes these issues in the Mattermost WOH:

1.  wrong construction of the JSON request
2. add call to `useSystemProperties` in the HTTP client construction (e.g. to use the configured HTTP proxy which is otherwise ignored)
3. extract Gson object to static object

### DIFF
--- a/modules/mattermost-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/mattermost/notification/MattermostNotificationWorkflowOperationHandler.java
+++ b/modules/mattermost-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/mattermost/notification/MattermostNotificationWorkflowOperationHandler.java
@@ -101,8 +101,6 @@ public class MattermostNotificationWorkflowOperationHandler extends AbstractWork
    */
   public static final String PUT = "put";
 
-
-
   /**
    * The logging facility
    */
@@ -127,6 +125,11 @@ public class MattermostNotificationWorkflowOperationHandler extends AbstractWork
    * The scale factor to the sleep time between two notification attempts
    */
   public static final int SLEEP_SCALE_FACTOR = 2;
+
+  /**
+   * Gson instance for JSON serialization.
+   */
+  private static Gson gson = new GsonBuilder().create();
 
   /**
    * {@inheritDoc}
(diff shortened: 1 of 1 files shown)


modules/mattermost-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/mattermost/notification/MattermostNotificationWorkflowOperationHandler.java
```diff
@@ -217,7 +217,7 @@ private String makeJson(String s, WorkflowInstance workflowInstance) {
     json.addProperty("text", s);
     GsonBuilder builder = new GsonBuilder();
     Gson gson = builder.create();
```
lkiesow: Would it make sense to have this as a static global object instead of initializing this each time? Afaik, Gson is thread-safe after all.

mtneug: I have updated the pull request.

modules/mattermost-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/mattermost/notification/MattermostNotificationWorkflowOperationHandler.java
```diff
@@ -128,6 +126,11 @@
    */
   public static final int SLEEP_SCALE_FACTOR = 2;
 
+  /**
+   * Gson instance for JSON serialization.
+   */
+  private static Gson gson = new GsonBuilder().create();
```
staubesv, line 132: Why not new Gson(); instead of new GsonBuilder().create(); ?

lkiesow: I think it was my original recommendation since I usually use `GsonBuilder`. Since we do not make any configuration, you could use `new Gson()` in this case, but it really does not matter which you use, does it?

staubesv: It doesn't really matter, but Gson() would be shorter ;-)

I'm also fine with the longer variant.

lkiesow: That's true :D

staubesv (approved):

PR #483, opened 2018-10-10
MH-13161, Simplify log statements

lkiesow:
This patch reduces the complexity of some log statements, in particular
by replacing the use of ExceptionUtils#getStackTrace by the same
functionality built into the logger used by Opencast.


lkiesow:
Fixed merge conflicts

PR #487, opened 2018-10-11
MH-13164, Load catalog for snapshot message effeciently

lkiesow:
Instead of loading the dublincore XML and de-serialize it into a Java
object only to directly serialize it again as XML, we can just load and
use the XML directly.

### DIFF
--- a/modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/assetmanager/AssetManagerItem.java
+++ b/modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/assetmanager/AssetManagerItem.java
@@ -22,12 +22,13 @@ package org.opencastproject.message.broker.api.assetmanager;
 
 import static com.entwinemedia.fn.Prelude.chuck;
 
+import org.opencastproject.mediapackage.Catalog;
 import org.opencastproject.mediapackage.MediaPackage;
+import org.opencastproject.mediapackage.MediaPackageElements;
 import org.opencastproject.mediapackage.MediaPackageException;
 import org.opencastproject.mediapackage.MediaPackageParser;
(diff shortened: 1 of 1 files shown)


smarquard:
Seems like a good efficiency fix to target at 6.x ?

lkiesow:
Rebased to `r/6.x`

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/assetmanager/AssetManagerItem.java
```diff
@@ -292,17 +294,16 @@ public String apply(DeleteEpisode a) {
...
+    for (Catalog catalog: mp.getCatalogs(MediaPackageElements.EPISODE)) {
+      try (InputStream in = workspace.read(catalog.getURI())) {
+        dcXml = IOUtils.toString(in, StandardCharsets.UTF_8);
+      } catch (Exception e) {
+        throw new IllegalStateException(String.format("Unable to load dublin core catalog for event '%s'",
+                mp.getIdentifier()), e);
       }
     }
-    return new TakeSnapshot(mp.getIdentifier().compact(), MediaPackageParser.getAsXml(mp), dc,
+    return new TakeSnapshot(mp.getIdentifier().compact(), MediaPackageParser.getAsXml(mp), dcXml,
```
staubesv, line 306: @lkiesow The semantic difference is that the former code would have failed early in case of an invalid XML while the new code will not check anything considering that data it loads from the workspace but simply pass the content to TakeSnapshot. 

Have you considered possible impacts this could have?

lkiesow: yes

staubesv (approved):

PR #488, opened 2018-10-12
MH-13160, Send actually required data in workflow messages

lkiesow:
Instead of sending a serialized workflow instance (which can be huge)
and later retrieving additional data from somewhere else about which we
do not know if it is still there or if it has been modified since, this
patch will send the required data–and only that–directly included in the
workflow messages.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
@@ -24,43 +24,21 @@ package org.opencastproject.index.service.message;
 import static org.opencastproject.index.service.impl.index.event.EventIndexUtils.getOrCreateEvent;
 import static org.opencastproject.index.service.impl.index.event.EventIndexUtils.updateEvent;
 
-import org.opencastproject.authorization.xacml.manager.api.AclServiceFactory;
-import org.opencastproject.authorization.xacml.manager.api.ManagedAcl;
 import org.opencastproject.index.service.impl.index.event.Event;
-import org.opencastproject.index.service.impl.index.event.EventIndexUtils;
-import org.opencastproject.index.service.util.AccessInformationUtil;
 import org.opencastproject.matterhorn.search.SearchIndexException;
 import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.message.broker.api.MessageSender;
 import org.opencastproject.message.broker.api.workflow.WorkflowItem;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalog;
-import org.opencastproject.metadata.dublincore.DublinCoreUtil;
-import org.opencastproject.security.api.AccessControlList;
-import org.opencastproject.security.api.AccessControlParser;
-import org.opencastproject.security.api.AclScope;
-import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.User;
 import org.opencastproject.util.NotFoundException;
-import org.opencastproject.util.data.Option;
-import org.opencastproject.util.data.Tuple;
-import org.opencastproject.workflow.api.WorkflowInstance;
-import org.opencastproject.workspace.api.Workspace;
 
-import com.entwinemedia.fn.data.Opt;
-
-import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.util.List;
 
 public class WorkflowMessageReceiverImpl extends BaseMessageReceiverImpl<WorkflowItem> {
 
   private static final Logger logger = LoggerFactory.getLogger(WorkflowMessageReceiverImpl.class);
-
-  private Workspace workspace;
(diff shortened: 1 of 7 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
```diff
@@ -71,110 +49,69 @@ public WorkflowMessageReceiverImpl() {
...
+    final MediaPackage mediaPackage = workflowItem.getMediaPackage();
+
+    // Load or create the corresponding recording event
+    try {
+      Event event = getOrCreateEvent(eventId, organization, user, getSearchIndex());
+      event.setCreator(user.getName());
+      event.setWorkflowId(workflowItem.getWorkflowInstanceId());
+      event.setWorkflowDefinitionId(workflowItem.getWorkflowDefinitionId());
+      event.setWorkflowState(workflowItem.getState());
+      event.setAccessPolicy(workflowItem.getAccessControlListJSON());
```
staubesv, line 96: The original source code did not update the ACL and Dublincore if the workflow instance is in state.isTerminated(). That seems reasonable as the workflow instance is not the data source for the event (the "archive" is).

To my understanding, the idea of creating and updating an event here is a workaround for the poor design idea to create an event as a side-effect of workflow execution during ingest - rather than creating the event at ingest and then start the workflow for this event which would make all this code totally obsolete as the workflow itself has neither metadata catalogs nor ACLs. The only thing this code would need to update is the event status as that one indeed does depend on the workflow instance...

... anyway, that is way out of scope of this PR, so back to this pull request:
Omitting the test !state.isTerminated() seems a step into the wrong direction to me and also seems to involve additional work load in case the workflow is in state.isTerminated().

Is there a specific reason you want to omit that test?

modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
```diff
@@ -71,110 +49,69 @@ public WorkflowMessageReceiverImpl() {
...
+
+      // Update metadata
+      DublinCoreCatalog dcCatalog = workflowItem.getEpisodeDublincoreCatalog();
+      if (dcCatalog != null) {
+        updateEvent(event, dcCatalog);
+      }
+      event.setSeriesName(mediaPackage.getSeriesTitle());
```
staubesv: updateEvent on line 106 also does this (see https://github.com/opencast/opencast/blob/develop/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventIndexUtils.java#L506) , so line 103 can be removed.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1375,6 +1374,18 @@ public void update(final WorkflowInstance workflowInstance) throws WorkflowExcep
         throw new WorkflowDatabaseException(e);
       }
 
+      // get metadata for index update
+      String dcXml = null;
```
staubesv: This code seems worth to be factored out as a utility method as used two times in this file and also in https://github.com/opencast/opencast/pull/487/files#diff-9593c210ac7711725ecfed6cfefc9039R297

staubesv (changes requested):

modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
lkiesow: >  the workflow instance is not the data source for the event (the "archive" is).

I'd like that to be true, but it isn't necessarily. There are some edge cases… It would also not  be sufficient to only update the workflow status. At least not right now. We can discuss that tomorrow if you need more information.

Now, the `isTerminated()` was actually a quite recent addition (commit e26c446953) as a workaround for the issue that data are requested in this message handler and the likelihood of this failing (remember that this is executed asynchronously) was very high compared to other message handling, were it was still a very real possibility.

Now, with the data already included in the message, there wasn't a real reason for this exception any longer.

modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
lkiesow: Makes sense → updated.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
lkiesow: Makes sense → updated.

modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowMessageReceiverImpl.java
lkiesow: Good catch → fixed

staubesv (approved):

PR #490, opened 2018-10-15
MH-13167, Republishing metadata does not update all metadata

lkiesow:
The merge option of the `publish-engage` operation is used for
republishing metadata like dublincore catalogs or access control lists.

Merge works in a way that every previously published element which is
not in the new catalog is included in the merged media package. Or in
other words, only newly published catalogs are updated.

This behavior, while technically well defined and implemented correctly,
causes severe usability issues in some edge cases since a metadata
update can include an unexpected removal of an element.

Example 1 (Series Catalog)
--------------------------

Given an episode which is assigned to a series. Both dublincore/episode
and dublincore/series are published to engage. The user updates the
metadata in the admin interface, removes the series and republishes the
metadata.

Since the episode does not contain a series catalog any longer, the
element is not published which causes the merge to republish the old one
(a new catalog is not published after all). Hence you end up with a
published episode without a series which still has a series catalog.

Example 2 (ACL)
---------------

Given an episode with a published episode ACL. A user now removes the
episode ACL, modifies the series ACL and republishes the metadata to
update the access status. Note that switching from episode to series ACL
may happen automatically in some cases.

The publication of the episode ACL is not updated and the merge option
will republish the old access rules which override the less specific
series ACL. Hence the new access rules have no effect at all.

The Fix
-------

This patch introduces a new merge-force-flavor option which can be used
to enforce the update of certain catalogs. By default this is done for
the dublincore catalogs and the ACLs which are usually part of a
metadata republication. This ensures the metadata are updated even when
a catalog is removed.

The default is set in a way that users will usually not need to specify
any other value. Additional values may be set for extended metadata
catalogs (for which the default does not make things worse though) or in
the case that someone republishes parts of the metadata which is rather
unlikely.

### DIFF
--- a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
+++ b/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
@@ -75,6 +75,7 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -84,7 +85,7 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.UUID;
-
+import java.util.stream.Collectors;
 
 /**
  * The workflow definition for handling "engage publication" operations
@@ -109,6 +110,9 @@ public class PublishEngageWorkflowOperationHandler extends AbstractWorkflowOpera
   private static final String STREAMING_TARGET_SUBFLAVOR = "streaming-target-subflavor";
   private static final String CHECK_AVAILABILITY = "check-availability";
   private static final String STRATEGY = "strategy";
+  private static final String MERGE_FORCE_FLAVORS = "merge-force-flavors";
+
+  private static final String MERGE_FORCE_FLAVORS_DEFAULT = "dublincore/*,security/*";
 
   /** The default path to the player **/
   protected static final String DEFAULT_PLAYER_PATH = "/engage/ui/watch.html";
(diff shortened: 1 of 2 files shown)


modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
```diff
@@ -693,14 +702,20 @@ protected MediaPackage merge(MediaPackage mediaPackageForSearch) throws Workflow
...
 
     MediaPackage mergedMediaPackage = (MediaPackage) updatedMp.clone();
     for (MediaPackageElement element : publishedMp.elements()) {
       String type = element.getElementType().toString().toLowerCase();
       if (updatedMp.getElementsByFlavor(element.getFlavor()).length == 0) {
+        if (forceFlavors.stream().anyMatch((f) -> element.getFlavor().matches(f))) {
+          logger.info("Forcing update of {} {} due to its flavor {}", type, element.getIdentifier(),
+                  element.getFlavor().toString());
+          continue;
+        }
```
karendolan, line 718: @lkiesow Can you clarify in the log message that this is "Forcing removal of {} {} due to ...." instead of the "Forcing update". The continue here omits this existing published element from being part of the merged updated mediapackage.

karendolan:
Hi @lkiesow  this looks like a good solution for removal based on the absence of "dublincore/*,security/*" flavored element without having to retract everything and republish from scratch.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
lkiesow: Makes sense. I've tried to clarify that the update is a removal. The message should now read somewhat like this:

> Forcing removal of attachment 153eff86-1578-4567-a8a4-174d915c2f94 due to the absence of a new element with flavor security/series+xacml

lkiesow:
Updated the log message and also added the missing documentation.
@karendolan, will you take the review?

karendolan:
Looks good to me. Removes force items as advertised.

karendolan:
@lkiesow can this still go into 5x?

lkiesow:
I would say it can since it is a bugfix.
@karendolan, any reason why you just closed this pull request without merging?
Did you just hit the “Close and comment” by accident?
Or is there something wrong with this patch?

karendolan:
@lkiesow H, I hit close and comment by accident! I see It can be re-openned via this comment...

lkiesow:
I nearly did that as well a few times ;-)
Thanks for the review.

PR #491, opened 2018-10-16
MH-13168 Testcases to test a captureagent with Opencast integration.

CGreweling:
A small documentation with test cases for captureagents with Opencast Integration.
The aim is to have a documentation, where anybody can test if a capture agent for opencast full fills the minimum requirements for a reliable device.


lkiesow (approved):

gregorydlogan:
Lots more testcases need to be added, but part of that is also updating the CA API itself so let's keep this as is :)

PR #493, opened 2018-10-16
MH-13172, Simplify Flavor Matching

lkiesow:
This patch just slightly simplifies the rules involved in matching two
media package element flavors.

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
@@ -399,14 +399,13 @@ public class MediaPackageElementFlavor implements Cloneable, Comparable<MediaPac
     if (subtype == null) {
       if (other.subtype != null && !isWildcard(other.subtype))
         return false;
-    } else if (!subtype.equals(other.subtype) && (!isWildcard(subtype) && !isWildcard(other.subtype)))
+    } else if (!subtype.equals(other.subtype) && !isWildcard(subtype) && !isWildcard(other.subtype))
       return false;
     if (type == null) {
-      if (other.type != null && !isWildcard(other.type))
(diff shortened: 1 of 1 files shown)


staubesv:
@lkiesow Somewhat out-of-scope of this PR, but: Do we really support flavors with type == null or subtype == null? The logic would get quite a bit simpler without that cases... 

... just saw that all the code handles this. Will continue the review... ;-)

lkiesow:
No idea but do we really want to risk changing that? :-D
I mean, we could. The release of OC7 is still a long time ahead.

staubesv:
I think we are lucky: 

https://github.com/opencast/opencast/blob/a5772303a92a7cb495a9b9d741037a85f0c590de/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java#L85

and

https://github.com/opencast/opencast/blob/a5772303a92a7cb495a9b9d741037a85f0c590de/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java#L98

lkiesow:
Good catch

lkiesow:
Unfortunately, if you look a few lines below, you find a constructor which allows you to have a null value:

https://github.com/opencast/opencast/blob/a5772303a92a7cb495a9b9d741037a85f0c590de/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java#L117-L118

Hence, you would need to start checking if this isn't used anywhere which for me is outside of the scope of this simple fix. If you can ensure that you could simplify the whole class quite a bit though.

staubesv:
Ok, will continue the review without the additional simplification.

staubesv:
@lkiesow I could not resists... see #498 ;-)

lkiesow:
I'm closing this in favor of pull request #498

PR #494, opened 2018-10-17
MH-13173, Do not hardcode value of ACL override

lkiesow:
Changing a series ACL in the admin interface will replace all access
rules set in the series' episodes. This behavior is hard-coded in the
logic for handling series ACL changes in the admin interface.

From there, the value is passed to the connection logic to the series
access REST endpoint where it is overwritten again which does not really
make any sense since we wouldn't need that parameter in the first place
if we just replace it again at that point.

This patch removes the second hard-coding of this value and respects the
value of the parameter passed to the logic instead.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/seriesAccessResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/seriesAccessResource.js
@@ -42,7 +42,7 @@ angular.module('adminNg.resources')
 
         return $.param({
           acl      : angular.toJson({acl: data.acl}),
-          override : true
+          override : data.override
         });
       }
     }


KatrinIhler:
The commit message is kind of misleading, no? I mean we're still hardcoding the override, we're just not doing it twice...

KatrinIhler:
The commit message is kind of misleading, no? I mean we're still hardcoding the override, we're just not doing it twice...

KatrinIhler:
The commit message is kind of misleading, no? I mean we're still hardcoding the override, we're just not doing it twice...

KatrinIhler:
The commit message is kind of misleading, no? I mean we're still hardcoding the override, we're just not doing it twice...

lkiesow:
The commit message clearly states that it's only removing one of the hard-coded values:

> This patch removes the second hard-coding of this value and respects the
> value of the parameter passed to the logic instead.

But I will try adjusting the title as well.

lkiesow:
The commit message clearly states that it's only removing one of the hard-coded values:

> This patch removes the second hard-coding of this value and respects the
> value of the parameter passed to the logic instead.

But I will try adjusting the title as well.

lkiesow:
Added a `twice` to the tile.

PR #497, opened 2018-10-17
MH-13155: Make weekday preselection optional

jaysuz:
When scheduling multiple events, this will make the preselection of the day of the week optional by introducing the ```prop.admin.event.new.preselect_day``` property.

Resolves: [MH-13155](https://opencast.jira.com/browse/MH-13155)

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
@@ -512,7 +512,14 @@ angular.module('adminNg.services')
             minute: parseInt(endDateTime.format('mm'))
           };
 
-          defaults.presentableWeekdays = chosenSlot.format('dd');
+
+          if (!orgProperties.hasOwnProperty('admin.event.new.preselect_day')
+            || orgProperties.hasOwnProperty('admin.event.new.preselect_day')
+            && orgProperties['admin.event.new.preselect_day'] === true) {
+            defaults.presentableWeekdays = chosenSlot.format('dd');
+          } else {
+            defaults.presentableWeekdays = '';
+          }
 
           if (self.captureAgents.length === 0) {
             //No capture agents, so user can only upload files
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
@@ -512,13 +512,10 @@ angular.module('adminNg.services')
             minute: parseInt(endDateTime.format('mm'))
           };
 
-
-          if (!orgProperties.hasOwnProperty('admin.event.new.preselect_day')
-            || orgProperties.hasOwnProperty('admin.event.new.preselect_day')
-            && orgProperties['admin.event.new.preselect_day'] === true) {
-            defaults.presentableWeekdays = chosenSlot.format('dd');
-          } else {
+          if (orgProperties['admin.event.new.preselect_day'] === 'false') {
             defaults.presentableWeekdays = '';
+          } else {
+            defaults.presentableWeekdays = chosenSlot.format('dd');
           }
 
           if (self.captureAgents.length === 0) {
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -403,6 +403,13 @@ prop.admin.shortcut.general.main_menu=m
 # Note: prop.admin.event.new.duration need not match prop.admin.event.new.interval;
 #       this is to allow short intermissions between consecutive events.
 
+# Preselect an appropriate weekday when scheduling multiple events.
+#
+# Format: Boolean
+# Default: true
+#
+#prop.admin.event.new.preselect_day=true
+
 #
 # Display durations for different notification types in SECONDS.
 #


modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
```diff
@@ -512,7 +512,14 @@ angular.module('adminNg.services')
             minute: parseInt(endDateTime.format('mm'))
           };
 
-          defaults.presentableWeekdays = chosenSlot.format('dd');
+
+          if (!orgProperties.hasOwnProperty('admin.event.new.preselect_day')
+            || orgProperties.hasOwnProperty('admin.event.new.preselect_day')
```
lkiesow: This is always `true` and can be removed, can it not?

lkiesow (changes requested):

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
```diff
@@ -512,7 +512,14 @@ angular.module('adminNg.services')
...
+            && orgProperties['admin.event.new.preselect_day'] === true) {
```
lkiesow: Are you sure this works? Isn't the value you get a string?

lkiesow: Thinking about this, wouldn't this single condition work as well (assuming it's really a boolean value)::

```js
if (orgProperties['admin.event.new.preselect_day'] === false) {
  defaults.presentableWeekdays = '';
} else {
  defaults.presentableWeekdays = chosenSlot.format('dd');
}
```

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
jaysuz: Please note that the boolean expression goes on, and since the following AND binds stronger than the preceding OR the entire expression is not always ```true```. One could, however, drop the second check for the wanted property because the second part of the OR clause is only evaluated if the existence of the wanted property is already established.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
jaysuz: When I tested this, it did work as expected. To be honest, though, I did not look into how these property files get parsed or in what form the parsed content ends up in the frontend.

Your proposed changes would not fulfil the contract of the ```'admin.event.new.preselect_day'``` property having to be ```true``` by default. (All of this is based on the premise that there is a scenario in which the desired property is not set in the configuration file, e.g. commented out.)

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
lkiesow: I'm not talking about the whole condition but about this line.

Assuming `orgProperties.hasOwnProperty('admin.event.new.preselect_day')` evaluates to `false`, you end up with `if (true || …) {…}` at which point the second part of the condition does not matter any longer. The whole thing will evaluate to `true`.

Assuming `orgProperties.hasOwnProperty('admin.event.new.preselect_day')` evaluates to `true`, you end up with `if (false || true && …) {…}`. Hence, if it's necessary for the evaluation of the whole condition to evaluate the second part, the `orgProperties.hasOwnProperty('admin.event.new.preselect_day')` is always `true` making it completely unnecessary.

This makes your statement identical to:

```js
if (!orgProperties.hasOwnProperty('admin.event.new.preselect_day')
    || orgProperties['admin.event.new.preselect_day'] === true) {
```

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
lkiesow: You might want to verify again that it works. I've now actually tested this and it behaves exactly as I suspected. Setting the value to `true` in the organization configuration file will give you:

```js
orgProperties['admin.event.new.preselect_day'] = "true";
```

With `"true" === true` evaluating to `false` this will cause the second part of the condition to always evaluate to `false` (actually regardless of the value since it is always a string). Hence, your code is equivalent to:

```js
if (!orgProperties.hasOwnProperty('admin.event.new.preselect_day')) {
  defaults.presentableWeekdays = chosenSlot.format('dd');
} else {
  defaults.presentableWeekdays = '';
}
```

Now, what you want is a default of `true` for `admin.event.new.preselect_day` meaning that you want to execute `defaults.presentableWeekdays = chosenSlot.format('dd');` if the key is set to `true`, it it is not defined or if it is set to a weird value (throwing an error would also be acceptable in that case but if it is set to a specific value, it should be the default value).

That are a lot of conditions and it's way easier to turn things around and handle the case it is set to `false` instead, bringing us back to my original code suggestion though this time we use a string comparison since we now know that the values are indeed strings:

```js
if (orgProperties['admin.event.new.preselect_day'] === 'false') {
  defaults.presentableWeekdays = '';
} else {
  defaults.presentableWeekdays = chosenSlot.format('dd');
}
```

Let's evaluate how this is executed given different configurations:
- if the key was not configured, `orgProperties['admin.event.new.preselect_day']` should be `undefined`. `undefined === "false"` evaluates to `false` and we end up in `else`. That's what we want for the default.
- if the key was configured to be `true`, `orgProperties['admin.event.new.preselect_day']` should be `"true"`. `"true" === "false"` evaluates to `false` and we end up in `else`. That's what we want.
- if the key was configured to be `false`, `orgProperties['admin.event.new.preselect_day']` should be `"false"`. `"false" === "false"` evaluates to `true` and we end up in the first case. That's what we want.
- if the key was configured to be `rubbish`, `orgProperties['admin.event.new.preselect_day']` should be `"rubbish"`. `"rubbish" === "false"` evaluates to `false` and we end up in `else`. That's the default case and a valid option for invalid input. An alternative would be to throw an error instead, not accepting invalid values.

One additional thing you could consider would be to also accept upper-/mixed-case boolean values. I don't mind either way. And btb. you should check that my code works if you want to use it since I did not test it ;-P

In any case: The current state is broken and this needs to be changed.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
jaysuz: I'm sorry, I thought you were talking about the whole boolean expression. Of course, the second check is obsolete.

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/source.js
jaysuz: Thank you for your lengthy explanation. I now realise that I made some hasty assumptions that apparently turned out to be wrong 😅I will go ahead and test this properly.

jaysuz:
@lkiesow, I have implemented your suggestion. After testing every imaginable scenario, I can confirm that in all of these scenarios we now get the desired result.

lkiesow:
Merged into `develop`.
Thanks for the patch.

PR #498, opened 2018-10-17
MH-13174 Simplify class MediaPackageElementFlavor

staubesv:
- Make default constructor private to ensure that type and subtype are never null
- Don't handle the cases type == null and subtype == null to simplify the boolean expressions
- Remove description and related functionality since unused
- Remove the ElementTypeEquivalent logic as unused

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
@@ -68,14 +68,12 @@ public class MediaPackageElementFlavor implements Cloneable, Comparable<MediaPac
   }
 
   /**
-   * Creates a new element type with the given type and subtype.
+   * Creates a new flavor with the given type and subtype.
    *
    * @param type
-   *          the major type
+   *          the type of the flavor
    * @param subtype
-   *          minor type
-   * @param description
-   *          an optional description
+   *          the subtype of the flavor
    */
   public MediaPackageElementFlavor(String type, String subtype) {
     this.type = checkPartSyntax(type);
@@ -115,24 +113,24 @@ public class MediaPackageElementFlavor implements Cloneable, Comparable<MediaPac
   }
 
   /**
-   * Returns the major type of this element type. Major types are more of a technical description.
-   * The major type is never null.
+   * Returns the type of this flavor.
+   * The type is never <code>null</code>.
    * <p>
-   * For example, if the element type is a presentation movie which is represented as <code>presentation/source</code>,
-   * this method will return <code>track</code>.
+   * For example, if the type is a presentation movie which is represented as <code>presentation/source</code>,
+   * this method will return <code>presentation</code>.
    *
-   * @return the type
+   * @return the type of this flavor
    */
   public String getType() {
     return type;
   }
 
   /**
-   * Returns the minor type of this element type. Minor types define the meaning.
-   * The minor type is never null.
+   * Returns the subtype of this flavor.
+   * The subtype is never <code>null</code>.
    * <p>
-   * For example, if the element type is a presentation movie which is represented as <code>presentation/source</code>,
-   * this method will return <code>presentation</code>.
+   * For example, if the subtype is a presentation movie which is represented as <code>presentation/source</code>,
+   * this method will return <code>source</code>.
    *
    * @return the subtype
    */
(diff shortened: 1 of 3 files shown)


modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
```diff
@@ -81,10 +77,9 @@
    * @param description
```
lkiesow: This parameter does not exist any longer and breaks `mvn site`

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
```diff
@@ -392,52 +254,27 @@ public MediaPackageElementFlavor unmarshal(String str) throws Exception {
...
+    return part1.equals(part2) || isWildcard(part1) || isWildcard(part2);
   }
 
   @Override
   public int hashCode() {
     final int prime = 31;
     int result = 1;
-    result = prime * result + ((subtype == null) ? 0 : subtype.hashCode());
-    result = prime * result + ((type == null) ? 0 : type.hashCode());
+    result = prime * result + subtype.hashCode();
```
lkiesow: We don't even need to initialize this before:

```suggestion
    int result = prime + subtype.hashCode();
```

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
```diff
@@ -392,52 +254,27 @@ public MediaPackageElementFlavor unmarshal(String str) throws Exception {
...
-    } else if (!subtype.equals(other.subtype))
-      return false;
-    if (type == null) {
-      if (other.type != null)
-        return false;
-    } else if (!type.equals(other.type))
-      return false;
-    return true;
+  public boolean equals(Object other) {
+    return (other != null) && (other instanceof MediaPackageElementFlavor)
```
lkiesow: `instanceof` should already take care of the null check:

```suggestion
    return other instanceof MediaPackageElementFlavor
```

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
lkiesow: Thinking about it again, the whole method is just:

```java
public int hashCode() {
￼     return 961 + subtype.hashCode() + type.hashCode();
}
```

lkiesow (changes requested):
Appreciate the pull request.
Everything except the `mvn site` build error caused is purely optional.

lkiesow:
Small addition:

https://github.com/opencast/opencast/blob/e6998f2153224084aef7282c6492f73d7be8559a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java#L250-L253

This could also be just

```java
return parseFlavor(str); 
```

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
staubesv: Yep, oversight...

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
staubesv: But this doesn't explain anymore that a prime is used to build the hash code... the code should be easy to understand - not necessarily as short as possible.

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
staubesv: Thanks, didn't know that.

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
lkiesow: True but irrelevant because it does not really matter that it's a prime number, does it?
In both cases, it's basically only a randomly chosen magic number.

But if you want to keep it, maybe:

```java
final int prime = 31;
return prime * prime + subtype.hashCode() + type.hashCode();
```

Because the `result` variable does not make it easier to read.

lkiesow: Shortly looked into this. With Java7+, this seems to be what we really want to use instead of writing our own code (sorry for all the updates):

```java
return Objects.hash(subtype, type);
```

staubesv: I did some quick research but it does not seem 100% clear whether Objects.hash is a good choice.

Maybe it is, but I don't really have time to look into this. While this code could be further simplified, it is not as terrible as some other code here was...

staubesv:
@lkiesow 

I've updated the PR and did some further simplifications. Note that there would be more desirable work to do, e.g. removing the weird functional-style parseFlavor method (how many functional libraries does Opencast have?) which is just used by two WOHs. Probably the format() in the log could also be removed..

... anyway, already spent too much time on this and while there is more room for simplification, this pull request does simplify the class quite a bit ;-)

lkiesow:
Thanks for the patch. Merged.

PR #501, opened 2018-10-18
MH-13177, Further Simplify MediaPackageElementFlavor

lkiesow:
This extends the work pull request #493 and pull request #498 started,
adding a few more simplifications of the media package elements flavor
class.


staubesv:
I we continue with the simplifications, we end up having the class not at all anymore ;-)

Thanks for the follow-up patch.

staubesv (approved):

lkiesow:
To be honest, I did shortly look at `parseFlavor` and the operations that were using it. But I stopped when I realized that you could overhaul that code quite a bit as well ;-D

staubesv:
Note that you could just re-use MH-13174: JIRA will list all pull request that relate to a ticket, so no problem having multiple pull requests for one issue as long as going into the same release. Would also be better for the change log as mentioning "Simplify MediaPackageElementFlavor" once is enough.

PR #503, opened 2018-10-18
MH-13179, Fix Video Editor Preview Mode Default

lkiesow:
The documentation claims the default for the video editor preview mode
setting to be `true` which is correct if it is not set at all. However,
if an incorrect value is set, the default incorrectly switches over to
`false`


staubesv:
You could argue that an incorrectly set configuration just yields in undefined behavior as a default usually is meant to be the value that is set if no value has been configured (vs. the value that is set if the configured value is invalid).

Anyway, I'm fine with this.

staubesv (approved):

PR #505, opened 2018-10-22
MH-13183 Add link to series details, out of the eventstable-view

CGreweling:
It would be nice to get directly to the details of a series out of the eventstable-view.
As it is now you have to open the Series tab and use the search field to get to the details.

series_link


CGreweling:
@staubesv Do you have any questions to this. I am not sure if this icon is the right one at the right place. May be it should be moved to the actions part with another icon in terms of ui styling guide ?

staubesv:
@CGreweling Sorry, haven't yet had a closer look... no questions. Will start review today.

staubesv (changes requested):
I've found two issues:

1. The symbol should not be shown if no series has been assigned to an event (or at least be inactive and greyed out). Currently, it opens an empty series details modal
2. In case of long titles, the line break seems suboptimal to me: The icon gets an own line (see screenshot)

screen shot 2018-10-29 at 11 48 43

On the "matter of taste"-side of life, I've found the following "things" that might be worth being tried out:

- I would rather expect an action item on the right side of the series it refers to

Maybe the following could be worth a try: Introduce an additional column (no border lines, just to ensure proper layout) for the series action icon. The icon could then be aligned right (horizontal) and middle (vertical). That would also address the somewhat weird line break but would require some attention if no event displayed has a series assigned. But that is just an idea...
It might also be worth trying out a "grey" icon (blue on mouse-over) to not distract the user visually from the data being displayed, but also just an idea...

Besides that, we had the "direct link to series detail" on our wishlist, too, and I'm happy to see it being addressed ;-)

CGreweling:
@staubesv The Icon is now only shown when there is Series for an event.
I have now added a new Icon and moved to the "actions Cell". This will keep the established look and feel of the admin ui.
Besides adding this icon next to the series cell in a good looking way, as you suggest is a good idea.
But not easy possible. To get this done a lot of css have to be rewritten.

I think this is the best way.
seriesdetials

staubesv (approved):

PR #515, opened 2018-10-24
NOJIRA: Documentation: wait_timeout should be bigger than max.idle.time

Rillke:
The MySQL `wait_timeout` should be bigger than `org.opencastproject.db.jdbc.pool.max.idle.time`. This especially not true if you use the mariadb Docker image from Docker Hub leaving the default values in place. This wasn't causing any harm in the OC 2.x releases but in OC 5.x I found it to be reason for aborted workflow instances.


lkiesow:
Thanks for the patch.

By the bye, seeing you changing the title multiple times: You don't have to have NOJIRA or Documentation in your title. Though you may if you want to ;-)

PR #518, opened 2018-10-24
MH-13054: Update angular-ui-sortable, adapting build pipeline

jaysuz:
This will update angular-ui-sortable to the highest version that is still compatible with the jQuery version in use. By doing this via bower, any direct dependencies will be added via bower as well to
make use of the build pipeline.

As a side-effect, manually injected scripts in the head tag having dependencies that are moved to the end of the body tag will be added to the build pipeline as well, allowing for the load order to stay consistent.

Resolves: [MH-13054](https://opencast.jira.com/browse/MH-13054)


modules/admin-ui/src/test/resources/karma.conf.js
```diff
@@ -6,6 +6,8 @@ module.exports = function (config) {
             '../../main/webapp/scripts/lib/jquery-1.10.2.js',
```
staubesv: @jaysuz Those lines need to be removed

KatrinIhler:
@jaysuz This PR has test failures!

jaysuz:
@KatrinIhler, this is weird because on my local machine the ```grunt test``` task succeeds without running into any problems. I will try to investigate.

JamesUoM:
This appears to work

KatrinIhler:
@JamesUoM In that case, can you do a review and merge this if it's okay?

PR #520, opened 2018-10-25
MH-13190, Factor out JpaGroupRoleProvider JaxRs REST to mitigate load cycle race

karendolan:
-  Move JpaGroupRoleProvider's REST concerns into a separate class to simplify the load of the JpaGroupRoleProvider OSGI service, simply the Group Role REST component, and mitigate the opencast-userdirectory getService load cycle race condition with JaxRsServiceTracker.addingService (/groups).
- Give external-api control of its API response

This pull does not add or remove any existing functionality. However, it forces the admin-ui, external-api, and index service to take resonsibility of their own REST API Response handling from now on.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
@@ -24,12 +24,17 @@ import static com.entwinemedia.fn.data.json.Jsons.arr;
 import static com.entwinemedia.fn.data.json.Jsons.f;
 import static com.entwinemedia.fn.data.json.Jsons.obj;
 import static com.entwinemedia.fn.data.json.Jsons.v;
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_CONFLICT;
+import static javax.servlet.http.HttpServletResponse.SC_CREATED;
 import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
+import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
 import static org.apache.commons.lang3.StringUtils.join;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getMessage;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
-import static org.apache.http.HttpStatus.SC_CONFLICT;
 import static org.opencastproject.util.doc.rest.RestParameter.Type.STRING;
 
 import org.opencastproject.external.common.ApiResponses;
@@ -74,7 +79,6 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.Status;
 
 @Path("/")
 @RestService(name = "externalapigroups", title = "External API Groups Service", notes = "", abstractText = "Provides resources and operations related to the groups")
@@ -169,7 +173,8 @@ public class GroupsEndpoint {
     } catch (UnauthorizedException e) {
       return Response.status(SC_FORBIDDEN).build();
     } catch (Exception e) {
-      throw new WebApplicationException(e);
+      logger.error("Unable to delete group {}", id, e);
+      throw new WebApplicationException(e, SC_INTERNAL_SERVER_ERROR);
     }
   }
 
@@ -191,7 +196,7 @@ public class GroupsEndpoint {
       indexService.updateGroup(id, name, description, roles, members);
     } catch (IllegalArgumentException e) {
       logger.warn(e.getMessage());
-      return Response.status(Status.BAD_REQUEST).build();
+      return Response.status(SC_BAD_REQUEST).build();
     } catch (UnauthorizedException ex) {
       return Response.status(SC_FORBIDDEN).build();
     }
@@ -206,8 +211,8 @@ public class GroupsEndpoint {
           @RestParameter(name = "description", description = "Group Description", isRequired = false, type = STRING),
           @RestParameter(name = "roles", description = "Comma-separated list of roles", isRequired = false, type = STRING),
           @RestParameter(name = "members", description = "Comma-separated list of members", isRequired = false, type = STRING) }, reponses = {
-                  @RestResponse(description = "A new group is created.", responseCode = HttpServletResponse.SC_CREATED),
-                  @RestResponse(description = "The request is invalid or inconsistent.", responseCode = HttpServletResponse.SC_BAD_REQUEST) })
+                  @RestResponse(description = "A new group is created.", responseCode = SC_CREATED),
+                  @RestResponse(description = "The request is invalid or inconsistent.", responseCode = SC_BAD_REQUEST) })
   public Response createGroup(@HeaderParam("Accept") String acceptHeader, @FormParam("name") String name,
           @FormParam("description") String description, @FormParam("roles") String roles,
           @FormParam("members") String members) {
@@ -215,13 +220,13 @@ public class GroupsEndpoint {
       indexService.createGroup(name, description, roles, members);
     } catch (IllegalArgumentException e) {
       logger.warn(e.getMessage());
-      return Response.status(Status.BAD_REQUEST).build();
+      return Response.status(SC_BAD_REQUEST).build();
     } catch (UnauthorizedException e) {
       return Response.status(SC_FORBIDDEN).build();
     } catch (ConflictException e) {
       return Response.status(SC_CONFLICT).build();
     }
-    return Response.status(Status.CREATED).build();
+    return Response.status(SC_CREATED).build();
   }
 
   @POST
@@ -230,9 +235,9 @@ public class GroupsEndpoint {
   @RestQuery(name = "addgroupmember", description = "Adds a member to a group.", returnDescription = "", pathParameters = {
           @RestParameter(name = "groupId", description = "The group id", isRequired = true, type = STRING) }, restParameters = {
                   @RestParameter(name = "member", description = "Member Name", isRequired = true, type = STRING) }, reponses = {
-                          @RestResponse(description = "The member was already member of the group.", responseCode = HttpServletResponse.SC_OK),
-                          @RestResponse(description = "The member has been added.", responseCode = HttpServletResponse.SC_NO_CONTENT),
-                          @RestResponse(description = "The specified group does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
+                          @RestResponse(description = "The member was already member of the group.", responseCode = SC_OK),
+                          @RestResponse(description = "The member has been added.", responseCode = SC_NO_CONTENT),
+                          @RestResponse(description = "The specified group does not exist.", responseCode = SC_NOT_FOUND) })
   public Response addGroupMember(@HeaderParam("Accept") String acceptHeader, @PathParam("groupId") String id,
           @FormParam("member") String member) {
     try {
@@ -247,7 +252,7 @@ public class GroupsEndpoint {
                     StringUtils.join(group.getRoles(), ","), StringUtils.join(group.getMembers(), ","));
           } catch (IllegalArgumentException e) {
             logger.warn(e.getMessage());
-            return Response.status(Status.BAD_REQUEST).build();
+            return Response.status(SC_BAD_REQUEST).build();
           } catch (UnauthorizedException ex) {
             return Response.status(SC_FORBIDDEN).build();
           }
@@ -291,7 +296,7 @@ public class GroupsEndpoint {
                     StringUtils.join(group.getRoles(), ","), StringUtils.join(group.getMembers(), ","));
           } catch (IllegalArgumentException e) {
             logger.warn(e.getMessage());
-            return Response.status(Status.BAD_REQUEST).build();
+            return Response.status(SC_BAD_REQUEST).build();
           } catch (UnauthorizedException ex) {
             return Response.status(SC_FORBIDDEN).build();
           }
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
@@ -24,9 +24,12 @@ import static com.entwinemedia.fn.data.json.Jsons.arr;
 import static com.entwinemedia.fn.data.json.Jsons.f;
 import static com.entwinemedia.fn.data.json.Jsons.obj;
 import static com.entwinemedia.fn.data.json.Jsons.v;
+import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static org.apache.commons.lang3.StringUtils.join;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getMessage;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
+import static org.apache.http.HttpStatus.SC_CONFLICT;
 import static org.opencastproject.util.doc.rest.RestParameter.Type.STRING;
 
 import org.opencastproject.external.common.ApiResponses;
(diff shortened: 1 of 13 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -166,7 +167,6 @@
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.WebApplicationException;
```
staubesv: The imports javax.ws.rs.WebApplicationException and  javax.ws.rs.core.Response.Status are only necessary because the method getGroups throws a WebApplicationException. Both imports are used only on this exact line: https://github.com/opencast/opencast/blob/4ba610ea7170617255cb512a30219a3473fe6804/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java#L1403

If the method getGroups would throw an IllegalArgumentException instead, the javax.ws.* imports could be removed. That would require the clients to handle the exception themselves... quite similar to the other changes you do here.

What do you think?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -81,33 +68,10 @@
...
-import javax.ws.rs.Produces;
-import javax.ws.rs.QueryParam;
-import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.core.MediaType;
-import javax.ws.rs.core.Response;
-import javax.ws.rs.core.Response.Status;
 
 /**
- * Manages and locates users using JPA. Note that this also provides a REST endpoint to manage user roles. Since this is
- * not intended to be production code, the REST concerns have not be factored into a separate class. Feel free to
```
staubesv, line 100: OMG, good to know that this was never intended to be production code ;-)

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -81,33 +68,10 @@
...
- * refactor.
+ * Manages and locates users using JPA. The REST concerns have been refactored to the GroupRoleEndpoint
```
staubesv: Suggestion: Just omit "The REST concerns have been refactored to the GroupRoleEndpoint"

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -448,24 +412,35 @@ private boolean like(final String str, final String expr) {
...
 
-  @GET
-  @Produces(MediaType.APPLICATION_JSON)
-  @Path("groups.json")
-  @RestQuery(name = "allgroupsasjson", description = "Returns a list of groups", returnDescription = "Returns a JSON representation of the list of groups available the current user's organization", restParameters = {
-          @RestParameter(defaultValue = "100", description = "The maximum number of items to return per page.", isRequired = false, name = "limit", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING) }, reponses = { @RestResponse(responseCode = SC_OK, description = "The groups.") })
-  public JaxbGroupList getGroupsAsJson(@QueryParam("limit") int limit, @QueryParam("offset") int offset)
+  /**
+   * Retrieves a group list based on input constraints. Called by GroupRoleEndpoint.
```
staubesv: Suggestion: Omit "Called by GroupRoleEndpoint."

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -448,24 +412,35 @@ private boolean like(final String str, final String expr) {
...
 
-  @GET
-  @Produces(MediaType.APPLICATION_XML)
-  @Path("groups.xml")
-  @RestQuery(name = "allgroupsasxml", description = "Returns a list of groups", returnDescription = "Returns a XML representation of the list of groups available the current user's organization", restParameters = {
-          @RestParameter(defaultValue = "100", description = "The maximum number of items to return per page.", isRequired = false, name = "limit", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING) }, reponses = { @RestResponse(responseCode = SC_OK, description = "The groups.") })
-  public JaxbGroupList getGroupsAsXml(@QueryParam("limit") int limit, @QueryParam("offset") int offset)
+  /**
+   * Returns a XML representation of the list of groups available the current user's organization. Used by
```
staubesv: Suggestion: Omit "Used by GroupRoleEndpoint"

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
```diff
@@ -0,0 +1,192 @@
...
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
+
+/**
+ * A REST EndPoint for JpaGroupRoleProvider. This endpoint class factors out the REST concerns of the
```
staubesv: Suggestion: Omit "This endpoint class factors out the REST concerns of the JpaGroupRoleProvider."

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
```diff
@@ -0,0 +1,192 @@
...
+
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("groups.json")
+  @RestQuery(name = "allgroupsasjson", description = "Returns a list of groups", returnDescription = "Returns a JSON representation of the list of groups available the current user's organization", restParameters = {
+          @RestParameter(defaultValue = "100", description = "The maximum number of items to return per page.", isRequired = false, name = "limit", type = RestParameter.Type.STRING),
+          @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING) }, reponses = { @RestResponse(responseCode = SC_OK, description = "The groups.") })
+  public JaxbGroupList getGroupsAsJson(@QueryParam("limit") int limit, @QueryParam("offset") int offset)
+          throws IOException {
+    return jpaGroupRoleProvider.getGroupsAsXml(limit, offset);
```
staubesv: /groups.json uses getGroupsAs**Xml**?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
```diff
@@ -0,0 +1,192 @@
...
+          @RestParameter(defaultValue = "100", description = "The maximum number of items to return per page.", isRequired = false, name = "limit", type = RestParameter.Type.STRING),
+          @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING) }, reponses = { @RestResponse(responseCode = SC_OK, description = "The groups.") })
+  public JaxbGroupList getGroupsAsXml(@QueryParam("limit") int limit, @QueryParam("offset") int offset)
+          throws IOException {
+    return jpaGroupRoleProvider.getGroupsAsXml(limit, offset);
+  }
+
+  @DELETE
+  @Path("{id}")
+  @RestQuery(name = "removegrouop", description = "Remove a group", returnDescription = "Return no content", pathParameters = { @RestParameter(name = "id", description = "The group identifier", isRequired = true, type = Type.STRING) }, reponses = {
```
staubesv: Typo "removegrou**o**p"

modules/userdirectory/src/test/java/org/opencastproject/userdirectory/JpaGroupRoleProviderTest.java
```diff
@@ -159,13 +165,13 @@ public void testUpdateGroupNotAllowedAsNonAdminUser() throws UnauthorizedExcepti
 
     try {
       // try add ROLE_USER
-      Response updateGroupResponse = provider.updateGroup(group.getGroupId(), group.getName(), group.getDescription(),
+      Response updateGroupResponse = endpoint.updateGroup(group.getGroupId(), group.getName(), group.getDescription(),
```
staubesv: In that case, you might want to rename the file and relocate this file to

directory/src/test/java/org/opencastproject/userdirectory/**endpoint**/GroupRole**Endpoint**Test.java

as it now seems to focus more on the endpoint than the provider.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
```diff
@@ -24,9 +24,12 @@
 import static com.entwinemedia.fn.data.json.Jsons.f;
 import static com.entwinemedia.fn.data.json.Jsons.obj;
 import static com.entwinemedia.fn.data.json.Jsons.v;
+import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static org.apache.commons.lang3.StringUtils.join;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getMessage;
 import static org.apache.commons.lang3.exception.ExceptionUtils.getStackTrace;
+import static org.apache.http.HttpStatus.SC_CONFLICT;
```
staubesv: Any special motivation to prefer this over javax.servlet.http.HttpServletResponse.SC_CONFLICT?

Seems a bit weird to grab the constants from different classes...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
```diff
@@ -154,7 +161,16 @@ public Response getGroup(@HeaderParam("Accept") String acceptHeader, @PathParam(
...
-    return indexService.removeGroup(id);
+    try {
+      indexService.removeGroup(id);
+      return Response.noContent().build();
+    } catch (NotFoundException e) {
+      return Response.status(SC_NOT_FOUND).build();
+    } catch (UnauthorizedException e) {
+      return Response.status(SC_FORBIDDEN).build();
+    } catch (Exception e) {
+      throw new WebApplicationException(e);
```
staubesv: I had a quick look at other endpoints of the external API and found that this most often is used like:


```
import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;

[...]
 logger.error("Unable to delete group {}", id, e);
 throw new WebApplicationException(e, SC_INTERNAL_SERVER_ERROR);

```

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
```diff
@@ -67,7 +72,9 @@
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Response.Status;
```
staubesv: Maybe instead import and use

javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST
javax.servlet.http.HttpServletResponse.SC_CREATED

for sake of consistency.

staubesv (changes requested):

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
karendolan: @staubesv  sounds good, I'll add that change.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
karendolan: Right! I kept coming back to this class and that uncomfortable comment when searching down the cycle chain after getting a cycle error message on load.

@staubesv , I added an extra commit to this pull to put immediate=false in the SCR configuration of the new Group REST and the existing textanalyzer REST endpoint class. Both endpoints load as usual with this setting at the same time as all the other Endpoints, via the RestPublisher.JaxRsServiceTracker.  I hunted through a lot of tickets from other projects, FELIX, and apache in general, related to the cycle error. Two comments said changing immediate=true to immediate=false in the REST class mitigated the cycle error. The general idea appears to be that there is an underlying race condition between service injection on the REST class and the JaxRs registration of the REST class at the time of REST class activation.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
karendolan: Thanks for seeing that, I'll fix it to be consistent.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
karendolan: Ironically,  jpaGroupRoleProvider.getGroupsAsJson sends the getGroupsAsXml result. I've been testing the endpoints and the result is in JSON. That "Produces JSON" annotation makes the JaxRs is convert the XML to JSON on the fly...I'll check if it's  easy to combine @Path("groups.{xml|json}) , @Produces({MediaType.APPLICATION_JSON,MediaType.APPLICATION_XML}) without changing the endpoint.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
karendolan: I'll fix, but it's changing the existing API ;)

modules/userdirectory/src/test/java/org/opencastproject/userdirectory/JpaGroupRoleProviderTest.java
karendolan: Yes, It does...  
"Feel free to refactor." But, I'm not feeling free!

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
staubesv: @RestQuery is just for the rest docs, isn't it? That does not actually change the API...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
karendolan: done

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
karendolan: done

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
karendolan: done

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
karendolan: done

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/endpoint/GroupRoleEndpoint.java
karendolan: done

modules/userdirectory/src/test/java/org/opencastproject/userdirectory/JpaGroupRoleProviderTest.java
karendolan: refactored

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
karendolan: done

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
karendolan: done

staubesv (approved):

PR #521, opened 2018-10-25
MH-13191 Improve performance of retrieving groups

pmiddend:
When groups have a large number of group members, the performance of the requests

- GET /admin-ng/groups
- GET /admin-ng/groups/ {group_identifier}

gets quite bad.

This is because of three reasons:

1. For each group and for each user in that group, an SQL query of its own is run. So there’s lots of queries, and apparently there are a number of caches involved to mitigate this already. However, as the
number of users in each group grows bigger, the caches don’t help anymore.
2. For each group, along with the actual users in that group, the users roles are retrieved. However, the roles aren’t actually _used_ in the JSON output of the request. So all that work is for nothing.
3. When deserializing the group output from the search index, the XML unmarshaller isn’t shared between runs.

Consequently, this PR does the following:

1. Add a function to retrieve a list of users (by their names) from the data base in bulk. This is used per-group. Currently supported backends for this are the JPA user provider and the JPA user reference
provider. The other backends behave as before. 
2. Don’t retrieve the users roles anymore.
3. Share the unmarshaller between deserialization runs for the group query.

We saw a performance increase from about 40s (!) to about 1s for a group with 1000 members.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/GroupsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/GroupsEndpoint.java
@@ -73,9 +73,14 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.DELETE;
 import javax.ws.rs.FormParam;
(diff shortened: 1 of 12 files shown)


smarquard:
The description and method names that refer to finding users by "names" is ambiguous and could be understood to mean by full name rather than username.

I would prefer to see "username" used (in appropriate case) in comments, javadoc and method names rather than "name". Username could be omitted from the method name if the parameter type is clear, e.g. change method signatures like this:

    default Iterator<User> findUsersByNames(Collection<String> names) {

to:

    default Iterator<User> findUsers(Collection<String> usernames) {

pmiddend:
@smarquard Thanks for the suggestion, you're right. I've changed all of the names to use "userName" or "username" where appropriate.

PR #522, opened 2018-10-25
MH-13192 Improve performance of list requests

pmiddend:
The following requests are quite slow when asking for a large number of objects:

- GET /admin-ng/events
- GET /api/events
- GET /admin-ng/series
- GET /api/series

An analysis revealed that most of the time is spent creating JAXB Unmarshaller objects for each single item. This can be sped up by reusing the same Unmarshaller between events/series.

Note that this is not thread-safe, of course, but we’re inside a single thread while deserializing, so this should be fine.

We saw the following performance improvement when displaying 100 items per page:

Events table: From approx. 850ms down to approx. 150ms
Series table: From approx. 900ms down to approx. 250ms

So the re-use of JAXB Unmarshaller objects results in a speedup of roughly 4-5.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/AbstractSearchIndex.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/AbstractSearchIndex.java
@@ -87,6 +87,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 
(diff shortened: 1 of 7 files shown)


pmiddend:
Sorry for screwing up the tests, I will fix it immediately.
Edit: fixed.

staubesv:
@pmiddend May redirect this to 6.x?

pmiddend:
@staubesv Thanks, forgot about that.

PR #523, opened 2018-10-25
MH-13193 Improve performance of event deletion (1)

krinnewitz:
When deleting events from the scheduler service, some unnecessary
queries related to properties are executed:

- Since we know, that we are deleting the whole episode, we do not have
to check whether there are snapshots for the properties to be deleted. I
am not sure anyway, why this is necessary at all.

- Since we know, that we take care of deleting all properties related to
the episode, we do not have to remove orphaned properties (because we are
not producing any).

It turned out to be quiet hard to control the behavior of the code
executed in AbstractADeleteQuery.runQueries() and I am not happy with
the flag attached to DeleteQueryContribution. However, I could not find a
better solution in the available time.

All in all, these changes seem to improve event deletion by about 50 or
60%, according to our measurements.

### DIFF
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/ADeleteQuery.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/ADeleteQuery.java
@@ -29,6 +29,13 @@ public interface ADeleteQuery {
   ADeleteQuery name(String queryName);
 
   /**
+   * When this is called with true, the AssetManager will assume that the whole media package will be removed. This is
+   * faster, but then it is your responsibility to remove all properties associated with the media package since no
+   * orphan removal will take place.
+   */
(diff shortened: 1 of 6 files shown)


KatrinIhler (approved):
Alright, I looked into this a bit and as far as I understand it, when deleting a scheduled event, the SchedulerService deletes the properties of its own namespaces and then the IndexService deletes the rest? On the one hand it feels weird to make two queries for that, on the other hand I get that one would want to keep this separated...opinions?
Anyway, all properties get deleted, and if it's definitely faster than before, so yay!

krinnewitz:
Yes @KatrinIhler, that's right. I decided to keep the deletion of properties in the scheduler since it seemed to not significantly impact performance and it makes the SchedulderService's removeEvent() method more "comprehensive". Right now, removeEvent() is only called by the IndexService, but if someone decides to call it from another place, he won't need to take care of the properties.

PR #524, opened 2018-10-25
MH-13193 Improve performance of event deletion (2)

krinnewitz:
The AssetManager used to "manually" control the relation between
snapshots and assets. I.e. assets actually have references to snapshots,
but this relationship was only handled on application level. On database
level, there was not explicitly declared relationship between assets and
snapshots.

So when deleting snapshots, the AssetManager had to first figure out,
which snapshots were going to be deleted by the given query and then
"manually" remove all assets referencing the snapshots to be deleted,
before actually deleting the snapshots.

With this patch, the relationship between assets and snapshots is
modeled explicitly on the database level by changing the shema
accordingly. This allows for ON DELETE CASCADE to be used resulting in
only one query being necessary when removing snapshots.

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -18,7 +18,12 @@ How to Upgrade
 Database Migration
 ------------------
 
-*So far, this is not required*
+As part of performance optimizations, a foreign key constraint was added to one table. This requires a database schema
+update. As with all database migrations, we recommend to create a database backup before attempting the upgrade.
+
+You can find the database upgrade script in `docs/upgrade/6_to_7/`. This script is suitable for both, MariaDB and
+MySQL.
+
 
 ActiveMQ Migration
 ------------------
--- a/docs/scripts/ddl/mysql5.sql
+++ b/docs/scripts/ddl/mysql5.sql
@@ -384,6 +384,7 @@ CREATE TABLE oc_assets_asset (
   size BIGINT NOT NULL,
   storage_id VARCHAR(256) NOT NULL DEFAULT 'local-filesystem',
   --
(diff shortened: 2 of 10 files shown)


docs/guides/admin/docs/upgrade.md
```diff
@@ -18,7 +18,12 @@ How to Upgrade
 Database Migration
 ------------------
 
-*So far, this is not required*
+As part of performance optimizations, a foreign key constraint was added to one table. This requires an database schema
```
KatrinIhler: 'a' instead of 'an'

docs/guides/admin/docs/upgrade.md
```diff
@@ -18,7 +18,12 @@ How to Upgrade
...
+update. As with all database migrations, we recommend to make a database backup before attempting the upgrade.
```
KatrinIhler: 'do' instead of 'make'?

docs/guides/admin/docs/upgrade.md
```diff
@@ -18,7 +18,12 @@ How to Upgrade
...
+
+You can find the database upgrade script in `docs/upgrade/6_to_7/`. This script is suitable for both, MariaDB, and
```
KatrinIhler: "for both MariaDB and MySQL" without the ,

KatrinIhler (changes requested):
I tested the migration script first and that didn't work for me, but it's possible that I skipped another necessary upgrade step before. When I deleted all tables and recreated them with the other script, it worked fine for me.

krinnewitz:
Thanks @KatrinIhler. Which database have you tried? mysql or MariaDB? What error message did you get?

KatrinIhler (approved):
@krinnewitz I use MariaDB. But I just tested this again by deleting all tables, recreating them with the script for 6 and then using the migration script from 6 to 7 and this time it worked, so it was probably my setup from before that was wonky. Lets get to merging! :)

PR #525, opened 2018-10-26
MH-12854, Get latest series when creating a new event

JamesUoM:
Quick fix, by pulling the metadata when the wizard is opened.


staubesv (approved):
While this might have some negativ impact on the performance of opening the modal, correctness seems more important.

PR #527, opened 2018-10-27
MH-13196, Unregister Resource Servlets of Bundles to be Removed

lkiesow:
Trying to replace a bundle with a static resource servelet registered by
Opencast (e.g. runtime-info-ui-ng or admin-ui) will result in an
org.osgi.service.http.NamespaceException when the updates bundle is
registered again since the servlet registered by the first bundle never
got removed and its alias is hence already in use.

Example:

[asciicast](https://asciinema.org/a/208822)

This patch ensures that servlets are unregistered as soon as the
associated bundle is unloaded. This makes it easily possible to
live-deploy a new version of that bundle, e.g. using Karaf's
`bundle:watch` feature without the need for restarting Opencast in the
process.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java
@@ -137,7 +137,7 @@ public class RestPublisher implements RestConstants {
   private Server server;
 
   /** The map of JAX-RS resource providers */
-  private Map<ServiceReference<?>, ResourceProvider> resourceProviders = new HashMap<>();
+  private Map<ServiceReference<?>, ResourceProvider> resourceProviders = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java
```diff
@@ -452,6 +450,8 @@ public URL getResource(String name) {
    */
   class StaticResourceBundleTracker extends BundleTracker<Object> {
 
+    private HashMap<Bundle, ServiceRegistration> servlets = new HashMap<>();
```
staubesv, line 453: What about thread-safety when accessing this data structure?

lkiesow: fixed

PR #528, opened 2018-10-27
MH-13197, Separate Admin Interface Index

lkiesow:
The search index of the admin interface is bundled in the same package
which also contains the REST endpoints and the frontend, resulting in a
re-start of a lot of components if you just want to deploy a simple
front-end fix.

This patch separated the index so that a replacement of the admin
interface bundle does not cause the index to re-start as well making
hot-deployments easier.

Note that commit a52a455 fixes a few problems with updating bundles
which will come handy if this is to be used.

*Work sponsored by SWITCH*


staubesv:
@lkiesow Please note that this PR currently has a merge conflict.

modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
```diff
@@ -13,68 +13,42 @@
     <provide interface="org.osgi.service.cm.ManagedService"/>
   </service>
   <reference bind="setAdminUIConfiguration"
-             cardinality="1..1"
```
JamesUoM, line 16: I'm not sure how changes to the cardinality and policy in this file (and not others) relates to this ticket?

modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
```diff
@@ -13,68 +13,42 @@
...
              interface="org.opencastproject.workflow.api.WorkflowService"
-             name="workflowService"
-             policy="static"/>
+             name="workflowService"/>
   <reference name="jobService"
              interface="org.opencastproject.adminui.endpoint.JobEndpoint"
-             cardinality="1..1"
-             policy="static"
              bind="setJobService"/>
-  <reference name="seriesService"
```
JamesUoM, line 30: Is this fixing a separate issue?

modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
lkiesow: I had to move and modify some of these anyway. The `1..1` is the default and technically unnecessary. I could move some of these changes to a separate pull request but it felt to be unnecessarily complex to do that with parts of them being modified anyway.

modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
lkiesow: This is just the name which I found to be quite confusing since it's not actually the binding of the series service but the admin interfaces series endpoint. Again, this could technically be a separate, one-line pull request.

lkiesow:
Thanks for picking up the review.

KatrinIhler:
@JamesUoM Is this ready for merging or do you still have questions?

PR #529, opened 2018-10-27
MH-13198, Properly Display Multiple Presenters

lkiesow:
If an event has multiple presenters, the main event table will display
them as one string, separated by comma. This can be a real usability
issue if the name format is in the form of “Doe, Jane”, where users will
end up with lines like “Doe, Jane, Doe, John, Mustermann, Max” which is
very confusing.

This patch tries to make the separate metadata entries visually distinct
by containing each entry in its own element, surrounded by a
lightweight border.

### DIFF
--- a/modules/admin-ui/src/main/webapp/styles/extensions/components/_tables.scss
+++ b/modules/admin-ui/src/main/webapp/styles/extensions/components/_tables.scss
@@ -25,6 +25,14 @@
     .main-tbl {
         border-top-left-radius: 0;
         border-top-right-radius: 0;
+
+        .metadata-entry {
+            border: 1px solid #ccc;
+            border-radius: 5px;
+            padding: 0 5px;
+            margin-right: 5px;
+        }
+
     }
 }
 
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventsController.js
@@ -51,7 +51,8 @@ angular.module('adminNg.controllers')
         label: 'EVENTS.EVENTS.TABLE.TITLE',
         sortable: true
       }, {
-        name:  'presenter',
+        template: 'modules/events/partials/eventsPresentersCell.html',
+        name:  'presenters',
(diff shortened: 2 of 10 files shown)


staubesv:
@lkiesow Any reason to not apply that style change to other multi-select fields (e.g. series.organizers, series.publishers, series.presenter)?

In theory, it would be desirable to also do this in the modals and wizards, but that might be out-of-scope for this PR.

staubesv (changes requested):
Should at least be done consistently on the tables - preferably also for the modals and wizards.

modules/admin-ui/src/main/webapp/styles/extensions/components/_tables.scss
```diff
@@ -25,6 +25,14 @@
     .main-tbl {
         border-top-left-radius: 0;
         border-top-right-radius: 0;
+
+        .metadata-entry {
```
staubesv, line 29: Maybe some more precise name for the style here? The table only contains "metadata entries", doesn't it?

lkiesow:
> Should at least be done consistently on the tables - preferably also for the modals and wizards.

That makes sense. I missed that the series table has the contributors and organizers fields. I will update this shortly.

modules/admin-ui/src/main/webapp/styles/extensions/components/_tables.scss
lkiesow: It's meant for a single metadata entry in a collection of metadata after all.
How would you prefer it to be named?

lkiesow:
Updates series table as well.

I deliberately chose to not set this for roles in the user/groups tables. These aren't exactly displayed well as it and due to their special format, I don't think they would benefit from this. They are also a different type of metadata making this clash less with these adjustments.

lkiesow:
Fixed merge conflicts.

modules/admin-ui/src/main/webapp/styles/extensions/components/_tables.scss
lkiesow: @staubesv 
> How would you prefer it to be named?

staubesv (approved):

PR #530, opened 2018-10-29
MH-12332, hide workflows whose tags don't explicitly match the source type

JamesUoM:
Add the workflow tags to the processing.json response. Use ng-options' "disable when" clause to grey out workflows that don't contain a tag relevant to the current source type.

### DIFF
--- a/modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/shared/partials/wizards/new-event.html
+++ b/modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/shared/partials/wizards/new-event.html
@@ -413,10 +413,8 @@
                         data-disable-search-threshold="8"
                         data-search_contains="true"
                         ng-model="wizard.step.ud.workflow"
-                        ng-options="w.title disable when 
-                          ((wizard.getStateControllerByName('source').ud.type==='UPLOAD' && !w.tags.includes('upload-ng')) ||
-                                  (wizard.getStateControllerByName('source').ud.type.startsWith('SCHEDULE') && !w.tags.includes('schedule-ng')))
-                          for (obj, w) in wizard.step.workflows"
+                        ng-options="w.title for (obj, w) in wizard.step.workflows |
+                                  filter: wizard.getStateControllerByName('source').ud.type==='UPLOAD' ? 'upload-ng' : 'schedule-ng'"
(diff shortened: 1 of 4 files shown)


modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/shared/partials/wizards/new-event.html
```diff
@@ -413,7 +413,10 @@
                         data-disable-search-threshold="8"
                         data-search_contains="true"
                         ng-model="wizard.step.ud.workflow"
-                        ng-options="w.title for (obj, w) in wizard.step.workflows"
+                        ng-options="w.title disable when
```
staubesv: Why would you want to display the workflows in the drop-down list if the user cannot select them? Can't they just be omitted in the list?

staubesv (changes requested):
Unclear to me why you don't just omit the workflows that a user is not allowed to select (see question in comment).

modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/shared/partials/wizards/new-event.html
JamesUoM: There is no inbuilt functionality in angular to omit them. It can be done with a filter if required but I was trying to make the fewest changes to the code. (see ticket)

staubesv: ... the filter seems not that hard to implement. If you don't need them "greyed", could you do the filter?

JamesUoM:
managed to format a filter that more easily that I'd thought. You can squash the commits on merge

staubesv (approved):

lkiesow:
Hi @JamesUoM, this patch caused larger merge conflicts when merging 4.x into 5.x which made me revert this as part of the merge. Can you help and create an updated pull request against 5.x for this?

JamesUoM:
> Hi @JamesUoM, this patch caused larger merge conflicts when merging 4.x into 5.x which made me revert this as part of the merge. Can you help and create an updated pull request against 5.x for this?

@lkiesow yeah sure

JamesUoM:
5.x https://github.com/opencast/opencast/pull/567

Sorry I missed 5.2 I was ill.

PR #532, opened 2018-10-30
MH-13194 wrong description for endpoint

tyfranck:
Corrected 'mediapackage' description for /ingest/{wdID} REST endpoint


modules/matterhorn-ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -1044,7 +1044,7 @@ public Response ingest(MultivaluedMap<String, String> formData) {
   @POST
   @Produces(MediaType.TEXT_HTML)
   @Path("ingest/{wdID}")
-  @RestQuery(name = "ingest", description = "Ingest the completed media package into the system, retrieving all URL-referenced files, and starting a specified workflow", pathParameters = { @RestParameter(description = "Workflow definition id", isRequired = true, name = "wdID", type = RestParameter.Type.STRING) }, restParameters = { @RestParameter(description = "The ID of the given media package", isRequired = true, name = "mediaPackage", type = RestParameter.Type.TEXT) }, reponses = {
+  @RestQuery(name = "ingest", description = "Ingest the completed media package into the system, retrieving all URL-referenced files, and starting a specified workflow", pathParameters = { @RestParameter(description = "Workflow definition id", isRequired = true, name = "wdID", type = RestParameter.Type.STRING) }, restParameters = { @RestParameter(description = "The given media package", isRequired = true, name = "mediaPackage", type = RestParameter.Type.TEXT) }, reponses = {
```
KatrinIhler: Would you mind removing the 'given' to make it consistent with /ingest without a workflow id?

KatrinIhler (changes requested):
Definitely a good idea to fix this. The merge commit confuses me though, why is that in here (especially since it doesn't seem to contribute anything)?

modules/matterhorn-ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
tyfranck: Does your comment about this not contributing to anything related to this pull request or the previous one I closed?

KatrinIhler: I'm specifically talking about the first of the two commits "Merge pull request #4 from opencast/r/4.x" - why is that here? What does it do?

lkiesow:
The changes itself look just fine. @tyfranck, do you mind if we just squash-merge this? That will automatically drop the additional merge commits in here. I guess that would be okay with you as well, @KatrinIhler?

KatrinIhler:
@lkiesow Sounds good to me, but Github tells me that squash & merge is not enabled for this repository, so how do we do this?

tyfranck:
@lkiesow  please go ahead. I've found a way to avoid those extra commits. So next time will be a cleaner one... :-)

lkiesow:
@KatrinIhler, found a way to do that.
Sometimes, being an admin helps ;-)

PR #533, opened 2018-10-30
MH-12561 correctly update config UI to reflect scheduled event settings

JamesUoM:
add, custom event to a confField that updates config UI, to reflect the read workflow config
add, trigger change event when indirectly setting  a confField input (hidden) to save the modifed config
fixed, unset checkbox inputs if not set.

### DIFF
--- a/etc/workflows/ng-schedule-and-upload.xml
+++ b/etc/workflows/ng-schedule-and-upload.xml
@@ -103,8 +103,17 @@
             var show = $(this).val() == 'publish';
             $('#publicationConfigurationPanel').toggle(show);
             $('#straightToPublishing').val(show);
+            // Trigger update of the schedule's workflow config
+            $('#straightToPublishing').change();
         });
 
+        <!-- called on loading an exisiting event -->
+        $('#straightToPublishing').bind('updateConfigUI', function() {
+            var publish = $(this).val().toLowerCase() === 'true';
+            $('#publicationConfigurationPanel').toggle(publish);
+            $('#holdOrPublishToggle_' + (publish ? 'publish':'hold')).prop('checked', true);
+        });
+        
         <!-- Remove the line below if you wish to publish to AWS S3 -->
         $('#publishToAws').prop('disabled', true);
         <!-- Remove the line above if you wish to publish to AWS S3 -->
--- a/etc/workflows/ng-schedule-and-upload.xml
+++ b/etc/workflows/ng-schedule-and-upload.xml
@@ -111,7 +111,7 @@
         $('#straightToPublishing').bind('updateConfigUI', function() {
             var publish = $(this).val().toLowerCase() === 'true';
             $('#publicationConfigurationPanel').toggle(publish);
-            $('#holdOrPublishToggle_' + (publish ? 'publish':'hold')).prop('checked', true);
+            $('#holdOrPublishToggle_' + (publish ? 'publish':'hold')).prop('checked', 'checked');
         });
         
         <!-- Remove the line below if you wish to publish to AWS S3 -->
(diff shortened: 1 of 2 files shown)


modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/modules/events/controllers/eventController.js
```diff
@@ -149,11 +149,16 @@ angular.module('adminNg.controllers')
                         if (el.length > 0) {
                             if (el.is('[type=checkbox]') || el.is('[type=radio]')) {
                                 if (value === 'true' || value === true) {
-                                    el.attr('checked','checked');
+                                    el.attr('checked', true);
```
lkiesow: [attr](http://api.jquery.com/attr/#attr2) should just set the value of the attribute and a checked attribute should afaik always set the element to checked regardless of its value. At least according to the specs…

Thus, it would make sense to remove the attribute instead like Opencast already does in recent versions:

https://github.com/opencast/opencast/blob/2d383df990cd8810876b4d6f1f1ec50b60fd16d9/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js#L158-L162

JamesUoM: I'm not sure I follow. Yes setting el.attr('foo',value); works in the main, but for checkboxes and radio button, to unset them (ie the value=false) either you have set value to null or call el.removeAttr('checked');  explictly

` if (value !== 'true' && value !== true ) {
    value = null; 
 }
 el.attr('checked', value)
`

would be an alternate way of writing it I guess?

staubesv: @JamesUoM The attribute "checked" needs to be set to the value "checked" (see https://www.w3schools.com/tags/att_input_checked.asp). There has been recently a fix the removes the attribute "checked" correctly: https://github.com/opencast/opencast/commit/eb7754f25b0d31a8a40eb0e91fffdc16e5f94935#diff-2ed317613602203533b1e42388734725

So the code @lkiesow should be correct but is not yet in 4.x

staubesv (dismissed):
See comments

modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/modules/events/controllers/eventController.js
JamesUoM: Sorry, I was confused I thought I was looking at my code the 6.x's. Yes I'll remove the attr.

staubesv:
Hi @JamesUoM Could you please set the value of the attribute to "checked" instead of "true"?

While the latter works, the value "checked" seems most common. The weird thing about the attribute "checked" is that its presence indicates "true" and its absence indicates "false". Seem from that point of view, checked=checked is equivalent to checked=true.

See

>  In XHTML, attribute minimization is forbidden, and the checked attribute must be defined as \<input checked="checked" \/>.

on https://www.w3schools.com/tags/att_input_checked.asp

etc/workflows/ng-schedule-and-upload.xml
```diff
@@ -103,8 +103,17 @@
...
             $('#straightToPublishing').val(show);
+            // Trigger update of the schedule's workflow config
+            $('#straightToPublishing').change();
         });
 
+        <!-- called on loading an exisiting event -->
+        $('#straightToPublishing').bind('updateConfigUI', function() {
+            var publish = $(this).val().toLowerCase() === 'true';
+            $('#publicationConfigurationPanel').toggle(publish);
+            $('#holdOrPublishToggle_' + (publish ? 'publish':'hold')).prop('checked', true);
```
staubesv: ...prop('checked', 'checked');

modules/matterhorn-admin-ui-ng/src/main/webapp/scripts/modules/events/controllers/eventController.js
staubesv: @JamesUoM Please also revert this change - the original code line is correct.

JamesUoM:
"Changes Requested" - I've no idea what github is referring to as I've closed all conversations AFAIK?

lkiesow:
Yes, you addressed all of the comments. But that does not imply that the general request for changes from @staubesv  is dismissed:

screenshot from 2018-11-23 17-58-53

[Github documentation: Dismissing a pull request review
](https://help.github.com/articles/dismissing-a-pull-request-review/)

staubesv (approved):

PR #535, opened 2018-10-31
MH-13200 Remove unused file acl-modal.html

staubesv:
This file is simply not used at all.


KatrinIhler (approved):
Looks like you're right, let's get rid of this!

PR #540, opened 2018-10-31
MH-13208, Create a short contributor guide

lkiesow:
This patrch tries to create a short contributor guide, linking the
development process documentation but avoiding its lengty explanations.
The guide is in Github's default contributing guideline formnat and
should hence be pointed out by Github:

  https://help.github.com/articles/setting-guidelines-for-repository-contributors/

The guide intentionally tries to avoid mentioning any special cases and
points this out to readers. If necessary, they can still be found in the
developer documentation but most of the time, just sticking to this
guide should be fine.

The guide is meant especially for new, maybe even one-time contributors
which likely would not have read the long documentation before. Using
this, they should be able to quickly get an overview of how to create
proper pull requests for Opencast.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,8 +11,7 @@ Issue Tracker
 -------------
 
 Opencast uses [Jira](https://opencast.jira.com) for tracking issues. Each pull request should be accompanied by a ticket
-in Jira. The issue identifier should also be used in the title of the pull request and the commits. E.g.: `MH-12345,
-Fixing Something Somewhere`.
+in Jira. The issue identifier should also be used in the title of the pull request and the commit message.
 
 
 Bug Fixes and Feature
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+document defining rules and recommendations as well as a detailed explanation of Opencast's branching model, take a look
+at [the development process documentation](https://docs.opencast.org/develop/developer/development-process/). But this
+guide should suffice for most contributions.
+
+
+Issue Tracker
+-------------
+
+Opencast uses [Jira](https://opencast.jira.com) for tracking issues. Each pull request should be accompanied by a ticket
+in Jira. The issue identifier should also be used in the title of the pull request and the commits. E.g.: `MH-12345,
```
staubesv: @lkiesow The comma after the issue number seems somewhat superfluous. What do you think about "MH-12345 Description" instead of "MH-12345, Description"?

lkiesow: This is the exact example from the development process docs. Also, to be honest, I like to have a separator between the ticket identifier and the summary. These are two different parts after all. But then, this is also just an example. If you like me to, I can remove it though, to just leave the choice to the contributor.

staubesv: No strong opinion from my side, just wanted to mention it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+When building Opencast, a set of unit tests is automatically run on the code. Passing these tests is a hard requirement.
+These (and a few more) tests are also run automatically on our CI system. Again, all tests need to pass.
+
+If the CI tests on your pull request fail and you are sure it is not your patch that did break it, please complain.
+Errors happen and committers can easily trigger a new build. But your patch will not be merged without these tests
+passing.
+
+Additionally, a reviewer will be assigned to your pull request to ensure that there are no further issues. Once
+everything is fine, the reviewer will merge the pull request.
```
gregorydlogan: "It may take some time before a review is assigned.  This is normal, so do not be concerned." ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+Checklist
+---------
+
+- [ ] Jira ticket created
+- [ ] Pull request has a proper title and description
```
gregorydlogan, line 51: Can we add a section about tagging your commits with the JIRA ticket as well?  That makes life so much easier.

gregorydlogan (approved):
Minor nitpicks, but looks good to me.

CONTRIBUTING.md
lkiesow: I've simply removed the suggestion. There should be enough real-world examples to follow.

CONTRIBUTING.md
lkiesow: Added

CONTRIBUTING.md
lkiesow: Done

PR #543, opened 2018-11-02
MH-13209: Add CAS security bundles to opencast distrubution packages

miesgre:
CAS security bundles are not added to the distribution packages. 
If you try to enable the CAS authentication you get an error.

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -146,6 +146,8 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-working-file-repository-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-workspace-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-workspace-impl/${project.version}</bundle>
+
+    <feature start-level="80">opencast-security-cas</feature>
   </feature>
 
   <feature name="opencast-allinone">


lkiesow:
Can you explain your reasoning behind this in a few more details?
- What error do you get?
- Why does it not suffice to add opencast-security-cas to the bootFeature configuration?

miesgre:
@lkiesow , The problem is that the [osgi bundles](https://github.com/opencast/opencast/blob/r/6.x/assemblies/karaf-features/src/main/feature/feature.xml#L738-L745) needed by the [opencast-security-cas](https://github.com/opencast/opencast/blob/r/6.x/assemblies/karaf-features/src/main/feature/feature.xml#L738-L745) are not packaged in the opencast distribution packages (admin, worker, presenter,...)

In other words, the jar files are not located in `system/org/opencastproject/`. So when you add the `opencast-security-cas` to bootFeature, the CAS bundles are not located.

If you try this in a system where you compiled opencast. this works because the bundles are loaded from your local maven repository ($HOME/.m2/repository), but If you try this using the docker images (or I think also in deb/rpm packages, I not tested it), you have a problem because the bundles are not found.

lkiesow:
I can see that that causes a problem. But wouldn't it then make more sense to add this to the features (not bootFeatures) configured for the karaf-maven-plugin instead of enabling this by default for all distributions?

miesgre:
I'm not an expert in karaf or maven build. I found that this PR solves my problem, but feel free to dismiss this PR and fix this in a most convenient way.

lkiesow:
@miesgre, I've submitted the suggested change as pull request #544. Since I'm not using CAS, can you check if that works for you?

miesgre:
Closing this PR, since #544 fixes the same issue.

PR #544, opened 2018-11-04
MH-13209, Put CAS Feature In Distributions

lkiesow:
This patch ensures that Opencast's CAS security feature is contained in
all distributions and can easily be installed/enabled without enabling
it by default or making it a core dependency.

*This is an alternative solution to pull request #543*

### DIFF
--- a/assemblies/pom.xml
+++ b/assemblies/pom.xml
@@ -117,6 +117,9 @@
             </libraries>
             <archiveZip>false</archiveZip>
             <archiveTarGz>false</archiveTarGz>
+            <installedFeatures>
+              <feature>opencast-security-cas</feature>
+            </installedFeatures>
           </configuration>
         </plugin>
         <plugin>


miesgre:
It works @lkiesow. Now the `opencast-secitity-cas` libraries are in the correct place (`system/org/opencastproject/`) and when you add the feature to `bootFeatures`, it os loaded correctly:
karaf@root> bundle:list | grep security
146 | Resolved |  82 | 6.0.0.SNAPSHOT          | Opencast :: security-aai, Hosts: 111
147 | Resolved |  82 | 6.0.0.SNAPSHOT          | Opencast :: security-cas, Hosts: 111
148 | Active   |  82 | 6.0.0.SNAPSHOT          | Opencast :: security-cas-client-wrapper
149 | Resolved |  82 | 6.0.0.SNAPSHOT          | Opencast :: security-shibboleth, Hosts: 111

But, why other authentication methods like ldap or shibboleth are in the core and loaded by default and CAS not? Should we create feature for each authentication method?

lkiesow:
Good point. I'll bring it up at tomorrows technical meeting.

I think this is the better solution since it's easy to activate and has less code running by default. It would also be a good idea to do something like this for e.g. LDAP as well. But let's see what others think.

miesgre:
Yes, I think so.
Another think is that the kernel module has a LDAP dependencies (https://github.com/opencast/opencast/blob/develop/modules/kernel/pom.xml#L267-L270). I think LDAP dependencies should be removed from kernel bundle and added it to the security-ldap bundle like CAS or shibboleth.  :)

miesgre:
@lkiesow Will this be merged before 6.0? Is there anything I can do to help?
I [tested](https://github.com/opencast/opencast/pull/544#issuecomment-435781645) it, and it works for me.

staubesv:
I don't have CAS in place for testing but as @miesgre tested this and it looks fine for me, I'll accept the PR.

staubesv (approved):

PR #547, opened 2018-11-05
MH-13211 engage-ui: Fix live schedule bug: event available before schedule

mliradelc:
The bug fix in this PR is to avoid the behaviour of allow users to access to the live stream before an event starts.

The problem was the check algorithm only worked if the `data.mediapackage.media.track` variable had an array type. The live streaming events have type dict. Thus the for statement never started.

The fix makes an immediate check considering that the live stream only have live tracks.

### DIFF
--- a/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
+++ b/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
@@ -675,19 +675,15 @@ function($, bootbox, _, alertify) {
 
                     if (data.mediapackage.media && data.mediapackage.media.track) {
                     	// Check if there's a 'live' track
-                        for (var i = 0; i < data.mediapackage.media.track.length; i++) {
-                            var track = data.mediapackage.media.track[i];
-                            if (track.live) {
-                            	// Is event in progress?
-                            	var start = new Date(data.mediapackage.start);
-                            	var end = new Date(start.getTime() + parseInt (data.mediapackage.duration));
-                            	var now = new Date(); 
-                                if (now < start || now > end) {
-                                	live = msg_live_not_in_progress;
-                                	canLaunch = false;
-                                } else live = msg_live_in_progress;
-                                break;
-                            }
+                        if (data.mediapackage.media.track.live) {
+                            // Is event in progress?
+                            var start = new Date(data.mediapackage.start);
+                            var end = new Date(start.getTime() + parseInt (data.mediapackage.duration));
+                            var now = new Date();
+                            if (now < start || now > end) {
+                               	live = msg_live_not_in_progress;
+                               	canLaunch = false;
+                            } else live = msg_live_in_progress;
                         }
                     }
                     tile = tile + "<div class=\"live\">" + live + "</div>";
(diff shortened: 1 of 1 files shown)


modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
```diff
@@ -675,19 +675,15 @@ function($, bootbox, _, alertify) {
...
-                                	canLaunch = false;
-                                } else live = msg_live_in_progress;
-                                break;
-                            }
+                        if (data.mediapackage.media.track.live) {
+                            // Is event in progress?
+                            var start = new Date(data.mediapackage.start);
+                            var end = new Date(start.getTime() + parseInt (data.mediapackage.duration));
+                            var now = new Date();
+                            if (now < start || now > end) {
```
lkiesow: Not your code but would you mind changing this to `now >= end`?
It's based on the inclusive start time and duration after all and should hence not be inclusive.

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
```diff
@@ -675,19 +675,15 @@ function($, bootbox, _, alertify) {
...
+                               	live = msg_live_not_in_progress;
+                               	canLaunch = false;
+                            } else live = msg_live_in_progress;
                         }
                     }
                     tile = tile + "<div class=\"live\">" + live + "</div>";
```
lkiesow, line 689: `… + _.escape(live) + …` would be safer

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
```diff
@@ -675,19 +675,15 @@ function($, bootbox, _, alertify) {
...
-                                } else live = msg_live_in_progress;
-                                break;
-                            }
+                        if (data.mediapackage.media.track.live) {
+                            // Is event in progress?
+                            var start = new Date(data.mediapackage.start);
+                            var end = new Date(start.getTime() + parseInt (data.mediapackage.duration));
+                            var now = new Date();
+                            if (now < start || now > end) {
+                               	live = msg_live_not_in_progress;
```
lkiesow: Indentation + please avoid mixing tabs and spaces.

lkiesow (changes requested):

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
mliradelc: I think the same, I applied that change.

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
mliradelc: `_.escape(live)` leaves a` <br>` in the not live videos, it is better to put in the messages before that line.

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
mliradelc: Indentation fixed

lkiesow:
Looks good now. Thanks for the additional changes. Merged.

PR #548, opened 2018-11-05
MH-13213, Separate External API Index

lkiesow:
The search index of the external API is bundled in the same package
which also contains the REST endpoints, resulting in a re-start of a lot
of components if you just want to deploy a simple fix.

This patch separated the index so that a replacement of the external-api
bundle does not cause the index to re-start making hot-deployments
easier.

*Work sponsored by SWITCH*


staubesv:
@lkiesow Please note that this PR has merge conflicts.

PR #549, opened 2018-11-05
MH-13215: WorkflowOperationTagUtil throws a null pointer

gregorydlogan:
Copying from list, this is the stack trace:
```
2018-11-05 15:14:27,152 | ERROR | (WorkflowOperationWorker:180) - Workflow operation 'operation:'prepare-av', position:13, state:'FAILED'' failed
java.lang.NullPointerException
at org.opencastproject.workflow.api.WorkflowOperationTagUtil.createTagDiff(WorkflowOperationTagUtil.java:63)[170:opencast-workflow-service-api:6.0.0.SNAPSHOT]
at org.opencastproject.workflow.handler.composer.PrepareAVWorkflowOperationHandler.mux(PrepareAVWorkflowOperationHandler.java:175)[61:opencast-composer-workflowoperation:6.0.0.SNAPSHOT]
at ...
```
Obvious fix is to make sure the list always exists.

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowOperationTagUtil.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowOperationTagUtil.java
@@ -22,6 +22,7 @@
 package org.opencastproject.workflow.api;
 
 import org.opencastproject.mediapackage.Track;
+import org.opencastproject.util.data.Collections;
 
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


uzk-rrzk:
I do not think that this will fix the NPE.
A query  like 
    if (targetTags != null) {
   }
is still missing

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowOperationTagUtil.java
```diff
@@ -54,7 +55,8 @@ private WorkflowOperationTagUtil() {
   private static final String MINUS = "-";
 
   public static TagDiff createTagDiff(final String tagList) {
-    final String[] targetTags = StringUtils.split(tagList, ",");
+    final ArrayList<String> targetTags = new ArrayList<>();
+    ((ArrayList) targetTags).addAll(Collections.toList(StringUtils.split(tagList, ",")));
```
lkiesow, line 59: To avoid unchecked assignments and Opencast's weird internal collection utils, maybe use [nullToEmpty(…)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ArrayUtils.html#nullToEmpty-java.lang.String:A-) from Apache commons lang instead? The library is included anyway:

```java
final String[] targetTags = ArrayUtils.nullToEmpty(StringUtils.split(tagList, ","));
```

staubesv (changes requested):
@gregorydlogan Could you change the code to something like @lkiesow suggests?

staubesv:
@gregorydlogan Since you will be away until January 2019, I have created an alternative solution for this and closing this in favor of #602

PR #553, opened 2018-11-06
MH-13144, only set Job startDate if no set before

JamesUoM:
Stops jobs that update while they are running (eg START_WORKFLOW) from having their startDate reset each time


KatrinIhler:
@JamesUoM Is there a way to see the startdate of the START_WORKFLOW job change? It's not visible in the admin ui, is it?

JamesUoM:
> @JamesUoM Is there a way to see the startdate of the START_WORKFLOW job change? It's not visible in the admin ui, is it?

No, I think the "start" date the Admin UI displays is actually the created date or at least is never updated. I started OC with a MySQL database, ran a long workflow and monitored the job in my database viewer.

KatrinIhler (approved):
I watched the job table in the database to confirm this bug, and the start date does indeed change, which doesn't make any sense. This fixes it, so let's merge this! :)

PR #555, opened 2018-11-08
MH-12414, JPA spec states positional parameters are indexed from 1

JamesUoM:


smarquard:
Seems like this:

      final Object value = params[i];

would give an out-of-bounds exception where i==params.length and what you should be doing instead is using 0 to params.length-1 but calling

    setParameter(i+1, ...)

? Or else

      final Object value = params[i-1];

JamesUoM:
> Seems like this:
> 
> ```
>   final Object value = params[i];
> ```
> would give an out-of-bounds exception where i==params.length and what you should be doing instead is using 0 to params.length-1 but calling
> 
> ```
> setParameter(i+1, ...)
> ```

And that is exactly what we have in our code, I thought it would be neater to write the other way *sigh*

KatrinIhler:
Maybe this should also be changed in persistencefn/Queries for consistency? (Although I have no idea how these two modules differ since they seem to contain the same functionality and are both used in Opencast...)

JamesUoM:
> Maybe this should also be changed in persistencefn/Queries for consistency? (Although I have no idea how these two modules differ since they seem to contain the same functionality and are both used in Opencast...)

The other is (Entwine's) functional programming equivalent ... *sigh* ... I'll duplicate the changes.

KatrinIhler (approved):

PR #557, opened 2018-11-08
MH-13225: Fix for exception in live scheduler service when rebuilding the admin ui index

rute-santos:
Please see the ticket for the complete description.

### DIFF
--- a/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
+++ b/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
@@ -232,7 +232,7 @@ public class LiveScheduleServiceImpl implements LiveScheduleService {
       // Check if capture not over. We have to check because we may get a notification for past events if
       // the admin ui index is rebuilt
       DCMIPeriod period = EncodingSchemeUtils.decodeMandatoryPeriod(episodeDC.getFirst(DublinCore.PROPERTY_TEMPORAL));
-      if (period.getEnd().getTime() < System.currentTimeMillis()) {
+      if (period.getEnd().getTime() <= System.currentTimeMillis()) {
         logger.info("Live media package {} not created in search index because event is already past (end date: {})",
                 mpId, period.getEnd());
         return false;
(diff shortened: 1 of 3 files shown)


modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -240,14 +255,23 @@ public boolean deleteLiveEvent(String mpId) throws LiveScheduleException {
     if (mp == null) {
       logger.debug("Live media package {} not found in search index", mpId);
       return false;
-    } else
+    } else {
+      if (!isLive(mp)) {
+        logger.info("Media package {} is not live. Not retracting.", mpId);
+        return false;
+      }
       return retractLiveEvent(mp);
```
lkiesow, line 263: No strong feelings about this and it depends if you really want/need logging here but the whole this is just:

```java
return mp != null
    && isLive(mp)
    && retractLiveEvent(mp);
```

…which might be easier to read/understand. I'll leave it to you to decide what you want to use ;-P
A similar handling would be possible in the other methods btw.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -229,9 +229,24 @@ protected void activate(ComponentContext context) {
   public boolean createOrUpdateLiveEvent(String mpId, DublinCoreCatalog episodeDC) throws LiveScheduleException {
     MediaPackage mp = getMediaPackageFromSearch(mpId);
     if (mp == null) {
+      // Check if capture not over. We have to check because we may get a notification for past events if
+      // the admin ui index is rebuilt
+      DCMIPeriod period = EncodingSchemeUtils.decodeMandatoryPeriod(episodeDC.getFirst(DublinCore.PROPERTY_TEMPORAL));
+      if (period.getEnd().getTime() < System.currentTimeMillis()) {
```
lkiesow: Not that it's very relevant but this should be `<=` since the end date-time should be non-inclusive.

lkiesow (changes requested):

mliradelc (approved):
I tested the code against 5.X the before and after this patch and I can confirm that solves the error message.

About the code, I think the same as Lars about the changes.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
rute-santos: I agree that it would be more compact but, since you said you don't have strong feelings, I prefer to keep the logging. It has proved to be very helpful when debugging problems.

rute-santos:
Sorry, I had forgotten about this PR.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
lkiesow: ok

lkiesow:
Thanks for fixing this → merged

PR #566, opened 2018-11-14
MH-13235 Fix bug x

KatrinIhler:
test


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -291,6 +291,11 @@ public void activate(ComponentContext cc) {
 
   }
 
+  /**
+   * this is a test comment
+   * @return
```
Rillke, line 296: What does it return?

Rillke (changes requested):
Looks like a test. Should not be merged.

lkiesow:
I'm closing this since it was a demo for the developer workshop

PR #567, opened 2018-11-14
MH-12332, disable workflows whose tags don't explicitly match the source type, UPLOAD|SCHEDULE 5.x

JamesUoM:
ported from the 4.x commit that conflicted with 5.x

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
@@ -473,7 +473,8 @@
                         data-disable-search-threshold="0"
                         data-search_contains="true"
                         ng-model="wizard.step.ud.workflow"
-                        ng-options="w.title for (obj, w) in wizard.step.workflows"
+                        ng-options="w.title for (obj, w) in wizard.step.workflows |
+                                  filter: wizard.getStateControllerByName('source').ud.type==='UPLOAD' ? 'upload-ng' : 'schedule-ng'"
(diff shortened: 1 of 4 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
```diff
@@ -473,7 +473,8 @@
                         data-disable-search-threshold="0"
                         data-search_contains="true"
                         ng-model="wizard.step.ud.workflow"
-                        ng-options="w.title for (obj, w) in wizard.step.workflows"
+                        ng-options="w.title for (obj, w) in wizard.step.workflows |
+                                  filter: wizard.getStateControllerByName('source').ud.type==='UPLOAD' ? 'upload-ng' : 'schedule-ng'"
```
mtneug: `-ng` was dropped in 5.x
```suggestion
                                  filter: wizard.getStateControllerByName('source').ud.type==='UPLOAD' ? 'upload' : 'schedule'"
```

mtneug (changes requested):

modules/admin-ui/src/main/webapp/scripts/shared/partials/wizards/new-event.html
lkiesow: @mtneug, will you take the review of this pull request?

mtneug: @lkiesow yes.

mtneug (approved):

mtneug:
Works as intended.

PR #568, opened 2018-11-15
MH-13237 Track previews do not work with stream security

staubesv:
Currently, video editor track previews cannot be loaded if stream security is enabled because their URLs are not signed.


lkiesow (changes requested):
Hi @staubesv, there is now no reasonable use of `signUrl(String)` anymore. The only two places where it is used look like this: `signUrl(distElement.getURI().toString())`. So, it makes more sense to only have `signUrl(URI)` instead of creating a second varant of the method. Can you change that?`

staubesv:
I'm closing this in favor of #576

staubesv:
@lkiesow Thanks for the finding, I've fixed this in #576

PR #569, opened 2018-11-15
MH-13238, check max attempts before setting error state for related services

JamesUoM:
The test had to be fixed as it appeared to be written to match the results not the expected behaviour

### DIFF
--- a/modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
+++ b/modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
@@ -2502,12 +2502,14 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
       // Services in WARNING state triggered by current job
       List<ServiceRegistrationJpaImpl> relatedWarningServices = getRelatedWarningServices(job);
 
-      // Sets all related services to error state
+      // The related services are already in WARNING state and max attempts is reached
       for (ServiceRegistrationJpaImpl relatedService : relatedWarningServices) {
-        logger.info("State set to ERROR for related service {} on host {}", currentService.getServiceType(),
-                currentService.getHost());
-        relatedService.setServiceState(ERROR, job.toJob().getSignature());
-        updateServiceState(em, relatedService);
+        if (!currentService.equals(relatedService) && getHistorySize(relatedService) >= maxAttemptsBeforeErrorState) {
+          logger.info("State set to ERROR for related service {} on host {}", relatedService.getServiceType(),
+                  relatedService.getHost());
+          relatedService.setServiceState(ERROR, job.toJob().getSignature());
+          updateServiceState(em, relatedService);
+        }
       }
 
     }
--- a/modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
+++ b/modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
@@ -2447,6 +2447,7 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
           if (currentService.equals(relatedService))
             continue;
 
+          // De-escalate the state of related services as the issue is most likely with the job not the service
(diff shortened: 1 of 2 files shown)


KatrinIhler:
Any advice on how to test this?

modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -2502,12 +2502,14 @@ else if (job.getStatus() == Status.FINISHED) {
...
-        logger.info("State set to ERROR for related service {} on host {}", currentService.getServiceType(),
-                currentService.getHost());
-        relatedService.setServiceState(ERROR, job.toJob().getSignature());
-        updateServiceState(em, relatedService);
+        if (!currentService.equals(relatedService) && getHistorySize(relatedService) >= maxAttemptsBeforeErrorState) {
+          logger.info("State set to ERROR for related service {} on host {}", relatedService.getServiceType(),
+                  relatedService.getHost());
+          relatedService.setServiceState(ERROR, job.toJob().getSignature());
+          updateServiceState(em, relatedService);
+        }
```
KatrinIhler: Can this actually ever happen? I would assume that if >= maxAttempts is true here, the relatedService would already be in ERROR state. In that case it would be better to remove the whole _if_ here.

KatrinIhler (changes requested):

modules/matterhorn-serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
JamesUoM: It begs the question of why it's modifying the state of related services at all as surely if they failed themselves they would be caught be the code that precedes this for the current service. Assuming the good reason for this has been lost in time, there is no reason for the related services to the thrown into ERROR state while the parent service remains in WARNING state until MAX_ATTEMPTS is reached.

KatrinIhler: Well, the general idea probably was that if a related service is in warning state and the job runs successfully on another service, it's obviously not a problem with the job, but the related service, so it would be thrown into error state regardless of the number of attempts. But this could lead to a related service being in error state after only one attempt, which really doesn't sound like a good idea. 
But I think your fix (considering the number of attempts before kicking a related service into error state) really makes the whole check of related services obsolete (which is maybe a good thing?)

JamesUoM: Confused. If the fix makes it work then the code around it can't be obsolete otherwise the fix wouldn't be there.

KatrinIhler: My suggestion would be to remove the whole block instead:

        if (!currentService.equals(relatedService) && getHistorySize(relatedService) >= maxAttemptsBeforeErrorState) {
          logger.info("State set to ERROR for related service {} on host {}", relatedService.getServiceType(),
                  relatedService.getHost());
          relatedService.setServiceState(ERROR, job.toJob().getSignature());
          updateServiceState(em, relatedService);
        }

and then check whether this results in your desired behavior that you want to achieve, because I think it would.

JamesUoM: OK I follow now (blame's lack of sleep)

JamesUoM:
also the first clause
` if (job.getStatus() == FAILED)`

then set the related services from ERROR to WARNING and WARNING to NORMAL. This is all messed up surely

JamesUoM:
Sorry, it's taken me a bit to get my head around the related services and why they were being blamed when the job succeded on the current service and visa-versa. I've added a comment in the first clause to make this clearer

KatrinIhler:
@JamesUoM Looks good! But this should probably not go into 4.x since there won't be another release. Since this changes the way services go into error state, maybe this should even go into develop...what do you think?

JamesUoM:
Move to develop https://github.com/opencast/opencast/pull/615

PR #570, opened 2018-11-15
MH-13214, Fix HTTP Digest Authentication

lkiesow:
On the first start of Opencast, sometimes, everything appears normal but
digest authentication does not function, including service heartbeats.
This causes Opencast to fail to perform any processing, as well as
disabling communication with capture agents, and other
digest-integrators.

The problem is basically that when the in-memory user-provider is
initialized, it will create a system user for every registered
organization allowing the system access we want.

But the initialization of the organizations is happening in parallel
and nothing guarantees that all, or in fact any, organizations are
created by the time the user is created. This means that it may always
happen that there is no digest user for some (or all) of the
organizations which then means that the system digest requests will
fail for that organization.

This patch changes the behavior to create the organization specific
users the first time that organization is requested, thus guaranteeing
the organizations existence even for organizations created at runtime.

It also switches to a more performant storage of these users (access
always happens on an organizational basis and it hence make sense to use
an organization based storage instead of one global list) and a
modern, less obscure delivery of users (removing Entwine's functional
library).

### DIFF
--- a/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
+++ b/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
@@ -27,7 +27,6 @@ import org.opencastproject.security.api.JaxbOrganization;
 import org.opencastproject.security.api.JaxbRole;
 import org.opencastproject.security.api.JaxbUser;
 import org.opencastproject.security.api.Organization;
-import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.Role;
 import org.opencastproject.security.api.RoleProvider;
 import org.opencastproject.security.api.SecurityConstants;
@@ -35,23 +34,21 @@ import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.User;
 import org.opencastproject.security.api.UserProvider;
 
-import com.entwinemedia.fn.Fn;
-import com.entwinemedia.fn.Stream;
-import com.entwinemedia.fn.StreamOp;
-
 import org.apache.commons.lang3.StringUtils;
 import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * An in-memory user directory containing the users and roles used by the system.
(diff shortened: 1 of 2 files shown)


modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
```diff
@@ -124,75 +118,86 @@ protected void activate(ComponentContext cc) {
...
 
+  /**
+   * Get a list of all users for the current organization, loading the users from configuration if the organization had
+   * not been loaded before.
+   *
+   * @return List of users.
+   */
+  private List<User> getOrganizationUsers() {
+    Organization organization = securityService.getOrganization();
+    if (!inMemoryUsers.containsKey(organization)) {
```
staubesv: Hmm... is it possible that this method could be called by multiple threads concurrently even for the same organisation? In that case, this PR would have a thread-safety issue.

@lkiesow?

staubesv (dismissed):
I have some concerns considering thread-safety... (see comment).

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
lkiesow: Makes sense. While it's highly unlikely, it is possible. I've fixed that.

staubesv:
@lkiesow But the methods put, get and containsKey can still be called concurrently... doesn't this require a ConcurrentHashMap?

lkiesow:
Updated.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
```diff
@@ -124,75 +118,90 @@ protected void activate(ComponentContext cc) {
...
```
staubesv: One last thing: 

```
users = inMemoryUsers.get(organization);
if (users == null) {
   createSystemUsers(organization);
   users = inMemoryUsers.get(organization);
}
return users;
```
... would be more efficient for the common case as we don't need to lookup the users in the map twice.

Not a big deal as this data structure won't contain too many users, but anyway, it does save CPU cycles.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
```diff
@@ -124,75 +118,90 @@ protected void activate(ComponentContext cc) {
...
-    createSystemUsers();
   }
 
   @Override
   public String getName() {
     return PROVIDER_NAME;
   }
 
+  /**
+   * Get a list of all users for the current organization, loading the users from configuration if the organization had
```
staubesv: I don't think this returns the list of "all users" for the current organisation, does it?

lkiesow: In the context of this user and role provider, it does.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/InMemoryUserAndRoleProvider.java
lkiesow: done

staubesv (approved):

PR #579, opened 2018-11-19
MH-12816 Make waveform size configurable in WOH

staubesv:
The size of the waveform image that is produced by the waveform service can currently only be configured globally.
This patch moves the configuration parameters relevant for waveform image size from the waveform service
configuration to the workflow operation configuration so that Opencast can produce different waveform sizes.

### DIFF
--- a/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
+++ b/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
@@ -59,7 +59,6 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.URI;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Dictionary;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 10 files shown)


modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
```diff
@@ -450,9 +387,13 @@ private Attachment extractWaveform(Track track) throws WaveformServiceException
    * Create an ffmpeg waveform filter with parameters based on input track and service configuration.
    *
    * @param track source audio/video track with at least one audio channel
+   * @param pixelsPerMinute width of waveform image in pixels per minute
+   * @param minWidth minimum width of waveform image
+   * @param maxWidth maximum width of waveform image
+   * @param height height of waveform image
```
KatrinIhler, line 391: Documented parameters don't match the actual parameters below.

KatrinIhler (changes requested):
Apart from the one thing it's looking good!

modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
staubesv: Fixed.

KatrinIhler (approved):

PR #513, opened 2018-10-23
MH-13186, Episode and Series ACL Handling

lkiesow:
Opencast's handling of access control lists always felt like a magic
black box from which you would get a defined output for each input but
no one (with a few exceptions) really knew what the output was or even
should be, especially when it comes to the relationship and handling of
series and episode rules.

This is an attempt to finally solve the issue by introducing ACL merge
modes and a well defined, configurable behavior of the admin interface
when it comes to modifications of series acl.

This task consists of three parts:

- Defining the admin interface's behavior for series acl changes
- Defining the permission handling in case multiple rule-sets (e.g.
  series and episode acl) are attached to an event.
- Document how access control lists work and how they are evaluated
  given different settings.

Additional to these primary tasks, this also includes a general code
cleanup for the access control management which most notably collapses
several evaluation paths to ensure that evaluation of access control
always behaves similar and which also removes the evaluation of the
`security/xacml` flavor which was deprecated and not used anymore since
before Opencast 1.2 (see discussion on list).

Series ACL Changes In The Admin Interface
-----------------------------------------

Until now, the admin interface would always automatically remove all
episode acls whenever a series' permissions where modified to ensure
these new permissions apply for all episodes.

Now, this behavior can be configured with the following modes of
operation being available. Note that the default has changed compared to
the previous behavior in the hope that it is easier to userstand.

- *always:*
  When modifying series permissions, automatically remove all permission
  rules specific to single episodes belonging to the series. This
  enforces that every episode has the rules of the series in effect as
  soon as they are changed.
- *never:*
  Only update the series permissions but never replace permissions set
  on event level. This may mean that updated rules have no effect on
  already existing events.
- *optional (default):*
  Like `never` but present users with a button in the series permission
  dialog which allows them to replace the event specific permissions
  easily if they want to.

Handling/Merging Series and Episode Rules
-----------------------------------------

Access control lists can be specified both on series and on episode
level. This means that multiple rule sets can be attaches to an episode
which is part of a series. Opencast can now handle this in multiple
ways. If multiple lists are attached, the following modes define
Opencast's behavior:

### Merge Mode “override” (Default)

The episode ACL takes precedence over the series ACL. This means that
the series ACL will be completely ignored as soon as the episode has an
ACL, no matter what rules are set in either. This allows users to define
general rules for a series which can be completely redefined on an
episode and which are not influenced by changes later made to the
series.  This is also a very simple rule and thus easy to understand.

Example:

[...]

PR #580, opened 2018-11-19
MH-13243, Asset Manager ACL Cache Updates

lkiesow:
This patch ensures that the internal asset manager ACL cache gets
updated and no old rules are left in the database.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -238,10 +238,17 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator {
   }
 
   private void storeAclAsProperties(Snapshot snapshot, AccessControlList acl) {
+    final String mediaPackageId =  snapshot.getMediaPackage().getIdentifier().toString();
+    // Drop old ACL rules
+    final AQueryBuilder queryBuilder = createQuery();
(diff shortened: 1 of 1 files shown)


lkiesow:
This fixes the bug discussed in pull request #513

slampunk:
I'll /review/ this

slampunk:
LGTM, can be merged.

lkiesow:
Thanks for reviewing this, @slampunk

staubesv:
@lkiesow Bad news: This patch literally kills Opencast in case mh_asset_properties contains a lot of properties (> 1 million). On our test environment, several requests take more than 20 minutes to execute and a lot of database deadlocks, lock timeouts, etc. happen.

smarquard:
@staubesv, is this running a query that does a full table scan? Can you run an "EXPLAIN" on the query?

Incidentally, sounds like you have a lot of properties in that table, possibly not all not necessary, e.g. from the scheduler as in MH-12659.

staubesv:
@smarquard Yes, besides the asset manager properties (ACLs) and workflow processing settings, there are certainly quite a lot properties from the "new" scheduler (which we use since 2016).

... but if Opencast would issue reasonable queries on that table, the number of entries should not be problematic.

staubesv:
... just did some benchmarks: To delete the properties in a given namespace for a given media package takes no more than 4-5ms using a naive SQL query (with our table with more than 1 million entries), see https://opencast.jira.com/browse/MH-13334

smarquard:
What's the actual query being run? Can you enable mysql slow query logging?

PR #581, opened 2018-11-20
MH-13244 Improve concurrency of OAIPMH republication

KatrinIhler:
Remove the snapshot after removing the episode ACL and move the whole step to the AssetManagerUpdatedEventHandler so that the likelihood that two OAIPMH distribution jobs run concurrently for the same mediapackage (which can result in a lost update) is minimized. (See [MH-13233](https://opencast.jira.com/browse/MH-13244).)

### DIFF
--- a/modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/series/SeriesItem.java
+++ b/modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/series/SeriesItem.java
@@ -51,6 +51,7 @@ public final class SeriesItem implements MessageItem, Serializable {
   private final Boolean optOut;
   private final String element;
   private final String elementType;
+  private final Boolean overrideEpisodeAcl;
 
   public enum Type {
     UpdateCatalog, UpdateElement, UpdateAcl, UpdateOptOut, UpdateProperty, Delete
@@ -62,7 +63,7 @@ public final class SeriesItem implements MessageItem, Serializable {
    * @return Builds {@link SeriesItem} for updating a series.
    */
   public static SeriesItem updateCatalog(DublinCoreCatalog series) {
-    return new SeriesItem(Type.UpdateCatalog, null, series, null, null, null, null, null, null);
+    return new SeriesItem(Type.UpdateCatalog, null, series, null, null, null, null, null, null, null);
   }
 
   /**
@@ -75,7 +76,7 @@ public final class SeriesItem implements MessageItem, Serializable {
    * @return Builds {@link SeriesItem} for updating series element.
    */
   public static SeriesItem updateElement(String seriesId, String type, String data) {
-    return new SeriesItem(Type.UpdateElement, seriesId, null, null, null, null, null, type, data);
+    return new SeriesItem(Type.UpdateElement, seriesId, null, null, null, null, null, type, data, null);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


staubesv:
Hi @KatrinIhler 

I've just had a look at this pull request. Here some comments:

- The issue title is not really describing what this change does. While I know that OAI-PMH is the reason for the change, this change is not directly related to OAI-PMH and has the potential to improve other things. @lkiesow In case Opencast is configured to do ACL merging of event and series ACLs, this patch seems quite a must, doesn't it?

- Personally, I would recommend not to apply this change to r/4.x (implies 5.x and 6.x). It seems quite risky for a maintenance release to me.

Anyway, looking forwards to see this in production ;-)

lkiesow:
> Personally, I would recommend not to apply this change to r/4.x (implies 5.x and 6.x). It seems quite risky for a maintenance release to me.

From what it seems like, there will not even be another 4.x release. 4.x EOL is  in four days and no release is planned at the moment.

Re ACL: I'll take a look at it…

staubesv:
@lkiesow What I mean is that if the series ACL and the event ACL are to be both considered (ACL merging), you certainly don't want the episode ACL to be removed when updating the series ACL.

lkiesow:
@staubesv, the `override` parameter existed before. That hasn't changed.
It's just that there might have been a possible concurrency side-effect.
Hence, the title and description make sense.

KatrinIhler:
@staubesv Rebased the PR, ready for review again!

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -882,11 +883,13 @@ protected MetadataList deserializeMetadataList(String json) throws ParseExceptio
   @Produces({ "application/json", "application/v1.0.0+json" })
   @RestQuery(name = "updateseriesacl", description = "Updates a series' access policy.", returnDescription = "", pathParameters = {
           @RestParameter(name = "seriesId", description = "The series id", isRequired = true, type = STRING) }, restParameters = {
-                  @RestParameter(name = "acl", isRequired = true, description = "Access policy", type = STRING) }, reponses = {
+                  @RestParameter(name = "acl", isRequired = true, description = "Access policy", type = STRING),
+                  @RestParameter(name = "override", isRequired = false, defaultValue = "false", description = "If true the series ACL will take precedence over any existing episode ACL", type = STRING)}, reponses = {
```
staubesv: I hate to say it, but this is a change to the External API which we should not do without introducing a new minor version of the External API.

Can you somehow split this out of this PR and ensure that the External API will still behave the same as before?

KatrinIhler: Actually I don't think this changes anything - if the override parameter isn't used here, it defaults to false, which is the same behavior as before since the override happened not in the series service but in the ACLService which isn't used by the External API.

staubesv: It does introduce the form parameter "override" to the request PUT /api/series/{series_id}/acl which is an API change. Without introducing a new version of the External API, there would then be different versions of 1.1.0, for example: One 1.1.0 that does not support this parameter, and one 1.1.0 that does. 
But that implies that a client cannot determine whether the parameter override is supported based on the version of the External API as returned by /api/versions. 
Besides that, you should update the specificaiton of the External API (https://docs.opencast.org/develop/developer/api/series-api/#put-apiseriesseries_idacl).

I suggest to remove the new argument and just default "override" to whatever makes it behave the same as before.

Later, we can introduce the parameter "override" in version 1.2.0 of the External API. For now, a JIRA issue ("Add parameter override to PUT /api/series/{series_id}/acl") should do.

KatrinIhler: You're right, I'm gonna update this PR accordingly.

KatrinIhler:
@staubesv I removed the part affecting the external API. Once this is merged I can create a separate PR for develop that reintroduces this (or is there a fixed point in time when the External API is updated?)

staubesv:
@KatrinIhler Thanks! There is not really a fixed point for the introduction of new minor versions of the External API. As a minor version implies added functionality, the only restriction is that the patch needs to go to develop. While a new version is in progress on develop, everybody else can take the chance to add backwards-compatible changes to that version if required.

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/series/SeriesItem.java
```diff
@@ -51,6 +51,7 @@
   private final Boolean optOut;
   private final String element;
   private final String elementType;
+  private final Boolean overrideEpisodeAcl;
```
lkiesow, line 54: This, unfortunately, broke sending any series related messages in develop.
The issue is fixed by pull request #659.
Please avoid using any Objects except Strings in these message items.

KatrinIhler: Thanks for noticing, I'll try to keep this in mind for the future.

PR #585, opened 2018-11-23
MH-13221 Improve multi-select metadata fields

KatrinIhler:
Improve the code and behavior of the multi-select metadata fields (e.g. presenters, contributors, ...): 
* Fix focusing issues in Firefox where one has to click twice to edit
* enter multiple values without leaving edit mode in between
* remove multiple values without leaving edit mode in between
* don't save the entered value when leaving the text field by hitting ESC or clicking somewhere else, instead allowing the user to continue typing when re-entering edit mode
* simplify code and remove unused functionality

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableMultiSelectDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableMultiSelectDirective.js
@@ -65,56 +65,82 @@ angular.module('adminNg.directives')
       }
 
       scope.enterEditMode = function () {
-        scope.parseValues();
         scope.editMode = true;
-        scope.focusTimer = $timeout(function () {
+        $timeout(function () {
           element.find('input').focus();
         });
       };
 
+      scope.onBlur = function (event) {
+        if (!scope.removedValue) {
+          scope.leaveEditMode();
+        } else {
+          $timeout(function () {
+            element.find('input').focus();
+          });
+        }
+        delete scope.removedValue;
+      };
+
       scope.leaveEditMode = function () {
-        scope.storeValues();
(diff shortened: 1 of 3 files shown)


staubesv (approved):

lkiesow:
> don't save the entered value when leaving the text field by hitting ESC or clicking somewhere else, instead allowing the user to continue typing when re-entering edit mode

I just noticed this drastically changing the behavior of the user interfaces which make them work completely inconsistent with other admin interface input fields and imho create a huge usability issue. I'm also a bit baffled that something like this was added as a bug fix shortly before the release.

The problem: Open „Add Event“, start adding the title and use tab to switch to the next field (or click outside of the input field). The result is as expected: The input is saved and you can continue to edit the next field.

That behavior continues until you run into a multi-select field like presenter. Using the same mechanism will cause a completely different behavior: The input simply disappears. No explanation why. My expectation: A display bug. After all, the text seemed to be saved since it reappeared after clicking on the field again. But the data did also not show up in the summery. My expectation: A bug in the input fields causing a problem when saving the data. I did not at all guess and nowhere is explained that I now need to hit `Enter` to finish editing for *only* this kind of field.

I'm really think we should fix this and possibly just drop this patch for now.

KatrinIhler:
@lkiesow You're right, I didn't consider the inconsistency this would cause between the different metadata fields. It's probably best to drop the patch and I'll file a fixed version for develop.

lkiesow:
Sorry for causing the trouble 🙀

staubesv:
@lkiesow & @KatrinIhler Sorry, seems I've missed that during the review. I'm fine, of course, with having the patch reverted if that makes more sense.

PR #587, opened 2018-11-23
MH-13252, Ineffective Synchronization of Elasticsearch Startup

lkiesow:
Object level synchronization is ineffective for the startup of the
internal Elasticsearch node since multiple objects are working on the
same class level variable which should only be initialized once. Using
class level synchronization instead will ensure the server is started
only once in Opencast.

### DIFF
--- a/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java
+++ b/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java
@@ -284,7 +284,7 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
     this.indexVersion = version;
 
     // Configure and start Elasticsearch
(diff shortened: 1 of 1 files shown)


staubesv (dismissed):
Could you please also adjust the locking granularity that protect elasticSearch and nodeClient from concurrent access in the method close() to class level, too?

I mean https://github.com/opencast/opencast/blob/f2e5b3d5d9e2de92229bde557cc305f814bfb51a/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java#L323 and https://github.com/opencast/opencast/blob/f2e5b3d5d9e2de92229bde557cc305f814bfb51a/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java#L327

staubesv (approved):

staubesv (approved):

PR #588, opened 2018-11-25
MH-13253, External Elasticsearch

lkiesow:
By default, Opencast runs its own, internal Elasticsearch node as
backend for the external API and the admin interface. While this makes
configuration and deployment a lot easier and works just fine even in
larger deployments, connecting to an external Elasticsearch instead
comes with some advantages as well:

- Lightweight Opencast admin distribution
- Being able to run Elasticsearch as a cluster
- Service redundancy (e.g. multiple instances of the admin interface)
- Service separation (e.g. external API on a separate node)

This patch allows to configure an external Elasticsearch. Doing so will
automatically stop Opencast from trying to spin up its own, internal
Elasticsearch node. The index structure is still created automatically
at the first connection if necessary. No need for any manual
preparations.

*Note that this does not yet updates the Elasticsearch version. Hence,
the external node should still be Elasticsearch 1.7.6.*


docs/guides/admin/docs/modules/searchindex/elasticsearch.md
```diff
@@ -0,0 +1,52 @@
...
+
+Next, configure the correct cluster name for all indexes in `etc/index/*/settings.yml`:
+
+```yml
+cluster.name: opencast
+```
+
+Make sure that the correct cluster name is set in the configuration file of each index.
+
+Once started, Opencast should now try to use the external Elasticsearch.
```
staubesv: Why not just "Opencast will now use the external Elasticsearch."?

docs/guides/admin/docs/modules/searchindex/elasticsearch.md
```diff
@@ -0,0 +1,52 @@
+Elasticsearch Configuration
+===========================
+
+Elasticsearch is powering the external API as well as the administrative user interface of Opencast. By default,
+Opencast will start its own, internal Elasticsearch node as part of the admin distribution and no special configuration
+or deployment is required. This should work just fine even in large deployments.
```
staubesv: I suggest to omit "This should work just fine even in large deployments.".

docs/guides/admin/docs/modules/searchindex/elasticsearch.md
```diff
@@ -0,0 +1,52 @@
...
+Configuring External Nodes
+--------------------------
+
+To configure an external node, set the server's address in `etc/custom.properties`:
+
+```properties
+org.opencastproject.elasticsearch.server.address=127.0.0.1
+```
+
+Once this is set, Opencast will not try to launch its own internal Elasticsearch anymore. If necessary, you can also
```
staubesv: Suggestion: Omit "try to"

docs/guides/admin/docs/modules/searchindex/index.md
```diff
@@ -0,0 +1,24 @@
+Search Indexes
+==============
+
+Opencast comes with a few search indexes which act both as a cache and as a fast way to perform full text searches on
```
staubesv: Maybe: "several" instead of "few"

docs/guides/admin/docs/modules/searchindex/index.md
```diff
@@ -0,0 +1,24 @@
...
+metadata. By default, all search indexes are created automatically and no additional external software is required.
+
+While this should even work and perform relatively well on a large cluster, all indexes can be deployed separately. This
```
staubesv: Suggestion: "While this does work on large clusters,"

docs/guides/admin/docs/modules/searchindex/index.md
```diff
@@ -0,0 +1,24 @@
...
+comes with the obvious drawback of a harder deployment but has also a few advantages like a smaller core system or being
+able to have some services redundancies which would not be possible otherwise.
+
+Opencast currently depends on two search index types: Solr and Elasticsearch.
```
staubesv: Maybe: "uses" instead of "depends"

docs/guides/admin/docs/modules/searchindex/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+---
+
+- Solr is mostly powering older services and replacing this index type is planned fur the future. But for now it is
```
staubesv: f**o**r

modules/search/pom.xml
```diff
@@ -131,7 +131,7 @@
     <dependency>
       <groupId>com.spatial4j</groupId>
       <artifactId>spatial4j</artifactId>
-      <version>0.4.1</version>
+      <version>0.4</version>
```
staubesv, line 134: Is there a special reason for the downgrade?

staubesv (dismissed):
I've done the skim-over part of the review and found some minor issues.

Please note that your wording is a bit on the "In case you really manage to get Opencast up and running, there is at least a slight chance that it might work relatively well unless it doesn't"-side ;-)

I think it is ok to just say "does work" vs. "should work relatively well".

modules/search/pom.xml
lkiesow: Yes. This is changed so that the library versions of what is actually used and included through other libraries. Just compile the module. Maven will even print a warning about this.

staubesv (approved):

PR #589, opened 2018-11-25
MH-13254, Automate Dependency Checking

lkiesow:
This patch enables *optional* automated dependency checks on Opencast
modules. While Opencast rpovides a default configuration, modules still
need to opt-in by including the maven-depenency-plugin.

If this works, long term, we could at some point enforce dependency
checking globally to prevent depedency problems right from the start.

---

Note that the dependency plugin is not perfect and may complain about
pure runtime dependencies (e.g. it does not know that we may want to
attach a logger to our test and may complain about it being an
unnecessary dependency).

But the plugin allows making exceptions for such cases. This means that
developers can still declare any dependencies they like, but they are
forced to consider if they actually make sense since they need to
possibly whitelist them if they are not directly used.

---

This patch provides some example usages, including some whitelistings in
the implementation classes.

### DIFF
--- a/modules/animate-api/pom.xml
+++ b/modules/animate-api/pom.xml
@@ -25,6 +25,10 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+      </plugin>
+      <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
         <extensions>true</extensions>
--- a/modules/animate-impl/pom.xml
+++ b/modules/animate-impl/pom.xml
@@ -57,6 +57,14 @@
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>jsr311-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
 
     <!-- testing -->
     <dependency>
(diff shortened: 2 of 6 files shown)


lkiesow:
This is an example of including the dependency check without any special cases:

https://github.com/opencast/opencast/blob/cfc4408b45a453bed1432024198c81149b91467a/modules/workspace-api/pom.xml#L26-L29

This is an example of including the dependency check where we have an indirect dependency we want to keep. The logger is not used directly and the module will compile, test and work just fine without it but we would not see any logging output when we run the tests. Hence we declare an exceptions:

https://github.com/opencast/opencast/blob/cfc4408b45a453bed1432024198c81149b91467a/modules/animate-impl/pom.xml#L88-L98

For more information, take a look at the [documentation of excluding dependencies from dependency analysis](https://maven.apache.org/plugins/maven-dependency-plugin/examples/exclude-dependencies-from-dependency-analysis.html).

gregorydlogan:
How would I turn this on, and what does the output look like?

lkiesow:
> How would I turn this on

See example right above your comment :-)

> and what does the output look like?

The output is identical to what the usual `mvn dependency:analyze` will give you.
When included, it's just executed as part of the build and will make Maven fail if a problem was detected.

Example:
```sh
…/working-file-repository-service-impl% mvn dependency:analyze
…
[WARNING] Used undeclared dependencies found:
[WARNING]    org.apache.commons:commons-lang3:jar:3.7:compile
[WARNING]    org.apache.httpcomponents:httpclient-osgi:jar:4.5.6:compile
[WARNING]    org.apache.httpcomponents:httpcore-osgi:jar:4.4.10:compile
[WARNING] Unused declared dependencies found:
[WARNING]    org.slf4j:slf4j-log4j12:jar:1.7.21:test
[WARNING]    log4j:log4j:jar:1.2.17:test
[WARNING]    org.apache.cxf:cxf-rt-frontend-jaxrs:jar:3.2.6:test
…
```

staubesv (approved):

PR #592, opened 2018-11-26
MH-13258, Broken User Provider Removal

lkiesow:
All user providers in Opencast are dynamically bound to the user
directory service which then acts as a unified interface to access all
users in Opencast.

This means that during at runtime, new user providers can dynamically be
added and old ones removed. However, the removal process in the
directory service is broken which can cause references to old user
services being left in the system. This can then lead to effects like
invalid users or multiple references to identical users.

This patch fixes the removal of user providers.

*Work sponsored by SWITCH*


mtneug:
@staubesv are you taking the review or can I assign myself? This seems to be a simple spelling mistake.

lkiesow:
@mtneug, it's a SWITCH patch. Hence, @staubesv shouldn't take it. If you can review this, it would be great.

mtneug:
Works as intended.

PR #597, opened 2018-11-28
MH-13257: Fix outdated command line argument for tesseract >= 4.0.0

EdmundoDelGusto:
As said in [MH-13257](https://opencast.jira.com/browse/MH-13257) tesseract >= 4.0.0 doesn't allow '-psm' as command line argument. Which is causing the tests to fail.

The right spelling is '--psm' which should also be accepted by tesseract 3

Without it, test failed with the following message:
`Error, unknown command line argument '-psm' `

### DIFF
--- a/docs/guides/admin/docs/modules/textextraction.md
+++ b/docs/guides/admin/docs/modules/textextraction.md
@@ -44,19 +44,19 @@ analysis.
 
 For example, if you want OCR for German content, you want to run something like this:
 
-    tesseract in.tif out.txt -l deu -psm 3
+    tesseract in.tif out.txt -l deu --psm 3
 
 * The arguments `in.tif` and `out.txt` are automatically set by Opencast.
 * The argument `-l` specifies the language files used by Tesseract. `deu` specifies the German language. Multiple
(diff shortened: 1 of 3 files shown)


mtneug:
I recently ran into the same issue having tesseract 4 installed locally. I would imagine, however, that most distributions still ship with tesseract 3. Do you know, if `--psm` is accepted by tesseract 3?

lkiesow:
Tesseract 3.05.02 accepts both variants and `tesseract --help` already lists `--psm` as the correct variant.

@mtneug, do you want to take the review of this pull request?

mtneug:
@EdmundoDelGusto looks good. Would you redirect this to r/4.x? I will propagate this change to the other releases.

mtneug:
Correction: could you redirect to `r/5.x`. 4.x is soon EOL.

mtneug:
@EdmundoDelGusto would it be possible for you to redirect this to `r/5.x` today? If possible I would like to include this fix in the 6.0 release.

EdmundoDelGusto:
@mtneug done

PR #598, opened 2018-11-28
MH-13263, Invalid Ingest Encoding

lkiesow:
The POST /ingest/ingest REST endpoints accept form data encoded as
`application/x-www-form-urlencoded` or as `multipart/form-data`.
Regardless of the encoding however, a url-decoding of all values is
performed. This may lead to corruption of data sent in this request.

A simple example for a problem is attaching an episode access control
list which is expected to have the flavor `security/xacml+episode`.
Ingesting this in a `multipart/form-data` request will force the
url-decoding of the non-encoded data and will convert this flavor to
`security/xacml episode`.

Opencast will now obviously not match this modified flavor anymore when
it is looking for ACLs and is furthermore not at all happy about the
space contained in the flavor causing further problems.

This patch fixes the decoding problems.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
@@ -62,6 +62,7 @@ import com.google.common.cache.CacheBuilder;
 
 import org.apache.commons.fileupload.FileItemIterator;
 import org.apache.commons.fileupload.FileItemStream;
+import org.apache.commons.fileupload.FileUploadException;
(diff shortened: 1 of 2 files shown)


mtneug (commented):
The patch looks good and fixes the problem.

What I don't understand is the `@Produces(MediaType.TEXT_HTML)`. These endpoints return the workflow instance as XML. Shouldn't `TEXT_XML` be used? Can it be safely changed without breaking compatibility with existing clients?

lkiesow:
> Shouldn't TEXT_XML be used?

Yes.

> Can it be safely changed without breaking compatibility with existing clients?

Likely yes, but you never know… I'll create a pull request against `develop`. That will give people enough time to spot possible problems.

PR #599, opened 2018-11-29
WIP MH-13240: first pass at removing solr index for workflows

slampunk:
First pass at removing workflows from the Solr index. Instead of getting bogged down in the code, I've decided to do this WIP PR for comments/feedback on my approach: is it the most optimal solution? should we use elasticsearch where we can? etc.

This patch will ultimately remove the need for workflow information to be saved to the Solr index. The approach put forth is to query the DB directly (via the ServiceRegistry interface) for workflow information.

The thrust of this patch at the most basic level is to replace every Solr index query for workflows with its analogue provided by the ServiceRegistry; if no such analogue exists, extend the ServiceRegistry to provide the functionality required. An example of such an extension is to make the ServiceRegistry aware of mediapackages (currently, the ServiceRegistry has no concept of a mediapackage, meaning that it cannot map mediapackages to their workflows).

Note: I've disabled a few tests that need to be updated to make this patch work.

Feel free to leave comments or feedback.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1486,6 +1486,43 @@ public abstract class AbstractEventEndpoint {
                   @RestResponse(description = "No event with this identifier was found.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
   public Response getEventWorkflows(@PathParam("eventId") String id)
           throws UnauthorizedException, SearchIndexException, JobEndpointException {
+    logger.info("Looking for workflows for {} via the DB", id);
+    Opt<Event> optEvent = getIndexService().getEvent(id, getIndex());
+    if (optEvent.isNone())
+      return notFound("Cannot find an event with id '%s'.", id);
+
+    try {
+      if (!optEvent.get().hasRecordingStarted()) {
+        List<Field> fields = new ArrayList<Field>();
+        Map<String, String> workflowConfig = getSchedulerService().getWorkflowConfig(id);
+        for (Entry<String, String> entry : workflowConfig.entrySet()) {
+          fields.add(f(entry.getKey(), v(entry.getValue(), Jsons.BLANK)));
+        }
+
+        Map<String, String> agentConfiguration = getSchedulerService().getCaptureAgentConfiguration(id);
+        return okJson(obj(f("workflowId", v(agentConfiguration.get(CaptureParameters.INGEST_WORKFLOW_DEFINITION), Jsons.BLANK)),
+                f("configuration", obj(fields))));
+      } else {
+        return okJson(getJobService().getTasksAsJSON(id));
+      }
+    } catch (NotFoundException e) {
+      return notFound("Cannot find workflows for event %s", id);
+    } catch (SchedulerException e) {
+      logger.error("Unable to get workflow data for event with id {}", id);
+      throw new WebApplicationException(e, SC_INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  @GET
+  @Path("{eventId}/solr/workflows.json")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RestQuery(name = "geteventworkflows", description = "Returns all the data related to the workflows tab in the event details modal as JSON", returnDescription = "All the data related to the event workflows tab as JSON", pathParameters = {
+          @RestParameter(name = "eventId", description = "The event id", isRequired = true, type = RestParameter.Type.STRING) }, reponses = {
+                  @RestResponse(description = "Returns all the data related to the event workflows tab as JSON", responseCode = HttpServletResponse.SC_OK),
+                  @RestResponse(description = "No event with this identifier was found.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
+  public Response getEventWorkflowsFromDB(@PathParam("eventId") String id)
+          throws UnauthorizedException, SearchIndexException, JobEndpointException {
+    logger.info("Looking for workflows for {} via the Solr", id);
     Opt<Event> optEvent = getIndexService().getEvent(id, getIndex());
     if (optEvent.isNone())
       return notFound("Cannot find an event with id '%s'.", id);
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
@@ -425,6 +425,55 @@ public class JobEndpoint {
   }
 
   /**
+   * Returns the list of tasks matching the given query as JSON Object
+   *
+   * @param query
+   *          The worklfow query
+   * @return The list of matching tasks as JSON Object
+   * @throws JobEndpointException
+   * @throws NotFoundException
+   */
+  public JObject getTasksAsJSON(String mediapackageId) throws JobEndpointException, NotFoundException {
+    // Get results
+    WorkflowSet workflowInstances = null;
+    long totalWithoutFilters = 0;
+    List<JValue> jsonList = new ArrayList<>();
+
+    try {
+      workflowInstances = workflowService.getWorkflowInstances(mediapackageId);
+      totalWithoutFilters = workflowService.countWorkflowInstances();
+    } catch (WorkflowDatabaseException e) {
+      throw new JobEndpointException(String.format("Not able to get the list of job from the database: %s", e),
+              e.getCause());
+    }
+
+    WorkflowInstance[] items = workflowInstances.getItems();
+
+    for (WorkflowInstance instance : items) {
+      long instanceId = instance.getId();
+      String series = instance.getMediaPackage().getSeriesTitle();
+
+      // Retrieve submission date with the workflow instance main job
+      Date created;
+      try {
+        created = serviceRegistry.getJob(instanceId).getDateCreated();
+      } catch (ServiceRegistryException e) {
+        throw new JobEndpointException(String.format("Error when retrieving job %s from the service registry: %s",
+                instanceId, e), e.getCause());
+      }
+
+      jsonList.add(obj(f("id", v(instanceId)), f("title", v(nul(instance.getMediaPackage().getTitle()).getOr(""))),
+              f("series", v(series, Jsons.BLANK)), f("workflow", v(instance.getTitle(), Jsons.BLANK)),
+              f("status", v(WORKFLOW_STATUS_TRANSLATION_PREFIX + instance.getState().toString())),
+              f("submitted", v(created != null ? DateTimeSupport.toUTC(created.getTime()) : ""))));
+    }
+
+    JObject json = obj(f("results", arr(jsonList)), f("count", v(workflowInstances.getTotalCount())),
+            f("offset", v(0)), f("limit", v(jsonList.size())), f("total", v(totalWithoutFilters)));
+    return json;
+  }
+
+  /**
    * Returns the single task with the given Id as JSON Object
    *
    * @param id
--- a/modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
+++ b/modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
@@ -389,6 +389,7 @@ public class AbstractEventEndpointTest {
   }
 
   @Test
+  @Ignore
   public void testGetEventWorkflows() throws Exception {
 
     given().pathParam("eventId", "notExists").expect().statusCode(HttpStatus.SC_NOT_FOUND).when()
--- a/modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
+++ b/modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
@@ -99,11 +99,13 @@ import javax.persistence.Version;
         @NamedQuery(name = "Job.children", query = "SELECT j FROM Job j WHERE j.parentJob.id = :id ORDER BY j.dateCreated"),
         @NamedQuery(name = "Job.withoutParent", query = "SELECT j FROM Job j WHERE j.parentJob IS NULL"),
         @NamedQuery(name = "Job.avgOperation", query = "SELECT j.operation, AVG(j.runTime), AVG(j.queueTime) FROM Job j GROUP BY j.operation"),
+        @NamedQuery(name = "Job.mediaPackage", query = "SELECT j FROM Job j where j.mediapackage = :mediapackage"),
 
         // Job count queries
         @NamedQuery(name = "Job.count", query = "SELECT COUNT(j) FROM Job j "
                 + "where j.status = :status and j.creatorServiceRegistration.serviceType = :serviceType"),
         @NamedQuery(name = "Job.count.all", query = "SELECT COUNT(j) FROM Job j"),
+        @NamedQuery(name = "Job.count.workflows", query = "SELECT COUNT(j) FROM Job j where j.operation = :operation"),
         @NamedQuery(name = "Job.count.nullType", query = "SELECT COUNT(j) FROM Job j " + "where j.status = :status"),
         @NamedQuery(name = "Job.count.nullStatus", query = "SELECT COUNT(j) FROM Job j "
                 + "where j.creatorServiceRegistration.serviceType = :serviceType"),
@@ -152,6 +154,9 @@ public class JpaJob {
 
   @Column(name = "status")
   private int status;
+
+  @Column(name = "mediapackage")
+  private String mediapackage;
 
   @Lob
   @Column(name = "operation", length = 65535)
@@ -276,6 +281,21 @@ public class JpaJob {
     this.dispatchable = dispatchable;
     this.jobLoad = load;
     this.status = Status.INSTANTIATED.ordinal();
+  }
+
+  public JpaJob(User currentUser, Organization organization, ServiceRegistrationJpaImpl creatingService,
+          String operation, List<String> arguments, String payload, boolean dispatchable, float load, String mediapackage) {
+    this.creator = currentUser.getUsername();
+    this.organization = organization.getId();
+    this.creatorServiceRegistration = creatingService;
+    this.jobType = creatingService.getServiceType();
+    this.operation = operation;
+    this.arguments = arguments;
+    this.payload = payload;
+    this.dispatchable = dispatchable;
+    this.jobLoad = load;
+    this.status = Status.INSTANTIATED.ordinal();
+    this.mediapackage = mediapackage;
   }
 
   public static JpaJob from(Job job) {
@@ -417,6 +437,10 @@ public class JpaJob {
     this.failureReason = failureReason;
   }
 
+  public void setMediapackageIdentifier(String mediapackage) {
+    this.mediapackage = mediapackage;
+  }
+
   public void setUri(URI uri) {
     this.uri = uri;
   }
@@ -488,6 +512,10 @@ public class JpaJob {
 
   public String getOrganization() {
     return organization;
+  }
+
+  public String getMediapackageIdentifier() {
+    return mediapackage;
   }
 
   @Override
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
@@ -1643,10 +1643,11 @@ public class IndexServiceImpl implements IndexService {
 
   @Override
   public Opt<WorkflowInstance> getCurrentWorkflowInstance(String mpId) throws IndexServiceException {
-    WorkflowQuery query = new WorkflowQuery().withMediaPackage(mpId);
+//    WorkflowQuery query = new WorkflowQuery().withMediaPackage(mpId);
     WorkflowSet workflowInstances;
     try {
-      workflowInstances = workflowService.getWorkflowInstances(query);
+//      workflowInstances = workflowService.getWorkflowInstances(query);
+      workflowInstances = workflowService.getWorkflowInstances(mpId);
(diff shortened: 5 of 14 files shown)


modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
```diff
@@ -153,6 +155,9 @@
   @Column(name = "status")
   private int status;
 
+  @Column(name = "mediapackage")
```
staubesv, line 158: This requires a data migration, doesn't it? Also, a job is associated to a workflow, not a media package (with exception of the job that represents the workflow). Seems from that point, I would avoid to create a direct link from a job to a media package.

smarquard: The job that represents the workflow is exactly the job that should have this mediapackage value set (it could be null for jobs that are operations as part of a workflow). The data migration will be a db conversion script using ExtractValue to get the mediapackage identified from the xml.

slampunk: Hi @staubesv 

Yes, as it stands this patch requires a data migration*.

I'm hoping to:
1. minimise the time spent performing an OC migration
2. make it "easier" to migrate events' true state... or at least make the migration process less destructive than what it currently happens to be (rebuilding indexes).

On your second point:
The direct link you refer to depends on the Solr index.

This patch attempts to remove Opencast's reliance on the Solr index for (almost) all things workflow related. The implication thereof means that the link between mediapackage -> workflow info via solr -> jobs breaks, hence the need for an alternative link between media packages and their associated jobs.

Also, only those jobs of operation START_WORKFLOW have a reference to their media package id (since actual workflows **are** jobs, but _not all_ jobs are workflows :)

Coming back to data migration*: using OC's current elasticsearch index schema, I could (in a bit of a roundabout way) get this patch to sidestep the need for data migration. But then I'm back to violating hope # 2 above.
Either way, I plan on implementing some elasticsearch querying to minimise hitting the DB for every workflow request.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
```diff
@@ -424,6 +424,55 @@ public JObject getTasksAsJSON(WorkflowQuery query) throws JobEndpointException,
...
+    WorkflowSet workflowInstances = null;
+    long totalWithoutFilters = 0;
+    List<JValue> jsonList = new ArrayList<>();
+
+    try {
+      workflowInstances = workflowService.getWorkflowInstances(mediapackageId);
+      totalWithoutFilters = workflowService.countWorkflowInstances();
+    } catch (WorkflowDatabaseException e) {
+      throw new JobEndpointException(String.format("Not able to get the list of job from the database: %s", e),
+              e.getCause());
```
lkiesow, line 447: Why skip parts of the stack trace here?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1486,6 +1486,43 @@ public Response getPublication(@PathParam("eventId") String eventId, @PathParam(
...
   public Response getEventWorkflows(@PathParam("eventId") String id)
           throws UnauthorizedException, SearchIndexException, JobEndpointException {
+    logger.info("Looking for workflows for {} via the DB", id);
+    Opt<Event> optEvent = getIndexService().getEvent(id, getIndex());
+    if (optEvent.isNone())
+      return notFound("Cannot find an event with id '%s'.", id);
+
+    try {
+      if (!optEvent.get().hasRecordingStarted()) {
+        List<Field> fields = new ArrayList<Field>();
```
lkiesow, line 1496: Nothing really wrong with this but the second `Field` is superfluous:
```java
List<Field> fields = new ArrayList<>();
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1486,6 +1486,43 @@ public Response getPublication(@PathParam("eventId") String eventId, @PathParam(
...
+        return okJson(obj(f("workflowId", v(agentConfiguration.get(CaptureParameters.INGEST_WORKFLOW_DEFINITION), Jsons.BLANK)),
+                f("configuration", obj(fields))));
+      } else {
+        return okJson(getJobService().getTasksAsJSON(id));
+      }
+    } catch (NotFoundException e) {
+      return notFound("Cannot find workflows for event %s", id);
+    } catch (SchedulerException e) {
+      logger.error("Unable to get workflow data for event with id {}", id);
+      throw new WebApplicationException(e, SC_INTERNAL_SERVER_ERROR);
```
karendolan, line 1512: Would you mind adding e to the end of that ERROR log to print the error stack track, otherwise the actual issue will be lost.

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
```diff
@@ -389,6 +389,7 @@ public void testGetAttachment() throws Exception {
   }
 
   @Test
+  @Ignore
```
karendolan, line 392: Adding a "//TODO: fix test"  will make it easier for other developers to see this ignore and address it in the future.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
karendolan: Hi @slampunk, I have a *strong* request for the term "mediapackage" to be renamed to "mediapackageId" or "mpId" or something that implies that that value has no relationship to the mediapackage manifest and is only the id of that package.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
```diff
@@ -424,6 +424,55 @@ public JObject getTasksAsJSON(WorkflowQuery query) throws JobEndpointException,
...
+    }
+
+    WorkflowInstance[] items = workflowInstances.getItems();
+
+    for (WorkflowInstance instance : items) {
+      long instanceId = instance.getId();
+      String series = instance.getMediaPackage().getSeriesTitle();
```
karendolan, line 454: I'm again reiterating my request to use the term mpId versus "mediapackage"  to represent the mediapackage id in this pull :) . Traditionally, calls like "getMediaPackage", shown above, return the media package, versus just the id of the package. The addition of the term "mediapackage" in this pull, to represent the mediapackage identifier, not the mediapackage and adds confusion with the traditional use of the term "mediapackage" to represent the media package.

rute-santos: +1

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
```diff
@@ -424,6 +424,55 @@ public JObject getTasksAsJSON(WorkflowQuery query) throws JobEndpointException,
...
+
+      // Retrieve submission date with the workflow instance main job
+      Date created;
+      try {
+        created = serviceRegistry.getJob(instanceId).getDateCreated();
+      } catch (ServiceRegistryException e) {
+        throw new JobEndpointException(String.format("Error when retrieving job %s from the service registry: %s",
+                instanceId, e), e.getCause());
```
lkiesow, line 462: Why not
```java
throw new JobEndpointException(String.format("Error when retrieving job %s from the service registry", instanceId), e);
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
```diff
@@ -424,6 +424,55 @@ public JObject getTasksAsJSON(WorkflowQuery query) throws JobEndpointException,
...
+
+      jsonList.add(obj(f("id", v(instanceId)), f("title", v(nul(instance.getMediaPackage().getTitle()).getOr(""))),
+              f("series", v(series, Jsons.BLANK)), f("workflow", v(instance.getTitle(), Jsons.BLANK)),
+              f("status", v(WORKFLOW_STATUS_TRANSLATION_PREFIX + instance.getState().toString())),
+              f("submitted", v(created != null ? DateTimeSupport.toUTC(created.getTime()) : ""))));
+    }
+
+    JObject json = obj(f("results", arr(jsonList)), f("count", v(workflowInstances.getTotalCount())),
+            f("offset", v(0)), f("limit", v(jsonList.size())), f("total", v(totalWithoutFilters)));
+    return json;
```
lkiesow, line 473: The variable `json´ is superfluous and you could just use `return obj(f(…`

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
lkiesow: Please don't. If this should be done: Do it.
Adding todos, … only leads to confusing code where in the end no one really knows what's supposed to happen.
If you know it should be done, do it.
If it doesn't matter, drop it.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
```diff
@@ -99,11 +99,13 @@
         @NamedQuery(name = "Job.children", query = "SELECT j FROM Job j WHERE j.parentJob.id = :id ORDER BY j.dateCreated"),
         @NamedQuery(name = "Job.withoutParent", query = "SELECT j FROM Job j WHERE j.parentJob IS NULL"),
         @NamedQuery(name = "Job.avgOperation", query = "SELECT j.operation, AVG(j.runTime), AVG(j.queueTime) FROM Job j GROUP BY j.operation"),
+        @NamedQuery(name = "Job.mediaPackage", query = "SELECT j FROM Job j where j.mediapackage = :mediapackage"),
 
         // Job count queries
         @NamedQuery(name = "Job.count", query = "SELECT COUNT(j) FROM Job j "
                 + "where j.status = :status and j.creatorServiceRegistration.serviceType = :serviceType"),
         @NamedQuery(name = "Job.count.all", query = "SELECT COUNT(j) FROM Job j"),
+        @NamedQuery(name = "Job.count.workflows", query = "SELECT COUNT(j) FROM Job j where j.operation = :operation"),
```
lkiesow, line 108: The name `Job.count.workflows` implies that this will count workflows which requires the operation to be `START_WORKFLOW`. Everything else wouldn't make sense. Hence, this shouldn't be a variable parameter, should it? If it stays, this should be called e.g. `Job.count.operations`.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
lkiesow: > The data migration will be a db conversion script using ExtractValue to get the mediapackage identified from the xml.

I think what @staubesv meant was most of all that this data migration and also the required SQL script adjustments are missing from this pull request.

I also kind of agree that what you are doing is a bit weird: You are trying to extend a database table to fit two different types of data which is generally a bad idea and separating the data instead if that is required would probably make more sense though I can see what you did it: It's easier after all and does require fewer changes.

Alternatives would be to:

- Separate jobs and workflows. This is certainly more work but you could store exactly what you need for a workflow, index the fields you want to search through fast, …
- Keep the generic jobs and use `oc_job_argument` to store the media package identifier instead. This would result in a query like `select job.* from oc_job job join oc_job_arguments arg on job.id = arg.id  where arg.argument = :mediapackage_id`. I think this should still be fast enough and would keep the job as a job.

In the end, I could probably live with adding the media package identifier to the jobs table but both other ways would be nicer (more consistent, less confusing) and should at least be considered.

…and I strongly agree to @karendolan that calling the ID just mediapackage is highly confusing ;-P

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1643,10 +1643,11 @@ public Source getEventSource(Event event) {
 
   @Override
   public Opt<WorkflowInstance> getCurrentWorkflowInstance(String mpId) throws IndexServiceException {
-    WorkflowQuery query = new WorkflowQuery().withMediaPackage(mpId);
+//    WorkflowQuery query = new WorkflowQuery().withMediaPackage(mpId);
```
lkiesow, line 1646: Can this be removed?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1643,10 +1643,11 @@ public Source getEventSource(Event event) {
...
     WorkflowSet workflowInstances;
     try {
-      workflowInstances = workflowService.getWorkflowInstances(query);
+//      workflowInstances = workflowService.getWorkflowInstances(query);
```
lkiesow, line 1649: Same here.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -1776,6 +1800,35 @@ public int compare(JpaJob job1, JpaJob job2) {
...
+    try {
+      em = emf.createEntityManager();
+      query = em.createNamedQuery("Job.mediaPackage", JpaJob.class);
+      query.setParameter("mediapackage", mediapackage);
+      List<JpaJob> jobs = query.getResultList();
+      for (JpaJob job : jobs) {
+        setJobUri(job);
+      }
+
+      return $(jobs).map(fnToJob()).toList();
```
lkiesow, line 1823: Can we avoid adding more usages of the entwine functional library?

If you want to use functional code, Java 8 supports this natively:
```java
jobs.stream().map(JpaJob::toJob).collect(Collectors.toList());
```

Or you could go the more traditional route and use something like:
```java
List<Job> jobs = new ArrayList<>();
for (JpaJob job : query.getResultList()) {
  setJobUri(job);
  jobs.add(job.toJob());
}
return jobs;
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1034,11 +1023,14 @@ public void remove(long workflowInstanceId) throws WorkflowDatabaseException, No
...
-      query.withId(Long.toString(workflowInstanceId));
-      WorkflowSet workflows = index.getWorkflowInstances(query, Permissions.Action.READ.toString(), false);
-      if (workflows.size() == 1) {
-        WorkflowInstance instance = workflows.getItems()[0];
+      Job job = serviceRegistry.getJob(workflowInstanceId);
+      if (Status.DELETED.equals(job.getStatus())) {
+        throw new NotFoundException("Workflow '" + workflowInstanceId + "' has been deleted");
+      }
+      if (!Operation.START_WORKFLOW.toString().equals(job.getOperation())) {
+        throw new NotFoundException("Job '" + workflowInstanceId + "' is an operation, not a workflow instance and thus cannot be deleted");
```
lkiesow, line 1031: You actually do not know what it is. You just know that it's no workflow ;-P

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1390,7 +1388,7 @@ public void update(final WorkflowInstance workflowInstance) throws WorkflowExcep
           messageSender.sendObjectMessage(WorkflowItem.WORKFLOW_QUEUE, MessageSender.DestinationType.Queue,
                   WorkflowItem.updateInstance(workflowInstance, dcXml, accessControlList));
         }
-        index(workflowInstance);
+//        index(workflowInstance);
```
lkiesow, line 1391: should be removed

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1442,7 +1440,11 @@ protected void index(final WorkflowInstance workflowInstance) throws WorkflowDat
    */
   @Override
   public long countWorkflowInstances() throws WorkflowDatabaseException {
-    return index.countWorkflowInstances(null, null);
+    try {
+      return serviceRegistry.countWorkflows();
```
lkiesow, line 1444: I'm just wondering if it wouldn't make sense to move the code of `serviceRegistry.countWorkflows()` here. In a way, it's weird that workflow specific code is not in the workflow service but in the service registry.

modules/workflow-service-impl/src/test/java/org/opencastproject/workflow/impl/WorkflowServiceImplTest.java
```diff
@@ -993,6 +993,7 @@ private OrganizationDirectoryService setupMockOrganizationDirectoryService() {
    *           if anything fails
    */
   @Test
+  @Ignore
```
lkiesow, line 996: Don't just ignore all tests.
Fix them or remove them.

lkiesow (commented):
Here are my comments :-)

> should we use elasticsearch where we can? etc.

I don't think a search index is required for this. It kind of made sense when it was directly powering the admin interface but that's mostly not the case any longer (except for very specific queries). Adding Elasticsearch will make things just more complex.

slampunk:
Hi All,

Thanks for the comments, pointers and recommendations! Hopefully I'll be able to implement them to get this patch into tip top shape.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
slampunk:
```
Keep the generic jobs and use oc_job_argument to store the media package 
identifier instead...
```
I had initially wanted to go down this route. Saving the media package identifier would be easy enough as an argument could pretty much be anything we'd like it to be.
But it left me with these questions:

- the `argument` column in `oc_job_argument` is of type `longtext`, and I'm not too clued up on how this affects indexing (I mean, we'd probably want to add an index on the mediapackage ids right?),
- suppose we query a (Jpa)Job, lets call it `job`. One of the entries in the list `job.arguments` would be the media package identifier. As a first approach, I'd probably use a regex to determine the `mpId`. But because we can add as many arguments as we'd like (those could be additional `mpId`s as arguments), we need further checks to assure that the mapping between workflows and media packages is well-defined. Ultimately, I'd prefer an approach that assures a well-defined mapping than having to code those checks.

```
Separate jobs and workflows...
```

It definitely would be more work, but imo _that_ shouldn't be a determining factor... code tends to stick around so do it right the first time round.

If you don't mind, I'll get you on IRC this week for a chat on best strategies for this patch.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
JamesUoM: aren't job argument getting deleted at the end of workflows, IIRC there was a PR to to remove them as part of workflow cleanup? If so a separate mh_mediapackage_workflow (string  mpId; long jobId) table.

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
JamesUoM: Ignore it, its a temporary ignore until the code is complete :grin:

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
smarquard: From a database performance point of view, the fastest is probably a separate table, with 2nd-best adding the column to mh_job. I don't think adding it into mh_job_argument, as requires that require joins to get what they need with very large tables can be suboptimal (as seen with various asset manager db performance issues at present).

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/AbstractEventEndpointTest.java
karendolan: Ok, I'm good with keeping the @ignore. Event though, IMHO, adding the //TODO makes it more obvious that this is an actual desired test which is temporarily under reconstruction and can be and *should* be implemented by someone, even if it is accidentally missed in this pull.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
staubesv: Without digging into it: I would not use the oc_job_argument table as the media package id is not really a job argument, is it? 
Without knowing the details, the most preferable solutions seem 1) to factor out workflows from the job table into an own table or 2) use a separate table to establish the relationship of media packages to workflows (mh_job_media_package consisting of the columns media_package_id and job_id) as @smarquard suggests.

lkiesow: Yep, that should work well.

KatrinIhler:
@slampunk Note that this PR has test failures.

smarquard:
This PR is looking for a new owner to complete the work, as @slampunk left UCT at the end of Dec 2018 and is no longer available to finalize it.

smarquard:
UCT has contracted ELAN to complete this work.

KatrinIhler:
I've picked this up and will file a new pull request as a draft as soon as possible.

PR #603, opened 2018-12-03
MH-13221 Open single-select with single click

KatrinIhler:
Open a single-select drop-down with a single click instead of having to click twice. No longer react to clicks outside of the span representing the drop-down menu.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableSingleSelectDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableSingleSelectDirective.js
@@ -99,6 +99,8 @@ angular.module('adminNg.directives')
       scope.collection = scope.ordered ? mapToArrayOrdered(scope.collection, scope.params.translatable) :
         mapToArray(scope.collection, scope.params.translatable);
 
+      scope.editMode = false;
+
       scope.submit = function () {
         // Wait until the change of the value propagated to the parent's
         // metadata object.
@@ -124,21 +126,30 @@ angular.module('adminNg.directives')
         $timeout.cancel(scope.submitTimer);
       });
 
+      scope.toggleEditMode = function () {
+
+        if (scope.editMode == false) {
+          scope.enterEditMode();
+        } else {
+          scope.leaveEditMode();
+        }
+      };
+
       scope.enterEditMode = function () {
+
         // Store the original value for later comparision or undo
         if (!angular.isDefined(scope.original)) {
           scope.original = scope.params.value;
         }
         scope.editMode = true;
-        scope.focusTimer = $timeout(function () {
+        $timeout(function () {
           if ($('[chosen]')) {
-            element.find('select').trigger('chosen:activate');
+            element.find('select').trigger('chosen:open');
           }
         });
       };
 
       scope.leaveEditMode = function () {
-        // does not work currently, as angular chose does not support ng-blur yet. But it does not break anything
         scope.editMode = false;
       };
     }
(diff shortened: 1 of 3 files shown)


staubesv:
@KatrinIhler In 2017, somebody did a patch on behalf of ETHZ (see https://opencast.jira.com/browse/MH-11849) to ensure that left-clicking into a metadata field will activate the field for editing. Limiting this function to left-clicks on the span element is indeed quite annoying if you have small texts (e.g. CC-BY) and is not consistent with the behaviour of the other metadata fields.

In this context: I'm just thinking about a patch that makes the drop downs 100% width as we have a lot of content which does not fit onto a single text line now. Maybe, the span gets 100% then, too?

What do you think about that?

KatrinIhler:
@staubesv The reason I disabled the left-clicking for the whole div was that it was creating other issues in conjuction with my changes (as this code often does), but yes, making the drop-downs take 100% of the width could solve these and sounds like a good idea overall. Do you want to take care of this and then I will update my PR accordingly?

staubesv:
@KatrinIhler The following two changes make this work with 100% drop down list width:

1. Revert https://github.com/opencast/opencast/pull/603/files#diff-9b8ee9895b36b3de07957518e1330e82L1
2. Change data-with to '100%' in https://github.com/opencast/opencast/pull/603/files#diff-9b8ee9895b36b3de07957518e1330e82R16

My feelings about the 100% width are ambivalent: Limiting the width to 250px is suboptimal as we have quite a number of values that require multi-line representation (e.g. series titles, licences) in our setup. On the other hand, having 100% width is quite width for very short names (e.g. licences in Opencast, Languages).

What do you think?

KatrinIhler:
@staubesv  I've put some more thought into this and I think what we actually want is to have a minimum width for the dropdowns similar to what we have now, and make the menus broader if an element doesn't fit, up to 100%. But that is not in the scope of this PR. Instead I have decided to make the whole div clickable again and made some more code adjustments to make it work. This has the added benefit that it makes leaveEditMode() work, which didn't previously. Check it out and tell me what you think!

modules/admin-ui/src/main/webapp/scripts/shared/directives/editableSingleSelectDirective.js
```diff
@@ -124,21 +126,30 @@ angular.module('adminNg.directives')
...
+        $timeout(function () {
           if ($('[chosen]')) {
-            element.find('select').trigger('chosen:activate');
+            element.find('select').trigger('chosen:open');
           }
         });
       };
 
       scope.leaveEditMode = function () {
-        // does not work currently, as angular chose does not support ng-blur yet. But it does not break anything
```
staubesv, line 141: Unfortunately, this still holds, although the comment is not correct: It is jquery-chosen that makes it impossible for angular-chosen to implement ng-blur (angular-chosen cannot fix this, it is a limitation of jquery-chosen).
Your idea to focus the pencil icon instead of the chosen dropdown would make this very, very nasty bug less important, but it still would exist: Left-click a chosen dropdown and then press the tab key will not call leaveEditMode when losing the keyboard focus.

staubesv:
@KatrinIhler Personally, I like the idea of not focusing on the chosen dropdowns when using the tab key as I consider this (not leaving edit mode for chosen dropdowns when losing focus) one of the most obvious and nasty bug of the admin UI - of course, it can be safely ignored, but what would you think of a product that has such obvious bugs?

What currently is missing: Now, the metadata editor cannot be operated by keyboard only as the user cannot open the dropdown using keys. I could live with that, but I think that there are other opinions about this out there ;-) 

Maybe just react to space, left/right arrow to enter edit mode so that keyboard only editing still works?

staubesv:
@KatrinIhler Hmm.... At this very moment of time, I think it might be a good idea to just stick with your original solution (that limits left-clicking to the content). Finally, not asking the user to have to click twice to select seems a good improvement and the left-click inconsistency seems less important.

What do you think?

KatrinIhler:
@staubesv An upated version of this PR is in the works, but not quite finished yet. That will hopefully improve things...

staubesv:
@KatrinIhler Thanks for the info! Looking forwards to an updated version of this PR.

staubesv:
@KatrinIhler Any news on this? Maybe close this PR and create a new one once a solution is available?

KatrinIhler:
@staubesv I'm planning to work on this today, but yeah, I'm gonna create a new pullrequest then, so I'll close this for now!

PR #606, opened 2018-12-04
MH-13266 Start date cross link does not work correctly

staubesv:
Left-clicking a date in the column "Date" in the table "Events" is supposed to enable the filter "Start date" for the date that has been clicked.

This feature has been recently broken: Left-clicking any date in the column "Date" will just activate the filter "Start date" for the current date of the day.

Steps to reproduce:
1. Left-click on the start date in the events table for a start date that is not "Today"
 
 Actual Results:
The crosslink for the start date will always enable the filter for the date "today". 
 
 Expected Results:
 The crosslink for the start date should enable the filter for the date "start date"
 
 Workaround (if any):


staubesv:
@KatrinIhler This is a regression bug caused by #348 which I didn't see during the code review. Could you please take the review?

lkiesow:
Can you add a short description to this pull request and fix the quoting issue?
The patch looks good otherwise.

KatrinIhler (approved):
Whoops, I don't know what I was thinking when I wrote that... Tested this and it works!

PR #608, opened 2018-12-05
MH-13269, Handle Authorization Errors

lkiesow:
A user trying to modify an ACL to which he does not have access causes
an internal server error in Opencast. This patch converts it to result
in a “forbidden” user error instead.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -728,8 +728,10 @@ public abstract class AbstractEventEndpoint {
         return ok();
       }
     } catch (AclServiceException e) {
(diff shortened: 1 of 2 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -728,8 +728,11 @@ public Response applyAclToEvent(@PathParam("eventId") String eventId, @FormParam
         return ok();
       }
     } catch (AclServiceException e) {
-      logger.error("Error applying acl '{}' to event '{}' because: {}",
-              accessControlList, eventId, ExceptionUtils.getStackTrace(e));
+
+      if (e.getCause() instanceof  UnauthorizedException) {
```
staubesv: @lkiesow Please remove the superfluous whitespace after instanceof to avoid further PRs that fix this cosmetic issue.

staubesv (approved):

PR #609, opened 2018-12-05
MH-13270, Fix Message Item Serialization

lkiesow:
This patch fixes a number of forbidden data type usages in the message
broker's item serialization.

### DIFF
--- a/modules/message-broker-api/pom.xml
+++ b/modules/message-broker-api/pom.xml
@@ -53,6 +53,10 @@
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.compendium</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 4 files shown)


lkiesow:
Since you found the problem, @krinnewitz, would you mind reviewing this patch?

staubesv:
@lkiesow Just wondering: Would the serialization of **b**oolean (vs. **B**oolean) work? Also, it seems somewhat overkill to me to use Gson to convert a string to a boolean and vice versa. Maybe just use Boolean.toString and Boolean.valueOf?

krinnewitz:
Thanks @lkiesow, I will test it. Of course. using Gson to serialize the Boolean is somewhat ugly, but it looks like you did that to allow for `null` to be transmitted, correct?

lkiesow:
@staubesv native data types like `boolean` work just fine. Objects like `Boolean` not without whiltelisting their package.

For using Gson: As @krinnewitz suspected, I'm using that to avoid all the null checks required otherwise. We have the serializer set-up and ready to go anyway.

EDIT: For the reason that the values may be null, I cannot use `boolean` for things like `optOut`, … in case you were wondering.

staubesv:
@lkiesow Haven't realised that "null" is used here, thanks for the explanation. Will continue the review...

krinnewitz:
Oh, I didn't notice that @staubesv was so kind to take the review. I won't test this, then ;-)

staubesv (approved):

PR #610, opened 2018-12-06
MH-13271, Remove Useless ACL Check

lkiesow:
This patch removes a useless check where the access control list used
for verifying the privileges of the current user to publish an event is
provided by the user itself right in the request. So there is no
security at all. It's just a useless check.

*Work sponsored by SWITCH*


modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -361,10 +360,6 @@ public void addSynchronously(MediaPackage mediaPackage) throws SearchException,
           IllegalArgumentException, UnauthorizedException {
     User currentUser = securityService.getUser();
```
staubesv: @lkiesow The variables currentUser and orgAdminRole are not used anymore after you remove the useless check.

lkiesow: good catch. removed.

PR #611, opened 2018-12-06
MH-13272 fix missing roles

mtneug:
Fix InMemoryUserAndRoleProvider that filtered out users and roles if the limit was set to 0 (= no limit).


lkiesow:
Ups, seems like I actually broke this. Sorry.
Merged.

staubesv:
Well, seems I have not seen this during the code review. Sorry for that, too...

PR #612, opened 2018-12-06
MH-13274 search by enddate  using filter=endDate:left/right

phoebem:
This adds back the functionality of "endsfrom" and "endsto" in the 1.x /recordings/recordings.xml

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -2341,6 +2341,15 @@ public abstract class AbstractEventEndpoint {
           return RestUtil.R.badRequest(e.getMessage());
         }
       }
+      if (EventListQuery.FILTER_ENDDATE_NAME.equals(name)) {
+        try {
+          Tuple<Date, Date> fromAndToCreationRange = RestUtils.getFromAndToDateRange(filters.get(name));
+          query.withTechnicalEndFrom(fromAndToCreationRange.getA());
+          query.withTechnicalEndTo(fromAndToCreationRange.getB());
+        } catch (IllegalArgumentException e) {
+          return RestUtil.R.badRequest(e.getMessage());
+        }
+      }
     }
 
     if (optSort.isSome()) {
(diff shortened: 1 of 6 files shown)


staubesv:
@phoebem The code works as expected. The only thing I don't understand is what a Admin UI user is supposed to do with a filter that filters events by end date. It seems like adding one more useless filter to the admin ui to me, or is there really a use case for this?

If you need this for an external application only as the JIRA issue suggests, would you mind adding the filter to the External API instead?
It is quite simple to add that filter to the External API (see technical_start filter on https://github.com/opencast/opencast/blob/develop/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java#L670, no need to put this into a list provider) and also not really a best practise to use the Admin UI facade as Opencast API as we officially declared the Admin UI facade as internal API.

phoebem:
@staubesv Thanks for reviewing this.
This endpoint and the capture agent endpoint changes are added in preparation for a live monitor UI that will come later.
The live monitor is built as part of admin-ui.
Our producers find the 1.x version of the monitor essential and asked that it goes into our 5.x cluster. 
Hopefully the community will find it useful too.
I want to keep each PR small and focused so that they are more manageable.

As for the external API, my understanding is that it uses a different index and can go out of sync with the admin index.

staubesv:
@phoebem Can we maybe set this onhold until the monitoring PR is available then?

phoebem:
The monitoring UI won't be ready for a while as we are still using rtmp and the version of the player in the admin ui does not work with rtmp, so we use a separate module.
We are transitioning to HLS and so it will be a better product when it is better integrated into system.

phoebem:
I am still working on an HLS live monitor. I should have something in a few weeks.

turro:
Hi @phoebem

I don't know if you need viewing the video, but just in case just to say that we are about releasing paella 6.1.0 with a lot of HLS improvements.

I expect it for next week, we are currently testing

phoebem:
Thanks Carlos, I will definitely check it out.
By the way, does Paella support the DVR feature for live streams?

Phoebe

staubesv:
@phoebem As the live monitor seems not to make it into Opencast 7.0 (feature freeze on Monday), I've labelled this PR as "onhold" so that we can review all related PRs together.

KatrinIhler:
@phoebem What do you think about closing this for now and refiling once the live monitor is ready?

phoebem:
Let’s so that. 
The latency from using HLS for live monitor was too big for our users,
so I don’t think that we will move away from rtmp any time soon. 
Our code is based on rtmp, so I don’t know useful it is to the community.

Phoebe.

staubesv:
@phoebem Why do you think RTMP is not interesting for the community for live monitoring?

PR #613, opened 2018-12-06
MH-13276, Drop org.opencastproject.fun

lkiesow:
Getting rid of one custom functional library…


pmiddend:
I can review this.

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageSupport.java
```diff
@@ -539,21 +540,21 @@ public String apply(MediaPackage mp) {
     public static final Function<MediaPackage, List<MediaPackageElement>> getElements = new Function<MediaPackage, List<MediaPackageElement>>() {
       @Override
       public List<MediaPackageElement> apply(MediaPackage a) {
-        return Immutables.list(a.getElements());
+        return Arrays.asList(a.getElements());
```
pmiddend, line 543: Very strictly speaking, you're losing something here, since `Arrays.asList` returns a mutable `ArrayList`. But I suppose `Collections.unmodifiableList(Arrays.asList(...))` is as ugly as it gets, so your solution fine with me. :)

modules/common/src/main/java/org/opencastproject/serviceregistry/api/Incidents.java
```diff
@@ -160,14 +161,14 @@ public void recordFailure(Job job, int code, Throwable t, List<Tuple<String, Str
    */
   public void recordFailure(Job job, int code, Throwable t, Map<String, String> params,
           List<Tuple<String, String>> details) {
-    List<Tuple<String, String>> detailList = Mutables.list(details);
+    List<Tuple<String, String>> detailList = new LinkedList<>(details);
```
pmiddend: I don't really care per se, but the "default" List implementation for Java seems to be `ArrayList` rather than `LinkedList`, probably because of better memory alignment and thus caching performance (see `Collectors.toList()`, which returns an `ArrayList`, for example, just as `toSet()` returns a `HashSet` instead of a `TreeSet`). But again, no blocker for me.

modules/common/src/main/java/org/opencastproject/util/persistence/PersistenceUtil.java
```diff
@@ -400,9 +398,9 @@ public static EntityManagerFactory newEntityManagerFactory(String emName, String
     pooledDataSource.setPassword(pwd);
 
     // Set up the persistence properties
-    final Map<String, Object> props = Immutables.<String, Object> map(persistenceProps,
-            tuple("javax.persistence.nonJtaDataSource", pooledDataSource),
-            tuple("eclipselink.target-database", vendor));
+    final Map<String, Object> props = new HashMap<>(persistenceProps);
```
pmiddend, line 401: Again, strictly speaking, we're losing the immutable guarantee here.

modules/incident-workflowoperation/src/main/java/org/opencastproject/workflow/handler/incident/IncidentCreatorWorkflowOperationHandler.java
```diff
@@ -72,13 +77,18 @@ public WorkflowOperationResult start(WorkflowInstance wi, JobContext ctx) throws
...
+            StringUtils.split(woi.getConfiguration(OPT_DETAILS), ";")))
+            .map((opt) -> opt.split("="))
+            .filter((t) -> t.length == 2)
+            .map((x) -> Tuple.tuple(x[0], x[1]))
+            .collect(Collectors.toList());
+    final Map<String, String> params = Arrays.stream(ArrayUtils.nullToEmpty(
+            StringUtils.split(woi.getConfiguration(OPT_PARAMS), ";")))
+            .map((opt) -> opt.split("="))
+            .filter((t) -> t.length == 2)
+            .collect(Collectors.toMap(x -> x[0], x -> x[1]));
```
pmiddend, line 91: Note that there is an extremely subtle difference between `Immutables` and the new code: the new code will fail if there is a duplicate key (because `toMap` checks this), whereas with `Immutables`, the latest key "wins". I'm not sure this is important though.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/AbstractIncidentService.java
```diff
@@ -279,15 +277,13 @@ public Incident apply(IncidentDto dto) {
...
-    final List<String> keys = mlist(locale).foldl(Mutables.list(base),
-            new Function2<List<String>, String, List<String>>() {
-              @Override
-              public List<String> apply(List<String> sum, String s) {
-                sum.add(sum.get(sum.size() - 1) + "." + s);
-                return sum;
-              }
-            });
-    return mlist(keys).reverse().value();
+    List<String> result = new ArrayList<>();
```
pmiddend: If you want to micro-optimize here: `LinkedList` plus `StringBuilder`. But unless this functions is called very often, it shouldn't matter.

pmiddend:
I added some comments, but there are no blockers. The code compiles, Opencast still seems to work and the tests succeed. This can be merged.

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageSupport.java
lkiesow: True, but I deliberately ignored that at multiple places since we do not really gain anything by making it immutable. It's a new list anyway and people could modify and further process it if they want to. It wouldn't have any effect on the media package. Hence, I'm going the easy route.

I will go through the other comments in a moment though and update a few of them.

modules/incident-workflowoperation/src/main/java/org/opencastproject/workflow/handler/incident/IncidentCreatorWorkflowOperationHandler.java
lkiesow: Good catch but I'm going to ignore that since I think this behavior is actually preferable. It shows a configuration error instead of silently ignoring it and doing something ;-)

lkiesow:
Thanks for your review, @pmiddend.
I've merged this based on your approval.
I'm looking forward to seeing more fn cleanup in the future ;-P

PR #604, opened 2018-12-04
MH-12177 TimeZone threadsafe and bulk schedule across DST (6x)

karendolan:
This pull is the old pull https://github.com/opencast/opencast/pull/189 but applied to r/6.x branch instead of to develop branch.

          MH-12177 fix TimeZone non-threadsafe issue in scheduler-api, scheduler-impl.
          Update ical4j lib (from net.fortuna.ical4j 1.0-rc2-patched to org.mnode.ical4j 2.2.0).
          DST border crossing in Recurrence inbound start and end dates must be in schedule to
                specific TimeZone, not be converted to UTC, in order for correct days of week
                and DST changeover to work with org.mnode.ical4j (new name of net.fortuna.ical4j).
          Move scheduler Util test to scheduler-api (the actual home of Scheduler Util).
          Bulk schedule to the specified timezone or fallback to local timezone.

PR #614, opened 2018-12-07
MH-13277 improve scheduler performance

krinnewitz:
This patchset improves the scheduler performance as part of the opencast crowdfunding 2018. As a benchmark, we scheduled 50k events and then added 6 events on top. Previously, adding the 6 events took between 19 and 20 seconds. Of these 20 seconds, the conflict check took about 80% of the time. After our changes, it takes about 2 seconds to add the 6 events on top. Of these 2 seconds, the conflict check now only takes a very small percentage of time. The benchmark was done using Opencast with MariaDB on an AMD FX(tm)-8350 Eight-Core Processor with 32GB RAM and an SSD.

The following changes were done to achieve the goal of faster conflict checks and higher throughput when inserting events:

Remove Unused YAGNI Code
-----------------------------------------

- The "transactional" feature of the scheduler was only used for the migration from opencast 3.x to 4.x and introduced a lot of complexity. In order to improve the maintainability of opencast, we decided to remove this code. If anyone wants to utilize it someday, it can be restored from the git history.
- The "blacklist" feature of the scheduler wasn't fully implemented. We removed it as well, again for the sake of maintainability.

Store Properties of Scheduled Events in a Dedicated Table
--------------------------------------------------------------------------------

Properties such as start date, end date, capture device id, of scheduled events were stored as asset manager properties which are basically key-value pairs which cannot be stored efficiently using a relational DBMS. For this reason, the conflict check was done in-memory, which was a workaround for the slow database queries. However, this still was slow since all events had to be loaded into memory before performing the conflict check. The scheduler now uses a proper "SQL-friendly" table to store these properties. This drastically improves the conflict check performance (which can be done on the database again) as well as the insertion speed when adding new events.

Reduce Number of ActiveMQ Messages When Adding Events
-------------------------------------------------------------------------------------

When scheduling multiple events, one message per event was sent over ActiveMQ. This seemed to cause some buffers filling up and blocking the message sender which in turn slowed down event insertion. We changed the behavior to send just one message for a whole bunch of events, which solves this problem.

Parallelize Adding Multiple Events
-----------------------------------------------

When scheduling multiple events (using an RRULE), these events cannot have conflicts with each other. This allows us to parallelize event insertion in this case, which also improves the throughput.

Other Changes
---------------------

Additionally, we improved the test coverage of the scheduler and added a migration which is necessary because of the database schema changes,


JamesUoM:
Is there any way to store custom properties, in scheduled events? We had planned to tighter integrate our Participation Manager (opt-out) with the using Asset managers properties to store addition choices like editing, camera and or screen recording, without having to parse the workflow and capture agent configs.

krinnewitz:
It depends.If you know the keys and types of your properties in advance, you can just add  one column per property to oc_scheduled_extended_event and store them there.

If the properties you want to store are arbitrary and don't have to be queried individually, you can add another string column, serialize the properties using gson and store them in this column. For reading those properties, you just read the string from the database and de-serialize them again. Just like we did it for the capture agent and workflow properties. 

The problem with the workflow and capture agent properties was, that they can consist of arbitrary key-value pairs, so we couldn't create one column per property. But since these properties weren't queried individually, we could just serialize them and store them as a whole.

And of course, you can still save and load asset manager properties. However, using the asset manager properties is expensive.

staubesv:
@JamesUoM & @krinnewitz Opencast 6.x introduces a new processing settings persistent layer that uses the asset manager to store workflow properties. At some point in the future, it likely makes sense to remove the workflow properties from the scheduler table and make the scheduler use the processing settings persistent layer of the asset manager. At the end of the day, a scheduled event is a event and it would be perfectly fine to store the workflow properties just for the event (vs. scheduled event).

staubesv:
@krinnewitz Please also adopt the page https://docs.opencast.org/develop/developer/scheduler/

krinnewitz:
Thanks @staubesv. I updated the documentation. 

Regarding the workflow properties: Maybe we could change that in a follow-up PR if there are not too many review findings with this one. However, the problem with storing anything in the asset manager properties is, that even just insertion is quite slow. So when scheduling multiple events at once, that could already be a problem.

JamesUoM:
> Thanks @staubesv. I updated the documentation.
> 
> Regarding the workflow properties: Maybe we could change that in a follow-up PR if there are not too many review findings with this one. However, the problem with storing anything in the asset manager properties is, that even just insertion is quite slow. So when scheduling multiple events at once, that could already be a problem.

Appologies for the diversion:

We schedule events for two weeks ahead which can be 2000+ new events being added on a Monday. If we moved away from our intermediate table of via recordings and user options is then 25,000 events at the start of the semester and a busy 72 hours when the majority opt-out decisions occur. I think we're likely to keep our user options in our own Participation Management (PM) tables if the Asset Manager was not able to cope with this, and I would be very concerned if the creation and modification performance of CA and WF configs was negatively impacted.

Very simply we ingest the whole time table and create potential recordings against which users opt-out, this is synchronized daily as the timetable changes often. We then created (delete and modify) scheduled events for a rolling two week window synchronized on a daily basis.

```
Timetable --1 semester--> PM --2 weeks--> OC
                          ^^
                     User options
```

gregorydlogan:
I've tagged this as a feature, even though it's not *really* a feature.  It *does* however warrant the extra testing that comes with being a feature IMO.

lkiesow:
Sorry for the late start of the review.

> When scheduling multiple events (using an RRULE), these events cannot have conflicts with each 
> other. This allows us to parallelize event insertion in this case, which also improves the throughput.

Does Opencast's internal processing of RRULES differ from what's possible in iCalendar? Because I think this would be a completely valid schedule with overlapping events using an RRULE (12 events á 2h scheduled hourly):

```ical
BEGIN:VEVENT
DTSTAMP:20190113T185826Z
UID:20190113T185826Z-1469192696
DTSTART;TZID=Europe/Berlin:20190201T120000
DTEND;TZID=Europe/Berlin:20190201T140000
RRULE:FREQ=HOURLY;INTERVAL=1;COUNT=12
SUMMARY:Test
DESCRIPTION:Some overlapping test events
END:VEVENT
```

pmiddend:
@lkiesow You're right. I hope the admin UI actually prevents this from happening, but surely there are other ways to "inject" such an RRULE into the system. Either way, I've added some code to check those overlaps.

docs/scripts/ddl/mysql5.sql
```diff
@@ -242,21 +242,28 @@ CREATE INDEX IX_oc_scheduled_last_modified_last_modified ON oc_scheduled_last_mo
...
+  capture_agent_id VARCHAR(128) NOT NULL,
+  start_date DATETIME NOT NULL,
+  end_date DATETIME NOT NULL,
+  source VARCHAR(255),
+  recording_state VARCHAR(255),
+  recording_last_heard BIGINT,
+  review_status VARCHAR(128),
+  review_date DATETIME,
+  presenters TEXT(65535),
+  optout TINYINT(1),
```
lkiesow, line 246:
```
review_status VARCHAR(128),
review_date DATETIME,
presenters TEXT(65535),
optout TINYINT(1),
last_modified_origin VARCHAR(255)
```
Are these used anywhere?

docs/scripts/ddl/mysql5.sql
```diff
@@ -242,21 +242,28 @@ CREATE INDEX IX_oc_scheduled_last_modified_last_modified ON oc_scheduled_last_mo
...
+  last_modified_origin VARCHAR(255),
+  last_modified_date DATETIME,
+  checksum VARCHAR(64),
+  capture_agent_properties MEDIUMTEXT,
+  workflow_properties MEDIUMTEXT,
```
lkiesow, line 250: Aren't these now stored in the asset manager?
Is this meant as cache?

docs/upgrade/6_to_7/mysql5.sql
```diff
@@ -1 +1,28 @@
...
+  checksum VARCHAR(64),
+  capture_agent_properties MEDIUMTEXT,
+  workflow_properties MEDIUMTEXT,
+  PRIMARY KEY (mediapackage_id, organization),
+  CONSTRAINT FK_oc_scheduled_extended_event_organization FOREIGN KEY (organization) REFERENCES oc_organization (id) ON DELETE CASCADE
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+CREATE INDEX IX_oc_scheduled_extended_event_organization ON oc_scheduled_extended_event (organization);
+CREATE INDEX IX_oc_scheduled_extended_event_capture_agent_id ON oc_scheduled_extended_event (capture_agent_id);
+CREATE INDEX IX_oc_scheduled_extended_event_dates ON oc_scheduled_extended_event (start_date, end_date);
+DELETE FROM oc_assets_properties WHERE namespace = 'org.opencastproject.scheduler.trx';
```
lkiesow: Can we just drop those? don't we need to move them to their new table?

docs/guides/admin/docs/upgrade.md
```diff
@@ -60,4 +60,23 @@ select * from oc_assets_snapshot where mediapackage_xml like '%"security/xacml"%
...
-- *TODO*
+- `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` no longer needs the `transaction_cleanup_offset`
+  option.
+- `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` has a new option `maintenance` which temporarily
+  disables the scheduler if set to `true`.
+
+Scheduler Migration
+-------------------
+
+The way the Scheduler stores its data was changed in Opencast 7 to improve performance when checking for conflicts.
```
lkiesow, line 71: Do you always have to run the migration or only if you have active schedules (scheduled events which have not yet been recorded)?

modules/migration/src/main/java/org/opencastproject/migration/EventIdPK.java
```diff
@@ -0,0 +1,68 @@
...
+    return mediaPackageId;
+  }
+
+  public String getOrganization() {
+    return organization;
+  }
+
+  @Override
+  public int hashCode() {
+    return EqualsUtil.hash(mediaPackageId, organization);
```
lkiesow: I think this would do the same without custom code:

```suggestion
    return Objects.hash(mediaPackageId, organization);
```

modules/migration/src/main/java/org/opencastproject/migration/EventIdPK.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == this)
+      return true;
+    if (!(obj instanceof EventIdPK))
+      return false;
+    EventIdPK pk = (EventIdPK) obj;
+    return EqualsUtil.eq(pk.getMediaPackageId(), mediaPackageId) && EqualsUtil.eq(pk.getOrganization(), organization);
```
lkiesow:
```suggestion
    return Objects.equals(pk.getMediaPackageId(), mediaPackageId) && Objects.equals(pk.getOrganization(), organization);
```

modules/migration/src/test/java/org/opencastproject/migration/SchedulerMigrationServiceTest.java
```diff
@@ -25,111 +25,268 @@
...
-import javax.sql.DataSource;
+import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
 
 public class SchedulerMigrationServiceTest {
 
-  /**
-   * Test class for the scheduler migration service
-   */
+  private static final File baseDir = new File(new File(IoSupport.getSystemTmpDir()), "schedulerservicetest");
```
lkiesow: Maybe use `TemporaryFolder`? We had a lot of problems with not properly cleaned up, self created tmp folders in the past.

https://github.com/opencast/opencast/blob/5a3437929bca3c5e05565331bfe192f2d70d2c8f/modules/search/src/test/java/org/opencastproject/matterhorn/search/impl/SearchIndexTest.java#L64-L65

https://github.com/opencast/opencast/blob/5a3437929bca3c5e05565331bfe192f2d70d2c8f/modules/search/src/test/java/org/opencastproject/matterhorn/search/impl/SearchIndexTest.java#L76

modules/migration/src/test/java/org/opencastproject/migration/SchedulerMigrationServiceTest.java
```diff
@@ -25,111 +25,268 @@
...
+    assertEquals("source of mp2", entity.getSource());
+    assertEquals("presenter of mp2", entity.getPresenters());
+    assertEquals("agent of mp2", entity.getCaptureAgentId());
+    assertEquals(new Date(202), entity.getStartDate());
+    assertEquals(new Date(203), entity.getEndDate());
+    assertFalse(entity.getOptOut());
+    assertEquals("origin of mp2", entity.getLastModifiedOrigin());
+    assertEquals(new Date(204), entity.getLastModifiedDate());
+    assertEquals("checksum of mp2", entity.getChecksum());
+    map = gson.fromJson(entity.getCaptureAgentProperties(), new HashMap<String, String>().getClass());
```
lkiesow: Suggestion (less unchecked casting + don't instantiate an object to get the class):

```java
Map map;
…
map = gson.fromJson(entity.getCaptureAgentProperties(), HashMap.class);
assertEquals("ca prop 1", map.get("agent testproperty"));
map = gson.fromJson(entity.getWorkflowProperties(), HashMap.class);
assertEquals("wf prop 1", map.get("workflow testproperty"));
…
map = gson.fromJson(entity.getCaptureAgentProperties(), HashMap.class);
assertEquals("ca prop 2", map.get("agent testproperty"));
map = gson.fromJson(entity.getWorkflowProperties(), HashMap.class);
assertEquals("wf prop 2", map.get("workflow testproperty"));
```

modules/migration/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+log4j.logger.org.mysema.querydsl=DEBUG,stdout
```
lkiesow: I don't think this is actually used, is it?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1036,6 +1009,68 @@ public Response updateEvent(@PathParam("id") String eventID, @FormParam("start")
...
+      Opt<MediaPackage> current = service.getCurrentRecording(agentId);
+      if (current.isNone()) {
+        return Response.noContent().build();
+      } else {
+        return Response.ok(MediaPackageParser.getAsXml(current.get())).build();
+      }
+    } catch (UnauthorizedException e) {
+      throw e;
+    } catch (Exception e) {
+      logger.error("Unable to get the current recording for agent '{}': {}", agentId, e);
```
lkiesow:
```suggestion
      logger.error("Unable to get the current recording for agent '{}'", agentId, e);
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1036,6 +1009,68 @@ public Response updateEvent(@PathParam("id") String eventID, @FormParam("start")
     }
   }
 
+  @GET
```
lkiesow, line 1000: are these used for anything?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1036,6 +1009,68 @@ public Response updateEvent(@PathParam("id") String eventID, @FormParam("start")
...
+      Opt<MediaPackage> upcoming = service.getUpcomingRecording(agentId);
+      if (upcoming.isNone()) {
+        return Response.noContent().build();
+      } else {
+        return Response.ok(MediaPackageParser.getAsXml(upcoming.get())).build();
+      }
+    } catch (UnauthorizedException e) {
+      throw e;
+    } catch (Exception e) {
+      logger.error("Unable to get the upcoming recording for agent '{}': {}", agentId, e);
```
lkiesow:
```suggestion
      logger.error("Unable to get the upcoming recording for agent '{}'", agentId, e);
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -1036,6 +1009,68 @@ public Response updateEvent(@PathParam("id") String eventID, @FormParam("start")
...
+  @Produces(MediaType.TEXT_PLAIN)
+  @RestQuery(name = "eventcount", description = "Get the number of scheduled events", returnDescription = "The number of scheduled events", reponses = {
+      @RestResponse(responseCode = HttpServletResponse.SC_OK, description = "The event count") })
+  public Response eventCount() throws UnauthorizedException {
+    try {
+      return Response.ok("" + service.getEventCount()).build();
+    } catch (UnauthorizedException e) {
+      throw e;
+    } catch (Exception e) {
+      logger.error("Unable to get the event count: {}", e);
```
lkiesow:
```suggestion
      logger.error("Unable to get the event count", e);
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -196,41 +179,24 @@
...
+  /**
+   * Deserializes properties stored in string columns of the extended event table
+   * @param props Properties as retrieved from the DB
+   * @return deserialized key-value pairs
+   */
+  private static Map<String,String> deserializeExtendedEventProperties(String props) {
+    if (props == null || props.trim().isEmpty()) {
+      return new HashMap<>();
+    }
+    return (Map<String, String>)new Gson().fromJson(props, Map.class);
```
lkiesow: You can have Gson deserialize into typed maps using something along the lines of 
```java
Type type = new TypeToken<Map<String, String>>(){}.getType();
Map<String, String> map = gson.fromJson(jsonString, type);
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -392,55 +368,21 @@ public void removeCatalogUIAdapter(EventCatalogUIAdapter catalogUIAdapter) {
...
-        } catch (SchedulerServiceDatabaseException e) {
-          logger.error("Unable to get transactions to cleanup incomplete transactions because", e);
-        }
-      });
-    }
-    logger.info("Finished checking for incomplete transactions from a shutdown or a restart.");
-  }
-
-  /** Callback from OSGi on service deactivation. */
+   /** Callback from OSGi on service deactivation. */
```
lkiesow: indentation

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -449,84 +391,14 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
         logger.info("Set last modified cache to default {}", getHumanReadableTimeString(DEFAULT_CACHE_EXPIRE));
       }
-      for (Integer offsetInSeconds : OsgiUtil.getOptCfg(properties, CFG_KEY_TRANSACTION_CLEANUP_OFFSET)
-              .bind(Strings.toInt)) {
-        transactionOffsetMillis = offsetInSeconds * DateTimeConstants.MILLIS_PER_SECOND;
+      final Option<Boolean> maintenance = OsgiUtil.getOptCfgAsBoolean(properties, CFG_KEY_MAINTENANCE);
+      if (maintenance.getOrElse(false)) {
+        final String name = SchedulerServiceImpl.class.getName();
+        logger.warn("Putting scheduler into maintenance mode. This only makes sense when migrating data. If this is not"
+                + " intended, edit the config file '{}.cfg' accordingly and restart opencast.", name);
```
lkiesow, line 403: 👍

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -732,12 +583,16 @@ private void addEventInternal(Date startDateTime, Date endDateTime, String captu
         mediaPackage.setIdentifier(id);
 
         // Update dublincore title and temporal
-        String newTitle = dc.getFirst(DublinCore.PROPERTY_TITLE) + String.format(" %0" + Integer.toString(periods.size()).length() + "d", ++counter);
+        String newTitle = dc.getFirst(DublinCore.PROPERTY_TITLE) + String.format(" %0" + Integer.toString(periods.size()).length() + "d", currentCounter + 1);
```
lkiesow, line 587: If this generates titles with ascending numbers, couldn't the parallelStream cause them to have a random order?

So, instead of

1. Linguistics 1
2. Linguistics 2
3. Linguistics 3

couldn't this cause the titles to be


1. Linguistics 1
2. Linguistics 3
3. Linguistics 2

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -732,12 +583,16 @@ private void addEventInternal(Date startDateTime, Date endDateTime, String captu
...
         dc.set(DublinCore.PROPERTY_TITLE, newTitle);
         DublinCoreValue eventTime = EncodingSchemeUtils.encodePeriod(new DCMIPeriod(startDate, endDate),
                 Precision.Second);
         dc.set(DublinCore.PROPERTY_TEMPORAL, eventTime);
-        mediaPackage = updateDublincCoreCatalog(mediaPackage, dc);
+        try {
+          mediaPackage = updateDublincCoreCatalog(mediaPackage, dc);
+        } catch (Exception e) {
+          throw new RuntimeException(e);
```
lkiesow: Is it a good idea to cast everything to RuntimeExceptions?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -833,63 +691,52 @@ private void updateEventInternal(final String mpId, String modificationOrigin, O
...
-        boolean propertyChanged = captureAgentId.isSome() || startDateTime.isSome() || endDateTime.isSome();
+      // Has a conflict related property be changed?
+      boolean propertyChanged = captureAgentId.isSome() || startDateTime.isSome() || endDateTime.isSome();
 
-        // Check for conflicting events
-        if (!isNewOptOut && (readyForRecording || (propertyChanged && !oldOptOut))) {
-          List<MediaPackage> conflictingEvents = $(findConflictingEvents(captureAgentId.getOr(agentId),
-                  startDateTime.getOr(start), endDateTime.getOr(end))).filter(new Fn<MediaPackage, Boolean>() {
+      // Check for conflicting events
+      if (!isNewOptOut && (readyForRecording || (propertyChanged && !oldOptOut))) {
```
lkiesow: too-complex-boolean-expression-sin ;-)

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -833,63 +691,52 @@ private void updateEventInternal(final String mpId, String modificationOrigin, O
...
-                    }
+                    return !mpId.equals(mp.getIdentifier().compact());
+                  }
                   }).toList();
-          if (conflictingEvents.size() > 0) {
-            logger.info("Unable to update event {}, conflicting events found: {}", mpId, conflictingEvents);
-            throw new SchedulerConflictException("Unable to update event, conflicting events found for event " + mpId);
-          }
+        if (conflictingEvents.size() > 0) {
+          logger.info("Unable to update event {}, conflicting events found: {}", mpId, conflictingEvents);
```
lkiesow, line 731: Does it make sense to log this if we throw it as exception anyway?
I'll bet the exception is caught and being logged somewhere else.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1288,237 +1063,66 @@ public boolean isOptOut(String mediaPackageId) throws NotFoundException, Schedul
...
-      List<Tuple<String, Boolean>> updated = persistence.updateBlacklist(blacklist);
-      for (Tuple<String, Boolean> tuple : updated) {
-        messageSender.sendObjectMessage(SchedulerItem.SCHEDULER_QUEUE, MessageSender.DestinationType.Queue,
-                SchedulerItem.updateBlacklist(tuple.getA(), tuple.getB()));
-      }
-    } catch (SchedulerServiceDatabaseException e) {
-      logger.error("Failed to update blacklist: {}", getStackTrace(e));
+      return persistence.countEvents();
+    } catch (Exception e) {
+      logger.error("Failed to search for events: {}", getStackTrace(e));
```
lkiesow:
```java
logger.error("Failed to search for events", e);
```

But not logging it at all before throwing it again probably makes more sense.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1288,237 +1063,66 @@ public boolean isOptOut(String mediaPackageId) throws NotFoundException, Schedul
...
-      return isBlacklisted(mediaPackageId, start, end, agentId, presenters);
-    } catch (NotFoundException e) {
-      throw e;
+      return persistence.search(captureAgentId, startsFrom, startsTo, endFrom, endTo, Opt.none()).parallelStream()
+          .map(ExtendedEventDto::getMediaPackageId)
+          .map(this::getEventMediaPackage).collect(Collectors.toList());
     } catch (Exception e) {
-      logger.error("Failed to get blacklist status of event with mediapackage '{}': {}", mediaPackageId,
-              getStackTrace(e));
+      logger.error("Failed to search for events: {}", getStackTrace(e));
```
lkiesow:
```suggestion
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1288,237 +1063,66 @@ public boolean isOptOut(String mediaPackageId) throws NotFoundException, Schedul
...
-      logger.error("Failed to get blacklist status of event with mediapackage '{}': {}", mediaPackageId,
-              getStackTrace(e));
+      final Date now = new Date();
+      List<ExtendedEventDto> result = persistence.search(Opt.some(captureAgentId), Opt.none(), Opt.some(now), Opt.some(now), Opt.none(), Opt.some(1));
+      if (result.isEmpty()) {
+        return Opt.none();
+      }
+      return Opt.some(getEventMediaPackage(result.get(0).getMediaPackageId()));
+    } catch (Exception e) {
+      logger.error("Failed to search for events: {}", getStackTrace(e));
```
lkiesow:
```suggestion
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1288,237 +1063,66 @@ public boolean isOptOut(String mediaPackageId) throws NotFoundException, Schedul
...
+  public Opt<MediaPackage> getUpcomingRecording(String captureAgentId) throws SchedulerException {
+    try {
+      final Date now = new Date();
+      List<ExtendedEventDto> result = persistence.search(Opt.some(captureAgentId), Opt.some(now), Opt.none(), Opt.none(), Opt.none(), Opt.some(1));
+      if (result.isEmpty()) {
+        return Opt.none();
       }
+      return Opt.some(getEventMediaPackage(result.get(0).getMediaPackageId()));
+    } catch (Exception e) {
+      logger.error("Failed to search for events: {}", getStackTrace(e));
```
lkiesow:
```suggestion
```

I'll just ignore these from now on

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/ExtendedEventDto.java
```diff
@@ -64,44 +60,207 @@
...
+  @Column(name = "recording_state")
+  private String recordingState;
+
+  /** recording last heard */
+  @Column(name = "recording_last_heard")
+  private Long recordingLastHeard;
+
+  /** review status */
+  @Column(name = "review_status")
+  private String reviewStatus;
```
lkiesow, line 91: In the SQL script, this is defined as
```sql
review_status VARCHAR(128),
```
while the `review_status` above which in here has the same definition, has the following SQL definition:
```sql
review_status VARCHAR(128),
```
The definition in code and in SQL should match whenever possible

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/ExtendedEventDto.java
```diff
@@ -64,44 +60,207 @@
...
+
+  /** capture agent properties */
+  @Column(name = "capture_agent_properties")
+  private String captureAgentProperties;
+
+  /** workflow properties */
+  @Column(name = "workflow_properties")
+  private String workflowProperties;
+
+  private String checksum;
```
lkiesow, line 120: does this not need a column name?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -117,7 +107,7 @@ public void touchLastEntry(String agentId) throws SchedulerServiceDatabaseExcept
     } catch (Exception e) {
       if (tx.isActive())
         tx.rollback();
-      logger.error("Could not updated last modifed date of agent {} status: {}", agentId, getStackTrace(e));
+      logger.error("Could not update last modified date of agent {} status: {}", agentId, getStackTrace(e));
```
lkiesow:
```suggestion
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -169,40 +159,79 @@ public Date getLastModified(String agentId) throws NotFoundException, SchedulerS
...
+      tx = em.getTransaction();
+      tx.begin();
+      Opt<ExtendedEventDto> entityOpt = getExtendedEventDto(mediapackageId, organizationId, em);
+      ExtendedEventDto entity = entityOpt.getOr(new ExtendedEventDto());
+      entity.setMediaPackageId(mediapackageId);
+      entity.setOrganization(organizationId);
+      if (captureAgentId.isSome()) {
+        entity.setCaptureAgentId(captureAgentId.get());
+      }
+      if (start.isSome()) {
```
lkiesow, line 178: Just wondering but is the start and end date really optional?
I expect things to blow up somewhere if there is none ;-)

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -169,40 +159,79 @@ public Date getLastModified(String agentId) throws NotFoundException, SchedulerS
...
+      if (lastModifiedOrigin.isSome()) {
+        entity.setLastModifiedOrigin(lastModifiedOrigin.get());
+      }
+      if (lastModifiedDate.isSome()) {
+        entity.setLastModifiedDate(lastModifiedDate.get());
+      }
+      if (checksum.isSome()) {
+        entity.setChecksum(checksum.get());
+      }
+      Gson gson = new Gson();
```
lkiesow: Maybe use a class level var? We don't need to initialize Gson every time, do we?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -169,40 +159,79 @@ public Date getLastModified(String agentId) throws NotFoundException, SchedulerS
...
+      if (workflowProperties.isSome()) {
+        entity.setWorkflowProperties(gson.toJson(workflowProperties.get()));
```
lkiesow, line 212: Do we now store the workflow properties in here? In the asset manager? In both places? Which version is the correct one if both diverge?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -169,40 +159,79 @@ public Date getLastModified(String agentId) throws NotFoundException, SchedulerS
...
-      throw e;
+      if (entityOpt.isNone()) {
+        em.persist(entity);
+      } else {
+        em.merge(entity);
+      }
+      tx.commit();
     } catch (Exception e) {
-      logger.error("Could not retrieve source for transaction with id '{}': {}", id, getStackTrace(e));
+      logger.error("Could not store extended event: {}", getStackTrace(e));
```
lkiesow:
```suggestion
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -211,96 +240,110 @@ public String getTransactionSource(String id) throws NotFoundException, Schedule
...
       }
+      em.remove(entity.get());
       tx.commit();
+    } catch (NotFoundException e) {
+      throw e;
     } catch (Exception e) {
-      logger.error("Could not store transaction: {}", getStackTrace(e));
       if (tx.isActive())
         tx.rollback();
+      logger.error("Could not delete extended event: {}", getStackTrace(e));
```
lkiesow:
```suggestion
```

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
```diff
@@ -309,81 +352,68 @@ public void storeTransaction(String id, String source) throws SchedulerServiceDa
...
-    // TODO updates last modified state
-    return new ArrayList<Tuple<String, Boolean>>();
+  public Opt<ExtendedEventDto> getEvent(String mediapackageId) throws SchedulerServiceDatabaseException {
+    try {
+      final String orgId = securityService.getOrganization().getId();
+      return getEvent(mediapackageId, orgId);
+    } catch (SchedulerServiceDatabaseException e) {
+      throw e;
+    } catch (Exception e) {
+      logger.error("Could not search for event {}: {}", mediapackageId, e.getStackTrace());
```
lkiesow:
```suggestion
```

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
```diff
@@ -961,6 +756,56 @@ public String getScheduleLastModified(String agentId) throws SchedulerException
...
+      } else if (SC_UNAUTHORIZED == response.getStatusLine().getStatusCode()) {
+        logger.info("Unauthorized to get current recording of agent {}", captureAgentId);
+        throw new UnauthorizedException("Unauthorized to get current recording of agent " + captureAgentId);
+      } else {
+        return Opt.none();
+      }
+    } catch (UnauthorizedException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new SchedulerException("Unable to get current recording from remote scheduler service: " + e);
```
lkiesow:
```suggestion
      throw new SchedulerException("Unable to get current recording from remote scheduler service: ", e);
```

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
```diff
@@ -961,6 +756,56 @@ public String getScheduleLastModified(String agentId) throws SchedulerException
...
+      } else if (SC_UNAUTHORIZED == response.getStatusLine().getStatusCode()) {
+        logger.info("Unauthorized to get upcoming recording of agent {}", captureAgentId);
+        throw new UnauthorizedException("Unauthorized to get upcoming recording of agent " + captureAgentId);
+      } else {
+        return Opt.none();
+      }
+    } catch (UnauthorizedException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new SchedulerException("Unable to get upcoming recording from remote scheduler service: " + e);
```
lkiesow:
```suggestion
      throw new SchedulerException("Unable to get upcoming recording from remote scheduler service: ", e);
```

lkiesow (changes requested):
Here is a number of questions, comments and suggestions.
In general, this looks good though.
I will do some actual testing later today/tomorrow.

lkiesow:
I've tested this in a production-like environment and automated scheduling of ~400 events worked very well. The only problem I really noticed was the weird naming I already found in the code earlier when manually scheduling multiple events. Apart from that, everything seems to work.

docs/scripts/ddl/mysql5.sql
krinnewitz: The problem with storing these properties in the asset manager is, that they are then extremely slow to access. So, for the sake of performance, we decided to store them in the scheduler as long as the event is owned by the scheduler.

docs/upgrade/6_to_7/mysql5.sql
krinnewitz: Since we removed the transactions mechanism, the properties of the namespace `org.opencastproject.scheduler.trx` should be safe to drop.

docs/guides/admin/docs/upgrade.md
krinnewitz: Well, if there are no snapshots with the scheduler as their owner, no data will be migrated. So I guess running the migration is not necessary in this case, but it won't hurt either.

modules/migration/src/main/java/org/opencastproject/migration/EventIdPK.java
krinnewitz: Absolutely. This class was copy-pasted. I will change it as you suggested, but I noticed that `EqualsUtil.hash()` and `EqualsUtil.eq()` are widely used in Opencast. I think it makes sense to refactor this when anybody has some time to kill :-)

modules/migration/src/test/java/org/opencastproject/migration/SchedulerMigrationServiceTest.java
krinnewitz: Sure, I will change it as you suggested.

modules/migration/src/test/java/org/opencastproject/migration/SchedulerMigrationServiceTest.java
krinnewitz: :+1:

modules/migration/src/test/resources/log4j.properties
krinnewitz: I will just remove the whole file.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
krinnewitz: Looks like `CaptureNowProlongingService` uses it.

lkiesow:
Btb, some review findings are just comments and may not necessarily require any change.
Just take a look at them and figure out if it makes sense to change anything ;-P

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
krinnewitz: We wanted to preserve the previous behavior of this method which was aborting as soon as an exception occurs. So I think in this case, it's fine to "uncheck" the exceptions since we are in a stream (and you cannot throw checked exceptions in streams).

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
krinnewitz: ok, I just re-activated my boolean algebra skills to get rid of those brackets ;)

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
krinnewitz: I would like to leave that in since the log entry provides some details which the exception doesn't contain.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
lkiesow: Exactly. Therefore, if the answer is that it does not make sense to cast everything to `RuntimeException`, it does not make sense to use streams here.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
krinnewitz: Since Asset Manager properties are extremely slow to query, we decided to store the workflow and capture agent properties in the scheduler. As long as the event is scheduled, the properties stored in the scheduler are the ones being used. When the event is archived, the properties stored in the asset manager are used.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/SchedulerServiceDatabaseImpl.java
krinnewitz: Well, to be honest, start and end are not optional, but the method is used to persist new events as well as to update them. So not providing start and end date only makes sense when updating some other fields of an existing event. This is not a good design, I admit. But since this is an internal class which is used by the scheduler implementation only, I think it doesn't make sense to be extra cautious and add a whole bunch of checks here.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
krinnewitz: That's true. I will remove the `AtomicInteger` and use the event's index in the list of periods instead. This way, the events are labeled as they would be without the `parallelStream`. Of course, they are not inserted in the database in that order, but I think that shouldn't be a problem.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/persistence/ExtendedEventDto.java
krinnewitz: Most likely yes. I will add it and see if everything still works. I don't see a reason to not have a `@Column` annotation.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
krinnewitz: In general I think it doesn't make sense to cast checked Exceptions to RuntimeExceptions since you are losing compile time safety.

However, in this particular case, I think it is justifiable. These RuntimeExceptions are caught outside the forEach and converted to checked SchedulerExceptions again. So from a caller's point of view, the behavior of the method doesn't really change.

Of course I could also fire up a `ThreadPool` and do a lot of foo instead of using a parallel stream, but I think that is more likely to introduce bugs since you have to shutdown the threadpool properly and so on.

Since Opencast has quite an old code base, many checked exceptions are used. Today, you would probably use less checked exceptions because java8's functional interfaces don't go well with checked exceptions.

krinnewitz: I will use `Misc.chuck` to be compliant with other Opencast components

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
```diff
@@ -859,42 +673,23 @@ public ReviewStatus getReviewStatus(String mediapackageId)
...
       }
-    } catch (NotFoundException e) {
-      throw e;
+      final String countString = EntityUtils.toString(response.getEntity(), UTF_8);
+      return Integer.parseInt(countString);
     } catch (UnauthorizedException e) {
       throw e;
     } catch (Exception e) {
-      throw new SchedulerException("Unable to get event blacklist status from remote scheduler service: " + e);
+      throw new SchedulerException("Unable to get event count from remote scheduler service: " + e);
```
karendolan: It would help debugging a lot here to add a log of the class name of the caught Exception and its message. My preference is a WARN log with the stack dump, since looks like an uncommon catch.

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
karendolan: Ditto with @lkiesow, to attach the original exception in the throw. But, because the reciever is not guaranteed to mention the originating exception, adding an INFO log here with the name of the exception and message and related captureAgentId, would help a lot. 

Is there a reason for not wanting to throw a regular Java class (non-Opencast) NotFoundException if the captureAgentId is not known?

karendolan:
Hi, since this pull has a lot of them, can someone explain why it is good for Opencast to make new iterations of NotFoundException, etc, that do not extend plain old Java NotFoundExceptions, etc? I do not understand the gained benefits over the complexity added.

krinnewitz:
> Hi, since this pull has a lot of them, can someone explain why it is good for Opencast to make new iterations of NotFoundException, etc, that do not extend plain old Java NotFoundExceptions, etc? I do not understand the gained benefits over the complexity added.

Hi @karendolan I think in most cases it makes sense to not throw exceptions which are part of the java API since you want to handle exceptions thrown by the java class library differently from the ones thrown by your own code. And in most cases you want to express a different situation than the "standard" java exception stands for.

For example here is a snippet from the javadoc of `javax.ws.rs.NotFoundException`:

> A runtime exception indicating a resource requested by a client was not found on the server.

That may not make sense in all places of your own code.

docs/scripts/ddl/mysql5.sql
krinnewitz: I will remove `last_modified_origin`. The other fields seem to be in use.

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
krinnewitz: I will change all the places in SchedulerRemoteServiceImpl where the `SchedulerException("some message: " + e)` pattern is used to `SchedulerException("some message",  e)`. However, I would like to not go any further at this point since it is not the scope of this PR to completely rework Opencasts Exception Handling.

krinnewitz:
Thanks @lkiesow and @karendolan.

I have fixed most of your comments.

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
lkiesow: A log is not a good idea. Why not

```
throw new SchedulerException("Unable to get event count from remote scheduler service", e);
```

That way you get a proper stack trace including the current class name without having several separated log entries. You can actually see where the error comes from, what caused it and which way it went through the code.

modules/scheduler-remote/src/main/java/org/opencastproject/scheduler/remote/SchedulerServiceRemoteImpl.java
lkiesow: Makes sense. Thanks.
I've also deliberately only complained about the newly introduced exceptions ;-P

smarquard:
Is there anything still outstanding on this review?

lkiesow:
Essentially just my final review. Unfortunately, I have been very busy last week.

One thing I like to take another look at is the separation of workflow parameters for which we now have two separate stores. I'd like to make sure that we do not get any inconsistencies there.

docs/upgrade/6_to_7/mysql5.sql
```diff
@@ -1 +1,28 @@
...
+  start_date DATETIME NOT NULL,
+  end_date DATETIME NOT NULL,
+  source VARCHAR(255),
+  recording_state VARCHAR(255),
+  recording_last_heard BIGINT,
+  review_status VARCHAR(128),
+  review_date DATETIME,
+  presenters TEXT(65535),
+  optout TINYINT(1),
+  last_modified_origin VARCHAR(255),
```
lkiesow: I noticed that using the database script and upgrading will give you two different database schemas right now:

I've created two databases, one updated from 6 to 7 and one straight from the 7 scripts:

```sh
%mysql>
create database oc6;
use oc6;
source …/opencast6/docs/scripts/ddl/mysql5.sql;
source …/opencast7/docs/upgrade/6_to_7/mysql5.sql;
create database oc7;
use oc7;
source …/opencast7/docs/scripts/ddl/mysql5.sql;
```

Then dumped the databases:
```sh
mysqldump oc6 > oc6.sql
mysqldump oc7 > oc7.sql
```
and diffed them which let me know that this column is only in the updated database:
```diff
3c3
< -- Host: localhost    Database: oc6
---
> -- Host: localhost    Database: oc7
1172d1171
<   `last_modified_origin` varchar(255) DEFAULT NULL,
1625c1624
< -- Dump completed on 2019-02-11 23:20:57
---
> -- Dump completed on 2019-02-11 23:21:03
```

smarquard:
I'm getting test failures with this branch on a local build. The expected and actual times are out by one hour, which makes me think it's timezone related. Building in SAST, which is GMT+2 without DST adjustment.

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.opencastproject.scheduler.impl.UtilTests
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.218 s - in org.opencastproject.scheduler.impl.UtilTests
[INFO] Running org.opencastproject.scheduler.impl.RecordingTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s - in org.opencastproject.scheduler.impl.RecordingTest
[INFO] Running org.opencastproject.scheduler.impl.SchedulerServiceImplTest
16:43:51  WARN (SchedulerServiceImpl:1379) - Invalid recording state: bad_state.
16:44:11  WARN (SchedulerServiceImpl:443) - Mediapackage with id '1' already exists!
[ERROR] Tests run: 34, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 25.53 s <<< FAILURE! - in org.opencastproject.scheduler.impl.SchedulerServiceImplTest
[ERROR] testEventManagement(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 0.24 s  <<< ERROR!
org.opencastproject.scheduler.api.SchedulerException: Event ID 5822c12b-5941-43d6-b8b9-1c22ba293d13 has already ended at 2019-02-12T13:45:11Z and now is 2019-02-12T14:44:11Z
	at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testEventManagement(SchedulerServiceImplTest.java:1442)

[INFO] Running org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest
[ERROR] Tests run: 11, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 0.885 s <<< FAILURE! - in org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest
[ERROR] testLastModifed(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 0.077 s  <<< FAILURE!
java.lang.AssertionError
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testLastModifed(SchedulerServiceDatabaseImplTest.java:85)

[ERROR] testUpdateEvent(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 0.078 s  <<< FAILURE!
java.lang.AssertionError: expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testUpdateEvent(SchedulerServiceDatabaseImplTest.java:225)

[ERROR] testCreateEvent(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 0.065 s  <<< FAILURE!
java.lang.AssertionError: expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testCreateEvent(SchedulerServiceDatabaseImplTest.java:139)

[ERROR] testResetRecordingState(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 0.064 s  <<< FAILURE!
java.lang.AssertionError: expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testResetRecordingState(SchedulerServiceDatabaseImplTest.java:803)

[INFO] Running org.opencastproject.scheduler.impl.SchedulerUtilTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.068 s - in org.opencastproject.scheduler.impl.SchedulerUtilTest
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   SchedulerServiceDatabaseImplTest.testCreateEvent:139 expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
[ERROR]   SchedulerServiceDatabaseImplTest.testLastModifed:85
[ERROR]   SchedulerServiceDatabaseImplTest.testResetRecordingState:803 expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
[ERROR]   SchedulerServiceDatabaseImplTest.testUpdateEvent:225 expected:<Mon Jan 07 09:00:00 SAST 2019> but was:<Mon Jan 07 08:00:00 SAST 2019>
[ERROR] Errors: 
[ERROR]   SchedulerServiceImplTest.testEventManagement:1442 » Scheduler Event ID 5822c12...
[INFO] 
[ERROR] Tests run: 51, Failures: 4, Errors: 1, Skipped: 0

```

lkiesow (changes requested):
I did a final in-depth review and this patch looks good except for one line of SQL update code which needs to be removed. Scheduling stuff works, capture agents work with the scheduler, migration works (I did upgrade from 6 to 7). @krinnewitz, can once the SQL line is removed, I'm going to merge this.

lkiesow:
Ups, missed @smarquard's comment. Is this issue new? Did the tests work in the same timezone without this patch?

smarquard:
Hmm, I get similar failures in the current develop branch. Maybe related to MH-12177 / #604 ?
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.opencastproject.scheduler.impl.UtilTests
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2.748 s - in org.opencastproject.scheduler.impl.UtilTests
[INFO] Running org.opencastproject.scheduler.impl.RecordingTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0 s - in org.opencastproject.scheduler.impl.RecordingTest
[INFO] Running org.opencastproject.scheduler.impl.SchedulerServiceImplTest
21:30:53  WARN (SchedulerServiceImpl:1767) - Invalid recording state: bad_state.
21:31:15 ERROR (SchedulerServiceImpl$SchedulerTransactionImpl:2326) - Unable to add events, conflicting events found: [d2b94471-95a6-451d-ad9a-8ce6b690a3e6]
21:31:15 ERROR (SchedulerServiceImpl$SchedulerTransactionImpl:2326) - Unable to add events, conflicting events found: [d2b94471-95a6-451d-ad9a-8ce6b690a3e6]
21:31:19  WARN (SchedulerServiceImpl:569) - Mediapackage with id '1' already exists!
[ERROR] Tests run: 36, Failures: 1, Errors: 3, Skipped: 0, Time elapsed: 48.981 s <<< FAILURE! - in org.opencastproject.scheduler.impl.SchedulerServiceImplTest
[ERROR] testFindConflictingEvents(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 0.452 s  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testFindConflictingEvents(SchedulerServiceImplTest.java:1908)

[ERROR] testTransactionConflicts(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 0.45 s  <<< ERROR!
org.opencastproject.scheduler.api.SchedulerException: Event ID 9a23667b-a7cc-4dd0-95d4-13653d8e3909 has already ended at 2019-02-12T18:32:16Z and now is 2019-02-12T19:31:16Z
	at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testTransactionConflicts(SchedulerServiceImplTest.java:1288)

[ERROR] testTransactionRollback(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 0.283 s  <<< ERROR!
org.opencastproject.scheduler.api.SchedulerException: Event ID 34079374-aabe-4704-bf4e-4ca19012cfbb has already ended at 2019-02-12T18:32:16Z and now is 2019-02-12T19:31:16Z
	at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testTransactionRollback(SchedulerServiceImplTest.java:1151)

[ERROR] testEventManagement(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 0.602 s  <<< ERROR!
org.opencastproject.scheduler.api.SchedulerException: Event ID 656643a8-6df1-43bd-a72d-4e596a496bab has already ended at 2019-02-12T18:32:18Z and now is 2019-02-12T19:31:19Z
	at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testEventManagement(SchedulerServiceImplTest.java:2106)

[INFO] Running org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest
[ERROR] Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.194 s <<< FAILURE! - in org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest
[ERROR] testLastModifed(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 0.092 s  <<< FAILURE!
java.lang.AssertionError
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testLastModifed(SchedulerServiceDatabaseImplTest.java:81)

[ERROR] testTransactions(org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest)  Time elapsed: 1.1 s  <<< FAILURE!
java.lang.AssertionError
	at org.opencastproject.scheduler.impl.persistence.SchedulerServiceDatabaseImplTest.testTransactions(SchedulerServiceDatabaseImplTest.java:150)

[INFO] Running org.opencastproject.scheduler.impl.SchedulerUtilTest
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.073 s - in org.opencastproject.scheduler.impl.SchedulerUtilTest
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   SchedulerServiceImplTest.testFindConflictingEvents:1908 expected:<1> but was:<0>
[ERROR]   SchedulerServiceDatabaseImplTest.testLastModifed:81
[ERROR]   SchedulerServiceDatabaseImplTest.testTransactions:150
[ERROR] Errors: 
[ERROR]   SchedulerServiceImplTest.testEventManagement:2106 » Scheduler Event ID 656643a...
[ERROR]   SchedulerServiceImplTest.testTransactionConflicts:1288 » Scheduler Event ID 9a...
[ERROR]   SchedulerServiceImplTest.testTransactionRollback:1151 » Scheduler Event ID 340...
[INFO] 
[ERROR] Tests run: 44, Failures: 3, Errors: 3, Skipped: 0

```

lkiesow:
That was my guess. While I've never seen it, @gregorydlogan has been hit by these multiple times. They seem to be timezone related. I haven't had a close look at the issue though.

docs/upgrade/6_to_7/mysql5.sql
krinnewitz: Thank you, I just pushed a commit which removes this column

lkiesow:
Btb, to properly test this on `develop`, I applied the following patches beforehand to remove some critical issues which would have otherwise made me testing this impossible:

- pull request #702 
- pull request #660
- pull request #659

krinnewitz:
I can reproduce test failures when building scheduler-impl with `mvn clean install -Duser.timezone=Africa/Johannesburg`

krinnewitz:
@lkiesow @smarquard I added yet another commit which should fix the test failures.

smarquard:
That solution to fixing the timezone issues is concerning, because at the beginning of our "timezone journey", we found that the default timezone could be set incorrectly when multiple threads were changing the timezone default at the same time. Thread 1 changes the default, thread 2 gets the default (which is now not the actual user/system default, but what thread 1 has changed it to), thread 1 changes it back to the user/system default but thread 2 has now hung on to an incorrect default and sets the default incorrectly. So we made this change:

https://github.com/opencast/opencast/commit/76dfc358d04926643e9e922d512f5488f6e23000

which your commit essentially reverses. IMO it's always unsafe to call setDefault(...) because it's changing a value which is JVM-wide, and you don't know what other threads are doing at that time.

krinnewitz:
Thanks for the hint. I removed the last commit. Since the problem with failing tests already exists in develop and is not introduced by this PR, I suggest to merge it.

smarquard:
I agree with merging this, despite the test failures which I'm hoping will be taken care of in #604 that @karendolan is working on. That PR (as it stands) removes all the timezone setDefault calls.

krinnewitz:
Just rebased onto develop to resolve merge conflict

smarquard:
The failing tests pass on  my local build if I use:

mvn -Duser.timezone=Europe/Zurich

lkiesow:
I'll then merge it like this.

Memo to myself: Mid-term, add `-Duser.timezone=America/St_Johns` to the Travis builds.

lkiesow:
It's merged. Thanks for the patch.
I think this does improve Opencast quite a bit.

lkiesow:
Seems like this has some issues after all. I'm now getting test failures on a regular basis (not every time!) on `develop`. This I have seen on multiple systems:

```
[INFO] Running org.opencastproject.scheduler.impl.SchedulerServiceImplTest
11:43:41  WARN (SchedulerServiceImpl:1377) - Invalid recording state: bad_state.
11:43:53  WARN (SchedulerServiceImpl:444) - Mediapackage with id '1' already exists!
[ERROR] Tests run: 34, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 13.256 s <<< FAILURE! - in org.opencastproject.scheduler.impl.SchedulerServiceImplTest
[ERROR] testAddMultipleEvents(org.opencastproject.scheduler.impl.SchedulerServiceImplTest)  Time elapsed: 2.657 s  <<< FAILURE!
java.lang.AssertionError: expected:<200> but was:<199>
        at org.opencastproject.scheduler.impl.SchedulerServiceImplTest.testAddMultipleEvents(SchedulerServiceImplTest.java:996)
```

It also happened on the Travis build here:
https://api.travis-ci.org/v3/job/495192840/log.txt

smarquard:
Curious that the test failure shows up unpredictably. I haven't seen any failures like this, either on develop or our 6.x custom branch with this code merged in.

These WARNs show up for successful builds, so are not an issue:

```
16:11:58  WARN (SchedulerServiceImpl:1377) - Invalid recording state: bad_state.
16:12:19  WARN (SchedulerServiceImpl:444) - Mediapackage with id '1' already exists!

```

pmiddend:
@lkiesow As I am now noticing, we indeed write into a `Map` in a thread-unsafe way (see `scheduledEvents` in `addMultipleEventInternal`). This might lead to sporadic problems like you mentioned (and is definitely a bug). I'll file another PR for that.
See #729.

lkiesow:
Thanks for taking a look at this.
I will review the patch tomorrow.

PR #615, opened 2018-12-07
MH-13238, don't throw related services straight into ERROR state just because job succeeded on current service

JamesUoM:
Moved from 4.x due to EoL https://github.com/opencast/opencast/pull/569


KatrinIhler (commented):
You actually don't need to create a new PR when you're moving to another targeted version, you can just rebase your code and change the base of the PR. But anyway, let's finally merge this!

PR #616, opened 2018-12-07
MH-13230: remove the need for passing an Accept header with external api requests

slampunk:
According to the [documentation<sup>1,2</sup>](https://docs.opencast.org/r/6.x/developer/api/usage/), the `Accept` header is optional when making requests against the external api:

> If that header (Accept) is not specified, or no version information can be extracted from the header, the assumption is that the request should be executed against the most recent version.

This patch allows for the execution of the latest version of the api on those requests which do not have an accompanying `Accept` header.

[1] https://docs.opencast.org/r/6.x/developer/api/usage/
[2] https://github.com/opencast/opencast/blame/r/6.x/docs/guides/developer/docs/api/usage.md#L26-L28


staubesv:
@slampunk Haven't yet seen a client that does not set the accept header... good catch ;-)

slampunk:
we've quite a number of scripts which pass ONLY what we tell them to... so imagine having to add an `Accept` header to each request! Luckily, that line in the documentation makes my life easier ;)

staubesv (approved):

PR #618, opened 2018-12-07
MH-13275, Allows the workflow to select the audio track for composite video

phoebem:
The composite audio is by default a mix of both lower and upper video 
 - "[0:a][1:a]amix=inputs=2[aout]", 
When both tracks have the same audio source, the resulting audio is echoey.
This patch allows the workflow to select one track (upper or lower) as audio source.
If the parameter is omitted, both tracks will be used.


docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
```diff
@@ -3,7 +3,7 @@
 ## Description
 
 The CompositeWorkflowOperationHandler is used to composite two videos (upper and lower) and an optional watermark into
-one video, including encoding to different formats. The audio track is always taken from the lower video. Everything is
+one video, including encoding to different formats. The audio track is always taken from the both videos by default. Everything is
```
staubesv: Suggestion: omit "always"

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
```diff
@@ -12,6 +12,10 @@ The internal ffmpeg command is using the following filters: scale for scaling th
 dimension including the background color, movie for adding additional videos and images and overlay for aligning the
 videos and images to the output dimension. More info can be found here: https://trac.ffmpeg.org/wiki/FilteringGuide
 
+If both upper and lower tracks have audio, "source-audio-name" can be set to "upper" or "lower" to choose only the
+audio from that track for the composite video.
+If this parameter is omitted, audio from both tracks are mixed into the composite media.
```
staubesv, line 17: Suggestion: It seems a bit weird that omitting a configuration key is the only way to select a specific option. Maybe allow "upper", "lower" and "both" as values for the configuration key.

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
```diff
@@ -12,6 +12,10 @@ The internal ffmpeg command is using the following filters: scale for scaling th
 dimension including the background color, movie for adding additional videos and images and overlay for aligning the
 videos and images to the output dimension. More info can be found here: https://trac.ffmpeg.org/wiki/FilteringGuide
 
+If both upper and lower tracks have audio, "source-audio-name" can be set to "upper" or "lower" to choose only the
```
staubesv: Suggestion: Omit "-name" in "source-audio-name" as "name" seems somewhat arbitrary/undefined/uncommon ("upper" and "lower" are not names, are they?). Maybe just "audio-source"?

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
```diff
@@ -22,6 +26,7 @@ Tags and flavors can be used in combination.
 
 |configuration keys|value type (EBNF)|example|description|default value|
 |------------------|-----------------|-------|-----------|-------------|
+|source-audio-name|String , { "," , String }    |upper|The "name" of track to use as a source audio.|EMPTY|
```
staubesv: Value type should be "upper" | "lower"

Also: We don't have "track" names, do we? Maybe something like "Which track(s) to use as audio source"

Note: If we had the option "both", you could specify the default value correctly here

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -125,8 +125,9 @@
   private static final int PROFILE_ID_INDEX = 0;
   private static final int UPPER_TRACK_INDEX = 3;
   private static final int UPPER_TRACK_LAYOUT_INDEX = 4;
-  private static final int WATERMARK_INDEX = 7;
-  private static final int WATERMARK_LAYOUT_INDEX = 8;
+  private static final int WATERMARK_INDEX = 8;
+  private static final int WATERMARK_LAYOUT_INDEX = 9;
+  private static final int AUDIO_SOURCE_INDEX = 7;
```
staubesv, line 130: Why not just add "AUDIO_SOURCE_INDEX" with 9 instead of changing existing indexes?

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -1668,8 +1671,12 @@ private static String buildCompositeCommand(Dimension compositeTrackSize, LaidOu
     if (upperLaidOutElement.isSome()) {
       // handle audio
       // if both videos contain audio mix it into a single audio stream
-      final boolean lowerAudio = lowerLaidOutElement.getElement().hasAudio();
-      final boolean upperAudio = upperLaidOutElement.get().getElement().hasAudio();
+      boolean lowerAudio = lowerLaidOutElement.getElement().hasAudio();
+      boolean upperAudio = upperLaidOutElement.get().getElement().hasAudio();
+      if (audioSourceName != null) {
+           lowerAudio  = lowerAudio  && "lower".equalsIgnoreCase(audioSourceName);
```
staubesv: Suggestions: Use constants for strings

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -1668,8 +1671,12 @@ private static String buildCompositeCommand(Dimension compositeTrackSize, LaidOu
     if (upperLaidOutElement.isSome()) {
       // handle audio
       // if both videos contain audio mix it into a single audio stream
```
staubesv: The comment is not correct anymore as you don't unconditionally mix into a single audio stream anymore

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
```diff
@@ -543,6 +543,7 @@ public Response imageSync(@FormParam("sourceTrack") String sourceTrackXml, @Form
           @RestParameter(description = "The watermark source attachment containing watermark image", isRequired = false, name = "watermarkTrack", type = Type.TEXT),
           @RestParameter(description = "The watermark layout containing the JSON definition of the layout", isRequired = false, name = "watermarkLayout", type = Type.TEXT),
           @RestParameter(description = "The background color", isRequired = false, name = "background", type = Type.TEXT, defaultValue = "black"),
+          @RestParameter(description = "The name of the audio source (lower or upper)", isRequired = false, name = "audioSourceName", type = Type.TEXT),
```
staubesv: Suggestion: The audio source (lower or upper). If we had "both" also, you could specify the default value here.

modules/composer-ffmpeg/src/test/java/org/opencastproject/composer/impl/ComposerServiceTest.java
```diff
@@ -353,14 +353,49 @@ public void testComposite() throws Exception {
             .get(1));
 
     Job composite = composerService.composite(outputDimension, Option.option(lowerLaidOutElement), upperLaidOutElement,
-            watermarkOption, "composite.work", "black");
+            watermarkOption, "composite.work", "black", null);
```
staubesv: If we just had that "both"... ;-)

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
```diff
@@ -389,6 +393,11 @@ private WorkflowOperationResult composite(MediaPackage src, WorkflowOperationIns
         watermarkLayout = singleLayouts.getB();
       }
 
+      // #DCE OPC-226 preview video echo
```
staubesv: Please don't refer to tickets only your site has access to

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
```diff
@@ -389,6 +393,11 @@ private WorkflowOperationResult composite(MediaPackage src, WorkflowOperationIns
...
+      // Check that source audio is upper, lower or use a combination of both
+      if (sourceAudioName != null && !("upper".equalsIgnoreCase(sourceAudioName) || "lower".equalsIgnoreCase(sourceAudioName)))
```
staubesv: Suggestions: Maybe prefer the use of string constants

staubesv (changes requested):
I've found some minor issues. Personally, I'm not a big fan of the source-audio-name as "name" is somewhat undefined/uncommon, but no strong opinion about that.

staubesv:
@phoebem Just wondering: What audio tracks do your recordings have that makes you know that you have two audio tracks and always need to use "upper", for example?

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
phoebem: Good catch in the value type!.
I thought a lot about the name. If I use a flavor or tag, then I have to worry about extra tracks in the mediapackage that matches the tag/flavor but is not "lower" or "upper".  I can't use tag or key or flavor, so I use name, because "upper" and "lower" are used throughout and I see them as named inputs. If I just use source-audio, then it is even more ambiguous.  I will add "both" though.

phoebem:
We have the same audio tracks on both presenter and presentation so that students can watch one or the other or both. It doesn't matter if we use upper or lower as long as we use only one.

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
phoebem: Actually I guess omission is a "don't care", since compose will try and use both if they exist or any one if only one has audio.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
phoebem: In ComposerServiceImpl:631,635, WATERMARK_* are optional arguments, and AUDIO_SOURCE_INDEX is not optional, so it has to come before the optional args.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
phoebem: Arrgh! I used a patch to move the edits from our in house repo (with a ton of customization) to this repo. I just missed those.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
staubesv: Thanks for clarifying and sorry for the noise.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
staubesv: Don't worry - That is what code reviews are good for ;-)

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
staubesv: Agree on omission being a "don't care", but still weird to only be able to get the default by omitting the configuration key. Finally, "don't care" is a valid option. Why should I be unable to explicitly set this? But no strong opinion ;-)

Also no strong opinion on the name of the configuration key.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
phoebem: It is not required, so the default isn't really needed.

staubesv: If it was required, there was no default value ;-)

Semantically, the default value of your solution is "mix audio of both lower and upper". That is what is does if you don't specify a value (implies a default behaviour is used), right?

docs/guides/admin/docs/workflowoperationhandlers/composite-woh.md
```diff
@@ -22,6 +26,7 @@ Tags and flavors can be used in combination.
 
 |configuration keys|value type (EBNF)|example|description|default value|
 |------------------|-----------------|-------|-----------|-------------|
+|source-audio-name|"lower", "upper" or "both" |upper|The "name" of track to use as a source audio.|EMPTY|
```
staubesv: Default value should now be "both" as omitting the configuration key source-audio-name will default to "both".

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
```diff
@@ -543,6 +543,7 @@ public Response imageSync(@FormParam("sourceTrack") String sourceTrackXml, @Form
           @RestParameter(description = "The watermark source attachment containing watermark image", isRequired = false, name = "watermarkTrack", type = Type.TEXT),
           @RestParameter(description = "The watermark layout containing the JSON definition of the layout", isRequired = false, name = "watermarkLayout", type = Type.TEXT),
           @RestParameter(description = "The background color", isRequired = false, name = "background", type = Type.TEXT, defaultValue = "black"),
+          @RestParameter(description = "The name of the audio source (lower or upper or both)", isRequired = false, name = "audioSourceName", type = Type.TEXT),
```
staubesv: Please add defaultValue = "both" and @DefaultValue("both")

modules/composer-ffmpeg/src/test/java/org/opencastproject/composer/impl/ComposerServiceTest.java
```diff
@@ -353,14 +353,49 @@ public void testComposite() throws Exception {
             .get(1));
 
     Job composite = composerService.composite(outputDimension, Option.option(lowerLaidOutElement), upperLaidOutElement,
-            watermarkOption, "composite.work", "black");
+            watermarkOption, "composite.work", "black", null); // null or "both" means that both tracks are checked for audio and both audio tracks are mixed into the final composite if they exist
```
staubesv: Suggestion: Just replace _null_ by _"both"_ (or the respective string constant if you have access to it) and remove the comment. I would recommend to not allow _null_ as actual parameter in composerService.composite() as you can expect clients to correctly chose between "lower", "upper" and "both".

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/ComposerService.java
```diff
@@ -43,6 +43,11 @@
   String AUDIO_ONLY = "a";
   String VIDEO_ONLY = "v";
 
+  /** sourceAudioName options for composite - use one or both, if null is passed, both will be used */
```
staubesv, line 46: Hmm... why do you allow _null_? You can expect clients to correctly choose from "lower", "upper" or "both" ("correct usage of interface").

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
```diff
@@ -345,6 +353,7 @@ private WorkflowOperationResult composite(MediaPackage src, WorkflowOperationIns
     CompositeSettings(MediaPackage mediaPackage, WorkflowOperationInstance operation)
             throws WorkflowOperationException {
       // Check which tags have been configured
+      sourceAudioName = StringUtils.trimToNull(operation.getConfiguration(SOURCE_AUDIO_NAME));
```
staubesv, line 356: Suggestion: I would recommend to default to "both" (not to null). This slightly simplifies further handling as you don't need to check for null anymore.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CompositeWorkflowOperationHandler.java
```diff
@@ -389,6 +398,10 @@ private WorkflowOperationResult composite(MediaPackage src, WorkflowOperationIns
         watermarkLayout = singleLayouts.getB();
       }
 
+      // Check that source audio is upper, lower or use a combination of both
+      if (sourceAudioName != null && !sourceAudioOption.matcher(sourceAudioName).matches())
+        throw new WorkflowOperationException("sourceAudioName if used, must be either upper or lower!");
```
staubesv: upper, lower or both

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/ComposerService.java
```diff
@@ -77,28 +82,30 @@ Job mux(Track sourceVideoTrack, Track sourceAudioTrack, String profileId) throws
...
    * @throws MediaPackageException
-   *           if the mediapackage is invalid
+   *     if the mediapackage is invalid
    */
-  Job composite(Dimension compositeTrackSize, Option<LaidOutElement<Track>> upperTrack, LaidOutElement<Track> lowerTrack,
-          Option<LaidOutElement<Attachment>> watermark, String profileId, String background) throws EncoderException,
-          MediaPackageException;
+  Job composite(Dimension outputDimension, Option<LaidOutElement<Track>> option,
+        LaidOutElement<Track> lowerLaidOutElement, Option<LaidOutElement<Attachment>> watermarkOption,
+        String identifier, String outputBackground, String sourceAudioName) throws EncoderException, MediaPackageException;
```
staubesv: Can you please either revert _outputBackground_ to _background_ or change the variable name consistently in the interface and its implementation?
The same interface now uses different names for the same formal parameter, e.g. the interface here names it _outputBackground_ but the implementation of the interface in ComposerServiceRemoteImpl names it _background_.  While this is correct, it seems suboptimal the use different names for the same formal parameter.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -1667,9 +1670,14 @@ private static String buildCompositeCommand(Dimension compositeTrackSize, LaidOu
 
     if (upperLaidOutElement.isSome()) {
       // handle audio
-      // if both videos contain audio mix it into a single audio stream
-      final boolean lowerAudio = lowerLaidOutElement.getElement().hasAudio();
-      final boolean upperAudio = upperLaidOutElement.get().getElement().hasAudio();
+      boolean lowerAudio = lowerLaidOutElement.getElement().hasAudio();
+      boolean upperAudio = upperLaidOutElement.get().getElement().hasAudio();
+      if (audioSourceName != null) {
```
staubesv: This check gets superfluous if you don't allow _null_ ....

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -1667,9 +1670,14 @@ private static String buildCompositeCommand(Dimension compositeTrackSize, LaidOu
...
+           if (! ComposerService.BOTH.equalsIgnoreCase(audioSourceName)) {
+             lowerAudio  = lowerAudio  && ComposerService.LOWER.equalsIgnoreCase(audioSourceName);
```
staubesv: Those two lines contain a number of superfluous space characters...

staubesv (changes requested):
Thanks for following the suggestion and also accept "both". I've added a number of minor comments considering "both" being implemented as default value.

Sorry, btw, for the large number of comments on this rather cosmetic issue.

The reasoning is that this WOH does support three different ways of how it can deal with audio. It seems cleaner to me to explicitly support all three ways (by having lower, upper and both) and document which one is used in case of the absence of the configuration key.
In the Java code, you can very much expect a (Java) client to explicitly chose lower, upper or both (no need to deal with _null_ here as _null_ is no valid "way" - also, _null_ as actual parameter isn't necessarily a best practise). So you just need to make sure that the implementation of the WOH defaults to "both" in case the configuration key is missing and check (you already do so) that the WOH calls the Java interface with either "lower", "upper" or "both" (but not _null_).

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
phoebem: Yes. That was the original default, I want it to stay the same.

modules/composer-service-api/src/main/java/org/opencastproject/composer/api/ComposerService.java
phoebem: That was an artifact of the patch and I didn't catch it when it was first changed. I will fix it.

staubesv:
@phoebem The only question left is why you would want to allow _null_ (which is now equivalent to "both") as actual parameter instead of just asking for "upper", "lower" or "both". 
The REST APIs will set source-audio-name to "both" if the parameter is omitted (because of defaultValue), so the only case left seems Java code calling the method composerService.composite() which you could expect to pass "both" instead of _null_.

Or have I misunderstood something?

phoebem:
Functionally it doesn’t really matter. 
Checking for null is cheaper than regex search against a pattern and if anyone would want to call the service directly can pass a null and it will throw an exception.
If you feel very strongly about it, I will take it out.

Phoebe

staubesv:
@phoebem I've just tested this and found a problem considering the default "both": If the configuration key "source-audio-name" is omitted, the result of composite will not contain any audio tracks. I suggest to ensure that the workflow operation defaults to "both" in case the configuration key is omitted. 

No strong opinion about the null check. Note, though, that performance really does not matter in this case (regex vs. null check in WOH parameter check).

KatrinIhler:
@phoebem Can you take another look at this?

phoebem:
I have submitted a new PR for this code. The rebase would have been messy.
As far as I know I have addressed all the comments above. This branch can be closed.

PR #619, opened 2018-12-07
MH-13279, Admin-ui Agents REST Details

phoebem:
Added optional parameter to /admin-ng/capture-agents/agents.json
to return details of all capture agents.


staubesv:
@phoebem As commented in the JIRA issue: Please do not add functionality to the Admin UI facade not used by the Admin UI. Please add this to the External API instead.

phoebem:
@staubesv Same as with the search-by-enddate change. This is in preparation for a live monitor UI (plugin in for admin-ui) that will come later. The monitor will allow the admin user to see the events and the live stream from the perspective of the capture agents.

staubesv:
@phoebem If this is preparation for some other piece of work, I would suggest to set this PR _onhold_ for now. Is this ok for you?

gregorydlogan:
I would agree with Sven in the sense that this should be wrapped up into the UI PR.  However, I'm very much against the idea of gating off anything that's not actively in use the existing UI functionality - if it makes life easier then I'm fine with this PR, especially since it changes nothing by default.  After all, there's a lot of stuff in the existing facade that's not used, or functional.

phoebem:
It will be a while before the monitor function is ready as we want to move away from rtmp before the PR.
If this can be merged now, then I won't forget about it and the next PR will be simpler for the reviewer.

mliradelc:
Hi, just a reminder, next Monday (April 1st) is the feature freeze of Opencast 7.0 😉

staubesv:
@phoebem As the live monitor seems not to make it into Opencast 7.0 (feature freeze on Monday), I've labelled this PR as "onhold" so that we can review all related PRs together.

KatrinIhler:
@phoebem This PR has conflicts now!

KatrinIhler:
@phoebem What do you think about closing this for now and refiling once the live monitor is ready?

PR #621, opened 2018-12-10
MH-13281, Added property keep-last-snapshot for asset-delete WOH

CGreweling:
With this PR it is possible to delete all snapshots of an episode and keep the last one.
The main idea is to save some Diskspace. 

It should also be possible to replace attachments/tracks. 
By un-tag the "archive-tag" of an attachment. Tag a new one with "archive-tag" and then delete all snapshots but the last one.


CGreweling:
After this one is merged,
this one will follow.
https://github.com/CGreweling/opencast/tree/f/MH-13294-Workflow-track-replacement-deletesnapshots

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerDeleteWorkflowOperationHandler.java
```diff
@@ -73,9 +78,22 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+
     try {
       final AQueryBuilder q = assetManager.createQuery();
-      final long deleted = q.delete(DEFAULT_OWNER, q.snapshot()).where(q.mediaPackageId(mpId)).run();
+      final long deleted;
+
+      if (keepLastSnapshot) {
+        deleted = q.delete(DEFAULT_OWNER, q.snapshot())
+                .where(q.mediaPackageId(mpId).and(q.version().isLatest().not())).run();
+        logger.info("Deleting all but latest Snapshot %s", mpId);
```
gregorydlogan: This renders as `Deleting all but latest Snapshot %s`

CGreweling: @gregorydlogan Why. I do not see the difference to the old code 2 lines below. And that works.
If there is a right way i will correct this in this class. Can you give me a hint?

smarquard: Try:

        logger.info("Deleting all but latest Snapshot {}", mpId);

The code below is explicitly using format(...).

CGreweling: My apologies. I had set the wrong upstream on my local git repo, corrected this here already. The magic of git, is not working with lack of coffee. 
Thanks @smarquard.

CGreweling: changed logging output variables to {}.

PR #623, opened 2018-12-14
MH-13248 Allow hidden workflow parameters

ppettit:
Workflow parameters in form elements of type "hidden" were ignored
when adding a new event, breaking the default schedule-and-upload
workflow.

### DIFF
--- a/modules/admin-ui/src/test/resources/test/unit/shared/services/wizards/new-event/proccessingSpec.js
+++ b/modules/admin-ui/src/test/resources/test/unit/shared/services/wizards/new-event/proccessingSpec.js
@@ -105,6 +105,18 @@ describe('Processing Step in New Event Wizard', function () {
                     .toEqual({ testID: 'testvalueA' });
             });
         });
+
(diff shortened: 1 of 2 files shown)


karendolan:
@ppettit  this looks good to me! Thank you for adding that test.

modules/admin-ui/src/test/resources/test/unit/shared/services/wizards/new-event/proccessingSpec.js
```diff
@@ -105,6 +105,18 @@ describe('Processing Step in New Event Wizard', function () {
                     .toEqual({ testID: 'testvalueA' });
             });
         });
+
+        describe('hidden field', function () {
```
karendolan: Not a blocker, but the test title should probably be "with a hidden field"

ppettit: done!

PR #624, opened 2018-12-15
MH-13283, Fix Custom CXF Error Handler

lkiesow:
This patch fixes two custom CFX error handlers which would not intercept
Opencast's own Exceptions otherwise and cause an internal server error.


slampunk:
This patch fixes the issue. Can be merged.

smarquard:
Thanks @lkiesow and @slampunk for the quick fix and review.

karendolan:
@lkiesow and @slampunk is the fix to explicitly override the toResponse() method and the functional version was not doing that?

lkiesow:
@karendolan, my guess is that CXF explicitly checks for the class type and I broke that check when I used the functional variant since the hook then became a generic callable. I did not dig into this in much detail though.

karendolan:
Ah, the toResponse() method *is* the ExceptionMapper API https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/ext/ExceptionMapper.html. It defaults to  Response.Status.INTERNAL_SERVER_ERROR

PR #625, opened 2018-12-17
MH-13284, Update Elasticsearch to 2.x

lkiesow:
As a first step in the process of bringing Elasticsearch to the latest
stable version, this patch updates Elasticsearch and its client
libraries to the latest 2.x version.

Unnecessary changes in how Opencast and Elasticsearch are avoided if
possible to make the update as smooth as possible though some queries
had to be replaces and the structure had to be adjusted slightly.

*This requires an index rebuild*

### DIFF
--- a/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchQueryBuilder.java
+++ b/modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchQueryBuilder.java
@@ -28,15 +28,10 @@ import org.opencastproject.matterhorn.search.SearchQuery;
 import org.opencastproject.util.DateTimeSupport;
 
 import org.apache.commons.lang3.StringUtils;
-import org.elasticsearch.ElasticsearchException;
-import org.elasticsearch.common.bytes.BytesReference;
 import org.elasticsearch.common.xcontent.XContentBuilder;
-import org.elasticsearch.common.xcontent.XContentType;
 import org.elasticsearch.index.query.BoolQueryBuilder;
-import org.elasticsearch.index.query.FilterBuilder;
-import org.elasticsearch.index.query.FilterBuilders;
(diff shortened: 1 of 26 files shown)


modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchQueryBuilder.java
```diff
@@ -45,33 +40,26 @@
...
 /**
  * Opencast implementation of the elastic search query builder.
  */
-public abstract class AbstractElasticsearchQueryBuilder<T extends SearchQuery> implements QueryBuilder {
+public abstract class AbstractElasticsearchQueryBuilder<T extends SearchQuery> extends QueryBuilder {
 
   /** Term queries on fields */
   private Map<String, Set<Object>> searchTerms = null;
 
-  /** Negative term queries on fields */
```
staubesv, line 63: @lkiesow Could you please provide a few words about why negative term queries, empty fields and non-empty fields are not required (anymore)?

staubesv (changes requested):
I've done a few test and the code seems to work... but please provide some words about the removal of the empty and non-empty fields and negative terms as I don't really understand what they were used for...

modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchQueryBuilder.java
lkiesow: Hm, it has been a while and a Christmas break does not help remembering :-D

But if I remember correctly then this code would need to be written differently since the API had changed and when I looked into this, I noticed that Opencast was never actually using this at all. Hence it seemed like the sensible way to remove this instead of having dead code lying around.

lkiesow:
@staubesv, what change are you requesting?

staubesv:
@lkiesow Just wanted to indicate that there is an open question - no actual request for change.

staubesv (approved):

PR #626, opened 2018-12-18
Provide location of `org.ops4j.pax.web.cfg`

countingpine:
(It's not clear to the uninitiated whether this is a file or a config entry)

### DIFF
--- a/docs/guides/admin/docs/configuration/basic.md
+++ b/docs/guides/admin/docs/configuration/basic.md
@@ -26,7 +26,7 @@ your own domain name:
 processed media. At least not without an extra amount of work involving modifications to the database. That is why you
 should think about this setting carefully.
 
-Second, adjust the binding address in `org.ops4j.pax.web.cfg`. The binding address can be set to `0.0.0.0` for general
+Second, adjust the binding address in `/etc/opencast/org.ops4j.pax.web.cfg`. The binding address can be set to `0.0.0.0` for general
 network access. The property to modify is:
 
     org.ops4j.pax.web.listening.addresses=127.0.0.1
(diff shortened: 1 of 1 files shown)


docs/guides/admin/docs/configuration/basic.md
```diff
@@ -26,7 +26,7 @@ your own domain name:
 processed media. At least not without an extra amount of work involving modifications to the database. That is why you
 should think about this setting carefully.
 
-Second, adjust the binding address in `org.ops4j.pax.web.cfg`. The binding address can be set to `0.0.0.0` for general
+Second, adjust the binding address in `/etc/opencast/org.ops4j.pax.web.cfg`. The binding address can be set to `0.0.0.0` for general
```
lkiesow: The problem with this change is that it's not necessarily true for all installations. E.g. for most custom installations it would be `/opt/opencast/etc/...` and it really could be anywhere else. Most of our documentation uses `etc/org.ops4j.pax.web.cfg` to indicate it's somewhere in the configuration folder. Do you think that wouzld work here as well?

This modification also breaks the style check which requires the line length to be leass than 120 characters. It would be great if you could change that as well.

lkiesow (changes requested):

countingpine:
Hi, thanks for reviewing my patch.  You're right, I neglected to consider there were two likely possible paths, and probably just went from the example edit command.

I've reworked things a bit.  The important thing to me is to be clear it's a file, and it's not necessary to include a path to do that.

I also decided to remove `custom.properties` filename from the file paths - it possibly puts too much emphasis on this one file, and just having it in the example edit command line is helpful enough.

lkiesow (approved):

lkiesow:
Thanks for the patch.
Update looks good
→ merged

PR #628, opened 2018-12-19
MH-13286 Remove unused functionality from MetadataField

KatrinIhler:


karendolan:
@KatrinIhler willl this prevent the temporal field from being accessible? The temporal is where tecnical start and start time  (used to) originate . If this pull is strictly for the use of the Admin UI, it should be applied in the Admin UI/Admin UI Index/AdminUI Search modules. This module is used by many other modules besides Admin UI related ones.

KatrinIhler:
@karendolan This PR only affects the TemporalDate- and TemporalDurationMetadatafields as they are not used by any opencast code as far as I could tell. The TemporalStartDate, TemporalStartTime- and TemporalMetadataField are unaffected by this.

karendolan:
@KatrinIhler Thank you for clarifying. Our site only uses the temporal period style syntax in the dublin core temporal field, as well. I suspect the other definitions were added for completeness of the API. I agree with the goal of simplifying code that has no use case and I believe it actually promotes interoperability for Opencast to narrow the definition of the Temporal dublincore field to DCMI Period syntax.

KatrinIhler:
@karendolan Do you want to take the review for this?

staubesv (approved):

PR #629, opened 2018-12-21
MH-13287 Fix incorrect text metadatafield types

KatrinIhler:
No longer create all text metadata fields with type TEXT, instead use the correct subtype, e.g. TEXT_LONG or TEXT_ORDERED.


staubesv (changes requested):
Please add the new parameter _type_ as @param in the comments.

KatrinIhler:
@staubesv Done!

staubesv (approved):

PR #630, opened 2018-12-21
MH-13288 Don't create incomplete metadata fields

KatrinIhler:
Don't allow the creation of incomplete metadata fields where the functions valueToJSON and JSONToValue are not defined.

~~This PR includes PR #629.~~

PR #631, opened 2018-12-21
MH-13289 Introduce Metadatafield Copy Constructor

KatrinIhler:
Introduce a copy constructor for MetadataField so other copy methods can be simplified or removed completely.

This PR includes the PRs #630  ~~and #629~~.

### DIFF
--- a/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
+++ b/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
@@ -60,7 +60,7 @@ import java.util.TimeZone;
  * @param <A>
  *          Defines the type of the metadata value
  */
-public final class MetadataField<A> {
+public class MetadataField<A> {
 
   private static final Logger logger = LoggerFactory.getLogger(MetadataField.class);
 
@@ -143,6 +143,35 @@ public final class MetadataField<A> {
   private Opt<Map<String, String>> collection = Opt.none();
   private Fn<Opt<A>, JValue> valueToJSON;
   private Fn<Object, A> jsonToValue;
+
+  /**
+   * Copy constructor
+   *
+   * @param other
+   *          Other metadata field
+   */
+  public MetadataField(MetadataField<A> other) {
+
+    this.inputID = other.inputID;
+    this.outputID = other.outputID;
+    this.label = other.label;
+    this.readOnly = other.readOnly;
+    this.required = other.required;
+    this.value = other.value;
+    this.translatable = other.translatable;
+    this.type = other.type;
+    this.jsonType = other.jsonType;
+    this.collection = other.collection;
+    this.collectionID = other.collectionID;
+    this.valueToJSON = other.valueToJSON;
+    this.jsonToValue = other.jsonToValue;
+    this.order = other.order;
+    this.namespace = other.namespace;
+    this.updated = other.updated;
+    this.pattern = other.pattern;
+    this.delimiter = other.delimiter;
+    this.listprovider = other.listprovider;
+  }
 
   /**
    * Metadata field constructor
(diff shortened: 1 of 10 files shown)


modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
```diff
@@ -1041,60 +993,84 @@ public static JObject mapToJSON(Map<String, String> map) {
...
-        this.readOnly = Boolean.valueOf(value);
-        break;
-      case CONFIG_REQUIRED_KEY:
-        this.required = Boolean.valueOf(value);
-        break;
-      case CONFIG_TYPE_KEY:
-        this.type = Type.valueOf(value.toUpperCase());
-        break;
+  public static MetadataField createMetadataField(Map<String,String> dublinCoreProperties) {
+    Opt<String> collectionID = dublinCoreProperties.containsKey(CONFIG_COLLECTION_ID_KEY)
```
pmiddend: Regarding this whole method: There is `Opt.nul`, which you can just pass the result of `dublinCoreProperties.get(...)` (which can be `null`). This way, you save a lot of ternary operators which are (strongly?) discouraged, anyway.

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataUtil.java
```diff
@@ -278,25 +278,39 @@ static void setDuration(DublinCoreCatalog dc, MetadataField<?> field, EName enam
...
+
+      Opt<String> propertyNameOpt = getDublinCorePropertyName(property);
+      Opt<String> propertyKeyOpt = getDublinCorePropertyKey(property);
+
+      if (propertyNameOpt.isSome() && propertyKeyOpt.isSome()) {
+
+        String propertyName = propertyNameOpt.get();
+        String propertyKey = propertyKeyOpt.get();
+
+        Map metadataFieldProperties;
```
pmiddend: Why omit the generic parameters here?

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataUtil.java
```diff
@@ -278,25 +278,39 @@ static void setDuration(DublinCoreCatalog dc, MetadataField<?> field, EName enam
...
+        if (!allProperties.containsKey(propertyName)) {
```
pmiddend: This can be shortened to

```
metadataFieldProperties = allProperties.computeIfAbsent(propertyName, key -> new HashMap<>());
```

pmiddend:
@KatrinIhler I had some time to comment on one or two things, but I have no time to "fully" review this. Sorry.

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataUtil.java
KatrinIhler: Interesting, thanks!

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataUtil.java
KatrinIhler: Whoops, thanks.

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
KatrinIhler: Thanks!

PR #632, opened 2018-12-22
MH-13290, Asset Manager Query Performance

lkiesow:
The asset manager security layer always generates a role based security
query part, even if the user is an administrator, in which case the
generated query part is just discarded later on.

Additionally, a never matching query part is always added in case a user
has no rules which is unlikely but possible. Nevertheless, this is
unnecessary if the user has any roles.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -21,7 +21,6 @@
 package org.opencastproject.assetmanager.impl;
 
 import static com.entwinemedia.fn.Prelude.chuck;
-import static com.entwinemedia.fn.Stream.$;
 import static java.lang.String.format;
 import static org.opencastproject.security.api.SecurityConstants.GLOBAL_ADMIN_ROLE;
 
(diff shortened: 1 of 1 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -153,22 +149,19 @@ private AQueryBuilder q() {
...
-               new Fn2<Predicate, Role, Predicate>() {
-                 @Override public Predicate apply(Predicate predicate, Role role) {
-                   return predicate.or(mkSecurityProperty(q, role.getName(), action).eq(true));
-                 }
-               })
-        .and(restrictToUsersOrganization());
+    return secSvc.getUser().getRoles().stream()
+            .map((role) -> mkSecurityProperty(q, role.getName(), action).eq(true))
+            .reduce(Predicate::or)
+            .orElseGet(() -> q.always().not())
```
lkiesow, line 163: Only add a never matching predicate if there are no role checks because a user has no roles.

In theory, we could even stop here since we already know that the query will never return anything. We wouldn't even need to run that query. But that would require more changes and it is actually very unlikely to happen since every user should usually have at least his own user role.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -181,15 +174,16 @@ private Predicate restrictToUsersOrganization() {
...
       case GLOBAL:
         return true;
       case ORGANIZATION:
         return true;
       default:
-        final AQueryBuilder q = delegate.createQuery();
-        return !q.select().where(p).run().getRecords().isEmpty();
+        return !delegate.createQuery().select()
+                .where(mkAuthPredicate(mediaPackageId, action))
+                .run().getRecords().isEmpty();
```
lkiesow, line 186: `mkAuthPredicate(…)` was called every time `isAuthorized(…)` was called through the result was not necessarily used. It makes more sense to call this in here and only if it is needed.

PR #634, opened 2019-01-07
MH-12290 SAXParserFactory slow to init when called by series listprovider

karendolan:
Adding a default SAXParserFactoryImpl to the system properties resolves the sluggish SAXParserFactory and SAXParser. This speeds up the series listprovider response back up to minmal lag when a system contains many hundreds of series. 

I tested with both "org.apache.xerces.jaxp.SAXParserFactoryImpl" and "com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl". The  first migitates the lag time. But the com.sum inpl was even faster. 

If approved, this patch is applicable to later version branches.

### DIFF
--- a/assemblies/resources/system.properties.append
+++ b/assemblies/resources/system.properties.append
@@ -5,4 +5,5 @@ org.terracotta.quartz.skipUpdateCheck=true
 
 # Configuring DocumentBuilderFactory and TransformerFactory
 javax.xml.parsers.DocumentBuilderFactory=org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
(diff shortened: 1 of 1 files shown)


staubesv:
Hi @karendolan Could you please redirect this r/5.x? There won't be any releases of Opencast 3.x or 4.x anymore as those versions are not supported anymore.

karendolan:
@staubesv thank you for taking this. I'll point it to 5x and also do a test updating the other two (Transformer and DocumentBuilder) to the native Impls. Might as well do all three at once.

PR #636, opened 2019-01-07
MH-12290 prevent SAXParserFactory and SAXParser class load lag in series listprovider

karendolan:
Adding a default SAXParserFactoryImpl to the system properties resolves the sluggish SAXParserFactory and SAXParser class load. This speeds up the series listprovider response back up to minmal lag when an Opencast system contains many hundreds of series.

I tested with both "org.apache.xerces.jaxp.SAXParserFactoryImpl" and "com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl".  Both migitate lag time. But the com.sun internal impl is faster.

Since the pull proposes using the com.sun internal impl, it also proposes updating the other two defaults to JDK javax.xml bundled Impls and to start weaning Opencast from unneeded external Xerces dependencies.

### DIFF
--- a/assemblies/resources/system.properties.append
+++ b/assemblies/resources/system.properties.append
@@ -3,6 +3,7 @@
 net.sf.ehcache.skipUpdateCheck=true
 org.terracotta.quartz.skipUpdateCheck=true
 
-# Configuring DocumentBuilderFactory and TransformerFactory
-javax.xml.parsers.DocumentBuilderFactory=org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
-javax.xml.transform.TransformerFactory=org.apache.xalan.processor.TransformerFactoryImpl
+# Configure for native JDK Impls
+javax.xml.parsers.DocumentBuilderFactory=com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl
+javax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl
+javax.xml.transform.TransformerFactory=com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl
--- a/modules/capture-admin-service-impl/pom.xml
+++ b/modules/capture-admin-service-impl/pom.xml
@@ -51,10 +51,6 @@
       <artifactId>jaxb-api</artifactId>
     </dependency>
     <dependency>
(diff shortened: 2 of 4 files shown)


staubesv (approved):

assemblies/resources/system.properties.append
```diff
@@ -3,6 +3,7 @@
 net.sf.ehcache.skipUpdateCheck=true
 org.terracotta.quartz.skipUpdateCheck=true
 
-# Configuring DocumentBuilderFactory and TransformerFactory
-javax.xml.parsers.DocumentBuilderFactory=org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
-javax.xml.transform.TransformerFactory=org.apache.xalan.processor.TransformerFactoryImpl
+# Configure for native JDK Impls
+javax.xml.parsers.DocumentBuilderFactory=com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl
+javax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl
+javax.xml.transform.TransformerFactory=com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl
```
staubesv, line 9: @karendolan Just found that replacing the xalan stuff here causes https://opencast.jira.com/browse/MH-13311. I verified that undo the change of line 9 fixes MH-13311. Not clear to me whether this requires the re-introduction of some dependencies you removed in the POMs... does it?

karendolan: Hi @staubesv, it's possible that changing back the TransformerFactory without changing the capture-admin and message-broker dependencies back could affect those 2 modules. However,  the coverimage module did not have difficulty finding TransformerFactory and did not have a declared dependency before. Our site is running in production with this change, but we haven't run the coverimage workflow. 

In short, it would be safest to revert all but line 8. Meanwhile, I am curious and will investigate the "serious configuration errors" that the coverimage module has with the Java8 TransformerFactory.

[EDIT] coverimage does declare xalan dependencies and karaf supplies it (this pull did not remove it from the karaf features assembly config). Therefore, it's unlikely the poms will be affected by converting the TransformerFactory. However coverimage module might also complain about the 7 (and possibly even the new inclusion of the 8). So just keeping the line 8, and reverting the rest, is still the safest, IMHO.

Meanwhile I'll look more at the coverimage. The API is the same for both TransformerFactories, but they probably have different default config.

PR #639, opened 2019-01-09
MH-13299 Make multi-select fields consistent again

KatrinIhler:
Make the multi-select metadata fields again consistent with the other fields by restoring the old behavior regarding TAB and blur (saving the value instead of merely hiding it) and discarding the current value completely on ESC.

This was broken by PR #585.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableMultiSelectDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/editableMultiSelectDirective.js
@@ -71,8 +71,9 @@ angular.module('adminNg.directives')
         });
       };
 
-      scope.onBlur = function (event) {
+      scope.onBlur = function () {
         if (!scope.removedValue) {
+          scope.addCurrentValue();
           scope.leaveEditMode();
         } else {
           $timeout(function () {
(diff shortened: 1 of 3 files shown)


staubesv (approved):

modules/admin-ui/src/main/webapp/scripts/shared/directives/editableMultiSelectDirective.js
```diff
@@ -71,8 +71,9 @@ angular.module('adminNg.directives')
         });
       };
 
-      scope.onBlur = function (event) {
+      scope.onBlur = function () {
         if (!scope.removedValue) {
+          scope.addCurrentValue();
```
staubesv, line 76: @KatrinIhler Just realised that this is a bad idea: the scope.AddCurrentValue() on blur causes the value to be stored. Problem: When navigating through the fields using the tab key, this will store the existing values (even if the user has not edited it at all) which creates a new version of the media package.

... we should fix this before 6.2 is released... do you have time?

KatrinIhler: @staubesv I can look into it. So I guess the solution is to only save if the values have changed, right?

staubesv: Yes, similar to editSingleValueDirective which keeps track of the original value.

KatrinIhler: Patch is out: #644

PR #640, opened 2019-01-09
MH-13300 Display multi-value fields correctly on summary pages

staubesv:
The case that the actual value of a metadata field is an array but the metadata field has no collection assigned was not handled correctly.

Steps to reproduce: 
1. Remove list providers from metadata configuration for event and/or series metadata for all multi-value fields (type mixed_text or iterable_text) 
2. Have a look at the summary pages of the Add Event and/or Add Series wizard 
  
 Actual Results: 
The values are not displayed in human readable formats:

screen shot 2019-01-09 at 16 14 45
  
 Expected Results: 
 The values should be displayed in human readable formats.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
@@ -84,33 +84,24 @@ angular.module('adminNg.services')
     };
 
     this.extractPresentableValue = function (field) {
-      var actualValue = field.value;
       var presentableValue = '';
-
-      if (actualValue !== undefined && actualValue !== '' && actualValue !== null) {
-        if (field.collection) {
-          if (angular.isArray(actualValue)) {
-            angular.forEach(actualValue, function (item, index) {
-              presentableValue += item;
-              if ((index + 1) < actualValue.length) {
-                presentableValue += ', ';
-              }
-            });
-            field.presentableValue = presentableValue;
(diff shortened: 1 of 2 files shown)


staubesv:
@KatrinIhler Since you recently came up with quite a number of PRs that relate to this... could you please take the review?

modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
```diff
@@ -84,33 +84,24 @@ angular.module('adminNg.services')
...
-          if (angular.isArray(actualValue)) {
-            angular.forEach(actualValue, function (item, index) {
-              presentableValue += item;
-              if ((index + 1) < actualValue.length) {
-                presentableValue += ', ';
-              }
-            });
-            field.presentableValue = presentableValue;
+      if (field.value !== undefined && field.value !== '' && field.value !== null) {
+        if (angular.isArray(field.value)) {
```
staubesv, line 89: Note that this is the actual bugfix: Before this PR, a field with no collection but an array as value would have led to the else case _presentableValue = field.value;_ which results in the non-human-readable format.
The other changes are just simplifications.

KatrinIhler (approved):
One thing: mixed_text doesn't make sense at all as a type if we have no collection, does it? Currently it behaves the same way as iterable_text...but anyway, that's not really the scope of this PR, which looks good, so lets merge this!

staubesv:
Yep, I should change this to iterable_text in our configuration ;-)

PR #642, opened 2019-01-09
MH-13244 Add override support to external api

KatrinIhler:
Split apart from PR #581 since this change requires a new version of the external API. Don't merge until the parent PR is merged and has reached develop since this change doesn't make any sense without it!

### DIFF
--- a/docs/guides/developer/docs/api/series-api.md
+++ b/docs/guides/developer/docs/api/series-api.md
@@ -461,11 +461,12 @@ __Example__
 
 Updates a series' access policy.
 
-Parameters | Required | Type                  | Description
-:----------|:---------|:----------------------|:-----------
-`acl`      | yes      | [`acl`](types.md#acl) | Access policy to be applied
+Parameters | Required | Type                        | Description                 | Default | Version
+:----------|:---------|:----------------------------|:----------------------------|:--------|:-------
+`acl`      | yes      | [`acl`](types.md#acl)       | Access policy to be applied |         |
+`override` | no       | [`boolean`](types.md#basic) | Whether the episode ACL of all events of this series should be removed | false | 1.2.0
 
-Note that the existing access policy will be overwritten.
+Note that the existing access policy of the series will be overwritten.
 
 __Response__
 
--- a/modules/external-api/src/main/java/org/opencastproject/external/common/ApiVersion.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/common/ApiVersion.java
@@ -24,10 +24,11 @@ package org.opencastproject.external.common;
 public enum ApiVersion {
 
   VERSION_1_0_0(1, 0, 0),
-  VERSION_1_1_0(1, 1, 0);
+  VERSION_1_1_0(1, 1, 0),
+  VERSION_1_2_0(1, 2, 0);
 
   /** The most recent version of the External API */
-  public static final ApiVersion CURRENT_VERSION = VERSION_1_1_0;
+  public static final ApiVersion CURRENT_VERSION = VERSION_1_2_0;
 
   private int major;
   private int minor;
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
@@ -79,7 +79,7 @@ import javax.ws.rs.core.Response;
  * supported API.
  */
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0 })
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0 })
 @RestService(name = "externalapiservice", title = "External API Service", notes = {}, abstractText = "Provides a location for external apis to query the current server of the API.")
 public class BaseEndpoint {
 
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
@@ -29,8 +29,10 @@ import static com.entwinemedia.fn.data.json.Jsons.v;
 import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
+import static org.opencastproject.external.common.ApiVersion.VERSION_1_2_0;
 import static org.opencastproject.util.DateTimeSupport.toUTC;
 import static org.opencastproject.util.RestUtil.getEndpointUrl;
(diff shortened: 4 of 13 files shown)


staubesv:
@KatrinIhler But this adds the new parameter override to all versions of the External API...

Quick intro into External API versioning: The @Produces annotation defines which API versions a request will answer to. This annotation can be set on class level and it can be overridden per method. The introduction of a new External API version requires all those annotations (for all endpoints) to be adapted.

Depending on whether the override parameter is required urgently by some Adopters, it might make sense to wait for other changes that require the introduction of version 1.2.0.

KatrinIhler:
@staubesv Thanks, I'm clearly lacking some knowledge regarding the External API, I'm gonna fix this later. This is not really urgent, so we could let this PR hang around until a new version is introduced by someone else with a more important change. I just wanted to get this out while I still remember what it's about.

KatrinIhler:
@staubesv Okay, next try at getting external API versioning right! ;)

modules/external-api/src/main/java/org/opencastproject/external/common/ApiVersion.java
```diff
@@ -24,7 +24,8 @@
 public enum ApiVersion {
 
   VERSION_1_0_0(1, 0, 0),
-  VERSION_1_1_0(1, 1, 0);
+  VERSION_1_1_0(1, 1, 0),
+  VERSION_1_2_0(1, 2, 0);
 
   /** The most recent version of the External API */
   public static final ApiVersion CURRENT_VERSION = VERSION_1_1_0;
```
staubesv: @KatrinIhler If the comment is correct, you should change CURRENT_VERSION to VERSION_1_2_0

modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
```diff
@@ -79,7 +79,7 @@
  * supported API.
  */
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0 })
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0 ,ApiMediaType.VERSION_1_2_0 })
```
staubesv: A space before the comma seems uncommon

modules/external-api/src/main/java/org/opencastproject/external/common/ApiVersion.java
KatrinIhler: You're right, fixed!

modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
KatrinIhler: Fixed!

staubesv (changes requested):
Can you please add an explanation of the new parameter on https://docs.opencast.org/develop/developer/api/?

KatrinIhler:
@staubesv Done!

docs/guides/developer/docs/api/series-api.md
```diff
@@ -461,11 +461,12 @@ __Example__
 
 Updates a series' access policy.
 
-Parameters | Required | Type                  | Description
-:----------|:---------|:----------------------|:-----------
-`acl`      | yes      | [`acl`](types.md#acl) | Access policy to be applied
+Parameters | Required | Type                        | Description                 | Default | Version
+:----------|:---------|:----------------------------|:----------------------------|:--------|:-------
+`acl`      | yes      | [`acl`](types.md#acl)       | Access policy to be applied |         |
+`override` | no       | [`boolean`](types.md#basic) | Whether the episode ACL of all events of this series should be removed | false | 1.2.0
```
staubesv, line 467: @KatrinIhler I would rather stay with appending " (version 1.2.0 or higher)" to the description (as other examples in the same page) as most fields are available for all External API versions. If we would consistently add a column version to all fields, most values would be 1.1.0 or higher.

What do you think?

KatrinIhler: Actually this is already inconsistent in the external api documentation (there is both this and what you described) and I think this is a bit more clean, but I don't mind either way...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -933,14 +935,22 @@ protected MetadataList deserializeMetadataList(String json) throws ParseExceptio
   @Path("{seriesId}/acl")
   @RestQuery(name = "updateseriesacl", description = "Updates a series' access policy.", returnDescription = "", pathParameters = {
           @RestParameter(name = "seriesId", description = "The series id", isRequired = true, type = STRING) }, restParameters = {
-                  @RestParameter(name = "acl", isRequired = true, description = "Access policy", type = STRING) }, reponses = {
+                  @RestParameter(name = "acl", isRequired = true, description = "Access policy", type = STRING),
+                  @RestParameter(name = "override", isRequired = false, defaultValue = "false", description = "If true the series ACL will take precedence over any existing episode ACL", type = STRING)}, reponses = {
```
staubesv: Should be of type BOOLEAN (requires import static org.opencastproject.util.doc.rest.RestParameter.Type.BOOLEAN;) so that the REST docs page displays a checkbox.

staubesv (changes requested):
I've found two issues:

1. Wrong parameter type for REST docs (see comment)

2. What is it supposed to do?

Reproduction:
1. Create Series A (in Admin UI)
2. Create Event B in Series A with some Event ACL (in Admin UI)
3. PUT /api/series/{id_of_series_a}/acl with '[]' (override: false)

Result: The ACL of Event B is deleted. 

Looks like the event ACL is overriden anyway, but what is the point of the newly introduced form parameter "override" then?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
KatrinIhler: Thanks, fixed!

KatrinIhler:
@staubesv There may be a misunderstanding here: What the 'override' parameter does is remove the episode ACLs when a series ACL is updated (so all events have the same access policies, namely that of the series, and no others). This is disabled by default, but can be enabled on request. This works for me, I tested with the rest docs and curl. (I didn't introduce the override parameter btw, it has been there for ages, I only made it possible to actually set this on a request.)

staubesv:
@KatrinIhler 

> What the 'override' parameter does is remove the episode ACLs when a series ACL is updated [...]

Yes, that is what I would have expected. The problem: I have set override to "false" in my test, but the episode ACL was removed...

staubesv:
@KatrinIhler Ok, now my tests worked.

There is one last cosmetic issue: The REST docs now per default selects the checkbox "override" although the default value is officially "false". I think you can just remove _defaultValue = "false"_ from _@RestParameter_ to fix this.

KatrinIhler:
@staubesv In my opinion the defaultValue shouldn't break the docs like this, but yeah, that's the solution.

staubesv:
@KatrinIhler Agree, I've filed https://opencast.jira.com/browse/MH-13416 for this.

staubesv (approved):

PR #649, opened 2019-01-14
MH-13309: return empty list when finding findUsersByUserName when the name param is empty.

TurRil:
When modifying a group in the Admin UI it might occur that the group does not have a list of users associated with it and when calling the function _findUsersByUserName_ for the appropriate table it throws an error like:

Call: SELECT id, email, last_login, login_mechanism, name, username, organization FROM oc_user_ref WHERE ((organization = ?) AND (username IN ()))

This pull request fixes that oversight and the groups load without further issue.


staubesv:
@TurRil Could you please redirect this to r/6.x?

staubesv (approved):

PR #650, opened 2019-01-16
WIP MH-13310: Fix event deletion

JulianKniephoff:
Ever since #523, events can't be deleted anymore, neither in the admin UI, nor using the external API. The problem only indirectly relates to that PR, though; it just surfaced it. The `AQueryBuilderImpl#always` implementation is to blame, introducing random variables into queries that might not declare them.

This improves the implementation. See also the commit message for more details.

Issue:  [MH-13310](https://opencast.jira.com/browse/MH-13310)

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
@@ -33,6 +33,7 @@ import org.opencastproject.assetmanager.api.AssetId;
 import org.opencastproject.assetmanager.api.AssetManager;
 import org.opencastproject.assetmanager.api.AssetManagerException;
 import org.opencastproject.assetmanager.api.Availability;
+import org.opencastproject.assetmanager.api.DeleteSnapshotHandler;
 import org.opencastproject.assetmanager.api.Property;
 import org.opencastproject.assetmanager.api.Snapshot;
 import org.opencastproject.assetmanager.api.Version;
@@ -140,6 +141,19 @@ public abstract class AbstractAssetManager implements AssetManager {
 
   @Override public void setAvailability(Version version, String mpId, Availability availability) {
     getDb().setAvailability(RuntimeTypes.convert(version), mpId, availability);
+  }
+
+  @Override public long removeEvent(String id, DeleteSnapshotHandler deletionHandler) {
+    AQueryBuilder q = createQuery();
+    long deletedItems = q.delete(DEFAULT_OWNER, q.propertiesOf())
+            .where(q.mediaPackageId(id))
+            .willRemoveWholeMediaPackage(true)
+            .run(deletionHandler);
+    deletedItems += q.delete(DEFAULT_OWNER, q.snapshot())
+            .where(q.mediaPackageId(id))
+            .willRemoveWholeMediaPackage(true)
+            .run(deletionHandler);
+    return deletedItems;
   }
 
   @Override public boolean setProperty(Property property) {
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
@@ -86,6 +86,18 @@ public interface AssetManager {
    */
   void setAvailability(Version version, String mpId, Availability availability);
 
+  /**
+   * Remove an event completely from the asset manager.
+   * @param id the media package ID of the event to remove
+   * @return the number of items deleted
+   */
+  long removeEvent(String id, DeleteSnapshotHandler nopDeleteSnapshotHandler);
+
+  /** Call {@link #removeEvent(String, DeleteSnapshotHandler)} with a callback that does nothing. */
+  default long removeEvent(String id) {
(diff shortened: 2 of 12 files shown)


KatrinIhler:
Note that this problem only appears for Non-Admins.

JulianKniephoff:
Unfortunately this is not enough to fix the problem. The deletion now seemingly "succeeds" without an error, but the event is still there in the admin UI and it also soeems to be in an inconsistent state; inspecting it's metadata for example now raises an error in the log about no snapshot being found, and just reveals empty fields.

I think I have a fix for this as well, though. I will update this PR some time this morning, after I tested it some more.

JulianKniephoff:
Wow, what a ride, but this should work now.

The problem, after fixing the `always` query, was that after #523, the deletion of an event in the asset manager was done by first deleting the properties and then the snapshots. Or was it the other way around? Turns out: It doesn't even matter, because neither order works when you are not an admin. The queries used to do these deletions then get enriched with security predicates which check (a) that there are asset manager properties representing ACL entries stating that you are allowed to delete snapshots, and (b) that you are in the same organization as the thing you are trying to delete.

Now if you delete the properties first, you are not allowed to delete the snapshots anymore, because said ACL properties were just deleted, and if you delete the snapshot first, you can't check whether you are in the correct organization to delete the properties, because that is only stored with the snapshots.

My solution is to introduce `AssetManager#removeEvent`, the implementation of which way down in the asset manager matroshka doll just deletes everything without any checks. This is then wrapped in `AssetManagerWithSecurity` with an extra `select` query that checks all the necessary things. And then I had to refactor a bit because of the complex structure of the asset manager. (You would think that with all this modularity and generality, a change like this should be trivial, but no. Lesson: Not only are you not going to need it (YAGNI), you will probably need something completely different from what you thought (YWNSCDFWYT ...?). :wink:)

I hope this does not destroy the performance gains of #523 too much; maybe @krinnewitz could take a look as well, but yeah, this should be good for review now. :smile:

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
```diff
@@ -142,6 +143,19 @@ public Snapshot takeSnapshot(MediaPackage mediaPackage) {
     getDb().setAvailability(RuntimeTypes.convert(version), mpId, availability);
   }
 
+  @Override public long removeEvent(String id, DeleteSnapshotHandler deletionHandler) {
+    AQueryBuilder q = createQuery();
+    long deletedItems = q.delete(DEFAULT_OWNER, q.propertiesOf())
```
lkiesow, line 148: Is it intentional that this will not delete anything owned e.g. by the scheduler

JulianKniephoff: Well kinda; `removeEvent` is pretty much a direct translation of a part of what the `IndexService` did in its `removeEvent`, where the deletion of "scheduler stuff" and "asset manager stuff" were two discrete steps. I just didn't have the time and/or confidence to touch the scheduling code as well. :sweat_smile:

So what I'm saying is: I agree with you in that it probably shouldn't be like this, but in my defense, it was like this before as well. :wink:

I mean we probably could just delete everything here, that shouldn't break anything, but then the scheduler deletion code would do some redundant work that I would then like to remove, but then who knows where that code is used, and also maybe you sometimes only want to delete the scheduler parts. So we would need two functions for that. And then it is kind of questionable that there is a method in the asset manager, that just deletes everything from itself, but potentially leaves other stuff around, that still depends on that data. (I have no idea whether that would actually be the case; does the scheduler for example store things outside of the asset manager? But if so, that's what I mean by this.) Or is this/should this be handled by `AssetManagerWithMessaging`? :thinking:

Opinions?

JulianKniephoff:
Aaaand another thing came up: When you delete an event that is scheduled but not processed yet, it **is** deleted (well, mostly ...) but the admin UI actually tells you that it failed to do so.

The reason is that `SchedulerService#removeEvent` first last snapshot if it belongs to it. If this is the only one, the `AssetManager#removeEvent` afterward fails in `AssetManagerWithSecurity` because there are no snapshots to join the properties with to determine whether you are allowed to delete them based on the organization you and the peroperties belong to ...

So I guess @lkiesow was right, `AssetManager#removeEvent` should just remove everything. I will see what needs to be refactored for that to work.

KatrinIhler:
@JulianKniephoff Is this gonna happen soon-ish? Otherwise it might be better to just decline this for now and file a new PR again later.

smarquard:
Event deletion in 6.x is very inefficient because of deleteOrphanedProperties() MH-13347, and it looks like this PR is the tail end of a series of changes addressing that.

So for the sake of 6.x adopters, it would be really helpful to get this PR finalized and merged, and then merge back the related changes into 6.x.

JulianKniephoff:
@KatrinIhler unfortunately I don't know yet how soon-ish I can really get back to this, especially considering that the necessary changes are not trivial.

@smarquard more like "tip of the iceberg", I'd say. :sweat_smile:

lkiesow:
Some additional info about this:

We are just evaluating the revert of pull request #523 since it seems to cause the problems with deletion and the patch loses any relevance if pull request #695 is merged since it contains the same performance improvements.

PR #651, opened 2019-01-22
SUREFIRE-1588: Resolving compilation issue on Debian and related distros

gregorydlogan:
This is not really an Opencast bug, but rather caused by a bad backport of a patch by the Debian maintainers.  Unfortunately, Debian Stretch appears to be stuck with the broken version of the JDK, and it's not obvious to someone attempting to build Opencast that the issue comes from upstream.

Upgrading maven-sure-plugin past 3.0.0-M1 (M3 is the most current) resolves the issue, and would need to be done for newer JDKs anyway.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -500,7 +500,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.20</version>
+          <version>3.0.0-M3</version>
         </plugin>
         <plugin>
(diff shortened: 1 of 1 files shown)


lkiesow:
Just to make sure: It's intentionally, that we go to a pre-release? The issue is not fixed in the latest stable release (2.22.1)?

lkiesow:
Read through it, it's fixed in M1. Hence, going with M3 makes sense.
We should go to the stable version as soon as it is out though.

PR #652, opened 2019-01-23
MH-13311 WOH cover-image is broken

staubesv:
This reverts #636 except of line 8 which improves XML parsing performance (see commits).

### DIFF
--- a/assemblies/resources/system.properties.append
+++ b/assemblies/resources/system.properties.append
@@ -5,4 +5,5 @@ org.terracotta.quartz.skipUpdateCheck=true
 
 # Configuring DocumentBuilderFactory and TransformerFactory
 javax.xml.parsers.DocumentBuilderFactory=org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
+javax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl
 javax.xml.transform.TransformerFactory=org.apache.xalan.processor.TransformerFactoryImpl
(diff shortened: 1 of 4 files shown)


karendolan:
This pull looks Ok to me. The comment like "reverts MH-12290 cleanup except for  javax.xml.parsers.SAXParserFactory" in the commit message is helpful.

Going forward, I'm concerned about mixing java8 and Xalan impls on the system class path. This morning, I'll do one more test around providing the Xalan TransformerFactory to cover-image-impl via it's pom imports/includes instead of it falling back to system provided impls. The latest Batik's batik-dom 10.0  has a direct dependency on Xalan 2.7.2. So, upgrading cover-image-impl's packages is not enough to migrate it completely. If that test doesn't work, I'm supressing my concern until a new issue points to it or upgrades make the issue obsolete.

staubesv:
@karendolan Thanks a lot for having a closer look at the causes!

mtneug:
@karendolan will you create a separat pull requests in case you find a fix?

@staubesv I have tested the changes and the `cover-image` operation in the default workflow now succeeds. I must admit that I'm not so familiar with how the internals of class loading works. @karendolan from your comment I take it that this can be merged as is or will this result in other compatibility issues with Xalan?

karendolan:
Hi @mtneug , I will create a separate new pull if other testing is successful. This pull is Ok to merge, in my opinion, no incompatibility is expected. 

Both implement API, so, in theory, they should not cause conflicts, except from older libraries that expect a specific implementation. Therefore, the benefit is that they will expose old libraries that need to be updated :)

PR #656, opened 2019-01-27
MH-13318, Update to Apache Karaf 4.2.2

lkiesow:
This patch updates Apache Karaf to its latest stable version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,7 +42,6 @@
     <osgi.compendium.version>5.0.0</osgi.compendium.version>
     <osgi.core.version>5.0.0</osgi.core.version>
     <osgi.enterprise.version>5.0.0</osgi.enterprise.version>
-    <pax.web.version>7.2.3</pax.web.version>
   </properties>
   <modules>
     <module>modules/admin-ui</module>
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
     <commons-lang3.version>3.8.1</commons-lang3.version>
     <commons-text.version>1.6</commons-text.version>
     <cxf.version>3.2.6</cxf.version>
-    <eclipselink.version>2.7.1</eclipselink.version>
+    <eclipselink.version>2.7.3</eclipselink.version>
     <functional.version>1.4.2</functional.version>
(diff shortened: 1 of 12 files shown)


JamesUoM:
Karaf 4.2.3 was released yesterday (11th Feb) if we want to be bleeding edge.

pom.xml
```diff
@@ -35,14 +35,14 @@
...
-    <karaf.version>4.2.0</karaf.version>
+    <karaf.version>4.2.2</karaf.version>
     <mina.version>2.0.19</mina.version>
     <node.version>v8.12.0</node.version>
     <npm.version>6.4.1</npm.version>
     <osgi.compendium.version>5.0.0</osgi.compendium.version>
     <osgi.core.version>5.0.0</osgi.core.version>
     <osgi.enterprise.version>5.0.0</osgi.enterprise.version>
-    <pax.web.version>7.0.0</pax.web.version>
+    <pax.web.version>7.2.3</pax.web.version>
```
JamesUoM: The news article (and pom) for the Karaf 4.2.2 release list pax web v7.2.5 (though confusingly the release note mention v7.2.4).

pax web 7.2.3 was supported by karaf 4.2.1

lkiesow: Not sure where I did take the version from. The notes about the versions are always kind of messy :-(
I'll look into it and fix it.

JamesUoM: Do we need to specify it at all if Karaf is pulling it in already?

lkiesow: > Do we need to specify it at all if Karaf is pulling it in already?

Good question and it seems like we don't need it. The specified version was used at a few more places but cleaning that out results in a fully functional system so I updated this pull request accordingly.

lkiesow:
> Karaf 4.2.3 was released yesterday (11th Feb) if we want to be bleeding edge.

I did take a short look at it but it seems like the included eclipselink update breaks the OAI-PMH tests. While it could very well be the fault of those tests, I'm leaving it at 4.2.2 for now. I might take a look at it again in a few weeks if things here are a bit quieter and I have more time on my hands.

PR #658, opened 2019-01-27
MH-13320, Asset Manager Performance

lkiesow:
The asset manager security layer ensures user are authorized to access
assets. It does that by checking all of the users roles against all of
the access control rules for a given event.

The problem is that a non-admin user with access to all of the admin
interface has about 100 extra roles due to the concept of role based
visibility in the admin interface. This leads to huge queries with
obviously bad performance.

For example, a default query would look like this:

    select distinct snapshotDto
    from Snapshot snapshotDto
    where snapshotDto.mediaPackageId = ?1 and ((not snapshotDto = snapshotDto or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?3 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?5 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?6 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?7 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?8 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?9 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?10 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?11 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?12 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?13 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?14 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?15 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?16 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?17 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?18 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?19 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?20 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?21 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?22 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?23 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?24 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?25 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?26 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?27 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?28 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?29 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?30 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?31 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?32 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?33 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?34 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?35 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?36 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?37 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?38 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?39 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?40 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?41 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?42 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?43 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?44 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?45 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?46 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?47 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?48 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?49 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?50 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?51 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?52 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?53 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?54 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?55 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?56 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?57 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?58 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?59 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?60 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?61 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?62 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?63 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?64 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?65 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?66 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?67 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?68 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?69 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?70 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?71 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?72 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?73 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?74 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?75 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?76 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?77 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?78 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?79 and p.boolValue = ?4)) and snapshotDto.organizationId = ?80)

UI roles are usually not used for any access control though and could
therefore be excluded in most cases. This patch adds an option to allow
for this pre-filtering of roles. The UI roles are excluded by default.

With the filtering, the same query looks like this:

    select distinct snapshotDto
    from Snapshot snapshotDto
    where snapshotDto.mediaPackageId = ?1 and ((exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?3 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?5 and p.boolValue = ?4) or exists (select p
    from Property p
    where snapshotDto.mediaPackageId = p.mediaPackageId and p.namespace = ?2 and p.propertyName = ?6 and p.boolValue = ?4)) and snapshotDto.organizationId = ?7)

This patch is based on pull request #632

*Work sponsored by SWITCH*

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -42,6 +42,7 @@ import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.security.api.AccessControlEntry;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AuthorizationService;
+import org.opencastproject.security.api.Role;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.security.api.User;
@@ -67,16 +68,23 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private final AuthorizationService authSvc;
   private final SecurityService secSvc;
 
-  private final boolean includeUIRoles;
+  // Settings for role filter
+  private boolean includeAPIRoles;
+  private boolean includeCARoles;
+  private boolean includeUIRoles;
 
   public AssetManagerWithSecurity(
       TieredStorageAssetManager delegate,
       AuthorizationService authSvc,
       SecurityService secSvc,
-      boolean includeUIRoles) {
+      final boolean includeAPIRoles,
+      final boolean includeCARoles,
+      final boolean includeUIRoles) {
     super(delegate);
     this.authSvc = authSvc;
     this.secSvc = secSvc;
+    this.includeAPIRoles = includeAPIRoles;
+    this.includeCARoles = includeCARoles;
     this.includeUIRoles = includeUIRoles;
   }
 
@@ -167,7 +175,7 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private Predicate mkAuthPredicate(final String action) {
     final AQueryBuilder q = q();
     return secSvc.getUser().getRoles().stream()
-            .filter((role) -> includeUIRoles || !role.getName().startsWith("ROLE_UI_"))
+            .filter(roleFilter)
             .map((role) -> mkSecurityProperty(q, role.getName(), action).eq(true))
             .reduce(Predicate::or)
             .orElseGet(() -> q.always().not())
@@ -200,7 +208,7 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
         // check acl rules
         logger.debug("Non admin user. Checking ACL rules.");
         final List<String> roles = secSvc.getUser().getRoles().parallelStream()
-                .filter((role) -> includeUIRoles || !role.getName().startsWith("ROLE_UI_"))
+                .filter(roleFilter)
                 .map((role) -> mkPropertyName(role.getName(), action))
                 .collect(Collectors.toList());
         return selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
@@ -261,4 +269,14 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private String mkPropertyName(String role, String action) {
     return role + " | " + action;
   }
+
+  /**
+   * Configurable filter for roles
+   */
+  private java.util.function.Predicate<Role> roleFilter = (role) -> {
+    final String name = role.getName();
+    return (includeAPIRoles || !name.startsWith("ROLE_API_"))
+        && (includeCARoles  || !name.startsWith("ROLE_CAPTURE_AGENT_"))
+        && (includeUIRoles  || !name.startsWith("ROLE_UI_"));
+  };
 }
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -64,13 +64,17 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private final AuthorizationService authSvc;
   private final SecurityService secSvc;
 
+  private final boolean includeUIRoles;
+
   public AssetManagerWithSecurity(
       TieredStorageAssetManager delegate,
       AuthorizationService authSvc,
-      SecurityService secSvc) {
+      SecurityService secSvc,
+      boolean includeUIRoles) {
     super(delegate);
     this.authSvc = authSvc;
     this.secSvc = secSvc;
+    this.includeUIRoles = includeUIRoles;
   }
 
   @Override public Snapshot takeSnapshot(String owner, MediaPackage mp) {
@@ -160,6 +164,7 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private Predicate mkAuthPredicate(final String action) {
     final AQueryBuilder q = q();
     return secSvc.getUser().getRoles().stream()
+            .filter((role) -> includeUIRoles || !role.getName().startsWith("ROLE_UI_"))
             .map((role) -> mkSecurityProperty(q, role.getName(), action).eq(true))
             .reduce(Predicate::or)
             .orElseGet(() -> q.always().not())
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -51,6 +51,9 @@ import com.entwinemedia.fn.data.Opt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 /**
  * Security layer.
  */
@@ -171,10 +174,6 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
             .and(restrictToUsersOrganization());
   }
 
-  private Predicate mkAuthPredicate(final String mpId, final String action) {
-    return q().mediaPackageId(mpId).and(mkAuthPredicate(action));
-  }
-
   /** Create a predicate that restricts access to the user's organization. */
   private Predicate restrictToUsersOrganization() {
     return q().organizationId().eq(secSvc.getUser().getOrganization().getId());
@@ -184,13 +183,29 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   private boolean isAuthorized(final String mediaPackageId, final String action) {
     switch (isAdmin()) {
       case GLOBAL:
+        // grant general access
+        logger.debug("Access granted since user is global admin");
         return true;
       case ORGANIZATION:
-        return true;
+        // ensure that the requested assets belong to this organization
+        logger.debug("User is organization admin. Checking organization. Checking organization ID of asset.");
+        return snapshotExists(mediaPackageId, secSvc.getOrganization().getId());
       default:
-        return !delegate.createQuery().select()
-                .where(mkAuthPredicate(mediaPackageId, action))
-                .run().getRecords().isEmpty();
+        // check organization
+        logger.debug("Non admin user. Checking organization.");
+        final String org = secSvc.getOrganization().getId();
+        if (!snapshotExists(mediaPackageId, org)) {
+          return false;
+        }
+        // check acl rules
+        logger.debug("Non admin user. Checking ACL rules.");
+        final List<String> roles = secSvc.getUser().getRoles().parallelStream()
+                .filter((role) -> includeUIRoles || !role.getName().startsWith("ROLE_UI_"))
+                .map((role) -> mkPropertyName(role.getName(), action))
+                .collect(Collectors.toList());
+        return selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
+                .map(p -> p.getId().getName())
+                .anyMatch(p -> roles.parallelStream().anyMatch(r -> r.equals(p)));
     }
   }
 
--- a/etc/org.opencastproject.assetmanager.impl.OsgiAssetManager.cfg
+++ b/etc/org.opencastproject.assetmanager.impl.OsgiAssetManager.cfg
@@ -1,0 +1,10 @@
+# Define if UI roles should be included in asset manager access control checks.
+# UI roles are defined as all roles starting with `ROLE_UI_`.
+#
+# Setting this to false will make Opencast's asset manager behave like users do not have these roles at all, leading to
+# far simpler and better performing SQL queries with the obvious downside that no access can be granted based on these
+# roles.
+#
+# Default: false
+#
+#includeUIRoles = false
--- a/etc/org.opencastproject.assetmanager.impl.OsgiAssetManager.cfg
+++ b/etc/org.opencastproject.assetmanager.impl.OsgiAssetManager.cfg
@@ -1,3 +1,25 @@
+# Define if API roles should be included in asset manager access control checks.
+# API roles are defined as all roles starting with `ROLE_API_`.
+#
+# Setting this to false will make Opencast's asset manager behave like users do not have these roles at all, leading to
+# far simpler and better performing SQL queries with the obvious downside that no access can be granted based on these
+# roles.
+#
+# Default: false
+#
+#includeAPIRoles = false
+
+# Define if CA roles should be included in asset manager access control checks.
+# CA roles are defined as all roles starting with `ROLE_CAPTURE_AGENT_`.
+#
+# Setting this to false will make Opencast's asset manager behave like users do not have these roles at all, leading to
+# far simpler and better performing SQL queries with the obvious downside that no access can be granted based on these
+# roles.
+#
+# Default: false
+#
+#includeCARoles = false
+
 # Define if UI roles should be included in asset manager access control checks.
 # UI roles are defined as all roles starting with `ROLE_UI_`.
 #
@@ -8,3 +30,4 @@
 # Default: false
 #
 #includeUIRoles = false
+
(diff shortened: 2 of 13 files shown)


lkiesow:
These logger settings might be handy for testing:
```diff
diff --git a/etc/org.ops4j.pax.logging.cfg b/etc/org.ops4j.pax.logging.cfg
index c494b1830..2d22bd49f 100644
--- a/etc/org.ops4j.pax.logging.cfg
+++ b/etc/org.ops4j.pax.logging.cfg
@@ -26,8 +26,8 @@ log4j2.logger.opencast.level = INFO
 
 # You can specify different log levels for different packages/modules by specifying their package component names. For
 # example, to raise the log level to DEBUG for the rest endpoints contained in the kernel module, set:
-#log4j2.logger.ingest.name = org.opencastproject.ingest
-#log4j2.logger.ingest.level = DEBUG
+log4j2.logger.ingest.name = org.opencastproject.assetmanager.impl.persistence
+log4j2.logger.ingest.level = DEBUG
 
 # For Karaf, Felix & CXF, we want to see some more details in the logs
 log4j2.logger.karaf.name = org.apache.karaf

```

smarquard:
Could this go into 6.x?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -153,22 +153,20 @@ private AQueryBuilder q() {
...
-    return $(secSvc.getUser().getRoles())
-        .foldl(q.always().not(),
-               new Fn2<Predicate, Role, Predicate>() {
-                 @Override public Predicate apply(Predicate predicate, Role role) {
-                   return predicate.or(mkSecurityProperty(q, role.getName(), action).eq(true));
-                 }
-               })
-        .and(restrictToUsersOrganization());
+    return secSvc.getUser().getRoles().stream()
+            .filter((role) -> includeUIRoles || !role.getName().startsWith("ROLE_UI_"))
```
smarquard: A less "magic string" way to do this would be to filter on roles which have a role target of ACL | ALL, or exclude roles with a a target of USER.

Role.java:
```

  /**
   * The target (intended purpose) of a set of roles
   *  USER - Roles which are assigned to users and/or groups to provide access to capabilities
   *  ACL  - Roles which are used to manage access to resources (Event, Series) in an ACL
   *  ALL  - All roles
   */
  enum Target {
    USER, ACL, ALL;
  }

```

smarquard: Actually the above is a nice idea but not currently workable, because I was mixing up Type and Target.

So roles are defined with a type (internal | external) but Target is only currently used for searching when a provider returns a set of roles.

To get this to work, we'd need to add Target to Role and its implementations and their constructors, which may be useful in future but more work than warranted here.

The other roles that probably never show up in event ACLs are ROLE_API_* and ROLE_CAPTURE_AGENT_*

smarquard:
It seems like the number of ACLs for an asset is probably very small, whereas the number of roles that a user could have is potentially very large (especially with some LMS integrations where each course site that you belong to maps to an Opencast role).

So it seems like it would be much more efficient to reverse the lookup order here, that is get the list of roles from the ACL list for the permission sought (e.g. "write"), then check if the user has any of those roles.

lkiesow:
@smarquard, you are right. Using something like this would probably be quite effective:

```sql
select p.propertyName from Property p
  where p.mediaPackageId = {mediaPackageId}
    and p.namespace = {security.namespace}
    and p.boolValue = {true}
```

We do also need to match the organization though.
But I'll take a look at what I can do.

lkiesow:
@smarquard, I've modified the code and used your suggestion where it was possible.

Unfortunately, this is possible only for some more specific queries since the generic `createQuery()` will not let you access any media package identifier or even guarantees its existence at all. The checks work since it's internally joining some tables based on media package identifiers and thus creating the connection.

So no chance for me to easily get the properties unless I start refactoring the whole thing which I won't do and in which case we should address some other problems first ;-)

smarquard:
The ROLE_UI_* and ROLE_CAPTURE_AGENT_*  also show up in the solr queries for the engage search index, and it might be worth filtering them there too. For a separate issue though.

lkiesow:
@smarquard, are there any remaining issues?

smarquard:
On the filtering of roles via prefix:

1. We need to add ROLE_CAPTURE_AGENT_ as a filtered prefix (in addition to ROLE_UI_), because if you have a significant number of CAs (we have > 100), then you have lot of unnecessary roles as well (requested change)

2. It might be cleaner to define a private boolean filterRole(String rolename) rather than repeat the list of role prefixes in a few places in the code (optional change)

3. I'm not that keen on the "magic prefix" approach here, though in the short term (Opencast 7) it's necessary for performance. Longer term (Opencast 8), I think we should add target explicitly to the role, so code that wants to filter can filter on role.target() rather than have a list of prefixes. The role provider that defines the roles can then define explicitly which roles are for ACLs (granting access to events and series), and which are for capabilities (granting access to UI abilities, capture agents, endpoints, etc.) (for future follow-up)

4. The other places which would benefit from filtering roles are where role lists are constructed for index queries (solr or elasticsearch) (for future follow-up)

We have been running the first path of this PR (with ROLE_CAPTURE_AGENT_ added) in our 6.x system for a while.

On the optimization of specific asset manager queries, I had hoped queries like this would show up in our slow query log and I'd have a near before-and-after test case, but they don't anymore. Can you suggest which code paths call these queries, i.e. how to test this?

staubesv:
@lkiesow & @smarquard We might also add ROLE_API_* as those roles are supposed to control the access to External API endpoints of individual API users (and not the access to objects with associated ACLs).

lkiesow:
- Updated to latest `develop`
- Added filter for `ROLE_API_`
- Added filter for `ROLE_CAPTURE_AGENT`
- Put the filters in a single Predicate

> On the optimization of specific asset manager queries, I had hoped queries like this would show up in our slow query log and I'd have a near before-and-after test case, but they don't anymore. Can you suggest which code paths call these queries, i.e. how to test this?

Simply opening the event details in the admin interface will trigger this code. If you want logs including the queries, add something like this to your logging configuration at `etc/org.ops4j.pax.logging.cfg`:

```properties
log4j2.logger.ingest.name = org.opencastproject.assetmanager
log4j2.logger.ingest.level = DEBUG
```

Anything else missing?

smarquard:
Tested on a dev build, no apparent regressions.

PR #659, opened 2019-01-27
MH-13321, Fix Series Item Serialization

lkiesow:
Opencast cannot serialize Boolean objects newly introduced in series
messages for ActiveMQ. We need to use native data types (boolean) or
Stings instead.

This patch fixes the sending of series related messages via ActiveMQ

*Work sponsored by SWITCH*


KatrinIhler (approved):
Reproduced the bug, tested the fix, looks good!

PR #663, opened 2019-01-29
MH-13326: No file/directory found when taking snapshot

rute-santos:


lkiesow (commented):
Generally, this looks good and fixes the problem. The following is nitpicking and I can merge it like this if you want me to:

- all `toString` calls are superfluous
- The logger's internal formatting is preferable since it's just executed if necessary while the `format()` will always generate a formatted sting even if the logs are never produced due to a lower log level.

```java
logger.debug("Archiving manifest of media package {} version {}", mpId, version);
final String manifestFileName = format("manifest_%s_%s.xml", pmp.getMediaPackage().getIdentifier(), version);
```

rute-santos:
Done! Thank you!
Rute

smarquard:
Can we target this at 6.x? Seems like a bugfix.

lkiesow:
I don't mind either way. It doesn't seem to be an urgent fix but it should also be a safe patch. @rute-santos, your call ;-)

lkiesow:
Merged.
Thanks for the patch.

PR #665, opened 2019-01-30
MH-13327 Disable save button on pending request

KatrinIhler:
Disable the save button in the video editor after clicking until the request is finished to avoid users clicking it multiple times.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
@@ -169,14 +169,14 @@ angular.module('adminNg.controllers')
     $scope.player = {};
     $scope.video  = ToolsResource.get({ id: $scope.id, tool: 'editor' });
 
-    $scope.activeTransaction = false;
+    $scope.activeRequest = false;
 
     $scope.submit = function () {
-      $scope.activeTransaction = true;
+      $scope.activeRequest = true;
       $scope.video.thumbnail.loading = $scope.video.thumbnail && $scope.video.thumbnail.type &&
               ($scope.video.thumbnail.type === 'DEFAULT');
       $scope.video.$save({ id: $scope.id, tool: $scope.tab }, function (response) {
-        $scope.activeTransaction = false;
+        $scope.activeRequest = false;
         if ($scope.video.workflow) {
           Notifications.add('success', 'VIDEO_CUT_PROCESSING');
           $location.url('/events/' + $scope.resource);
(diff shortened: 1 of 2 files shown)


modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
```diff
@@ -169,14 +169,14 @@ angular.module('adminNg.controllers')
...
 
     $scope.submit = function () {
-      $scope.activeTransaction = true;
+      $scope.activeRequest = true;
       $scope.video.thumbnail.loading = $scope.video.thumbnail && $scope.video.thumbnail.type &&
               ($scope.video.thumbnail.type === 'DEFAULT');
       $scope.video.$save({ id: $scope.id, tool: $scope.tab }, function (response) {
-        $scope.activeTransaction = false;
+        $scope.activeRequest = false;
         if ($scope.video.workflow) {
```
staubesv, line 180: In case the default thumbnail changes because the user cuts the video, $scope.video.$save will overwrite $scope.video.workflow which causes this logic to fail.

Something like...

```
var closeVideoEditor = $scope.video.workflow;
$scope.video.$save({ id: $scope.id, tool: $scope.tab }, function (response) {
        $scope.activeRequest = false;
        if (closeVideoEditor) {
         ....
```

...makes this work correctly.

staubesv:
@KatrinIhler I've fixed some more things in our code base: https://opencast.jira.com/browse/MH-13330. Also, the "Close" button should be disabled while an active request is going on:

  ```
  <a class="return-button" ng-class="{disabled: activeRequest}" ng-click="leave()" translate="VIDEO_TOOL.BUTTONS.CLOSE">
        <!-- Close -->
 </a>
```

Since those changes are trivial and would likely conflict with this PR, could you please include them in this PR?

staubesv:
@KatrinIhler Actually, quite all controls should be disabled when activeRequest is true.

KatrinIhler:
@staubesv I don't quite agree, I think all controls that don't have any effect outside of the editor (so anything were you'd need to hit save after) should be fine, no? Same goes for Close, why shouldn't that be okay?

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
KatrinIhler: Thanks, done!

gregorydlogan:
This was merged as d7be8edbe9c48ee9e7f1956ac2cb0f04434e90ae, but I screwed up the pull process so this isn't automagically resolving.  I'm going to do that manually.

PR #666, opened 2019-01-30
MH-13328 Remove save button at top of videoeditor

KatrinIhler:
Remove the save button at the top of the video editor as it is rather unintuitive and the duplicate code leads to inconsistencies.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/partials/tools.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/partials/tools.html
@@ -24,22 +24,6 @@
             </span>
             <span class="video-details-series">{{(video.series && video.series.title) ? " - " + video.series.title : "" }}</span></h4>
         </div>
-        <div class="video-save-panel video-more-options video-editor-actions-toolbar" ng-if="tab === 'editor' && player.adapter">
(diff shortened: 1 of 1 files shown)


smarquard:
I'm not convinced this is an improvement from a UX / usability perspective. I personally often use the top save button.

KatrinIhler:
@smarquard I'm convinced for two reasons:
1. Having a save button at the top instead of at the bottom is completely unexpected and unusual, so much that a colleague of mine who also works on Opencast didn't know it existed until I pointed it out to him. Having it exist twice even more so. This makes for bad usability because software should adhere to common conventions and not reinvent the wheel at every turn.
2. We have a lot of duplicate code in the Opencast UI, which can and does lead to inconsistencies because even developers simply don't expect there to be two save buttons. Currently the rules that define when the save button is deactivated are not the same for the top and bottom.

I also don't really understand why you'd need a save button at the top, if you're in the video editor you're probably doing some cutting or other stuff which is near the bottom anywhere. Just starting a workflow can easily been done from the Admin UI.

gregorydlogan:
@KatrinIhler conflicts due to #665

KatrinIhler:
@gregorydlogan Fixed!

PR #668, opened 2019-02-01
MH-13329 Removing a capture agent resets the password of all Opencast users

staubesv:
The method getPassword() returns the encoded user password. If we set this as new password, it will be encoded again resulting in the user password being changed. If we set an empty string, the old encoded password will be re-used.

Steps to reproduce: 
1. Login as admin:opencast 
2. Create another user test:test 
3. Login as test 
4. Login as admin:opencast again 
5. Create a capture agent "test" 
6. Delete the capture agent "test" 
7. Login as test:test 
  
 Actual Results: 
 You cannot login as test:test because the password "test" is not valid anymore. 
  
 Expected Results: 
 Deleting a capture agent should not reset the passwords of users. 
  
 Workaround (if any): 
Do not delete capture agents ;-)

### DIFF
--- a/modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentAdminRoleProviderImpl.java
+++ b/modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentAdminRoleProviderImpl.java
@@ -146,7 +146,7 @@ public class CaptureAgentAdminRoleProviderImpl implements RoleProvider, CaptureA
               .map(role -> (JpaRole)role)
               .collect(Collectors.toSet());
       try {
-        this.userAndRoleProvider.updateUser(new JpaUser(user.getUsername(), user.getPassword(),
+        this.userAndRoleProvider.updateUser(new JpaUser(user.getUsername(), "",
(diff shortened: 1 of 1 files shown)


smarquard:
Weird bug. Isn't there a more efficient way to remove a role than iterating through all the users? Sounds inefficient, especially if you had 100K internal users or something.

staubesv:
@smarquard I thought about removing this completely, but as this bug is quite crazy for people that use a lot of Opencast users, it might be worth fixing it quickly...

smarquard:
Interesting example of how a suboptimal API design can induce coding errors, as the before code above looks quite reasonable and it's not obvious that getPassword() is get encrypted password, and setPassword() is set password from plain text string.

Probably we should only have setPassword(String plaintext) and boolean checkPassword(String plaintext) with no getPassword() method.

It's also a little odd to pass an empty string for the password value to mean "no password" rather than null (though obviously here that's what updateUser() requires).

smarquard (approved):
Haven't tested, but code looks good.

lkiesow:
This provides a nice quick-fix → merged and will be included in the next maintenance release.

I agree that this behavior is weird in general. That is why I created pull request #675 to make the user provider handle `null` as empty password and also pull request #676 which removes the user modifications when removing a capture agent alltogether.

PR #669, opened 2019-02-01
MH-13333 Do not import properties in publish WF

Rillke:
Do not import the Java properties file in publish workflow because
that would override user choices, if not correctly filtered by tag.

### DIFF
--- a/etc/workflows/publish.xml
+++ b/etc/workflows/publish.xml
@@ -105,15 +105,26 @@
 
     <!-- Apply the default workflow configuration -->
 
+    <!--
+    The following operation allows you to import properties from
+    a Java properties file, which might have been created by the
+    upload-and-schedule workflow.
+    However, whatever you import here will override the choices
+    your users made in the configuration panel. That's why it is
+    commented.
+    -->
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Hi @Rillke, can you give me an example of how this causes issues?  Are you saying the user's publishing channel choices are discarded at runtime?

pawohl:
Hi @gregorydlogan - the issue is described in detail in the JIRA ticket https://opencast.jira.com/projects/MH/issues/MH-13333?filter=allissues

> Are you saying the user's publishing channel choices are discarded at runtime?

In fact all of the user's choices in the publish workflow are overridden, including but not limited the channels and the video quality.

Screen recording (no sound, only 2 minutes): https://users.informatik.uni-halle.de/~abrcs/OC5-WF-Publish.mp4

At 00:53 I am unticking some publication channels and have already decided I want a full-HD video. At 01:05 you can see how the values were overridden during the WF execution. At 01:22 you see the outcome: Publication to all channels and at 01:44 - low resolution video despite full-HD requested.

PR #674, opened 2019-02-05
MH-13339, Handle Bad User Update Requests

lkiesow:
The admin interface REST endpoint for updating user data does not handle
invalid requests gracefully but just produces internal server errors.

This patch properly parses the given data and returns an HTTP Bad
Request if invalid data is sent.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
@@ -28,6 +28,7 @@ import static com.entwinemedia.fn.data.json.Jsons.v;
 import static java.lang.String.CASE_INSENSITIVE_ORDER;
 import static org.apache.commons.lang3.StringUtils.trimToEmpty;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
+import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
 import static org.apache.http.HttpStatus.SC_CONFLICT;
 import static org.apache.http.HttpStatus.SC_CREATED;
 import static org.apache.http.HttpStatus.SC_FORBIDDEN;
@@ -72,6 +73,9 @@ import com.entwinemedia.fn.data.json.Field;
 import com.entwinemedia.fn.data.json.JObject;
 import com.entwinemedia.fn.data.json.JValue;
 import com.entwinemedia.fn.data.json.Jsons;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.reflect.TypeToken;
 
 import org.apache.commons.lang3.StringUtils;
 import org.json.simple.JSONArray;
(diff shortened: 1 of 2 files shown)


lkiesow:
I've added another patch which fixes a number of inconsistencies I noticed in the user endpoint.

smarquard:
Is there something you can do in the UI to produce these internal server errors (before this PR), or would this only happen if you passed bad data to this endpoint from source other source like a script?

Put another way, is this testable via the UI?

lkiesow:
I don't think the admin interface would send data which would cause this issue but you can test the issue.
For example, go to the `PUT /{username}.json` endpoint, and use `123` as the data for roles. It should give you a BAD REQUEST response but it's actually producing an internal server error.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
```diff
@@ -200,74 +193,79 @@ public Response getUsers(@QueryParam("filter") String filter, @QueryParam("sort"
...
       // Filter list
-      if (filterName.isSome() && !filterName.get().equals(user.getName())
-              || (filterRole.isSome()
-                  && !Stream.$(user.getRoles()).map(getRoleName).toSet().contains(filterRole.get()))
-              || (filterProvider.isSome()
-                  && !filterProvider.get().equals(user.getProvider()))
-              || (filterText.isSome()
-                  && !TextFilter.match(filterText.get(), user.getUsername(), user.getName(), user.getEmail(), user.getProvider())
-                  && !TextFilter.match(filterText.get(), Stream.$(user.getRoles()).map(getRoleName).mkString(" ")))) {
+      final String finalFilterRole = filterRole;
```
staubesv, line 196: What is the purpose of finalFilterRole? This variable seems superfluous to me...

staubesv (changes requested):
Found a superfluous variable... besides that, seems ok.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
lkiesow: This is important since it declares the value as `final`. I couldn't use it later otherwise.

staubesv: Weird... Please explain: Why could you not use the variable filterRole directly in lines 198/199?

lkiesow: Because `filterRole` is neither final nor effectively final but Java requires that for use in lambdas, …

If you still do not believe me, please try it or read the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se10/html/jls-15.html#jls-15.27.2)

staubesv: Ok, thanks for the explanation.

staubesv (approved):

PR #677, opened 2019-02-05
Documentation: Dictionary service

Rillke:
- Fixed some copy & paste glitches in code's inline documentation.
- Added a point to the admin docu: How does this regexp-thingy
  exactly work (so admins can craft a suitable RegExp).


lkiesow:
Looks good → merged.
Thanks for the patch.

PR #678, opened 2019-02-05
MH-13342 Don't try to create events with empty metadata

KatrinIhler:
This is an attempt to fix a bug where an event cannot be created because the metadata sent to the backend are empty, resulting in a server error complaining about an empty title. So far this could only be reproduced in Firefox when trying to create an event for the first time, and it doesn't happen consistently.

This PR rewrites the save method that currently effectively doesn't save anything (the metadata fields that are attached to the userdata under their field id are duplicates and are discarded in the backend on event creation) and only works because of the metadata fields in the userdata being the exact same objects used as models for the input fields. When this isn't true (and it seems it sometimes isn't) no values are saved. This PR saves the metadata fields explicitly and directly into the userdata array to work around this problem.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-event/metadata.js
@@ -51,7 +51,7 @@ angular.module('adminNg.services')
           // preserve default value, if set
           if (field.hasOwnProperty('value') && field.value) {
             field.presentableValue = me.extractPresentableValue(field);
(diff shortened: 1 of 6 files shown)


JamesUoM:
A general comment is that the metadata.js and metdataExtended.js files for the event and series appear to all intents identical.

KatrinIhler:
@JamesUoM Well, they _should_ be identical, there are multiple places where they currently are not, probably mostly because metadataExtended is forgotten on updates... it's somewhere on my to-do-list to merge this into one, but no idea when/if that's gonna happen.

PR #679, opened 2019-02-05
MH-13343 Load track into workspace with unique ID

KatrinIhler:
When extracting multiple thumbnails during one workflow operation in the composer service, load the track into the workspace with a unique id so the different jobs don't step on each other's toes.


smarquard:
As this is a bugfix affecting 6.x, please target at the 6.x branch

KatrinIhler:
@smarquard Done, and I fixed a test as well.

PR #680, opened 2019-02-05
MH-13344 Enable AssetManager to reply NOT_MODIFIED

KatrinIhler:
This enables the Asset Manager to reply NOT_MODIFIED in case the Workspace already has the file with a matching md5 hash, similarly to how the Working File Repository does it.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
@@ -75,6 +75,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.security.NoSuchAlgorithmException;
 import java.util.Date;
 
 /**
@@ -154,13 +155,22 @@ public abstract class AbstractAssetManager implements AssetManager {
     // try to fetch the asset
     for (final AssetDtos.Medium asset : getDb().getAsset(RuntimeTypes.convert(version), mpId, mpeId)) {
       for (final InputStream assetStream : getLocalAssetStore().get(StoragePath.mk(asset.getOrganizationId(), mpId, version, mpeId))) {
+
+        Checksum checksum = null;
+        try {
+          checksum = Checksum.fromString(asset.getAssetDto().getChecksum());
+        } catch (NoSuchAlgorithmException e) {
+          logger.warn("Invalid checksum for asset {} of media package {}", e, mpeId, mpId);
+        }
+
         final Asset a = new AssetImpl(
                 AssetId.mk(version, mpId, mpeId),
                 assetStream,
                 asset.getAssetDto().getMimeType(),
                 asset.getAssetDto().getSize(),
                 asset.getStorageId(),
-                asset.getAvailability());
+                asset.getAvailability(),
+                checksum);
         return Opt.some(a);
       }
     }
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
@@ -39,6 +39,7 @@ import org.opencastproject.assetmanager.impl.storage.DeletionSelector;
 import org.opencastproject.assetmanager.impl.storage.Source;
 import org.opencastproject.assetmanager.impl.storage.StoragePath;
 import org.opencastproject.mediapackage.MediaPackageElement;
+import org.opencastproject.util.Checksum;
 import org.opencastproject.util.MimeTypes;
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.RequireUtil;
@@ -56,6 +57,7 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.security.NoSuchAlgorithmException;
 import java.util.Date;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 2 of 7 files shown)


smarquard:
There's no description in the JIRA for this issue. Under what circumstances is this helpful? What performance improvement does it produce?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -303,6 +328,14 @@ public Response getAsset(@PathParam("mediaPackageID") final String mediaPackageI
     }
   }
 
+//
```
smarquard, line 331: Looks like this was left in by mistake.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
```diff
@@ -154,13 +155,22 @@ public Snapshot takeSnapshot(MediaPackage mediaPackage) {
     // try to fetch the asset
     for (final AssetDtos.Medium asset : getDb().getAsset(RuntimeTypes.convert(version), mpId, mpeId)) {
       for (final InputStream assetStream : getLocalAssetStore().get(StoragePath.mk(asset.getOrganizationId(), mpId, version, mpeId))) {
+
+        Checksum checksum = null;
+        try {
+          checksum = Checksum.fromString(asset.getAssetDto().getChecksum());
+        } catch (NoSuchAlgorithmException e) {
+          logger.warn("Invalid checksum for asset {} of media package {}", e, mpeId, mpId);
```
lkiesow, line 163:
```suggestion
          logger.warn("Invalid checksum for asset {} of media package {}", mpeId, mpId, e);
```

…otherwise `e` would be put into the first placeholder.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
```diff
@@ -326,13 +328,22 @@ private boolean isRemoteAssetStoreId(String storeId) {
       for (final String storageId : getSnapshotStorageLocation(version, mpId)) {
         for (final AssetStore store : getAssetStore(storageId)) {
           for (final InputStream assetStream : store.get(StoragePath.mk(asset.getOrganizationId(), mpId, version, mpeId))) {
+
+            Checksum checksum = null;
+            try {
+              checksum = Checksum.fromString(asset.getAssetDto().getChecksum());
+            } catch (NoSuchAlgorithmException e) {
+              logger.warn("Invalid checksum for asset {} of media package {}", e, mpeId, mpId);
```
lkiesow, line 336: same here:

```suggestion
              logger.warn("Invalid checksum for asset {} of media package {}", mpeId, mpId, e);
```

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -278,13 +285,31 @@ public Response getMediaPackage(@PathParam("mediaPackageID") final String mediaP
...
+
     try {
       for (final Version v : getAssetManager().toVersion(version)) {
         for (Asset asset : getAssetManager().getAsset(v, mediaPackageID, mediaPackageElementID)) {
+
+          if (StringUtils.isNotBlank(ifNoneMatch)) {
+            Checksum checksum = asset.getChecksum();
+
+            if (checksum == null || !checksum.getType().equals((ChecksumType.DEFAULT_TYPE))) {
+              checksum = Checksum.create(ChecksumType.DEFAULT_TYPE, asset.getInputStream());
```
staubesv, line 299: @KatrinIhler Hmm... just wondering: How expensive is this operation for large files and when will it be called? Not sure whether the asset manager allows this, but if it does: Maybe check for equal file size before doing a checksum (in means of an optimization)?

Oh, just realised that you only create a new checksum when the asset does not have one. Hmm... maybe log a warn then or think about updating it in the database...

staubesv: @KatrinIhler Just thought about this again: 

I would recommend to not calculate the checksum in case it is missing (or of wrong type) but rather just log a warning message ("checksum missing").
At the end of the day, not having a checksum is a data consistency error and just masking this error by recalculating it on-the-fly should not be necessary. Note that of more than 3 million assets in our production database, none has no checksum.

KatrinIhler: I agree, I'm gonna change this.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManagerWithTieredStorage.java
KatrinIhler: Thanks!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AbstractAssetManager.java
KatrinIhler: Thanks!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
KatrinIhler: Whoops, you're right.

KatrinIhler:
Deleted the remote branch by accident which closed this PR and apparently I can't reopen it, so I created #685 with an updated version instead.

PR #664, opened 2019-01-29
MH-13147: OptimisticLockException in ServiceRegistry dispatchJob

rute-santos:
It turns out that the blocking/blocked job info stored in the database was not used. It is enough to keep that info in memory, even when Opencast is restarted. Removing the need to update it in the database prevents the optimistic error.
The problem is difficult to reproduce so it's hard to test the fix, but we have been successfully using this fix in production (our version is "5.x + tiered asset manager") for a few months.

PR #683, opened 2019-02-06
MH-13147 (6.x): OptimisticLockException in ServiceRegistry dispatchJob

rute-santos:


lkiesow:
@rute-santos, can you please add a description to this pull request and explain what you are trying to do here? The title only gives you a vague idea and the commit description does not contain a lot of information as well.

And this is modifying the database, isn't it? That would cause massive problems within a maintenance update. Please redirect this to `develop` for Opencast 7. Additionally, this is missing a database upgrade script and the necessary modification to the database set-up script. Can you add that?

smarquard:
@lkiesow, this is a 6.x version of #664 - see discussion there.

smarquard:
We have this code running in our 6.x production system and hasn't caused any issues so far.

smarquard:
Hm, apologies, I merged the wrong PR here...

smarquard:
This looks like a clean change, and applies to 6.x without requiring any database modifications. The db conversions will get handled in PR #664.

We have been running this at UCT in production for a few days, and have seen no further OptimisticLockExceptions and also no regressions in workflow or job handling.

PR #685, opened 2019-02-07
MH-13344 Enable AssetManager to reply NOT_MODIFIED

KatrinIhler:
Updated version of #680, deleted my remote branch by accident which closed the PR...

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
@@ -25,6 +25,7 @@ import static javax.servlet.http.HttpServletResponse.SC_CREATED;
 import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
 import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_MODIFIED;
 import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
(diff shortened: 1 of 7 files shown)


KatrinIhler:
Currently the workspace can download assets directly into the workspace, and will do so even if the exact same asset is already contained in the workspace. This PR tries to emulate the behavior of the WFR, which will return NOT_MODIFIED instead if the workspace already has the asset with a matching md5 checksum. This will only happen if the "If-none-match" header is set on request, otherwise the behavior of the Endpoint is unchanged.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -278,13 +285,33 @@ public Response getMediaPackage(@PathParam("mediaPackageID") final String mediaP
...
+                           @HeaderParam("If-None-Match") String ifNoneMatch) {
+
     try {
       for (final Version v : getAssetManager().toVersion(version)) {
         for (Asset asset : getAssetManager().getAsset(v, mediaPackageID, mediaPackageElementID)) {
+
+          if (StringUtils.isNotBlank(ifNoneMatch)) {
+            Checksum checksum = asset.getChecksum();
+
+            if (checksum != null & checksum.getType().equals(ChecksumType.DEFAULT_TYPE)) {
```
gregorydlogan: Single `&` here, which I suspect is not the goal.  Also, we're checking the type, under what circumstances can this *not* be `DEFAULT_TYPE`?  Why do we care what the type is if we're getting the checksum and checking it against the `ifNoneMatch` field?

KatrinIhler: I guess right now it can't, I mainly wanted to make sure that both checksums were created by the same algorithm even in the future (although for them to match when that isn't the case is not very probable..)

KatrinIhler: Fixed the &.

PR #688, opened 2019-02-08
MH-13350 Theodul core HTML validation

Rillke:
+ alt atttributes to img
+ use IDs only once
+ separate attributes by a space character

Validator:
- https://html5.validator.nu/

Reference:
- https://www.w3.org/TR/2012/WD-html-markup-20120320/syntax.html

### DIFF
--- a/modules/engage-theodul-core/src/main/resources/ui/core.html
+++ b/modules/engage-theodul-core/src/main/resources/ui/core.html
@@ -13,7 +13,7 @@
   <body>
     <noscript>
       <div class="noJavaScript">
-        <img src="img/opencast.svg" class="loadingImg" />
+        <img src="img/opencast.svg" alt="" class="loadingImg" />
         <h1>Error</h1>
         <p id="noJavaScript-container">
           JavaScript is not enabled in your browser.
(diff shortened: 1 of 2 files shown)


modules/engage-theodul-core/src/main/resources/ui/core.html
```diff
@@ -13,7 +13,7 @@
   <body>
     <noscript>
       <div class="noJavaScript">
-        <img src="img/opencast.svg" class="loadingImg" />
+        <img src="img/opencast.svg" alt="Opencast word and figurative mark" class="loadingImg" />
```
smarquard: I would just call this "Opencast logo" which is simpler and shorter for screen-readers (and also where it occurs below)

lkiesow (approved):
Looks good. What do you think about @smarquard's suggestion?

modules/engage-theodul-core/src/main/resources/ui/core.html
Rillke: After reading https://www2.le.ac.uk/webcentre/plone/build/basics/add-images/alt-text and https://ux.stackexchange.com/questions/21518/how-detailed-should-a-text-alternative-version-of-an-infographic-be - still not sure and I have no personal experience with screen readers and how it feels if they read possibly not strictly relevant information. So I trust your experience here and will change it to `Opencast logo`.
It all depends on whether you'd like to make people aware what the logo is comprised of or whether you just want to tell there is a logo. The latter is probably enough for a media player.

smarquard: Looking more closely at this and https://www.webcredible.com/blog/writing-effective-alt-text-images/, it might be appropriate to just use alt="" as the logo could be regarded as decorative here. The context for this is the no-javascript warning, and below the error page and loading animation. So this image isn't functional for the user in any way. I don't have any particular screen-reader or accessibility expertise though, so this is just from googling.

Rillke: > Conversely, you could argue that the images on your site create a brand identity and by hiding them from screen reader users you're denying this group of users the same experience. Accessibility experts tend to favour the former argument, but there certainly is a valid case for the latter too.

If the missing branding is not an issue for Opencast, I'd say not mentioning the image is the best solution for those who depend on screen readers.

lkiesow:
Okay, seems the discussion came to a final conclusion and in any case, this is definitely an improvement → merged
Thanks for the patch, @Rillke

PR #689, opened 2019-02-08
MH-13275 - Allows the workflow to select the audio track for composite videos

phoebem:
This is a redo of an older pull request #618


modules/admin-ui/package-lock.json
```diff
@@ -2339,8 +2339,7 @@
         "ansi-regex": {
```
lkiesow, line 2339: You probably did not want to submit this

gregorydlogan: Concur with removing this.

staubesv (approved):

PR #690, opened 2019-02-08
Revert "MH-13243, Asset Manager ACL Cache Updates"

staubesv:
This reverts commit 271df274c22423e73dd43f89f6ee3206b61230d6 because is causes massive performance problems in systems that have large number of objects in the tables oc_assets_asset and oc_assets_properties.

The problem addressed by MH-13243 should be fixed at some point of time, but as we don't have this on our radar in the near future (and it does not seem like anybody else does), it's probably best to revert it for now.

@wsmirnow & @ts23 (5.x) and @mtneug & @lkiesow (6.x): This is a highly critical release blocker that essentially totally breaks Opencast if the tables mentioned above are large enough.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -210,17 +210,10 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator {
   }
 
   private void storeAclAsProperties(Snapshot snapshot, AccessControlList acl) {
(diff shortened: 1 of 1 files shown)


lkiesow:
> The problem addressed by MH-13243 should be fixed at some point of time, but as we don't have this on our radar in the near future (and it does not seem like anybody else does), it's probably best to revert it for now.

Not true. Here is a patch which fixes the problem properly;

- https://github.com/lkiesow/opencast/commit/88f04a18c396e2436c843912203a9dcb28489eb3

This has already been tested on UCT's production cluster and seems to solve the performance issue. I'll bring out a proper pull request for this plus a few more performance patches in the next days.

I'm closing this since the problem gets a proper fix and this would introduce a security issue.

PR #691, opened 2019-02-09
Documentation: Developer Console: How to shutdown

pawohl:
>> osgi:shutdown

> Command not found: osgi:shutdown

Help says there is `system:shutdown`.


lkiesow:
`system:shutdown` makes sense → merged

PR #692, opened 2019-02-09
MH-13310, Simplify `AQueryBuilderImpl#always`

lkiesow:
The previous implementation basically created a predicate of the form
`q == q`, where `q` is some query variable (here `Q_SNAPSHOT`),
to create a predicate that is always true.
This works but introduces `q` as a variable in the `WHERE` clause
of every query that uses this combinator, which in turn requires
a matching declaration in the `FROM` clause, even though a query
in might not even need `q`.

The new implementation is still a bit roundabout. I would have
preferred the empty condition to just look like `true` when rendered
to SQL, but that is not valid JPQL. `true = true` works, though.

@JulianKniephoff, please forgive me, I bluntly stole your commit from the closed pull request #650 since I think that this one does make sense in any case.


smarquard (approved):
Code looks good. We have merged and deployed this in our 6.x build and has not caused any issues that I'm aware of.

JulianKniephoff:
@lkiesow it's fine; I'm glad to see at least some of the work from #650 bear fruit, seeing as how that one turned into a bit of a trainwreck. :sweat_smile:

PR #693, opened 2019-02-10
MH-13352, Unnecessary Format

lkiesow:
The asset manager database class is constantly preparing formatted query
strings for logging even though they are only necessary when the log
level is set to debug.

This patch uses the formatter build into the logger instead to only
prepare the strings when they are actually needed.


smarquard:
Looks good.

PR #694, opened 2019-02-11
MH-13354, Asset Manager Property Performance

lkiesow:
All asset manager delete queries suffer from very poor performance in
large systems since they always probing for orphaned properties of any
media packages and will always create a joined table of all snapshots
and all properties in the process, even if it is completely unnecessary
like in the case of updating the ACL properties.

This patch fixes the problem by introducing a way to directly delete
specific properties in a single, well performing database query.

### DIFF
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
@@ -96,6 +96,24 @@ public interface AssetManager {
    */
   boolean setProperty(Property property);
 
+  /**
+   * Delete all properties for a given media package identifier
+   *
+   * @param mediaPackageId
+   *          Media package identifier
+   */
+  void deleteProperties(String mediaPackageId);
+
+  /**
+   * Delete all properties for a given media package identifier and namespace.
+   *
+   * @param mediaPackageId
+   *          Media package identifier
+   * @param namespace
+   *          A namespace prefix to use for deletion
+   */
+  void deleteProperties(String mediaPackageId, String namespace);
+
   /** Create a new query builder. */
   AQueryBuilder createQuery();
 
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
@@ -77,6 +77,7 @@ public class AssetManagerWithSecurity extends AssetManagerDecorator<TieredStorag
   }
 
   @Override public Snapshot takeSnapshot(String owner, MediaPackage mp) {
+
     final String mediaPackageId = mp.getIdentifier().toString();
     final AQueryBuilder q = q();
     final AResult r = q.select(q.snapshot())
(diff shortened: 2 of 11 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -208,13 +209,18 @@ private AdminRole isAdmin() {
...
+   *          ACL to set
+   */
   private void storeAclAsProperties(Snapshot snapshot, AccessControlList acl) {
     final String mediaPackageId =  snapshot.getMediaPackage().getIdentifier().toString();
     // Drop old ACL rules
-    final AQueryBuilder queryBuilder = createQuery();
-    queryBuilder.delete(snapshot.getOwner(), queryBuilder.propertiesOf(SECURITY_NAMESPACE))
-            .where(queryBuilder.mediaPackageId(mediaPackageId))
-            .run();
+    super.deleteProperties(mediaPackageId, SECURITY_NAMESPACE);
```
JulianKniephoff, line 223: Note that this no longer does any authorization. Is this okay?

I get that we don't need to check against ACL properties here, since this is the place that creates these, but the "are you in the right organization (and potentially even its admin)" check is circumvented as well.

lkiesow: That's why I documented this at the method :-)

This is a private method which is only used in `takeSnapshot()` after assuring that the user has the necessary privileges.

JulianKniephoff: Right, I read the JavaDoc comment of course ... *cough* :wink: 

Forget I ever said anything. :sweat_smile:

JulianKniephoff:
Looks good to me. :smile:

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
```diff
@@ -96,6 +96,24 @@
    */
   boolean setProperty(Property property);
 
+  /**
+   * Delete all properties for a given media package identifier
+   *
+   * @param mediaPackageId
+   *          Media package identifier
+   */
+  void deleteProperties(String mediaPackageId);
```
gregorydlogan, line 105: I would be tempted to rename this delete**All**Properties so that it's clearer.

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
```diff
@@ -96,6 +96,24 @@
...
+
+  /**
+   * Delete all properties for a given media package identifier and namespace.
+   *
+   * @param mediaPackageId
+   *          Media package identifier
+   * @param namespace
+   *          A namespace prefix to use for deletion
+   */
+  void deleteProperties(String mediaPackageId, String namespace);
```
gregorydlogan, line 115: I would be tempted to rename this delete**All**Properties so that it's clearer.

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManager.java
lkiesow: But this does not delete all properties. It only deletes those who fit the given criteria which is the media package identifier here and the media package identifier and the namespace for the other method.

To me `assetManager.deleteAllProperties(…)` sounds like it's dropping all properties the asset manager has stored.

PR #695, opened 2019-02-11
MH-13347, Don't always look for orphaned properties

lkiesow:
For every delete query executed by the asset manager and regardless of
what is being deleted, the asset manager will look for any orphaned
asset manager properties (defined as properties for media packages for
which no snapshot exists) by joining the complete snapshot table with
the complete property table which is a quite expensive operation.

This patch will remove this procedure and instead add a simple,
performant cleanup procedure for a single media package before the first
snapshot is being created.

*Note that this uses the fast property deletion of pull request #694*

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/query/AbstractADeleteQuery.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/query/AbstractADeleteQuery.java
@@ -29,15 +29,12 @@ import org.opencastproject.assetmanager.impl.AbstractAssetManager;
 import org.opencastproject.assetmanager.impl.RuntimeTypes;
 import org.opencastproject.assetmanager.impl.VersionImpl;
 import org.opencastproject.assetmanager.impl.persistence.Conversions;
-import org.opencastproject.assetmanager.impl.persistence.Database;
 import org.opencastproject.assetmanager.impl.persistence.EntityPaths;
 import org.opencastproject.assetmanager.impl.persistence.QPropertyDto;
 import org.opencastproject.assetmanager.impl.persistence.QSnapshotDto;
 import org.opencastproject.assetmanager.impl.storage.DeletionSelector;
-import org.opencastproject.util.persistencefn.Queries;
 
 import com.entwinemedia.fn.Fn;
-import com.entwinemedia.fn.Unit;
 import com.entwinemedia.fn.data.SetB;
 import com.mysema.query.Tuple;
 import com.mysema.query.jpa.JPASubQuery;
@@ -53,8 +50,6 @@ import org.slf4j.LoggerFactory;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-
-import javax.persistence.EntityManager;
 
 public abstract class AbstractADeleteQuery implements ADeleteQuery, DeleteQueryContributor, EntityPaths {
   private static final Logger logger = LoggerFactory.getLogger(AbstractADeleteQuery.class);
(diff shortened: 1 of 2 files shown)


staubesv:
> This patch will remove this procedure and instead add a simple,
> performant cleanup procedure for a single media package before the first
> snapshot is being created.

AFAIK, the term "orphaned properties" refers to properties (oc_assets_properties) which refer to a media package that does not exist anymore because no snapshots for this media package exist anymore (properties refer to media packages, not to snapshots).

Doesn't seem that "...before the first snapshot is being created" is the right time to handle properties where no snapshots for a given media package exist anymore...

... How are those orphaned properties handled with this patch?

lkiesow:
The only time that this might become problematic is when these data affect new recordings which might happen if you delete and then re-ingest recordings. Additionally to fixing the performance issues, this patch will assure that there are no leftover properties for new events which are being added.

staubesv:
@lkiesow But how are the orphan properties handled?

Example:
TakeSnapshotOf(MediaPackageA) -> Creates Version 0
TakeSnapshotOf(MediaPackageA) -> Creates Version 1
TakeSnapshotOf(MediaPackageA) -> Creates Version 2
AddPropertiesTo(MediaPackageA)
DeleteSnapshotVersion(0 and 1) of MediaPackageA
DeleteSnapshotVersion(2) of MediaPackageA (no snapshots anymore)

... but now the properties of Media Package A are still there (hence "orphaned"), aren't they?

While this won't cause any harm (except re-ingest MediaPackageA), there are still leftovers in the database.

lkiesow:
Yes, they are still there but that does hurt far less than joining this table every time and you can easily clean up the table anytime you want to. Unless, of course, the delete process handles the properties which e.g. happens in case of completely deleting an event from the admin interface.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java
```diff
@@ -86,6 +87,10 @@ public AssetManagerWithSecurity(
     // Allow this if:
     //  - no previous snapshot exists
     //  - the user has write access to the previous snapshot
+    if (r.getSize() < 1) {
```
gregorydlogan: If the user does not have write permissions we still nuke the old properties?

lkiesow: A non-existent snapshot automatically means that you have write permissions. There are no permissions to check against after all which could forbid you from writing something. Look a few lines down and you will find:

https://github.com/opencast/opencast/blob/c93265c87de283ffcca825c1d5c68648a928f907/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java#L94-L97

gregorydlogan:
Are there plans to address @staubesv's comment long term?  I don't like the idea of accumulating cruft.  Also, conflicts now that I've merged #696.

lkiesow:
Fixed conflict

- Mid-term: I will add a cleanup line to the OC7 database upgrade script
- Long-term: I have an idea for a proper solution. But I will need to see if I can get the time.

JulianKniephoff (approved):
I talked to Lars directly and basically raised the same questions as @staubesv and @gregorydlogan, so I have nothing to add in this review, sadly. :sweat_smile:

Long term we should really rethink the structure of at least this part of the asset manager code, since trying to fix things in the current setup only leads to crazy problems as observed in #650.

smarquard:
deleteOrphanedProperties() is a non-starter for any production system, so the goal for 6.x is to make it perform well and correctly.

Overall, this seems a crazy problem to have, and hopefully we can refactor somehow for a better solution in Opencast 7.

There is also the outstanding issue of not being able to delete an event completely as a non-admin user in 6.x which needs work (but beyond the scope of this PR).

So I support merging this PR as-is because it is part of the fix for critical performance problems (and are also running this change in our 6.x production system).

smarquard (approved):

lkiesow:
To give a perspective on performance: On my test machine, I could usually run any select query on `oc_assets_properties` in under 100ms while having > 8 000 000 properties in the table (for comparison: SWITCH production has ~1.3M). When I tested the clean-up from #718, it took 3 min 12.553 sec to run the query and I found the query used to be faster than the join which was generated by `deleteOrphanedProperties()`.

smarquard:
Merging this because it resolves a blocker issue in 6.x, and there is cleanup code in the 6-to-7 db conversion to remove the orphaned property records.

If there's a cleaner solution, it can be applied to develop (or even a subsequent 6.x PR), but 6.x needs to move forward in a performant state.

PR #696, opened 2019-02-11
MH-13356, Unnecessary Snapshots

lkiesow:
# MH-13356, Introduce native existence check

This patch introduces a native query as a fast check of the existence of
any snapshot for a given media package identifier instead of linking
this check to a sub-query looking for the latest snapshot of a given
media package.

# MH-13356, Don't always take snapshots

This patch makes the asset manager's property util check if it is
actually necessary to create a snapshot to store the properties and does
it only if it is.

*Note that this uses the native database access introduced by pull request #694 *


wsmirnow:
This PR has conflicts with #695

lkiesow:
I know. And I will solve them on this pull request if #695 gets merged before this.
Right now, there is nothing I can do about that.

gregorydlogan (approved):
Looks good, waiting on comments and conflict resolution with #694 and #695

PR #697, opened 2019-02-11
MH-13357 Enable being able to disable 2 confusing Admin UI metadata: "duration" & "created"

karendolan:
These 2 fields throw NullPointer currently, when omitted from  org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg.
But, there is no technical reason why they need to be required in the Event Details Metadata display.

Making them optional allows sites to reduce confusion they can cause to users.  Display of "duration" display cause confusion with actual published and uploaded media duration times and does not correspond to a changed schedule time. The "created' display causes confusion with schedule, event, media package start time, and published event start time.

### DIFF
--- a/etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
+++ b/etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
@@ -85,7 +85,7 @@ property.description.readOnly=false
 property.description.required=false
 property.description.order=2
 
-# A Duration property (Display is optional)
+# A Duration property
 property.duration.inputID=temporal
 property.duration.outputID=duration
 property.duration.label=EVENTS.EVENTS.DETAILS.METADATA.DURATION
@@ -94,7 +94,7 @@ property.duration.readOnly=false
 property.duration.required=false
 property.duration.order=11
 
-# A Start Date property (Display is required)
+# A Start Date property
 property.startDate.inputID=temporal
 property.startDate.outputID=startDate
 property.startDate.label=EVENTS.EVENTS.DETAILS.METADATA.START_DATE
@@ -120,7 +120,7 @@ property.source.readOnly=false
 property.source.required=false
 property.source.order=13
 
-# Created (Display is optional)
+# Created
 property.created.inputID=created
 property.created.label=EVENTS.EVENTS.DETAILS.METADATA.CREATED
 property.created.type=date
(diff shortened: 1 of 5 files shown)


etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
```diff
@@ -85,7 +85,7 @@ property.description.readOnly=false
 property.description.required=false
 property.description.order=2
 
-# A Duration property
+# A Duration property (Display is optional)
```
staubesv: @karendolan Note that the setting "required" is not just about indicating "required" vs "optional" in the UI - both the Admin UI facade and the External API will actually react on this, e.g. you cannot set an empty value for a required field using the Admin UI facade or the External API. Seem from that point of view, the comment "Display is optional" seems not very helpful and somewhat misleading as this depends on the actual value of the field "required"

modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventUtils.java
```diff
@@ -159,7 +159,8 @@ public static MetadataCollection getEventMetadata(Event event, EventCatalogUIAda
       }
     }
 
-    if (event.getDuration() != null) {
+    // The duration event details metadata is an optional display field
```
staubesv: Whether this field is optional or not depends on the metadata configuration...

karendolan: I can remove the comment to avoid confusion.

etc/org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg
karendolan: The intention was to show that the field could be commented out of the cfg file. [EDIT] removed the comment change to avoid confusion.

staubesv: @karendolan Sorry... I did not read the description in detail enough... I thought that those fields would already be optional ;-) 

Nevertheless, thanks for removing the "Display is optional" comments as the metadata configuration has more effect than just "display" in Admin UI, so those comments were indeed somewhat misleading.

staubesv (approved):

PR #700, opened 2019-02-12
MH-13355 Increase the default timeout for TrustedHttpClientImpl 

smarquard:
Inter-server operations like ingest -> admin to start a new workflow on an ingested mediapackage use a POST to start a workflow.

Under some conditions, this can take longer than 60s, in which case the calling node (ingest server) thinks the request has failed, whereas it actually executes fine.

This increases the default timeout for HttpTrustedClient from 1 minute to 5 minutes so it doesn't fail prematurely, which will make Opencast a little more robust and avoid services being flagged as WARN or ERROR unnecessarily.

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/TrustedHttpClientImpl.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/TrustedHttpClientImpl.java
@@ -114,7 +114,7 @@ public class TrustedHttpClientImpl implements TrustedHttpClient, HttpConnectionM
   public static final int DEFAULT_CONNECTION_TIMEOUT = 60 * 1000;
 
   /** The default time between packets that causes a connection to fail */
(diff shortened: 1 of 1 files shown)


CGreweling:
@smarquard  we had also problems with this at switch. I wrote a patch which makes this configurable. Had no time so far to make pr against community version.  In there i changed the DEFAULT_CONNECTION_TIMEOUT. Do you know the difference behavior between this and the DEFAULT_SOCKET_TIMEOUT?

smarquard:
Connection timeout is how long to wait while attempting to open the TCP connection, and the socket timeout is the read response (how long to wait between receiving data, or for the request to return the final response).

I think it's reasonable to leave the connection timeout as-is (e.g. if a worker node was down in such a way that it didn't return connection refused but just no response to tcp connections), then you'd want to move on sooner rather than later. On the other hand once the TCP connection has opened, then you're waiting for the server to return something in response to the GET/POST which is the socket timeout.

PR #702, opened 2019-02-12
MH-13361, Fix Scheduler Item Serialization

lkiesow:
Opencast cannot serialize Long whcih are used in a few scheduler
messages given the right (or wrong) circumstances.  We need to use
native data types (long) or Stings instead.

This patch fixes the sending of some scheduler related messages via
ActiveMQ.


smarquard:
Looks good. Tested in conjunction with #614

PR #705, opened 2019-02-13
WIP: Added missing documentation for the restore from remote WOH

ebbertd:
This PR adds the missing documentation for the restore-from-remote WOH

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/restore-from-remote.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/restore-from-remote.md
@@ -1,0 +1,11 @@
+# RestoreFromRemoteOperationHandler
+
+## Description
+The RestoreFromRemoteOperationHandler can be used to move files from remote storate to local storage.
+
+## Operation Example
+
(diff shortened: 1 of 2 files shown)


ebbertd:
This PR adds the missing documentation for the restore-from-remote WOH. However, I'm not sure if I'm missing some details on how to use it. I only spotted that it was missing.

gregorydlogan:
@ebbertd, now I know why I was confused on list: That workflow operation handler no longer exists.  I unified offloading and restoring to a single workflow operation handler (move-storage).  Are there references to the old one that I've missed?

ebbertd:
@gregorydlogan Yes, you missed the example workflow. https://github.com/opencast/opencast/blob/develop/etc/workflows/restore.xml#L39

gregorydlogan:
@ebbertd Bah, thanks for pointing that out!  I've directly fixed it in develop.  Are there any other references you can find?

ebbertd:
No, couldn't find any more and therefore I will close this.

PR #708, opened 2019-02-14
MH-13366: Add `REFERENCES` permission to standard Opencast `GRANT` statement

JulianKniephoff:
You apparently need `REFERENCES` permissions in MySQL (as opposed to MariaDB), to create foreign key constraints. In MariaDB, this permission is simply ignored.


lkiesow:
Looks good. Works on MariaDB as well. Merged.

PR #709, opened 2019-02-14
MH-13365 inbox ingest into series and inbox retry

doofy:
This patch adds the ability to ingest videos via the inbox directly into a existing series. Furthermore a retry strategy was implemented to make the inbox somewhat more robust to failure.

@reviewer: I suggest to squash all commits of this PR as there is a lot of back and forth with some changes

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -211,8 +211,8 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-incident-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-index-service/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-api/${project.version}</bundle>
-    <bundle start-level="83">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
-    <bundle start-level="83">mvn:org.opencastproject/opencast-ingest-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-ffmpeg/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-workflowoperation/${project.version}</bundle>
@@ -324,7 +324,7 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-incident-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-index-service/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-api/${project.version}</bundle>
-    <bundle start-level="83">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-remote/${project.version}</bundle>
@@ -449,7 +449,7 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-incident-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-index-service/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-api/${project.version}</bundle>
-    <bundle start-level="83">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-remote/${project.version}</bundle>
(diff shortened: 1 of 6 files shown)


assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -211,8 +211,8 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-incident-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-index-service/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-ingest-workflowoperation/${project.version}</bundle>
+    <bundle start-level="83">mvn:org.opencastproject/opencast-ingest-service-impl/${project.version}</bundle>
```
lkiesow: Is there a specific reason to increase the start level?
Usually, that's a sign for broken dependencies

etc/org.opencastproject.ingest.scanner.InboxScannerService-inbox.cfg
```diff
@@ -3,7 +3,7 @@ user.name=admin
 user.organization=mh_default_org
 
 # Specify the workflow definition (by its identifier) to run for media, ingested from the inbox
-workflow.definition=schedule-and-upload
+workflow.definition=fast
```
lkiesow: Do you really want to set the fast testing workflow as default?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
```diff
@@ -289,8 +304,18 @@ public static String getCfg(Dictionary d, String key) throws ConfigurationExcept
     return ps;
   }
 
+  public static String getCfgOrDefault(Dictionary d, String key, String defaultValue) throws ConfigurationException {
```
lkiesow: This is basically a custom re-implementation of Java's built-in [`Objects.toString(Object o, String nullDefault)`](https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#toString(java.lang.Object,%20java.lang.String)) with the only difference that

```java
getCfgOrDefault(properties, WORKFLOW_DEFINITION, null);
```
would be
```java
Objects.toString(properties.get(WORKFLOW_DEFINITION), null);
```
which isn't much of a difference.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
```diff
@@ -305,11 +330,28 @@ public static String getCfg(Dictionary d, String key) throws ConfigurationExcept
...
+
+  public static int getCfgAsIntOrDefault(Dictionary d, String key, int defaultValue) throws ConfigurationException {
+    String valStr = getCfgOrDefault(d, key, null);
+    if (StringUtils.isEmpty(valStr))
+      return defaultValue;
+    try {
+      return Integer.parseInt(valStr);
+    } catch (Exception e) {
+      logger.warn("The inbox configuration {} value {} is not an integer. Use default value of {}", key, valStr,
+              defaultValue);
```
lkiesow: I'm not a huge fan of running with actual errors and would recommend to rather fail fast.
While it might seem nice that things still work even though the configuration is broken, it can lead to unexpected, weird behavior.

Hence, my suggestion here would be to just go with something like:

```java
if (StringUtils.isEmpty(valStr)) {
  return defaultValue;
} else {
  return Integer.parseInt(valStr);
}
```

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
```diff
@@ -68,100 +80,218 @@
...
+                  logger.info("Ingest from inbox into series with id {}", seriesID);
+                } else {
+                  seriesID = null;
+                }
+              }
+
+              /* Add title */
+              DublinCoreCatalog dcc = DublinCores.mkOpencastEpisode().getCatalog();
+              dcc.add(DublinCore.PROPERTY_TITLE, artifact.getName());
+              if (StringUtils.isNotBlank(seriesID))
```
lkiesow: While in practice this should always work and this is nitpicking, you already checked that the id is valid or `null` and this is more than a check against `null` ;-)

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
```diff
@@ -68,100 +80,218 @@
...
+                dcc.add(DublinCore.PROPERTY_IS_PART_OF, seriesID);
+              logger.debug("episode dublincore for the inbox file {}: {}", artifact.getName(), dcc.toXml());
```
lkiesow: You always serialize this as XML regardless of the log level. you can check if the log level is debug using the logger.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
doofy: I debated this a lot and came to the conclusion that I wont touch this

assemblies/karaf-features/src/main/feature/feature.xml
doofy: I think this was a try to reduce inbox unreliability.

doofy: I reverted this change

etc/org.opencastproject.ingest.scanner.InboxScannerService-inbox.cfg
doofy: I reverted this change

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
doofy: I implemented this change

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
doofy: I implemented this change

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
doofy: I implemented this change

doofy:
I worked through all of your comments @lkiesow although I did not adapt all suggestions.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
gregorydlogan: While I agree that our codebase perhaps has a little bloat (c.f OsgiUtil.getOptCfg, which does basically the same thing here without the default), why add a bunch of boilerplate to this relatively simple built-in?  I'm fine adding boilerplate, but I feel like in this case the extra code doesn't actually add anything.

gregorydlogan:
My understanding is that Michael is still working on this.  I'm going to be away until the 20th of March, so if someone wants to merge this prior to my return feel free.

gregorydlogan:
Reminder: This is a feature which means it needs to be done soon to get into 7.x!

lkiesow:
@gregorydlogan, I fixed the remaining issues since @doofy is still on vacation.
Should hopefully be ready to go now.

PR #710, opened 2019-02-14
adds an WOH, which can add catalogs to the MediaPackage of an workflow instance

darolfes:
This WorkflowOperationHandler can add a catalog to the MediaPackage.
It supports the following workflow configuration keys:
- `catalog-path`: The path to the catalog, which will be added.
- `catalog-flavor`: The flavor of the catalog, also used to identify catalogs of the same type.
- `catalog-name`: Name of the catalog in the workspace.
- `catalog-tags`: List of comma seperated catalog tags.
- `catalog-type-collision-behavior` the action to perform, if an catalog of the same flavor already exists. Three options are supported: `skip` the adding of the catalog, `fail` the workflow operation or `keep` the new catalog, resulting in two or more catalogs of the same type coexisting.

### DIFF
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
@@ -66,7 +66,7 @@ public class AddCatalogWorkflowOperationHandler extends AbstractWorkflowOperatio
 
   /** The logging facility */
   private static final Logger logger = LoggerFactory
-    .getLogger(ProbeResolutionWorkflowOperationHandler.class);
+    .getLogger(AddCatalogWorkflowOperationHandler.class);
 
   /** config key for the catalog name */
   private static final String CFG_KEY_CATALOG_NAME   = "catalog-name";
@@ -215,17 +215,13 @@ public class AddCatalogWorkflowOperationHandler extends AbstractWorkflowOperatio
    */
   private CatalogTypeCollisionBehavior parseCollisionBehavior(String rawBehavior)
       throws WorkflowOperationException {
-    switch (rawBehavior) {
-      case "skip":
-        return CatalogTypeCollisionBehavior.SKIP;
-      case "keep":
-        return CatalogTypeCollisionBehavior.KEEP;
-      case "fail":
-        return CatalogTypeCollisionBehavior.FAIL;
-      default:
-        throw new WorkflowOperationException("Workflowoperation configured incorrectly, the configuration '"
-                                             + CFG_KEY_CATALOG_TYPE_COLLISION_BEHAVIOR
-                                             + "' only accepts 'skip', 'keep', 'fail'");
+    try {
+      return CatalogTypeCollisionBehavior.valueOf(rawBehavior.toUpperCase());
+    }
+    catch (IllegalArgumentException e) {
+      throw new WorkflowOperationException("Workflowoperation configured incorrectly, the configuration '"
+                                           + CFG_KEY_CATALOG_TYPE_COLLISION_BEHAVIOR
+                                           + "' only accepts 'skip', 'keep', 'fail'");
     }
   }
 }
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
@@ -43,7 +43,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.Arrays;
-import java.util.List;
 import java.util.UUID;
 
 /**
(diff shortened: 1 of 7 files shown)


staubesv:
@darolfes Not expecting you to implement this, but it would be nice to generalize this WOH to add just any type of media package element. Just one more configuration key "type" (track|catalog|attachment) and some additional logic that adds track|catalog|attachment instead of just catalogs (they all have a path, flavor, name, tags and some potential for collisions)... just mentioning in case you had a lot of time ;-)

... and just wondering: What is the use case for adding a catalog in a workflow?

lkiesow:
> ... and just wondering: What is the use case for adding a catalog in a workflow?

Setting a default extended metadata catalog for ingests from unreliable sources (like the inbox)

lkiesow:
@staubesv 

> it would be nice to generalize this WOH to add just any type of media package element.

We've discussed that today and added it to our todo window (←yes, it's a real window) but it will probably not happen for OC7 since we want to get a few other things ready in time.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+
+  /**
+   * Checks whether the catalogFlavor exists in the array of catalogs
+   *
+   * @param catalogFlavor
+   * @param catalogs
+   * @return true, if the catalogFlavor exists in the array of catalogs, else false
+   */
+  private boolean doesCatalogFlavorExist(MediaPackageElementFlavor catalogFlavor, Catalog[] catalogs) {
+    List<Catalog> ctls = Arrays.asList(catalogs);
```
staubesv: In theory, you could write just...

`
return Arrays.asList(catalogs).stream().anyMatch(catalog -> catalogFlavor.matches(catalog.getFlavor()));
`

here.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+   *          used to get the value of the configuration
+   * @param cfgKey
+   *          specifies the value to get
+   * @param isOptional
+   *          is the configuration value allowed to be not set
+   * @return the value of the configuration key or null if isOptional is set and the configuration value is not set
+   * @throws WorkflowOperationException
+   *           if isOptional is false and the configuration value is not set
+   */
+  private String getCfgValueFromOp(WorkflowInstance inst, String cfgKey, boolean isOptional)
```
staubesv: Hmm.... this kind of code should go to WorkflowInstance or some Workflow utility class, shouldn't it? Also, have you checked that WorkflowInstance does not already offer similar methods?

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+
+  /**
+   * Parses the rawFlavor String to an MediaPackageElementFlavor
+   * or throws an WorkflowOperationException if the parsing failed
+   *
+   * @param rawFlavor
+   * @return
+   * @throws WorkflowOperationException
+   */
+  private MediaPackageElementFlavor parseFlavor(String rawFlavor)
```
staubesv: This methods seems somewhat overkill/superfluous to me: Can't you just use a try-catch block wherever you need to parse the flavor?

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+      throws WorkflowOperationException {
+    try {
+      return MediaPackageElementFlavor.parseFlavor(rawFlavor);
+    }
+    catch (IllegalArgumentException e) {
+      throw new WorkflowOperationException("Flavor needs to be an correct Flavor MimeType");
```
staubesv: There is no such thing as a "Flavor MimeType"

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+
+  /** The logging facility */
+  private static final Logger logger = LoggerFactory
+    .getLogger(AddCatalogWorkflowOperationHandler.class);
+
+  /** config key for the catalog name */
+  private static final String CFG_KEY_CATALOG_NAME   = "catalog-name";
+  /** config key for the catalog flavor */
+  private static final String CFG_KEY_CATALOG_FLAVOR = "catalog-flavor";
+  /** config key, which locates the catalog on the filesystem */
```
staubesv: no comma here

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
```diff
@@ -0,0 +1,227 @@
...
+  private static final String CFG_KEY_CATALOG_PATH   = "catalog-path";
+  /** config key for the catalog tags */
+  private static final String CFG_KEY_CATALOG_TAGS   = "catalog-tags";
+  /** config key, which defines the behavior if a catalog of the same flavor already exists */
```
staubesv: no comma here

staubesv (changes requested):
@darolfes Just skimmed over the code and found some minor issues.

Most importantly, though: Please do add a documentation for this new WOH (see https://docs.opencast.org/develop/admin/workflowoperationhandlers/).

Will continue the review once a documentation is available.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AddCatalogWorkflowOperationHandler.java
darolfes: fixed it, the AbstractWorkflowOperationHandler provides similar methods

darolfes:
I added the documentation and fixed the issues

staubesv (approved):

PR #711, opened 2019-02-15
MH-13367 External API series acl returns null pointer with missing acl

CGreweling:
External API series acl returns Server Error null pointer with missing acl.
This patch returns http not found.

This is a problem when a series is created. And directly after this accessed before
creation is completed.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
@@ -722,12 +722,11 @@ public class SeriesEndpoint {
       // The ACL is stored as JSON string in the index. Parse it and extract the part we want to have in the API.
       JSONObject acl = (JSONObject) parser.parse(series.getAccessPolicy());
 
-      if (((JSONObject) acl.get("acl")).containsKey("ace")) {
-        return ApiResponses.Json.ok(requestedVersion, ((JSONArray) ((JSONObject) acl.get("acl")).get("ace")).toJSONString());
-      } else if (!((JSONObject) acl.get("acl")).containsKey("ace")) {
+      if (!((JSONObject) acl.get("acl")).containsKey("ace")) {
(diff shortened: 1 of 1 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -721,6 +721,12 @@ public Response getSeriesAcl(@HeaderParam("Accept") String acceptHeader, @PathPa
     for (final Series series : indexService.getSeries(id, externalIndex)) {
       // The ACL is stored as JSON string in the index. Parse it and extract the part we want to have in the API.
       JSONObject acl = (JSONObject) parser.parse(series.getAccessPolicy());
+
+      if (((JSONObject) acl.get("acl")).containsKey("ace")) {
+        return ApiResponses.Json.ok(requestedVersion, ((JSONArray) ((JSONObject) acl.get("acl")).get("ace")).toJSONString());
+      } else if (!((JSONObject) acl.get("acl")).containsKey("ace")) {
```
smarquard: This is a bit confusing to read. Why put a check for the negation of the condition that's already checked in the if above (unless I'm misreading this). Seems you could just do:

```
if (!((JSONObject) acl.get("acl")).containsKey("ace")) {
        return ApiResponses.notFound("Cannot find acl for series with id '%s'.", id);
}
```

Is there a possibility of an empty acl, that is it exists but has no ace elements?

CGreweling: Yes, ok you are right, this easier to read. And brings not that much benefits with it. I will change it.

Yep Unfortunately yes. 
If you create a series with the external api, which is async.
And directly try to acces this series after creating without a time out.
So the Series is created, but the acl is not at that time. It comes some seconds later in.
I think this mainly happens in installations with lots of data.

Or if you have a Series with all acls being deleted. This should not be the case, but sometimes... :(.

PR #713, opened 2019-02-15
MH-13294 Workflow for track replacement and cleanup Snapshots

CGreweling:
This Workflow replaces the delivery presenter video with a local stored video. And removes every trimmed and source flavours from the mediapackage. The metadata is kept. 
All older Snapshots are deleted from the asset manager.
 
This is implemented to replace a failed recording, when for example the video data is corrupt,
with a place holder video and its metadata. To save some space the original not needed data is deleted.


etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>failedrecording</id>
+  <title>Recording failed</title>
+  <tags>
+  <tag>editor</tag>
+  <tag>archive</tag>
+  </tags>
+  <description/>
```
lkiesow: Can you add a description so that people understand for what this workflow is meant to be used?

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,100 @@
...
+    </operation>
+
+    <!-- Copy placeholder -failed- Video for publication -->
+    <operation
+      id="compose"
+      fail-on-error="true"
+      exception-handler-workflow="error"
+      description="copy failed Video">
+      <configurations>
+        <configuration key="source-flavor">*/source</configuration>
```
lkiesow: If there are `presenter/source` and `presentation/source`, wouldn't this run twice and create two identical `presenter/delivery` tracks?

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,100 @@
...
+
+    <!-- Archive the current state of the media package -->
+    <operation
+      id="snapshot"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error"
+      description="Archiving">
+      <configurations>
+         <configuration key="source-flavors">dublincore/*,security/*,*/delivery</configuration>
+         <configuration key="source-tags">archive</configuration>
```
lkiesow:
```suggestion
         <configuration key="source-tags">failed</configuration>
```

Using the `failed` tag here would mean that you could remove the `tag` operation from above since you would not need to remove any tags.

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,100 @@
...
+      id="asset-delete"
+      fail-on-error="true"
+      exception-handler-workflow="error"
+      description="Delete every older Snapshot from AssetManager">
+      <configurations>
+        <configuration key="keep-last-snapshot">true</configuration>
+      </configurations>
+    </operation>
+
+  </operations>
```
lkiesow: We should run cleanup as the last operation to not leave all artifacts in the workspace and working file repository.

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,100 @@
...
+      <configurations>
+        <configuration key="download-source-flavors">dublincore/*,security/*,*/delivery</configuration>
+        <configuration key="download-source-tags">engage-download</configuration>
+        <configuration key="check-availability">false</configuration>
+      </configurations>
+    </operation>
+
+    <!-- Copy last mediapackage version to workspace -->
+    <operation
+      id="ingest-download"
```
lkiesow: Why do we need to copy all files into the local workspace?

etc/workflows/failedRecording.xml
CGreweling: Cleanup is enforced anyway.

etc/workflows/failedRecording.xml
CGreweling: Otherwise cleanup will throw a lot of error, because of not found links to the archive referenced in the workflow mp.

etc/workflows/failedRecording.xml
CGreweling: I don't get that. Maybe you can explain this. You need to tag the items you want to keep in the last snapshot. And untag the Items you want to delete.

etc/workflows/failedRecording.xml
CGreweling: No, but i can change it for better understanding.

etc/workflows/failedRecording.xml
CGreweling: Okay, after our discussion. Here is the explanation. Maybe it could be done in a different way, but to keep it in  the most "common" state as other workflows do. I will stay witch the archive tag.

etc/workflows/failedRecording.xml
CGreweling: Done, good point.

etc/encoding/failed.properties
```diff
@@ -0,0 +1,6 @@
+# copy video into mp workspace
+profile.failed.http.name = Failed video
+profile.failed.http.input = visual
+profile.failed.http.output = visual
+profile.failed.http.suffix = -failed.mp4
+profile.failed.http.ffmpeg.command = -i /tmp/failed1.mp4 -c copy #{out.dir}/#{out.name}#{out.suffix}
```
staubesv: Hmm... While I'm not a big fan of having binary files in a GIT repo, this profile obviously will cause the workflow to fail which seems very undesirable.

Can you please somehow make this profile and the workflow that uses it work out-of-the-box?

Also, please add somewhat more description to the profile - it does have a purpose more than "copy video into mp workspace", doesn't it?

Note that you also could use an image and transform it into a video using WOH image-to-video.

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>failedrecording</id>
+  <title>Recording failed</title>
+  <tags>
+  <tag>editor</tag>
```
staubesv: Indentation

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
...
+    <!-- Untag elements for deletion -->
+    <operation
+      id="tag"
+      max-attempts="2"
+      fail-on-error="true"
+      exception-handler-workflow="error"
+      description="Tagging media package elements for deletion">
+      <configurations>
+        <configuration key="source-flavors">*/trimmed,*/source</configuration>
+        <configuration key="target-tags">-atom,-archive</configuration>
```
staubesv: -archive should be enough here.

Also: Does */trimmed really ever get the tag "archive"? It should not.

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
...
+      </configurations>
+    </operation>
+
+    <!-- Copy placeholder -failed- Video for publication -->
+    <operation
```
staubesv: Once you convinced @darolfes to generalize his WOH (see https://github.com/opencast/opencast/pull/710), you can just add the video without mis-using an encoding profile to copy files ;-)

Note that it would be an option to use WOH image-to-video in combination with an failed.jpg instead of failed.mp4... but just mentioning this.

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
...
+    </operation>
+
+    <!-- Archive the current state of the media package -->
+    <operation
+      id="snapshot"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error"
+      description="Archiving">
+      <configurations>
+         <configuration key="source-flavors">dublincore/*,security/*,*/delivery</configuration>
```
staubesv: Are you really sure that you want to archive */delivery?

etc/workflows/failedRecording.xml
staubesv: Really? Why are any workflows then using it at all? Please explain...

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>failedrecording</id>
+  <title>Recording failed</title>
```
staubesv: Imagine a user seeing "Recording failed" in the workflow selection dropdown. Will she or he understand what this is supposed to do? 

Maybe something more descriptive like "Publish failed placeholder"...

etc/workflows/failedRecording.xml
```diff
@@ -0,0 +1,103 @@
...
+    <!-- Copy placeholder -failed- Video for publication -->
+    <operation
+      id="compose"
+      fail-on-error="true"
+      exception-handler-workflow="error"
+      description="copy failed Video">
+      <configurations>
+        <configuration key="source-flavor">presenter/source</configuration>
+        <configuration key="target-flavor">presenter/delivery</configuration>
+        <configuration key="target-tags">failed,+archive,+engage-download</configuration>
```
staubesv: Should be +failed

staubesv (changes requested):
@CGreweling I've found a number of minor issues.

CGreweling:
May be i was a little too eager, to push this to the community. If this confuses you at some points, i think this should not got to an "enduser". 
The main Point is here the placeholdervideo you have to put somewhere on the disk.
I Think i should decline this PR an do another one when i have an easier way for the video.
@staubesv  @lkiesow  are you okay with this?

staubesv:
@CGreweling & @lkiesow I think the workflow is useful. But it really should work out-of-the-box if it is offered in the workflow selection. Alternatively, you could remove the tags "archive" and "editor" (to hide the workflow from being used) and add a comment to the workflow definition (requires encoding profile XYZ to be adjusted).

etc/workflows/failedRecording.xml
CGreweling: done

etc/workflows/failedRecording.xml
CGreweling: So, i don't know how but before my rebase it was enforced i do not know how... 
Now it is not anymore so i added the cleanup.

CGreweling:
So i removed the editor/archive tag. And resolved your comments.
Now it is up to you to decide.
I do not like the idea of having binary files like pictures or videos kept in git for this. 
And i think every institution would do his own branded placeholder video.

For a standard workflow maybe we can generate something in the woh in the future?

staubesv (changes requested):
Tested and found some minor issues:

- The filename failedRecording.xml does not meet the unwritten convention of workflow filenames in etc/workflows (no camel case here). Maybe something like "publish-failed-recording.xml" and a corresponding workflow definition id like "publish-failed-recording"

- The workflow does not unpublish existing publications. Use case: A recording has been published successfully but the actual content is unusable (no audio, signal problems,...). In that case, you might also want to use this workflow to publish a placeholder. This does not work, however

Can you please address those two comments?

CGreweling:
@staubesv rewrote this to a more default wf and default encoding.
I hope i got everything is needed for this.
The mediapackaged is retracted from the standard channels, which are enabled for publication by default.
All source material is deleted.
Dublicore Series/Episode and Acls will not be deleted.
Now a 10sec video is generated with text Sorry video not available.
The reusulting mediapackage will be published to the channels, it was published before.

mliradelc:
Hi, just a reminder, next Monday (April 1st) will be the feature freeze of Opencast 7.0 😉

staubesv (approved):

PR #714, opened 2019-02-15
MH-12533, Re-introduce ability to avoid data loss during ingest

karendolan:
This pull re-introduces the ability for sites preserve metadata edits to scheduled events made from the Opencast Admin UI when using a capture agent that ingests with potentially stale material, such as series and event catalogs.

Tracks from the ingest always given priority from the CA. But, attachments and catalogs from the CA have a lower precedence than existing ones in the Opencast system.

This config should go into the docs, somewhere, to describe the various configs, but I'm having a hard time finding a spot. I might make an Ingest Service page.

### DIFF
--- a/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
+++ b/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
@@ -1,28 +1,37 @@
 #This is the config file for the IngestServiceImpl
 
-# This is an option to update Series metadata during ingest
-# The default is true. This will updated an existing series with catalog found in the ingest
-# false means the exiting series catalog will not be updated using new series metadata found in the ingest
-# In both cases, the series catalog will be created if it does not already exist in Opencast
-# In both cases, the catalog for the episode will be created in Opencast
-
-# TODO: consider moving this key to the system.properties if used by other services, such as mediapackage update
-
-org.opencastproject.series.overwrite=false
-
-# Control if catalogs sent by capture agents for scheduled events are skipped. Not skipping them means that they will
+# Control if all catalogs sent by capture agents for scheduled events are skipped. Not skipping them means that they will
 # potentially overwrite existing metadata catalogs in Opencast.
 #
-# Default: true
+# Default: false
 #
-#skip.catalogs.for.existing.events=true
+#skip.catalogs.for.existing.events=false
 
-# Control if attachments sent by capture agents for scheduled events are skipped. Not skipping them means that they will
+# Control if all attachments sent by capture agents for scheduled events are skipped. Not skipping them means that they will
 # potentially overwrite existing attachments in Opencast.
 #
+# Default: false
+#
+#skip.attachments.for.existing.events=false
+
+# Control if some attachments and catalogs sent by capture agents for scheduled are added only if they do not overwrite
+# existing attachments and catalogs of the same flavor in Opencast. Not restricting to only new catalogs and attachments means
+# that ingested ones with flavors that match existing catalogs and attachments in Opencast will be overwritten.
+#
+# Note: if the skip catalogs and skip attachments parameters, above, are set to true, new catalogs and attachments are also skipped.
+#
 # Default: true
 #
-#skip.attachments.for.existing.events=true
+#add.only.new.catalogs.attachments.for.existing.events=true
+
+# *Deprecated*
+# Control to allow an Opencast series to be modified if an event is ingested. Enabling this feature is not recommended.
+# Enabling series overwrite during event ingest will potentially start an event update chain reaction affecting all events associated
+# to that series in Opencast. Every time an event is ingested.
+#
+# Default is false
+# *Deprecated*
+#org.opencastproject.series.overwrite=false
 
 # The approximate load placed on the system by ingesting a file
 # Since these jobs are *not* dispatched there is no current way to limit the simultaneous number of ingests, but these jobs will block further jobs from running on an already busy admin node
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -156,9 +156,6 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
 
   public static final String JOB_TYPE = "org.opencastproject.ingest";
 
-  /** Managed Property key to overwrite existing series */
-  public static final String PROPKEY_OVERWRITE_SERIES = "org.opencastproject.series.overwrite";
-
   /** Methods that ingest zips create jobs with this operation type */
   public static final String INGEST_ZIP = "zip";
 
@@ -202,13 +199,31 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
   public static final String DOWNLOAD_PASSWORD = "org.opencastproject.download.password";
 
   /** By default, do not allow event ingest to modify existing series metadata */
-  static final boolean DEFAULT_ALLOW_SERIES_MODIFICATIONS = false;
+  public static final boolean DEFAULT_ALLOW_SERIES_MODIFICATIONS = false;
+
+  /** The default is to preserve existing Opencast flavors during ingest. */
+  public static final boolean DEFAULT_ALLOW_ONLY_NEW_FLAVORS = true;
+
+  /** The default is not to automatically skip attachments and catalogs from capture agent */
+  public static final boolean DEFAULT_SKIP = false;
+
+  /** Managed Property key to allow Opencast series modification during ingest
+   * Deprecated, the param potentially causes an update chain reaction for all
+   * events associated to that series, for each ingest */
+  @Deprecated
+  public static final String MODIFY_OPENCAST_SERIES_KEY = "org.opencastproject.series.overwrite";
+
+  /** Managed Property key to allow new flavors of ingested attachments and catalogs
+   * to be added to the existing Opencast mediapackage. But, not catalogs and attachments
+   * that would overwrite existing ones in Opencast.
+   */
+  public static final String ADD_ONLY_NEW_FLAVORS_KEY = "add.only.new.catalogs.attachments.for.existing.events";
 
   /** Control if catalogs sent by capture agents for scheduled events are skipped. */
-  private static final String SKIP_CATALOGS_KEY = "skip.catalogs.for.existing.events";
+  public static final String SKIP_CATALOGS_KEY = "skip.catalogs.for.existing.events";
 
   /** Control if attachments sent by capture agents for scheduled events are skipped. */
-  private static final String SKIP_ATTACHMENTS_KEY = "skip.attachments.for.existing.events";
+  public static final String SKIP_ATTACHMENTS_KEY = "skip.attachments.for.existing.events";
 
   /** The approximate load placed on the system by ingesting a file */
   private float ingestFileJobLoad = DEFAULT_INGEST_FILE_JOB_LOAD;
@@ -274,11 +289,13 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
   private Cache<String, Long> partialTrackStartTimes = CacheBuilder.newBuilder().expireAfterWrite(1, TimeUnit.DAYS)
           .build();
 
-  /** Option, if an event ingest may modify metadata from existing series */
-  private boolean allowSeriesModifications = DEFAULT_ALLOW_SERIES_MODIFICATIONS;
+  /** Option to overwrite matching flavors (e.g. series and episode metadata) on ingest,
+   *  tracks are always taken on ingest */
+  protected boolean isAddOnlyNew = DEFAULT_ALLOW_ONLY_NEW_FLAVORS;
+  protected boolean isAllowModifySeries = DEFAULT_ALLOW_SERIES_MODIFICATIONS;
 
-  private boolean skipCatalogs = true;
-  private boolean skipAttachments = true;
+  private boolean skipCatalogs = DEFAULT_SKIP;
+  private boolean skipAttachments = DEFAULT_SKIP;
 
   /**
    * Creates a new ingest service instance.
@@ -319,6 +336,7 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
    */
   @Override
   public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+
     if (properties == null) {
       logger.info("No configuration available, using defaults");
       return;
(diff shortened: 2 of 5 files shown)


smarquard:
I'm just wondering how stale the CA's metadata might be? A CA would generally poll for calendar updates and get event updates when they're changed server-side (including metadata changes), so it seems the window of "staleness" is quite short, for the period when the CA is recording?

Or are there CAs that don't update metadata when it changes in the calendar? Or is this a scenario when you perhaps re-ingest something a while after the recording but have changed the metadata in the meantime on the admin ui?

karendolan:
Hi @smarquard, yes, our data loss concern is mostly for the situation when the recording has stopped but the ingest has not yet happened. We were staggering ingests at one point and had a hard time with this issue. 

But, also, in the past, we've used CAs that read and re-write series and episode catalogs with its own data model and augments custom variations to the catalog from Opencast. We don't want to preclude using CA's that might have that issue. This config option allows us to safely avoid the danger of data loss to catalogs from the CA while still getting tracks and any new items the CA can provide.

etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
```diff
@@ -6,9 +6,14 @@
...
-# TODO: consider moving this key to the system.properties if used by other services, such as mediapackage update
+# MH-12533 Option to prevent ingested flavors (except for tracks) to be overritten during ingest.
+# If true (default), ingested elements that match existing flavors take precedence and overwrite the existing.
+#    For example, metadata changes made from a capture agent UI will override metadata changes made from the Admin UI prior to the ingest.
+# If false, tracks and only new element flavors are accepted from the ingested mp. Existing flavors, like episode and series catalogs,
+#    are not overwritten and retain all updates made from the Admin UI.
+#    Changed the default from true to false to allow metadata edits from the Admin UI to take precedence.
 
-org.opencastproject.series.overwrite=true
+org.opencastproject.ingest.overwrite=true
```
lkiesow: Can you use the following form:

```
# Some description
# Default: xy
#some.key=default.value
```

That way you can easily spot what the default is and you know right away that you did not modify the default since it's still commented out. This is used in a lot of other configuration files as well.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -266,11 +267,11 @@
   private Cache<String, Long> partialTrackStartTimes = CacheBuilder.newBuilder().expireAfterWrite(1, TimeUnit.DAYS)
           .build();
 
-  /** The default is to overwrite series catalog on ingest */
-  protected boolean defaultIsOverWriteSeries = true;
+  /** The OC default is to overwrite existing flavors on ingest. The default is used if the configuration param is not found */
+  protected boolean defaultOverWriteValue = true;
```
lkiesow: I'm wondering if `false` wouldn't be a better default since my guess would be that it is more common for users to modify data on the admin interface than on the capture agents since most capture agents (SMP, NCast, pyCA, ARec) by default do not even give you the option to modify this data.

lkiesow: And you could make this `final` since it's basically a constant isn't it?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -325,14 +326,22 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
     ingestZipJobLoad = LoadUtil.getConfiguredLoadValue(properties, ZIP_JOB_LOAD_KEY, DEFAULT_INGEST_ZIP_JOB_LOAD,
             serviceRegistry);
     // try to get overwrite series option from config, use default if not configured
+    isOverwrite = defaultOverWriteValue;
     try {
-      isOverwriteSeries = Boolean.parseBoolean(((String) properties.get(PROPKEY_OVERWRITE_SERIES)).trim());
+      String overwriteParam = (String) properties.get(PROPKEY_OVERWRITE);
+      if (overwriteParam != null) {
+        isOverwrite = BooleanUtils.toBoolean(overwriteParam.trim().toLowerCase(), "true", "false");
```
lkiesow: Most services just go with the default which afaik also make the lowercase, … unnecessary:
```suggestion
        isOverwrite = BooleanUtils.toBoolean(overwriteParam);
```

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -325,14 +326,22 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
+      } else {
+        logger.info("Did not find '{}' in configuration, using default value '{}'", PROPKEY_OVERWRITE, isOverwrite);
```
lkiesow: I wouldn't log this since you log the value anyway.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -325,14 +326,22 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
+      }
     } catch (Exception e) {
-      isOverwriteSeries = defaultIsOverWriteSeries;
-      logger.warn("Unable to update configuration. {}", e.getMessage());
+      isOverwrite = defaultOverWriteValue;
+      logger.warn("Unable to use '{}' from Ingest configuration file, expected 'true' or 'false'. {}",
+              PROPKEY_OVERWRITE,
+              e.getMessage());
```
lkiesow: I would probably log an error and definitely return a proper stack trace instead of just the message: Especially since you are catching just all kinds of errors

```java
logger.warn("Unable to use '{}' from Ingest configuration file, expected 'true' or 'false'.", PROPKEY_OVERWRITE, e);
```

lkiesow: Thinking about this again: Fail early and don't accept broken configurations: Just don't catch the error. Having a broken configuration may lead to all kinds of weird behavior and having people realize and fix that right away may be the best solution.

lkiesow (dismissed):
Just a few minor things

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: Using 'false' as default would be good for us. However, currently, in OC v2x-v7x, ingesting catalogs overwrite the existing catalogs. The 'true' default, and the current 'true' behavior, works well for Galicaster capture agents. Galicaster  has the use case that event titles, and other event metadata, are adjusted on the CA UI, in the classroom, at the point of capture. I am concerned that flipping the default behavior from true to false will cause metadata data loss to Galicaster adopters until they adjust their config.

etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
karendolan: @lkiesow thank you for this suggestion. It is changed to this format.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: I agree! done.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1349,44 +1351,99 @@ private MediaPackage mergeScheduledMediaPackage(MediaPackage mp) throws IngestEx
...
+        logger.debug("Ignoring {}, not adding to ingested mediapackage {}", MediaPackageElement.Type.Publication, mp);
         continue;
+      } else if (mp.getElementsByFlavor(element.getFlavor()).length > 0) {
+        // The default is to overwrite matching flavored elements in the Asset managed mediapackage (e.g. catalogs)
+        // If isOverwrite is true, changes made from the CA overwrite (update/revert) changes made from the Admin UI.
+        // If isOverwrite is false, changes made from the CA do not overwrite (update/revert) changes made from the Admin UI.
+        // regardless of overwrite, always keep new ingested tracks.
+        if (isOverwrite || MediaPackageElement.Type.Track.equals(element.getElementType())) {
+          // Allow updates made from the Capture Agent to overwrite existing metadata in Opencast
+          logger.debug(
```
gregorydlogan: While I understand the logic here, I can see people being confused by messages about omitting existing elements when it's talking about tracks.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1349,44 +1351,99 @@ private MediaPackage mergeScheduledMediaPackage(MediaPackage mp) throws IngestEx
...
 
+  /**
+   *
+   * The previous OC behaviour is for metadata in the ingested mediapackage to be updated by the
+   * Asset Managed metadata *only* when the field is blank on the ingested mediapackage.
+   * However, that field may have been intentionally emptied by
+   * removing its value from the Capture Agent UI (e.g. Galicaster)
+   *
+   * If isOverwrite is true, metadata values in the ingest mediapackage take precedence and are *not* replaced Asset Managed metadata fields.
+   * If isOverwrite is false, the ingest mediapackage metadata is updated with Asset Managed mediapackage metadata (i.e. Asset Manged mp data prevails, and can empty out values)
```
gregorydlogan: `and can empty out value`.  Can't both cases lead to that?  If I empty out a field on the CA, and isOverwrite is true, then that field *should* be empty at the end of the day.  Likewise, if I populate a field on the CA, and isOverwrite is false, then that fields should also end up empty.  I'm fine with the logic here, but the comment clouds things.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
JamesUoM: I vote for keeping the existing behaviour as the default. (Edit, I should read the code more carefully)

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: Hi @gregorydlogan, actually tracks are always taken from the ingest, it's all the other potential attachments and catalogs that come in on the ingest that are being addressed here, not the tracks.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: Thanks for that observation, @gregorydlogan , I'll fix the comment.

karendolan:
Hi, for anyone looking at this pull, I'm starting to wonder if the community would prefer to continue separating the series catalog overwrite preference with preference to prevent overwrite of all catalogs and attachments (i.e. episode catalog, and potentially, ACL files). Right now, series can be flagged not be overwritten, but the event catalog can be overwritten. This pull changes the existing series overwrite flag into a general all-or-nothing overwrite (except tracks that are always taken)

mliradelc:
Hi, just a reminder, next monday (April 1st) is the feature freeze for Opencast 7.0 😉

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
JamesUoM: @karendolan do you have time to fix the comment and resolve the merge conflicts today?

karendolan: @JamesUoM I have 30 min right now...

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: @JamesUoM I realize that overwrite value will change from default true to default false pretty soon. Not sure if it's worth changing now or wait for the merge back from @lkiesow 's other overwrite pull.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: removed log

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: I fixed comments to clarify

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
karendolan: I made the default constants final

JamesUoM:
I'm changing this to an enhancement for consistency with  https://github.com/opencast/opencast/pull/809

lkiesow:
@karendolan, only now I realize that you are working on the ingest as well. I totally forgot what this pull request is about. Sorry! (Now your question on #809 makes more sense :-)

How do we want to proceed? Shall I merge our two pull requests so we have four options?
The options would then be:

- merge, scheduler media package elements take precedence
- merge, ingest media package elements take precedence
- drop attachments
- drop catalogs

JamesUoM:
> @karendolan, only now I realize that you are working on the ingest as well. I totally forgot what this pull request is about. Sorry! (Now your question on #809 makes more sense :-)

Looks like @gregorydlogan has already merged #809. 

I would like to have the option to add additional catalogs and attachments from the CA not just a  blanket skip all. @lkiesow, can we "fix" your code with  @karendolan's post feature freeze as I'm not sure there is time to resolve this this morning.

karendolan:
@lkiesow Your pull was a bug fix. This pull was designated as a "feature", by you, to optionally prevent ingest from overwriting the series catalog and other catalogs. IMHO, there should not be third feature introduced here to optionally omit new catalogs and attachments from the ingested mp.

karendolan:
I'm in progress of repointing this pull to r/7.x...

karendolan:
@JamesUoM @lkiesow [EDIT][EDIT] Ok, I'll update this pull to reflect the 4 options of

skip.catalogs.for.existing.events //Default true
skip.attachments.for.existing.events //Default true
merge.without.overwrite.for.existing.events // Default true, False is merge with overwrite

The skip catalogs & skip attachments takes precedence over the merge. In other words, if skip.catalogs is true, any new [flavors of] catalogs, from the ingest, are not added to the existing ones.

If skip catalogs is true, and merge without overwrite is false, the existing catalog is not wiped out but is preserved in the updated mp. This could be argued either way. But, in practice, it makes sense for a site to want a new inbound ingest episode catalog to wipe out the existing, but if the ingest doesn't contain a catalog,  to keep the original in the updated mp.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1376,45 +1406,80 @@ private void mergeMediaPackageElements(final MediaPackage mp, final MediaPackage
     }
 
     for (MediaPackageElement element : scheduledMp.getElements()) {
-      // Asset manager media package may have a publication element (for live) if retract live has not run yet
-      if (element.getFlavor() != null
-              && !MediaPackageElement.Type.Publication.equals(element.getElementType())
```
karendolan, line 1381: TODO: check if it's really possible for a mediapackage element to not have a flavor.

karendolan:
@gregorydlogan I'm adding the bug label to be compatible with the companion pull and that this fixes a bug in that pull.

KatrinIhler:
@karendolan Is this ready for review?

KatrinIhler:
@karendolan If this still needs some work, it might be better to close this for now and refile later.

karendolan:
> @karendolan If this still needs some work, it might be better to close this for now and refile later.

Hi @KatrinIhler, I missed your early comment. I believe this pull can be reviewed.

PR #384, opened 2018-08-16
MH-12690: Add i18n support for capture agent statuses

jaysuz:
This will ensure that capture agent statuses are translated in every place they appear (in the main table, in the filter list and in the details modal). For the purpose of showcasing, throw-away German translations are supplied. Additionally, the mockup data has been adapted to reflect said changes.

Regarding the translation prefix constant: I was not sure if a convention existed for such cases (where a constant is used in multiple locations), so I figured I place the prefix in the AgentState interface.

PR #715, opened 2019-02-15
MH-13369, Delete Capture Agents

lkiesow:
Pull request #384 broke the mechanism to show the icon for deleting
capture agents which are offline from within the admin interface since
the JavaScript code does no longer see the actual status but the
translation key used to represent this status.


smarquard:
Works as advertised

PR #716, opened 2019-02-15
MH-13370, Don't configure unnecessary default credentials

lkiesow:
The default configuration for the ingest service contains unnecessary
credentials which are used if the configured regular expresseion (which
has a typo) matches. This could (given the type get's fixed) cause
trouble with ingests from that system.

This just goes with the empty default configuration instead.

### DIFF
--- a/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
+++ b/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
@@ -10,23 +10,23 @@
 
 org.opencastproject.series.overwrite=true
 
-#The approximate load placed on the system by ingesting a file
-#Since these jobs are *not* dispatched there is no current way to limit the simultaneous number of ingests, but these jobs will block further jobs from running on an already busy admin node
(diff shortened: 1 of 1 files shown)


staubesv (dismissed):
The ingest service does not set those defaults, see https://github.com/opencast/opencast/pull/401/files

lkiesow:
> The ingest service does not set those defaults

Yes, it does once `updated()` is executed.

staubesv:
@lkiesow Can you please point me to the lines of code that set those defaults (I mean the default values http://localhost/.*, opencast_system_account and CHANGE_ME you commented in https://github.com/opencast/opencast/pull/716/files#diff-4cf1facdb31b5379cd031f5edd84d605R30)? I really did not find them...

staubesv:
@lkiesow Ah sorry, now I got it: There is not need to set a value at all.... sorry for the noise...

staubesv (approved):

lkiesow:
This does still mean that there is a problem if there is no
configuration file at all. But that's anither problem.

PR #720, opened 2019-02-15
MH-13374, Update Node.js

lkiesow:
Update Node.js to the [latest stable LTS version](https://nodejs.org/en/download/).


staubesv (approved):
This required the deletion of the folders node* to work, but then it worked.

PR #723, opened 2019-02-18
MH-13377 Fix scheduler rrule TimeZone issue

krinnewitz:
When scheduling mutliple events, the admin UI sends an RRULE with UTC
hours and minutes plus start and end date  to the server.
The server than retrieves the timezone from the capture agent and
adjusts start and end date accordingly, which is fine. However, what was
missing was converting the hour and minute in the RRULE.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
@@ -155,7 +155,6 @@ import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -971,9 +970,7 @@ public String createEvent(EventHttpServletRequest eventHttpServletRequest) throw
         }
         return mediaPackage.getIdentifier().compact();
       case SCHEDULE_MULTIPLE:
-        List<Period> periods = schedulerService.calculatePeriods(rRule, start.toDate(), end.toDate(), duration, tz);
```
krinnewitz, line 974: `periods` was unused here

modules/index-service/src/test/java/org/opencastproject/index/service/impl/IndexServiceImplTest.java
```diff
@@ -754,25 +738,12 @@ public void testCreateEventInputNormalExpectsCreatedRecurringEvent() throws Exce
...
     assertTrue(schedRRule.hasCaptured());
     assertTrue(schedTz.hasCaptured());
-    List<Period> pCheck = calculatePeriods(schedRRule.getValue(), schedStart.getValue(), schedEnd.getValue(), (Long) schedDuration.getValue(), schedTz.getValue());
-    List<Period> pExpected = calculatePeriods(rrule.getValue(), recurrenceStart.getValue(), recurrenceEnd.getValue(), (Long) duration.getValue(), tz.getValue());
-
-    //Assert that the first capture time is the same as the recurrence start
-    assertEquals(pExpected.get(0).getStart(), pCheck.get(0).getStart());
-    //Assert that the end of the last capture time is the same as the recurrence end
-    assertEquals(pExpected.get(pExpected.size() - 1).getEnd(), pCheck.get(pCheck.size() - 1).getEnd());
   }
```
krinnewitz, line 747: Parts of this test case were based on the unnecessary call of `calculatePeriods()` in schedulerServiceImpl

smarquard (approved):
This fixes the issue for recurring event scheduling and conflict checks for us.

PR #724, opened 2019-02-18
MH-13368, Added color property to waveform operation handler

tyfranck:
color property added to waveform operation handler


lkiesow (changes requested):
While I haven't tested this in action, a first glance at the code looks good except for the broken JavaDoc (see Travis CI test failure). Can you fix that?

smarquard:
What is the range of possible values for the color property? There should be some reference to that in the WOH doc.

lkiesow:
Valid options should be valid options to the color option of FFmpeg's [showwavespic](https://ffmpeg.org/ffmpeg-filters.html#showwavespic) filter

tyfranck:
Comments addressed, thanks guys

PR #725, opened 2019-02-18
MH-13378: Add mimetype audio/m4a

miesgre:


lkiesow:
The file already contains a definition for `.m4a` files and while I don't mind changing the mime type if it makes sense, I guess that there should be only one definition. Or is there a specific reason to have both?

https://github.com/opencast/opencast/blob/4b8dc6d2851d725495ec9c09e005ad6fea91ea1c/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml#L177-L181

lkiesow:
Just read through some code and while it does not seem to be a good solution in general, this specifically makes sense for the mine type to file extension mapping.

miesgre:
@lkiesow The `x-` prefix is deprecated (https://tools.ietf.org/html/rfc6648). We can keep `x-` for compability.

And there are some other mimetypes duplicated...
Example (`.m4v`):
https://github.com/opencast/opencast/blob/4b8dc6d2851d725495ec9c09e005ad6fea91ea1c/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml#L432-L436
https://github.com/opencast/opencast/blob/4b8dc6d2851d725495ec9c09e005ad6fea91ea1c/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml#L387-L391

Or (`.mp4` tagged ad `video/mp4` and `audio/mp4`):
https://github.com/opencast/opencast/blob/4b8dc6d2851d725495ec9c09e005ad6fea91ea1c/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml#L387-L391
https://github.com/opencast/opencast/blob/4b8dc6d2851d725495ec9c09e005ad6fea91ea1c/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml#L122-L126

PR #726, opened 2019-02-18
MH-13379, Simplify Mime Type Handling

lkiesow:
I glanced at the mime type handling code in Opencast and could not
resist but had to simplify it :D

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/elementbuilder/AbstractElementBuilderPlugin.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/elementbuilder/AbstractElementBuilderPlugin.java
@@ -22,24 +22,15 @@
 
 package org.opencastproject.mediapackage.elementbuilder;
 
-import org.opencastproject.util.Checksum;
 import org.opencastproject.util.MimeType;
-import org.opencastproject.util.MimeTypes;
-import org.opencastproject.util.UnknownFileTypeException;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.io.File;
-import java.io.IOException;
-import java.security.NoSuchAlgorithmException;
 import java.util.List;
 
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathFactory;
 
 /**
- * This general inplementation of a media package element builder supports specialized implementations by providing
+ * This general implementation of a media package element builder supports specialized implementations by providing
  * tests on the filename an mime type of the file in question.
  */
 public abstract class AbstractElementBuilderPlugin implements MediaPackageElementBuilderPlugin {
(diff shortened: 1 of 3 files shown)


modules/common/src/main/java/org/opencastproject/mediapackage/elementbuilder/AbstractElementBuilderPlugin.java
```diff
@@ -143,15 +143,9 @@ else if (!file.isFile())
...
-    if (mimeTypes != null) {
-      if (mimeTypes.size() > 0) {
-        for (MimeType m : mimeTypes) {
-          if (m.equals(MimeTypes.fromURL(file.toURI().toURL())))
-            return true;
-        }
-      } else {
-        return false;
-      }
+    for (MimeType m : mimeTypes) {
```
staubesv: I would rather recommend to replace all code of this method by 

```
  if (mimeTypes != null) {
    return checkMimeType(file, mimeTypes);
  } else {
    throw new IllegalStateException("Mime types have not been initialized");
  } 

```

;-)

modules/common/src/main/java/org/opencastproject/mediapackage/elementbuilder/AbstractElementBuilderPlugin.java
```diff
@@ -112,7 +112,7 @@ else if (!file.isFile())
     if (mimeTypes != null) {
       if (mimeTypes.length > 0) {
         for (MimeType m : mimeTypes) {
-          if (m.equals(MimeTypes.fromURL(file.toURI().toURL())))
+          if (m.equals(MimeTypes.fromString(file.getPath())))
```
staubesv: Can you remove the null and length (lines 112 & 113) checks here, too?

modules/common/src/main/java/org/opencastproject/util/MimeTypes.java
```diff
@@ -122,58 +115,21 @@ private MimeTypes() {
...
-      String definitions = null;
-      is = MimeTypes.class.getResourceAsStream(DEFINITION_FILE);
-      StringBuilder buf = new StringBuilder();
-      if (is == null)
-        throw new FileNotFoundException(DEFINITION_FILE);
-
-      isr = new InputStreamReader(is);
-      char[] chars = new char[1024];
-      int count = 0;
-      while ((count = isr.read(chars)) > 0) {
```
staubesv, line 138: OMG!!!

staubesv (dismissed):
@lkiesow Can you please apply your simplification (remove null and length checks) in the method checkMimeType(File file) also in the method checkMimeType(File file, MimeType[] mimeTypes) and then make the first method just call the other one? I think that is worth the little effort...

modules/common/src/main/java/org/opencastproject/mediapackage/elementbuilder/AbstractElementBuilderPlugin.java
staubesv: @lkiesow Actually, it seems like the object instance variable "mimeTypes" is never initialised... does this method work at all?

staubesv (approved):

PR #731, opened 2019-02-20
MH-13385: Simplify the configuration of the URL signing components

JulianKniephoff:
Instead of having to consecutively number certain configuration
properties, these changes allow you to specify them in a more free
form manner. This makes generating the corresponding files
automatically much simpler and also makes the manual editing process
less error prone.

### DIFF
--- a/etc/org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider.cfg
+++ b/etc/org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider.cfg
@@ -5,22 +5,13 @@
 # This provider supports multiple configuration blocks, providing the settings for separate distributions (i.e. download
 # or streaming servers, services or paths).
 
-# Each configuration block consists of the following items:
+# A URL signing key has the following attributes:
+
     # Key ID: Key Identifier, e.g. ‘demoKeyOne’
     # Key secret: Key value, e.g. ‘25DA2BA549CB62EF297977845259A’. The key-length is not predefined, but a key length of
         # at least 128 bit is recommended. Any larger value will not increase security of the underlying algorithm.
     # URL prefix: The URL Signing Provider will only sign URLs that start with this value. This allows to support
         # multiple distributions and different key pairs.
-    # Organization: Optional tenant identifier of organization allowed to sign using this key. Can be set to "*"
-        # (default) to indicate that any organization is allowed to use this key for URL signing.
-
-# The properties defined in the configuration file take a numeric suffix that must start at “1" and increase in single
-# increments. In the example below these can be seen as: “.1” and “.2”. As soon as there is a missing number it will
-# stop looking for further entries so be careful not to remove configurations with numbers lower than others. For
-# example if there are configurations using number suffixes from 1 to 5, then commenting out the number 2 configurations
-# will prevent the 3, 4, 5 configurations from being used.
-
-# Note that id and key form a fixed pair, while the same key can be used in more than one configuration block.
 
 # Important: For all URL prefixes, no URL prefix may be a prefix of any other URL prefix
 
@@ -29,12 +20,10 @@
 
 # A typical configuration looks like this:
 
-# id.1=demoKeyOne
-# key.1=6EDB5EDDCF994B7432C371D7C274F
-# url.1=http://download.opencast.org/engage
-# organization.1=mh_default_org
+# key.demoKeyOne.secret=6EDB5EDDCF994B7432C371D7C274F
+# key.demoKeyOne.url=http://download.opencast.org/engage
+# key.demoKeyOne.organization=mh_default_org
 
-# id.2=demoKeyOne
-# key.2=6EDB5EDDCF994B7432C371D7C274F
-# url.2=http://download.opencast.org/custom
-# organization.2=mh_default_org
+# key.demoKeyTwo.secret=6EDB5EDDCF994B7432C371D7C274F
+# key.demoKeyTwo.url=http://download.opencast.org/custom
+# key.demoKeyTwo.organization=mh_default_org
--- a/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
+++ b/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
@@ -39,22 +39,27 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Dictionary;
+import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
-import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, ManagedService {
-  /** The prefix in the configuration file to define the id of the key. */
-  public static final String ID_PREFIX = "id";
-  /** The prefix in the configuration file to define the encryption key. */
-  public static final String KEY_PREFIX = "key";
-  /** The prefix in the configuration file to define the matching url. */
-  public static final String URL_PREFIX = "url";
-  /** The prefix in the configuration file to define the organization owning the key. */
-  public static final String ORGANIZATION_PREFIX = "organization";
+  /** The prefix for key entry configuration keys */
+  public static final String KEY_ENTRY_PREFIX = "key";
+  /** The postfix in the configuration file to define the encryption key. */
+  public static final String SECRET = "secret";
+  /** The postfix in the configuration file to define the matching url. */
+  public static final String URL = "url";
+  /** The postfix in the configuration file to define the organization owning the key. */
+  public static final String ORGANIZATION = "organization";
 
   /** Value indicating that the key can be used by any organization */
   public static final String ANY_ORGANIZATION = "*";
@@ -75,32 +80,14 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
   /**
    * A class to contain the necessary key entries for url signing.
    */
-  private class KeyEntry {
-    private final String keyId;
-    private final String key;
-    private final String organization;
-
-    KeyEntry(String keyId, String key, String organization) {
-      this.keyId = keyId;
-      this.key = key;
-      this.organization = organization;
-    }
-
-    public String getId() {
-      return keyId;
-    }
-
-    public String getKey() {
-      return key;
-    }
-
-    public String getOrganization() {
-      return organization;
-    }
+  private static class KeyEntry {
+    private String key = null;
+    private String url = null;
+    private String organization = ANY_ORGANIZATION;
   }
 
   /** The map to contain the list of keys, their ids and the urls they match. */
-  private Map<String, KeyEntry> keys = new TreeMap<>();
+  private Map<String, KeyEntry> keys = new HashMap<>();
 
   /**
    * @param securityService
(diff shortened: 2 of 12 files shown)


mtneug:
Note that there are other configuration files using this pattern. I know of at least `org.opencastproject.kernel.security.LtiLaunchAuthenticationHandler.cfg` and `org.opencastproject.kernel.security.OAuthConsumerDetailsService.cfg`.

Personally, I find the simplification visually more pleasing and yeah this probably makes it easer to generate configuration files. However, I'm not a fan of allowing arbitrary non-prefixed keys. What if `GenericUrlSigningProvider` suddenly gets additional configuration keys such as the LTI configuration? I would therefore suggest to add a common prefix to all dynamic keys to distinguish them from possible future configuration options.

JulianKniephoff:
@mtneug good point, I think I agree. I'll try to get that in soon-ish.

As for the other files: I think it should not be a problem to only address the files of one logical "feature group" in this PR; it does not really reduce consistency or something. But I will definitely try to take a look where else we could apply this, and maybe even how we can factor this out, so that we don't duplicate too much of this admittedly kind of awkward code.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -110,37 +93,13 @@ public void setSecurityService(SecurityService securityService) {
...
-      throw new IllegalStateException("Url matcher '" + url + "' already registered");
-    keys.put(url, new KeyEntry(keyId, key, organization));
-  }
-
   /**
    * @return The current set of url beginnings this signing provider is looking for.
    */
   public Set<String> getUris() {
-    return Collections.unmodifiableSet(keys.keySet());
+    return keys.values().stream()
```
gregorydlogan, line 104: Question: Does `Collectors.toSet` yield an unmodifiable set?  It's not the end of the world in my eyes, but this would technically change the API here

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -150,121 +109,113 @@ public void addKeyEntry(String keyId, String url, String key, String organizatio
...
-      if (organizationValue == null) {
-        // organization is optimal. Set it to default
-        organizationValue = ANY_ORGANIZATION;
-        getLogger().debug("No organization set for key id {}, using default organization {}", keyIdValue,
-              organizationValue);
+      String[] parts = Arrays.stream(propertyKey.split("\\."))
+              .map(String::trim)
+              .toArray(String[]::new);
+      if (parts.length != 2) {
+        throw new ConfigurationException(propertyKey, "wrong property key format");
```
gregorydlogan, line 146: That's nice.  Which key? :)

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -150,121 +109,113 @@ public void addKeyEntry(String keyId, String url, String key, String organizatio
...
-      } catch (IllegalStateException e) {
-        throw new ConfigurationException(urlValue, e.getMessage());
+      String id = parts[0];
+      KeyEntry currentKeyEntry = keys.computeIfAbsent(id, __ -> new KeyEntry());
+
+      String attribute = parts[1];
+      String propertyValue = Objects.toString(properties.get(propertyKey), "").trim();
+      switch (attribute) {
+        case ORGANIZATION:
+          currentKeyEntry.organization = propertyValue;
```
gregorydlogan, line 159: If `propertyValue` is the empty string here, is this a valid thing to do?

gregorydlogan (changes requested):

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Looking at the old code again, none of the property values were allowed to be `null` or empty. Somehow I missed adopting that behavior. :sweat_smile: Will fix.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Not sure what you are getting at here. You mean I should add which key was the culprit to the error message? Shouldn't the first parameter to `ConfigurationException` suffice? My hope was that the container displays that somewhere. :thinking:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: `Collectors.toSet` does not guarantee unmodifiability, it turns out. In fact, it's contract is pretty minimalistic. It says "Returns a `Collector` that accumulates the input elements into a new `Set`. [...]" from which **I think** we can assume that the `Set`s returned by it are independent of each other and of the stream it was fed from.

So the main rationale to make the return value unmodifiable -- i.e. not handing out a thing by which the outside world can tamper with our internals, as was the case with `keySet` -- **should** be gone.

But maybe it is the moral thing to wrap it anyway; however, is not doing so really an API change? Do we ever actually state that this returns an unmodifiable map?

JulianKniephoff:
@mtneug's suggestion should be implemented as well, now.

etc/org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider.cfg
```diff
@@ -29,12 +20,10 @@
...
+# demoKeyOne.url=http://download.opencast.org/engage
+# demoKeyOne.organization=mh_default_org
 
-# id.2=demoKeyOne
-# key.2=6EDB5EDDCF994B7432C371D7C274F
-# url.2=http://download.opencast.org/custom
-# organization.2=mh_default_org
+# demoKeyTwo.key=6EDB5EDDCF994B7432C371D7C274F
+# demoKeyTwo.url=http://download.opencast.org/custom
+# demoKeyTwo.organization=mh_default_org
```
mtneug: Shouldn't this be prefixed as well?

JulianKniephoff: Good catch, thanks! :smile:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
gregorydlogan: Not really I guess, but it does change previous (undocumented) behaviour.  I would wrap it out of extra caution, if only to prevent someone accidentally mangling the URIs for everyone else.  I have a vague recollection of an unmodifiableStream() or something, but offhand I don't remember where (or even if) I saw it.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
gregorydlogan: I believe (although I haven't actually tested) that the container will display the stack trace - `ConfigurationException` extends `RuntimeException`, but doesn't have any fancy logic saying to output the key name.  Generally I would drop a log entry at ERROR, then throw the `ConfigurationException`.

JulianKniephoff: There might not be specific logic, but `ConfigurationException` formats its message nicely. It looks like this:

```
2019-03-07 09:10:36,408 | ERROR | (Log:186) - [org.osgi.service.cm.ManagedService, org.opencastproject.security.urlsigning.provider.UrlSigningProvider, id=795, bundle=178/mvn:org.opencastproject/opencast-urlsigning-service-impl/7-SNAPSHOT]: Updating property key.foo.demoKeyOne.secret of configuration org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider caused a problem: wrong property key format
```

followed by the stack trace, as you said.

I can add an additional log entry if you insist, of course, but I think that message contains everything we need to know, and I'm generally not a friend of excessive, unnecessary logging. :wink:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Note that this is resolved [here](https://github.com/opencast/opencast/pull/731/files#diff-34839f84f847d48880db263f855ba1c9R155).

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Okay, [here](https://github.com/opencast/opencast/pull/731/files#diff-34839f84f847d48880db263f855ba1c9R108) you go. :wink:

JulianKniephoff: Also just for reference: What you are thinking of might be `Collectors.toUnmodifiableSet`, which is only in Java 10, though, unfortunately. :disappointed:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
gregorydlogan: I would argue that the average user isn't going to parse the whole message and will just complain on list, but technically you are correct (the best kind).

gregorydlogan (approved):

JulianKniephoff:
Not to be impatient here, but is there anything still blocking this? :sweat_smile:

gregorydlogan:
Not that I can see, aside from me not having my signing keys on me. @mtneug, can you merge this when you get a chance? I'm afk until the 22nd or so.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -75,32 +80,14 @@
...
-    }
+  private static class KeyEntry {
+    private String key = null;
+    private String url = null;
+    private String organization = ANY_ORGANIZATION;
   }
 
   /** The map to contain the list of keys, their ids and the urls they match. */
-  private Map<String, KeyEntry> keys = new TreeMap<>();
+  private Map<String, KeyEntry> keys = new HashMap<>();
```
staubesv: @JulianKniephoff Looking at the lines 

https://github.com/opencast/opencast/pull/731/files#diff-34839f84f847d48880db263f855ba1c9R85

and

https://github.com/opencast/opencast/pull/731/files#diff-34839f84f847d48880db263f855ba1c9R119

it seems that either me or you have misunderstood the purpose of "keys".

To my understanding, the URL signing provider needs fast lookup of keys based on the URL. That's why they used a TreeMap with Map<String(=URL),KeyEntry>.

If that is correct, you have "un-"optimized that code and made it slower ;-)

staubesv:
@JulianKniephoff Would be nice if you could raise a PR that briefly notes that the URL signing configuration needs to be adapted. A first Adopter already ran into this: https://groups.google.com/a/opencast.org/forum/#!topic/users/hmArC_ylKik

JulianKniephoff:
@staubesv you mean something like #909? :wink:

staubesv:
@JulianKniephoff Yes, thanks!

PR #735, opened 2019-02-23
MH-13390, Quick-Filter by Presenter

lkiesow:
Like with series meta-data, it would be nice to quickly filter the
events table by clicking on a presenter's name.

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -546,7 +546,8 @@
            "EDITOR": "Open video Editor",
            "EDITOR_NEEDS_CUTTING": "Open video Editor (a comment indicates that cutting the video has been requested)",
            "COMMENTS": "View comments",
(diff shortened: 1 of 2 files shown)


staubesv (changes requested):
Is there a specific reason to not implement the same quick-filter on series.presenter (maybe even series.organizer)? Seems just somewhat unexpected that the same field (presenter) can be filtered on one table but not the other...

... what do you think about this?

lkiesow:
In general, that would be nice to have on several other places but I don't have time for that right now ;-)
Here is where it did bother me. And I don't think that it is too inconsistent since it is implemented on some fields already.

staubesv (changes requested):
Basically works as expected, but there is a cosmetic issue: The filter value is not shown in the selected filter.

Example: I've left-clicked on the presenter "Sven Stauber". The event table does now filter for "Sven Stauber", but the filter value "Sven Stauber" is not shown in the filter:

screen shot 2019-02-25 at 12 34 01

Can you please have a look at this?

[Update]: The value "Sven Stauber" was not yet in the list of persons available to the filter, but that issue is not caused by this PR. I've created https://opencast.jira.com/browse/MH-13391 for that.

staubesv (approved):

PR #736, opened 2019-02-26
Revert #523: Special handling of asset manager event removal

lkiesow:
This patch reverts pull request #523 which introduced a special query option to indicate that a query will remove a whole media package which would disable a set of checks and cleanup operations.

Since pull request #696 (alone with the other asset manager performance fixes) removed the whole expensive handling which was circumvented, this now only removes some left-over complexity which had no gain any longer.

Note that the performance patches are not yet merged into `develop` (which really means that `r/6.x` is not merged) which is intentional since they cause merge conflicts with pull request #523 meaning that the merge should be way cleaner after this is merged.

### DIFF
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/ADeleteQuery.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/ADeleteQuery.java
@@ -29,13 +29,6 @@ public interface ADeleteQuery {
   ADeleteQuery name(String queryName);
 
   /**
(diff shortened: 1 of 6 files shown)


lkiesow:
Note that additionally, pull request #650 claims that the changes introduced with pull request #523 also cause problems with event deletion. Even more reasons to drop this again.

smarquard:
There is such a complex chain of changes to asset manager for performance fixes that the only rational thing to do here is sync up develop to match 6.x, as 6.x has been more thoroughly tested than develop. So I think this revert makes sense, and after 6.x has been merged back to develop, we can take a look again at the outstanding functional and performance issues.

staubesv (approved):

PR #737, opened 2019-02-27
MH-13392: Added allowConflict parameter to methods and implemented

TurRil:
Add a rest parameter to the EventsEndpoint > @PUT > @Path("{eventId}/scheduling") endpoint; 
called _allowConflict_ to send to _SchedulerServiceImpl_ to allow conflicts and not perform the conflict check when updating an event, only if the call was made by an admin user and the flag is checked.

If not handled properly this will likely cause two or more events to be scheduled on a particular capture agent at the same time, which will then cause a capture failure for all but one of the events.

Now the reason that this exists is that it will be possible to move around events from capture agents and have them overlap (conflict) for a short period of time until all the shuffling of events to different locations have been completed. 

ONLY IF THE FLAG IS SET AND THE CALL IS MADE WITH ADMIN PRIVILEGES.

So it will be up to the person writing the script to add logic in terms of conflict checking afterwards, but by setting the flag and using an admin user to make the call; that's what they signed up for.

If this is not implemented then a method for moving events to temporary locations will have to be implemented, and shuffling all the correct locations back might cause huge processing overheads. This is equivalent to a Towers of Hanoi problem (the https://en.wikipedia.org/wiki/Tower_of_Hanoi) but with potentially many more events/locations/discs to move around.

https://opencast.jira.com/browse/MH-13392

### DIFF
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -42,9 +42,6 @@ Configuration changes
 API changes
 -----------
 
-Due to [MH-13392](https://opencast.jira.com/browse/MH-13392):
-- Modified PUT /api/events/{id}/scheduling: Added optional field 'allowConflict'
-
 Due to [MH-13397](https://opencast.jira.com/browse/MH-13397):
 
 - Modified GET /recordings/{id}/technical.json: Removed field `optOut`
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -42,6 +42,9 @@ Configuration changes
 API changes
 -----------
 
+Due to [MH-13392](https://opencast.jira.com/browse/MH-13392):
+- Modified PUT /api/events/{id}/scheduling: Added optional field 'allowConflict'
+
 Due to [MH-13397](https://opencast.jira.com/browse/MH-13397):
 
 - Modified GET /recordings/{id}/technical.json: Removed field `optOut`
--- a/docs/guides/developer/docs/api/events-api.md
+++ b/docs/guides/developer/docs/api/events-api.md
@@ -886,9 +886,10 @@ Available since API version 1.1.0.
 
 Update the scheduling information of the event with id `{event_id}`.
 
-Form Parameters             |Type            | Description
-:---------------------------|:---------------|:----------------------------
-`scheduling`                | `string`       | The scheduling information.
+Form Parameters             |Type            | Description                              | Default | Version   
+:---------------------------|:---------------|:-----------------------------------------|:--------|:------- 
+`scheduling`                | `string`       | The scheduling information.              | <span class="required">Required</span>| 1.1.0    
+`allowConflict`             | `boolean`      | Allow conflicts when updating scheduling.| false   | 1.2.0
 
 __Sample__
 
@@ -919,3 +920,9 @@ In case of a conflict:
   }
 ]
 ```
+
+`allowConflict` was introduced in version 1.2.0, it allows the schedule to be updated without checking for conflicts. To allow conflicts (`true`) the call **MUST** be made with a user that has an _Administrative Role_.
+
+If not handled properly this will likely cause two or more events to be scheduled on a particular capture agent at the same time, which will then cause a capture failure for all but one of the events.
+
+The person making this call and allowing conflicts to exist, will bear the responsibility of resolving the conflicts that might result.
--- a/docs/guides/developer/docs/api/events-api.md
+++ b/docs/guides/developer/docs/api/events-api.md
@@ -921,9 +921,8 @@ In case of a conflict:
 ]
 ```
 
-`allowConflict` was introduced in version 1.2.0, it allows the schedule to be
-updated without checking for conflicts. To allow conflicts (`true`) the call
-**MUST** be made with a user that has an _Administrative Role_.
+`allowConflict` allows the schedule to be updated without checking for conflicts.
+To allow conflicts (`true`) the call **MUST** be made with a user that has an _Administrative Role_.
 
 If not handled properly this will likely cause two or more events to be
 scheduled on a particular capture agent at the same time, which will then
--- a/docs/guides/developer/docs/api/events-api.md
+++ b/docs/guides/developer/docs/api/events-api.md
@@ -886,9 +886,9 @@ Available since API version 1.1.0.
 
 Update the scheduling information of the event with id `{event_id}`.
 
-Form Parameters             |Type            | Description                              | Default | Version   
-:---------------------------|:---------------|:-----------------------------------------|:--------|:------- 
-`scheduling`                | `string`       | The scheduling information.              | <span class="required">Required</span>| 1.1.0    
+Form Parameters             |Type            | Description                              | Default | Version
+:---------------------------|:---------------|:-----------------------------------------|:--------|:-------
+`scheduling`                | `string`       | The scheduling information.              | <span class="required">Required</span>| 1.1.0
 `allowConflict`             | `boolean`      | Allow conflicts when updating scheduling.| false   | 1.2.0
 
 __Sample__
@@ -921,8 +921,13 @@ In case of a conflict:
 ]
 ```
 
-`allowConflict` was introduced in version 1.2.0, it allows the schedule to be updated without checking for conflicts. To allow conflicts (`true`) the call **MUST** be made with a user that has an _Administrative Role_.
+`allowConflict` was introduced in version 1.2.0, it allows the schedule to be
+updated without checking for conflicts. To allow conflicts (`true`) the call
+**MUST** be made with a user that has an _Administrative Role_.
 
-If not handled properly this will likely cause two or more events to be scheduled on a particular capture agent at the same time, which will then cause a capture failure for all but one of the events.
+If not handled properly this will likely cause two or more events to be
+scheduled on a particular capture agent at the same time, which will then
+cause a capture failure for all but one of the events.
 
-The person making this call and allowing conflicts to exist, will bear the responsibility of resolving the conflicts that might result.
+The person making this call and allowing conflicts to exist, will bear the
+responsibility of resolving the conflicts that might result.
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -457,7 +457,7 @@ public class EventsEndpoint implements ManagedService {
         if (eventHttpServletRequest.getScheduling().isSome() && !requestedVersion.isSmallerThan(VERSION_1_1_0)) {
           // Scheduling is only available for version 1.1.0 and above
           Optional<Response> clientError = updateSchedulingInformation(
-              eventHttpServletRequest.getScheduling().get(), eventId, requestedVersion);
+              eventHttpServletRequest.getScheduling().get(), eventId, requestedVersion, false);
           if (clientError.isPresent()) {
             return clientError.get();
           }
@@ -1706,18 +1706,18 @@ public class EventsEndpoint implements ManagedService {
   @RestQuery(name = "updateeventscheduling", description = "Update an event's scheduling information.", returnDescription = "", pathParameters = {
       @RestParameter(name = "eventId", description = "The event id", isRequired = true, type = Type.STRING) }, restParameters = {
       @RestParameter(name = "scheduling", isRequired = true, description = "Scheduling Information", type = Type.STRING),
-      @RestParameter(name = "allowConflict", description = "Allow conflicts when updating scheduling", defaultValue = "false", isRequired = false, type = Type.BOOLEAN) }, reponses = {
+      @RestParameter(name = "allowConflict", description = "Allow conflicts when updating scheduling", isRequired = false, type = Type.BOOLEAN) }, reponses = {
       @RestResponse(description = "The  scheduling information for the specified event is updated.", responseCode = HttpServletResponse.SC_NO_CONTENT),
       @RestResponse(description = "The specified event has no scheduling information to update.", responseCode = HttpServletResponse.SC_NOT_ACCEPTABLE),
       @RestResponse(description = "The scheduling information could not be updated due to a conflict.", responseCode = HttpServletResponse.SC_CONFLICT),
       @RestResponse(description = "The specified event does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
   public Response updateEventScheduling(@HeaderParam("Accept") String acceptHeader, @PathParam("eventId") String id,
                                  @FormParam("scheduling") String scheduling,
-                                 @FormParam("allowConflict") boolean allowConflict) throws Exception {
+                                 @FormParam("allowConflict") @DefaultValue("false") boolean allowConflict) throws Exception {
     final ApiVersion requestedVersion = ApiMediaType.parse(acceptHeader).getVersion();
     final Opt<Event> event = indexService.getEvent(id, externalIndex);
 
-    if (!requestedVersion.isSmallerThan(ApiVersion.VERSION_1_2_0)) {
+    if (requestedVersion.isSmallerThan(ApiVersion.VERSION_1_2_0)) {
         allowConflict = false;
     }
     if (event.isNone()) {
@@ -1733,14 +1733,6 @@ public class EventsEndpoint implements ManagedService {
     }
     Optional<Response> clientError = updateSchedulingInformation(parsedJson, id, requestedVersion, allowConflict);
     return clientError.orElse(Response.noContent().build());
-  }
-
-  private Optional<Response> updateSchedulingInformation(
-      JSONObject parsedScheduling,
-      String id,
-      ApiVersion requestedVersion) throws Exception {
-
-    return updateSchedulingInformation(parsedScheduling, id, requestedVersion, false);
   }
 
   private Optional<Response> updateSchedulingInformation(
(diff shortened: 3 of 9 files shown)


smarquard:
We discussed whether to add a second method signature here for SchedulerService.updateEvent() or just change the existing one (similarly for updateSchedulingInformation()).

It might be cleaner just to change the existing one, which will require some changes elsewhere in the coed base where that's called. Comments welcome.

staubesv:
@smarquard Can you please do this in External API version 1.2.0? You're lucky: There is another PR that already introduces version 1.2.0 (see #642), so adding 1.2.0 to this one is trivial. Please also update the External API specification found on https://docs.opencast.org/develop/developer/api/.

TurRil:
Hi @staubesv, 
I rebased the branch to follow on from #642 (version 1.2.0) support and added the documentation for the new parameter.

staubesv:
@smarquard  I disagree on this being equivalent to the Tower of Hanoi problem as that one would be trivial to solve if you had an unlimited number of rods. Anyway, I see the point that you would not want to move events around for efficiency reasons.

smarquard:
Sorry @staubesv I wish I followed your comment above, but am lost. More context? Or did this end up here from another thread on another PR by mistake?

smarquard:
Tower of Hanoi is @TurRil's description, but yes. Short version is we allow admins to introduce temporary conflicts that they are responsible for resolving, for efficiency reasons.

staubesv:
@smarquard Ah, sorry. It was @TurRil that mentioned the Tower of Hanoi problem in the description of this PR.

Anyway, no need for action here - that was just a comment about the equivalence of problems ;-)

staubesv:
@TurRil The pull request has merge conflicts. Please resolve them so that I can start the review.

docs/guides/admin/docs/releasenotes.md
```diff
@@ -42,6 +42,9 @@ Configuration changes
 API changes
 -----------
 
+Due to [MH-13392](https://opencast.jira.com/browse/MH-13392):
```
staubesv: This is confusing as this PR does not change this request for existing External API versions so there is no API change here. I suggest to just omit this.

docs/guides/developer/docs/api/events-api.md
```diff
@@ -886,9 +886,10 @@ Available since API version 1.1.0.
 
 Update the scheduling information of the event with id `{event_id}`.
 
-Form Parameters             |Type            | Description
-:---------------------------|:---------------|:----------------------------
-`scheduling`                | `string`       | The scheduling information.
+Form Parameters             |Type            | Description                              | Default | Version
+:---------------------------|:---------------|:-----------------------------------------|:--------|:-------
+`scheduling`                | `string`       | The scheduling information.              | <span class="required">Required</span>| 1.1.0
```
staubesv, line 891: Should we really use HTML in markdown? Seems somewhat superfluous here.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1704,17 +1704,22 @@ public Response getEventScheduling(@HeaderParam("Accept") String acceptHeader, @
   @Path("{eventId}/scheduling")
   @Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0 })
   @RestQuery(name = "updateeventscheduling", description = "Update an event's scheduling information.", returnDescription = "", pathParameters = {
-      @RestParameter(name = "eventId", description = "The event id", isRequired = true, type = STRING) }, restParameters = {
-      @RestParameter(name = "scheduling", isRequired = true, description = "Scheduling Information", type = STRING) }, reponses = {
+      @RestParameter(name = "eventId", description = "The event id", isRequired = true, type = Type.STRING) }, restParameters = {
+      @RestParameter(name = "scheduling", isRequired = true, description = "Scheduling Information", type = Type.STRING),
+      @RestParameter(name = "allowConflict", description = "Allow conflicts when updating scheduling", defaultValue = "false", isRequired = false, type = Type.BOOLEAN) }, reponses = {
```
staubesv: Due to MH-13416 REST docs defaultValue not correctly set for boolean parameters, setting the defaultValue for a boolean REST parameter will cause the REST docs to always check the checkbox representing the value of this parameter. As far now, the overall result is better when you just omit the defaultValue (will default to false in that case)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1704,17 +1704,22 @@ public Response getEventScheduling(@HeaderParam("Accept") String acceptHeader, @
...
       @RestResponse(description = "The scheduling information could not be updated due to a conflict.", responseCode = HttpServletResponse.SC_CONFLICT),
       @RestResponse(description = "The specified event does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
   public Response updateEventScheduling(@HeaderParam("Accept") String acceptHeader, @PathParam("eventId") String id,
-                                 @FormParam("scheduling") String scheduling) throws Exception {
+                                 @FormParam("scheduling") String scheduling,
+                                 @FormParam("allowConflict") boolean allowConflict) throws Exception {
     final ApiVersion requestedVersion = ApiMediaType.parse(acceptHeader).getVersion();
     final Opt<Event> event = indexService.getEvent(id, externalIndex);
 
+    if (!requestedVersion.isSmallerThan(ApiVersion.VERSION_1_2_0)) {
```
staubesv: This should be

`if (requestedVersion.isSmallerThan(ApiVersion.VERSION_1_2_0)) { [...]`

shouldn't it?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1726,14 +1731,24 @@ public Response updateEventScheduling(@HeaderParam("Accept") String acceptHeader
       logger.debug("Client sent unparsable scheduling information for event {}: {}", id, scheduling);
       return RestUtil.R.badRequest("Unparsable scheduling information");
     }
-    Optional<Response> clientError = updateSchedulingInformation(parsedJson, id, requestedVersion);
+    Optional<Response> clientError = updateSchedulingInformation(parsedJson, id, requestedVersion, allowConflict);
     return clientError.orElse(Response.noContent().build());
   }
 
   private Optional<Response> updateSchedulingInformation(
```
staubesv, line 1739: As this private method is called only once, I would rather remove it and adapt the caller to use the extended version with allowConflict set to false.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -944,7 +943,7 @@ protected MetadataList deserializeMetadataList(String json) throws ParseExceptio
   @RestQuery(name = "updateseriesacl", description = "Updates a series' access policy.", returnDescription = "", pathParameters = {
           @RestParameter(name = "seriesId", description = "The series id", isRequired = true, type = STRING) }, restParameters = {
                   @RestParameter(name = "acl", isRequired = true, description = "Access policy", type = STRING),
-                  @RestParameter(name = "override", isRequired = false, description = "If true the series ACL will take precedence over any existing episode ACL", type = BOOLEAN)}, reponses = {
+                  @RestParameter(name = "override", isRequired = false, defaultValue = "false", description = "If true the series ACL will take precedence over any existing episode ACL", type = STRING)}, reponses = {
```
staubesv: This change does not belong to this PR, does it?

modules/scheduler-api/src/main/java/org/opencastproject/scheduler/api/SchedulerService.java
```diff
@@ -175,6 +175,48 @@ void updateEvent(String mediaPackageId, Opt<Date> startDateTime, Opt<Date> endDa
           Opt<Map<String, String>> caMetadata)
                   throws NotFoundException, UnauthorizedException, SchedulerConflictException, SchedulerException;
 
+  /**
+   * Updates event with specified ID and possiblyDictionary checking for conflicts.
```
staubesv: possiblyDictionary?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -791,6 +803,17 @@ public Boolean apply(MediaPackage mp) {
     }
   }
 
+  private boolean isAdmin() {
```
staubesv, line 806: Hmm... I think that the chance that you find this logic somewhere else in the Opencast codebase (-> code re-use)  is very high. Can you please check?

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -688,8 +698,10 @@ private void updateEventInternal(final String mpId, Opt<Date> startDateTime,
       String agentId = extendedEventDto.getCaptureAgentId();
       Opt<String> seriesId = Opt.nul(record.getSnapshot().get().getMediaPackage().getSeries());
 
+      // Check for conflicting events
       // Check scheduling conficts in case a property relevant for conflicts has changed
-      if (captureAgentId.isSome() || startDateTime.isSome() || endDateTime.isSome()) {
+      if ((captureAgentId.isSome() || startDateTime.isSome() || endDateTime.isSome())
+            && (!isAdmin() || (isAdmin() && !allowConflict))) {
```
staubesv: Slight performance optimization:

`&& (!allowConflicts || !isAdmin()) 
`

is more efficient as the user's role needs not to be determined if conflicts are not allowed anyway which is the case in 99.99% of calls to this method.

We likely can save one or more trees (think of energy consumption) by avoiding the execution of that code ;-)

docs/guides/developer/docs/api/events-api.md
```diff
@@ -919,3 +920,14 @@ In case of a conflict:
   }
 ]
 ```
+
+`allowConflict` was introduced in version 1.2.0, it allows the schedule to be
```
staubesv: I would omit "was introduced in version 1.2.0, it" as the table above already indicates that.

staubesv (changes requested):

staubesv:
@smarquard Besides the minor review comments, is this still "Work in progress"? (you set that label according to github)

smarquard:
I added work-in-progress at the time because Corne was away for a while.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
smarquard: Is it necessary to do this at all? It's optional and the default is false anyway. Seems like this would override the value if you requested an API version < 1.2.0 and you provided a true value here, it would get changed to false. Doesn't seem like a necessary thing to enforce, unless I'm missing something.

staubesv: Without that, the older versions would essentially support this new parameter, too. So strictly spoken: It is necessary.

smarquard: I think it's potentially confusing to silently change the value like that. If we really want to be strict about not accepting a parameter in an earlier API version that was added in a later version, then we should return a Bad Request rather than accept it and ignore it.

docs/guides/admin/docs/releasenotes.md
TurRil: Ok, will remove it.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
staubesv: @smarquard AFAIK, form parameters that are not specified in the method signature will simply be ignored, i.e. if you pass an unsupported form parameter, no request would respond with "bad request".  Setting the value to false therefore makes sense as it prevents the value to be set by API versions not supposed to support the option.

docs/guides/developer/docs/api/events-api.md
TurRil: The styling of "Required" text adds value to this case where you can immediately see which field is actually required. It enriches the documentation. With markdown you can't really control how it will look.

The style can be reused to update the rest of the documentation later on.

docs/guides/developer/docs/api/events-api.md
TurRil: Ok, will remove.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
TurRil: Ok, changing this and it's good to know.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
TurRil: Yes, it should be
`if (requestedVersion.isSmallerThan(ApiVersion.VERSION_1_2_0)) { [...]`

Corrected in my code.

I think that if it is not explicitly set to version 1.2.0 or above (by default it takes the newest version anyway) this should be set to false so that it reacts like the parameter doesn't exist for older versions.

Even if it is set to false by default which "Due to MH-13416 REST docs defaultValue not correctly set for boolean parameters..." is uncertain at this point.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
TurRil: Nope, removed it.

modules/scheduler-api/src/main/java/org/opencastproject/scheduler/api/SchedulerService.java
TurRil: :p should just be possibly

docs/guides/developer/docs/api/events-api.md
staubesv: Ok.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
TurRil: 👍 :D

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
staubesv: @TurRil The @RestParameters annotations are Opencast-specific. The only thing they are used for is the dynamic creation of the REST docs page. The only thing that "defaultValue = "false"" does it to set the default value in the REST docs page - this won't affect the actual REST call (except when performed by the REST docs page).

To set the default value used if calls don't set a value for the form parameter "allowConflict", please change

` @FormParam("allowConflict") boolean allowConflict`

to 

` @FormParam("allowConflict") @DefaultValue("false") boolean allowConflict`

@FormParam and @DefaultValue are Java annotations used by JAX-RS.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
TurRil: So this kind of function does exist in _modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerWithSecurity.java_ [192]

And a similar call is made in 11 files in the project, so I'm thinking that maybe it would be better if this was a public function in the "user class" (_modules/common/src/main/java/org/opencastproject/security/api/User.java_) ref: MH-13470

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
TurRil: 👍 ok removing the extra function.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
TurRil: Thanks - added it.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
staubesv: Ok, thanks for checking! 

While the class User would be a better location for this logic, this seems somewhat out-of-scope for this pull request.

I therefore suggest to leave the logic here.

If you have some more time,  you could simplify this into:

```
return (securityService.getUser().hasRole(GLOBAL_ADMIN_ROLE)
     || securityService.getUser().hasRole(securityService.getOrganization().getAdminRole());

```

staubesv:
@TurRil We should hurry up as the feature freeze is approaching... Building this PR currently fails:

```
[ERROR] /home/travis/build/opencast/opencast/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java:[1716,63] cannot find symbol
  symbol:   class DefaultValue
  location: class org.opencastproject.external.endpoint.EventsEndpoint
```

You need to add the import 

`import javax.ws.rs.DefaultValue;`

staubesv (approved):

PR #740, opened 2019-02-27
MH-13346 Thumbnail generation from Admin UI can fail on longer videos

mliradelc:
This is a temporary fix that disables completely the generation of a thumbnail in the Admin UI editor.
This fix also disables the thumbnail for the track view in the editor.

Original discussion: https://groups.google.com/a/opencast.org/forum/#!topic/dev/_6qvemO50bg


KatrinIhler:
@mliradelc This PR has test failures!

mliradelc:
> @mliradelc This PR has test failures!

You have right!, the error is in the check style, I will fix ASAP

staubesv:
@mliradelc There is no need to disable the preview thumbnail of the track preview as that one is generated by the workflow (vs. interactively).

There is also no need to disable the feature completely: The only thing that works slow in the community version is the "interactive" generation of the thumbnail because downloading the source file into the workspace is terribly slow for very large files (the operations on the file are cheap).

Consider adding a configuration setting that allows Adopters to choose whether they want to enable/disable thumbnail previews (in the video editor). Depending on that setting, the backend can just not generate a preview image in case it would imply downloading the file into the workspace (no problems with uploaded images which work well).

That way, Opencast users can still select the thumbnail but just don't get a preview in the editor.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
```diff
@@ -649,24 +640,24 @@ public Response editVideo(@PathParam("mediapackageid") final String mediaPackage
...
-          }
-        } catch (UrlSigningException | URISyntaxException e) {
-          logger.error("Error while trying to serialize the thumbnail url because: {}", getStackTrace(e));
-          return R.serverError();
-        } catch (IOException | EncoderException | PublicationException | UnknownFileTypeException | MediaPackageException e) {
-          logger.error("Error while updating default thumbnail because: {}", getStackTrace(e));
-          return R.serverError();
-        }
-      }
+//      MediaPackageElement distributedThumbnail = null;
```
mliradelc, line 643: @staubesv When I made the tests, even if you didn't touch the interactive thumbnail generation, if  you hit save and continue (When you made a cut in the video), that calls the composer to create the thumbnail as a default configuration (If you didn't upload any thumbnail or created one using the interactive method).

Nevertheless this is an emergency fix. I will work in a next PR to solve better and restore this feature.

smarquard: Agreed, this code has to go, as it can otherwise cause extremely slow responses to "Save and Continue" (POST to editor.json). This operation should happen in a workflow anyway, so it's unclear why it had to be included here and run before you get a response from the POST to editor.json.

staubesv: For this kind of job, workflows are not a valid option as workflows are a) far(!) too slow and b) appear in the history of processing. Here, we don't actually process the event. We just needs a preview image used by the editor. The actual thumbnail will be generated by the workflow.

POST /admin-ng/editor.json needs to update the preview because cutting the video will result in another default thumbnail which the user should see if he presses "Save".

Agree that the extraction is too slow in the community version and should not happen, though.

staubesv: @mliradelc I'll have a quick look at this now to see whether I can come up with a solution - at the end of the day, it was us that caused the problem.

staubesv:
@mliradelc & @smarquard Admittedly not a totally pleasing solution, but seen as a "emergency" fix, #743 manages to still allow users to select thumbnail (although usability does suffer quite a bit) and avoids the performance problem.

The mode configuration property would in theory allow the contribution of our custom patch that solves the performance issue in a rather pragmatic way... I'll bring the topic up in one of the next technical meetings I can join (currently, Tuesday does not work for me).

mliradelc:
@staubesv I just Saw the #743 PR and is a better solution than disable from the code directly from what I did.
I will close this PR

PR #741, opened 2019-02-27
MH-13394 Remove unfinished feature "User Blacklisting"

staubesv:
Reason: The implementation of this feature was never finished and the related code is therefore dead code.

The module admin UI contains quite some code that is related to a feature that has never been finished: User Blacklisting.

The idea seemed to be that black lists can be created for individual users that black list the user within a given time period including a reason for blacklisting and a comment.

The module admin UI contained to following code:

- User blacklist table (username, date_from, date_to and reason)
- User blacklist modal
- User blacklist wizard ("Add user blacklist")

IMO, that seems quite a overkill. Or is there anybody out there that would be willing to manage all those blacklists if the feature was completed?

The KISS principle would reduce this to a new user status field ("active", "inactive") and the ability to set this status.

The goal of this task is to remove the unused code of a never finished feature that existed since Opencast 2.x in our code basis.


smarquard:
Seems like a good cleanup.

lkiesow:
@staubesv there seems to be something wrong with the changes to the `lang-de_DE.json` and `lang-en_US.json`. The changes are huge and seem to be mostly unrelated.

staubesv:
@lkiesow Those are test resources required by the mockup of the Admin UI. I really just copied to language files of the actual web application (where I have removed the user blacklisting language keys) over these, so that should be fine.

staubesv:
@lkiesow I had another look at the language files and realised that those are not used by the mockup but rather by some unit tests. Therefore, it indeed does not make much sense to copy the web app translation files over those. Will update the PR soon.

staubesv:
@lkiesow Your turn again.

PR #744, opened 2019-02-28
MH-13399, Update Elasticsearch Configuration

lkiesow:
This patch updates the meachanism of how Opencast's internal
Elasticsearch is configured. It moves most of the configuration into the
bundle resources and mostly unifies the externalapi and adminui index
configuration.

Reasons for the change are:

1. A lot of the exposed configuration options were actually fixed
   settings and changing them could cause errors in Opencast. In other
   words, there was nothing to configure at all.

2. The current configuration was mixing node and index configuration
   which is no longer allowed starting with Elasticsearch 5.0. This
   patch properly separates these settings.

3. All relevant configuration setting for both indexes were just
   duplicated. Even worse, they had to be since the main node and some
   index settings would only be loaded from one configuration location
   and not defining it twice could lead to inconsistent behaviour of
   Opencast after restarts.


lkiesow:
Fixed merge conflict

modules/search/src/main/resources/elasticsearch/indexSettings.json
```diff
@@ -1,12 +1,15 @@
...
+  "max_result_window": 2147483647,
   "analysis": {
     "analyzer": {
       "lowercasespaceanalyzer": {
         "type": "custom",
         "tokenizer": "whitespace",
         "filter": [
           "lowercase"
-        ]
+          ]
```
gregorydlogan: Was this extra tab deliberate?

gregorydlogan:
I'm not terribly familiar with the guts of ES, but if I wanted to run it as a separate service external to Opencast, is that possible with this new PR?  Otherwise, looks good to me!

etc/elasticsearch.yml
```diff
@@ -0,0 +1,59 @@
...
+# Set a custom allowed content length:
+#http.max_content_length: 100mb
+
+# Disable HTTP completely:
+#http.enabled: false
+
+
+################################## Shutdown ###################################
+
+# Disable the Elasticsearch Shutdown API (recomended).
```
staubesv: Typo: recom**m**ended

etc/elasticsearch.yml
```diff
@@ -0,0 +1,59 @@
+## Elasticsearch Configuration
+
+# Cluster name identifies your cluster for auto-discovery. If you're running
+# multiple clusters on the same network, make sure you're using unique names.
+cluster.name: opencast
+
+#################################### Paths ####################################
+
+path.home: ${karaf.home}/
```
staubesv, line 8: No comment here?

etc/elasticsearch.yml
```diff
@@ -0,0 +1,59 @@
...
+
+# Path to directory where to store index data allocated for this node.
+#
+#path.data: /path/to/data
```
staubesv: This line seems superfluous

etc/elasticsearch.yml
```diff
@@ -0,0 +1,59 @@
...
+#
+# Can optionally include more than one location, causing data to be striped across
+# the locations (a la RAID 0) on a file level, favouring locations with most free
+# space on creation. For example:
```
staubesv: Omit "For example:" (it is an actual configuration setting, isn't it?)

modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java
```diff
@@ -444,111 +435,49 @@ private void createIndex(String idx) throws SearchIndexException, IOException {
...
-      }
-    }
-
-    // If no local settings were found, read them from the bundle resources
-    if (settings == null) {
-      String resourcePath = PathSupport
-              .concat(new String[] { "/elasticsearch/", index, "settings.json" });
-      try (InputStream is = this.getClass().getResourceAsStream(resourcePath)) {
+  private String loadResources(final String fileName) throws IOException {
+    final String resourcePath =  "/elasticsearch/" + fileName;
```
staubesv: file**n**ame

modules/search/src/main/java/org/opencastproject/matterhorn/search/impl/AbstractElasticsearchIndex.java
```diff
@@ -444,111 +435,49 @@ private void createIndex(String idx) throws SearchIndexException, IOException {
...
-    return settings;
+    return null;
   }
 
   /**
-   * Loads the mapping configuration. An initial attempt is made to get the configuration from
-   * <code>${opencast.home}/etc/index/&lt;index&gt;/&lt;type&gt;-mapping.json</code>. If this file can't be found, the
-   * default mapping loaded from the classpath.
+   * Loads the settings for the elastic search configuration. An initial attempt is made to get the configuration from
+   * <code>${opencast.home}/etc/index/&lt; index &gt;/settings.yml</code>.
```
staubesv: This PR removed /etc/index. Please adjust comment.

modules/search/src/test/java/org/opencastproject/matterhorn/search/impl/ElasticsearchUtils.java
```diff
@@ -42,29 +41,23 @@ private ElasticsearchUtils() {
 
   /**
    * Creates an elastic search index configuration inside the given directory by loading the relevant configuration
-   * files from the bundle. The final location will be <code>homeDirectory/etc/index</code>.
+   * files from the bundle. The final location will be <code>configDirectory/etc/index</code>.
```
staubesv: The location is not correct anymore

modules/search/src/test/java/org/opencastproject/matterhorn/search/impl/ElasticsearchUtils.java
```diff
@@ -42,29 +41,23 @@ private ElasticsearchUtils() {
...
+  public static void createIndexConfigurationAt(File configDirectory) throws IOException {
 
     // Load the index configuration and move it into place
-    File configurationRoot = new File(PathSupport.concat(new String[] { homeDirectory.getAbsolutePath(), index }));
-    FileUtils.deleteQuietly(configurationRoot);
-    if (!configurationRoot.mkdirs())
-      throw new IOException("Error creating " + configurationRoot);
-
-    String[] files = new String[] { "content-mapping.json", "names.txt", "settings.yml", "version-mapping.json" };
+    String[] files = new String[] { "content-mapping.json", "elasticsearch.yml", "version-mapping.json" };
```
staubesv, line 54: content-mapping.json?

lkiesow:
> […] if I wanted to run it as a separate service external to Opencast, is that possible with this new PR? 

Yes, you can. This configuration does not change at all:
https://docs.opencast.org/develop/admin/modules/searchindex/elasticsearch/

How this works is basically, that a connection to ES is established (that is still configurable) and then Opencast will initialize the index if it does not already exist. This works with both the internal and external Elasticsearch.

For the configuration, you now have:

- The connection configuration in `custom.properties`
- The ES node configuration in `etc/elasticsearch.yml`. This is what you would use to configure an external ES as well
- The index configuration which is tied to our code and is now in the search module's resources since you  should not change that without changing the code

lkiesow:
Fixed the space and re-based onto `develop`

staubesv (dismissed):
Just found some minor issues (see comments).

Please also adapt documentation and comments that mention the old index configuration path:

screen shot 2019-03-06 at 17 20 03

modules/search/src/test/java/org/opencastproject/matterhorn/search/impl/ElasticsearchUtils.java
lkiesow: I'm not sure what your question here is.

staubesv: The file does not exist. But anyway, it is not this PR that "introduced" it.

staubesv: Just realised that this is test code and there indeed is a content-mapping.json in the test resources... sorry for the noise.

lkiesow: Yes, it's a bit confusing that the utils classes are only used for tests.

staubesv (approved):

lkiesow:
Thanks for the review!
ES5 will follow soon.

PR #745, opened 2019-02-28
MH-13400, Admin Index Test Cleanup

lkiesow:
This patch removes some unused code from the admin interface index
tests.


modules/admin-ui-index/src/test/java/org/opencastproject/adminui/index/AdminUIElasticsearchUtils.java
```diff
@@ -60,16 +60,16 @@ public static String createIndexConfigurationAt(File homeDirectory, String index
     if (!configurationRoot.mkdirs())
       throw new IOException("Error creating " + configurationRoot);
 
-    String[] files = new String[] { "default-mapping.json", "event-mapping.json", "group-mapping.json", "names.txt",
```
staubesv, line 63: Is "names.txt" still in use after removing it here?

lkiesow: It's unnecessary but it's already removed as part of the Elasticsearch configuration patch. This patch is basically a follow up of some more or less unrelated stuff which is also not needed.

staubesv: Ok, thanks.

staubesv (approved):

PR #748, opened 2019-02-28
MH-13404, Improve Workspace Logging

lkiesow:
The workspace is spamming the logs with lines about what files it is
downloading which is not a very helpful or important information unless
the download fails. This drown other, important messages right now.

This patch reduces the log level to debug.


smarquard:
Looks good

PR #754, opened 2019-03-02
MH-13412, Deprecate PathSupport.concat(…)

lkiesow:
Most of Opencast's PathSupport class are re-implementations of of
functionality built into Java's standard library and it would make much
more sense to use the standard library instead.

This patch deprecates a number of PathSupprt methods, describing the
alternatives from the standard library.

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/util/PathSupport.java
+++ b/modules/common/src/main/java/org/opencastproject/util/PathSupport.java
@@ -38,7 +38,7 @@ public final class PathSupport {
   private static final Logger logger = LoggerFactory.getLogger(PathSupport.class);
 
   /**
-   * This class should not be instanciated, since it only provides static utility methods.
+   * This class should not be instantiated, since it only provides static utility methods.
    */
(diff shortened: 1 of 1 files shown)


gregorydlogan:
I'm 90% certain I was just advising @rute-santos to use these methods, sooo... I'll take this PR :-)

lkiesow:
Happy to have a reviewer ;-)

> I was just advising @rute-santos to use these methods

Any particular reason for using PathSupport over the built-ins? Pull request #755 contains an example for switching to the built-ins:
```diff
- File configFile = new File(PathSupport.concat(new String[] { uiConfigFolder, orgId, component, filename }));
+ File configFile = Paths.get(uiConfigFolder, orgId, component, filename).toFile();
```
Not only do I think that it is actually simpler to use but I'm also all for having less custom code.

gregorydlogan:
Apologies, I was thinking of UrlSupport.concat, not PathSupport.  I'm all for reducing custom code in favour of built-ins.

PR #756, opened 2019-03-04
MH-13414 extended metadata multivalue fields are not handled properly

geichelberger:
This Patch corrects the behaviour of multivalue extended metadata fields.
If non multivalue fields contain multiple values, the last value is used to not change the behaviour.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataCollection.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataCollection.java
@@ -29,6 +29,7 @@ import org.opencastproject.metadata.dublincore.EncodingSchemeUtils;
 import org.opencastproject.metadata.dublincore.MetadataField;
 
 import com.entwinemedia.fn.data.Opt;
+import com.google.common.collect.Iterables;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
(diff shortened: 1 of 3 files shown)


staubesv (changes requested):
To my understanding, only metadata fields of the types ITERABLE_TEXT, MIXED_TEXT and ORDERED_TEXT are supposed to be multi-value metadata fields.

It probably would make more sense to throw an exception if a client attempts to add a list of values to a single-value field instead of just taking the last value as this basically means the other values are silently ignored.

Or maybe at least log a warning...

What do you think about this?

geichelberger:
> To my understanding, only metadata fields of the types ITERABLE_TEXT, MIXED_TEXT and ORDERED_TEXT are supposed to be multi-value metadata fields.

Thats a problem of the data model. The DublinCoreCatalog returns a List of DublinCoreValues for each property. To map this to a a single-value field there is only the possibility to take the first or the last value. If the DublinCore property contains only one value, the first value in the list is also the last value.

> It probably would make more sense to throw an exception if a client attempts to add a list of values to a single-value field instead of just taking the last value as this basically means the other values are silently ignored.

At this point the data does not come from a user, but from a persisted MediaPackage.

My solution behaves exactly the same way for single-value fields as the original version.

staubesv:
> Thats a problem of the data model. The DublinCoreCatalog returns a List of DublinCoreValues for each property. To map this to a a single-value field there is only the possibility to take the first or the last value. If the DublinCore property contains only one value, the first value in the list is also the last value.

At the end of the day, it is a mismatch between the actual metadata and the configuration of the metadata fields. For whatever reasons the data source contains multiple values for a single-value field, this will cause data to be "lost" in means of being ignored which should not happen silently.

Don't get me wrong: This problem is not directly introduced with your PR. But now addField explicitly takes a list of values while implicitly knowing that this only works for specific metadata field types.

Could you please log a warning if addField is asked to put multiple values into a single-value field?
It is fine to just take the last value, but don't be silent about this as it implies a mismatch between data and data configuration.

geichelberger:
I got your point, and I've added the logging.

staubesv (approved):

staubesv:
@geichelberger Thanks for the patch!

PR #757, opened 2019-03-05
MH-13415: Timelinepreviews process first one only  

TurRil:
Documentation states:
In the engage player only the preview images of one video are shown (the first that is found), so to make sure the correct preview images are shown, better generate them only for one video.

### Flavors:
presenter/trimmed
presentation/trimmed
presentation2/trimmed

#### Running:
<configuration key="source-flavor">*/trimmed</configuration>

#### Output (3 flavors):
presenter/timeline+preview
presentation/timeline+preview
presentation2/timeline+preview

#### Running:
<configuration key="source-flavor">*/trimmed</configuration>
<configuration key="process-first-match-only">true</configuration>

#### Output (1 Flavor > which one depends on input flavors):
presenter/timeline+preview

or
presentation/timeline+preview

or
presentation2/timeline+preview

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/timelinepreviews-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/timelinepreviews-woh.md
@@ -19,6 +19,7 @@ Parameter Table
 |target-flavor|\*/timeline+preview|Specifies the flavor the new files will get. This should use the \*-operator if it was used in the source-flavor too. This flavor has to contain the words "timeline" and "preview" for the file to be found by the player.|EMPTY|
 |target-tags|engage-download|Specifies the tags the new files will get.|EMPTY|
 |image-count|100|Specifies the number of generated timeline preview images. In the example 100 timeline preview images will be generated and stored in a 10x10 grid in the output image|100|
+|process-first-match-only|true|Only use the first resource that matches the source-flavor and/or target-tag|false|
 
 Operation Example
 -----------------
@@ -32,6 +33,8 @@ Operation Example
     <configuration key="target-flavor">*/timeline+preview</configuration>
     <configuration key="target-tags">engage-download</configuration>
     <configuration key="image-count">100</configuration>
+    <!-- If there is more than one file that match the source-tags, use only the first one -->
+    <configuration key="process-first-match-only">true</configuration>
   </configurations>
 </operation>
 ```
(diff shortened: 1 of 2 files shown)


gregorydlogan:
How is the order determined?  Raw mediapackage ordering?  I can see end users being very confused in terms of how to change their ordering as this PR stands right now...

TurRil:
Hi @gregorydlogan,
I'm not 100% sure - I think the order is determined by the track id. Track-0 being before Track-1. So as it is added to the TrackSelector that order should be preserved.

What I would like to have is something more akin to the prepare-av workflow handler where the parameter can be set to:

`presenter/trimmed, presentation/trimmed, */trimmed`

"going from left to right through this comma-separated list of source flavors"

Not entirely sure how to do this and where to place the code:
a) Add the method to _TimelinePreviewsWorkflowOperationHandler.java_

That looks very similar to _findAudioTrack_ (PrepareAVWorkflowOperationHandler.java [line 403]) and caling it like _findVideoTrack_ then calling it in _start_ [line 138]

b) Add a method to _AbstractMediaPackageElementSelector.java_ to be able to handle a list of flavors in a comma separated list.

But this commit is a simple first step to just do less processing and see what the results might be in order of the flavor order and then figure out the next bit.

Paths:
- modules/timelinepreviews-workflowoperation/src/main/java/org/opencastproject/workflow/handler/timelinepreviews/TimelinePreviewsWorkflowOperationHandler.java
- modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/PrepareAVWorkflowOperationHandler.java
- modules/common/src/main/java/org/opencastproject/mediapackage/selector/AbstractMediaPackageElementSelector.java

gregorydlogan:
I would be tempted to add that code, and its filtering, to the step you're already working on.  Add the filtering list as a parameter, then filter/sort the list at https://github.com/opencast/opencast/pull/757/files#diff-95ef405635a64bcd4e93769a41a324c9R195.  If the filtering parameter is set then process the first one, otherwise process all of them.

KatrinIhler:
@TurRil This branch has conflicts now!

KatrinIhler:
@TurRil Is this still a work in progress or ready for review?

KatrinIhler:
@TurRil If this is not ready, it might make more sense to close this PR now and refile once it's done.

gregorydlogan:
@TurRil any ETA for further development here?

TurRil:
I think this can be reviewed.

PR #759, opened 2019-03-06
MH-13420 ngRepeat does not allow duplicates

geichelberger:
This patch allows duplicate values to be displayed in multivalue fields if they are present in the data source.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableMultiSelect.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/editableMultiSelect.html
@@ -1,6 +1,6 @@
 <div ng-click="enterEditMode()" ng-form="innerForm">
   &nbsp;<ul ng-show="!editMode">
-    <li ng-repeat="value in params.value">
+    <li ng-repeat="value in params.value track by $index">
       <span ng-if="value">{{ getText(value).value | translate }}</span>
     </li>
   </ul>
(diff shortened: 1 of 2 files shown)


KatrinIhler:
Is there a use case we need this for?

geichelberger:
The data model allow duplicates and with the external api it is possible to add duplicate contributor entries for example.

If there are duplicates for any reason, they should be displayed. At least to be able to delete them.

KatrinIhler:
@geichelberger Okay, I can see that, but maybe it would make more sense to not allow to create duplicate entries in the first place?

geichelberger:
@KatrinIhler Perhaps someone has a use case for duplicates, this concerns not only the normal metadata but also the extended ones. Why should the API forbid something if it is supported by the data model? Furthermore, this would be a change that is not backward compatible.

I think the reason why no duplicates are allowed in the Admin-UI is because the value of an entry was used as identifier and not the index. (function removeValue)

See Pull-Request MH-1323 commit: https://github.com/opencast/opencast/pull/564/commits/e8b6ac07f15ae389959dc7fb643993348025f764

Maybe we should allow to enter duplicates in the Admin-UI?

KatrinIhler:
@geichelberger Okay, I see your point. And yes, in that case we should probably allow multiple values in the Admin-UI as well to be consistent.

KatrinIhler:
One more thing: Presenters are also not displayed at all in the admin ui table if there are duplicate values.

geichelberger:
You mean Event/Presenters? With the patched version it works for me.

KatrinIhler:
It's not working for me: 
image

I create duplicate presenters by editing the dublincore in the archive and then rebuilding the Admin UI index, no idea if that's affecting things...once I delete the duplicates, the table shows the presenters just fine.

geichelberger:
Should be fixed now. The easiest way to create duplicate entries is to resend a manipulated event update request.

KatrinIhler:
Still not working for me, I think you'd actually have to update EventsPresentersCell.html or something to fix this in the main table. Anyway, we could consider this a separate issue and this improves the current state, so I'm gonna merge this as is.

KatrinIhler (approved):

PR #761, opened 2019-03-07
MH-13423 Possible NPE if debugging is enabled

geichelberger:
If debugging is enabled and the media package contains no elements to distribute, an NPE is thrown.


lkiesow:
Thanks for the patch. Looks good. Merged.

PR #762, opened 2019-03-07
MH-13424, Elasticsearch 5.6.15

lkiesow:
This patch updates Opencast's Elasticsearch compatibility and the
internal Elasticsearch node to version 5.6.15, finally btinging us to a
recent (Feb 19, 2019), still supported version.

### DIFF
--- a/modules/search/pom.xml
+++ b/modules/search/pom.xml
@@ -13,7 +13,7 @@
   <properties>
     <opencast.basedir>${project.basedir}/../..</opencast.basedir>
     <checkstyle.skip>false</checkstyle.skip>
-    <lucene.version>5.5.4</lucene.version>
+    <lucene.version>6.6.1</lucene.version>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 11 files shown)


miesgre:
@lkiesow Why elasticsearch 5 and not elasticsearch 6?
https://www.elastic.co/support/eol

lkiesow:
Elasticsearch 6 is still planned but it may not be feasible to integrate it in the same manner.
It also does not make sense to wait until I finish the 6.x upgrade when 5 is already ready to go.
As announced earlier, the plan is to go 1.6.x → 2.4.x → 5.6.x → *6.x*.
There are enough changes already when going these small steps.

modules/search/pom.xml
```diff
@@ -39,9 +39,44 @@
...
       <groupId>org.elasticsearch</groupId>
       <artifactId>elasticsearch</artifactId>
-      <version>2.4.6</version>
+      <version>5.6.15</version>
     </dependency>
-    <!-- elasticsearch dependencies -->
+    <dependency>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>transport</artifactId>
+      <version>5.6.15</version>
```
staubesv: Is there a special reason for listing the version many times vs. using a variable like ${elasticsearch.version} similar to ${lucene.version}?

staubesv (dismissed):
Just the question about using a variable for the Elasticsearch version, besides that looks good to me.

staubesv (approved):

PR #763, opened 2019-03-08
MH-13381, Use Organization Identifier in Workflows

lkiesow:
The maximum size of the serialized workflow instances stored in the
database is huge, causing an unnecessary performance overhead for the
frequent serialization and deserialization to/from XML. One cause for
this is the large organization data stored multiple times in the XML.

This patch drops the use of the serialized organization which is not
frequently being used anyway and replaces it by referencing the
organization via its unique identifier.

The old organization is still kept as an optional field, ensuring
backward compatibility to old serialized workflows. This means that no
migration is required.

There are more references to organizations in the XML caused by the user
data stored in the XML, but they will be dealt with separately.

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
@@ -37,6 +37,9 @@ import org.opencastproject.workflow.api.WorkflowOperationInstance.OperationState
 
 import com.entwinemedia.fn.Fn;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -61,6 +64,8 @@ import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter;
 @XmlRootElement(name = "workflow", namespace = "http://workflow.opencastproject.org")
 @XmlAccessorType(XmlAccessType.NONE)
 public class WorkflowInstanceImpl implements WorkflowInstance {
+
+  private final Logger logger = LoggerFactory.getLogger(WorkflowInstance.class);
 
   @XmlAttribute()
   private long id;
@@ -87,6 +92,9 @@ public class WorkflowInstanceImpl implements WorkflowInstance {
   @XmlJavaTypeAdapter(OrganizationAdapter.class)
   @XmlElement(name = "organization", namespace = "http://org.opencastproject.security")
   private JaxbOrganization organization;
+
+  @XmlElement(name = "organizationid", namespace = "http://org.opencastproject.security")
+  private String organizationId;
 
   @XmlElement(name = "mediapackage", namespace = "http://mediapackage.opencastproject.org")
   private MediaPackage mediaPackage;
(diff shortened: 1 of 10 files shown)


smarquard:
Confirmed by looking at payload in oc_job that the workflow serialized XML now has:

    <sec:organizationid>mh_default_org</sec:organizationid>

rather than

    <sec:organization id="mh_default_org"> ... (etc) ... </sec:organization>

Still to do some basic regression testing, but looks good.

A similar change for the sec:creator user info will be a big win as well.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
```diff
@@ -202,8 +210,15 @@ public User getCreator() {
    * @see org.opencastproject.workflow.api.WorkflowInstance#getOrganization()
    */
   @Override
-  public Organization getOrganization() {
-    return organization;
+  public String getOrganization() {
```
staubesv: getOrganizationId() would be more consistent with other similar situations in the codecase

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
```diff
@@ -217,14 +232,11 @@ public void setCreator(User creator) {
...
-   *          the organization
+   * @param organizationId
+   *          the organization identifier
    */
-  public void setOrganization(Organization organization) {
-    if (organization == null)
-      this.organization = null;
-    else
-      this.organization = JaxbOrganization.fromOrganization(organization);
+  public void setOrganization(final String organizationId) {
```
staubesv: setOrganizationId() would be more consistent with other similar situations in the codecase

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2387,7 +2387,13 @@ public void repopulate(final String indexName) throws Exception {
           logger.warn("Skipping restoring of workflow. Error parsing: {}", workflow, e);
           continue;
         }
-        Organization organization = instance.getOrganization();
+        Organization organization = null;
+        try {
+          organization = organizationDirectoryService.getOrganization(instance.getOrganization());
```
staubesv: ... getOrganization() expects the ID of the organization. While instance.getOrganization() now indeed does return the ID, instace.getOrganizationId() would make this more obvious.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
```diff
@@ -88,6 +93,9 @@
   @XmlElement(name = "organization", namespace = "http://org.opencastproject.security")
```
staubesv: That likely can be removed...

staubesv (dismissed):
I've found a minor issue:

- The setter/getter method set-/getOrganization() should be renamed to set-/getOrganizationId()

[Update] Just found the serialized workflow instance in the DB....

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstanceImpl.java
```diff
@@ -88,6 +93,9 @@
...
   private JaxbOrganization organization;
 
+  @XmlElement(name = "organizationid", namespace = "http://org.opencastproject.security")
```
staubesv: Maybe organization_id or organizationId for sake of readability?

lkiesow: I changed it to `organization-id` to be consistent with other XML tags we use like for example `time-in-queue`

staubesv (approved):

PR #768, opened 2019-03-08
MH-13430, Fix Opencast Offline Builds

lkiesow:
The SMIL service tests required an internet connection to verify that
the test XML document is valid, causing the whole build to fail in an
offline environment.

### DIFF
--- a/modules/smil-impl/src/test/java/org/opencastproject/smil/impl/SmilServiceImplTest.java
+++ b/modules/smil-impl/src/test/java/org/opencastproject/smil/impl/SmilServiceImplTest.java
@@ -66,8 +66,7 @@ public class SmilServiceImplTest {
   /**
    * Test SMIL document
    */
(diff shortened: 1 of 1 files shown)


lkiesow:
To test this:

- build Opencast
- disconnect your machine from the internet and build again (the build will fail)
- apply the patch, stay offline and build again. It should now work

staubesv:
@lkiesow Hmm... not in the focus of this PR, but just wondering: Is it just the tests that fail without Internet connection or would a productive Opencast system also want to download the DTD to validate the XML document?

staubesv (approved):

lkiesow:
If in production the XML would contain a dtd then it would try to access the specs. At least the parser the smil service uses would do that.

staubesv:
Hmm... I've created https://opencast.jira.com/browse/MH-13435 as I don't think it is a good idea to validate the XML based on external resources.

lkiesow:
Are there references to external resources in the production smile XML files?

staubesv:
Good question. Will check.

PR #770, opened 2019-03-12
MH-13432 Remove unused modals "Job Details" and "Server Details"

staubesv:
The modals "Job details" and "Server details" are just mockups that are not in use.

The goal of this task is to remove them from the Opencast codebase.


lkiesow:
These are actually still referenced in the code though I guess that the references can just be removed as well:

```
.../admin-ui% grep -rn server-details-modal src
src/main/webapp/styles/views/modals/_modal-dialog.scss:26:    &#server-details-modal {
src/main/webapp/scripts/modules/systems/partials/serverActionsCell.html:1:<a data-open-modal="server-details-modal"
                                          
.../admin-ui% grep -rn job-details-modal src
src/main/webapp/scripts/modules/systems/partials/jobActionsCell.html:2:  data-open-modal="job-details-modal"
```

staubesv:
Good catch! Will update the PR tomorrow.

staubesv:
@lkiesow PR updated.

PR #771, opened 2019-03-13
MH-13381, Use Organization Identifier In Roles

lkiesow:
…instead of storing the whole organization and all its properties itself
since that causes a huge overhead when it comes to serialization of
users (and thus also workflows).

This patch uses a similar method to pull request #763 which means that this again does not need any migration of old data. They can still be used in their old form but new ones will not any longer include the whole organization.

Serialized roles now look like this (excerpt from a workflow instance XML):
```xml
<wf:workflow …>
  <wf:template>fast</wf:template>
  …
  <sec:creator>
    <sec:username>opencast_system_account</sec:username>
    <sec:name>System User</sec:name>
    <sec:provider>system</sec:provider>
    <sec:manageable>false</sec:manageable>
    <sec:roles>
      <sec:role>
        <sec:name>ROLE_ADMIN</sec:name>
        <sec:organizationid>mh_default_org</sec:organizationid>
        <sec:type>INTERNAL</sec:type>
      </sec:role>
      <sec:role>
        <sec:name>ROLE_SUDO</sec:name>
        <sec:organizationid>mh_default_org</sec:organizationid>
        <sec:type>INTERNAL</sec:type>
      </sec:role>
    </sec:roles>
    …
```


modules/common/src/main/java/org/opencastproject/security/api/JaxbRole.java
```diff
@@ -45,10 +47,12 @@
   @XmlElement(name = "description")
   protected String description;
 
-  /** The description */
   @XmlElement(name = "organization")
   protected JaxbOrganization organization;
 
+  @XmlElement(name = "organizationid")
```
staubesv: Maybe organization_id or organizationId for sake of readability?

smarquard: We have several precedents to choose from :-)

```
@XmlElement(name = "warning_state_trigger")
@XmlElement(name = "processingHost")
@XmlElement(name = "samplingrate")
```

but of these, camelCase seems to be the more common style used in our code base so far.

smarquard:
I've tested this and PR #763 by merging both into our 6.x dev system and they have the desired effect and do not seem to have broken anything.

modules/common/src/main/java/org/opencastproject/security/api/JaxbRole.java
lkiesow: In #763, I've switched to `organization-id` because of `execution-history` and `time-in-queue`, `fail-on-error`, `retry-strategy`, … it seemed to me that `-` was commonly used ;-D

I tend to use it here as well unless someone wants something else in particular.

staubesv: For me, organization-id is fine.

smarquard: Sounds good.

staubesv: @lkiesow #763 is merged, so let's go for organization-id ...

modules/common/src/main/java/org/opencastproject/security/api/JaxbRole.java
```diff
@@ -134,11 +157,15 @@ public String getDescription() {
 
   /**
    * {@inheritDoc}
-   *
-   * @see org.opencastproject.security.api.Role#getOrganization()
    */
-  public Organization getOrganization() {
-    return organization;
+  public String getOrganizationId() {
```
staubesv, line 160: Matter of taste thing: Personally, I would write

```
if (organizationId != null) {
  return organizationId;
} else if (organization != null) {
  return organization.getId();
} else {
  return null;
}
```

... but fine for me if you prefer the version without else if.

staubesv (dismissed):
Please rename organizationid to organization-id as discussed. Besides that, LGTM.

modules/common/src/main/java/org/opencastproject/security/api/JaxbRole.java
lkiesow: I'll keep the current version.

staubesv (approved):

PR #773, opened 2019-03-14
Make button `Edit on Github` link works on documentation page

miesgre:
The documentation page has a link `Edit on Github`, but the link does not work because is linking to an incorrect page.

This PR fix that. This allow people reading the documentation to fix errors easily.

### DIFF
--- a/docs/guides/admin/mkdocs.yml
+++ b/docs/guides/admin/mkdocs.yml
@@ -1,5 +1,6 @@
 site_name: Opencast - Administration Guide
-repo_url: https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs
+repo_url: https://github.com/opencast/opencast
+edit_uri: edit/r/6.x/docs/guides/admin/docs
 
 # Default theme used is readthedocs
 theme: readthedocs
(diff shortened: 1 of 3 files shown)


KatrinIhler:
This would work for 6.x, but would break if it's merged ahead into develop. Additionally, the URL would have to be updated for each new release branch. I wonder if there is a more permanent solution for this?

miesgre:
@KatrinIhler Yes, this works for `r/6.x`branch.. for develop we need to change the URL.

I can make a PR for develop also. Then each time a new branch is cut the URL needs to be modified and this can be done with a `sed` command.
Now, all pom.xml files are modified each time we make a new release.

KatrinIhler:
The link actually already works for develop, it was fixed a while ago.
And good point...

lkiesow:
**-1** to this. I'm strongly against introducing anything that needs manual intervention for every version change. I'm fine with just linking to develop (that's also already in the develop branch) to update new docs. Linking specific version could also cause people to create pull requests against inactive branches.

In any case, if you want specific link, please propose a fully automated process (e.g. solve it in JS).

miesgre:
For me it is ok to link to develop version. So, I'm closing this PR.

PR #774, opened 2019-03-14
MH-12139 Admin UI upload progress bar for event create with media upload

karendolan:
This pull causes a new event media upload to show a prominent yellow progress bar containing the percentage of upload across the top of the Admin UI, below the main banner.

The motivation is to let the user know what is going on the background when a large file upload is started and an indication of when it will be complete. Showing this information helps to stop the user from accidentally closing the browser, or starting another upload, or anything that will jeopardize the current large file upload. 

This pull adds a new resource, chieffancypants angular-loading-bar tool. There is already a chieffancypants hot key tool with the same license. The angular-loading-bar is included in the same way the first tool was included (a checked in bower_component). 

This pull also changes the behavior of the event success/fail notification, in that it does not automatically remove the notification after a few seconds, and it gives the notification a global context so the user can see it if the user has moved onto other tabs while the event was still being processed. 

Screen Shot 2019-03-14 at 5 33 13 PM


karendolan:
@staubesv I'm going to move that onUploadFileProgress method and it's baggage (translate) out of the newEventResource into a service to keep newEventResource clean(er) and to reuse the same service for the asset upload resource. I'll also add some tests. I notice the travis test fail is related to abstract test scanner in the common module. Not sure what's going on there.

staubesv:
@karendolan Ok, please let me know once I should start the review. The test failure is not related to this PR.

lkiesow:
> The test failure is not related to this PR.

Yes, it is. This pull request breaks the style checks.

staubesv:
@lkiesow The first build of this PR failed because of a unit test of the module "common" was failing. That was not related to this PR. I've then restarted the build and now it fails indeed because of the PR ;-)

karendolan:
@staubesv  The pull is updated and should be good to go. 
 
Here's a screen capture that shows the progress bar when partially across the screen. I tried to use the recommended Opencast yellows, but please feel free to suggest variations. The prominence of the progress bar was requested by our users at our site because they often do something else after starting a big upload and want other people, and themself, to know not to close the screen while it's still progressing. We use even brighter caution "yellow" in our production.
Screen Shot 2019-03-15 at 11 52 36 AM

karendolan:
Sorry, I'll fix the eslint!

modules/admin-ui/src/main/webapp/login.html
```diff
@@ -78,6 +78,7 @@
     <script src="../../bower_components/angular-local-storage/dist/angular-local-storage.js"></script>
     <script src="../../bower_components/angular-wizard/dist/angular-wizard.js"></script>
     <script src="../../bower_components/angular-hotkeys/build/hotkeys.js"></script>
+    <script src="../../bower_components/angular-loading-bar/build/loading-bar.js"></script>
```
staubesv, line 81: Why would you want this in the login page? Mis-injected by grunt-wiredep?

staubesv:
@karendolan Please adapt NOTICES

modules/admin-ui/src/main/webapp/login.html
karendolan: @staubesv it was injected by the auto-wiring. I'll remove it. 
[EDIT] @staubesv wow, I can't remove it, it's added back in during each mvn build. The Grunt file wires the bower dependencies into html files that take in bower dependencies, automatically. In theory, this could be moved to the package.json file. But, it feels more natural to pull it in the same way as other similar component via the bower_components area.

lkiesow:
The design looks a bit weird to me. It feels like going back to the flashy warning colors of the old interface and my first thought seeing this was that it looks like Opencast is telling me that something very important is broken right now and I have to intervene immediately before it's too late ;-)

I'm also not entirely sure about the positioning in the header. It feels a bit weird to have these elements above the top navigation.

I've just loaded your image into Gimp and came up with a few other suggestons (please ignore the poor quality, I can make it look nice if you like any of them):

54444458-20b50700-4719-11e9-808a-79cd8f1a5b3a

The first idea was to have a somewhat more natually looking bar at the top. The colors are less flashy and I specifically decided to not go with white since it seemed like it would just device the top and the bottom half of the header. At 100% both header parts would then be completely separated.

The secons bar is what I would prefer. It reuses some of the colors used in the admin interface (note that I wasn't very precise) but is clearly separated from the header. You could also put more information there like estimated time, … and you could potentially deal with parallel ingests (something we might look into soon).

I'm also wondering if we shouldn't remove the green notification.

karendolan:
@lkiesow  The easiest compromise is to adopt your first UI, that "subtle" top bar, and add  a configurable  [bold|subtle] option. The "subtle" would be the thin quiet bar. The "bold" would be the thicker bright in-your-face yellow progress bar. 

The original requirement at our site, was that too many people are loosing their mega-multiGB uploads and needed something visually prominent to let them know all was good. Thus,  the big bold yellow "everything is going Ok, but not out of danger yet!" progress bar.

A benefit of having the progress bar just below the header is that the progress bar has a constant position when changing the view to the series, or system, or any other section, while your upload is still uploading. Users can continue working but still see that the upload is OK. Without the bar jumping around, disappearing, or overlaying something important.

staubesv:
@karendolan Note that this is a feature and therefore should go to develop.

karendolan:
Moving this from 6x to develop target https://github.com/opencast/opencast/pull/784

karendolan:
This pull is rebased against develop branch. It's the same changes, except the CSS is more subtle and the NOTIFICATIONS file is updated. PULL: https://github.com/opencast/opencast/pull/784/files

PR #775, opened 2019-03-15
MH-13434, Opencast Common Clean-up

lkiesow:
This patch removes some unused code from opencast-common.


smarquard:
Looks good to me except for the test failure:

[ERROR] Failures: 
[ERROR]   EqualUtilTest.testHash:43 expected:<126145> but was:<236095>

lkiesow:
Ups. Missed pushing that. Sorry.

smarquard:
The CaptureParameters time definitions (MILLISECONDS, SECONDS, MINUTES, HOURS) could go in favour of TimeUnit references.

staubesv:
@lkiesow Note that some capture agents do set values for some of the configuration keys you remove here. For example, Extorn SMPs set capture.schedule.remote.polling.interval and capture.agent.name.

Just mentioning...

lkiesow:
And they can continue to do so if they like to. But Opencast does not use these Constants anywhere and hasn't before this patch.

smarquard:
Same issue described in https://opencast.jira.com/browse/MH-12654

It doesn't really matter if Opencast knows about a property or not; it can still be reported and retrieved. It would be helpful to have a controlled vocabulary in common across CAs (like reporting free disk space) but we don't at present.

Having these in the code promoted bad behaviour in CAs which set them regardless for some notion of compatibility with the original reference CA, e.g.

https://github.com/teltek/Galicaster/issues/572

staubesv:
Makes sense.

staubesv (approved):

PR #776, opened 2019-03-15
MH-13421 Remove unused workflowservice exceptions

KatrinIhler:

PR #777, opened 2019-03-15
MH-13387 Get ACLs of finished workflows from AssetManager

KatrinIhler:
Deleting an event as a Non-Admin can fail if there are finished workflows to remove that were started by another user, since in that case the ACL will be looked for in the WFR, where it might no longer be. This PR gets the ACL for workflows that are no longer running from the AssetManager.

This also contains PR #776.


staubesv:
@KatrinIhler Hmm.... shouldn't this go to 6.x?

KatrinIhler:
@staubesv You're right, I made a wrong assumption. Fixed and tested!

PR #778, opened 2019-03-15
MH-13436: Improve error message for out of bounds image extraction

JulianKniephoff:
The error message for when you try to extract an image from a time
past the end of the video is confusing because of the lack of explicit units
and also inconsistency of the units implied.

Example:

```
org.opencastproject.composer.api.EncoderException: Can not extract an image at time 1.0 f
rom a track with duration 520
```

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -98,6 +98,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -1085,14 +1086,22 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
         throw new EncoderException("Unable to extract an image from a track with unknown duration");
       }
       if (time < 0 || time * 1000 > sourceTrack.getDuration()) {
+        Duration duration = Duration.ofMillis(sourceTrack.getDuration());
+        long durationSeconds = duration.getSeconds();
+        long durationMillis = duration.minusSeconds(durationSeconds).toMillis();
+        String formattedDuration = String.valueOf(durationSeconds) + '.'
+                + StringUtils.leftPad(String.valueOf(durationMillis), 3, '0');
+
         Map<String, String> params = new HashMap<>();
         params.put("track-id", sourceTrack.getIdentifier());
         params.put("track-url", sourceTrack.getURI().toString());
-        params.put("track-duration", sourceTrack.getDuration().toString());
+        params.put("track-duration", formattedDuration);
         params.put("time", Double.toString(time));
         incident().recordFailure(job, IMAGE_EXTRACTION_TIME_OUTSIDE_DURATION, params);
-        throw new EncoderException("Can not extract an image at time " + time + " from a track with duration "
-                + sourceTrack.getDuration());
+
+        throw new EncoderException("An image could not be extracted from the track " + sourceTrack.getURI()
+                + " with id " + sourceTrack.getIdentifier() + " because the extraction time (" + time + " second(s)) is "
+                + "outside of the track's duration (" + formattedDuration + " second(s))");
       }
     }
 
--- a/modules/composer-ffmpeg/src/main/resources/incident-l10n/org.opencastproject.composer.properties
+++ b/modules/composer-ffmpeg/src/main/resources/incident-l10n/org.opencastproject.composer.properties
@@ -93,7 +93,7 @@
 #  track-duration The source track duration
 #  time The extraction time
 17.title = Image extraction failed - time outside duration
-17.description = An image could not be extracted from the track '\#{track-url}' with id '\#{track-id}' because the extraction time '\#{time}' is outside of the track's duration '\#{track-duration}'.
+17.description = An image could not be extracted from the track '\#{track-url}' with id '\#{track-id}' because the extraction time (\#{time} second(s)) is outside of the track's duration (\#{track-duration} second(s)).
 # params:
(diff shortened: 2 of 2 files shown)


modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -1091,8 +1091,18 @@ public Job image(Track sourceTrack, String profileId, Map<String, String> proper
         params.put("track-duration", sourceTrack.getDuration().toString());
         params.put("time", Double.toString(time));
         incident().recordFailure(job, IMAGE_EXTRACTION_TIME_OUTSIDE_DURATION, params);
-        throw new EncoderException("Can not extract an image at time " + time + " from a track with duration "
-                + sourceTrack.getDuration());
+
+        // Format duration (a `long`, in milliseconds) as seconds, with a fractional part.
+        // Why not just cast to `double` and divide by 1000? Because of precision loss,
+        // both during the cast and the division.
+        String duration = Long.toString(sourceTrack.getDuration());
```
staubesv: Please remove the comment and replace the code by something like:

```
Duration duration = Duration.ofMillis(sourceTrack.getDuration());
throw new EncoderException("Cannot extract image at position " + time + " second(s) from a track with the duration "
                + duration.getSeconds() + "." + duration.minusSeconds(duration.getSeconds()).toMillis() + "second(s)");
```

Admittedly still quite ugly as Java 8 does not yet support the toMillisPart() method introduced later in Java 9.

Also: This is just the exception. It might be preferable to (also) adjust the incident IMAGE_EXTRACTION_UNKNOWN_DURATION

[Update] Admittedly, I've forgot to format the fraction with leading zeros if necessary ;-) That could be done using String.format().

staubesv (changes requested):
I think that can be solved "better" using the Java 8 Duration class. Also, it would be nice to improve the quality of the incident message.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
JulianKniephoff: @staubesv thanks for the chance to finally take a look at the "new" time API :smile: I fixed that part and added 0-padding, which I also forgot. :wink:

As for the incident message: I'm not quite sure what you mean. First, you probably mean `IMAGE_EXTRACTION_TIME_OUTSIDE_DURATION`, not `IMAGE_EXTRACTION_UNKNOWN_DURATION`, as the latter is for a different error, right? But even then: `IMAGE_EXTRACTION_TIME_OUTSIDE_DURATION` ist just a number (`17` to be precise). Where can I find the message you are talking about?

staubesv: @JulianKniephoff I still don't get what it means to  "Extract an image [...] into a second track":

> Can not extract an image from " + time + " second(s) into a " + formattedDuration + "second track

I think that should be something like: "Cannot extract image at position " + time + " second(s) " from a track with duration " + formattedDuration

The incident is supposed to be the user-readable error message that will be displayed in the Admin UI. You can find the incident messages for the composer service in https://github.com/opencast/opencast/tree/develop/modules/composer-ffmpeg/src/main/resources/incident-l10n

JulianKniephoff: The message is not "extract an image [...] into a second track", but something like "extract an image from 6 seconds into a 5 second track."

I'm no native speaker, of course, but as far as I know "n <units> into an m <unit> thing" is a perfectly valid phrasing of the two facts that the thing is m units long and that we are talking about the point in time n units from the start of the thing.

If you insist, though, I will change it of course. :wink:

The incident message should now get the formatted duration as well.

staubesv: @JulianKniephoff What confuses me most about the message is how can you extract an image **into** a track? It is extracted **from** a track, isn't it?

JulianKniephoff: It is extracted **from** some time into the track. The "into" belongs to "some time" in this case, not to the verb. "n seconds into the track" specifies a point in time, relative to the start of the track.

staubesv: @JulianKniephoff I'm also not a native speaker, but I don't think the current wording makes too much sense.

@karendolan As you are a native speaker, could you please have a quick look at the error message under discussion and suggest an understandable and correct wording? 

The error message currently is:

> Can not extract an image from " + time + " second(s) into a " + formattedDuration + " second track"

whereas _formattedDuration_ (in seconds) is the duration of the track where the image is extracted at position _time_ (in seconds)

JulianKniephoff: Okay so apparently Google Groups thinks I'm a spam bot, so I can't answer on list right now, unfortunately.

Here is what I drafted (slightly adapted to GitHub)

> Oh no, this got out of control, I'm sorry! ^.^"
>
> @smarquard is right, an error message should be clear to the users who have to read and act on it in the end, regardless of whether it is correct and/or natural. Seeing as this made it to list this is not the case here, so it was a bad choice on my part. I will change it.
>
> But I have to say, @staubesv, it feels good to be backed up by native speakers on this (thanks, @JamesUoM!), so I guess in the end everyone is happy. ;D

I took a hint from the incident message now, which is very clear in my opinion, especially after adding units to it as well.

Can we merge this now? :see_no_evil: :hear_no_evil: :speak_no_evil:

staubesv: @JulianKniephoff Well... seems I need some English lessons ;-)

I'm on it.

staubesv (approved):

PR #779, opened 2019-03-15
MH-13425, feed links

eukaiser:
I added the feedlinks to the series modal view.

### DIFF
--- a/modules/admin-ui/package-lock.json
+++ b/modules/admin-ui/package-lock.json
@@ -2281,12 +2281,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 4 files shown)


staubesv:
@eukaiser ... but your code seems to assume that those feeds are always available which is not necessarily the case as not every Opencast Adopter offers feeds. Can you please somehow address that issue?

staubesv:
@eukaiser Also the URLs are not really metadata and should probably not just be appended to the actual metadata of the series to not confuse different data sets.

Maybe a new tab "Links" that lists configurable links (e.g. using a list provider) and a basic template replacement mechanism ({{seriesId}}-> actual series Id)?

lkiesow:
> […] your code seems to assume that those feeds are always available which is not necessarily the case as not every Opencast Adopter offers feeds. Can you please somehow address that issue?

@staubesv, you are right. I've checked and one would probably need to check the list of registered feed generators for an instance of a `SeriesFeedService` and check which pattern it registered. If there is one, that's what can be used to display the links

https://github.com/opencast/opencast/blob/afa95263ae8cb9a8ca3f23e668bfc65021069ff2/modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java#L91-L92

> (e.g. using a list provider) and a basic template replacement mechanism ({{seriesId}}-> actual series Id)?

That would work as well but since we can actually ask Opencast if series feeds are available, I suggest we do that. Not sure if you want to create a list provider for that or a separate facade endpoint though.

lkiesow:
Thinking about it again, creating the connection is actually not that easy since the feed service would usually be on a different server than admin. So, maybe a list provider is a good idea ;-D

lkiesow:
declining this for now. we will open a new one.

PR #782, opened 2019-03-17
MH-13439, Dynamic Player Redirect

lkiesow:
This patch adds a `/watch/<id>` endpoint as a universal place for
Opencast players, regardless of which specific player is configured to
be used right now. The endpoint will redirect users to the correct place
using an HTTP redirect.

Furthermore, this patch integrates the new endpoint into all relevant
Opencast components like the publication channels and the engage tools.

Note that this behavior is similar to the JavaScript redirect included
in `/engage/ui/watch.html` in commit aac56d7 which was later partially
broken by commit 23dbfa9 causing the direct player links to be written
into the media packages publication elements directly.

But unlike the old mechanism, the new endpoint does not require the web
browser to render the page before issuing the redirect or even load any
additional resources.

To sum up what this patch does:

- Add a new player redirection endpoint

- Use this endpoint wherever a player link is required to unify the
  behavior

- Update the player settings in the organization configuration file

- Deprecate the old `/engage/ui/watch.html`

- Provide instructions to migrate all old links if desired

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -14,7 +14,8 @@ How to Upgrade
 5. [Upgrade the ActiveMQ configuration](#activemq-migration)
 6. Review the [configuration changes](#configuration-changes) and adjust your configuration accordingly
 7. Migrate the scheduled events
-8. Rebuild the Elastic Search indexes
+8. [Optionally update the player links](#update-player-links)
+9. Rebuild the Elastic Search indexes
 
 Database Migration
 ------------------
@@ -82,6 +83,57 @@ change `maintenance` back to `false` to put the scheduler back into its normal m
 You should avoid running Opencast 7 without migrating the scheduled events first. Otherwise, your capture agents may
 fetch an empty calendar.
 
+
+Update Player Links
+-------------------
+
+> This step is optional
+
+Opencast 7 comes with the capability of dynamically switching the configured player without requiring the republication
+of all published material by providing the dynamic target `https://example.opencast.org/watch/<id>`. This new target
+will be used for all new engage publications.
+
+But old publications still reference the players directly in the admin interface and the external API. This does not
+pose any immediate problem unless you actually want to switch players in which case you would need to republish the
+material once to update the publication links.
+
+Alternatively, you can rewrite the old links all at once without re-publication using the following method:
+
+1. Find the archive directory and run the following command (replacing `<playerlink>`):
+
+        sed -i 's_<playerlink>_/watch/_g' .../archive/*/*/*/manifest.xml
+
+    For Theodul, Paella and the Engage player the specific commands would be:
+
+        sed -i 's_/engage/theodul/ui/core.html?id=_/watch/_g' .../archive/*/*/*/manifest.xml
+        sed -i 's_/paella/ui/watch.html?id=_/watch/_g' .../archive/*/*/*/manifest.xml
+        sed -i 's_/engage/ui/watch.html?id=_/watch/_g' .../archive/*/*/*/manifest.xml
+
+2. Run the following SQL commands on your opencast database (replacing `<playerlink>`):
+
+        UPDATE oc_assets_snapshot
+          SET mediapackage_xml =
+          REPLACE(mediapackage_xml, '<playerlink>', '/watch/')
+          WHERE INSTR(mediapackage_xml, '<playerlink>') > 0;
+
+    For Theodul, Paella and the Engage player the specific commands would be:
+
+        UPDATE oc_assets_snapshot
+          SET mediapackage_xml =
+          REPLACE(mediapackage_xml, '/engage/theodul/ui/core.html?id=', '/watch/')
+          WHERE INSTR(mediapackage_xml, '/engage/theodul/ui/core.html?id=') > 0;
+        UPDATE oc_assets_snapshot
+          SET mediapackage_xml =
+          REPLACE(mediapackage_xml, '/paella/ui/watch.html?id=', '/watch/')
+          WHERE INSTR(mediapackage_xml, '/paella/ui/watch.html?id=') > 0;
+        UPDATE oc_assets_snapshot
+          SET mediapackage_xml =
+          REPLACE(mediapackage_xml, '/engage/ui/watch.html?id=', '/watch/')
+          WHERE INSTR(mediapackage_xml, '/engage/ui/watch.html?id=') > 0;
+
+Please ensure to execute these steps before rebuilding the index.
+
+
 Rebuild the Elastic Search Indexes
 ----------------------------------
 
@@ -89,4 +141,3 @@ Due to [MH-13396](https://opencast.jira.com/browse/MH-13396), the configuration 
 the Admin UI and the External API have changed.
 
 Therefore, those indexes both need to be rebuilt.
-
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -90,7 +90,7 @@ Update Player Links
 > This step is optional
 
 Opencast 7 comes with the capability of dynamically switching the configured player without requiring the republication
-of all published material by providing the dynamic target `https://example.opencast.org/watch/<id>`. This new target
+of all published material by providing the dynamic target `https://example.opencast.org/play/<id>`. This new target
 will be used for all new engage publications.
 
 But old publications still reference the players directly in the admin interface and the external API. This does not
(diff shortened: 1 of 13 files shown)


docs/guides/admin/docs/upgrade.md
```diff
@@ -82,11 +83,61 @@ change `maintenance` back to `false` to put the scheduler back into its normal m
...
+
+1. Find the archive directory and run the following command (replacing `<playerlink>`):
+
+        sed -i 's_<playerlink>_/watch/_g' .../archive/*/*/*/manifest.xml
+
+    For Theodul, Paella and the Engage player the specific commands would be:
+
+        sed -i 's_/engage/theodul/ui/core.html?id=_/watch/_g' .../archive/*/*/*/manifest.xml
+        sed -i 's_/paella/ui/watch.html?id=_/watch/_g' .../archive/*/*/*/manifest.xml
+        sed -i 's_/engage/ui/watch.html?id=123_/watch/_g' .../archive/*/*/*/manifest.xml
```
JulianKniephoff: Is `123` some kind of magic Engage ID prefix? (No sarcasm, I genuinely don't know.) :sweat_smile:

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -679,23 +677,19 @@ MediaPackage replaceAndDistributeAcl(MediaPackage previousMp, AccessControlList
...
               "Using 'server.url' as a fallback for the non-existing organization level key '{}' for the publication url",
               ENGAGE_URL_PROPERTY);
     }
-    if (playerPath == null)
-      playerPath = DEFAULT_PLAYER_PATH;
 
     try {
       // Create new distribution element
-      URI engageUri = URIUtils.resolve(new URI(engageUrlString), playerPath + "?id=" + mp.getIdentifier().compact());
+      URI engageUri = URIUtils.resolve(new URI(engageUrlString), PLAYER_PATH + mp.getIdentifier().compact());
```
JulianKniephoff, line 692: This code, building the link to the watch page, is still duplicated. We might want to use this opportunity to factor this into some kind of helper method, so that every module that needs a link to a player page can just say `whateverService.getPlayerLink(mpId)`. :thinking:

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -224,9 +224,12 @@ prop.adminui.user.external_role_display=false
...
 # specific videos will be appended automatically.  Common values include:
-# - theodul player: /engage/theodul/ui/core.html
-# - paella  player: /paella/ui/watch.html
-prop.player=/engage/theodul/ui/core.html
+#
+# - theodul player: /engage/theodul/ui/core.html?id={{id}}
+# - paella  player: /paella/ui/watch.html?id={{id}}
+#
+# Default: /engage/theodul/ui/core.html?id={{id}}
+#prop.player=/engage/theodul/ui/core.html?id={{id}}
```
karendolan: @lkiesow  is theodul still a default?

smarquard:
What would happen if a player has additional URL parameters (for example to start playing at a particular point, or an embed view) - are they passed through?

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -103,9 +103,7 @@
   /** The engage base url property **/
   static final String ENGAGE_URL_PROPERTY = "org.opencastproject.engage.ui.url";
   /** The default path to the player **/
-  static final String DEFAULT_PLAYER_PATH = "/engage/ui/watch.html";
-  /** The configuration property value for the player location. */
-  static final String PLAYER_PROPERTY = "player";
+  static final String PLAYER_PATH = "/watch/";
```
karendolan, line 107: @lkiesow would you mind clarifying the comment to be something like "The default path to the engage-ui module's player REST endpoint", so future developer's know which module that path is linked to.

Is there any OC standard'ish way to indicate which modules have implicit dependencies between them? [EDIT], just read @JulianKniephoff 's comment regarding moving the path creation to a common location, and moving this path creation to it's own service, something like engage-player-redirect, or a better name, will make the common dependency very clear and developer-user-friendly.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
karendolan: +1

karendolan: Moving the redirect-path creation path to a separate common service will also make it easier for a site, like us, that uses 2 different engage path's for one organization during one workflow (the first published path forces a re-direct to Engage login before re-redirecting to the player, the second publish path doesn't) to encapsulate our custom[...]ness into one isolated service.

modules/engage-ui/src/main/java/org/opencastproject/engage/ui/PlayerRedirect.java
```diff
@@ -0,0 +1,84 @@
...
+@RestService(
+  name = "PlayerRedirect",
+  title = "Configurable Player Endpoint",
+  abstractText = "This service redirects to configured players.",
+  notes = {})
+public class PlayerRedirect {
+
+  private static final Logger logger = LoggerFactory.getLogger(PlayerRedirect.class);
+
+  private static final String PLAYER_DEFAULT = "/engage/theodul/ui/core.html?id={{id}}";
```
karendolan, line 58: Is Theodul still a good default player path for OCv7x?

docs/guides/admin/docs/upgrade.md
```diff
@@ -82,11 +83,61 @@ change `maintenance` back to `false` to put the scheduler back into its normal m
 You should avoid running Opencast 7 without migrating the scheduled events first. Otherwise, your capture agents may
 fetch an empty calendar.
 
+
+Update Player Links
+-------------------
+
+> This step is optional
+
+Opencast 7 comes with the capability of dynamically switching the configured player without requiring the republication
```
karendolan, line 92: @lkiesow can you clarify that this the publish link from the Admin UI. For example, something like: "...dynamically switching the configured player that is referenced from the Admin UI without requiring...."

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
```diff
@@ -114,7 +114,7 @@
   private static final String MERGE_FORCE_FLAVORS_DEFAULT = "dublincore/*,security/*";
 
   /** The default path to the player **/
-  protected static final String DEFAULT_PLAYER_PATH = "/engage/ui/watch.html";
+  protected static final String PLAYER_PATH = "/watch/";
```
karendolan: It would help to clarify in the comment that it reference an Engage UI REST Endpoint that returns a path to the player at runtime. Or, as @JulianKniephoff  suggested, provide a common service that returns the common player redirect path.

docs/guides/admin/docs/upgrade.md
lkiesow: No, it's not. It's my test id. Removed.

lkiesow:
> What would happen if a player has additional URL parameters (for example to start playing at a particular point, or an embed view) - are they passed through?

The same as right now. If you configure them in the organization configuration, they are passed to the player. If you add them to the actual player page, you can still use that link. At that point, it's specific to that player and event anyway.

Using the player agnostic redirect to pass player specific URL parameters on an event by event basis will obviously not work.

[edit] Correction, actually it can work since the redirect accepts whatever you throw at it as id. But you still shouldn't do that. It does not make much sense.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
lkiesow: I've looked into this and I agree that if these URLs are an integral part of the organization, it might make sense to add methods like `getAdminUrI()` or `getPlayerLink(…)` to the organization at which point it would also make sense to move the configuration for those from the general, optional properties to dedicated configuration values.

But this is somewhat unrelated to this since it wouldn't make much sense to do this only with the engage URL and another, likely larger patch which I'm happy to review if one of you wants to provide the code ;-P

modules/engage-ui/src/main/java/org/opencastproject/engage/ui/PlayerRedirect.java
lkiesow: Yes, it definitely is. Paella still does not work with the default configuration which the default player should do ;-)

docs/guides/admin/docs/upgrade.md
lkiesow: But it's not just the admin interface. It's basically all places: admin interface, external API, LTI tools, media module, search endpoint (I think that's all)

lkiesow:
I'm just wondering if, instead of `/watch/`, we should use something short like `/v/`. E.g.:

- https://example.opencast.org/v/1ce397a0-396f-4446-861a-f00866e7e55c

smarquard:
I would tend to think of "/v/" as "version". /play/ or /view/ maybe?

karendolan:
The term "/viewer/<id>"  helps clarify that it's returning a link to the player page.

lkiesow:
I was thinking about YouTube's `v=`:

- https://www.youtube.com/watch?v=o1dBg__wsuo

I like `/watch/` better than `/view/` for videos. `/watch/` vs `/play/`… I don't care ;-)

karendolan:
(between the two options, /play/  sounds more "interactive", less passive than /watch/. Thus, better for PR.)

lkiesow:
Changed the path to `/play/`. @karendolan, @smarquard, will one of you officially take the review and merge it if there are no more problems?

JamesUoM:
I've tested the code and the various links as well as gonethrough updating the publication for existing archives. All appears to work.

PR #784, opened 2019-03-18
MH-12139  Admin UI new event media upload progress bar

karendolan:
This pull causes a new event media upload to show a prominent yellow progress bar containing the percentage of upload across the top of the Admin UI, below the main banner.

The motivation is to let the user know what is going on the background when a large file upload is started and an indication of when it will be complete. Showing this information helps to stop the user from accidentally closing the browser, or starting another upload, or anything that will jeopardize the current large file upload.

This pull adds a new resource, chieffancypants angular-loading-bar tool. There is already a chieffancypants hot key tool with the same license. The angular-loading-bar is included in the same way the first tool was included (a checked in bower_component).

This pull also changes the behavior of the event success/fail notification, in that it does not automatically remove the notification after a few seconds, and it gives the notification a global context so the user can see it if the user has moved onto other tabs while the event was still being processed.

### DIFF
--- a/modules/admin-ui/src/main/webapp/index.html
+++ b/modules/admin-ui/src/main/webapp/index.html
@@ -162,6 +162,7 @@
     <script src="scripts/shared/services/regexService.js"></script>
     <script src="scripts/shared/services/schedulingHelperService.js"></script>
     <script src="scripts/shared/services/eventHelperService.js"></script>
+    <script src="scripts/shared/services/progressBarService.js"></script>
     <script src="scripts/shared/services/hotkeysService.js"></script>
     <script src="scripts/shared/services/restServiceMonitor.js"></script>
     <script src="scripts/shared/resources/resources.js"></script>
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/newEventResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/newEventResource.js
@@ -21,7 +21,8 @@
 'use strict';
 
 angular.module('adminNg.resources')
-.factory('NewEventResource', ['$resource', 'JsHelper', function ($resource, JsHelper) {
+.factory('NewEventResource', ['$resource', 'JsHelper', 'ProgressBar', function ($resource, JsHelper, ProgressBar) {
+
   return $resource('/admin-ng/event/new', {}, {
     save: {
       method: 'POST',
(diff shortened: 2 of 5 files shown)


karendolan:
The progress bar UI in the current state of this pull becomes conflicted when more than one event is at the same time. I'm modifying the code to add a new progress bar row for each active upload. I'm hesitant to remove the green "uploading in progress"  banner in the lower area, that duplicates the upload message, because users are accustomed to looking there.

staubesv:
@karendolan Ok, thanks for the info. Please let me know once I should continue the review.

Note that you might encounter similar problems when it comes to the creation of (multiple) scheduled events.

mliradelc:
Hi, just a reminder, next Monday (April 1st) is the feature freeze for opencast 7.0 😉

staubesv:
@karendolan Just to be sure: The ball is at your side currently, isn't it?

staubesv:
@karendolan Please note that https://github.com/opencast/opencast/pull/852 (still looking for a reviewer...) might be relevant for this work and that this PR currently has conflicts.

KatrinIhler:
@karendolan If there is nothing happening here, it might be better to close this PR and refile once it's ready.

gregorydlogan:
@karendolan any ETA for further development?

gregorydlogan:
Conflicts @karendolan

karendolan:
Hi @gregorydlogan I'm actually on vacation this week, I fixed the immediate conflicts but not the issue of the multiple long uploads, nor resolved the color of the progress bar. As I was looking at it today, also, I think it's worth revisiting the idea of putting the progress bar right into that existing green "event is being uploaded..." notice bar. since it is already there and already trying to tell the user that something is happening. So it might as well be giving status during a very long upload. 

Do you mind if I attempt to migrate the bar into the green "upload in progress notification" and revisit the reasons why this was rejected originally? If I cannot get it resolved by Aug 17. Then I will close this pull.

gregorydlogan:
Per discussion in the technical meeting today, Karen's going to continue with with this pull. If we still can't get this done then we can revisit closing this later.

karendolan:
The pull is now simplified to add an upload percentage to an already existing new event upload notification (see screen shot below). The controversial progress bar UI element is omitted.

Purpose: Adding a percentage indicator to the existing upload notification gives the user confidence that a large upload is still progressing (and not a stick spinner). The user can gage how fast the upload is progressing: if the user should stay and wait for a fast upload, or go out for a coffee and come back for a slow one. 

fileUploadPercentageIndicator

gregorydlogan:
I'll take a look at this in the next couple of days since I've been bugging you 'bout it :)

gregorydlogan:
This looks good to me.  My only complaint is that my tests froze at 98.x%, but that freeze appears to be related to waiting on the system to percolate.  When the pause ended the was an update message sent via ActiveMQ, so I'm guessing that it might have been waiting on disk or some other processor intensive thing to actually finish.  Regardless, this works better than nothing, so let's merge it.

PR #785, opened 2019-03-18
MH-13240 Remove WorkflowStatistics

KatrinIhler:
This is part of the bigger endeavor to remove the Workflow service index. Instead of adapting the WorkflowStatistics to read them from the database instead of the index, this PR removes them since they are not in use.

_This work is sponsored by UCT._


staubesv:
@KatrinIhler We don't use those statistics at SWITCH, but as the statistics can be retrieved by Opencast's REST API (and JMX?), you actually cannot know whether somebody uses this or not ;-)

I would at least document the changes to the REST API similar to https://github.com/opencast/opencast/pull/751/files in the release notes and maybe bring this up on list.

KatrinIhler:
@staubesv You're right, my point was that Opencast does not use this anymore, so I'd rather not invest the work to rewrite those endpoints to use the database.

Thanks for pointing out the documentation, I'm gonna add to it and think about putting it on list.

smarquard:
We actually do use these at the moment for graphing workflows like this:

Screenshot 2019-03-18 at 19 56 27

We use the top level of info (workflow ids and counts per workflow ids), excluding the operations.

Screenshot 2019-03-18 at 19 58 58

So I wouldn't like to see this disappear (though we don't care about operations inside workflows), unless there's an equivalent way to query this efficiently via the external api.

Our main interest is running workflows. In fact it would be nice if the Admin UI could show this info (e.g. Systems | Workflows) as an intermediate level of detail between System | Jobs and System | Servers. So the Admin UI change is out-of-scope, but would be helpful to have the support for the admin ui endpoints to call.

KatrinIhler:
@smarquard  Unfortunately the external API can't be queried for more than one workflow right now, I'm afraid. Though I'm wondering, if it's running workflows you're interested in, is that not what the Admin UI shows as running events? The number should be accurate, since workflows can't be run on the same event at the same time. Same for instantiated.

smarquard:
The total number of running workflows we could get from the Admin UI endpoint, but we're interested in what type of workflows are running too (by workflow id). So one more level of detail, of a type that the Admin UI doesn't present at the moment.

KatrinIhler:
@smarquard Okay, I understand now. In that case that kind of information would have to be queried from the database once the index is gone... I'll have to think about this.

PR #786, opened 2019-03-18
MH-13444, Insecure Series Creation

lkiesow:
This patch prevents users from being able to overwrite series metadata
by default when ingesting event data. Users can still enable this, if
they need this functionality.

In addition to this, Opencast will not any longer add anonymous access
to series if they are created as part of the event ingest.

*Note that the behavior of both features is questionable and we may want
to remove this entirely in the future.*

### DIFF
--- a/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
+++ b/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
@@ -8,7 +8,7 @@
 
 # TODO: consider moving this key to the system.properties if used by other services, such as mediapackage update
 
-org.opencastproject.series.overwrite=true
+org.opencastproject.series.overwrite=false
 
 #The approximate load placed on the system by ingesting a file
 #Since these jobs are *not* dispatched there is no current way to limit the simultaneous number of ingests, but these jobs will block further jobs from running on an already busy admin node
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -56,8 +56,6 @@ import org.opencastproject.metadata.dublincore.DublinCoreValue;
 import org.opencastproject.metadata.dublincore.EncodingSchemeUtils;
 import org.opencastproject.scheduler.api.SchedulerException;
 import org.opencastproject.scheduler.api.SchedulerService;
-import org.opencastproject.security.api.AccessControlEntry;
-import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.SecurityService;
@@ -249,7 +247,7 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
           .build();
 
   /** The default is to overwrite series catalog on ingest */
-  protected boolean defaultIsOverWriteSeries = true;
+  protected boolean defaultIsOverWriteSeries = false;
 
   /** Option to overwrite series on ingest */
   protected boolean isOverwriteSeries = defaultIsOverWriteSeries;
@@ -880,9 +878,6 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
             logger.info("Creating new series {} with default ACL", id);
             seriesService.updateSeries(dc);
             isUpdated = true;
-            String anonymousRole = securityService.getOrganization().getAnonymousRole();
-            AccessControlList acl = new AccessControlList(new AccessControlEntry(anonymousRole, "read", true));
-            seriesService.updateAccessControl(id, acl);
           }
 
         } catch (Exception e) {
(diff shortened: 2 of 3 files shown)


modules/ingest-service-impl/src/test/java/org/opencastproject/ingest/impl/IngestServiceImplTest.java
```diff
@@ -561,7 +561,7 @@ public void testFailedJobs() throws Exception {
   private void testSeriesUpdateNewAndExisting(Dictionary<String, String> properties) throws Exception {
 
     // default expectation for series overwrite is True
-    boolean isExpectSeriesOverwrite = true;
+    boolean isExpectSeriesOverwrite = false;
```
karendolan: Update the comment, too. Or, better, remove the "is True" end portion of it.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -249,7 +247,7 @@
           .build();
 
   /** The default is to overwrite series catalog on ingest */
-  protected boolean defaultIsOverWriteSeries = true;
+  protected boolean defaultIsOverWriteSeries = false;
```
karendolan: Update or remove the old comment.

etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
```diff
@@ -8,7 +8,7 @@
```
karendolan, line 8: Update that "default" comment to false.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -880,9 +878,6 @@ protected boolean updateSeries(URI uri) throws IOException, IngestException {
             logger.info("Creating new series {} with default ACL", id);
             seriesService.updateSeries(dc);
             isUpdated = true;
-            String anonymousRole = securityService.getOrganization().getAnonymousRole();
-            AccessControlList acl = new AccessControlList(new AccessControlEntry(anonymousRole, "read", true));
-            seriesService.updateAccessControl(id, acl);
```
karendolan, line 885: @lkiesow so what default ACL does the new series ultimately get? Is it OC admin read-write-view?  Or can a default be defined in an organizational default series ACL? Our site actually uses a default series ACL when populating a semester's worth of series from a 3rd party.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
lkiesow: I went with a static constant I could use here as well

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
lkiesow: Not having an ACL, Opencast should fall back to its global default ACL which contains no rules at all. This essentially means that people with exceptions are allowed to access the series. People with exceptions are usually global and organizational administrators.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -191,6 +191,9 @@
   /** The key to look for in the service configuration file to override the {@link DEFAULT_INGEST_ZIP_JOB_LOAD} */
   public static final String ZIP_JOB_LOAD_KEY = "job.load.ingest.zip";
 
+  /** The default is to leave series catalogs untouched on ingest */
+  static final boolean DEFAULT_IS_OVERWRITE_SERIES = false;
```
karendolan: Ah, that constant, when said out loud, sounds a lot like it's going to overwrite the series ~ "The default is to overwrite the series (false)(??)". It might be better to just to put static final boolean FALSE  = false; :)

lkiesow:
```java
  static final boolean DEFAULT_ALLOW_OVERWRITE_SERIES = false;
 ```
?

Comment (karendolan) [reply modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java]: Ah, ok the "IS" version is more programer friendly, just not human friendly.

Comment (lkiesow) [reply modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java]: or maybe 
```java
static final boolean DEFAULT_ALLOW_SERIES_WRITE = false;
```
since we are dealing with writes in general not necessarily only modifications but also series creation.

lkiesow: Updated. Chose yet another name though :D

lkiesow:
@karendolan, can you officially take the review for this pull request and merge it if there are no more issues?

PR #791, opened 2019-03-21
MH-13451 Improve delete-event submit button

KatrinIhler:
- Rename button to 'delete' if there is nothing to retract
- Make delete-event button red if event8s) can be deleted immediately (makes users aware, consistent with delete-series)

### DIFF
--- a/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -272,6 +272,7 @@
       "BACK": "Previous",
       "CREATE": "Create",
       "RETRACT": "Retract",
+      "RETRACT_DELETE": "Retract/Delete",
       "DELETE": "Delete",
       "UPDATE": "Update",
       "DURATION": {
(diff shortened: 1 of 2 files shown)


KatrinIhler:
@staubesv Just noticed that this doesn't quite work as intended yet, give me a few more minutes! ;)

KatrinIhler:
@staubesv Ready for review!

staubesv (changes requested):
The behaviour is not really consistent:

Case 1:
User selects one or more events without publications.
Button: "Delete" (red)

Case 2:
User selects one or more events with publications.
Button: "Retract" (green)

Case 3:
User selects one or more events without publications and one or more events with publications.
Button: "Retract" (green)

Case 3 is suggesting that no events are deleted.

What do you think about the handling of case 3?

KatrinIhler:
@staubesv Good question. I assume in that case some events are deleted and others just retracted? In that case the button should reflect that...

staubesv:
@KatrinIhler Yes, both the delete and retract operations will be executed.

KatrinIhler:
@staubesv Alright, I'm gonna try to take a quick look at this today.

KatrinIhler:
@staubesv How about this? Not quite sure about "Retract/Delete" yet, maybe an "or" would be preferable...opinions?

staubesv:
@KatrinIhler I suggest "Retract & Delete" as this label is only shown if both retractions and deletions will be performed.

KatrinIhler:
@staubesv I think that wording is confusing because users could assume that opencast retracts the published events and then deletes all events, which...would make sense, but is not what is happening.

staubesv (approved):

PR #795, opened 2019-03-24
MH-13455, Opencast Plug-in Features

lkiesow:
This patch extracts some core features into separate Karaf features.
These features are installed by default, but not started by default.
This means that they can be easily enabled but will not launch
regardless of users needing those features or not.

The idea behind this is to extract modules used by just a few users to
minimize the code which is running by default, needing less resources,
booting up faster and also minimizing the risk for security
vulnerabilities.

This patch extracts the Moodle and Sakai user provider as an example.
More modules may follow.

Long-term, it may make sense to still boot all features in the
development distribution to ensure they are not easily broken without
someone noticing. That will be easier once KARAF-6207 is fixed:

- https://github.com/apache/karaf/pull/791

### DIFF
--- a/docs/guides/admin/docs/configuration/security.user.moodle.md
+++ b/docs/guides/admin/docs/configuration/security.user.moodle.md
@@ -36,12 +36,12 @@ After the installation, a new user with the capabilities
 Then generate a new web service token and add that user to the "Opencast web
 service" service.
 
-### Step 0
+### Step 1
 
 Edit `etc/org.apache.karaf.features.cfg` and make sure the `opencast-moodle` feature is listed in the `featuresBoot`
 option.
 
-### Step 1
+### Step 2
 
 To enable the Moodle User Provider, copy and rename the bundled configuration
 template from
@@ -57,7 +57,7 @@ org.opencastproject.userdirectory.moodle.url=http://localhost/webservice/rest/se
 org.opencastproject.userdirectory.moodle.token=mytoken1234abcdef
 ```
 
-### Step 2
+### Step 3
 
 Verify that the Moodle User Provider starts up with the correct Moodle URL by looking
 for a log entry like this:
@@ -78,7 +78,7 @@ by adding an entry to `OPENCAST/etc/org.ops4j.pax.logging.cfg`:
 log4j.logger.org.opencastproject.userdirectory.moodle=DEBUG
 ```
 
-### Step 3
+### Step 4
 
 You can grant additional roles to all Moodle users in Opencast by creating a
 group with the name 'Moodle'. You can then add additional roles to this group,
--- a/docs/guides/admin/docs/configuration/security.user.moodle.md
+++ b/docs/guides/admin/docs/configuration/security.user.moodle.md
@@ -36,6 +36,11 @@ After the installation, a new user with the capabilities
 Then generate a new web service token and add that user to the "Opencast web
 service" service.
 
+### Step 0
+
+Edit `etc/org.apache.karaf.features.cfg` and make sure the `opencast-moodle` feature is listed in the `featuresBoot`
+option.
+
 ### Step 1
 
 To enable the Moodle User Provider, copy and rename the bundled configuration
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -20,6 +20,10 @@ New Features
 - Add Internationalization support for series LTI tools
 - Display responsible person for workflows
 - Allow the Ingest Service to make authenticated requests to other servers
+- Some modules are now plugins. These are not started by default to reduce the amount of code running unnecessarily.
+  They can easily be enabled in `etc/org.apache.karaf.features.cfg`. Modified modules are:
+    - Moodle user directory
+    - Sakai user directory
 
 Improvements
 ------------
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -133,8 +133,6 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-static-file-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory-ldap/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory-moodle/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory-sakai/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-user-interface-configuration/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-urlsigning-common/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-urlsigning-service-api/${project.version}</bundle>
@@ -753,5 +751,14 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-security-cas/${project.version}</bundle>
   </feature>
 
+  <feature name="opencast-moodle" version="${project.version}">
+    <feature start-level="80">opencast-core</feature>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory-moodle/${project.version}</bundle>
+  </feature>
+
+  <feature name="opencast-sakai" version="${project.version}">
+    <feature start-level="80">opencast-core</feature>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-userdirectory-sakai/${project.version}</bundle>
+  </feature>
 
 </features>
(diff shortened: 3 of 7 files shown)


smarquard:
What are the implications of this for someone installing and running from rpms or other packages?

docs/guides/admin/docs/releasenotes.md
```diff
@@ -24,6 +24,9 @@ New Features
 - Display responsible person for workflows
 - Allow the Ingest Service to make authenticated requests to other servers
 - Crop service
+- Some features are now plugins and will not anymore started, regardless of people using them or not:
```
smarquard: Suggest rephrase to:

These features are now plugins, and will not be started by default unless enabled in `etc/org.apache.karaf.features.cfg`

lkiesow:
> What are the implications of this for someone installing and running from rpms or other packages?

The same as for everyone else: They need to add the feature they want to the boot-feature configuration. Since the features are still installed, they are still part of the distributions. The only difference is that they are not started by default. The type of installation makes no difference.

docs/guides/admin/docs/releasenotes.md
lkiesow: Rephrased although it's just a suggestion for the release managers anyway.

gregorydlogan:
Grabbing this since it directly impacts the packaging system as well.

lkiesow:
> Grabbing this since it directly impacts the packaging system as well.

How? This does not change the assemblies or what they contain. Apart from the configuration, that is. It changes nothing for the RPM packaging at least and I cannot see what it changes for the Debian packages.

smarquard:
I don't think this change has much benefit, and it adds an unnecessary additional step to activate the chosen provider.

The activate() method of these user directory providers does nothing except log startup, so there is no code running until at least one organization is configured for a provider (by the *.cfg..template config being rename to a .cfg file) and it gets instantiated through the updated() method.

lkiesow:
You forget that general OSGI activation magic is involved as well and e.g. the Sakai user directory factory is a managed service which constantly look for configuration based on which it would launch a user directory.

I agree that this is a low-risk code and does not require a lot of resources, but it is still completely unnecessary for most people.

Long run, I would like to see more modules becoming optional, at which point you will likely notice that you are running 30 bundles less. These two bundles are, apart from the already existing CAS plugin, just a very good starting point since they are simply to externalize and completely unnecessary for most users.

And a single line of configuration to activate a feature not many people are using isn't that hard, is it?

smarquard:
Long run, I think we should aim for a system where you can install it from packages, start it up, point a web browser at the default URL and configure everything else from there (i.e. through a UI), including which optional services are activated or not.

I don't strong objections to this though because the changes are small either way. I would suggest adding a comment to `etc/org.opencastproject.userdirectory.sakai-default.cfg.template` and the equivalent moodle config pointing out the requirement to enable the service in `etc/org.apache.karaf.features.cfg`.

lkiesow:
Can you make a specific suggestion about what you want me to add? Because right now those configuration files hold no information at all about what to do when you want to enable one of those user providers. and just adding  “edit `etc/org.apache.karaf.features.cfg` and make sure the `opencast-moodle` feature is listed in the `featuresBoot` option to enable the moodle user provider” seems a bit misleading since that is only part of what people need to do. Would you add the whole guide to the configuration file? Would you add a summary? Would you link the documentation?

Btb. you could use Karaf's FeaturesService to get these features, list them in a UI and enable them.

gregorydlogan:
I'm happy with this as is, or with @smarquard's suggestions.  I was concerned that this might cause issues but it doesn't :)

gregorydlogan:
@lkiesow merge conflicts

lkiesow:
Merge conflict fixed. Anything else you want me to do?

docs/guides/admin/docs/configuration/security.user.moodle.md
```diff
@@ -36,6 +36,11 @@ After the installation, a new user with the capabilities
 Then generate a new web service token and add that user to the "Opencast web
 service" service.
 
+### Step 0
+
+Edit `etc/org.apache.karaf.features.cfg` and make sure the `opencast-moodle` feature is listed in the `featuresBoot`
```
smarquard, line 41: We don't seem to have a etc/org.apache.karaf.features.cfg file?

Also I think this should be Step 1 and the next steps renumbered. Otherwise we'll end up with Step -1, -2, ...

smarquard:
For commenting in the config template files, I would suggest (and similarly for moodle):

```
# Sakai UserDirectoryProvider configuration

# This is an an optional service which is not enabled by default. To enable it,
# edit etc/org.apache.karaf.features.cfg and add opencast-sakai to the featuresBoot option.
```

lkiesow:
`etc/org.apache.karaf.features.cfg` is generated when building the assemblies based on your assembly configuration. Opencast will not start without this configuration file, so I'm pretty sure it exists in your installation :-)

---

I've updated the step numbers

---

I've added your text to the configuration files though I still find it kind of confusing to put half of the instruction into those files. Users still have to do some additional steps to make this work. The old steps are not documented in there. Just the one new step is.

smarquard:
Hm, I see. So in develop, karaf has added the version numbers. Does it work with a version?

Still don't like it. Hopefully @gregorydlogan feels more warmly about this change.

```
#
# Comma separated list of features to install at startup
#
featuresBoot = \
    (wrap), \
    cxf-http-jetty/3.2.6, \
    package/4.2.2, \
    scr/4.2.2, \
    log/4.2.2, \
    cxf-jaxrs/3.2.6, \
    aries-blueprint/4.2.2, \
    opencast-allinone/7.0.0.SNAPSHOT, \
    framework/4.2.2, \
    system/4.2.2, \
    feature/4.2.2, \
    shell/4.2.2, \
    eclipselink/2.7.3, \
    service/4.2.2, \
    jndi/4.2.2, \
    http/4.2.2, \
    jpa/2.7.0, \
    deployer/4.2.2, \
    http-whiteboard/4.2.2, \
    bundle/4.2.2, \
    config/4.2.2, \
    transaction/2.0.0
```

gregorydlogan:
Listing it without the version seems to pull the current version out of the Karaf jars, so I think this works as expected.  It might be nice to fix the documentation to explicitly mention that it should include the version regardless.

lkiesow:
> It might be nice to fix the documentation…

I don't think there is anything to fix. The documentation states:

> make sure the `opencast-moodle` feature is listed in the `featuresBoot`

Assuming that people are intelligent enough to add the feature in the same way the other Opencast features are listed or read the Karaf documentation. Writing a novel about this one single configuration key does not help. The format is really obvious already.

And yes, they could ommit it so that Karaf would pick the latest available version, though I wouldn't recommend that.

---

> Still don't like it. Hopefully @gregorydlogan feels more warmly about this change.

If we decide not to merge this, we really should revisit the `opencast-cas-security` feature since we discussed not to allow @miesgre to add it to the Opencast core as it would then always run on all machines regardless of people needing the features or not.

Back then we said it would be nice to have features not a lot of people use as optional components and I promised to implement that for a few other features. Deciding here that we want to have every other optional component running after all (but not cas) would be unfair.

I'll prepare a pull request adding all of the cas features to the core as an alternative to this one ion a moment.

gregorydlogan:
I'm going to drop my reasoning here rather than #814 (which I'll close in a second here, thanks for preparing it @lkiesow).

I personally don't care either way, but I'm going to merge this PR in favour of #814.  My reasoning is this:

* Less code running everywhere shrinks our potential security vulnerability surface.  It might be a very minor change right now, but future changes might not be.
* If we leave this in, we'll need to leave all other integrations in as they develop.  The kitchen sink is _already_ present in Karaf, we don't need the garage too.
* Re-enabling the modules removed here is trivial, especially since it's relatively well documented in the release notes, markdown docs, and config files.
* I can see an extreme case where one might want to turn off parts of our current node profiles (c.f: a worker which *only* encodes, nothing else).  Right now that can somewhat be accomplished with stupid job dispatch tricks (pending bugs), but it would be nice to be able to really enforce that by not running the code in the first place.  To that end, we should get used to being able to turn off features - and if we can get there then maybe we save some of the work of developing part of that UI @smarquard talked about by coopting the existing feature selection work inside Karaf.
* The argument that having this code enabled all the time surfaces bugs faster fails in the light of the code _not actually doing anything unless configured_.  Code that does nothing can't surface bugs.  If an institution feels strongly that the (CAS/Sakai/Moodle/Foobar) module(s) need testing, they should provide access and credentials, and/or provision hardware to test against.  Whether that's managed by the community, or said institution takes ownership of testing a given module is up to the institution I suppose.credentials,

Note: This is explicitly not a slam against UCT or @smarquard, both of whom are tremendously important to the community, but more a general rant by the QA person who has been asking for that last bullet point for years :)

lkiesow:
To ease things a bit further, let me promise that I will personally try to release a prototype of a web-based plugin management off which we can potentially build an easier way to enable plugins (if you consider handling web interfaces easier than handling a configuration file).

smarquard:
On the "last bullet point", I don't particularly feel that the Sakai module needs testing (we run it in production so if it was broken we'd know pretty soon): the important thing is that all the optional modules are still built in a full build and by the CI servers. Sakai specifically has public QA servers so there's no obstacle to testing the integration if anyone wants to.

I think moving to web-based configuration as much as possible is valuable for new adopters, and may take a while to get there but hope we can move incrementally in that direction. So +1 to a prototype.

PR #801, opened 2019-03-24
MH-13461, Dependency Fixes & Dependency Checks

lkiesow:
This patch fixes a number of dependency issues and adds the maven
dependency plugin  to modules to automatically check for incorrect
dependencies.


staubesv:
@lkiesow Just wondering: What is the reason to not add the dependency checks to all modules?

staubesv (approved):

lkiesow:
The reason is simple: Time ;-P
Many modules are more work than these ones

staubesv:
@lkiesow Expected that ;-)

PR #38, opened 2018-01-15
MH-12643, Allow worspace to directly read files from asset manager

lkiesow:
The workspace will get all files from the asset manager via HTTP, even
    if the files are accessible locally on the filesystem. While this will
    ensure the data's integrity as no one is able to accidentally write to
    files in the asset manager, this protection is unnecessary for a
    read-only access as we have for example with `Workspace.read(…)`.
    
This patch now tests if the asset manager storage is accessible via
    filesystem and if it is, will try to read files from there directly. If
    it fails to access the file, it will still fall back to the old method
    of using `Workspace.get(…)` and download the files.
    
Note that the existence test for the local asset manager may fail the
    first time Opencast starts (the first time ever!) due to the directory
    being created by Opencast itself. However, the worst case is that the
    fallback to the current state is used and the direct access is used only
    after a restart of Opencast.
    
Also note that this patch makes the workspace access assets directly via
    file system and it is not necessary to have the asset manager running on
    the same node. This means, that this will also work for example on
    worker or presentation nodes.

- *Based on pull request #7 *
- *Work sponsored by SWITCH*

PR #804, opened 2019-03-25
MH-12643, Allow workspace to read from asset manager

lkiesow:
The workspace will get all files from the asset manager via HTTP, even
if the files are accessible as local files. While this will ensure the
data's integrity as no one is able to accidentally write to files in the
asset manager, this protection is unnecessary for a read-only access.

Additional to the slow HTTP download, the current mechanism also
generates a lot more network trafic since all archived files are first
transferred from the storage server to the admin node (which basically
acts as proxy) and then transferred to the workspace (potentially on
another server).

This means that, for example, if a worker requests a media file, it is
transferred twice over the network interface of the admin node,
transforming it into a bottleneck for performance.

This patch now tests if the asset manager storage is accessible via
filesystem and if it is, will try to read files from there directly. If
it fails to access the file, it will still fall back to the old method
downloading the files.

This means that if e.g. the asset manager is not mounted on a worker,
everything will still work the same way it does right now.

Note that this patch makes the workspace access assets directly via the
file system and it is not necessary to have the asset manager running on
the same node (only the asset manager utils which are on all nodes
anyway). This means, that this will work on worker or presentation nodes
without modifications to what is running there.

Also note that the protection against modification of arcgived data
still holds since no general access is granted to the asset manager's
storage system.

---

This patch is an updated version of pull request #38 which was declined
back then for the reason of the workspace getting asset manager specific
code. This is what has changed since then:

- Those who were complaining back then (and many more) are now *actively
  using the old patch in production* ;-)
- Most asset manager specific code (e.g. resolving the asset manager
  URLs) has been moved into the asset manager utils which are readily
  available on all node types.
- Additional to directly reading files for `workspace::read` requests,
  files for `workspace::get` are copied löcally if possible to avoid
  tunneling all traffic through the admin node.

### DIFF
--- a/modules/asset-manager-util/pom.xml
+++ b/modules/asset-manager-util/pom.xml
@@ -62,64 +62,24 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
-    <!--
-      - test
-      -->
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <scope>test</scope>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.easymock</groupId>
-      <artifactId>easymock</artifactId>
-      <scope>test</scope>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.hamcrest</groupId>
-      <artifactId>hamcrest-all</artifactId>
-      <version>1.3</version>
-      <scope>test</scope>
-    </dependency>
-    <!--
-      - Parameterized tests for JUnit
-      - https://github.com/Pragmatists/JUnitParams
-      -->
-    <dependency>
-      <groupId>pl.pragmatists</groupId>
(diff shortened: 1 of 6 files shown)


smarquard:
Thank you for resurrecting this with a cleaner solution.

For the case where we get the file, can we use the same approach used elsewhere to hardlink the files where possible rather than a filesystem copy? This avoids a lot of IO in cases where the asset manager and WFR/workspace are on the same NFS share, which we have done locally for performance reasons because the asset manager will do a hardlink rather than copy if possible when copying files into the asset manager.

This assumes that the file in asset manager will not change, but I think that's a safe assumption as the asset manager versions media assets, so that specific version will never change.

lkiesow:
> For the case where we get the file, can we use the same approach used elsewhere to hardlink the files where possible rather than a filesystem copy?

No. That would violate the data protection put in place by the asset manager. Hard linking or handing out File objects to the files in the asset manager directly, which would effectively be the same, would allow anyone to simply write into the archive directly.

If you need read-only access: Use `workspace::read`.
Everything else, I would put through a discussion on list.

wsmirnow (approved):
This patch is fine for me.

smarquard (approved):
If a consensus emerges from the list discussion, it can be a follow up change.

Happy with this PR as-is.

PR #807, opened 2019-03-26
MH-13465, Prevent NullPointerException

lkiesow:
Restarting a server while the admin interface is still open it happens
from time to time that the table refresh kicks in at the wrong time and
requests resources from Opencast which end in the following
NullPointerException:

    (AbstractFaultChainInitiatorObserver:114) - An unexpected error
    occurred during error handling. No further error processing will
    occur.
    org.apache.cxf.interceptor.Fault: null
      ...
    Caused by: java.lang.NullPointerException
            at org.opencastproject.adminui.endpoint.AbstractEventEndpoint.getEvents
	    (AbstractEventEndpoint.java:2050)

This patch adds a simple null check preventing this.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -36,6 +36,7 @@ import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 import static javax.servlet.http.HttpServletResponse.SC_NO_CONTENT;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
+import static javax.servlet.http.HttpServletResponse.SC_SERVICE_UNAVAILABLE;
 import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
 import static org.opencastproject.index.service.util.RestUtils.conflictJson;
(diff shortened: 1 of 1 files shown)


staubesv (dismissed):
There are actually quite a number of requests in the Admin UI facade that would fail similarly if securityService.getOrganization() returns null...

... does it make sense to "fix" this for exactly one view leaving all others untouched?

Maybe it would be worth adding a utility function to the Admin UI facade that returns a HTTP 503 (Service Unavailable) in case getOrganization() or getUser() return null and just call that in front of any methods affected. 

Also not 100% clear to me whether this should be a 500 (vs. 503). Does getOrganization() == null || getUser() == null imply that Opencast is being restarted?

What do you think?

smarquard:
It would be nice to just not have the http endpoints come up until all the dependent services are started. I would have thought OSGI would do that for us in an ideal world.

lkiesow:
@smarquard, OSGI ensures that all static bound services are available. It cannot make any guarantees about inner service states.

@staubesv, 500 is more a there is something wrong in Opencast but we do not know what it is. But here we do: Some of the required resources are not yet available. So declaring that the service is unavailable does make sense.

In general it *may* make sense to have the organization available as part of the activation of the security service, but I'm not going to look into that.  This is just the place where I have seen the error multiple times due to the auto-reload and it should fix most exceptions. So, it's a “take it or don't” kind of patch. I'm not going to work on this any more right now.

staubesv:
@lkiesow I had another look at this. The same error will happen for any polling request of the Admin UI (not just events.json) and even on that page, /admin-ng/event/85071c96-a654-4e2b-b426-86fffb5d78b9/hasActiveTransaction caused a null pointer exception during startup. Besides that, it happens for almost all requests the result from user interaction, too.

Personally, I recommend to close this PR and wait until somebody has time to fix the problem for all requests instead of just picking the most seen one.

... as I don't have a strong opinion, I unassigned the review to not block it based on my personal opinion ;-)

lkiesow:
My argument would be that while it only fixes one endpoint, it actually fixes that endpoint and we shouldn't decline valid fixes because they do not fix all problems we have at once (That would be a big patch ;-P). But since this particular patch is mostly a cosmetical fix, this is not that important and I will not fight for it to be included. Whoever picks this up, feel free to decide either way.

smarquard:
This seems a reasonable improvement to me, and "we could do more" is not an argument against improving something.

PR #808, opened 2019-03-27
MH-13467: opencast-security-cas feature can not be started

miesgre:
There is a dependency problem. 
The`opencast-security-cas` feature need the bundle `mvn:commons-collections/commons-collections/3.2.1`


lkiesow:
The patch seems to work and definitely does not harm anything.
Though I noticed that activating this feature *on-thy-fly* more or less breaks Opencast.
But that should be unrelated to this patch.

PR #809, opened 2019-03-27
MH-13466, Prevent Capture Agents From Modifying Metadata

lkiesow:
Capture agents usually do not need to handle metadata at all but for
historical reasons, all capture agent data take precedence over
Opencast's data. This has shown to be troublesome as agents have broken
data for several reasons in the past.

This patch introduces two options causing the ingest service to discard
metadata catalogs and/or attachments sent by capture agents for
scheduled events, using the metadata available in Opencast instead.

The default is to discard the capture agent metadata.

### DIFF
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -36,6 +36,10 @@ Configuration changes
 - `KARAF_NOROOT` is now set to `true` by default, preventing Opencast to be started as root user unless the
   configuration is changed.
 - The default configuration for the Paella player has been moved to `etc/ui-config/mh_default_org/paella/config.json`
+- By default, metadata catalogs and attachments sent by capture agents are discarded since this data is usually
+  controlled by Opencast and the routing through capture agents which existed for historical reasons was just an
+  additional source for errors. If you rely on the old behavior, it can be configured in
+  `etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg`.
 
 API changes
 -----------
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -1393,25 +1393,29 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
     // Merge media package fields
     if (mp.getDate() == null)
       mp.setDate(scheduledMp.getDate());
-    if (isBlank(mp.getLicense()))
+
+    if (skipCatalogs || isBlank(mp.getLicense()))
       mp.setLicense(scheduledMp.getLicense());
-    if (isBlank(mp.getSeries()))
+    if (skipCatalogs || isBlank(mp.getSeries()))
       mp.setSeries(scheduledMp.getSeries());
-    if (isBlank(mp.getSeriesTitle()))
+    if (skipCatalogs || isBlank(mp.getSeriesTitle()))
       mp.setSeriesTitle(scheduledMp.getSeriesTitle());
-    if (isBlank(mp.getTitle()))
+    if (skipCatalogs || isBlank(mp.getTitle()))
       mp.setTitle(scheduledMp.getTitle());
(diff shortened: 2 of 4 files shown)


lkiesow:
Rebased onto the last non-broken commit in `develop`

docs/guides/admin/docs/releasenotes.md
```diff
@@ -39,6 +39,10 @@ Configuration changes
 - `KARAF_NOROOT` is now set to `true` by default, preventing Opencast to be started as root user unless the
   configuration is changed.
 - The default configuration for the Paella player has been moved to `etc/ui-config/mh_default_org/paella/config.json`
+- By default, metadata catalogs and attachments sent by capture agents are discarded since these data is usually
```
smarquard: "these data are" or "this data is" (subject/verb concord). Strictly speaking data is plural, so "these data are" is more grammatically correct but most often it's used as singular ("this data is")

lkiesow: I notice a correlation between me writing docs late at night and you complaining about grammar issues ;-D
Fixed.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1348,7 +1364,22 @@ private MediaPackage mergeScheduledMediaPackage(MediaPackage mp) throws IngestEx
...
+      }
+    }
+
+    // drop attachments the capture agent sent us in favor of Opencast's attachments
+    // e.g. prevent capture agents from modifying security rules of schedules events
+    if (skipAttachments) {
+      for (MediaPackageElement element : mp.getAttachments()) {
+        mp.remove(element);
+      }
+    }
```
karendolan, line 1376: @lkiesow Does it make sense to omit the opportunity for the ingest service merge new attachments and catalogs into the existing mediapackage while avoiding overwriting existing the catalogs and attachments (defined by flavor)?

lkiesow: I'm not going to do that. I have/see no use case for that. Quite the contrary, I think it might even cause confusion about what will ultimately end up in the media package and it will add further code making the module even more complex. That said, I'm not against you adding that functionality if you have a specific use case for it. We just don't.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1348,7 +1364,22 @@ private MediaPackage mergeScheduledMediaPackage(MediaPackage mp) throws IngestEx
     }
   }
 
-  private void mergeMediaPackageElements(MediaPackage mp, MediaPackage scheduledMp) {
+  private void mergeMediaPackageElements(final MediaPackage mp, final MediaPackage scheduledMp) {
+    // drop catalogs sent by the capture agent in favor of Opencast's own metadata
+    if (skipCatalogs) {
```
gregorydlogan, line 1364: We do this with the elements, do we also need to do this with the metadata?  In other words, does `mergeMediaPackageMetadata` also needs these changes?

lkiesow: Done.

lkiesow:
Merge conflict fixed.

JamesUoM:
OK just to be clear this ignores all CA catalogs and attachments whereas  @karendolan PR https://github.com/opencast/opencast/pull/714 only ignores those that match existing ones.

Frankly, I can't see a problem with a CA introducing _additional_ catalogs and attachments and seems superfluous.

gregorydlogan:
Adding new things is fine IMO.  Unless you modify your workflow, additional bits aren't going to be processed and thus likely won't break anything.  If you are modifying your workflow then YMMV, but you're also at the point where you need to start learning about how the workflow engine works :)

lkiesow:
The reason is simply that we have had some proprietary capture agents break stuff by ingesting things we did not want (broken metadata among other things) and normally there is no reason for agents to modify any data except the video tracks.

PR #810, opened 2019-03-27
MH-13468, Capture Agent Registration Exception

lkiesow:
When switching from a lazy capture agent state update to a new, accurate
one for pyCA which would cause the agent to first register itself as
offline and once it is available (which can be really fast) stating it's
status as idle, I noticed that Opencast will often fail if two agent state
requests are handled concurrently.

In short, what happens is that for both requests, Opencast (or rather
the Java persistence code) decides to use an SQL insert statement,
ultimately causing a primary key conflict for the slower one.

This patch fixes the problem by catching the conflict and repeating the
process, now being certain that the resource already exists which causes
an SQL update to be used instead.

Note that the underlying eclipselink library still logs a warning when
this problem occurs.


gregorydlogan:
Doesn't this run the risk of an infinite recursion if there is some other issue causing this?

lkiesow:
Shouldn't happen with the current code but I see your point of it being somewhat dangerous if e.g. someone modifies the code. Added a maximum number of retries.

PR #811, opened 2019-03-27
MH-13469, Drop LastHeardFrom On Scheduler Messages

lkiesow:
LastHeardFrom is part of the scheduler messages sent over for each
scheduled event. It is not used at all and can thus be dropped. Less
data to send.


lkiesow:
Rebased onto non-broken `develop`

PR #812, opened 2019-03-29
MH-13471, Shibboleth SSO plugin to add roles for users on OC according to their EDUPERSONAFFILIATION. eg: "ROLE_AAI_USER_AFFILIATION_student" for "student"

aliyevTU:
Shibboleth SSO plugin to add roles for users on OC according to their EDUPERSONAFFILIATION.

Example 1: if Alice as EDUPERSONAFFILIATION has "student;" value then "ROLE_AAI_USER_AFFILIATION_student" will be added to OC for Alice.

Example 2: if Bob as EDUPERSONAFFILIATION has "student;member;staff;alum" value then
"ROLE_AAI_USER_AFFILIATION_student"
"ROLE_AAI_USER_AFFILIATION_member"
"ROLE_AAI_USER_AFFILIATION_staff"
"ROLE_AAI_USER_AFFILIATION_alum" roles will be added to OC for Bob.

Developed for the following initial ticket: https://opencast.jira.com/browse/MH-13223

### DIFF
--- a/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
+++ b/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
@@ -27,6 +27,9 @@
 #header.home_organization = homeOrganization
 
 # Optional header to extract the affiliation
+# Note: Only if "header.affiliation" is uncommented and set to the correct Shibboleth header value, then
+# the corresponding roles will be added for each user with "role.affiliation.prefix" value prefix.
+# Default: none
 #header.affiliation = EDUPERSONAFFILIATION
 
 # Roles assigned to Shibboleth authenticated users
--- a/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
+++ b/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
@@ -26,6 +26,9 @@
 # Optional header to extract the home organization from
 #header.home_organization = homeOrganization
 
+# Optional header to extract the affiliation
+#header.affiliation = EDUPERSONAFFILIATION
+
 # Roles assigned to Shibboleth authenticated users
 
 # Based on information from the HTTP request headers, the configurable AAI login handler
@@ -50,4 +53,8 @@
 # Federation membership role name
 # Default: ROLE_AAI_USER
 #role.federation = ROLE_AAI_USER
+
+# Role prefix for affiliation
+# Default: ROLE_AAI_USER_AFFILIATION_
+#role.affiliation.prefix = ROLE_AAI_USER_AFFILIATION_
 
--- a/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
+++ b/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
@@ -32,11 +32,12 @@
 # Roles assigned to Shibboleth authenticated users
 
 # Based on information from the HTTP request headers, the configurable AAI login handler
-# assigns three basic roles to authenticated users:
+# assigns these basic roles to authenticated users:
 #
 # 1. The user role unique for each user
 # 2. The organization membership role indicating the home organization a user belongs to
 # 3. The federation membership role indicating a Shibboleth authenticated user
+# 4. The affiliation role based on the Shibboleth affiliation
 
 # User role prefix
 # Default: ROLE_AAI_USER_
--- a/modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
+++ b/modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
@@ -273,7 +273,7 @@ public class ConfigurableLoginHandler implements ShibbolethLoginHandler, RolePro
       logger.warn("Optional header '{}' is not configured ", CFG_HEADER_HOME_ORGANIZATION_KEY);
     }
 	
-	String cfgAffiliation = StringUtils.trimToNull((String) properties.get(CFG_HEADER_AFFILIATION_KEY));
+    String cfgAffiliation = StringUtils.trimToNull((String) properties.get(CFG_HEADER_AFFILIATION_KEY));
     if (cfgAffiliation != null) {
       headerAffiliation = cfgAffiliation;
       logger.info("Header '{}' set to '{}'", CFG_HEADER_AFFILIATION_KEY, headerAffiliation);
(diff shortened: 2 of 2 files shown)


smarquard:
Adding these as groups would provide a bit more flexibility than individual roles. e.g the Sakai provider will add the group ROLE_GROUP_SAKAI_INSTRUCTOR if you have instructor role in one or more sites.

So you could add

ROLE_GROUP_AAI_student
ROLE_GROUP_AAI_staff

which means you can use these groups in ACLs, but also give access to capabilities through the groups (e.g. differing Admin UI permissions, capture agent access, etc.)

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -425,6 +465,14 @@ private String extractEmail(HttpServletRequest request) {
     if (StringUtils.equals(id, bootstrapUserId)) {
       roles.add(new JpaRole(GLOBAL_ADMIN_ROLE, organization));
     }
+	if (headerAffiliation != null) {
+      String affiliation = request.getHeader(headerAffiliation);
+	  List<String> affiliations = Arrays.asList(affiliation.split(";"));
+	  for (String var : affiliations) {
```
JulianKniephoff: `var` is such a generic name. Why is this not called `affiliation`?

aliyevTU: I changed it as you suggested.

aliyevTU: > 
So, that was why I named it "var". Changed it to "eachAffiliation".

`[ERROR] /home/travis/build/opencast/opencast/modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java:[471,23] variable affiliation is already defined in method extractRoles(java.lang.String,javax.servlet.http.HttpServletRequest)`

> I changed it as you suggested.

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -252,6 +272,14 @@ public void updated(Dictionary properties) throws ConfigurationException {
     } else {
       logger.warn("Optional header '{}' is not configured ", CFG_HEADER_HOME_ORGANIZATION_KEY);
     }
+	
+	String cfgAffiliation = StringUtils.trimToNull((String) properties.get(CFG_HEADER_AFFILIATION_KEY));
```
staubesv: Indentation ;-)

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -298,6 +326,18 @@ public void updated(Dictionary properties) throws ConfigurationException {
       roleOrganizationSuffix = CFG_ROLE_ORGANIZATION_SUFFIX_DEFAULT;
       logger.info("AAI organization membership role suffix '{}' is not configured, using default '{}'",
               CFG_ROLE_ORGANIZATION_SUFFIX_KEY, roleOrganizationSuffix);
+    }
+	
+	String cfgRoleAffiliationPrefix = StringUtils.trimToNull((String) properties.get(
```
staubesv: Indentation

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -425,6 +465,14 @@ private String extractEmail(HttpServletRequest request) {
     if (StringUtils.equals(id, bootstrapUserId)) {
       roles.add(new JpaRole(GLOBAL_ADMIN_ROLE, organization));
     }
+	if (headerAffiliation != null) {
+      String affiliation = request.getHeader(headerAffiliation);
```
staubesv, line 469: Depending on the Shibboleth federation and the configuration of the Shibboleth service provider, it might be possible that this header is not set at all.

Therefore, you need to add the check "if (affiliation != null)".

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
aliyevTU: Thank you for bringing it to my attention. Fixed all the indentations :)

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -425,6 +465,14 @@ private String extractEmail(HttpServletRequest request) {
...
+	  List<String> affiliations = Arrays.asList(affiliation.split(";"));
+	  for (String eachAffiliation : affiliations) {
```
staubesv: No strong opinion, by eachAffiliation seems a somewhat weird choice for a variable name to me (e.g. read roleAffiliationPrefix + eachAffiliation). I suggest to use simply "affiliation" here.

etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
```diff
@@ -51,3 +54,7 @@
 # Default: ROLE_AAI_USER
 #role.federation = ROLE_AAI_USER
 
+# Role prefix for affiliation
+# Default: ROLE_AAI_USER_AFFILIATION_
+#role.affiliation.prefix = ROLE_AAI_USER_AFFILIATION_
```
staubesv, line 63: Maybe add a forth point after line https://github.com/opencast/opencast/pull/812/files#diff-6861a5481b6de717f8e1169f75581425R39

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
aliyevTU: Since there already exists a variable called "affiliation", and "var" is too generic, I settled with "eachAffiliation".

staubesv: You could rename _affiliation_ to _affiliationHeader_, then the name _affiliation_ is free 

Anyway, no strong opinion here ;-)

etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg
aliyevTU: Thank you, done.

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
aliyevTU: Fixed

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
aliyevTU: Since this feature is optional, only if on the configuration file "header.affiliation" is uncommented and set to the correct value, this feature will work, therefore no check for "affiliation" value is necessary.

aliyevTU: In case of a faulty configuration/typo by system administrator "affiliation" check added, as a safety measure.

staubesv (approved):

PR #814, opened 2019-03-29
MH-13209, Add CAS Feature to Opencast's Core

lkiesow:
This is an alternative to pull request #795.

The discussion on pull request #795 is about wanting to put all optional
features/modules into the Opencast core. Pull request #543 was declined
for exactly the reason that we wouldn't want to do that. If we decide to
do this way after all, we should also accept the old pull request.

This patch reverts pull request #544 and applies the previously declined
pull request #543 instead. The pull request adds all of the CAS modules
to Opencast's core feature, always starting those modules with Opencast,
regardless of users needing the feature or not.

This change has the advantage of making the feature slightly easier to
configure as well as better tested since problems with those features
will affect all installations.


gregorydlogan:
Declining in favour of #795.  See my logic in that PR.  Thanks for preparing this @lkiesow.

PR #815, opened 2019-03-30
MH-13389 More informative job load logging

smarquard:
This adds more job load logging so it's easier to understand job allocation across worker nodes (and also then adjust job load configurations to more closely match real-world conditions)


lkiesow (commented):
The change looks reasonable except maybe for the thread information now being logged which is part of the logger information already. It is just not configured by default. In the [PatternLayout](https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout) we are using, you can just add [`threadId` as  pattern](https://logging.apache.org/log4j/2.x/manual/layouts.html#Patterns).

Does it make sense to log it twice? What do you think?

smarquard:
I didn't change the thread logging (it was there before) but am happy to remove it as I agree it shouldn't be logged explicitly (locally we log it as part of the pattern).

lkiesow:
Ups, you are right. It was there before. Sorry.
Your call if you want to change it or not.

smarquard:
Removed the thread logging. Please squash the commits on merge.

lkiesow:
Thanks. Looks good. Merged.

PR #816, opened 2019-03-30
MH-13474  Change the default composer job load from 0.8 to 1.5

smarquard:
Based on production experience, the default composer job load of 0.8 is too low.

### DIFF
--- a/modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
+++ b/modules/composer-service-api/src/main/java/org/opencastproject/composer/api/EncodingProfileImpl.java
@@ -85,7 +85,7 @@ public class EncodingProfileImpl implements EncodingProfile {
   protected HashMap<String,String> suffixes = new HashMap<String, String>();
 
   @XmlElement(name = "jobLoad")
-  protected Float jobLoad = 0.8f;
+  protected Float jobLoad = 1.5f;
 
   /**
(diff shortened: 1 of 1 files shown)


wsmirnow:
Please update the documentation too. See https://docs.opencast.org/develop/admin/configuration/load/#step-1-determine-your-load-values.

mtneug:
All values were changed below 1.0 for 5.x (see https://github.com/opencast/opencast/pull/272#discussion_r191875606). Maybe @lkiesow could provide the context for the decision?

smarquard:
I think there may have been some discussion about the implications of higher job load values on test systems with very small resources?

Maybe the problem is that this default is not configurable. I don't quite recall exactly why we needed to change it in code, but there seemed to be cases where the job load was still 0.8 despite setting explicit job load values wherever we could.

gregorydlogan:
@smarquard it can be overridden on a per-encoding settings basis, so perhaps that's just being missed?

I'm fine with increasing it to 1.0, but 1.5 might cause issues with low-end systems configured not accept oversize jobs.

wsmirnow:
If the job load exceed the max load of the server, then you will see a warning message. But the job will be processed anyway. This is fine for me.

wsmirnow (approved):

mtneug:
@wsmirnow not that this is only the default configuration (see `org.opencastproject.job.load.acceptexceeding=true`).

PR #817, opened 2019-03-30
MH-13475 role target filtering

smarquard:
To allow role providers to define the purpose of provided roles (ACLs, or capabilities), add Role.Target to the role object so it can be used for filtering in Asset Manager and index queries.

This is in some ways a simplification of:

MH-13320, Asset Manager Performance #658

in that it allows the asset manager to filter roles without needing to know anything about the role prefixes, and also future role providers to introduce new roles that don't show up in ACL or index queries (which can have a performance overhead).

OTOH it does touch quite a bit of code to do so.


smarquard:
Not tested yet, so just putting this up here as an early preview.

The role filtering here by Target.ACL can also be extended to the solr queries.

smarquard:
Related: https://opencast.jira.com/browse/MH-13143 and #818 

It should not be possible (in the Admin UI) to add roles with Target.USER to Event and Series ACLs, but it is currently possible for Event ACLs because of a regression.

smarquard:
This is not quite working yet because we need to persist the role target properly to oc_role, and doing that is slightly complex (currently new roles are added by UserDirectoryPersistenceUtil.java).

I'm not totally sure this is all worth it though. Might be simpler just to add a global (custom.properties) list of role prefixes which are user roles only and should be filtered, and a utility method to filter a set of roles.

KatrinIhler:
@smarquard Info for next time: You can create draft pull requests now for work that is still in progress so it can't be merged by accident! :)

gregorydlogan:
Does it make sense to decline this for now and refile once it's ready?

KatrinIhler:
@gregorydlogan I've been wondering the same thing...

gregorydlogan:
@smarquard  any ETA for further development here?

gregorydlogan:
Per discussion on the dev list, let's decline this for now.  @smarquard feel free to refile either as a draft, or when ready.

PR #818, opened 2019-03-30
MH-13476 Filter capture agent roles for ACLs

smarquard:
ROLE_CAPTURE_AGENT_* roles are intended to provide access to CA scheduling, and not for event and series ACLs. 

They should therefore not be returned by findRoles() called with a target of ACL. This affects how the Admin UI offers roles for event and series ACLs (well, the event ACL is not doing this properly at present which is a regression, but the series ACL is).


wsmirnow (approved):
looks good

PR #819, opened 2019-03-31
MH-13477, Faster Asset Manager Property Access

lkiesow:
This patch uses the new native asset manager property access for
workflow property retrieval.

*Work sponsored by SWITCH*


smarquard:
Does this improve the underlying mysql query that gets generated?

lkiesow:
Yes. The current one still joins the snapshot and property tables

smarquard:
This is a performance optimization rather than feature, so perhaps redirect to 7.x ?

lkiesow:
Moved to `r/7.x`.

Note that I'm on vacation for the next two weeks.

I specifically allowed edits from maintainers. Hence, feel free to modify anything that needs modification if you do not want to wait or contact someone else at [ELAN e.V.](mailto:opencast-support@elan-ev.de) to fix problems

smarquard:
We have merged this into our (modified) 6.x and it is running in production. If no issues show up in the next day or two, I will merge this.

PR #822, opened 2019-04-01
NOJIRA: Migrate to Kotlin

pmiddend:
# Abstract

After a short discussion, plapadoo has decided to migrate the Opencast code-base from Java to [Kotlin](https://kotlinlang.org/) (currently in 39th place on the prestigious [TIOBE Index](https://www.tiobe.com/tiobe-index/)!). This PR contains careful hand-migrations for almost all modules. Note that this branch will fail CI because of some…precautions that have to be taken care of, first.

# Benefits

- Kotlin is more **concise** and **safe** than Java.
- We can actually transition to using Kotlin **everywhere**, since it can compile to JVM bytecode **and Javascript code**. Thus, it is one language to rule them all. And who likes Javascript, anyway?
- IntelliJ’s Kotlin plugin is superb, others are practically nonexistant, so Eclipse and Vim users will finally be drawn away from Opencast.
- As a young and up-and-coming language, Kotlin will rejuvenate Opencast’s code base and attract the “cool kids”.

# Migration path

Of course, everybody has to learn Kotlin first. It’s also advised to “un-learn” all other programming languages, so as to fully immerse oneself into Kotlin.
Unfortunately, we don’t have much time, as this PR will become outdated soon. We propose to merge it immediately (and disregard the any merge conflicts). 
There will surely be people with doubts about this, and who might close this PR. In that case, we will open it up again, without modifications, until it finally passes.

This work was sponsored by ~~Jetbrains~~plapadoo.


smarquard:
I am thinking Rust would be a better choice. Kotlin is rather 2017 by now, and soon it will be very last-decade. Rust is like C, but cooler.

We spend a lot of time dealing with unnecessary abstractions in high-level languages in Opencast, and as video is basically just binary data, we should just get right down there to the lowest levels.

mtneug:
@pmiddend nice work!

Will this be part of Opencast 7.0?

staubesv:
@pmiddend Could you maybe redirect this PR to r/6.x? Seems people can't wait to get Kotlin-ed ;-)

gregorydlogan:
Wait, I thought we were doing the rewrite in Python?

pmiddend:
We'll see ourselves out.

PR #823, opened 2019-04-01
MH-13265, Automatic caption using Google speech to text api

tyfranck:
Implementation of automatic captions/transcriptions using Google speech to text. 
Please review :-)


lkiesow:
Quick first glance:

- test failures
- database migration script is missing

modules/transcription-service-google-speech-impl/src/main/resources/OSGI-INF/google-speech-transcription-rest.xml
```diff
@@ -0,0 +1,22 @@
...
+	<implementation
+		class="org.opencastproject.transcription.googlespeech.endpoint.GoogleSpeechTranscriptionRestService" />
+	<property name="service.description" value="Google Speech Transcription REST Endpoint" />
+	<property name="opencast.service.type" value="org.opencastproject.transcription.googlespeech" />
+	<property name="opencast.service.path" value="/transcripts/googlespeech" />
+	<property name="opencast.service.jobproducer" value="true" />
+
+	<service>
+		<provide
+			interface="org.opencastproject.transcription.googlespeech.endpoint.GoogleSpeechTranscriptionRestService" />
```
lkiesow: indentation/tabs

modules/transcription-service-google-speech-impl/src/main/resources/OSGI-INF/google-speech-transcription-service.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0"
+	name="org.opencastproject.transcription.googlespeech.GoogleSpeechTranscriptionService"
+	activate="activate" immediate="true">
+	<implementation
+		class="org.opencastproject.transcription.googlespeech.GoogleSpeechTranscriptionService" />
```
lkiesow: tabs

tyfranck:
@lkiesow what do you mean by 'database migration script is missing'. 
I have added my table scripts to 'mysql5.sql', is that not enough?

lkiesow:
That's for new installations.
There are also the migration scripts: `docs/upgrade/6_to_7/mysql5.sqldocs/upgrade/6_to_7/mysql5.sql`

smarquard:
Just noting here that the persistent service here is suitable for use by multiple services. We have an implementation of a transcription service for Nibity (https://opencast.jira.com/browse/MH-13308) that we will submit in due course, and it will use these generalized persistence modules (so we don't have to have new tables for every new transcription or captioning service).

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
+Transcripts (Automated by Google Speech)
+========================================
+
+Overview
+--------
+
+The GoogleSpeechTranscriptionService invokes the Google Speech-to-Text service via REST API to translate audio to text.
```
gregorydlogan: `translate` should perhaps be `transcribe`?

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
...
+
+During the execution of an Opencast workflow, an audio file is extracted from one of the presenter videos and sent 
+to the Google Speech-to-Text service. When the results are received, they are converted to the desired caption format 
+and attached to the media package.
+
```
gregorydlogan, line 12: To me it was unclear that this is an async operation - at least from a naive user's point of view.  This could be clarified with something like:

`Because Google's Speech-to-Text service can take a while to process a recording, we do not wait for it to finish before proceeding with the rest of Opencast's normal processing.  Logically, this looks something like this:`

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
...
+    * File with results is converted and attached to media package
+    * Media package is republished with captions/transcripts
+
+Google Speech-to-Text service documentation, including which languages are currently supported, can be found
+ [here](https://cloud.google.com/speech-to-text/docs/basics).
+
+Configuration
+-------------
+
+### Step 1: Activate Google Speech and Google Cloud Storage APIs
```
gregorydlogan, line 35: So, my experience writing the docs for the AWS integration tells me that by the time someone's actually reading these docs the UI will have changed *drastically*, or been removed completely.  A warning here saying that this is how it's currently working and to consult Google's documentation for $keywords is probably warranted.

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
...
+* Use **_OAuth Client ID_**, **_OAuth Client secret_**, **_Refresh token_**, **_Token endpoint_** and **_storage bucket_** created above to respectively set _google.cloud.client.id_ , _google.cloud.client.secret_ , _google.cloud.refresh.token_ , _google.cloud.token.endpoint.url_ and _google.cloud.storage.bucket_
+* Enter the appropriate language in _google.speech.language_, default is (_en-US_). List of supported language: [https://cloud.google.com/speech-to-text/docs/languages](https://cloud.google.com/speech-to-text/docs/languages)
+* Remove profanity (bad language) from transcription by using _google.speech.profanity.filter_, default is (_false_), not removed by default
+* In _workflow_, enter the workflow definition id of the workflow to be used to attach the generated transcripts/captions
+* Enter a _notification.email_ to get job failure notifications. If not entered, the email in _etc/custom.properties_ (org.opencastproject.admin.email) will be used.
+If no email address specified in either _notification.email_ or _org.opencastproject.admin.email_,
+email notifications will be disabled. 
+
+```
+# Change enabled to true to enable this service.
```
gregorydlogan, line 113: This example of the file should either match the instructions above, or be removed IMO.

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
...
+
+# Email to send notifications of errors. If not entered, the value from
+# org.opencastproject.admin.email in custom.properties will be used.
+notification.email=localadmin@domain
+```
+
+### Step 4: Add encoding profile for extracting audio
+
+The Google Speech-to-Text service has limitations on audio types. [Supported audio type are here](https://cloud.google.com/speech-to-text/docs/reference/rest/v1p1beta1/RecognitionConfig#AudioEncoding).
+Try using the encoding profile suggested in etc/encoding/googlespeech-audio.properties.
```
gregorydlogan: Doesn't this PR include those settings?  If they're in the encodings directory there's no action required.

`Try using...` -> `By default Opencast will use the encoding settings in $file`

docs/guides/admin/docs/modules/googlespeechtranscripts.md
```diff
@@ -0,0 +1,251 @@
...
+        <configuration key="transcription-job-id">${transcriptionJobId}</configuration>
+        <configuration key="line-size">80</configuration>
+        <configuration key="target-flavor">captions/timedtext</configuration>
+        <configuration key="target-tag">archive</configuration>
+        <configuration key="target-caption-format">vtt</configuration>
+      </configurations>
+    </operation>
+
+    <!-- Merge caption/transcript to existing publication and republish -->
+
```
gregorydlogan: Is there an operation missing here?  If not, this comment should be removed.

docs/guides/admin/docs/workflowoperationhandlers/google-speech-attach-transcription-woh.md
```diff
@@ -0,0 +1,63 @@
...
+```xml
+ <!-- Attach caption/transcript -->
+
+    <operation id="google-speech-attach-transcription"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error" 
+      description="Attach captions/transcription">
+      <configurations>
+        <!-- This is filled out by the transcription service when starting this workflow -->
+        <configuration key="transcription-job-id">${transcriptionJobId}</configuration>
```
gregorydlogan, line 29: If this parameter's value can't change without breaking things, why are we exposing it to the user?  Is there a usecase where changing this value is useful?

docs/guides/admin/docs/workflowoperationhandlers/google-speech-start-transcription-woh.md
```diff
@@ -0,0 +1,61 @@
...
+text.
+
+## Parameter Table
+
+|configuration keys|description|default value|example|
+|------------------|-------|-----------|-------------|
+|source-flavor|The flavor of the audio file to be sent for translation.|EMPTY|presenter/delivery|
+|source-tag|The flavor of the audio file to be sent for translation.|EMPTY|transcript|
+|skip-if-flavor-exists|If this flavor already exists in the media package, skip this operation.<br/>To be used when the media package already has a transcript file. Optional|false|captions/timedtext|
+|language-code|The language code to use for the transcription. Optional. If set, it will override the configuration language code|EMPTY|en-US|
```
gregorydlogan: Does Google have a link for valid language codes?

etc/org.opencastproject.transcription.googlespeech.GoogleSpeechTranscriptionService.cfg
```diff
@@ -0,0 +1,48 @@
+# Change enabled to true to enable this service.
+enabled=false
+
+# Google Cloud Service details
+google.cloud.client.id=
```
gregorydlogan, line 5: Config keys with default values set should be commented out.  The exception is for things like `enabled` above, which can remain uncommented.

etc/workflows/google-speech-attach-transcripts.xml
```diff
@@ -0,0 +1,79 @@
...
+        <!-- This is filled out by the transcription service when starting this workflow -->
+        <configuration key="transcription-job-id">${transcriptionJobId}</configuration>
+        <configuration key="line-size">80</configuration>
+        <configuration key="target-flavor">captions/timedtext</configuration>
+        <configuration key="target-tag">archive</configuration>
+        <configuration key="target-caption-format">vtt</configuration>
+      </configurations>
+    </operation>
+
+    <!-- Merge caption/transcript to existing publication and republish -->
```
gregorydlogan: Same as the docs above, is there a missing step here?

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
```diff
@@ -0,0 +1,188 @@
...
+  private static final int LINE_SIZE = 100;
+
+  @Override
+  public List<Caption> importCaption(InputStream inputStream, String languageLineSize) throws CaptionConverterException {
+    List<Caption> captionList = new ArrayList<Caption>();
+    JSONParser jsonParser = new JSONParser();
+    int transcriptionLineSize = 0;
+    try {
+      // No language to specify so define size of a transcripts line
+      transcriptionLineSize = Integer.parseInt(languageLineSize.trim());
```
gregorydlogan, line 64: Probably makes sense to log the size selected here as well.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
```diff
@@ -0,0 +1,188 @@
...
+    int h = (int) (ms / 3600000L);
+    int m = (int) ((ms % 3600000L) / 60000L);
+    int s = (int) ((ms % 60000L) / 1000L);
+    ms = (int) (ms % 1000);
+
+    return new TimeImpl(h, m, s, (int) ms);
+  }
+
+  private String removeCharacter(String str, char car) {
+    if (str != null && str.length() > 0 && str.charAt(str.length() - 1) == car) {
```
gregorydlogan: `if (str != null && str.endsWith(car)))` ?

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  private static final String DEFAULT_LANGUAGE = "en-US";
+  private static final String GOOGLE_SPEECH_URL = "https://speech.googleapis.com/v1";
+  private static final String GOOGLE_AUTH2_URL = "https://www.googleapis.com/oauth2/v4/token";
+  private static final String REQUEST_PATH = "/speech:longrunningrecognize";
+  private static final String RESULT_PATH = "/operations/";
+  private static final String INVALID_TOKEN = "-1";
+  private static final String PROVIDER = "Google Speech";
+
+  // Global configuration (custom.properties)
+  public static final String ADMIN_URL_PROPERTY = "org.opencastproject.admin.ui.url";
```
gregorydlogan: This key is in `OpencastConstants.java`, the rest (IMO) should be but that's kinda out of scope for this PR :)

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  private static final String ADMIN_EMAIL_PROPERTY = "org.opencastproject.admin.email";
+  private static final String DIGEST_USER_PROPERTY = "org.opencastproject.security.digest.user";
+
+  // Cluster name
+  private static final String CLUSTER_NAME_PROPERTY = "org.opencastproject.environment.name";
```
gregorydlogan: This is already in the Watson version, and should be moved up either into `OpencastConstants.java` or into a translation service constants file.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+
+  private enum Operation {
+    StartTranscription
+  }
+
+  /**
+   * Service configuration options
+   */
+  public static final String ENABLED_CONFIG = "enabled";
+  public static final String GOOGLE_SPEECH_LANGUAGE = "google.speech.language";
```
gregorydlogan, line 169: I feel like the keys with prefixed `google.speech` *really* ought to live in a Google provided jar pulled out of Maven somewhere, but I can't find them in their docs.  Something to look out for with revision 2 of this code I guess.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+        logger.info("Workflow dispatch interval is {} seconds", workflowDispatchInterval);
+        // How long to wait after a transcription is supposed to finish before starting checking
+        Option<String> bufferOpt = OsgiUtil.getOptCfg(cc.getProperties(), COMPLETION_CHECK_BUFFER_CONFIG);
+        if (bufferOpt.isSome()) {
+          try {
+            completionCheckBuffer = Long.parseLong(bufferOpt.get());
+          } catch (NumberFormatException e) {
+            // Use default
+            logger.warn("Invalid configuration for {} : {}. Default used instead: {}",
+                    new Object[]{COMPLETION_CHECK_BUFFER_CONFIG, bufferOpt.get(), completionCheckBuffer});
```
gregorydlogan: I seem to remember @lkiesow went through and changed a ton of logging lines either to this style, or away from it.  Check the rest of the codebase and make sure it matches.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+          }
+        }
+        logger.info("Completion check buffer is {} seconds", completionCheckBuffer);
+        // How long to wait after a transcription is supposed to finish before marking the job as canceled in the db
+        Option<String> maxProcessingOpt = OsgiUtil.getOptCfg(cc.getProperties(), MAX_PROCESSING_TIME_CONFIG);
+        if (maxProcessingOpt.isSome()) {
+          try {
+            maxProcessingSeconds = Long.parseLong(maxProcessingOpt.get());
+          } catch (NumberFormatException e) {
+            // Use default
```
gregorydlogan: This should be logged, since it's a configuration error (either an uncommented blank, or not numeric).

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+          }
+        }
+        logger.info("Maximum time a job is checked after it should have ended is {} seconds", maxProcessingSeconds);
+        // How long to keep result files in the working file repository
+        Option<String> cleaupOpt = OsgiUtil.getOptCfg(cc.getProperties(), CLEANUP_RESULTS_DAYS_CONFIG);
+        if (cleaupOpt.isSome()) {
+          try {
+            cleanupResultDays = Integer.parseInt(cleaupOpt.get());
+          } catch (NumberFormatException e) {
+            // Use default
```
gregorydlogan: Logging here.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+
+        Option<String> optCluster = OsgiUtil.getOptContextProperty(cc, CLUSTER_NAME_PROPERTY);
+        if (optCluster.isSome()) {
+          clusterName = optCluster.get();
+        }
+        logger.info("Environment name is {}", clusterName);
+
+        logger.info("Activated!");
+      } else {
+        logger.info("Service disabled. If you want to enable it, please update the service configuration.");
```
gregorydlogan: In general I would reverse the order of this if-else clause.

```
if (!enabled) {
  logger.info("Service disabled...
  return;
}
<the rest of the method>
```

This avoids indenting the whole thing, and clarifies the logic.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      }
+    } else {
+      throw new IllegalArgumentException("Missing component context");
```
gregorydlogan: I'm not *opposed* to throwing this here since (IMO) it is an error, but does this ever actually happen aside from testing?  IIRC, we don't check for this anywhere else, and if the CC is null something is *very* wrong anyway :)

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      switch (code) {
+        case HttpStatus.SC_OK: // 200
+          accessToken = (String) jsonObject.get("access_token");
+          long duration = (long) jsonObject.get("expires_in"); // Duration in second
+          tokenExpiryTime = (System.currentTimeMillis() + (duration * 1000)); // time in millisecond
+          if (!INVALID_TOKEN.equals(accessToken)) {
+            logger.info("Google Cloud Service access token created");
+            return accessToken;
+          }
+          return INVALID_TOKEN;
```
gregorydlogan: I would be tempted to log an error, throw and exception here specifically saying that the 200 happened but you received bad data from Google.  This is a different case than any non-200 result, which is a (hopefully) temporary failure which should be retried.

If you're getting the invalid token back on an HTTP 200 then something is _very_ wrong.

This also has the side effect of not needing to check whether it's the invalid token everywhere you request it.  Either the token is cached (and good), or you request a new one and it's either good, or Google is broken and you throw an exception.  There's no case where you need it and it *hasn't* be checked immediately before.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      // Delete audio file from Google storage
+      deleteStorageFile(mpId, token);
+
+      // Save results in file system if there exist
+      if (resultsArray != null) {
+        saveResults(jobId, jsonObj);
+      }
+    } catch (IOException e) {
+      logger.warn("Could not save transcription results file for mpId {}, jobId {}: {}",
+              new Object[]{mpId, jobId, jsonObj == null ? "null" : jsonObj.toJSONString()});
```
gregorydlogan: Logging type again.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      throw new TranscriptionServiceException("Could not save transcription results file", e);
+    } catch (TranscriptionDatabaseException e) {
+      logger.warn("Transcription results file were saved but state in db not updated for mpId {}, jobId {}", mpId,
```
gregorydlogan, line 409: I'm assuming this exception is triggered by line 378 (which IMO sounds like a race condition).  If so, this sounds like something that should be ignored (or better yet, doesn't happen at all), except we throw an exception right after this.  What other cases could trigger this exception?

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      database.updateJobControl(jobId, TranscriptionJobControl.Status.Error.name());
+      TranscriptionJobControl jobControl = database.findByJob(jobId);
+      logger.warn(String.format("Error received for media package %s, job id %s",
+              jobControl.getMediaPackageId(), jobId));
+      // Send notification email
+      sendEmail("Transcription ERROR",
+              String.format("There was a transcription error for for media package %s, job id %s.",
+                      jobControl.getMediaPackageId(), jobId));
+    } catch (TranscriptionDatabaseException e) {
+      logger.warn("Transcription error. State in db could not be updated to error for mpId {}, jobId {}", mpId, jobId);
```
gregorydlogan, line 437: Same comment as above, won't continue banging on this.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+    CloseableHttpClient httpClient = makeHttpClient();
+    CloseableHttpResponse response = null;
+    String mpId = "unknown";
+    JSONArray resultsArray = null;
+    String token = getRefreshAccessToken();
+    if (token.equals(INVALID_TOKEN)) {
+      return false;
+    }
+    try {
+      HttpGet httpGet = new HttpGet(GOOGLE_SPEECH_URL + RESULT_PATH + jobId);
```
gregorydlogan: `UrlSupport.concat` has the benefit of handling the `/` characters for you, thus removing them from your constants.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      }
+      throw new TranscriptionServiceException(
+              String.format("Could not check recognition job for media package %s, job id %s. Status returned: %d",
+                      mpId, jobId, code), code);
+    } catch (TranscriptionServiceException e) {
+      throw e;
+    } catch (Exception e) {
+      String msg = String.format("Exception when calling the recognitions endpoint for media package %s, job id %s",
+              mpId, jobId);
+      logger.warn(String.format(msg, mpId, jobId), e);
```
gregorydlogan: `msg` is being formatted twice

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+          throws TranscriptionServiceException, IOException {
+    CloseableHttpClient httpClient = makeHttpClient();
+    CloseableHttpResponse response = null;
+    String token = getRefreshAccessToken();
+    if (token.equals(INVALID_TOKEN)) {
+      logger.warn("Invalid access token");
+      return "No results found";
+    }
+    try {
+      HttpGet httpGet = new HttpGet(GOOGLE_SPEECH_URL + RESULT_PATH + jobId);
```
gregorydlogan: `UrlSupport` again, not going to keep banging on this so check the rest of this PR.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  }
+
+  /**
+   * Asynchronous Requests and Responses call to Google Speech API
+   * https://cloud.google.com/speech-to-text/docs/basics
+   */
+  void createRecognitionsJob(String mpId, Track track, String languageCode) throws TranscriptionServiceException, IOException {
+    // Use default language if not set by workflow
+    if (languageCode == null || languageCode.isEmpty()) {
+      languageCode = language;
```
gregorydlogan: Just realized that `language` here is a service-wide default, and thus should be renamed to `defaultLangauge` or something.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  /**
+   * Get mediapackage transcription status
+   *
+   * @param mpId, mediapackage id
+   * @return transcription status
+   * @throws TranscriptionServiceException
+   */
+  public String getTranscriptionStatus(String mpId) throws TranscriptionServiceException {
+    try {
+      for (TranscriptionJobControl jc : database.findByMediaPackage(mpId)) {
```
gregorydlogan, line 738: So, we're returning one result from this method, but the database can contain multiple job controls.  I'm assuming we support re-transcribing the same mediapackage (say, a retry with a new/better speech model).  But don't we want to get the *latest* job control?  The underlying query here would return them in whatever the default order is for the table, which I'm guessing is actually the opposite of what you're looking for with this method.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+        return jc.getStatus();
+      }
+    } catch (TranscriptionDatabaseException e) {
+      throw new TranscriptionServiceException("Mediapackage id transcription status unknown", e);
+    }
+    return "Unknown";
+  }
+
+  protected CloseableHttpClient makeHttpClient() throws IOException {
```
gregorydlogan, line 747: The wrapping in this method is wonky.  I would do something like
```
RequestConfig reqConfig = RequestConfig.custom()
        .setConnectTimeout(CONNECTION_TIMEOUT)
        .setSocketTimeout(SOCKET_TIMEOUT)
        .setConnectionRequestTimeout(CONNECTION_TIMEOUT)
        .build();
return HttpClients.custom().setDefaultRequestConfig(reqConfig).build();
```

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  }
+
+  protected String refreshAccessToken(String clientId, String clientSecret, String refreshToken)
+          throws TranscriptionServiceException, IOException {
+    CloseableHttpClient httpClient = makeHttpClient();
+    CloseableHttpResponse response = null;
+
+    try {
+      HttpPost httpPost = new HttpPost(tokenEndpoint + String.format(
+              "?client_id=%s&client_secret=%s&refresh_token=%s&grant_type=refresh_token",
```
gregorydlogan, line 763: UrlSupport probably has bits for query params.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+              clientId, clientSecret, refreshToken));
+      httpPost.addHeader(HttpHeaders.CONTENT_TYPE, "application/x-www-form-urlencoded");
+      response = httpClient.execute(httpPost);
+      int code = response.getStatusLine().getStatusCode();
+      String jsonString = EntityUtils.toString(response.getEntity());
+      JSONParser jsonParser = new JSONParser();
+      JSONObject jsonObject = (JSONObject) jsonParser.parse(jsonString);
+      switch (code) {
+        case HttpStatus.SC_OK: // 200
+          accessToken = (String) jsonObject.get("access_token");
```
gregorydlogan: These should be constants.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+  private JSONArray getTranscriptionResult(JSONObject jsonObj) {
+    JSONObject responseObj = (JSONObject) jsonObj.get("response");
+    JSONArray resultsArray = (JSONArray) responseObj.get("results");
+    return resultsArray;
+  }
+
+  protected void deleteStorageFile(String mpId, String token) throws IOException {
+    CloseableHttpClient httpClientDel = makeHttpClient();
+    GoogleSpeechTranscriptionServiceStorage storage = new GoogleSpeechTranscriptionServiceStorage();
+    storage.deleteGoogleStorageFile(httpClientDel, storageBucket, mpId + ".flac", token);
```
gregorydlogan: What if the user changes the encoding settings?  Is this always a `.flac`?

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+    try {
+      jsonObj = (JSONObject) obj;
+      jobId = (String) jsonObj.get("name");
+      // Update state in database
+      database.updateJobControl(jobId, TranscriptionJobControl.Status.Error.name());
+      TranscriptionJobControl jobControl = database.findByJob(jobId);
+      logger.warn(String.format("Error received for media package %s, job id %s",
+              jobControl.getMediaPackageId(), jobId));
+      // Send notification email
+      sendEmail("Transcription ERROR",
```
gregorydlogan: I would be tempted to turn most of the parameters for your usage of `sendEmail` into constants.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+     *
+     * @see java.lang.Thread#run()
+     */
+    @Override
+    public void run() {
+      logger.debug("WorkflowDispatcher waking up...");
+
+      try {
+        // Find jobs that are in progress and jobs that had transcription complete
+        List<TranscriptionJobControl> jobs = database.findByStatus(TranscriptionJobControl.Status.Progress.name(),
```
gregorydlogan: IMO, `findBy*` which takes `Status` enum types should directly take the type rather than Strings.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+                TranscriptionJobControl.Status.TranscriptionComplete.name());
+        for (TranscriptionJobControl j : jobs) {
+          String mpId = j.getMediaPackageId();
+          String jobId = j.getTranscriptionJobId();
+
+          // If the job in progress, check if it should already have finished.
+          if (TranscriptionJobControl.Status.Progress.name().equals(j.getStatus())) {
```
gregorydlogan: I would rename `Progress` to be `InProgress`, otherwise this is quite confusing unless read aloud.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+            // If job should already have been completed, try to get the results. Consider a buffer factor so that we
+            // don't try it too early.
+            if (j.getDateCreated().getTime() + j.getTrackDuration() + completionCheckBuffer * 1000 < System
```
gregorydlogan: Aside: They're saying they take at least real time?  I would have expected that they could process a track faster than real-time....

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+              try {
+                if (!getAndSaveJobResults(jobId)) {
+                  // Job still running, not finished, so check if it should have finished more than N seconds ago
+                  if (j.getDateCreated().getTime() + j.getTrackDuration()
+                          + (completionCheckBuffer + maxProcessingSeconds) * 1000 < System.currentTimeMillis()) {
+                    // Processing for too long, mark job as cancelled and don't check anymore
+                    database.updateJobControl(jobId, TranscriptionJobControl.Status.Canceled.name());
+                    // Delete file stored on Google storage
+                    String token = getRefreshAccessToken();
+                    deleteStorageFile(mpId, token);
```
gregorydlogan, line 976: We don't want to optionally initiate a retry here?

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+              continue; // Not time to check yet
+            }
+          }
+
+          // Jobs that get here have state TranscriptionCompleted
+          try {
+
+            // Apply workflow to attach transcripts
+            Map<String, String> params = new HashMap<String, String>();
+            params.put("transcriptionJobId", jobId);
```
gregorydlogan: Constant

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -0,0 +1,1064 @@
...
+      }
+    } catch (NotFoundException e) {
+      logger.warn("Organization {} not found for media package {}.", org, mpId);
+      return null;
+    }
+    securityService.setOrganization(organization);
+
+    try {
+      WorkflowDefinition wfDef = workflowService.getWorkflowDefinitionById(wfDefId);
+      Workflows workflows = wfUtil != null ? wfUtil : new Workflows(assetManager, workspace, workflowService);
```
gregorydlogan: We try and avoid ternaries, although this is not enforced by checkstyles.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/endpoint/GoogleSpeechTranscriptionRestService.java
```diff
@@ -0,0 +1,130 @@
...
+    @RestResponse(responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR, description = "An error occurred")})
+  // GET v1/operations/7022084740052439959
+  public Response getTranscriptionResult(@QueryParam("jobId") String jobId,
+          @Context HttpServletRequest request) throws TranscriptionServiceException, IOException {
+    logger.debug("REST endpoint getTranscriptionResult called with job id: '{}'", jobId);
+    String jobResult = null;
+    try {
+      jobResult = service.getTranscriptionResults(jobId);
+    } catch (TranscriptionServiceException e) {
+      throw new TranscriptionServiceException("Could not get transcription result", e);
```
gregorydlogan: IIRC we need to build a Response object still, or we get an ugly stack trace in the browser in this case.  Check some of the other common services to see.  This comment applies to all of the REST endpoints in this PR.

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionDatabaseImpl.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  @Override
+  public TranscriptionProviderControl findProviderById(Long id) throws TranscriptionDatabaseException {
+    TranscriptionProviderControlDto dtoProvider = TranscriptionProviderControlDto.findProviderById(emf.createEntityManager(), id);
+    if (dtoProvider != null) {
+      return dtoProvider.toTranscriptionProviderControl();
+    }
+    return null;
```
gregorydlogan, line 156: What's the advantage here vs throwing an exception?  If someone's asking for an invalid/missing provider then there's a fairly major error.

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionDatabaseImpl.java
```diff
@@ -0,0 +1,166 @@
...
+    if (dtoProvider != null) {
+      return dtoProvider.toTranscriptionProviderControl();
+    } else {
+      // store provider and retrieve id
+      TranscriptionProviderControl dto = storeProviderControl(provider);
+      if (dto != null) {
+        dtoProvider = TranscriptionProviderControlDto.findIdByProvider(emf.createEntityManager(), provider);
+        return dtoProvider.toTranscriptionProviderControl();
+      }
+      return null; // not found
```
gregorydlogan, line 146: Should be an exception IMO, see comment below.

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionJobControlDto.java
```diff
@@ -0,0 +1,246 @@
...
+  @NamedQuery(name = "TranscriptionJobControl.findByStatus", query = "SELECT jc FROM TranscriptionJobControl jc WHERE jc.status IN :status")})
+public class TranscriptionJobControlDto implements Serializable {
+
+  @Id
+  @GeneratedValue(strategy = GenerationType.AUTO)
+  @Column(name = "id", length = 128)
+  @XmlAttribute
+  private long id;
+
+  @Column(name = "media_package_id", nullable = false, length = 128)
```
gregorydlogan: `media_package_id` should be `mediapackage_id`.  You may have gotten this from my AWS S3 dist code, or from the Watson code.  Both sets are wrong :D  I'll file a PR shortly.

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionJobControlDto.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * Find all job controls by status.
+   */
+  @SuppressWarnings("unchecked")
+  public static List<TranscriptionJobControlDto> findByStatus(EntityManager em, final String... status)
+          throws TranscriptionDatabaseException {
+    Collection<String> statusCol = new HashSet<String>();
+    for (String st : status) {
```
gregorydlogan, line 179: I see what you're doing, and understand why.  I would still break this into two methods, one which takes a single Status, the other which takes a Set<Status>.  The single input would wrap its input into a Set, then call the Set input to keep the logic in one place.

Hint: What if status is `null`?

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionProviderControlDto.java
```diff
@@ -0,0 +1,134 @@
...
+   */
+  public static TranscriptionProviderControlDto findProviderById(EntityManager em, long id)
+          throws TranscriptionDatabaseException {
+    Query query = null;
+    try {
+      query = em.createNamedQuery("TranscriptionProvider.findProviderById");
+      query.setParameter("id", id);
+      return (TranscriptionProviderControlDto) query.getSingleResult();
+    } catch (NoResultException e) {
+      return null; // Not found
```
gregorydlogan, line 111: I would throw a `NotFoundException` here, which we use in a few places for things like this.

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechAttachTranscriptionOperationHandler.java
```diff
@@ -0,0 +1,178 @@
...
+        transcription = MediaPackageElementParser.getFromXml(job.getPayload());
+      }
+
+      // Set the target flavor
+      transcription.setFlavor(flavor);
+
+      // Add tags
+      if (targetTagOption != null) {
+        for (String tag : asList(targetTagOption)) {
+          if (StringUtils.trimToNull(tag) != null) {
```
gregorydlogan: `!StringUtils.isBlank` would be a better fit here.

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechStartTranscriptionOperationHandler.java
```diff
@@ -0,0 +1,185 @@
...
+    if (StringUtils.isNotBlank(sourceFlavorOption)) {
+      String flavor = StringUtils.trim(sourceFlavorOption);
+      try {
+        elementSelector.addFlavor(MediaPackageElementFlavor.parseFlavor(flavor));
+      } catch (IllegalArgumentException e) {
+        throw new WorkflowOperationException("Source flavor '" + flavor + "' is malformed");
+      }
+    }
+    if (StringUtils.isNotBlank(langCode)) {
+      language = StringUtils.trim(langCode);
```
gregorydlogan, line 142: `langCode` has already been trimmed to null above.  IMO line 120 should read `language = ...` since `language` as implemented ends up as either the non-blank `langCode` value, or `null`.

gregorydlogan (changes requested):
This looks really good, thanks for the hard work @tyfranck.  I know this looks like a ton of complaints, but the majority is style and nitpicks.  Ideally the database bits would be broken out into their own PR, but at this point I've gone through it all so we might as well leave it as is :)

tyfranck:
@gregorydlogan thanks for the reviews and comments, really useful. 
I am not clear or don't understand one or two but will get back to you in due course.

smarquard:
@tyfranck, I was wondering what the value is of the provider table. Is there a possibility that we may store may provider-related info in this table in future? Otherwise it would seem simpler to just have a string providerid in the transcription jobs table.

tyfranck:
@smarquard, yes, the provider table provides the ability to add more infos about providers in the future if needed. It also allows those who sign up to many providers (google, ibm) to easily identify them.  I can see Amazon, Microsoft being added to the list in the future. So having a human readble way to identify them is necessary.
providerId only, I agree is simpler but might be an issue later on when it comes to identify who is who.
I am opened for suggestions though.

gregorydlogan:
I really like the idea of the provider table.  We're going to have multiple implementations as soon as this goes in anyway, we should get ahead of the technical debt and do it right the first time :)

KatrinIhler:
@tyfranck Conflicts in the database migration script!

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechAttachTranscriptionOperationHandler.java
lkiesow: [`StringUtils.isNotBlank(…)`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNotBlank-java.lang.CharSequence-) ;-)

lkiesow:
Note that this pull request has conflicts in `docs/upgrade/6_to_7/mysql5.sql￼` which should be `docs/upgrade/7_to_8/mysql5.sql￼`anyway since this will be an Opencast 8 feature

docs/guides/admin/docs/workflowoperationhandlers/google-speech-attach-transcription-woh.md
tyfranck: This workflow is not accessible by user via admin. Only admin can access it via workflow directory.  And if they know what they're doing they can set it to attach a given job id.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
tyfranck: "car" for character :-) , this code is to remove last character from the provided google timestamp which has the format: 100s.  See line 117 on how it's used: (removeCharacter((wordTSList.get("endTime").toString()), 's'))
@gregorydlogan  do you want me to write it differently?

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: @lkiesow can you please advice on the logging style @gregorydlogan is mentioning in his comment above.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: yes line 378, it could be a race condition but main motivation is for example if Google storage services or its API's are inaccessible at the time of deletion, job should not be mark as completed to avoid orphan file on google storage.

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechAttachTranscriptionOperationHandler.java
gregorydlogan: @lkiesow has a better solution :)

docs/guides/admin/docs/workflowoperationhandlers/google-speech-attach-transcription-woh.md
gregorydlogan: Ok, fair enough.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
gregorydlogan: @tyfranck My initial comment was my suggestion :)

You're checking if it's null, that its length is greater than zero, and then if the last character matches.  I'm checking if it's null, then if it ends with a given character.  I would even simplify my initial comment with

`if (str.endsWith(car))`, since I'm 99% sure `.endsWith` checks for nulls.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
gregorydlogan: Ah, good point.  Ok, let's leave this then.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
tyfranck: I get it now :-) . Completely misread it.

KatrinIhler:
@tyfranck Test failures!

gregorydlogan:
@tyfranck, any idea what the ETA for movement on this would be?

tyfranck:
@gregorydlogan, working my way through the comments and addressing them when possible. Most of the trivial changes have been done locally. Some comments require more thinking and have knock on effect on other part so need a bit more time to look into them. I'll spend some time on it this week and merge some of the new changes. The plan is to get it done by the end of this month.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: The underlying query order the results by date created. This is specify in  'TranscriptionJobControlDto.java' in the database module. The latest result should be the first one.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Nope, UrlSupport doesn't seem to support query params

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Good point I'll put this in the config file and set default as 'flac'

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: yes indeed.  I've set it to half of of the track duration

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Possible but not part of the initial release, I'll leave it for the next version. 
It needs couple of tweaks and designed properly..

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionDatabaseImpl.java
tyfranck: The advantage here is that if provider is missing then create it. 
Providers are not created by default so we don't want this throwing an exception/ending the application, we want to create the provider instead and continue the application.

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionDatabaseImpl.java
tyfranck: same reply as above

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionProviderControlDto.java
tyfranck: Same reply as previously explained

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionJobControlDto.java
tyfranck: I don't understand the 2nd sentence. 
The current implementation will handle both single status and a set of statuses.  I've added a check for null status. 
What's the reasoning behind splitting into 2 methods?

gregorydlogan: Right now, if I were to call this method with a null as the final parameter, I would see a null pointer exception on 176 when Java attempts to iterate on a null.  I don't see a null check, but that might be a local change that's not here yet.  If it's checked we can leave it as is.

docs/guides/admin/docs/modules/googlespeechtranscripts.md
tyfranck: fixed

docs/guides/admin/docs/modules/googlespeechtranscripts.md
tyfranck: I've added a note

docs/guides/admin/docs/modules/googlespeechtranscripts.md
tyfranck: fixed

docs/guides/admin/docs/modules/googlespeechtranscripts.md
tyfranck: fixed

docs/guides/admin/docs/workflowoperationhandlers/google-speech-start-transcription-woh.md
tyfranck: link added

etc/org.opencastproject.transcription.googlespeech.GoogleSpeechTranscriptionService.cfg
tyfranck: fixed

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: yes, something to keep an eye on but doesn't currently exist

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Not sure what to change here

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/endpoint/GoogleSpeechTranscriptionRestService.java
tyfranck: fixed

modules/transcription-service-persistence/src/main/java/org/opencastproject/transcription/persistence/TranscriptionJobControlDto.java
tyfranck: changed was local, I've now pushed it.

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechAttachTranscriptionOperationHandler.java
tyfranck: fixed

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/GoogleSpeechStartTranscriptionOperationHandler.java
tyfranck: fixed

tyfranck:
@gregorydlogan, apologies for the delay but I've now addressed your comments and pushed all the changes after a bit of fire fighting with various conflict :-).
In addition the pull request have successfully passed all automated checks/build stage.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
```diff
@@ -0,0 +1,189 @@
...
+
+  @Override
+  public List<Caption> importCaption(InputStream inputStream, String languageLineSize) throws CaptionConverterException {
+    List<Caption> captionList = new ArrayList<Caption>();
+    JSONParser jsonParser = new JSONParser();
+    int transcriptionLineSize = 0;
+    try {
+      // No language to specify so define size of a transcripts line
+      transcriptionLineSize = Integer.parseInt(languageLineSize.trim());
+      logger.info("Transcripts line size {} used", LINE_SIZE);
```
gregorydlogan: This is logging the default, not the parsed line size

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
gregorydlogan: Is there a reason this wasn't changed if you addressed the other ones?

gregorydlogan:
@tyfranck two final nits to pick then we're done!

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/GoogleSpeechCaptionConverter.java
tyfranck: fixed

PR #824, opened 2019-04-01
MH-13381, Use Username In Workflows

lkiesow:
Instead of using a serialized version of a user object which can be
potentially huge, use a username for identifying creators.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
@@ -39,6 +39,7 @@ import org.opencastproject.matterhorn.search.SearchQuery;
 import org.opencastproject.matterhorn.search.SortCriterion;
 import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.security.api.User;
+import org.opencastproject.security.api.UserDirectoryService;
 import org.opencastproject.serviceregistry.api.IncidentL10n;
 import org.opencastproject.serviceregistry.api.IncidentService;
 import org.opencastproject.serviceregistry.api.IncidentServiceException;
@@ -123,6 +124,7 @@ public class JobEndpoint {
   private WorkflowService workflowService;
   private ServiceRegistry serviceRegistry;
   private IncidentService incidentService;
+  private UserDirectoryService userDirectoryService;
 
   /** OSGi callback for the workflow service. */
   public void setWorkflowService(WorkflowService workflowService) {
@@ -137,6 +139,10 @@ public class JobEndpoint {
   /** OSGi callback for the incident service. */
   public void setIncidentService(IncidentService incidentService) {
     this.incidentService = incidentService;
+  }
+
+  public void setUserDirectoryService(UserDirectoryService userDirectoryService) {
+    this.userDirectoryService = userDirectoryService;
   }
 
   protected void activate(BundleContext bundleContext) {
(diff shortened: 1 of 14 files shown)


staubesv:
@lkiesow Hmm... I see the point that just serializing the Java object User is not necessarily a good idea as this potentially includes much more information than required. I do have the following concerns here, though:

- At least in our case, we will never(!) delete the workflow instances from the database as we want to know the processing history of every event over all the event lifecycle (yes, we do have events that likely will "never" be deleted). So can we expect the legacy field "user" to stay forever? Maybe add a comment to that field. The cleaner way would be to migrate the data...
- In our case, we mainly need the user information for Event Details->Workflows and in the future maybe for a new view "Systems->Workflows". The later would need this data (at least the name of the user) for approx. 100 workflows (-> pagination). Will this still be fast enough with additional 100 calls to the user directory?

Note that in Shibboleth environments, there is no such thing as a username (the field is used to map a technical user identifier), so this might be less critical in non-Shibboleth environments. Anyway, displaying the name of the user seems more "human-readable" than displaying usernames even in non-Shibboleth environments. The e-mail address is less critical as it is detail information typically not displayed in lists.

smarquard:
@staubesv, on your specific concerns:

* The proposed work being done to remove the solr workflow index may introduce a new table with one row per workflow, so you could then delete the job info with the bloated xml (assuming you don't want to keep the history of every operation in every job, rather just that the job ran).

* The User Directory should cache user info, so I don't think performance would be an issue. Event Details > Workflows would typically be a fairly small number of workflows per event. If Systems > Workflows (a new view) has particular performance implications, than that should be considered as part of implementing that feature. Maybe a new workflow table will solve that, or maybe the admin UI ES index needs to index that in a particular way to support fast searching or sorting.

smarquard:
Can we target this at 7.x? It is a performance optimization (reducing memory usage and GC) rather than feature.

smarquard:
Tested by merging this into a (somewhat custom) 6.x dev build, and it has the desired effect without appearing to cause any regressions.

lkiesow:
@staubesv, this patch should not be a problem for getting user information for individual workflows and if you start building admin interface tables with hundreds per view: Please refrain from loading all data from large XML structures. That's not a good idea ;-)

My suggestion would be to add the data you need to the Elasticsearch index used for the admin interface. We put some workflow information there already. Extending that and adding what you need would mean that you could easily retrieve and even search in lots of data efficiently.
 
---

We could add a migration script updating all workflows or do it in code. Would you prefer one way over the other? In any case, if we do that, I would do that as well for the organization information we handled in the same manner in an earlier pull request.

---

Note that I'm on vacation for the next two weeks.

I specifically allowed edits from maintainers. Hence, feel free to modify anything that needs modification if you do not want to wait or contact someone else at [ELAN e.V.](mailto:opencast-support@elan-ev.de) to fix problems.

staubesv:
@lkiesow 

> My suggestion would be to add the data you need to the Elasticsearch index used for the admin interface. We put some workflow information there already. Extending that and adding what you need would mean that you could easily retrieve and even search in lots of data efficiently.

The Admin UI search index does not contain workflow instances and actually also should not contain those (we don't do full text searches on workflow instances, do we?).
But let's hope that the workflows will be represented well in the database at some point of time as it would actually be quite trivial to retrieve the required information using the database if we had a reasonable database model. But that is out-of-scope considering this PR ;-)

> We could add a migration script updating all workflows or do it in code. Would you prefer one way over the other?

Providing a migration is the correct way as just leaving the old@XmlElemen "creator" field for compatibility reasons indeed raises the question at which point of time it finally will be removed. So the migration clearly is my preference.

smarquard:
I would suggest merging this as-is and doing the migration (for this and the organization field) in a new issue, mostly because these commits can be cherry-picked back to 6.x to improve memory usage and performance without requiring migration.

For the migration, the easiest may be:
https://dev.mysql.com/doc/refman/5.7/en/xml-functions.html#function_updatexml

smarquard (approved):
We are running this change successfully in our 6.x production system.

PR #769, opened 2019-03-12
MH-13431 Remove unfinished feature "Bulk Messaging"

staubesv:
Opencast contains quite a bit of code to the unfinished feature "Bulk messaging". The idea behind that feature is that Opencast allows users to send e-mails to some technical contact of events.

The feature would include:

- Management of e-mail templates
- Management of e-mail signatures
- Persistence of SMTP configurations (per tenant)
- Bulk action "Send message" to actually send messages

The goal of this task is to removed that unfinished feature from the Opencast codebase.

PR #825, opened 2019-04-01
MH-13480 - Remove leftover service

isabellakutger:
The service has been removed in #769 but the service definition was left over.


staubesv (approved):
Thanks!

PR #826, opened 2019-04-03
url query string incorrect

kristofkeppens:
The url query string should use ? instead of ; to work in an lti context.

### DIFF
--- a/docs/guides/admin/docs/modules/ltimodule.md
+++ b/docs/guides/admin/docs/modules/ltimodule.md
@@ -100,8 +100,8 @@ custom parameters to be defined separately in each place where an LTI tool is us
 custom parameters to be defined globally.
 
 * To show the Opencast Media Module, use `tool=engage/ui/`
-* To show all videos for a single series, use `tool=ltitools/series/index.html;series=SERIESID`
-* To show a single video, use `tool=engage/theodul/ui/core.html;id=MEDIAPACKAGEID`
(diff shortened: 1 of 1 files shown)


lkiesow:
Looks good. Two open questions:

- Does it make sense to already have this in the 6.x documentation?
- Does it make sense to change `tool=engage/theodul/ui/core.html?id=MEDIAPACKAGEID` to `tool=/play/MEDIAPACKAGEID`? In 7.x, this will automatically redirect you to the configured default player even if you later decide to change it.

kristofkeppens:
This change is definitely relevant for 6.x as well, you need to add the tool parameter the same way then. 
For 7.x it would be an improvement to use the /play/MEDIAPACKAGEID instead of the player specific url. I wasn't aware of this change in 7.x, but good to know so I can change the LTI deep linking integration to use this url as well.

Should I make 2 different pull requests, 1 to develop and 1 to 6.x to add this?

lkiesow:
Please do if you don't want me to do the work ;-)

kristofkeppens:
@lkiesow I'll make a new PR for the 6.x branch without the /play/MEDIAPACKAGE url. This change can go into r/7.x as well.

lkiesow (approved):

lkiesow:
Thanks for the patch. Merged in `develop` and cherry-picked to `r/7.x`

PR #829, opened 2019-04-05
MH-13486 Cleanup NOTICES 6.x

staubesv:
Considering the JavaScript libraries used in the module admin-ui, NOTICES contains some incorrect directories and is not sorted alphabetically. Some notices are missing, too.

Note that there will be a follow-up PR against develop which requires some more adjustments.


gregorydlogan:
Hi @staubesv, was this done manually, or do you have an automated way to do this?

staubesv:
@gregorydlogan Manually.

PR #830, opened 2019-04-08
MH-13482 Prevent incorrect local job load total

smarquard:
A worker node that has been restarted after an unclean shutdown can have an incorrect local job load value.

On startup (activation of the serviceregistry) set the job load to 0.

If the job load cache has no entries (which should correspond to no running jobs) and the job load is non-zero, reset it to 0 if it's non-zero and log a warning (because this situation should not arise).


smarquard:
Needs a small change to avoid floating-point errors like this:

/opt//opencast/data//log/opencast.log:2019-04-08 19:18:18,203 | WARN  | pool-28-thread-5 | (ServiceRegistryJpaImpl:953) - No jobs in the job load cache, but load is 0.0: resetting job load to 0

smarquard:
Fixed. Please squash commits on merge.

gregorydlogan:
This was squashed by c985171069bc01d06c9cd2093b56c149360b624d, but I clearly didn't do something right since GitHub still wants to do a noop merge.  I'm going to close this to keep this history cleaner, but future readers should refer to the commit above for this changeset.

PR #633, opened 2019-01-02
MH-13293 Add Canvas user and role provider

paulgration:
This adds a Canvas user and role provider to Opencast analogous to the existing Sakai and Moodle user providers.

PR #831, opened 2019-04-10
MH-13488 userprovider for the d2l brightspace LMS

kristofkeppens:
A userprovider for the d2l brightspace LMS, similar to the moodle and sakai userprovider. 
It uses the brightspace API to retrieve the courses a user is enrolled in and uses this to create extra roles for the user.


modules/userdirectory-brightspace/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+    <name>Opencast :: userdirectory-brightspace</name>
+    <parent>
+        <groupId>org.opencastproject</groupId>
+        <artifactId>base</artifactId>
+        <version>8-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <properties>
+        <opencast.basedir>${project.basedir}/../..</opencast.basedir>
+        <checkstyle.skip>true</checkstyle.skip>
```
gregorydlogan: This needs to be removed, and checkstyles needs to pass.

modules/userdirectory-brightspace/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>be.ugent.brightspace</groupId>
+            <artifactId>valence-sdk-java</artifactId>
+            <version>1.0.1</version>
```
gregorydlogan: This version should move up into the main pom.

modules/userdirectory-brightspace/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+            <version>1.10.19</version>
```
gregorydlogan: Version up to main pom.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+
+    Organization org;
+    try {
+      org = this.orgDirectory.getOrganization(organization);
+    } catch (NotFoundException nfe) {
+      logger.warn("Organization {} not found!", organization);
+      throw new ConfigurationException(ORGANIZATION_KEY, "not found");
+    }
+
+    logger.debug("creating new MoodleUserProviderInstance for pid=" + pid);
```
gregorydlogan: MoodleUserProviderInstance?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+
+  @Override
+  public Iterator<Role> findRoles(String query, Target target, int offset, int limit) {
+    return Collections.emptyIterator();
+  }
+
+  private User loadUserFromBrightspace(String username) {
+    if (this.cache == null) {
+      throw new IllegalStateException("The Brightspace user detail service has not yet been configured");
+    } else if (!"admin".equals(username) && !"".equals(username) && !"anonymous".equals(username)) {
```
gregorydlogan: Magic strings should be constants IMO.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+
+          return new JaxbUser(brightspaceUser.getUserName(), null, brightspaceUser.getDisplayName(),
+                  brightspaceUser.getExternalEmail(), this.getName(), true, jaxbOrganization, roles);
+        } else {
+          cache.put(username, nullToken);
+          logger.debug("User {} not found in Brightspace system", username);
+          return null;
+        }
+      } catch (BrightspaceClientException e) {
+        logger.error("User {} not found in Brightspace system", username);
```
gregorydlogan: Is this really that the user is not found?  This seems more like a communication error to me.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+
+  private User loadUserFromBrightspace(String username) {
+    if (this.cache == null) {
+      throw new IllegalStateException("The Brightspace user detail service has not yet been configured");
+    } else if (!"admin".equals(username) && !"".equals(username) && !"anonymous".equals(username)) {
+
+      JaxbOrganization jaxbOrganization = JaxbOrganization.fromOrganization(organization);
+
+      this.brightspaceWebServiceRequests.incrementAndGet();
+      Thread currentThread = Thread.currentThread();
```
gregorydlogan, line 253: Why are we messing with the classloader?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/BrightspaceClient.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.opencastproject.userdirectory.brightspace.client.api.BrightspaceUser;
+import org.opencastproject.userdirectory.brightspace.client.api.OrgUnitItem;
+
+import java.util.List;
+import java.util.Set;
+
+public interface BrightspaceClient {
+
+  BrightspaceUser findUser(String username) throws BrightspaceClientException;
```
gregorydlogan, line 32: Ideally these would have javadocs.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/api/BrightspaceUser.java
```diff
@@ -0,0 +1,103 @@
...
+  private String middleName;
+  private String lastName;
+  private String userName;
+  private String externalEmail;
+  private String orgDefinedId;
+  private String uniqueIdentifier;
+  private Activation activation;
+  private String displayName;
+
+  @JsonCreator
```
gregorydlogan, line 41: We're using a different library for these kinds of annotations across the rest of the codebase.  Any chance this can be adapted to use that library?  I can see this kind of module not getting security updates since it's using a different library that we might not be checking as closely.

modules/userdirectory-brightspace/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <version>1.3</version>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+    <build>
+        <plugins>
```
lkiesow: Can we maybe include the `maven-dependency-plugin` to automatically check for broken dependencies? Here is an example including a few configured exceptions:

https://github.com/opencast/opencast/blob/3bc4af82ce74b3270b117f5bbacd93061714a728/modules/workspace-impl/pom.xml#L105-L117

This isn't strictly necessary but we have lately started to introduce this to catch dependency problems early.

modules/userdirectory-brightspace/pom.xml
```diff
@@ -0,0 +1,153 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>opencast-userdirectory-brightspace</artifactId>
+    <packaging>bundle</packaging>
+    <name>Opencast :: userdirectory-brightspace</name>
+    <parent>
+        <groupId>org.opencastproject</groupId>
```
lkiesow: Not extremely important but all other `pom.xml` files use two spaces for indentation ;-)

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
lkiesow:
```suggestion
    logger.debug("creating new brightspace user provider for pid={}", pid);
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.osgi.service.cm.ManagedServiceFactory#deleted(java.lang.String)
+   */
+  @Override
+  public void deleted(String pid) {
+    logger.debug("delete BrightspaceUserProviderInstance for pid=" + pid);
```
lkiesow:
```suggestion
    logger.debug("delete BrightspaceUserProviderInstance for pid={}", pid);
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+    ServiceRegistration registration = providerRegistrations.remove(pid);
+    if (registration != null) {
+      registration.unregister();
+
+      try {
+        ManagementFactory.getPlatformMBeanServer().unregisterMBean(getObjectName(pid));
+      } catch (Exception var4) {
```
lkiesow:
```suggestion
      } catch (Exception e) {
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+        logger.warn("Unable to unregister mbean for pid='{}': {}", pid, var4.getMessage());
```
lkiesow:
```suggestion
        logger.warn("Unable to unregister mbean for pid='{}'", pid, e);
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+      }
+    }
+  }
+
+  private int parseCacheExpirationProperty(Dictionary properties) {
+    try {
+      if (properties.get(CACHE_EXPIRATION) != null) {
+        return Integer.parseInt(properties.get(CACHE_EXPIRATION).toString());
+      }
+    } catch (NumberFormatException nfe) {
```
lkiesow: My general suggestion for configuration parsing would be to rather fail early if a configuration is broken than to fall back to a default value. That way, as a system administrator you will notice that problem immediately and can fix it instead of accidentally running with an incorrect value for whatever time ;-)

In general, I would also suggest using common-lang3's [NumberUtils::toLong(String str, long defaultValue)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#toLong-java.lang.String-long-). Makes your life much easier and basically any Opencast module uses it anyway.

lkiesow: I mean [toInt(String str, int defaultValue)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#toInt-java.lang.String-int-)

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+      logger.warn("{} could not be loaded, default value is used: {}", CACHE_EXPIRATION, DEFAULT_CACHE_EXPIRATION_VALUE);
+    }
+    return DEFAULT_CACHE_EXPIRATION_VALUE;
+  }
+
+  private int parseCacheSizeProperty(Dictionary properties) {
```
lkiesow, line 189: same here…

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+      if (properties.get(CACHE_SIZE) != null) {
+        return  Integer.parseInt((String) properties.get(CACHE_SIZE));
+      }
+    } catch (NumberFormatException nfe) {
+      logger.warn("{} could not be loaded, default value is used: {}", CACHE_SIZE, DEFAULT_CACHE_SIZE_VALUE);
+    }
+    return DEFAULT_CACHE_SIZE_VALUE;
+  }
+
+  private void validateSystemUserKey(String userKey) throws ConfigurationException {
```
lkiesow: The following four methods are identical except for the key which you could just pass along to the method as arhument.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+    if (StringUtils.isBlank(userId)) {
+      throw new ConfigurationException(BRIGHTSPACE_USER_ID, "is not set");
+    }
+  }
+
+  private void validateUrl(String urlStr) throws ConfigurationException {
+    if (StringUtils.isBlank(urlStr)) {
+      throw new ConfigurationException(BRIGHTSPACE_URL, "is not set");
+    } else {
+      URI url;
```
lkiesow: This is never used

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -0,0 +1,246 @@
...
+      try {
+        url = new URI(urlStr);
+      } catch (URISyntaxException var21) {
```
lkiesow:
```suggestion
      } catch (URISyntaxException e) {
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+
+public class BrightspaceUserProviderInstance implements UserProvider, RoleProvider, CachingUserProviderMXBean {
+
+  private static final Logger logger = LoggerFactory.getLogger(BrightspaceUserProviderInstance.class);
+  private static final String PROVIDER_NAME = "brightspace";
+
+  private String pid;
+  private BrightspaceClient client;
+  private Organization organization;
+  private LoadingCache<String, Object> cache;
```
lkiesow, line 73: I'm wondering if it wouldn't make sense to make this

```java
  private LoadingCache<String, User> cache;
```

Instead of casting the Objects to users later on.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.security.api.UserProvider#countUsers()
+   */
+  @Override
+  public long countUsers() {
+    return 0L;
```
lkiesow, line 194: Wouldn't that suggest that this user provider never holds any users?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.security.api.UserProvider#findUsers(java.lang.String, int, int)
+   */
+  @Override
+  public Iterator<User> findUsers(String query, int offset, int limit) {
+    return Collections.emptyIterator();
```
lkiesow, line 214: Same here. Why never return any user?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
lkiesow: But these are configurable in `custom.properties` so we should never hard-code them:

```
org.opencastproject.security.admin.user=admin
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/BrightspaceClientImpl.java
```diff
@@ -0,0 +1,220 @@
...
+    String request = GET_USER_BY_USERNAME + userName;
+    String response = httpGetRequest(request);
+
+    try {
+      List<BrightspaceUser> brightspaceUserList = objectMapper
+              .readValue(response, new TypeReference<List<BrightspaceUser>>() {
+              });
+      return brightspaceUserList.stream().findFirst().orElse(null);
+    } catch (IOException e) {
+      logger.error(UNEXPECTED_JSON_RESPONSE);
```
lkiesow: No need to log this if you throw the same message as exception anyway.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/BrightspaceClientImpl.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  @Override
+  public List<BrightspaceUser> findAllUsers() throws BrightspaceClientException {
+    String response = httpGetRequest(GET_ALL_USERS);
+    try {
+      UsersResponse usersResponse = objectMapper.readValue(response, UsersResponse.class);
+      return usersResponse.getItems();
+    } catch (IOException e) {
+      logger.error(UNEXPECTED_JSON_RESPONSE);
```
lkiesow: same here

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/BrightspaceClientImpl.java
```diff
@@ -0,0 +1,220 @@
...
+
+  private URL createUrl(String request) throws BrightspaceClientException {
+    URI uri = userContext.createAuthenticatedUri(request, "GET");
+    URL url;
+    try {
+      url = uri.toURL();
+    } catch (MalformedURLException mue) {
+      throw new BrightspaceClientException("url was malformed", mue);
+    }
+    logger.debug("about to make GET request to : {}", uri.toString());
```
lkiesow:
```suggestion
    logger.debug("about to make GET request to : {}", uri);
```

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/BrightspaceClientImpl.java
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  private String readInputStream(InputStream inputStream) throws BrightspaceClientException {
+    StringBuilder content = new StringBuilder();
+    try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream))) {
+      String inputline;
+      while ((inputline = bufferedReader.readLine()) != null) {
+        content.append(inputline);
+      }
+      logger.debug("call to brightspace api: {}", content.toString());
```
lkiesow:
```suggestion
      logger.debug("call to brightspace api: {}", content);
```

lkiesow (changes requested):
I did a quick run-through without any actual testing. Most comments aren't very important. Hard-coding `admin` is probably the most important one.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/api/BrightspaceUser.java
```diff
@@ -0,0 +1,103 @@
...
+  public BrightspaceUser(@JsonProperty("OrgId") String orgId, @JsonProperty("UserId") String userId, @JsonProperty("FirstName") String firstName, @JsonProperty("MiddleName") String middleName, @JsonProperty("LastName") String lastName, @JsonProperty("UserName") String userName, @JsonProperty("ExternalEmail") String externalEmail, @JsonProperty("OrgDefinedId") String orgDefinedId, @JsonProperty("UniqueIdentifier") String uniqueIdentifier, @JsonProperty("Activation") Activation activation, @JsonProperty("DisplayName") String displayName) {
```
JamesUoM: This line is too long (by about 0.5km ;-) )

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
smarquard: These providers have all inherited this code from the original Sakai provider, so changes like this should really be made across all of them, or we move the change higher up the stack.

The logic for user providers not handling admin and anonymous is it's not necessary to resolve them further (because admin has all permissions and anonymous is, well, anonymous). So that check could go into the UserDirectoryService or whatever's calling these providers.

lkiesow: Then we should fix it in the Sakai user provider as well since that's clearly a bug.

> because admin has all permissions and anonymous is, well, anonymous

The admin user's name in configurable and it could be any other string so there is no guarantee that a user named `admin` has all permissions.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: Honestly no idea, the code was developed by an external consultant and it seems that this was copy-pasted from the sakai userprovider.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: I changed the error to "A Brightspace API error occurred, user {} could not be retrieved". Is this clean enough?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/api/BrightspaceUser.java
kristofkeppens: I can have a look, could you point me in the right direction to where to find the different library ?

modules/userdirectory-brightspace/pom.xml
kristofkeppens: I prefer 2 spaces as well, I'll change the pom.xml :)

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
smarquard: I looked into why that's there in the Sakai provider, and can't find an answer - it's been there for quite some time and I can't find the commit or issue that added it.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: I'm assuming that this was copied over from the sakai userprovider as well, not sure why this is this way.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: I think it is because we don't need this method, not sure. From where would you use this method, is this to search for users from the UI?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: Seems like a valid argument :), checked the code and was copied over from the other userdirectories as well. I can try to have a look at changing it.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
smarquard: The LMS user providers are intended to support looking up users, not "holding" users like the internal user provider does. There's no value in providing an arbitrary user count here.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
gregorydlogan: Agree with @lkiesow, good catch.  I'm wondering if we should merge this as is, and then fix all the providers with a separate PR after this one and #633 go in.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
gregorydlogan: Yep, perfect.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
gregorydlogan: @smarquard I don't see anything obvious regarding why this is needed.  Can you try yanking the loader at your end and see if it breaks anything.  @flyapen, if you have a system to test this against, can you yank it here too and see if it breaks things?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/api/BrightspaceUser.java
gregorydlogan: Sure thing.  Most of our classes with these annotations are prefixed with "Jaxb".  A simple one would be https://github.com/opencast/opencast/blob/develop/modules/common/src/main/java/org/opencastproject/job/api/JaxbJob.java, and its other Jaxb friends in that package.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
smarquard: I can try on our test system, but not keen to make a possibly-breaking-change in production right now especially as there's no particular problem that we're trying to solve. I think it may be related to caching or something but it would have been done like that in 1.x days.

smarquard: It turns out the Sakai provider was copied from the Ldap provider, and we find this classloader switching there:

https://github.com/opencast/opencast/blob/ceb51e5a42d21997f478b0d84ef674ab4fc310ee/modules/userdirectory-ldap/src/main/java/org/opencastproject/userdirectory/ldap/LdapUserProviderInstance.java#L320

gregorydlogan: Oh man, old code is old.  Now that I look at how this is being used by the LDAP (and Sakai) code, I'm guessing this switching is more or less required.  The factory which creates this provider has the relevant Karaf service bits and hooks into the security system, I think.  It might be worth testing in a test environment to see if this is still required, if you get a chance.  It's always nice to be able to remove black magic from our code :)

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
lkiesow: Let's get it in properly. I'll try to provide a reference implementation. Hopefully tomorrow.

lkiesow: As promised, I created a reference implementation by fixing the Moodle user provider in pull request #902

KatrinIhler:
This PR has test failures!

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderInstance.java
kristofkeppens: I have implemented it in the same way for this user provider

KatrinIhler:
Test failures and conflicts!

gregorydlogan:
@flyapen Any idea when there will be movement on this?

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/client/api/BrightspaceUser.java
kristofkeppens: I had a look to change this myself but not really confident enough that I won't break more than it'll fix. We'll have a look to see if we can get this fixed by a developer.

kristofkeppens:
@gregorydlogan I think I've been able to resolve most comments, there are still some open issues. If these prohibit the merge we'll have a look at asking outside help with the remaining issues.

gregorydlogan:
I see two remaining issues, assuming no new issues were introduced:

1) The serialization library is not the same one the rest of the project uses
2) There are some missing Javadocs

I really only see 1 as a serious issue (we've got lots of missing javadocs), and I'm not sure it's worth blocking this issue.  @lkiesow, @JamesUoM, @smarquard, thoughts?

gregorydlogan:
Per discussion in the technical meeting today, the libraries in question are already actually in use elsewhere in the codebase, so this is going in right... now.

PR #834, opened 2019-04-12
MH-13483 Introduce basic statistics visualization capabilities

pmiddend:
This pull request provides front-end and back-end changes to implement statistics visualization and retrieval.

# Abstract #

The idea is that there can be multiple statistics-collecting backends, each delivering data to Opencast by exposing so-called “statistics providers”. There is currently only one provider *type* available, which, given a time frame and bucket size, delivers *time-series data* for that time frame.

To illustrate these concepts: this PR includes a statistics backend which connects to an [InfluxDB](https://www.influxdata.com/) database and retrieves episode view counts, with the organization ID and series ID attached to each view. This InfluxDB backend exposes three statistics providers: one for time-series data about an episode, one for time-series data about a series, and the same for a whole organization.

The Admin-UI queries these providers (see below) and displays one chart for each provider.

# Querying statistics #

There are two new queries that you can make via the External API. The Admin-UI also implements two new analogous endpoints.

The first endpoint retrieves a list of the providers mentioned above, along with metadata like a textual description of the provider (“Episode Views”, for example). Each provider also has a unique *ID*.

The second endpoint receives a list of these provider IDs, as well as temporal data such as “from” and “to”, and delivers the corresponding (presumably numerical) statistical results, aka “views per month”.

This is just a summary. The endpoints are documented in the External API documentation.

# User Interface changes #

356d131f3b83d00ddeef4a9dc3704fbf

There are *three* new sections in the Admin-UI

  1. a “Statistics” tab in the event details modal
  2. a “Statistics” tab in the series details modal
  3. a new main navigation point for the statistics of an organization

There are new roles for the series and event details modals, and the tabs are only displayed if the user is authorized and providers have been found.

For the charts, [chart.js](https://www.chartjs.org/) is used, along with its AngularJS wrapper.

# API changes #

This pull request introduces External API version 1.3.0. The following requests have been added:

GET /api/statistics/providers
GET /api/statistics/providers/{providerId}
POST /api/statistics/data/query

# Testing #

You can test this in one of two ways: You can set up an InfluxDB database and specify its credentials as described in the documentation. To fill the InfluxDB with data, we provided a script in the admin docs.

To test this without setting up an InfluxDB instance, we also included a random provider that returns different random data on each request. However, when you build Opencast, both the InfluxDB providers and the random providers are set up, so you have to explicitly remove the InfluxDB providers from the `etc/statistics` directory to *just* use the random provider.

If the merge is approved, we will remove all providers by default.

# InfluxDB Statistic Provider Implementation

https://github.com/opencast/opencast-influxdb-adapter/pull/1 provides an implementation of the StatisticsProvider interface for InfluxDB.

# Attribution #


wsmirnow:
The ETH is interesting in this feature. So we will review this PR. If you also interesting in this feature, feel free to join and help us to review such a big PR.

smarquard:
Do we need to pull in these JS components with full source? This PR updates 520 files which is huge.

staubesv:
@wsmirnow Thanks for taking the review! Note that the PR actually isn't that big. As @smarquard pointed out, most of the changes come from the included libraries (in particular, chart.js) in the bower_components folder.

We also noticed that including everything that bower install chartjs will download results in a lot of files to be included. Maybe we can only check in the dist folder of chartjs. What do others think about this?

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
+Overview
+========
+
+In Opencast, the "Statistics" feature can be seen as a set of charts which can be displayed in the Admin UI. Currently,
+statistics for three so called "resource types" are available:
```
KatrinIhler: so-called

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
...
+username, password, and database name.
+
+Statistics Providers
+--------------------
+
+To support the detailed configuration of the charts to be shown in the Admin UI, Opencast has a concept called
+_Statistics Providers_. Each statistics provider can be configured separately and for each provider, there is one chart
+displayed in the Admin UI.
+
+The configuration files of the providers have to stored under `etc/statistics` and they have to follow a certain naming
```
KatrinIhler: to be stored at /are located at

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
...
+pattern. Configuration files of providers using InfluxDB have to be named starting with `influx.`. All provider
```
KatrinIhler: naming convention?

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
...
+configurations have to be in json format. So e.g. _influx.views.episode.sum.json_ would be a valid name.
+
+For each provider, the following properties have to be configured:
+
+- **`id`** has to be a unique identifier and can be chosen freely.
+- **`title`** is the title to be displayed with the chart. This can be a translation key.
+- **`description`** is the description to be displayed with the chart. This can be a translation key.
+- **`resourceType`** tells Opencast to which type if entity the chart refers to. Valid values are `EPISODE`, `SERIES`,
```
KatrinIhler: type of entity

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
...
+  "source": "influx:infinite.impressions_daily:SUM:value:episodeId",
+  "type": "timeseries"
+}
+```
+
+Verifying Your Setup<a name="verify"></a>
+====================
+
+If you want to test your setup, you can put the following test data into InfluxDB and check if Opencast displays all
+charts correctly. First, create a series and an event as part of that series using the Opencast Admin UI. Seconds. copy
```
KatrinIhler: Second,

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+`withparameters`       | [`boolean`](types.md#basic) | Whether support parameters should be included in the response
+
+__Sample request__
+```xml
+https://opencast.domain.com/api/statistics/providers?filter=resourceType:episode
+```
+
+__Response__
+
+`200 (OK)`: A (potentially empty) list of providers is returned as JSON array contained JSON objects describing the series:
```
KatrinIhler: as a JSON array containing

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+
+Field          | Type                                      | Description
+:--------------|:------------------------------------------|:-----------
+`identifier`   | [`string`](types.md#basic)                | The unique identifier of the provider
+`title`        | [`string`](types.md#basic)                | The title of the provider
+`description`  | [`string`](types.md#basic)                | The description of the provider
+`type`\*       | [`string`](types.md#basic)                | Type of the provider
```
KatrinIhler: The type

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+
+```
+[
+  {
+    "identifier": "influxdb-episode-views-provider",
+    "title": "Episode Views",
+    "description": "Episode Views, Lorem Ipsum",
+    "type": "timeSeries",
+    "resourceType": "episode",
+    "para": ["monthly", "yearly"]
```
KatrinIhler: parameters

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+  }
+]
+```
+
+### GET /api/statistics/providers/{providerId}
+
+Returns a statistics providers.
```
KatrinIhler: provider

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+`withparameters`       | [`boolean`](types.md#basic) | Whether support parameters should be included in the response
+
+__Sample request__
+```xml
+https://opencast.domain.com/api/statistics/providers/a-timeseries-provider?withparameters=true
+```
+
+__Response__
+
+`200 (OK)`: A (potentially empty) list of providers is returned as JSON array contained JSON objects describing the series:
```
KatrinIhler: a JSON array containing

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+
+Field          | Type                                      | Description
+:--------------|:------------------------------------------|:-----------
+`identifier`   | [`string`](types.md#basic)                | The unique identifier of the provider
+`title`        | [`string`](types.md#basic)                | The title of the provider
+`description`  | [`string`](types.md#basic)                | The description of the provider
+`type`\*       | [`string`](types.md#basic)                | Type of the provider
```
KatrinIhler: The type

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+      "name": "dataResolution",
+      "optional": false,
+      "type": "enumeration"
+    }
+  ]
+}
+```
+
+
+### POST /api/statistics/data/query
```
KatrinIhler, line 134: Why is this a post request?

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+        "parameters": {
+          "resourceId": "23413432-5a15-328e-aafe-562a2bae6800",
+        }
+      }
+    ]
+
+
+__Response__
+
+`200 (OK)`: A (potentially empty) list of query results is returned as JSON array contained JSON objects
```
KatrinIhler: a JSON array containing

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+`parameters` | yes      | [`property`](#extended)  | A JSON object containing the parameters
+
+The JSON object `provider` has the following fields:
+
+Field        | Required | Type                | Description
+:------------|:---------|:--------------------|:-----------
+`identifier` | yes      | [`string`](#basic)  | A JSON object with information about the statistics provider to be queried
+
+The JSON object `parameters` has the following fields:
+
```
KatrinIhler: Is there something missing here?

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+
+__Response__
+
+`200 (OK)`: A (potentially empty) list of query results is returned as JSON array contained JSON objects
+`400 (BAD REQUEST)`: The request was not valid
+
+Field name       |Type                                  | Description
+:----------------|:-------------------------------------|:-----------
+`provider`       | [`property`](#extended)              | A JSON object describing the statistics provider as described below
```
KatrinIhler, line 190: above?

docs/guides/developer/docs/api/statistics-api.md
```diff
@@ -0,0 +1,251 @@
...
+          "labels": ["2019-04-10T13:45:32Z", "2019-04-11T00:00:00Z", "2019-04-12T00:00:00Z"],
+          "values": [20, 100, 300],
+          "total": 420
+        }
+    ]
+
+#### Time Series Statistics Provider
+
+Time Series Statistics Providers (type = timeseries) support some well-defined parameters and deliver a well-defined
+dataformat.
```
KatrinIhler: data format?

docs/guides/developer/docs/api/types.md
```diff
@@ -432,3 +456,30 @@ Field                    | Type
 `retry_strategy`         | [`workflow_retry_strategy`](#workflow_retry_strategy)   | The retry strategy of this workflow operation instance in case of an error
 `max_attempts`           | [`integer`](#basic)                                     | The number of attempts made to execute this workflow operation instance
 `failed_attempts`        | [`integer`](#basic)                                     | The number of failed attempts to execute this workflow operation instance
+
+## Statistics
+
+### parameter
```
KatrinIhler: parameters?

docs/guides/developer/docs/statistics.md
```diff
@@ -0,0 +1,99 @@
...
+
+Interfaces & Classes
+--------------------
+
+The Opencast `StatisticsService` implements the two interfaces `StatisticsProviderRegistry` and
+`StatisticsService.`
+
+**StatisticsProviderRegistry**
+
+This interface is used by `StatisticsProvider` implementation to register and unregister themselves at the
```
KatrinIhler: the StatisticsProvider implementations

docs/guides/user/docs/advanced/role-based-visibility.md
```diff
@@ -92,6 +99,7 @@ There are quite a number of roles to provide selective access to the tabs offere
 |ROLE_UI_EVENTS_DETAILS_SCHEDULING_VIEW   |Display tab *Scheduling*                 |
 |ROLE_UI_EVENTS_DETAILS_ACL_VIEW          |Display tab *Access Policy*              |
 |ROLE_UI_EVENTS_DETAILS_COMMENTS_VIEW     |Display tab *Comments*                   |
+|ROLE_UI_EVENTS_DETAULS_STATISTICS_VIEW   |Display tab *Statistics*                 |
```
KatrinIhler: DETAILS

KatrinIhler (changes requested):
Some nagging about the documentation, but it looks pretty good overall! Might take a look at the other things later.

docs/guides/developer/docs/api/statistics-api.md
staubesv: @KatrinIhler Because the parameters we pass are potentially to large for a query parameter.

modules/statistics-service-remote/src/main/java/org/opencastproject/statistics/remote/StatisticsServiceRemoteImpl.java
```diff
@@ -0,0 +1,216 @@
...
+    for (Object object : providersJson) {
+      JSONObject jsonObject = (JSONObject) object;
+      final String idJson = (String) jsonObject.get("id");
+      final String titleJson = (String) jsonObject.get("title");
+      final String descriptionJson = (String) jsonObject.get("description");
+      final String resourceTypeJson = (String) jsonObject.get("resourceType");
+      if (jsonObject.containsKey("dataResolutions")) {
+        final JSONArray resolutionsJson = (JSONArray) jsonObject.get("dataResolutions");
+        providers.add(new TimeSeriesProvider() {
+          @Override
```
KatrinIhler, line 152: Same question here: Why not call a constructor with the attributes?

modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/DateUtil.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public final class DateUtil {
+
+  private DateUtil() {
+  }
+
+  public static Instant min(final Instant i1, final Instant i2) {
+    return i1.isBefore(i2) ? i1 : i2;
```
KatrinIhler: Is there no library that provides this functionality, at least min/max? I'm pretty sure we have more than one DateUtil in opencast already, we shouldn't add another unless it's really necessary...

modules/statistics-provider-random/src/main/java/org/opencastproject/statistics/provider/random/StatisticsProviderRandomService.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public void deactivate(ComponentContext cc) {
+    logger.info("Deactivating Statistics Provider Random Service");
+    fileNameToProvider.values().forEach(provider -> statisticsProviderRegistry.removeProvider(provider));
+  }
+
+  private StatisticsProvider createRandomStatisticsProvider(ConfiguredProvider providerCfg) {
+    return new RandomStatisticsProvider() {
+      @Override
```
KatrinIhler: Wouldn't it be easier to make the RandomStatisticProvider not abstract and call a constructor instead of overriding the methods here?

modules/external-api/src/main/java/org/opencastproject/external/util/statistics/DataResolutionUtils.java
```diff
@@ -0,0 +1,55 @@
...
+  public static JSONArray toJson(Set<DataResolution> dataResolutions) {
+    JSONArray result = new JSONArray();
+    for (DataResolution resolution : dataResolutions) {
+      result.add(toString(resolution));
+    }
+    return result;
+  }
+
+  public static String toString(DataResolution dataResolution) {
+    return dataResolution.toString().toLowerCase();
```
KatrinIhler, line 44: Instead of creating a lot of utility classes, why can't this kind of functionality be part of the classes, here DataResolution, themselves? They are a lot harder to miss that way, so chances are slimmer that people create duplicate utility functionality, as often seen in opencast...

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -0,0 +1,378 @@
...
+    }
+    return result;
+  }
+
+  private String dataResolutionToJson(DataResolution dataResolution) {
+    return dataResolution.toString().toLowerCase();
+  }
+
+  private static Optional<DataResolution> dataResolutionFromJson(final String dataResolution) {
+    try {
```
KatrinIhler, line 262: Isn't this similar to what DataResolutionUtil does?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -0,0 +1,378 @@
...
+  public void setSearchIndex(AdminUISearchIndex searchIndex) {
+    this.searchIndex = searchIndex;
+  }
+
+  public void setStatisticsService(StatisticsService statisticsService) {
+    this.statisticsService = statisticsService;
+  }
+
+  private static final class RawProviderQuery {
+    private String providerId;
```
KatrinIhler, line 119: I'm confused - why do we need RawProviderQuery when we have ProviderQuery?

KatrinIhler (commented):
Part two of my review, some questions I had while scrolling through the java files. I'll be on vacation starting tomorrow for almost two weeks, so responses may take a while.

modules/external-api/src/main/java/org/opencastproject/external/util/statistics/DataResolutionUtils.java
staubesv: ... because changing the classes in the statistics service would imply changing the External API which makes it hard to detect unintentional changes to the External API.
At least for the representation of data, the External API should really use its own code to create representations as not having dependencies to shared code seems more important than to avoid code duplication in the case of the External API. 

In this particular example, though, the External API is not actually changed when messing around with DataResolution as the exact values of data resolutions are not part of the External API specification.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -352,6 +356,10 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-search-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-search-service-remote/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-series-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-provider-influx/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-api/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-remote/${project.version}</bundle>
```
wsmirnow: Why are the impl and the remote module in the same assembly? These modules are conflicting.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -481,6 +489,10 @@
     <bundle start-level="84">mvn:org.opencastproject/opencast-search-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-search/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-series-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-provider-influx/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-api/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-remote/${project.version}</bundle>
```
wsmirnow: impl and remote are in conflict

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -647,6 +659,10 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-search-service-remote/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-search/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-series-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-provider-influx/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-api/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-statistics-service-remote/${project.version}</bundle>
```
wsmirnow: impl and remote…

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -0,0 +1,146 @@
...
+- **`resolutions`** is a list of resolutions supported by this provider. Opencast allows the user to select a
+  _resolution_ with which the data is displayed. Valid values are `DAILY`, `WEEKLY`, `MONTHLY` and `YEARLY`. E.g. when a
+  chart shows data of two years, a `DAILY` resolution will lead to 2x365=730 values to be plotted while a `MONTHLY`
+  resolution would leave us with 24 values being plotted in the chart.
+- **`source`** is a composite string which holds information about your data schema and how to calculate the
+  actual data to be displayed in the Admin UI. For InfluxDB, the `source` string consists of 5 parts, separated by
+  colons (:). A valid `source` string would be `influx:infinite.impressions_daily:SUM:value:episodeId` which has the
+  following meaning:
+
+    1. `influx` tells Opencast that InfluxDB is is the data source.
```
wsmirnow: …InfluxDB is the data… (double is)

docs/guides/admin/docs/configuration/admin-ui/statistics.md
krinnewitz: fixed

docs/guides/admin/docs/configuration/admin-ui/statistics.md
krinnewitz: fixed

docs/guides/admin/docs/configuration/admin-ui/statistics.md
krinnewitz: fixed

docs/guides/admin/docs/configuration/admin-ui/statistics.md
krinnewitz: fixed

docs/guides/admin/docs/configuration/admin-ui/statistics.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/statistics-api.md
krinnewitz: fixed

docs/guides/developer/docs/api/types.md
krinnewitz: fixed

docs/guides/developer/docs/statistics.md
krinnewitz: fixed

docs/guides/user/docs/advanced/role-based-visibility.md
krinnewitz: fixed

modules/statistics-provider-random/src/main/java/org/opencastproject/statistics/provider/random/StatisticsProviderRandomService.java
krinnewitz: Good point. Fixed.

modules/statistics-service-remote/src/main/java/org/opencastproject/statistics/remote/StatisticsServiceRemoteImpl.java
krinnewitz: Good point. Fixed.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
krinnewitz: I think here it is the same as above. We want to keep the admin-ui API and the external API separated.

modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/DateUtil.java
krinnewitz: fixed

docs/guides/developer/docs/statistics.md
```diff
@@ -0,0 +1,99 @@
...
+
+The StatisticsService consists of the following modules:
+
+* `statistics-service-api`
+An API module defining the core StatisticsService and StatisticsProvider functions.
+* `statistics-service-impl`
+The default implementation of the StatisticsService as an OSGi service.
+* `statistics-service remote`
+The remote implementation of the StatisticsService.
+* `statistics-service-influx`
```
wsmirnow: isn't it named `statistics-provider-influx`?

docs/guides/developer/docs/statistics.md
```diff
@@ -0,0 +1,99 @@
...
+An implementation of the StatisticsProvider for InfluxDB
+* `statistics-service-random`
```
wsmirnow: same here, `statistics-provider-random`

etc/security/security_sample_ldap.xml-example
```diff
@@ -96,6 +96,7 @@
     <sec:intercept-url pattern="/admin-ng/resources/providers.json" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
     <sec:intercept-url pattern="/admin-ng/resources/THEMES.json" method="GET" access="ROLE_ADMIN, ROLE_UI_THEMES_VIEW" />
     <sec:intercept-url pattern="/admin-ng/resources/*.json" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
+    <sec:intercept-url pattern="/admin-ng/statistics/views/**" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
```
wsmirnow: 1. see comment in the `mh_default_org.xml`
2. why are you using `/admin-ng/statistics/views/**` here and not `/admin-ng/statistics/*.json` like in `mh_default_org.xml`

etc/security/mh_default_org.xml
```diff
@@ -98,6 +98,7 @@
     <sec:intercept-url pattern="/admin-ng/resources/providers.json" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
     <sec:intercept-url pattern="/admin-ng/resources/THEMES.json" method="GET" access="ROLE_ADMIN, ROLE_UI_THEMES_VIEW" />
     <sec:intercept-url pattern="/admin-ng/resources/*.json" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
+    <sec:intercept-url pattern="/admin-ng/statistics/*.json" method="GET" access="ROLE_ADMIN, ROLE_ADMIN_UI" />
```
wsmirnow, line 101: Here I'd expect a `ROLE_ADMIN_UI_STATISTICS_VIEW` or something like that because all admin UI users have the `ROLE_ADMIN_UI` applied. Currently all Users, that have (partial) access to the admin UI can query and view the statistics.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -0,0 +1,378 @@
...
+  @Produces(MediaType.APPLICATION_JSON)
+  @RestQuery(name = "getproviderdata", description = "Returns the statistical data for a specific provider and a specific resource", returnDescription = "The statistical data as JSON", restParameters = {
+    @RestParameter(name = "data", isRequired = true, description = "A list of statistical data requests, containing a provider id, from, to, the resource id and a resolution - all as JSON", type = RestParameter.Type.TEXT) },
+    reponses = {
+      @RestResponse(description = "Returns the statistical data for the given resource type as JSON", responseCode = HttpServletResponse.SC_OK),
+      @RestResponse(description = "If the current user is not authorized to perform this action", responseCode = HttpServletResponse.SC_UNAUTHORIZED)
+    })
+  public Response getProviderData(@FormParam("data") String data) {
+    if (StringUtils.isBlank(data)) {
+      logger.warn("No data set");
```
wsmirnow: This log message is meaningless. I would drop the log level to debug at least or drop the log operation at all.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/StatisticsEndpoint.java
```diff
@@ -0,0 +1,300 @@
...
+        description = "A comma seperated list of filters to limit the results with. A filter is the filter's name followed by a colon \":\" and then the value to filter with so it is the form <Filter Name>:<Value to Filter With>.",
+        type = RestParameter.Type.STRING),
+      @RestParameter(
+        name = "withparameters", isRequired = false,
+        description = "Whether the parameters should be included in the response.",
+        type = RestParameter.Type.BOOLEAN)
+    },
+    reponses = {
+      @RestResponse(
+        description = "Returns the requested statistics providersas JSON",
```
wsmirnow: providers as JSON (missing space)

modules/external-api/src/main/java/org/opencastproject/external/util/statistics/ResourceTypeUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    String result;
+    switch (resourceType) {
+      case EPISODE:
+        result = RESOURCE_TYPE_EPISODE;
+        break;
+      case SERIES:
+        result = RESOURCE_TYPE_SERIES;
+        break;
+      case ORGANIZATION:
+        result = RESOURCE_TYPE_EPISODE;
```
wsmirnow: Is it right? I'd expect `RESOURCE_TYPE_ORGANIZATION` here.

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/StatisticsProviderInfluxService.java
```diff
@@ -0,0 +1,180 @@
...
+  @Override
+  public void update(File file) throws Exception {
+    uninstall(file);
+    install(file);
+  }
+
+  @Override
+  public void updated(Dictionary<String, ?> dictionary) {
+    if (dictionary == null) {
+      logger.info("No configuration available. Not connecting to influx.");
```
wsmirnow: Not connecting to influx - sounds wiered. What is about - Skip connecting to Influx DB - or - Please provide configuration for connection to Influx DB?

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxStatisticsProvider.java
```diff
@@ -0,0 +1,179 @@
...
+  public String getDescription() {
+    return description;
+  }
+
+  protected static String dataResolutionToInfluxGrouping(DataResolution dataResolution) {
+    switch (dataResolution) {
+      case DAILY:
+        return " GROUP BY time(1d)";
+      case WEEKLY:
+        return " GROUP BY time(1w, -3d)";  // -3d because otherwise, influx starts weeks on Thursdays (wtf?)
```
wsmirnow: Thanks for the explanation, what the -3d is for. This is really uncommon, but the influx DB developers will have a reason for that, for sure. So please remove your personal statement on the end of the comment!

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxTimeSeriesStatisticsProvider.java
```diff
@@ -0,0 +1,111 @@
...
+
+  @Override
+  public TimeSeries getValues(String resourceId, Instant from, Instant to, DataResolution resolution, ZoneId zoneId) {
+    final String influxGrouping = dataResolutionToInfluxGrouping(resolution);
+    final List<Tuple<Instant, Instant>> periods = getPeriods(from, to, resolution, zoneId);
+    final List<String> labels = new ArrayList<>();
+    final List<Double> values = new ArrayList<>();
+    for (final Tuple<Instant, Instant> period : periods) {
+      final Query query = BoundParameterQuery.QueryBuilder
+          .newQuery("SELECT " + aggregation + "(" + aggregationVariable + ") FROM " + measurement + " WHERE " + resourceIdName + "=$resourceId AND time>=$from AND time<=$to" + influxGrouping)
```
wsmirnow: break the line at 120 chars

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/endpoint/StatisticsRestService.java
```diff
@@ -0,0 +1,210 @@
...
+public class StatisticsRestService {
+
+  /** Logging utility */
+  private static final Logger logger = LoggerFactory.getLogger(StatisticsRestService.class);
+
+  /** Statistics Service */
+  private StatisticsService statisticsService;
+
+  /** Default server URL */
+  protected String serverUrl = "http://localhost:8080";
```
wsmirnow: Remove unused variable.

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/endpoint/StatisticsRestService.java
```diff
@@ -0,0 +1,210 @@
...
+
+  /** Service url */
+  protected String serviceUrl = null;
```
wsmirnow: This variable is also never used. Please remove.

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/endpoint/StatisticsRestService.java
```diff
@@ -0,0 +1,210 @@
...
+
+  private static final JSONParser jsonParser = new JSONParser();
```
wsmirnow: Again, unused variable. Please remove.

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/endpoint/StatisticsRestService.java
```diff
@@ -0,0 +1,210 @@
...
+    } catch (Exception e) {
+      logger.error("Could not retrieve providers: {}", e.getMessage());
+      throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+    }
+  }
+
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  @Path("timeseries/{providerId:.+}.json")
+  @RestQuery(name = "getAsJson", description = "Returns the time series data for the given providerId", returnDescription = "Returns the time series data JSON document",
```
wsmirnow: The name `getAsJson` is allready used by `getProviders()`. Please rename.

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/endpoint/StatisticsRestService.java
```diff
@@ -0,0 +1,210 @@
...
+      pathParameters = {@RestParameter(name = "providerId", isRequired = true, description = "The provider identifier", type = STRING)},
+      restParameters = {
+          @RestParameter(name = "resourceId", description = "The id of the resource to get the data for. E.g. episode id.", isRequired = true, type = STRING),
+          @RestParameter(name = "from", description = "Start of the time series as ISO 8601 UTC date string", isRequired = true, type = STRING),
+          @RestParameter(name = "to", description = "End of the time series as ISO 8601 UTC date string", isRequired = true, type = STRING),
+          @RestParameter(name = "resolution", description = "Data aggregation level. Must be one of 'hourly', 'daily', 'weekly', 'monthly', 'yearly'", isRequired = true, type = STRING),
```
wsmirnow: `hourly` resolution is not implemented by `DataResolution`. please update the REST documentation or provide an implementation.

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/impl/StatisticsServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Override
+  public Set<StatisticsProvider> getProviders() {
+    return new HashSet<>(providers.values());
+  }
+
+  @Override
+  public Set<StatisticsProvider> getProviders(ResourceType resourceType) {
+    return providers.values().stream().filter(p -> p.getResourceType().equals(resourceType)).collect(Collectors.toSet());
```
wsmirnow: line break at 120 chars

modules/statistics-service-impl/src/main/java/org/opencastproject/statistics/impl/StatisticsServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Override
+  public Optional<StatisticsProvider> getProvider(String providerId) {
+    return providers.values().stream().filter(p -> p.getId().equals(providerId)).findAny();
+  }
+
+  @Override
+  public TimeSeries getTimeSeriesData(StatisticsProvider provider, String resourceId, Instant from, Instant to, DataResolution resolution, ZoneId zoneId) {
```
wsmirnow: line break at 120 chars

modules/statistics-service-remote/src/main/java/org/opencastproject/statistics/remote/StatisticsServiceRemoteImpl.java
```diff
@@ -0,0 +1,216 @@
...
+@RestService(name = "statisticsservice", title = "Statistics Service Remote", abstractText = "This service provides statistics.", notes = {
+    "All paths above are relative to the REST endpoint base (something like http://your.server/files)",
+    "If the service is down or not working it will return a status 503, this means the the underlying service is "
+        + "not working and is either restarting or has failed",
+    "A status code 500 means a general failure has occurred which is not recoverable and was not anticipated. In "
+        + "other words, there is a bug! You should file an error report with your server logs from the time when the "
+        + "error occurred: <a href=\"https://opencast.jira.com\">Opencast Issue Tracker</a>"})
+public class StatisticsServiceRemoteImpl extends RemoteBase implements StatisticsService {
+
+  private static final Logger logger = LoggerFactory.getLogger(StatisticsServiceRemoteImpl.class);
```
wsmirnow: unused variable, please remove

modules/statistics-service-remote/src/main/resources/OSGI-INF/statistics-service-remote.xml
```diff
@@ -0,0 +1,19 @@
...
+  <implementation class="org.opencastproject.statistics.remote.StatisticsServiceRemoteImpl"/>
+  <property name="service.description" value="Statistics Remote Service Proxy"/>
+
+  <property name="opencast.service.type" value="org.opencastproject.statistics"/>
+  <property name="opencast.service.path" value="/statistics"/>
+  <property name="opencast.service.publish" value="false"/>
+
+  <service>
+    <provide interface="org.opencastproject.statistics.api.StatisticsService"/>
+    <provide interface="org.opencastproject.statistics.remote.StatisticsServiceRemoteImpl"/>
```
wsmirnow: We should not never bind on remote implementations. I guess you can remove this provide statement.

wsmirnow (changes requested):
I've finished the theoretical code review. Some points have to be fixed but all in one it looks good. I'll continue the review by running and playing with the service tomorrow.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
pmiddend: I've added some Javadoc to clarify. We're using GSON to parse the JSON directly into a Java POJO (`RawProviderQuery`) and then, to actually work with it, we convert it to a proper query object (`ProviderQuery`).

etc/security/mh_default_org.xml
wsmirnow: Ok, the admin ui facade access policy is not ready for this request. It is ok to mark the statistics facade endpoints with `ROLE_ADMIN_UI`.

modules/statistics-service-remote/src/main/resources/OSGI-INF/statistics-service-remote.xml
krinnewitz: fixed

docs/guides/developer/docs/statistics.md
krinnewitz: fixed

docs/guides/developer/docs/statistics.md
krinnewitz: fixed

assemblies/karaf-features/src/main/feature/feature.xml
krinnewitz: fixed

assemblies/karaf-features/src/main/feature/feature.xml
krinnewitz: fixed

assemblies/karaf-features/src/main/feature/feature.xml
krinnewitz: fixed

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxTimeSeriesStatisticsProvider.java
```diff
@@ -0,0 +1,112 @@
...
+    final List<Double> values = new ArrayList<>();
+    for (final Tuple<Instant, Instant> period : periods) {
+      final Query query = BoundParameterQuery.QueryBuilder
+              .newQuery("SELECT " + aggregation + "(" + aggregationVariable + ") FROM " + measurement + " WHERE "
+                                + resourceIdName + "=$resourceId AND time>=$from AND time<=$to" + influxGrouping)
+              .bind("resourceId", resourceId)
+              .bind("from", period.getA())
+              .bind("to", period.getB())
+              .create();
+      final QueryResult results = service.getInfluxDB().query(query);
```
wsmirnow: `influxDB.query(…)` may fail on some conditions. Especially if the Influx DB is not running or not reachable. It would make sense to throw an exception in that case and mark the method as "throws exceptions" in the API.

etc/security/mh_default_org.xml
wsmirnow: The `/admin-ng/statistics/*.json` POST endpoint configuration is missing.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -0,0 +1,383 @@
...
+  }
+
+  private void checkOrganizationAccess(final String orgId) throws UnauthorizedException {
+    final User currentUser = securityService.getUser();
+    final Organization currentOrg = securityService.getOrganization();
+    final String currentOrgAdminRole = currentOrg.getAdminRole();
+    final String currentOrgId = currentOrg.getId();
+
+    boolean authorized = currentUser.hasRole(GLOBAL_ADMIN_ROLE)
+      || (currentUser.hasRole(currentOrgAdminRole) && currentOrgId.equals(orgId));
```
wsmirnow: There is no way to view organization statistics if you are not an admin or an org admin? What is the `ROLE_UI_STATISTICS_ORGANIZATION_VIEW` for?

I think we can change the access check to:
1. if the use is admin, provide access
2. if the user is org admin, provide access
3. if the user have role `ROLE_UI_STATISTICS_ORGANIZATION_VIEW` and the `orgID` is the ID oth the users organization, provide access
4. otherwise throw an exception

Are you agree?

staubesv: No, a user that has not access to all events of an organization should not see that statistics.

Or why should that be the case?

etc/security/mh_default_org.xml
krinnewitz: good catch. Fixed.

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxTimeSeriesStatisticsProvider.java
krinnewitz: Actually, `queryResultToTimeSeries()` checks for errors and throws an exception if any. I would like to leave that a `RuntimeException` since checked exceptions don't go well with java8's functional features (which are used in `StatisticsEndpoint`).

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
wsmirnow: The user you describing may not have the `ROLE_UI_STATISTICS_ORGANIZATION_VIEW` if he/she shouldn't see the org statistics. But in case you want to show your stats, you can add the role for the user and you are not limited to org admins.

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxTimeSeriesStatisticsProvider.java
wsmirnow: Unfortunately true. I will loose my `chuck(e)` war in opencast only because exceptions are not intended in functional programming.
And no, `queryResultToTimeSeries()` does not filter the case, the connection to influx db is blocked by the firewall. You will see a really ugly and long stacktrace with a simple reason:

```
Caused by: org.influxdb.InfluxDBIOException: java.net.ConnectException: Failed to connect to /127.0.0.1:8086
```

Is there a way to handle this case properly? You can easy reproduce it by stopping influx db and call the /statistics/timeseries endpoint.

assemblies/karaf-features/src/main/feature/feature.xml
wsmirnow: As I understood, you want to provide statistics in the admin UI and external API. Both services are currently running on admin or partial admin node (like adminpresentation, adminworker). It make sense to put the statistics-service-impl on that node IMO. 

The remote is currently not really needed, unless you separate external API and admin UI. In that case we will make use of the statistics-service-remote.

Please update.

assemblies/karaf-features/src/main/feature/feature.xml
wsmirnow: As described before, please drop remote and put the impl back

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/provider/InfluxTimeSeriesStatisticsProvider.java
krinnewitz: I have added special handling for connect exceptions. In this case, now only the error message is logged.

assemblies/karaf-features/src/main/feature/feature.xml
krinnewitz: fixed

assemblies/karaf-features/src/main/feature/feature.xml
krinnewitz: fixed

modules/statistics-provider-influx/src/main/java/org/opencastproject/statistics/provider/influx/StatisticsProviderInfluxService.java
krinnewitz: OK, I have added "DB" to clarify what is meant.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
krinnewitz: Fixed. Users with `ROLE_UI_STATISTICS_ORGANIZATION_VIEW` can now view the org stats if they belong to the organization.

krinnewitz:
Thanks for the review @wsmirnow and @KatrinIhler. I have fixed your findings so far. Is your review still ongoing or are we good to go now?

wsmirnow:
After some failures to get influxdb docker running with the testdata from the admin guide, I have it now. The influxdb statistics provider configuration contains a retantion policy name `infinite`. This is missing on a clean influxdb instance and have to be created. Or in my case, I've removed it from all `etc/statistics/influx*` configurations <- only for testing purposes. 

@pmiddend as you mentioned, you will remove these predefined `etc/statistics/*` configuration, then it is fine. Otherwise we have to solve this issue.

wsmirnow:
Ok, I'm done with influx DB provider tests. All in one it works as expected. But if not, the error message isn't shown in some cases. The InfluxDB documentation provides some examples of the query responses, see https://docs.influxdata.com/influxdb/v1.7/tools/api/#status-codes-and-responses-1. Here thay are:

1. HTTP 200: success. This is handled properly
2. HTTP 400: syntactic error. This is also handled properly at the beginning of `InfluxTimeSeriesStatisticsProvider::queryResultToTimeSeries(QueryResult)`
3. HTTP 403: invalid authentication. This may be handled properly, see the case HTTP 400
4. HTTP 200: successful request that returns an error (see InfluxDB documentation example 2): This is not handled at all but is really easy to handle. In the `InfluxTimeSeriesStatisticsProvider::queryResultToTimeSeries(QueryResult)` you iterate over all results. There (line 113+) you can also query the result status like  `if (result.hasError())…`. I would expect at least an warning message in the log output.

krinnewitz:
> @pmiddend as you mentioned, you will remove these predefined `etc/statistics/*` configuration, then it is fine. Otherwise we have to solve this issue.

I have now removed the example configurations, since I sense we are close to the merge :)

krinnewitz:
> This is not handled at all but is really easy to handle. In the `InfluxTimeSeriesStatisticsProvider::queryResultToTimeSeries(QueryResult)` you iterate over all results. There (line 113+) you can also query the result status like  `if (result.hasError())…`. I would expect at least an warning message in the log output.

Good catch. I am now checking every single result for errors, too.

wsmirnow (approved):

wsmirnow:
Good job. Thank you.

@KatrinIhler are you done with the review?

KatrinIhler:
@krinnewitz Looking good now, I guess the only thing that still bothers me are all those utility classes. I mean, since the external API is often dependent on other parts of opencast, it's always in danger of being accidentally changed, isn't it? I think I'd still rather see that functionality in the actual classes.

staubesv:
@KatrinIhler It was actually me the refactored that code and created the utility classes ;-) I did that by intention (not as shortcut to save time) as I really prefer having the code the generates the representation for the External API to be separated and not re-used (API stability seems more important to me than re-use of simple code).

>  I mean, since the external API is often dependent on other parts of opencast, it's always in danger of being accidentally changed, isn't it? 

Yes, unfortunately that is true.
At least the separation approach in this PR makes accidental changes to the External API less likely - that is exactly the point of it.

Can we agree to disagree and keep the current state?

KatrinIhler (approved):
Alright, giving in! ;)

staubesv:
Thanks ;-)

wsmirnow:
@pmiddend  would you squash the commits to one (or some)? I will then merge the branch.

krinnewitz:
@wsmirnow done

lkiesow:
While this got merged, it seems like there are still errors. Just running develop (ingested two recordings, looked at the admin interface) with default settings gives me a lot of error logs:

```
2019-05-11 14:04:35,633 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,636 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,637 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,639 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,640 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,642 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,643 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,644 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,645 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,646 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,648 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086                                                                        
2019-05-11 14:04:35,649 | ERROR | (InfluxTimeSeriesStatisticsProvider:89) - Influx connect exception: java.net.ConnectException: Failed to connect to /127.0.0.1:8086
```

krinnewitz:
That is because the configs under etc/statistics got merged by accident with PR #876. Removing those json files should fix the problem.

lkiesow:
So, that directory should be dropped completely?

krinnewitz:
I'd rather keep the directory, so users configuring this stuff already have it. I think it already contains a `.gitdummy`.

lkiesow:
Can you just create a pull request? :-)

krinnewitz:
https://github.com/opencast/opencast/pull/890

lkiesow:
@pmiddend, I noticed that the unit tests for `statistics-service-api` are extremely slow, taking ~45sec to run on my fairly fast workstation. While most testing for me is run on build-servers, assuming that on average I build Opencast at least once per day with tests enabled, that means that I alone will spend approximately 4:30h per year just waiting on this particular test. And I'm far from the only Opencast developer.

I know that there are a few other modules with slow tests but I don't think it's nice to add more slow ones. Could you take another look at it and fix those tests?

krinnewitz:
Sorry, didn't notice that these tests are too slow. Here is a PR: https://github.com/opencast/opencast/pull/894

PR #835, opened 2019-04-12
MH-13494 use metadata in publish configureable woh

rrolf:
This feature uses data from metadata catalogs to create the URL to which publish-configureable is "publishing". 
So if you miss-use a DC catalog field or create and extended metadata catalog where you write an URL you can use this in the WOH.

As the Freemarker template engine that Opencast is already using here, might stumple about missing data, you can now use the "configure-by-dcterm" WOH to check if the field is available with the expected data.
For this "configure-by-dcterm" is now able to access extended metadata catalogs with a new parameter and the matching is enhanced in a way that also regular expressions can be used. 

Example configuration for testing (included in partial-publish.xml workflow), you would need an extended series metadata catalog "myterms": 

```xml
<operation
  id="configure-by-dcterm"
  fail-on-error="true"
  description="Configure publication channel by dcterm">
  <configurations>
    <configuration key="externalcatalog">myterms</configuration>
    <configuration key="dccatalog">series</configuration>
    <configuration key="dcterm">portalUrl</configuration>
    <configuration key="match-value">^(http|https):\/\/(.*)</configuration>
    <configuration key="portalurlset">true</configuration>
  </configurations>
</operation>

    <operation
      id="defaults"
      description="Applying default configuration values">
      <configurations>
        <configuration key="portalurlset">false</configuration>
      </configurations>
    </operation>

    <operation
            id="publish-configure"
            if="${portalurlset}"
      description="Publish to external api publication channel">
      <configurations>
        <configuration key="channel-id">portal</configuration>
        <configuration key="source-tags">engage-download,engage-streaming</configuration>
        <configuration key="url-pattern">${(myterms.series.portalUrl?lower_case)?ensure_ends_with('/')}${event_id}</configuration>
      </configurations>
    </operation>
```


KatrinIhler:
This PR contains unrelated commits, the branch might need a rebase?

KatrinIhler:
@rrolf This PR has conflicts!

staubesv:
@rrolf Please note that this PR currently has merge conflicts.

modules/engage-paella-player/package.json
```diff
@@ -10,6 +10,6 @@
     "eslint-plugin-header": "^3.0.0",
     "gulp": "^4.0.0",
     "merge-stream": "^1.0.1",
-    "paellaplayer": "github:polimediaupv/paella#6.1.0"
+    "paellaplayer": "github:polimediaupv/paella#6.1.2"
```
lkiesow: You don't want to modify the Paella player. This also applies to `modules/engage-paella-player/package-lock.json`

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
```diff
@@ -96,29 +100,48 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
 
     String configuredCatalog = StringUtils.trimToEmpty(currentOperation.getConfiguration(DCCATALOG_PROPERTY));
     String configuredDCTerm = StringUtils.trimToEmpty(currentOperation.getConfiguration(DCTERM_PROPERTY));
+    String configuredExternalCatalog = StringUtils.trimToEmpty(currentOperation.getConfiguration(EXTERNAL_CATALOG_PROPERTY));
```
lkiesow:
```suggestion
    String configuredExternalCatalog = currentOperation.getConfiguration(EXTERNAL_CATALOG_PROPERTY);
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
```diff
@@ -96,29 +100,48 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
     String configuredDefaultValue = StringUtils.trimToNull(currentOperation.getConfiguration(DEFAULT_VALUE_PROPERTY));
     String configuredMatchValue = StringUtils.trimToEmpty(currentOperation.getConfiguration(MATCH_VALUE_PROPERTY));
 
+    // Set Default
+    EName dcterm = null;
+    if (("").equals(configuredExternalCatalog)) {
```
lkiesow:
```suggestion
    if (StringUtils.isBlank(configuredExternalCatalog)) {
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
```diff
@@ -96,29 +100,48 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+      List<DublinCoreValue> values = new LinkedList<>();
       for (Catalog catalog : catalogs) {
         DublinCoreCatalog dc = DublinCoreUtil.loadDublinCore(workspace, catalog);
-        // Match Value
-        List<DublinCoreValue> values = dc.get(dcterm);
-        if (values.isEmpty()) {
-          // Use default
-          if (configuredDefaultValue != null) {
-            foundValue = configuredDefaultValue.equals(configuredMatchValue);
+          for (Map.Entry<EName, List<DublinCoreValue>> entry : dc.getValues().entrySet()) {
```
lkiesow: indentation

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
```diff
@@ -96,29 +100,48 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+            if (dcterm != null && dcterm.equals(entry)) {
```
lkiesow: It's a bit weird that namespaces are used in case it's a dublincore catalog but that they are completely ignored otherwise. It would be less confusing to always (or never but I prefer always) specify them.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
```diff
@@ -96,29 +100,48 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
-        } else {
-          foundValue = values.contains(DublinCoreValue.mk(configuredMatchValue));
+      }
+      // Match Value
+      if (values.isEmpty()) {
+        // Use default
+        if (configuredDefaultValue != null) {
+          // check for direct character match or regex match
+          foundValue = configuredDefaultValue.equals(configuredMatchValue)
+                  || configuredDefaultValue.matches(configuredMatchValue);
```
lkiesow, line 138: `.matches(…)` should always be true if `.equals(…)` if true, won't it? That means that `.equals(…)` is just superfluous.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/ConfigurablePublishWorkflowOperationHandler.java
```diff
@@ -124,6 +136,36 @@ protected DownloadDistributionService getDistributionService() {
...
+          for (Map.Entry<EName, List<DublinCoreValue>> entry : dc.getValues().entrySet()) {
+            String value = entry.getValue().get(0).getValue();
+            metadata.put(entry.getKey().getLocalName(), value);
+          }
+          if (!values.containsKey(mainflavor)) {
+            values.put(mainflavor, new HashMap<>());
+          }
+          ((Map<String, Object>)values.get(mainflavor)).put(subflavor, metadata);
+        } catch (Exception e) {
+          logger.error("Catalog from URL %s with flavor %s could not be loaded", catalog.getURI(),
```
lkiesow:
```suggestion
          logger.error("Catalog from URL {} with flavor {}/{} could not be loaded", catalog.getURI(),
                  mainflavor, subflavor);
```

But wouldn't it make more sense to not catch the exception and let the operation fail if such an error occurs?

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/ConfigurablePublishWorkflowOperationHandler.java
```diff
@@ -124,6 +136,36 @@ protected DownloadDistributionService getDistributionService() {
...
+
+    for (Catalog catalog : mp.getCatalogs()) {
+      if (catalog.getFlavor() != null) {
+        String mainflavor = catalog.getFlavor().getType();
+        String subflavor = catalog.getFlavor().getSubtype();
+        Map<String, Object> metadata = new HashMap<>();
+        try {
+          DublinCoreCatalog dc = DublinCoreUtil.loadDublinCore(workspace, catalog);
+          for (Map.Entry<EName, List<DublinCoreValue>> entry : dc.getValues().entrySet()) {
+            String value = entry.getValue().get(0).getValue();
```
lkiesow, line 154: This will discard all but the first value in case you have e.g. multiple presenters. That's generally okay but is not what's documented. either adjust this or the documentation.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/ConfigurablePublishWorkflowOperationHandler.java
```diff
@@ -124,6 +136,36 @@ protected DownloadDistributionService getDistributionService() {
     return distributionService;
   }
 
+  private void addMetadataCatalogsToTemplateData(Map<String, Object> values, MediaPackage mp) {
+    if (! mp.hasCatalogs()) {
+      logger.info("No metdata catalogs found.");
+      return;
+    }
+    if (values == null) values = new HashMap<>();
```
lkiesow, line 144: Since `values` is not returned by this method, in case you create a new Map here, all data you add to it will be lost at the end of the method.

docs/guides/admin/docs/workflowoperationhandlers/configure-by-dcterm-woh.md
```diff
@@ -15,7 +15,8 @@ Tags and flavors can be used in combination.
 |------------------|-------|-----------|-------------|
 |dccatalog         |"episode" or "series"|the type of catalog in which to search for `dcterm`|EMPTY|
 |dcterm            |"creator"            |the name of the Dublin Core term which to check|EMPTY|
-|match-value       |"Joe Bloggs"         |the Dublin Core term value to check for|EMPTY|
+|externalcatalog   |i.e. "myterms"       |the type of external metadata catalog in which to search for `dcterm`|dublincore|
```
lkiesow, line 18: Maybe, instead of treating dublincore as a special case, it would make sense to have a more generic configuration of source catalog and source element like the [transfer-metadata operation](https://github.com/opencast/opencast/blob/d1aedd50967a2fc8564e0f6bc3462f5b06a8a370/docs/guides/admin/docs/workflowoperationhandlers/transfer-metadata-woh.md#operation-example)

lkiesow (changes requested):
I did a quick review and found a few problems additional to the merge conflicts.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
rrolf: First I check if the strings are identical (current default case) second I check if "configuredMatchValue" might be a regex that could match. If I only would have an match, if the regex always includes the identical value also.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConfigureByDublinCoreTermWOH.java
rrolf: The namespaces can be arbitrary selected by the user that introduces a new metadata catalogue. Where should I get these arbitrary namespaces from? Do we demand the users to provide them in another config file somewhere?

Removing the previously existing check for DC would be an option although.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/ConfigurablePublishWorkflowOperationHandler.java
rrolf: I don't think that failing is an option. We are looking for metadata catalogs to look for the key-value pairs there. If a catalog is not available for whatever reason we might have corrupt metadata, but in most cases that must not have any consequences why the publishing should not still work.

wsmirnow:
I've rebased this PR to develop, solved the merge conflicts and squashed all commits to one. As I havn't access to @rrolf GitHub repo, I will push it to my own and refile the PR.

wsmirnow:
Rebased to current develop and solved merge conflicts in PR #1279

PR #837, opened 2019-04-16
NOJIRA: Fix adaptive streaming configuration guide

JulianKniephoff:
`etc/org.opencastproject.streaming.wowza.StreamingDistributionService.cfg ` does not exist, AFAIK.

`etc/org.opencastproject.distribution.streaming.wowza.WowzaAdaptiveStreamingDistributionService.cfg` does, though.

P.S.: Is the NOJIRA "label" accepted for cases like these? If I were a commiter, I could just merge a change like this, too, right? :wink:

### DIFF
--- a/docs/guides/admin/docs/modules/adaptivestreaming-wowza.md
+++ b/docs/guides/admin/docs/modules/adaptivestreaming-wowza.md
@@ -61,8 +61,8 @@ below.
 ## Configuration
 
 
-1. Edit the file etc/org.opencastproject.streaming.wowza.StreamingDistributionService.cfg with your preferred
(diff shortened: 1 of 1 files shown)


gregorydlogan:
NOJIRA is fine here, especially for minor doc fixes like this.  If you were a committer you could indeed just commit and push :)

gregorydlogan:
How far back does this error exist?  This is targetting `develop`, should it go into say `r/6.x`?

JulianKniephoff:
Yip, it is in 6.x as well. Rebased and retargeted. :slightly_smiling_face:

PR #838, opened 2019-04-16
MH-13346 Thumbnail preview mode 'optimized'

staubesv:
This PR is based on https://github.com/opencast/opencast/pull/743.

It fixes the performance issue that results from downloading the source video files into the workspace when thumbnails are extracted on behalf of the video editor.

While #743 avoids the download by not extracting a thumbnail preview at all (partial-mode), the optimized mode avoids the download by calling the requires ffmpeg commands directly from the Admin UI facade accessing the source file directly.

Pros:
- Very fast
- Easy to do

Contras:
- Not compliant to the way Opencast is supposed to work
- Exposes video files in archive to ffmpeg command
- Only works on shared volume setups
- Requires ffmpeg to be installed on Admin Node


gregorydlogan:
A description here would be handy ;)

gregorydlogan:
A few top level comments:

* In a shared volume environment shouldn't hardlinks be possible?  Therefore linking the file into the workspace should be cheap, which would resolve one of the negative points here.  I suppose that doesn't help in environments where the shared space *doesn't* support hardlinks, but I'd be curious how many adopters are running such a configuration.
* Installing ffmpeg on the admin node is not a negative IMO, since one of our default node profiles requires it :)
* This could be accomplished cleanly by creating a lobotomized composer on the admin node which only handles the relevant operations.  I could see this being useful in the future as other admin-side composer operations pop up.

docs/guides/admin/docs/configuration/admin-ui/thumbnails.md
```diff
@@ -40,6 +40,19 @@ The thumbnail preview is automatically downscaled if necessary using the followi
...
+image can be configured:
+
+    # Default: partial
+    #thumbnail.preview.mode=partial
+
+If set to `always`, the thumbnail preview image is updated whenever required. If set to `partial`, the thumbnail
+preview image is only updated for uploaded thumbnails to avoid long waiting times that could cause timeouts in
+case of very large files.
+If set to `optimized`, the thumbnail preview image is updated whenever required, but instead of downloading
+files into the workspace (slow for large files), `ffmpeg` is called directly on the Admin node and accesses the
```
gregorydlogan, line 53: Needs a big warning about adding ffmpeg to the admin node, since it's normally not installed.

etc/org.opencastproject.adminui.cfg
```diff
@@ -53,6 +53,27 @@
...
+# The ad hoc updates can cause performance problems since potentially large video files need to be downloaded into the
+# workspace to be able to extract the image.
+# The thumbnail preview mode affects how Opencast does ad hoc thumbnail preview image updates:
+#
+# always:  The thumbnail preview image is updated whenever required (inefficient for large video files)
+# optimized: The thumbnail preview image is updated whenever required (efficient, see notes below).
+# partial: The thumbnail preview image is only updated for uploaded thumbnails (efficient)
+#
+# Notes on "optimzied":
+# The optimized mode requires ffmpeg to be installed on the Admin node and to be included in the search path.
```
gregorydlogan, line 70: If we're requiring the `episode.rootdir` key we should be using the composer ffmpeg key, no?

modules/admin-ui/src/main/java/org/opencastproject/adminui/impl/ThumbnailImpl.java
```diff
@@ -161,6 +183,11 @@ public ThumbnailImpl(final AdminUIConfiguration config, final Workspace workspac
     final OaiPmhPublicationService oaiPmhPublicationService,
     final ConfigurablePublicationService configurablePublicationService, final AssetManager assetManager,
     final ComposerService composerService) {
+    this.archiveBasePath = config.getArchiveBasePath();
+    this.previewMode = config.getThumbnailPreviewMode();
+    if ((this.previewMode == AdminUIConfiguration.ThumbnailPreviewMode.OPTIMIZED) && (this.archiveBasePath == null)) {
+      logger.error("Cannot enable thumbnail preview mode OPTIMIZED (archive path unknown). Falling back to ALWAYS.");
```
gregorydlogan, line 189: This log statement does not match the actual logic, correct?

KatrinIhler:
@staubesv The ball is in your court. ;)

gregorydlogan:
Per discussion on the dev list, let's decline this for now.  @staubesv (or other SWITCH contact) please feel free to refile as a draft or when ready.

PR #839, opened 2019-04-17
MH-13495 Ignore old requests instead of cancelling

KatrinIhler:
Instead of cancelling old requests, which leads to a 'server unavailable' notification, we ignore the response of a request if a more recent request finished earlier.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/services/tableService.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/services/tableService.js
@@ -271,10 +271,6 @@ angular.module('adminNg.services')
          */
       this.fetch = function (reset) {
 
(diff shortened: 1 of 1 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/services/tableService.js
```diff
@@ -321,8 +317,14 @@ angular.module('adminNg.services')
 
         (function(resource){
 
-          me.lastRequest = me.apiService.query(query);
-          me.lastRequest.$promise.then(function (data) {
+          var startTime = new Date();
+          me.apiService.query(query).$promise.then(function (data) {
```
gregorydlogan: Indentation doesn't match here.

PR #840, opened 2019-04-18
MH-13501 Match against user pattern for loadUser() lookups

smarquard:
When entering presenter information, Opencast tries to resolve the
presenter name as a user. This avoids trying to look up full names
(including spaces) as user eids from Sakai.


smarquard:
ping @gregorydlogan ...

PR #841, opened 2019-04-19
MH-12179 Expand log messages to add error detail

smarquard:
e.getMessage() can be null, and a null log entry is unhelpful.

### DIFF
--- a/modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
+++ b/modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
@@ -243,7 +243,8 @@ public class AssetManagerUpdatedEventHandler {
         }
       }
     } catch (IOException | NotFoundException e) {
-      logger.warn("Unable to handle update event for series {}: {}", seriesItem, e.getMessage());
+      logger.warn("Unable to handle update event for series {} for user {}: {}",
+                  seriesItem, prevUser.getUsername(), e.getMessage());
     } finally {
       securityService.setOrganization(prevOrg);
       securityService.setUser(prevUser);
--- a/modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
+++ b/modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
@@ -242,10 +242,8 @@ public class AssetManagerUpdatedEventHandler {
           logger.error("Error updating mediapackage {}", mp.getIdentifier().compact(), e);
         }
       }
-    } catch (NotFoundException e) {
-      logger.warn(e.getMessage());
-    } catch (IOException e) {
-      logger.warn(e.getMessage());
+    } catch (IOException | NotFoundException e) {
+      logger.warn("Unable to handle update event for series {}: {}", seriesItem, e.getMessage());
     } finally {
       securityService.setOrganization(prevOrg);
       securityService.setUser(prevUser);
--- a/modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
+++ b/modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
@@ -278,7 +278,8 @@ public class SeriesUpdatedEventHandler {
       logger.warn("Unable to find mediapackages for series {} in search: {}", seriesItem, e.getMessage());
     } catch (UnauthorizedException | MediaPackageException | ServiceRegistryException
              | NotFoundException | IOException | DistributionException e) {
-      logger.warn("Unable to update mediapackages for series {}: {}", seriesItem, e.getMessage());
+      logger.warn("Unable to update mediapackages for series {} for user {}: {} {}",
+                  seriesId, prevUser.getUsername(), e.getClass().getSimpleName(), e.getMessage());
     } finally {
       securityService.setOrganization(prevOrg);
       securityService.setUser(prevUser);
--- a/modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
+++ b/modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
@@ -275,19 +275,10 @@ public class SeriesUpdatedEventHandler {
         barrier.waitForJobs();
       }
     } catch (SearchException e) {
-      logger.warn("Unable to find mediapackages in search: ", e.getMessage());
-    } catch (UnauthorizedException e) {
-      logger.warn(e.getMessage());
-    } catch (MediaPackageException e) {
-      logger.warn(e.getMessage());
-    } catch (ServiceRegistryException e) {
-      logger.warn(e.getMessage());
-    } catch (NotFoundException e) {
-      logger.warn(e.getMessage());
-    } catch (IOException e) {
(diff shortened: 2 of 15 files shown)


modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
```diff
@@ -242,10 +242,8 @@ public void handleEvent(final SeriesItem seriesItem) {
           logger.error("Error updating mediapackage {}", mp.getIdentifier().compact(), e);
         }
       }
-    } catch (NotFoundException e) {
-      logger.warn(e.getMessage());
-    } catch (IOException e) {
-      logger.warn(e.getMessage());
+    } catch (IOException | NotFoundException e) {
+      logger.warn("Unable to handle update event for series {}: {}", seriesItem, e.getMessage());
```
karendolan: Is it Ok, privacy wise, to also add the user "preUser"'s username? Since you have it handy already?

modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
```diff
@@ -275,19 +275,10 @@ public void handleEvent(final SeriesItem seriesItem) {
...
-    } catch (NotFoundException e) {
-      logger.warn(e.getMessage());
-    } catch (IOException e) {
-      logger.warn(e.getMessage());
-    } catch (DistributionException e) {
-      logger.warn(e.getMessage());
+      logger.warn("Unable to find mediapackages for series {} in search: {}", seriesItem, e.getMessage());
+    } catch (UnauthorizedException | MediaPackageException | ServiceRegistryException
+             | NotFoundException | IOException | DistributionException e) {
+      logger.warn("Unable to update mediapackages for series {}: {}", seriesItem, e.getMessage());
```
karendolan: This is great to tighten up the catch. But, the exception message might be null, as mentioned, and then we won't know the class name of the original exception (because we can't go to the line of the code where the log was made and see the catch class type).

Since this is a WARN anyway, and should not be too frequent in the logs. How do you feel about including 3 more bits of info here: the username, mpId (if not null at this point), the exception class name being caught?

modules/conductor/src/main/java/org/opencastproject/event/handler/AssetManagerUpdatedEventHandler.java
smarquard: Added

modules/conductor/src/main/java/org/opencastproject/event/handler/SeriesUpdatedEventHandler.java
smarquard: I added username and exception class name. Hesitant to add the mp because the exceptions could be caused elsewhere in the try/catch where the mpid is not relevant. So to log an mpid accurately maybe needs additional try/catch blocks or some refactoring, a task I will be happy to leave to other maintainers :-)

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -270,7 +270,7 @@ public Response discardMediaPackage(@FormParam("mediaPackage") String mpx) {
       ingestService.discardMediaPackage(mp);
       return Response.ok().build();
     } catch (Exception e) {
-      logger.warn(e.getMessage(), e);
+      logger.warn("Unable to discard mediapackage", e);
       return Response.serverError().status(Status.INTERNAL_SERVER_ERROR).build();
```
karendolan, line 274: it would be useful to include the mpId that could not be discarded.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -545,7 +545,7 @@ protected Response addMediaPackageElement(HttpServletRequest request, MediaPacka
       }
       return Response.ok(MediaPackageParser.getAsXml(mp)).build();
     } catch (Exception e) {
-      logger.warn(e.getMessage(), e);
+      logger.warn("Unable to add mediapackage element", e);
       return Response.serverError().status(Status.INTERNAL_SERVER_ERROR).build();
     } finally {
       IOUtils.closeQuietly(in);
```
karendolan, line 551: for these logs, it would also be useful to include the affected mpId.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -1281,7 +1281,7 @@ public Response addDCCatalog(@FormParam("mediaPackage") String mp, @FormParam("d
       logger.error("Could not write catalog to disk: {}", e.getMessage());
       return Response.serverError().build();
     } catch (Exception e) {
-      logger.error(e.getMessage());
+      logger.error("Unable to add catalog: {}", e.getMessage());
```
karendolan, line 1284: Not requiring a change for this pull, I'm just noting the oddity that the failed mediapackage gets a WARN log. But, a failed add catalog is a full blown ERROR [system about to go down in a fiery mess] log.

IMHO, ERROR logs warrant a full stack trace, not just a potentially missing message.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowOperationWorker.java
```diff
@@ -169,7 +169,7 @@ public WorkflowInstance execute() {
       workflow = service.handleOperationResult(workflow, result);
       return workflow;
     } catch (JobCanceledException e) {
-      logger.info(e.getMessage());
+      logger.info("Job cancelled: {}", e.getMessage());
       return workflow;
```
karendolan, line 173: It would be useful to add the workflow Id

karendolan:
@smarquard I think this pull makes the logs more immediately useful, so one is not forced to look up logged line numbers in the code to see what's going on. The adding of a mpid/wfId/jobId when ever possible is also hugely helpful, but could be a future task for some other maintainer...

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
smarquard: Not sure we have an mpId accessible here

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
smarquard: An IOException from filesystem full, incorrect filesystem permissions or something would be  ERROR-worthy.

Anyway, changes to log level outside the scope of this PR.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowOperationWorker.java
smarquard: Added

smarquard:
Thanks for the detailed review and suggestions, @karendolan

smarquard:
Anything else to do here, or good to merge?

karendolan:
Looks good, merging.

PR #842, opened 2019-04-23
MH-13502: Harmonizing the column names

gregorydlogan:
In a few database column names we're using `media_package` rather than `mediapackage`.  This PR harmonizes this to match the rest of the code base.


lkiesow:
Though I wouldn't say this change is necessary since we are using *media package* all over the place, it's certainly less common in the database and I wouldn't expect this to cause huge migration problems. The change itself looks fine as well though I haven't tested it so far due to the merge conflict. Can you fix that?

gregorydlogan:
@lkiesow should be fixed now!

PR #845, opened 2019-04-25
MH-13447, Admin Interface Feeds Endpoint

ek926m:
I need this PR for an another ticket. I can use this backend functionality to show available feed data for the admin ui.


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
```diff
@@ -0,0 +1,65 @@
...
+  public FeedEndpoint() {
+    super("org.opencastproject.feed.impl.FeedServiceImpl");
+  }
+
+  @GET
+  @Path("/listFeedServices")
+  @Produces(MediaType.APPLICATION_JSON)
+  public String listFeedServices() {
+
+    HttpGet get = new HttpGet("/listFeedServices");
```
staubesv: I would have rather expected a method like FeedGenerator[] getFeedGenerators() in the FeedServiceImpl. The FeedEndpoint (respectively the module AdminUI) would then require a dependency to the FeedService and use that method here to retrieve the feed generators (as Java objects) and create the response (JSON).

That way you also could avoid the weird mixup of XML and JSON in the FeedServiceImpl.

BTW: Personally, I would prefer GET /admin-ui/feeds over GET /admin-ui/feeds/listFeedServices

modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java
```diff
@@ -94,6 +97,34 @@
...
+   * /feeds/listFeedServices
+   *
+   */
+
+  @GET
+  @Path("/listFeedServices")
+  @Produces(MediaType.APPLICATION_JSON)
+  public String listFeedServices() {
+
+    Gson gson = new Gson();
```
staubesv: Gson is thread-safe. Please avoid creating per-request instances but rather re-use a static instance.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
lkiesow: `FeedServiceImpl` isn't necessarily available on the same node as the admin interface. It usually resides on the presentation node.

---

`GET /admin-ui/feeds` is probably more REST-like

modules/search-service-impl/pom.xml
```diff
@@ -15,6 +15,11 @@
     <checkstyle.skip>false</checkstyle.skip>
   </properties>
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.8.5</version>
```
lkiesow: Drop the version in here.
It's managed by the main pom.xml.

modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java
```diff
@@ -94,6 +97,34 @@
   /** The security service */
   private SecurityService securityService = null;
 
+  /*
+   *
+   * TEST
```
lkiesow: test?

lkiesow (changes requested):

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
ek926m: the current URL is /feeds/listFeedServices, I could change it to /feeds/admin-ui/feeds but this looks weird, probably /feeds/admin-ui, but I don't know If I can change the root "feeds" to something else, it seems that "feeds" is fix

staubesv: The URL `/admin-ui/feeds` comes from https://github.com/opencast/opencast/pull/845/files#diff-fbcfee710538a23c249dab87b06f3769R8 and is fine.

What I meant is to directly use this URL as `GET /admin-ui/feeds` already suggests that it retrieves the list of feeds. Looking at other similar requests in the Admin UI Facade, e.g. `GET /admin-ui/series/series.json` (see https://github.com/opencast/opencast/blob/develop/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java#L536), it seems that `GET /admin-ui/feeds/feeds.json` would be most consistent with other list retrieval requests in the Admin UI (vs. the current URL `GET /admin-ui/feeds/listFeedServices`. Depending on what a FeedService is supposed to be, maybe `GET /admin-ui/feeds/services.json`

ek926m: I changed it to:
/feeds/feeds.json (backend)
/admin-ng/feeds/feeds.json (frontend -> req to backend in another branch)

I hope that this is okay

KatrinIhler:
I guess this is ready for review again?

ek926m:
> I guess this is ready for review again?

yes please, and the PR #912 is the other PR mentioned in the initial sentence

KatrinIhler:
@ek926m So does this mean it's no longer just a draft but could be merged? In that case please change the type of pullrequest! :)

staubesv:
@lkiesow We are not using feeds, but it seems you are. While looking at the code here, I've found https://opencast.jira.com/browse/MH-13572 which would be easy to fix.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /** Logging facility */
+  private static Logger logger = LoggerFactory.getLogger(FeedEndpoint.class);
+
+  @GET
+  @Path("/feeds")
+  @Produces(MediaType.APPLICATION_JSON)
+  public String listFeedServices() {
+
+    HttpGet get = new HttpGet("/feeds");
```
staubesv, line 56: @ek926m I don't think this works on distributed setups where the search service is running on the presentation node and whereas the Admin UI feed endpoint runs on the admin node, does it?

I would tend to say that instead of just redirecting the call to the feed service, a dependency to the search service should be added (if not yet there). 

@lkiesow Or what would you think?

lkiesow: The search service is not present on an admin node and this actually implements a remote for that service so that it will automatically locate the remote service location and act more or less as a proxy. So I think this should work. I haven't tested it though…

staubesv: Ah sorry, have just looked at HttpGet("/feeds"); and not the RemoteBase stuff...

ek926m:
@staubesv is the review done? could this be merged?

staubesv:
@ek926m Sorry, I did not have time to work on this. I unassigned myself from the review as I want have time anymore to do it...

ek926m:
@staubesv okay

gregorydlogan:
I'll take this since it's blocking #912

gregorydlogan:
`http://localhost:8080/admin-ui/feeds" spits out the following:

```json
{"version":"v1.3.0","url":"http:\/\/localhost:8080\/api"}
```

by default.  I'm assuming it should be outputting more/better information than that?

gregorydlogan:
@ek926m any idea when this will be moved forwards?

ek926m:
@gregorydlogan could you give me an example, how http://localhost:8080/admin-ng/feeds could/should look?

ek926m:
@gregorydlogan are you looking for information like in #912 http://localhost:8080/feeds/feeds

modules/search-service-impl/src/main/java/org/opencastproject/feed/impl/FeedServiceImpl.java
```diff
@@ -94,6 +97,36 @@
...
+  @GET
+  @Path("/feeds")
+  @Produces(MediaType.APPLICATION_JSON)
+  public String listFeedServices() {
+
+    List<Map<String, String>> feedServices = new ArrayList<>();
+
+    for (FeedGenerator generator : feeds) {
+      Map<String, String> details = new HashMap<>();
+      details.put("identifier", generator.getIdentifier());
```
gregorydlogan: The identifier you are getting here is not correct.  I'm seeing this render as (for example) `http://localhost:8080/series`, when it should be rendering as:
* `http://localhost:8080/feeds/rss/2/series/338f6880-2ccc-4869-8133-c53d3c4de430` (for rss)
* `http://localhost:8080/feeds/atom/1/series/338f6880-2ccc-4869-8133-c53d3c4de430` (for atom).

Ideally this method would take a parameter setting the type of feed to return, but if that's not possible then it should return both, or rss.

I get that given the endpoint's parameters you can't know the some of the parameters for the feed (like the series ID), but returning an identifier with a marker for substitution would make sense.  In other words, `http://localhost:8080/feeds/atom/1/series/$seriesid$`, or some other easy-to-parse marker so that the admin ui can replace the marker with the correct series id.

ek926m: so instead of showing a list of feedgenerator data I just need to show the atom/rss links to the series itself?

rss: 
https://develop.opencast.org/feeds/rss/2.0/series/20f4cc55-db54-421d-b593-254b44262cae

atom:
https://develop.opencast.org/feeds/atom/1.0/series/20f4cc55-db54-421d-b593-254b44262cae


**instead** of this:

61053502-f6850480-a3ed-11e9-903e-195718176147

So my solution was fundamentally wrong?!

gregorydlogan: No, you are fundamentally right :)  Your initial code is basically right, and actually nice and clean despite my nit picking.  You just need to modify the identifier the service is sending you into something which works when clicked on.  I would leave the feed service alone, and just fix the URLs using some string magic in the UI to render the correct values (like the ones you posted above) in the identifier column.  You might need to do some CSS tricks to wrap the column nicely so it doesn't run off the side of the screen, but you're very very close to this being done!

ek926m: @gregorydlogan I am working on it, I just have not commited by now, I send you an update until I need feedback

ek926m: @gregorydlogan my current feed tab view looks like this, are there links missing? I still need to git commit if this view is right.
Screenshot from 2019-09-05 09-37-21

gregorydlogan: Yep, those urls look right to me!

ek926m: @gregorydlogan pls look at #1050, this is a updated one from this

PR #848, opened 2019-04-25
MH-13508 Remove method canLogin from interface User

staubesv:
The interface User offers a method canLogin which is not really used in means of it is used to set the isEnabled of https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/core/userdetails/User.html but it always sets it to true.

The goal of this maintenance task is to remove that code from our codebase since it is not really actually used.

In case we need this in the future, we might also decide to make the field persistent in the database and call it "active" or something like that.


KatrinIhler (approved):
Looks good, let's get rid of this!

PR #850, opened 2019-05-01
MH-13514 Add node name to host registration as a UI searchable alternative to hostname

JamesUoM:
Define in custom.properties and use in Admin UI : Systems

- Field is “Node name” not “Nodename” hence camelCase and underscores in variable and constant names
- Service API’s that use hostname have not been changed.


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/JobEndpoint.java
```diff
@@ -743,4 +787,18 @@ public int compare(Job job1, Job job2) {
       return ascending ? result : -1 * result;
     }
   }
+
+  /**
+   * @param hostname of server to find in list
+   */
+  private Optional<HostRegistration> findServerByHost(String hostname, List<HostRegistration> servers) {
+    // Java 1.8 return servers.stream().filter(o -> o.getBaseUrl().equals(hostname)).findFirst();
```
staubesv: You can use Java 8 in opencast/develop

JamesUoM: Yeah, I remembered last night I hadn't updated the code for java  8 (ported from our 3.7 java 7 repo)

staubesv (approved):

lkiesow:
The default of `AllInOne` does not fit in most cases and is hence not the best default. Can we set the default to something which usually works by default? Using the hostname would be an option. Or we could generate the default based on the installed distribution (identified by the Karaf feature)?

JamesUoM:
> The default of `AllInOne` does not fit in most cases and is hence not the best default. Can we set the default to something which usually works by default? Using the hostname would be an option. Or we could generate the default based on the installed distribution (identified by the Karaf feature)?

If not set, then nodename is null by default. The value set custom.properties could be injected when the assemblies are built?

lkiesow:
You could but that would be a bit hacky. My suggestion would be to
either obtain and use the hostname by default or to get the installed
features from the karaf features service.

Since the prior method is simpler and leyy likely to break, I would
prefer that.

You mentioned that it's `null` by default, what happens if the value is
is set to `null`? Would that be a sensible default as well?

–Lars

PR #851, opened 2019-05-01
MH-13496 Highlight main table rows on hover

krinnewitz:
This patch adds an optional "highlight" flags to the table directive. When this flag is set to true, table rows are highlighted when hovered.

The following image illustrates the feature (unfortunately my screenshot tool hides my mouse pointer):

highlight


lkiesow:
Looks good → merged

PR #853, opened 2019-05-02
MH-13330 The video editor does not always close after the user presses "Publish"

staubesv:
Steps to reproduce: 
1. Open the video editor 
2. Remove the first 10 seconds of video 
3. Press on "Publish" 
  
 Actual Results: 
 The video editor should close after the user presses on publish 
  
 Expected Results: 
 The video editor does not close after the user presses on publish 
  
 Workaround (if any): 
 Ignore 

Analysis: 
The logic that determines whether the video editor should close after the request returns uses the variable $scope.video.workflow. 
In case the default thumbnail changes because of the cutting, this variable is overwritten by the save request. 
The solution is to use a separate variable to store the state of $scope.video.workflow before performing the save reques


KatrinIhler (approved):
Bug reproduced, fix tested, yadda yadda, looks good.

lkiesow:
@staubesv, since this patch conflicts with the `r/7.x` code, can you maybe merge `r/6.x` to `r/7.x`? You know the code already.

PR #854, opened 2019-05-03
MH-13518 move ingest-download Operation to worker

CGreweling:
The aim is to reduce the file io load on the admin node.

So This moves the ingest-download-operation to the worker nodes.
The Ingestdownload WOH is rewritten as a service with impl and remote operation.

Work sponsored by SWITCH


smarquard:
Or you can just mount the shared workspace on the ingest nodes directly, and leave out the ingest-download WOH step completely.

staubesv:
@smarquard In our case, we need to download from a separate storage which is not part of the Opencast cluster.

wsmirnow:
@smarquard the `ingest-download` operation is not hosted on the ingest node like the `ingest` operation. This means, all `ingest-download` jobs are currently running on the admin node. This patch should change it.

The `ingest-download` operation can reduce the IO load by (selective) downloading the media files into the shared workspace (e.g. all files from Asset Manager). This it is only useful, if this operation is not running on the admin node but distributed to the workers.

modules/ingest-download-service-impl/src/main/java/org/opencastproject/ingestdownloadservice/impl/endpoint/IngestDownloadServiceEndpoint.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /** The service */
+  private IngestDownloadService service;
+  private ServiceRegistry serviceRegistry;
+
+  @POST
+  @Path("ingestdownload")
+  @Produces(MediaType.TEXT_XML)
+  @RestQuery(name = "ingestdownload",description = "Downloads mediapackage elements to workspace",
+          restParameters = { @RestParameter(description = "mediapackage as xml", isRequired = true, name = "mediapackage",
```
gregorydlogan, line 69: I feel like these docs need work.  A few don't specify what you're separating by, and the booleans don't specify what they actually do.

modules/ingest-download-service-impl/src/main/java/org/opencastproject/ingestdownloadservice/impl/IngestDownloadServiceImpl.java
```diff
@@ -0,0 +1,348 @@
...
+
+      // Store original URI for deletion
+      externalUris.add(originalUri);
+    }
+
+    if (!deleteExternal || externalUris.size() == 0)
+      return MediaPackageParser.getAsXml(mediaPackage);
+
+    // Find all external working file repository base Urls
+    logger.debug("Assembling list of external working file repositories");
```
gregorydlogan, line 286: Ok, wait a minute.  You're checking for the case where there is more than one WFR in the cluster, and then telling it to remove the file(s) if they exist.  What's the usecase here?  If we're fetching a file from an external host, how can said file already be present in another WFR?

CGreweling: Hmm good question. I just copied the old code here. 
Maybe this is possible if we have a cluster without a shared (nfs)medium?
But to have a workspace and a workingfile-repository seems to me, that we have one storage to much... 

If this check is not necessary, i would be happy to remove it.

modules/ingest-download-service-impl/src/main/java/org/opencastproject/ingestdownloadservice/impl/endpoint/IngestDownloadServiceEndpoint.java
CGreweling: done. :)

modules/ingest-download-service-impl/src/main/java/org/opencastproject/ingestdownloadservice/impl/IngestDownloadServiceImpl.java
gregorydlogan: If this was copy and pasted then likely the check originated back in the days when the WFR could be in multiple places.  I don't think it's likely to be useful now, but there might be some weird edge case that would show up later.  Let's leave it for now I guess, at some point someone will rip out the WFR entirely, and they will remove this check at that point.

PR #855, opened 2019-05-03
MH-13519 Migrate mappings to Elastic Search 5.x

staubesv:
As for Opencast 7.0, we are about to upgrade ElasticSearch from 1.x up to at least 5.x. While the 1.x mapping configuration still seems to work, some types have been deprecated.

This work addresses the following issues:

- The type "string" has been split up into the types "keyword" and "text" and should not be used in ElasticSearch 5.x or higher anymore
- The mapping parameter "index" expects the values true or false (default: true)
- The mapping parameter "store" expects the values true or false (default: false)
- Remove unused dynamic templates
- Use type "text" for the fields "objects" to be able to store large values here

### DIFF
--- a/modules/search/src/main/resources/elasticsearch/event-mapping.json
+++ b/modules/search/src/main/resources/elasticsearch/event-mapping.json
@@ -1,117 +1,117 @@
 {
     "event" : {
         "_source" : { "enabled" : true },
-        "dynamic": "true",
+        "dynamic": true,
         "properties" : {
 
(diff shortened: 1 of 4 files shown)


lkiesow:
@staubesv, since the ES upgrade is in `r/7.x`, can you redirect this pull request?
…and thanks for taking a look at this.

staubesv:
@lkiesow Done.

PR #856, opened 2019-05-03
MH-13503: Job Dispatch Fairness

gregorydlogan:
This pull request changes the logic around job dispatch from lowest current load first, to lowest load factor first.  For instance, in a situation with three nodes, with maximum loads of 1, 2, and 4 the dispatching order looks like this:

Node | Max load
-------|-------------
Node 1 | 1
Node 2 | 2
Node 3 | 4

Current number of jobs running | Node 1 Load | Node 1 Factor | Node 2 Load | Node 2 Factor | Node 3 Load | Node 3 Factor | Notes
-|-|-|-|-|-|-|-
0 | 0 | 0.0 | 0 | 0.0 | 0 | 0.0 |
1 | 1 | 1.0 | 0 | 0.0 | 0 | 0.0 | Already suboptimal, Node 1 is maxed out, should have gone to Node 3
2 | 1 | 1.0 | 1 | 0.5 | 0 | 0.0 | This job should have gone to Node 3 as well
3 | 1 | 1.0 | 1 | 0.5 | 1 | 0.25 | Finally!
4 | 1 | 1.0 | **2** | **1.0** | 1 | 0.25 | This is the opposite of what we want, Node 3 is basically idle!
5 | 1 | 1.0 | 2 | 1.0 | 2 | 0.5 |  From here on, the jobs go to Node 3, but only because everyone else is already maxed out!
6 | 1 | 1.0 | 2 | 1.0 | 3 | 0.75 | 
6 | 1 | 1.0 | 2 | 1.0 | 4 | 1.0 | 

The new dispatch looks like this

Current number of jobs running | Node 1 Load | Node 1 Factor | Node 2 Load | Node 2 Factor | Node 3 Load | Node 3 Factor | Notes
-|-|-|-|-|-|-|-
0 | 0 | 0.0 | 0 | 0.0 | 0 | 0.0 |
1 | 0 | 0.0 | 0 | 0.0 | 1 | 0.25 | In the case of a tie, dispatch to the node with the largest maximum load
2 | 0 | 0.0 | 1 | 0.5 | 1 | 0.25 | Dispatching to lowest load factor, breaking the tie by max load
3 | 1 | 1.0 | 1 | 0.5 | 1 | 0.25 | 
4 | 1 | 1.0 | 1 | 0.5 | 2 | 0.5 | Largest node is being hit hardest
5 | 1 | 1.0 | 1 | 0.5 | 3 | 0.75 | Same tie breaking rules
6 | 1 | 1.0 | 2 | 1.0 | 3 | 0.75 |
7 | 1 | 1.0 | 2 | 1.0 | 4 | 1.0 |

Note that the smaller worker only becomes saturated after 6 jobs are running, vs 4 with the original dispatch logic.  This becomes even more obvious as the maximum load scales - the larger a worker is relative to the rest of the cluster, the more jobs it will attract before the system will start dispatching to weaker workers.

Long term it would be nice to scale things such that especially weak workers are used only as a last resort, however this approach seems good enough for now, and clusters with completely crippled workers present are likely to be pretty rare.

### DIFF
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -161,7 +161,7 @@ public class ServiceRegistryJpaImplTest {
     serviceRegistryJpaImpl.setUserDirectoryService(userDirectoryService);
 
     final Capture<HttpUriRequest> request = EasyMock.newCapture();
-    final BasicHttpResponse successRespone = new BasicHttpResponse(
+    final BasicHttpResponse successResponse = new BasicHttpResponse(
             new BasicStatusLine(new HttpVersion(1, 1), HttpStatus.SC_NO_CONTENT, "No message"));
     final BasicHttpResponse unavailableResponse = new BasicHttpResponse(
             new BasicStatusLine(new HttpVersion(1, 1), HttpStatus.SC_SERVICE_UNAVAILABLE, "No message"));
@@ -178,7 +178,7 @@ public class ServiceRegistryJpaImplTest {
         if (request.getValue().getURI().toString().contains(TEST_PATH_3))
           return unavailableResponse;
 
-        return successRespone;
+        return successResponse;
       }
     }).anyTimes();
     EasyMock.replay(trustedHttpClient);
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -406,9 +406,13 @@ public class ServiceRegistryJpaImplTest {
     JobBarrier barrier = new JobBarrier(null, serviceRegistryJpaImpl, testJob);
     try {
       barrier.waitForJobs(2000);
+      //We should never successfully complete the job, so if we get here then something is wrong
       Assert.fail();
     } catch (Exception e) {
       testJob = serviceRegistryJpaImpl.getJob(testJob.getId());
+      //Some explanation here: If the load exceeds the global maximum node load (ie, jobLoad > all individual node max
+      // loads), then we dispatch to the biggest, even if it's not going to normally accept the job.  That node may still
+      // reject the job, but that's AbstractJobProducer's job, not the service registry's
       Assert.assertEquals(TEST_HOST_OTHER, testJob.getProcessingHost());
     }
   }
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -39,6 +39,7 @@ import org.opencastproject.security.api.User;
 import org.opencastproject.security.api.UserDirectoryService;
 import org.opencastproject.serviceregistry.api.ServiceRegistration;
 import org.opencastproject.serviceregistry.api.ServiceRegistryException;
+import org.opencastproject.serviceregistry.api.SystemLoad;
 import org.opencastproject.systems.OpencastConstants;
 import org.opencastproject.util.NotFoundException;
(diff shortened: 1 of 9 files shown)


gregorydlogan:
Tagging committers for review who I know have large systems.  Still interested in hearing from those who I haven't tagged!

JamesUoM (approved):
Looks fine to me. As I understand it this would not affect a system of homogenous nodes as the current load and load factor are equivalent.

smarquard:
The logic and implementation of this look good to me. We will try this out in our 6.x production system in due course.

smarquard:
Job allocation does indeed look fairer. These 3 columns show cores, completed, running:

image

staubesv:
> Long term it would be nice to scale things such that especially weak workers are used only as a last resort

Hmm... I don't think so: The job dispatcher should dispatch jobs to all available workers even if some are much weaker than others. Not dispatching jobs to  weak workers might be a bad idea as an idle weak worker with maxLoad 1.0/currentLoad 0.0 (1.0 available) might be faster to execute a 1.0 job load than a very strong worker with maxLoad 100.0/currentLoad:98 (2.0 available) even though the load suggests that the stronger worker has more "power " left.

smarquard (approved):
This code change does the right thing on our 6.x production cluster.

smarquard:
Sorry @staubesv didn't notice your assignment until after I merged this.

Any remaining concerns?

staubesv:
@smarquard No, I was just about to hit the merge button ;-)

gregorydlogan:
> > Long term it would be nice to scale things such that especially weak workers are used only as a last resort
> 
> Hmm... I don't think so: The job dispatcher should dispatch jobs to all available workers even if some are much weaker than others. Not dispatching jobs to weak workers might be a bad idea as an idle weak worker with maxLoad 1.0/currentLoad 0.0 (1.0 available) might be faster to execute a 1.0 job load than a very strong worker with maxLoad 100.0/currentLoad:98 (2.0 available) even though the load suggests that the stronger worker has more "power " left.

@staubesv Ideally there would be a costing system for both the jobs, and the workers.  Right now, you can make a job really expensive to keep it off of weaker workers, but that might mean artificially boosting the cost of other jobs too.  What we really need is a way to mark workers are higher priority to accept jobs (ie, they get first crack), but that's waaaay outside the scope of this PR :)

lkiesow:
> Right now, you can make a job really expensive to keep it off of weaker workers,

If you want a specific job type of a certain worker type you could also build an assembly with no service accepting that particular job type. For example, the ETH is running workers dedicated to composer jobs and ones which will never do encodings. Let's not make the job dispatching even more complex as it already is.

modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
```diff
@@ -350,9 +406,13 @@ public void testDispatchingJobsHigherMaxLoad() throws Exception {
...
     try {
       barrier.waitForJobs(2000);
+      //We should never successfully complete the job, so if we get here then something is wrong
       Assert.fail();
     } catch (Exception e) {
       testJob = serviceRegistryJpaImpl.getJob(testJob.getId());
+      //Some explanation here: If the load exceeds the global maximum node load (ie, jobLoad > all individual node max
+      // loads), then we dispatch to the biggest, even if it's not going to normally accept the job.  That node may still
+      // reject the job, but that's AbstractJobProducer's job, not the service registry's
       Assert.assertEquals(TEST_HOST_OTHER, testJob.getProcessingHost());
```
karendolan, line 416: @gregorydlogan Should this job really have a processing Host here while still in QUEUED state? This job is in QUEUE'd state because it was rejected by the host_other. What is the point of preserving the attempted  dispatch if that host will not take the job? [UPDATE] @rute-santos found an issue with one JpaJob being unexpectedly updated by another JpaJob's  data after one is updated from persistence data, when the two, seemingly separate, JpaJob  objects share the same id attribute. It might be related to this testJob still being in QUEUE'd state instead of DISPATCHING state.

rute-santos: The issue @karendolan refers to is explained in https://opencast.jira.com/browse/MH-13675 (see problem 2 there). I will provide a fix soon.

PR #860, opened 2019-05-06
MH-13520 Remove unused fields from search index

staubesv:
The Elastic Search index contains a number of fields which we actually do not use at all.

The goal of this task is to remove those fields.


staubesv:
@lkiesow I couldn't resist ;-)

PR #861, opened 2019-05-06
MH-13522 Add option to configure state mappings for workflows

krinnewitz:
This patch adds the possibility to define so-called state mappings for
workflows. State mappings tell what to display in the admin UI when a
workflow is in one of the mapped workflow states. For example, you can
configure that the admin UI should display "Retracting" instead of
"Running" when the retract workflow is active. When no state mappings
are given, the default mappings are used (exactly as before). This is
mainly an usability improvement.

localhost_8080_admin-ng_index html

No migration or index rebuild is needed to make use of the new mappings.
Just extend your workflow defintions accordingly.


docs/scripts/activemq/activemq.xml
```diff
@@ -146,6 +146,16 @@
               </compositeQueue>
             </virtualDestinations>
           </virtualDestinationInterceptor>
+          <virtualDestinationInterceptor>
```
lkiesow: Just a minor thing but can you add a section to the update docs that the activemq configuration needs to be update?

krinnewitz: sure, done

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -507,7 +507,13 @@
...
          "RECORDING_FAILURE": "Recording failure",
          "PROCESSING_FAILURE": "Processing failure",
-         "PROCESSING_CANCELED": "Processing canceled"
+         "PROCESSING_CANCELED": "Processing canceled",
+         "RETRACTING": "Retracting",
+         "DUPLICATING": "Duplicating",
+         "PUBLISHING": "Publishing",
+         "IMPORTING": "Importing",
+         "REPUBLISHING": "Republishing",
+         "UPDATING_PREVIEWS": "Updating Previews"
```
staubesv: "Updating **p**reviews" to be consistent with the other statuses

docs/guides/admin/docs/workflowoperationhandlers/index.md
```diff
@@ -92,3 +92,34 @@ The following table contains the workflow operations that are available in an ou
...
+|**instantiated**| The workflow is queued and will be started as soon as possible | "Pending" |
+|**running**| The workflow is running, no problems so far | "Running" |
+|**stopped**| The workflow was aborted by the user | "Processing canceled" |
+|**paused**| The workflow was paused and can be continued | "Paused" |
+|**succeeded**| The workflow has completed successfully | "Finished" |
+|**failed**| The workflow failed due to an error | "Processing failure" |
+|**failing**| The workflow is still running, but there were errors. It will fail. | "Running" |
+
+Using so-called state mappings, it is possible to refine the labels displayed in the Admin UI events table for a
+particular workflow.
```
KatrinIhler: re-define?

etc/workflows/republish-metadata.xml
```diff
@@ -114,4 +114,9 @@
 
   </operations>
 
+  <state-mappings>
+    <state-mapping state="running">EVENTS.EVENTS.STATUS.REPUBLISHING</state-mapping>
+    <state-mapping state="failing">EVENTS.EVENTS.STATUS.REPUBLISHING</state-mapping>
+  </state-mappings>
+
```
KatrinIhler, line 121: Maybe call this "publishing" as well since we can also publish an already published event, therefore republishing? Or being more specific and calling it "Republishing Metadata"?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -507,7 +507,13 @@
...
          "RECORDING_FAILURE": "Recording failure",
          "PROCESSING_FAILURE": "Processing failure",
-         "PROCESSING_CANCELED": "Processing canceled"
+         "PROCESSING_CANCELED": "Processing canceled",
+         "RETRACTING": "Retracting",
+         "DUPLICATING": "Duplicating",
+         "PUBLISHING": "Publishing",
+         "IMPORTING": "Importing",
+         "REPUBLISHING": "Republishing",
+         "UPDATING_PREVIEWS": "Updating previews"
```
KatrinIhler, line 518: Maybe separate the actual event statuses and the displayable statuses here?

modules/admin-ui/src/main/webapp/scripts/modules/events/partials/eventsStatusCell.html
```diff
@@ -1,5 +1,5 @@
 <a class="crosslink"
    ng-click="table.addFilterToStorage('status', row.event_status_raw)"
    title="{{ 'EVENTS.EVENTS.TABLE.TOOLTIP.STATUS' | translate }}">
-  {{row.event_status}}
+  {{row.displayable_status}}
```
KatrinIhler, line 4: Maybe make displayable status optional and fall back to event status if it's not defined?
This would it also make unnecessary to update all test data.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowStateMappingImpl.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * @param state
+   *          The state to map
+   * @param value
+   *          The value to use for the state
+   */
+  public WorkflowStateMappingImpl(WorkflowInstance.WorkflowState state, String value) {
+    this();
```
KatrinIhler: Why is this necessary?

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/workflow/SerializableWorkflowStateMapping.java
```diff
@@ -0,0 +1,45 @@
...
+ * the License.
+ *
+ */
+
+package org.opencastproject.message.broker.api.workflow;
+
+import java.io.Serializable;
+
+public class SerializableWorkflowStateMapping implements Serializable {
+
```
KatrinIhler: Why does this not extend WorkflowStateMapping / why is WorkflowStateMapping itself not simply serializable?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
```diff
@@ -124,6 +154,23 @@ public boolean accept(File arg0, String name) {
...
+        WorkflowDefinitionItem.WORKFLOW_DEFINITION_QUEUE,
+        MessageSender.DestinationType.Queue,
+        new WorkflowDefinitionItem(def.getId(),
+            def.getStateMappings()
+                .stream()
+                .map(m -> new SerializableWorkflowStateMapping(m.getState().name(), m.getValue()))
+                .toArray(SerializableWorkflowStateMapping[]::new)
+        )
+      )
+    );
```
KatrinIhler: Do we really need to send messages for this? I feel like it would be nicer to keep and update the workflowStateMappings somewhere in the WorkflowDefinitionScanner, create an interface for the Scanner in workflow.api (as the IndexService and others already depends on that module for the WorkflowService) and then ask the scanner to do the mapping or hand the mappings over each time? I feel like Event shouldn't hold that much knowledge (although it already does for other things...).

KatrinIhler: (Also does this mean that on system start, there's a message sent for each workflow definition?)

modules/index-service/src/main/java/org/opencastproject/index/service/message/WorkflowDefinitionMessageReceiverImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  protected void execute(WorkflowDefinitionItem workflowDefintionItem) {
+    logger.debug("Received workflow definition item for definition id '{}'", workflowDefintionItem.getWorkflowDefinitionId());
+
+    if (workflowDefintionItem.getStateMappings().length > 0) {
+      final Map<String, String> mappings = Arrays
+          .stream(workflowDefintionItem.getStateMappings())
+          .collect(Collectors.toMap(SerializableWorkflowStateMapping::getState, SerializableWorkflowStateMapping::getValue));
+      Event.customWorkflowStatusMapping.put(workflowDefintionItem.getWorkflowDefinitionId(), mappings);
```
KatrinIhler: I feel like updating the map from the outside is a bit invasive, I'd prefer a method in Event for this.

modules/admin-ui-index/src/main/resources/OSGI-INF/message-receiver-workflow-definition.xml
```diff
@@ -0,0 +1,33 @@
...
+
+  <reference name="search-index"
+             interface="org.opencastproject.adminui.index.AdminUISearchIndex"
+             bind="setSearchIndex"/>
+
+  <reference name="security-service"
+             interface="org.opencastproject.security.api.SecurityService"
+             bind="setSecurityService"/>
+
+</scr:component>
```
KatrinIhler: I don't quite understand how there are two message receivers with only one implementation, could you elaborate how this works?

KatrinIhler (changes requested):
Some thoughts below.

docs/guides/admin/docs/workflowoperationhandlers/index.md
staubesv: It indeed is **refine**. Example: The event status "Running" can be refined into "Deleting" and "Publishing" in means of providing additional information about what "Running" means. This does, however, not change the status of the event which is still "Running". It is really just some additional information about the status "Running".

KatrinIhler: Ah, now I get it, alright then.

smarquard: I would just drop the "so-called". You've called it, so that's what it's called :-) No need to undercut the term assigned for the feature.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
staubesv: As @krinnewitz has not yet answered: From the point of view of the Admin UI it does not really matter whether the string to be translated is an actual event status or displayable event status. It's always correct to just use those translation keys (for translating both the "actual" event statuses and "displayable" event statuses).

Why would you want too separate them? Seems to make thing more complicated...?

modules/admin-ui/src/main/webapp/scripts/modules/events/partials/eventsStatusCell.html
staubesv: ... then again, the "Status" row is supposed to just display the displayable status. Why should there be a fallback? Updating the test data is not an issue, is it?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
KatrinIhler: It might not matter for the Admin UI, but there is a difference between those two kinds, and when defining state mappings for workflows, it might be nice to be able to differentiate between the general statuses and the specific ones. I don't think this makes it more complicated, on the contrary I think it makes things clearer.

modules/admin-ui/src/main/webapp/scripts/modules/events/partials/eventsStatusCell.html
KatrinIhler: Thought about this some more and I guess it doesn't really matter.

krinnewitz:
Thanks @KatrinIhler. I like your proposal of not using activemq here, but extending the WorkflowService API. I have changed the code accordingly and also fixed your other findings.

docs/guides/admin/docs/upgrade.md
```diff
@@ -1,143 +1,42 @@
...
-
-```sql
--- Check OAI-PMH publications:
-select * from oc_oaipmh_elements where flavor = 'security/xacml';
--- Check engage publications:
-select * from oc_search where mediapackage_xml like '%"security/xacml"%';
--- Check asset manager:
-select * from oc_assets_snapshot where mediapackage_xml like '%"security/xacml"%';
-```
+Due to [MH-13522](https://opencast.jira.com/browse/MH-13522) one new queue was added, so you have to update your
```
staubesv: @krinnewitz I think that is now obsolete and the PR does not need to be mentioned in the upgrade guide anymore.

staubesv:
@KatrinIhler Any more open comments from your side?

modules/workflow-service-remote/src/main/java/org/opencastproject/workflow/remote/WorkflowServiceRemoteImpl.java
```diff
@@ -722,4 +725,20 @@ public void cleanupWorkflowInstances(int lifetime, WorkflowState state) throws W
...
     throw new WorkflowDatabaseException("Unable to successfully request the workflow cleanup endpoint");
   }
+
+  @Override
+  public Map<String, Map<String, String>> getWorkflowStateMappings() {
+    HttpGet get = new HttpGet("/statemappings.json");
+    HttpResponse response = getResponse(get);
+    try {
+      if (response != null) {
+        return (Map<String, Map<String, String>>) new JSONParser().parse(IOUtils.toString(response.getEntity().getContent(), "utf-8"));
```
krinnewitz, line 672: I am using JSON here while the rest of this class seems to use xml. However, the counterpart (WorkflowRestService) seems to provide JSON *and* xml for most endpoints. I opted in for JSON for the getWorkflowStateMappings() implementation since that was very easy to implement. However, I am not really happy with this. 

@staubesv @KatrinIhler any thoughts?

staubesv: @krinnewitz I don't know why those endpoints deliver both JSON and XML... likely, people missed the point that this would double the testing effort in case somebody would test this ;-)

I don't see an advantage in adding XML. In case somebody would actually want to use it, it could be added anytime.

KatrinIhler: Same.

KatrinIhler (changes requested):
A few things I noticed when testing:
1. This does not change the workflow status in the event details, is this intended?
2. This doesn't seem to workflows that are included in other workflows. Which maybe makes sense (what would that look like anyway?), but should maybe be documented somewhere.
3. "Succeeded" can't be mapped, probably because the "Finished" status in the event table is a mix of different statuses? Should likewise be documented.

krinnewitz:
> A few things I noticed when testing:
> 
>     1. This does not change the workflow status in the event details, is this intended?
> 
>     2. This doesn't seem to workflows that are included in other workflows. Which maybe makes sense (what would that look like anyway?), but should maybe be documented somewhere.
> 
>     3. "Succeeded" can't be mapped, probably because the "Finished" status in the event table is a mix of different statuses? Should likewise be documented.

Thanks @KatrinIhler. I just rebased the PR to fix merge conflicts and documented, that only the state of the including workflow is displayed if a workflow includes other workflows.

I just tried mapping "succeeded" and it works for me.

In the events details, I think it is okay to display the unmapped states because there, you can also see the name of the workflow, so it should be exactly clear what is going on.

KatrinIhler:
Mapping "Succeeded" still does not work for me, the table says "Finished" once the workflow is done...

KatrinIhler (approved):

KatrinIhler:
Never mind, my mistake. Let's merge this!

PR #862, opened 2019-05-06
MH-13516 add single step event deletion

krinnewitz:
This patch changes the event deletion via admin UI such that you don't
have to retract events before you are able to delete them. While that
was the case in the past, now, Opencast itself takes care of retracting
the events before deleting them if, necessary. This also applies for
bulk deletion.

This is achieved by first checking if the event to delete still has
publications and if so, running a retract workflow and waiting for its
completion. Once, a retract workflow is completed, the associated event
is deleted.

The criteria to tell if an event needs to be retracted is equal to the
one that was used in the admin UI frontend before.

The id of the workflow to use for event retraction is configurable.


modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -271,9 +258,7 @@
       "NEXT_STEP": "Next",
       "BACK": "Previous",
       "CREATE": "Create",
-      "RETRACT": "Retract",
       "RETRACT_DELETE": "Retract/Delete",
```
staubesv: @krinnewitz The button is still named "Retract/Delete". Please just name it "Delete" (and use the translation key "DELETE") as at the end of the day, pressing the button will delete the events. While this might involve retraction, that is a detail not of concern to the user as it also results in deletion.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -123,9 +123,11 @@
      "SERIES_NOT_UPDATED":  "The series could not be saved",
      "SERIES_NOT_UPDATED_ID":  "The series {{ id }} could not be saved",
      "SERIES_NOT_UPDATED_ALL":  "None of the series could be saved",
-     "EVENTS_DELETED":      "The event has been deleted",
-     "EVENTS_NOT_DELETED":  "The event(s) could not be deleted",
-     "EVENTS_NOT_DELETED_NOT_AUTHORIZED":  "The event could not be deleted, because you don't have the permission to do so.",
+     "EVENT_DELETED":      "The event has been deleted.",
+     "EVENT_WILL_BE_DELETED": "The event will be unpublished first, and will then be deleted automatically.",
+     "EVENTS_DELETED":      "The events are being deleted. Published events will be unpublished first, and then be deleted automatically.",
```
staubesv: @krinnewitz "The events are being deleted." is enough as this is what happens.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -123,9 +123,11 @@
      "SERIES_NOT_UPDATED":  "The series could not be saved",
      "SERIES_NOT_UPDATED_ID":  "The series {{ id }} could not be saved",
      "SERIES_NOT_UPDATED_ALL":  "None of the series could be saved",
-     "EVENTS_DELETED":      "The event has been deleted",
-     "EVENTS_NOT_DELETED":  "The event(s) could not be deleted",
-     "EVENTS_NOT_DELETED_NOT_AUTHORIZED":  "The event could not be deleted, because you don't have the permission to do so.",
+     "EVENT_DELETED":      "The event has been deleted.",
+     "EVENT_WILL_BE_DELETED": "The event will be unpublished first, and will then be deleted automatically.",
```
staubesv: @krinnewitz Just say "The events are being deleted.". Finally, this is what is happening.

staubesv:
@krinnewitz Please add a note to the upgrade guide explaining that the retract-id must be configured.

staubesv:
@krinnewitz When deleting an event with publications but not having configured the retraction workflow correctly, the Admin UI tells me that the events will be deleted (green message) and does not do anything.
In the log, I see 

`ERROR | (IndexServiceImpl:1533) - Could not find workflow with id retract to retract the event. Check your configuration.`

While this case is not of high relevance as it implies an incorrect configuration which won't happen in production environments (fingers crossed), the Admin UI could at least display an (possible unspecific) error notification ;-)

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -383,24 +385,30 @@ public Response getEventResponse(@PathParam("eventId") String id) throws Excepti
...
-    try {
-      checkAgentAccessForEvent(id);
-      if (!getIndexService().removeEvent(id))
-        return Response.serverError().build();
-    } catch (NotFoundException e) {
-      // If we couldn't find any trace of the event in the underlying database(s), we can get rid of it
-      // entirely in the index.
+    final Opt<Event> event = checkAgentAccessForEvent(id);
+    if (event.isNone()) {
+      return Response.serverError().build();
```
staubesv: Is this a really a server error? Seems more like a "NOT FOUND" or "UNAUTHORIZED"

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -432,31 +440,37 @@ public Response deleteEvents(String eventIdsContent) throws UnauthorizedExceptio
...
+        IndexService.EventRemovalResult currentResult;
+        if (event.isSome()) {
+          currentResult = getIndexService().removeEvent(
+            event.get(),
+            () -> { result.addNotFound(eventId); },
+            getAdminUIConfiguration().getRetractWorkflowId()
+          );
         } else {
-          result.addOk(eventId);
+          currentResult = IndexService.EventRemovalResult.FAILED;
```
staubesv: Same here: Why not "NOT FOUND" or "UNAUTHORIZED"?

modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/delete-events-modal.html
```diff
@@ -60,166 +51,18 @@ <h2 translate="BULK_ACTIONS.DELETE.EVENTS.CAPTION"><!-- Delete Events --></h2>
             </div> <!-- obj -->
           </div><!-- full-col -->
 
-          <div class="full-col obj" ng-if="!events.unpublished.has">
-            <p translate="BULK_ACTIONS.DELETE.EVENTS.UNPUBLISHED.NONE">
```
staubesv, line 64: This PR should also remove the BULK_ACTIONS.* translation keys if they are not used anymore.

krinnewitz: Actually there still was one unused BULK_ACTIONS.* key which I just removed. The other BULK_ACTIONS.* keys are still in use.

krinnewitz:
> @krinnewitz When deleting an event with publications but not having configured the retraction workflow correctly, the Admin UI tells me that the events will be deleted (green message) and does not do anything.
> In the log, I see
> 
> `ERROR | (IndexServiceImpl:1533) - Could not find workflow with id retract to retract the event. Check your configuration.`
> 
> While this case is not of high relevance as it implies an incorrect configuration which won't happen in production environments (fingers crossed), the Admin UI could at least display an (possible unspecific) error notification ;-)

Fixed

krinnewitz:
> @krinnewitz Please add a note to the upgrade guide explaining that the retract-id must be configured.

Fixed

staubesv:
@krinnewitz Please note that #926 causes the build of this PR to fail. Reason is a changed interface: https://github.com/opencast/opencast/pull/926/files#diff-2739771e472673d0bd918d001162c406R698

Can you please make this PR build again?

krinnewitz:
@staubesv done

staubesv:
@lkiesow & @KatrinIhler Just feeling amazed while deleting events on our test environment... any news on the review of this PR?

staubesv:
@krinnewitz Oh, I've found a missing peace myself...

Could you please...
- Refactor the workflow definition `retract` into `retract`, `delete` and `partial-retract` (the latter does the WOH retract-* and is included by both`retract` and `delete`, `delete` is only used by the admin ui to retract events on deletion in case needed)
- Set the status mappings of the workflow definition `delete` to "Deleting"
- Add a translation key for the workflow status mapping to "Deleting"

Goal: If a user deletes an event, we do want to show "Deleting..." rather than "Retracting..." in case the publications of the event to be deleted need to be retracted. At the end of the day, the user asked Opencast to delete the events (not to retract them even if deletion implies retraction).

Please set the description of the workflow definition delete to something like "This workflow is used by the Admin UI to retract all publications of an event upon deletion".

krinnewitz:
@staubesv done

staubesv:
@KatrinIhler & @lkiesow Any open comments to be resolved?

docs/guides/admin/docs/upgrade.md
```diff
@@ -13,3 +13,10 @@ How to Upgrade
 4. [Upgrade the database](#database-migration)
 
 You can find the database upgrade script in `docs/upgrade/7_to_8/`.
+
+Configuration Changes
```
KatrinIhler: Since Sven's proposal to move the configuration changes to the upgrade guide didn't pass as far as I remember, I'd prefer listing them shortly in the release notes (see 7.0 release for reference), since right now we have some at both places, which is confusing.

docs/guides/admin/docs/upgrade.md
```diff
@@ -13,3 +13,10 @@ How to Upgrade
...
+---------------------
+
+The admin UI configuration (`etc/org.opencastproject.adminui.cfg`) now has a new option `retract.workflow.id` which
+holds the id of the workflow used to retract events when deleting. This is used by the new single step event deletion
```
KatrinIhler: 'contains' instead of 'holds'

docs/guides/admin/docs/upgrade.md
```diff
@@ -13,3 +13,10 @@ How to Upgrade
...
+feature [MH-13516](https://opencast.jira.com/browse/MH-13516).
```
KatrinIhler: I know Sven disagrees with me, but I don't think it's necessary to list the issue number here, it's sufficient to know what has changed and maybe why (single step event deletion).

pmiddend:
Note that there will soon be a PR that is based on this one, so I'd be very nice to get this merged quickly.

staubesv:
@krinnewitz Please note that this PR currently has merge conflicts...

etc/workflows/delete.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>delete</id>
```
KatrinIhler, line 4: I don't really like that we now have three workflow files for basically the same task since the only thing that differs is the state mappings. Can't we solve this in a better way?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1518,6 +1536,41 @@ public void createGroup(String name, String description, String roles, String me
     return Opt.some(result.getItems()[0].getSource());
   }
 
+  @Override
+  public EventRemovalResult removeEvent(Event event, Runnable doOnNotFound, String retractWorkflowId)
+      throws UnauthorizedException, WorkflowDatabaseException, NotFoundException {
+    final boolean hasOnlyEngageLive = event.getPublications().size() == 1
+        && EventUtils.ENGAGE_LIVE_CHANNEL_ID.equals(event.getPublications().get(0).getChannel());
+    final boolean retract = event.hasPreview()
+        || (!event.getPublications().isEmpty()  && !hasOnlyEngageLive && this.hasSnapshots(event.getIdentifier()));
```
KatrinIhler, line 1545: In which case can we have publications, but no snapshots?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2166,4 +2219,47 @@ public boolean isWorkflowActive(String workflowState) {
             || WorkflowState.RUNNING.toString().equals(workflowState)
             || WorkflowState.PAUSED.toString().equals(workflowState);
   }
+
+  private static final class Retraction {
```
KatrinIhler: The IndexService is already long enough, I'd vote for putting this into its own file.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2166,4 +2219,47 @@ public boolean isWorkflowActive(String workflowState) {
...
+    private Runnable doOnNotFound;
+
+    private Retraction(User user, Organization organization, Runnable doOnNotFound) {
+      this.user = user;
+      this.organization = organization;
+      this.doOnNotFound = doOnNotFound;
+    }
+  }
+
+  private final class RetractionListener implements WorkflowListener {
```
KatrinIhler: Same for this. Could be the same file as Retraction.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -239,6 +240,7 @@
   //TODO Move to a constants file instead of declaring it at the top of multiple files?
   protected static final String WORKFLOW_DEFINITION_DEFAULT = "org.opencastproject.workflow.default.definition";
 
+
```
KatrinIhler: Unnecessary line.

KatrinIhler (commented):
God, I feel bad about procrastinating on this for so long...
I think it looks good, so I just have some minor points.

lkiesow:
I'm not sure but how did the discussion about the now missing retraction go?
Also, it seems like the first users start complaining about it being too easy to accidentally delete events.

staubesv:
@lkiesow The retraction is reachable via Actions->Start Task->Retract. I don't think that it is now missing as the delete button neither indicates the function "Retraction" nor was consistent with the delete button across the Admin UI nor was particular user-friendly. Besides that, if retract was available on that level, you might also expect publish and possibly other actions to be there.

Maybe there is a need of having faster access to other tasks (retract, publish, ...) directly in the events table (as discussed on list), but that is not scope of this PR.

IIRC, there will be a confirmation window that asks the user whether he/she really wants to delete the event(s). That seems very much the expected behaviour users should be familiar with from virtually any other software they use. Why should this be different in Opencast?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: Well, I more or less took these conditions from the existing code. I remember that there were changes to the event deletion code in the past to make it more robust against data inconsistencies. I agree that there should be no publications for events without snapshots, but reality might proof us wrong. So I would prefer to keep the condition as is to not introduce a bug.

etc/workflows/delete.xml
krinnewitz: Okay, so now there are two workflow defintions which have the same content. I think it's not easy to avoid the redundancy here. Also, the two workflow definitions have different use-cases. Maybe someone wants to do additional/other things when deleting an event instead of just retracting it. So this at least gives us more flexibility than before :)

krinnewitz:
Thanks for the review @KatrinIhler . I have fixed your findings and answered the two comments where I did not take further actions.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
KatrinIhler: Sounds reasonable.

KatrinIhler:
@krinnewitz There have been some worries that users might delete events accidentally. Maybe we could clarify in the dialog that events that are deleted are truly, irrevocably gone and this cannot be undone? (Since a lot of software has some sort of undo function, but opencast does not). Maybe even in a very eye-catching way? (Not sure exactly about the how without it getting too ugly, but maybe in red?)

krinnewitz:
> @krinnewitz There have been some worries that users might delete events accidentally. Maybe we could clarify in the dialog that events that are deleted are truly, irrevocably gone and this cannot be undone? (Since a lot of software has some sort of undo function, but opencast does not). Maybe even in a very eye-catching way? (Not sure exactly about the how without it getting too ugly, but maybe in red?)

Ok, I added a red warning to the event deletion confirmation dialog. The bulk-delete dialog already had a red warning message.

Screenshot_2019-08-21 Opencast

krinnewitz:
@KatrinIhler hopefully, I have addressed all your findings ;)

KatrinIhler:
@krinnewitz I'm happy with this so far, but @lkiesow was planning to discuss the implications of this PR on list shortly, so I'm gonna hold back from merging for now, sorry!

etc/org.opencastproject.adminui.cfg
```diff
@@ -231,3 +231,8 @@ sourcetrack.right.flavor=presentation/prepared
 #
 # Default: video+preview
 #preview.video.subtype=video+preview
+
+# ID of the workflow used to unpublish events when deleting.
+#
+# Default: delete
+# retract.workflow.id=delete
```
karendolan, line 238: @krinnewitz what is the Use Case for changing this default? Is the alternative to delete from Opencast asset manager/adminUI/adminSearch/adminIndex but not un-publish? 
A possibly unintended benefit of this config item is that it clearly identifies which workflow can be adjusted to change the default retractions.

krinnewitz: > @krinnewitz what is the Use Case for changing this default? Is the alternative to delete from Opencast asset manager/adminUI/adminSearch/adminIndex but not un-publish?
> A possibly unintended benefit of this config item is that it clearly identifies which workflow can be adjusted to change the default retractions.

@karendolan I don't have a specific use case in mind, but the alternative to making it configurable would be to hard-code the workflow id. I think by making it configurable, we have maximum flexibility as well as the documental benefit you mentioned. However, I agree that changing the workflow ID is a rare use case (if any at all) and people would rather edit the workflow itself.

staubesv: @karendolan & @krinnewitz Our use case is that we use a naming scheme where the workflow definition identifier includes both a prefix to clearly identify SWITCH custom workflow definitions and a workflow definition version number (e.g. switch-delete-1.0).
Besides that, I strongly dislike the idea of hard-coded identifiers ;-)

krinnewitz:
@KatrinIhler Since there were no objections on [the mailing list post](https://groups.google.com/a/opencast.org/forum/#!topic/dev/sguFnfA3TMc), I guess we can finally merge this :)

KatrinIhler (approved):

PR #863, opened 2019-05-06
MH-13512 Improve URL signing performance

staubesv:
The performance of URL signing is critical for External API request that retrieve publications of events where each URI of each published elements needs to be signed.

The current implementation does a brute force comparison of the URL to be signed vs. the URL prefixes that can be signed. This is not only because of the algorithm that is in O(n) but it also seems that the functional implementation requires some high-overhead transformations of data structures.

The goal of this task is to use a O(log(n)) algorithm and avoid expensive transformations of data structures.

The performance of URL signing is critical for External API request that retrieve publications of events where each URI of each published elements needs to be signed.

The current implementation does a brute force comparison of the URL to be signed vs. the URL prefixes that can be signed. This is not only because of the algorithm that is in O\(n) but it also seems that the functional implementation requires some high-overhead transformations of data structures.

The goal of this task is to use a O(log\(n)) algorithm and avoid expensive transformations of data structures.

Benchmark:
GET /api/events/{event_id}?withpublications=true&sign=true for a event with approx. 40 URLs to be signed takes approx. 40ms with the current code and 30ms with this optimization, whereas approx. 20ms are required without URL signing. So the overhead of enabling URL signing is cut in half.

Open questions:
- This code will not work with prefixes that are equal except of lowercase vs. uppercase differences.

### DIFF
--- a/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
+++ b/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
@@ -20,6 +20,7 @@
  */
 package org.opencastproject.security.urlsigning.provider.impl;
 
+import org.opencastproject.security.api.Organization;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.urlsigning.exception.UrlSigningException;
 import org.opencastproject.security.urlsigning.provider.UrlSigningProvider;
@@ -47,18 +48,17 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.TreeMap;
 
 public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, ManagedService {
-  /** The prefix for key entry configuration keys */
-  public static final String KEY_ENTRY_PREFIX = "key";
-  /** The postfix in the configuration file to define the encryption key. */
+  /** The prefix for key configuration keys */
+  public static final String KEY_PROPERTY_PREFIX = "key";
+  /** The attribute name in the configuration file to define the encryption key. */
   public static final String SECRET = "secret";
-  /** The postfix in the configuration file to define the matching url. */
+  /** The attribute name in the configuration file to define the matching url. */
   public static final String URL = "url";
-  /** The postfix in the configuration file to define the organization owning the key. */
+  /** The attribute name in the configuration file to define the organization owning the key. */
   public static final String ORGANIZATION = "organization";
 
   /** Value indicating that the key can be used by any organization */
@@ -78,16 +78,24 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
   public abstract Logger getLogger();
 
   /**
-   * A class to contain the necessary key entries for url signing.
+   * A class representing a URL signing key.
    */
-  private static class KeyEntry {
-    private String key = null;
-    private String url = null;
-    private String organization = ANY_ORGANIZATION;
+  private static class Key {
+    private String id = null;
+    private String secret = null;
+    private String organizationId = ANY_ORGANIZATION;
+
+    Key(String id) {
+      this.id = id;
+    }
+
+    boolean supports(String organizationId) {
+      return this.organizationId.equals(ANY_ORGANIZATION) || this.organizationId.equals(organizationId);
+    }
   }
 
-  /** The map to contain the list of keys, their ids and the urls they match. */
-  private Map<String, KeyEntry> keys = new HashMap<>();
+  /** A mapping of URL prefixes to keys used to lookup keys for a given URL. */
+  private TreeMap<String, Key> urls = new TreeMap<>();
 
   /**
    * @param securityService
@@ -101,24 +109,34 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
    * @return The current set of url beginnings this signing provider is looking for.
    */
   public Set<String> getUris() {
-    return keys.values().stream()
-            .map(keyEntry -> keyEntry.url)
-            .collect(Collectors.collectingAndThen(
-                    Collectors.toSet(),
-                    Collections::unmodifiableSet));
+    return Collections.unmodifiableSet(urls.keySet());
   }
 
   /**
-   * If available get a {@link KeyEntry} if there is a matching Url matcher.
+   * Get{@link Key} for a given URL.
+   * This method supports multi-tenancy in means of only returning keys that can be used by the current
+   * organization. In case the current organization cannot be determined, no key will be returned. 
    *
    * @param baseUrl
-   *          The url to check against the possible matchers.
-   * @return The {@link KeyEntry} if it is available.
+   *          The URL that needs to be signed.
+   * @return The {@link Key} if it is available.
    */
-  private Optional<Map.Entry<String, KeyEntry>> getKeyEntry(String baseUrl) {
-    return keys.entrySet().stream()
-            .filter(entry -> baseUrl.startsWith(entry.getValue().url))
-            .findAny();
+  private Key getKey(String baseUrl) {
+    /* Optimization: Use TreeMap.floorEntry that can retrieve the greatest URL equal to or greater than 'baseUrl'
+       in O(log(n)). As we are trying to find an URL that is a prefix of 'baseUrl', candidate.getKey() either is
+       that URL (needs to be checked!) or there is no such URL. */
+    Map.Entry<String, Key> candidate = urls.floorEntry(baseUrl);
+    if (candidate != null && baseUrl.startsWith(candidate.getKey())) {
+      Key key = candidate.getValue();
+
+      // Don't accept URLs without an organization context
+      // (for example from the ServiceRegistry JobProducerHeartbeat)
+      Organization organization = securityService.getOrganization();
+      if (organization != null && key.supports(organization.getId())) {
+        return key;
+      }
+    }
+    return null;
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -196,39 +219,26 @@ public boolean accepts(String baseUrl) {
...
-    if (!accepts(policy.getBaseUrl())) {
+    Key key = getKey(policy.getBaseUrl());
+    if (key == null) {
       throw UrlSigningException.urlNotSupported();
     }
 
     policy.setResourceStrategy(getResourceStrategy());
 
     try {
-      // Get the key that matches this URI since there must be one that matches as the base url has been accepted.
```
staubesv, line 223: Strictly spoken, it could have been removed meanwhile. Besides that, we don't want to lookup the key twice.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -130,62 +146,69 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
-                  .map(keyEntry -> keyEntry.url)
-                  .filter(Objects::nonNull)
-                  .anyMatch(url -> propertyValue.startsWith(url) || (url != null && url.startsWith(propertyValue)))) {
+          if (urls.keySet().stream().anyMatch(v -> propertyValue.startsWith(v) || (v.startsWith(propertyValue)))) {
             throw new ConfigurationException(propertyKey,
-                    "there is already a key configuration for a URL with the prefix " + propertyValue);
+                    "There is already a key configuration for a URL with the prefix " + propertyValue);
           }
-          currentKeyEntry.url = propertyValue;
+          /* We explicitely support multiple URLs that map to the same key */
```
staubesv: As properties is a set, it is currently not possible to add multiple URLs. That will be added by https://opencast.jira.com/browse/MH-13513 later.

staubesv:
@JulianKniephoff As you have worked on this before, would you mind taking the review?

JulianKniephoff:
@staubesv sure :slightly_smiling_face:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -78,16 +78,24 @@
   public abstract Logger getLogger();
 
   /**
-   * A class to contain the necessary key entries for url signing.
+   * A class representing an url signing key.
```
JulianKniephoff: "**a** url" or preferrably even "a URL", IMO. :wink:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -101,24 +109,32 @@ public void setSecurityService(SecurityService securityService) {
...
-            .findAny();
+  private Key getKey(String baseUrl) {
+    /* Optimization: Use TreeMap.floorEntry that can retrieve the greatest URL equal to or greater than 'baseUrl'
+       in O(log(n)). As we are trying to find an URL that is a prefix of 'baseUrl', candidate.getKey() either is
+       that URL (needs to be checked!) or there is no such URL. */
+    Map.Entry<String, Key> candidate = urls.floorEntry(baseUrl);
+    if (candidate != null && baseUrl.startsWith(candidate.getKey())) {
+      Key key = candidate.getValue();
+
+      // Don't accept URLs without an organization context
```
JulianKniephoff, line 132: I don't think this code belongs into this method, semantically, as it exists right now. **I** would probably pull it out, but I don't actually have a strong opinion, here. We should, however, at least document that behavior in the JavaDoc comment, and maybe give it a bit more of a descriptive name. :thinking:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -101,24 +109,32 @@ public void setSecurityService(SecurityService securityService) {
...
    * @param baseUrl
    *          The url to check against the possible matchers.
-   * @return The {@link KeyEntry} if it is available.
+   * @return The {@link Key} if it is available.
    */
-  private Optional<Map.Entry<String, KeyEntry>> getKeyEntry(String baseUrl) {
-    return keys.entrySet().stream()
-            .filter(entry -> baseUrl.startsWith(entry.getValue().url))
-            .findAny();
+  private Key getKey(String baseUrl) {
```
JulianKniephoff, line 124: I see you don't like `Optional`? Any particular reason?

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -130,62 +146,69 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
+    // Temporary list of key entries to simplify building up the keys
+    Map<String, Key> keys = new HashMap<>();
 
     Enumeration<String> propertyKeys = properties.keys();
     while (propertyKeys.hasMoreElements()) {
       String propertyKey = propertyKeys.nextElement();
 
-      String keyEntryProperty = StringUtils.removeStart(propertyKey, KEY_ENTRY_PREFIX + ".");
-      if (keyEntryProperty == propertyKey) continue;
+      if (!propertyKey.startsWith(KEY_PROPERTY_PREFIX + ".")) {
```
JulianKniephoff: Is this change also for performance reasons? Did you measure it? Because I can hardly imagine this making any meaningful difference. Note that the result of `removeStart` probably shares the data with the original string (strings are immutable), so there should be no copying here. Plus you have the advantage of having to process less of the string in the following code, too.

But of course this loop will probably be executed just once for the lifetime of the program, anyway, so it really doesn't matter, and if you just like this style better, I'm fine with it, I guess. x)

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -130,62 +146,69 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
         case ORGANIZATION:
-          currentKeyEntry.organization = propertyValue;
+          currentKey.organizationId = propertyValue;
           break;
         case URL:
-          if (keys.values().stream()
-                  .map(keyEntry -> keyEntry.url)
-                  .filter(Objects::nonNull)
-                  .anyMatch(url -> propertyValue.startsWith(url) || (url != null && url.startsWith(propertyValue)))) {
+          if (urls.keySet().stream().anyMatch(v -> propertyValue.startsWith(v) || (v.startsWith(propertyValue)))) {
```
JulianKniephoff: You have a `TreeMap` now, shouldn't you be using that here? x)

But see above; this method's performance is probably irrelevant.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Good point! Not looking up the key twice was in my notes for this code as well, but I don't remember why I didn't implement it. Probably just time. But I guess I can scratch that now. :smile:

modules/urlsigning-service-impl/src/test/java/org/opencastproject/security/urlsigning/provider/impl/GenericUrlSigningProviderTest.java
```diff
@@ -283,4 +291,5 @@ public void testMultitenantSign() throws UrlSigningException, ConfigurationExcep
     }
     assertTrue(exceptionThrown);
   }
+
```
JulianKniephoff: Meh, I don't like this. x)

JulianKniephoff (changes requested):
Okay, from just looking at the code, this looks fine, I think. I mostly remarked on some minor nitpicks. None of these is a deal breaker for me, but maybe you can still give some feedback, @staubesv. :wink:

Testing now.

JulianKniephoff:
By the way @staubesv, what about the open question you raised? Will you address this, still? I guess this is out of scope for this PR, since the original code was case sensitive, as well, right?

JulianKniephoff:
Alright, it compiles and seems to work fine, as far as I can tell.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: No, this change is not related to performance but rather about code readability:

```
       String keyEntryProperty = StringUtils.removeStart(propertyKey, KEY_ENTRY_PREFIX + ".");
       if (keyEntryProperty == propertyKey) continue;
      // This configuration property key starts with KEY_ENTRY_PREFIX + "." so we have to handle it
```

just seems a crazy way to say

```
  if (!propertyKey.startsWith(KEY_PROPERTY_PREFIX + ".")) {
      // This configuration property key starts with KEY_ENTRY_PREFIX + "."so we have to handle it
```

to me. It indeed took me a second look to understand what it does and what the superfluous variable _keyEntryProperty_ (there is enough confusion about keys in this code) actually is used for.

[Update] When writing this, I just realised I did: 

```
  if (!propertyKey.startsWith(KEY_PROPERTY_PREFIX + ".")) {
      // This configuration property key does not start with KEY_ENTRY_PREFIX + "."so we don't have to handle it
     continue; // don't handle it
  }
  // handle it
```

which has room for improvement... having a look at this ;-)

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: There is simply no reason why we should pack the answer into a wrapper object.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: Depends on what semantics you assume. Here, the method returns a key that can be used to sign "baseUrl". Since the keys explicitly support organizations, a key that cannot be used by the current organization to sign "baseUrl" is therefore not returned as it cannot be used to sign "baseUrl" in this context.

I've added some information to the comment, though.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: Here, the code seems fine as performance optimizations are not required.

modules/urlsigning-service-impl/src/test/java/org/opencastproject/security/urlsigning/provider/impl/GenericUrlSigningProviderTest.java
staubesv: Me too ;-)

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: Fixed.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: I've changed it to

```
  if (propertyKey.startsWith(KEY_PROPERTY_PREFIX + ".")) {
      // This configuration property key starts with KEY_ENTRY_PREFIX + "."so we have to handle it

```
so that the code matches the semantics.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: I disagree. Expressing the exact type of object a method can return as explicitly as possible is very useful to force consumers to think about all necessary (edge) cases. This method returns `null` sometimes, which every consumer needs to consider. That's the main purpose of things like `Optional` in languages like Java.

Now, maybe it's justified to remove this overhead here, this being a PR about performance and all. Also this is a `private` method, so the effects of being less explicit are localized at least, but I still maintain that `Optional` could help when working on the module itself. Also I think that this might be a bit of a premature optimization.

But okay, I won't block the PR for this. Just wanted to spread some :heart: for type safety.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Everything always depends on the semantics you assume. :wink: But okay, your point is fair, and I can live with a note in the docs. :slightly_smiling_face:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Okay, I take your first point, but (and we had this discussion already) I really think that

```
if (condition) {
  // many many lines of code
  // that handle the thing
  // and introduce multiple further levels of indentation
  // [...]
        }
      }
    }
  }
}
// Where does `if (condition)` end?! WHERE AM I?!! O_O
```

**much** less readable than early bailouts. :thinking:

Again, not gonna insist here, but weren't you the one who was against changing things not directly related to the PR? :wink:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Yeah okay sure. :slightly_smiling_face:

JulianKniephoff (approved):

PR #864, opened 2019-05-06
MH-13500 Multitenancy support for workflows

pmiddend:
This PR adds a new (optional) `organization` field to workflow definitions. The principle behind this is as follows: There can be two workflows with the same id now, but the one corresponding to the user’s organization overrides the ‘global’ one. This pertains workflow dropdowns (for example, the “Add new event” dropdown) as well as workflows included from other workflows via the `include` workflow operation handler.

Note that I took the liberty of fixing a few more stylistics issues in the `WorkflowDefinitionScanner`.

### DIFF
--- a/docs/guides/admin/docs/configuration/workflow.md
+++ b/docs/guides/admin/docs/configuration/workflow.md
@@ -55,6 +55,8 @@ Start by naming the workflow and giving it a meaningful description:
 
       <!-- Description -->
       <id>example</id>
+      <!-- Optionally specify an organization -->
+      <organization>mh_default_org</organization>
       <title>Encode Mp4, Distribute and Publish</title>
       <tags>
         <!-- Tell the UI where to show this workflow -->
@@ -77,7 +79,11 @@ Start by naming the workflow and giving it a meaningful description:
     </definition>
 
 * The `id` is used in several Opencast endpoints to identify and select this workflow. Make sure that this identifier
-  is unique among all endpoints in the system.
+  is unique among all endpoints in the system (except in multitenant workflows, see `organization` below).
+* The `organization` specifies the organization this workflow is valid for (thus, it only makes sense in multitenant
+  installations). If there are two workflows with the same id, the one corresponding to the user’s organization is
+  always chosen. This pertains workflow dropdowns (for example, the “Add new event” dropdown) as well as workflows
+  included in other workflows via the `include` workflow operation handler.
 * The `tags` define where the user interfaces may use these workflows. Useful tags are:
     * *upload*: Usable for uploaded media
     * *schedule*: Usable for scheduled events
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
@@ -155,6 +155,12 @@ public interface WorkflowDefinition extends Comparable<WorkflowDefinition> {
   void clearTags();
 
   /**
+   * Gets the organization associated with this workflow (or <code>null</code>, if it's global)
+   * @return the organization
+   */
+  String getOrganization();
+
+  /**
    * Appends the operation to the workflow definition.
    *
    * @param operation
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
@@ -61,6 +61,9 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
   @XmlElement(name = "title")
   private String title;
 
+  @XmlElement(name = "organization")
+  private String organization;
+
   @XmlElementWrapper(name = "tags")
   @XmlElement(name = "tag")
   protected SortedSet<String> tags = new TreeSet<String>();
@@ -347,12 +350,20 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
     this.title = title;
   }
 
+  @Override
+  public String getOrganization() {
+    return organization;
+  }
+
   /**
    * {@inheritDoc}
    *
    * @see java.lang.Object#toString()
    */
   public String toString() {
+    if (organization != null) {
+      return "Workflow definition {" + id + "/" + organization + "}";
+    }
     return "Workflow definition {" + id + "}";
   }
 
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIdentifier.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIdentifier.java
@@ -1,0 +1,75 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.opencastproject.workflow.api;
+
+import java.util.Objects;
+
+/**
+ * Represents a workflow's "primary key", for use in maps and sets (immutable)
+ */
+public final class WorkflowIdentifier {
+  private final String id;
+  private final String organization;
+
+  public WorkflowIdentifier(String id, String organization) {
+    this.id = id;
+    this.organization = organization;
+  }
+
+  public String getId() {
+    return id;
+  }
+
+  public String getOrganization() {
+    return organization;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o)
+      return true;
+    if (o == null || getClass() != o.getClass())
+      return false;
+    WorkflowIdentifier that = (WorkflowIdentifier) o;
+    return id.equals(that.id) && Objects.equals(organization, that.organization);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(id, organization);
+  }
+
+  public String toString() {
+    if (organization == null) {
+      return id;
+    }
+    return String.format("%s/%s", organization, id);
+  }
+
+  /**
+   * Return a new identifier without an organization (idempotent)
+   * @return See above
+   */
+  public WorkflowIdentifier withoutOrganization() {
+    return new WorkflowIdentifier(id, null);
+  }
+}
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
@@ -23,12 +23,14 @@ package org.opencastproject.workflow.impl;
 
 import static org.opencastproject.util.ReadinessIndicator.ARTIFACT;
 
+import org.opencastproject.security.api.Organization;
+import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.util.ReadinessIndicator;
 import org.opencastproject.workflow.api.WorkflowDefinition;
+import org.opencastproject.workflow.api.WorkflowIdentifier;
 import org.opencastproject.workflow.api.WorkflowParser;
 import org.opencastproject.workflow.api.WorkflowStateMapping;
 
-import org.apache.commons.io.IOUtils;
 import org.apache.felix.fileinstall.ArtifactInstaller;
 import org.osgi.framework.BundleContext;
 import org.slf4j.Logger;
@@ -36,7 +38,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FilenameFilter;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Dictionary;
@@ -45,6 +46,7 @@ import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 /**
  * Loads, unloads, and reloads {@link WorkflowDefinition}s from "*workflow.xml" files in any of fileinstall's watch
@@ -54,25 +56,31 @@ public class WorkflowDefinitionScanner implements ArtifactInstaller {
   private static final Logger logger = LoggerFactory.getLogger(WorkflowDefinitionScanner.class);
 
   /** An internal collection of workflows that we have installed */
-  protected Map<String, WorkflowDefinition> installedWorkflows = new HashMap<String, WorkflowDefinition>();
+  protected Map<WorkflowIdentifier, WorkflowDefinition> installedWorkflows = new HashMap<>();
 
   /** All workflow state mappings which are configured for the workflow defintions */
   protected Map<String, Set<WorkflowStateMapping>> workflowStateMappings = new HashMap<>();
 
   /** An internal collection of artifact id, bind the workflow definition files and their id */
-  protected Map<File, String> artifactIds = new HashMap<File, String>();
+  protected Map<File, WorkflowIdentifier> artifactIds = new HashMap<>();
 
   /** List of artifact parsed with error */
-  protected List<File> artifactsWithError = new ArrayList<File>();
+  protected List<File> artifactsWithError = new ArrayList<>();
 
   /** OSGi bundle context */
   private BundleContext bundleCtx = null;
 
   /** Tag to define if the the workflows definition have already been loaded */
-  private boolean isWFSinitiliazed = false;
+  private boolean isWFSinitialized = false;
 
   /** The current workflow definition being installed */
   private WorkflowDefinition currentWFD = null;
+
+  private OrganizationDirectoryService organizationDirectoryService;
+
+  public void setOrganizationDirectoryService(OrganizationDirectoryService organizationDirectoryService) {
+    this.organizationDirectoryService = organizationDirectoryService;
+  }
 
   /**
    * OSGi callback on component activation. private boolean initialized = true;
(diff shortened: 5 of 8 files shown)


lkiesow:
How does this effect default workflows and workflows selected at other places like the ingest service?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
```diff
@@ -738,18 +739,21 @@ public Response registerWorkflowDefinition(@FormParam("workflowDefinition") Work
   }
 
   @DELETE
-  @Path("/definition/{id}")
```
staubesv: Hmm.... that breaks API compatibility to existing clients... I see the point that the ID alone is not unique anymore which works out well for almost anycase except this one ;-)

Ideas for a backwards-compatible change?

pmiddend:
@lkiesow Good question. I'm not sure :D If the new field isn't specified, everything should work just like before. I think the key method that hasn't changed, at least to the "outside world", is `getWorkflowDefinitionById`. This method still receives a `String` and tries to find the workflow. First using the organization, then without it.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
pmiddend: Hm, but this only breaks the internal REST API, doesn't it? The one from the remote to the implementation. Am I missing something? Are there people who update their remotes, but not their core, or vice-versa?

staubesv:
@lkiesow & @pmiddend I would expect the same lookup behaviour everywhere, so looking for a workflow definition with the identifier "test-id" should work like:

1.  If a workflow definition with the identifier "test-id" exists for the current organization, return it
2. If step 1 does not return a workflow definition, return the workflow definition with the identifier "test-id" without any organization specified
3. If step 2 does not return a workflow definition, return none

staubesv:
@pmiddend I think the method "start" also needs to be adapted to ensure that a workflow configured for tenant "abc" can only be started from tenant "abc".

Unclear to me how to property handle deletion of workflow definitions, though.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
staubesv: The world of APIs in Opencast currently looks like:

1. The Admin UI Facade is an internal API
2. The External API is an external API
3. All other APIs are neither nor ;-)

staubesv: Maybe the following approach would make sense: DELETE /definition/{id} keeps the signature but does the following: If caller has ROLE_ADMIN, unregister all workflow definitions (can be 0, 1 or 2) that have the identifier "id". If not, just unregister the workflow definition (can be 0 or 1) with "id" and the organization set to the current organization.

Note that ROLE_ADMIN is explicitly not including the tenant administration role.

Although all that needs to work correctly when a system administrator changes the XMLs on disk... in that case, removing all workflow definitions would not be correct...

lkiesow: > If caller has ROLE_ADMIN, unregister all workflow definitions (can be 0, 1 or 2) that have the identifier "id". If not, just unregister the workflow definition (can be 0 or 1) with "id" and the organization set to the current organization.

Please don't. That's extremely confusing. Especially since the global admin would still access the endpoint as part of an organization.

staubesv: @lkiesow That was just an idea ;-)

With reading, there is no problem with this model because there is always exactly none or one workflow definition with a given WDID from point of view of an organization: Either none, the one specific to the organization OR the one available for all organizations.

When updating workflow definitions, however, the question is which one to take and whether an organization is allowed to update workflow definitions that are also used by other organizations. So there are two endpoints to be considered: DELETE WD & PUT WD.

As for PUT WD, I wonder whether this makes sense at all (as the change is not made persistent anyway and will therefore be lost when restarting Opencast). Actually, I wonder how much sense DELETE makes for the same reason.

Writing this, I'm tempted to propose the removal of those two REST functions as they actually don't make much sense without persistency which we don't have (except the files, but when thinking REST, usually DBs are more reasonable then files).

Removing those functions would result in a sound solution as all other situations work well.

What do you think? Ideas?

lkiesow: > With reading, there is no problem with this model because there is always exactly none or one workflow definition with a given WDID from point of view of an organization: Either none, the one specific to the organization OR the one available for all organizations.

That's still a very weird behavior, isn't it? As a user, if I delete something, I expect it to be deleted and not replaced by another version.

If we can get rid of all the endpoints, that's, of course, a solution :D

staubesv: You cannot delete workflow definitions as a user - they can only be configured by an admin having access to Opencast's configuration files.
While there are two functions to delete or update workflow definitions using a REST API, those don't seem to make much sense in the current implementation (lack of persistence) and I personally would not aim at adding persistence as the configuration files are working just fine for at least us.

As for the configuration files, the motivation to allow "overrides" of the global workflow configuration is the assumption that most of the configuration is the same for all tenants but a provider might want to tweak individual tenants. Of course, we could just say each workflow definition has exactly one organization, that would - in our case - result in approx. 1600 workflow definition files where most of them are the same (except the organization). While Ansible can create them based on templates, I still would prefer to not see so many files in etc/workflows ...

lkiesow: admins are still users ;-P

But I'm not arguing against having a tenant workflow definition. It's just that if you delete a workflow for an organization it would be very weird if it's not deleted but replaced.

But that's no problem anymore if you remove these endpoints.

staubesv: Yep, absolutely, but as an admin I would rather adjust the configuration in etc/workflows than use a REST endpoint ;-)

I'll check the state of the current implementation of the DELETE/PUT requests and bring the topic up at the tomorrows technical meeting.

staubesv:
@pmiddend Since #872 is merged, can you please adjust this PR and resolve the merge conflicts?

staubesv:
@mliradelc IIRC, Ruth has this on her wish list... would you be willing to take the review?

docs/guides/admin/docs/configuration/workflow.md
```diff
@@ -55,6 +55,8 @@ Start by naming the workflow and giving it a meaningful description:
 
       <!-- Description -->
       <id>example</id>
+      <!-- optionally specify an organization -->
```
KatrinIhler: Optionally

docs/guides/admin/docs/configuration/workflow.md
```diff
@@ -77,7 +79,11 @@ Start by naming the workflow and giving it a meaningful description:
     </definition>
 
 * The `id` is used in several Opencast endpoints to identify and select this workflow. Make sure that this identifier
-  is unique among all endpoints in the system.
+  is unique among all endpoints in the system (except in multitenant workflows, see `organization` below).
+* The `organization` specifies the organization this workflow is valid for (as such, it only makes sense in multitenant
```
KatrinIhler: "thus" instead of "as such"?

docs/guides/admin/docs/configuration/workflow.md
```diff
@@ -77,7 +79,11 @@ Start by naming the workflow and giving it a meaningful description:
...
+  installations). If there are two workflows with the same id, the one corresponding to the user’s organization is
+  always chosen. This pertains workflow dropdowns (for example, the “Add new event” dropdown) as well as workflows
+  included from other workflows via the `include` workflow operation handler.
```
KatrinIhler: included in?

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
```diff
@@ -333,13 +336,23 @@ public void setTitle(String title) {
...
+  }
+
   /**
    * {@inheritDoc}
    *
    * @see java.lang.Object#toString()
    */
   public String toString() {
-    return "Workflow definition {" + id + "}";
+    return "Workflow definition {" + id + "/" + organization + "}";
```
KatrinIhler: Maybe make this dependent on whether there is an organization?

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIdentifier.java
```diff
@@ -0,0 +1,93 @@
...
+public final class WorkflowIdentifier {
+  private final String id;
+  private final String organization;
+
+  /**
+   * Create an identifier directly from a workflow definition
+   * @param definition the workflow definition
+   * @return the constructed identifier
+   */
+  public static WorkflowIdentifier fromDefinition(final WorkflowDefinition definition) {
```
KatrinIhler: Why aren't these simply constructors?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -887,9 +900,14 @@ else if (!FAILED.equals(workflow.getState())) {
 
   @Override
   public WorkflowDefinition getWorkflowDefinitionById(String id) throws NotFoundException {
-    WorkflowDefinition def = workflowDefinitionScanner.getWorkflowDefinition(id);
-    if (def == null)
-      throw new NotFoundException("Workflow definition '" + id + "' not found");
+    final WorkflowIdentifier workflowIdentifier = new WorkflowIdentifier(id, securityService.getOrganization().getId());
+    WorkflowDefinition def = workflowDefinitionScanner.getWorkflowDefinition(workflowIdentifier);
+    if (def == null) {
+      def = workflowDefinitionScanner.getWorkflowDefinition(workflowIdentifier.withoutOrganization());
```
KatrinIhler: Can't getWorkflowDefinition() simply fall back to the global workflow definition of there isn't one for this specific organization instead of us having to ask twice?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -357,12 +359,23 @@ protected void fireListeners(final WorkflowInstance oldWorkflowInstance, final W
...
+              final WorkflowDefinition definition = workflowDefinitionScanner
+                      .getWorkflowDefinition(new WorkflowIdentifier(x, organization));
+              if (definition != null) {
+                return definition;
+              }
+              return Objects.requireNonNull(
+                      workflowDefinitionScanner.getWorkflowDefinition(WorkflowIdentifier.withoutOrganization(x)));
+            })
+            .sorted()
+            .collect(Collectors.toList());
```
KatrinIhler: This is quite complicated. So we first build workflowIdentifiers based on the workflowDefinitions and then ask the workflowDefinitionScanner for each WorkflowDefinition seperately and sometimes twice? Maybe we should integrate this functionality (something like getWorkflowDefinitions(Organization org)) into the workflowDefinitionScanner to make it simpler.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
```diff
@@ -103,22 +103,22 @@ public void install(File artifact) throws Exception {
...
-        return name.endsWith(".xml");
-      }
-    });
+    String[] filesInDirectory = artifact.getParentFile().list((arg0, name) -> name.endsWith(".xml"));
+    if (filesInDirectory == null) {
+      throw new RuntimeException("error retrieving files from directory \"" + artifact.getParentFile() + "\"");
+    }
 
-    logger.info("Worfkflow definition '{}' from file '{}' installed", def.getId(), artifact.getName());
+    logger.info("Workflow definition '{}' from file '{}' installed", workflowIdentifier, artifact.getName());
```
KatrinIhler, line 138: And I had gotten so used to the worfkflows.. ;)

staubesv:
@pmiddend Please note that this PR currently has a merge conflict.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIdentifier.java
pmiddend: It's simply a matter of style, of course, but I find multiple overloaded constructors

1. hard to read
2. don't scale

Let me illustrate the first point:

```
new WorkflowDefinition("foo", null)
```

Reading this, do you know what the `null` is for?

As for the second point, there are lots of examples in Opencast, but you might take a look at `SchedulerItem`. As you can see, not only has this class many constructors, but they partly overlap. One takes a `String`, one takes `String` and a `long`, which isn't even used. This is done (generally speaking) so that the compiler can figure out which constructor you mean.

In order to combat this, I invent this named, static "constructors".

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
pmiddend: It's a valid point. However, no matter what you do, you're going to have code that checks twice. Currently, this code is in `WorkflowServiceImpl`, making `WorkflowDefinitionScanner` dumb. We could make `WorkflowDefinitionScanner` smarter and put the check in there, but then we'd lose the ability to get the definition without the fallback mechanism.
I vote for keeping it the way it is.

Edit: I spoke to fast, it seems. In combination with the restructuring below, such a method makes much more sense. Be prepared for a cleanup commit :)

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
pmiddend: While the complexity is somewhat inherent because of the fallback mechanism, I was able to come up with a better solution.

pmiddend:
@KatrinIhler Thanks for the review, I've pushed some changes.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIdentifier.java
KatrinIhler: Okay, I get what you're saying, but I still think this is relatively straight-forward in this case: One constructor that takes both id and organization, one that only takes the id, one that takes a workflow definition. So no need for WorkflowIdentifier("id", null), except internally. These are all easily distinguishable by the compiler.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
```diff
@@ -169,61 +170,54 @@ public void update(File artifact) throws Exception {
...
 
-  /**
-   * Gets the workflow definitions with the given id.
-   *
-   * @param id
-   * @return the workflow definition if exist or null
-   */
-  public WorkflowDefinition getWorkflowDefinition(String id) {
-    return installedWorkflows.get(id);
+  public Stream<WorkflowDefinition> getAvailableWorkflowDefinitionsForOrganization(final Organization organization) {
```
KatrinIhler: I vote for leaving "forOrganization" out of the method name (and re-add some javadoc instead) since this is currently the only way to get workflow definitions anyway and otherwise it's terribly long...

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
```diff
@@ -169,61 +170,54 @@ public void update(File artifact) throws Exception {
...
   }
 
-  /**
-   * Get the list of installed workflow definitions.
-   *
-   * @return the collection of installed workflow definitions id
-   */
-  public Map<String, WorkflowDefinition> getWorkflowDefinitions() {
-    return installedWorkflows;
+  public WorkflowDefinition getWorkflowDefinitionOrDefault(WorkflowIdentifier workflowIdentifier) {
```
KatrinIhler: Same here - leave "orDefault". I think that's confusing anyway, it's not like we have one default workflow definition... and it's currently the only way to get a workflow definition.
Also Javadoc!

KatrinIhler (changes requested):
So I tried this out and ran into some weird behavior when adding new workflows to a running opencast (which should imo be possible).
So when I copy a global workflow definition and then change the copy to a tenant-specific workflow, this workflow is available, but the global workflow disappears, and only reappears if I edit that workflow definition or I restart. My guess is that the copy becomes the new global workflow and after changing it to a tenant-specific one, the global workflow is uninstalled (since there's some delay between copying and editing the copy). Imo opencast should not install another global workflow with the same id, but from the log it appears to do that.

KatrinIhler (changes requested):
I've tested this and encountered some weird behaviour when adding workflows while opencast is running. This is what I do:
- Copy global workflow
- Change copy to be tenant specific
Result: Tenant-specific workflow is available, but global definition is gone until I edit the global workflow definition or restart.
My guess is that between copying and editing the copy, opencast accepts the copy as the new global workflow (which it shouldn't, since there already is one) and then uninstalls it after the editing.

pmiddend:
@KatrinIhler I've fixed the code according to your suggestions and added a new check for duplicate workflow definitions.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
```diff
@@ -154,6 +154,18 @@
    */
   void clearTags();
 
+  /**
+   * Sets the organization associated with this workflow (or <code>null</code>, if it's global)
+   * @param organization the organization
+   */
+  void setOrganization(String organization);
```
KatrinIhler: Shouldn't this be removed as per the discussion in #874?

KatrinIhler (commented):
Alright, final test. ;) Looks good now except for two minor things:
- Defining more than one organization should probably lead to the workflow definition not being accepted.
- The same should be true if the tenant doesn't exist.

pmiddend:
@KatrinIhler Hmmm, not sure how you would define multiple organizations? Sure you could repeat the XML tag, but the XML isn't really checked via a schema, or is it? Maybe I simply misunderstood.

KatrinIhler:
You're right, though it'd be nice if it would be...as it is, the first organization is accepted. But I guess this is out of the scope of this PR.

pmiddend:
Okay, another round of fixes. I also renamed a private variable that had a typo in it, while I was at it.

KatrinIhler (approved):

PR #865, opened 2019-05-06
MH-13524 Empty node name causes exception

staubesv:
Steps to reproduce: 
1. Ensure oc_host_registration has node_name = NULL 
2. Go to System->Servers 
  
 Actual Results: 
The table does not show any results. In the Opencast log, you find: 

Caused by: java.lang.IllegalArgumentException: Value of field 'nodeName' must not be null 
at com.entwinemedia.fn.data.json.Field.<init>(Field.java:31) ~[?:?] 
at com.entwinemedia.fn.data.json.Jsons.f(Jsons.java:89) ~[?:?] 
at org.opencastproject.adminui.endpoint.ServerEndpoint.getServersListAsJson(ServerEndpoint.java:345) ~[?:?] 
at org.opencastproject.adminui.endpoint.ServerEndpoint.getServers(ServerEndpoint.java:317) ~[?:?] 
  
 Expected Results: 
The Admin UI should still display the servers even if the backend does not deliver a valid nodename

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ServerEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ServerEndpoint.java
@@ -21,6 +21,7 @@
 
 package org.opencastproject.adminui.endpoint;
 
+import static com.entwinemedia.fn.data.json.Jsons.BLANK;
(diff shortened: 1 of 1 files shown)


staubesv:
@JamesUoM I've missed this when reviewing #850. Would you mind taking the review?

PR #866, opened 2019-05-06
MH-13523 Allow removal of locations anytime

staubesv:
Currently, the Admin UI will not show the "Remove location" button in the table Capture->Locations unless the capture agent is in status OFFLINE or UNKNOWN.

At least at our side, this is annoying organization administrators that expect to be able to remove a capture agent at any time they want.

As removing the capture agent doesn't actually do much (the capture agent just register itself again in case it is still there), there does not seem to be much benefit in preventing users from removing a location in case it is not OFFLINE or UNKNOWN.
Also, the backend would not prevent anybody from removing the location - it is just the front-end that does not show the button.

The goal of this task is to allow uses to always remove locations if they want to.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/recordings/partials/recordingActionsCell.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/recordings/partials/recordingActionsCell.html
@@ -7,7 +7,6 @@
 <a data-confirmation-modal="confirm-modal"
    data-callback="table.delete"
    data-object="row"
-   ng-if="row.removable"
    class="remove"
    title="{{ 'RECORDINGS.RECORDINGS.TABLE.TOOLTIP.DELETE' | translate }}"
    with-role="ROLE_UI_LOCATIONS_DELETE">
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/captureAgentsResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/captureAgentsResource.js
@@ -41,7 +41,6 @@ angular.module('adminNg.resources')
           row.inputs = r.inputs;
           row.roomId = r.roomId;
           row.type = 'LOCATION';
-          row.removable = 'AGENTS.STATUS.OFFLINE' === r.Status || 'AGENTS.STATUS.UNKNOWN' === r.Status;
           return row;
         };
 


smarquard:
It reduces UI clutter only to show the X if the CA is not online. Is there a point to removing a CA that's online? It will just reappear again.

lkiesow:
No strong feelings here but bear in mind that this change could lead to weird situations in which a user removes a capture agent which will send the regular ping around that time which will register the agent anew so that for the user, it looks like the delete option is broken.

staubesv:
@smarquard & @lkiesow We originally did not follow the community version and always allowed users to remove locations at any time. We synced to the community version behaviour end of January this year and now have first complaints about people being unhappy with this "new" behaviour.
Likely people do things like renaming a CA and are then annoyed having to wait until Opencast detects that the CA is offline - I will ask them to see what motivates them to want the old behaviour back...

smarquard:
It could be annoying to wait as the default is 2 hours:

````
# Timeout for capture agent status, in minutes.
# Capture agents which have not sent status updates for this period will be marked as offline.
# Default: 120 minutes (2 hours)
````

though locally we have this configured to 3 minutes which is appropriate for CAs that send status updates regularly. I think originally it was 2 hours as a compromise because pyca at that time didn't send status updates during a recording as it was single-threaded.

smarquard:
Another UI option to reduce clutter would have the X show always for offline/unknown CAs, and show only on mouse-over of that row for online CAs.

lkiesow:
> Another UI option to reduce clutter would have the X show always for offline/unknown CAs, and show only on mouse-over of that row for online CAs.

Please let us not go that way. I think that's more confusing than always having the button. At least, unless we don't do that for all action icons.

smarquard:
Revealing actions on mouseover is quite a common UI pattern nowadays. For example, Outlook web UI:

Screenshot 2019-05-06 at 20 33 57

lkiesow:
Sorry

> At least, unless we don't do that for all action icons.

should have been:

> At least, unless we <strike>don't</strike> do that for all action icons.

I'm aware that it's used and I'm fine with going that route but not only for this single button because that's something no one would expect.

staubesv:
I think that lowering the "time to offline" will work in our case.

staubesv:
@lkiesow & @smarquard 

> Revealing actions on mouseover is quite a common UI pattern nowadays.

Yes, but IMO only makes sense on small surfaces where mouseover implies that the mouse cursor is near to where the controls are revealed as in the Outlook example.
In the Admin UI, I would consider it a bit "weird" if a mouseover on the left side of the screen reveals controls on the right side of the screen as the "distance" is too large so I cannot actually reach the controls (without moving the mouse cursor across the entire screen). Not revealing the controls on row mouseover would somewhat indicate no controls available for the row...

KatrinIhler:
So, is this even necessary anymore? Sounds like lowering the time to offline is a reasonable thing to do (and maybe even set that default a bit lower?)

staubesv:
@KatrinIhler Yep, lowering the time will do. I close this PR. Thanks for the reminder!

PR #872, opened 2019-05-06
MH-13530 Remove pseudo-mechanism for workflow definition registration

staubesv:
Opencast workflow definitions are stored as XML files in etc/workflows.

There is an additional mechanism that suggests that the workflow configuration can be managed used the REST endpoint of the workflow service that offers those two functions:

- DELETE /workflow/definition/{id} to unregister a workflow definition
- PUT /workflow/definition to add a workflow definition

While those functions are implemented, they aren't actually useful as there is no persistence layer that would in any way ensure that the changes survive a restart of Opencast where Opencast would simply start over with the configuration found in etc/workflows.

Instead of using those functions, administrators can delete the workflow definition XML file in etc/workflows to unregister a workflow definition or add a file to add a workflow definition.

### DIFF
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -1,105 +1,31 @@
-Opencast 7: Release Notes
+Opencast 8: Release Notes
 =========================
 
 New Features
 ------------
 
-- Overhaul of the permission management with the newly added possibility to define how access control lists are
-  evaluated and how series permission changes are populated to episodes. For more details take a look at the [access
-  control configuration guide](configuration/acl.md).
-- Update Elasticsearch and make is possible to run Elasticsearch as an external service.
-- Per-Tenant Capture Agent Users
-- Asset manager snaphots clean-up to remove older snapshots for a given media package. In some cases, this can
-  drastically reduce Opencast's storage consumption. This feature is implemented as an option for the [asset-delete workflow
-  operation handler](workflowoperationhandlers/asset-delete-woh.md).
-- Allow the workflow to select the audio track for composite video
-- Add multi-tenant support for all list providers
-- Make waveform size configurable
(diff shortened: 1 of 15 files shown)


KatrinIhler:
I'm usually all for removing unused functionality, but I'm not sure I agree with this one. It's true that the changes don't survive a system restart, but I've used this quite a lot when testing on remote servers because I found it easier than dealing with the filesystem, file permissions and so on, just to throw a quick testing workflow in.

staubesv:
@KatrinIhler This work is motivated by #864 which aims at making it possible to have workflow definitions that are valid for either all tenants or just a specific one. That idea works perfectly fine with the file-based workflow configuration but has some smells when starting to allow API access.
Looking at the API code it seems to be useful for testing purposes at most. The motivation to remove it is that this is not real functionality (except testing - for anything else there is no way around configuring the workflows per file) and removing it yields are more concise design without some weird smells.

An alternative to removing the functionality is to adapt the interface to support organizations, although that does make much sense (e.g. an organization should not be allowed to remove workflow definitions of other organizations or all organizations). On option would be to explicitly mark the methods as "for testing only" so that people at least realize that the methods don't make much sense except being convenient and should be deactivated in production systems.

staubesv:
@KatrinIhler ...one cigarette later: Actually, the convenience argument (no need to access files on the server) is quite weak: If you want this, I would recommend you to write an endpoint that just allows you to put files into the etc folder. That way, you can also change any other configuration (e.g. encoding profiles, list providers, ...) that is automatically loaded by Opencast which is even more convenient ;-)

lkiesow:
@staubesv, this has conflicts

staubesv:
@lkiesow Merge conflicts fixed.

lkiesow (approved):

PR #873, opened 2019-05-07
MH-13531 - upgrade spring-security and jasig cas library to fix issue…

kristofkeppens:
… with https and reverse proxies

See https://opencast.jira.com/browse/MH-13531 for more explanation.


lkiesow:
If no one else wants's to take the review, I'll test it as best as I can until tomorrow without having an actual CAS server.

lkiesow:
Seems to work as far as I can tell → merged

PR #874, opened 2019-05-07
MH-13532 Role support for workflows

pmiddend:
This PR adds the ability to define roles for workflows, as such:

- A workflow definition can specify a list of roles.
- If the workflow does not specify roles, the previous behavior is in effect: everyone can use this workflow.
- Users with `ROLE_ADMIN` can also use all workflows.
- Otherwise, if the current user has roles in common with the roles specified in the workflow, the workflow can be started.

Note that the PR depends on #864, as it was easier to implement this way.

### DIFF
--- a/docs/guides/admin/docs/configuration/workflow.md
+++ b/docs/guides/admin/docs/configuration/workflow.md
@@ -57,6 +57,10 @@ Start by naming the workflow and giving it a meaningful description:
       <id>example</id>
       <!-- Optionally specify an organization -->
       <organization>mh_default_org</organization>
+      <!-- optionally specify roles for this workflow -->
+      <roles>
+        <role>ROLE_ADMIN</role>
+      </roles>
       <title>Encode Mp4, Distribute and Publish</title>
       <tags>
         <!-- Tell the UI where to show this workflow -->
@@ -84,6 +88,10 @@ Start by naming the workflow and giving it a meaningful description:
   installations). If there are two workflows with the same id, the one corresponding to the user’s organization is
   always chosen. This pertains workflow dropdowns (for example, the “Add new event” dropdown) as well as workflows
   included in other workflows via the `include` workflow operation handler.
+* The `roles` define which user roles are allowed to see and start this workflow (a user needs one of the roles provided
+  in the definition). If this is omitted or no roles are specified, everyone can see and start the workflow (provided
+  the `organization` constraints are satisfied). Also, users with `ROLE_ADMIN` can see and start every workflow. Note
+  that the workflows included in Opencast do not set roles.
 * The `tags` define where the user interfaces may use these workflows. Useful tags are:
     * *upload*: Usable for uploaded media
     * *schedule*: Usable for scheduled events
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
@@ -150,6 +150,13 @@ public interface WorkflowDefinition extends Comparable<WorkflowDefinition> {
   String[] getTags();
 
   /**
+   * Returns the roles for this workflow definition
+   *
+   * @return the tags
+   */
+  Collection<String> getRoles();
+
+  /**
    * Removes all tags associated with this workflow definition
    */
   void clearTags();
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -66,7 +67,11 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
 
   @XmlElementWrapper(name = "tags")
   @XmlElement(name = "tag")
-  protected SortedSet<String> tags = new TreeSet<String>();
+  protected SortedSet<String> tags = new TreeSet<>();
+
+  @XmlElementWrapper(name = "roles")
+  @XmlElement(name = "role")
+  protected SortedSet<String> roles = new TreeSet<>();
 
   @XmlElement(name = "description")
   private String description;
@@ -165,7 +170,7 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
    */
   public List<WorkflowOperationDefinition> getOperations() {
     if (operations == null)
-      operations = new ArrayList<WorkflowOperationDefinition>();
+      operations = new ArrayList<>();
     return operations;
   }
 
@@ -185,7 +190,7 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
   @Override
   public WorkflowOperationDefinition get(int position) throws IndexOutOfBoundsException {
     if (operations == null)
-      operations = new ArrayList<WorkflowOperationDefinition>();
+      operations = new ArrayList<>();
     if (position < 0 || position >= operations.size())
       throw new IndexOutOfBoundsException();
     return operations.get(position);
@@ -199,7 +204,7 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
   @Override
   public void add(WorkflowOperationDefinition operation) {
     if (operations == null)
-      operations = new ArrayList<WorkflowOperationDefinition>();
+      operations = new ArrayList<>();
     add(operation, this.operations.size());
   }
 
@@ -212,7 +217,7 @@ public class WorkflowDefinitionImpl implements WorkflowDefinition {
   @Override
   public void add(WorkflowOperationDefinition operation, int position) {
     if (operations == null)
-      operations = new ArrayList<WorkflowOperationDefinition>();
+      operations = new ArrayList<>();
 
     if (operation == null)
       throw new IllegalArgumentException("Workflow operation cannot be null");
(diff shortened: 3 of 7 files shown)


modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -949,11 +980,34 @@ else if (!FAILED.equals(workflow.getState())) {
     return processingOperation;
   }
 
+  private boolean userCanAccessWorkflow(WorkflowDefinition wd) {
+    if (wd.getRoles().isEmpty() || securityService.getUser().getRoles().stream()
```
staubesv: securityService.getUser().hasRole(GLOBAL_ADMIN_ROLE)

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -949,11 +980,34 @@ else if (!FAILED.equals(workflow.getState())) {
...
+            .anyMatch(role -> role.getName().equals(GLOBAL_ADMIN_ROLE))) {
+      return true;
+    }
```
staubesv: Might be a matter of taste, but I would prefer...

` } else {`

... over ...

```
  return true;
}
... the else case
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -949,11 +980,34 @@ else if (!FAILED.equals(workflow.getState())) {
...
+    final Set<Role> userRoles = securityService.getUser().getRoles();
+    final Set<Role> workflowRoles = wd.getRoles().stream()
+            .map(roleStr -> roleDirectoryService.getRolesStream().filter(role -> role.getName().equals(roleStr))
```
staubesv: What is the role for the RoleDirectoryService here? At the end of the day, the only thing this code needs to do is to check whether usersRoles contains at least one of the roles in wd.getRoles()...?

That is something like:

```
User user = securityService.getUser();
for (String Role : wd.getRoles()) {
  if (user.hasRole(role.getName())) {
    return true;
  }
}
return false;
```

pmiddend: Yeah, that's actually true. I just thought I maybe shouldn't rely on string comparisons but instead try to actually get an existing `Rule` object. If that's not necessary, great, I'll simplify the code. :)

modules/workflow-service-impl/src/main/resources/OSGI-INF/workflow-service.xml
```diff
@@ -39,4 +39,6 @@
              cardinality="1..1" policy="static" bind="setMessageReceiver"/>
   <reference name="message-broker-sender" interface="org.opencastproject.message.broker.api.MessageSender"
              cardinality="1..1" policy="static" bind="setMessageSender"/>
+  <reference name="roleDirectoryService" interface="org.opencastproject.security.api.RoleDirectoryService"
```
staubesv: The code does not depend on the RoleDirectoryService anymore, does it?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2439,4 +2484,7 @@ public String getSystemUserName() {
     return SecurityUtil.getSystemUserName(componentContext);
   }
 
+  public void setRoleDirectoryService(RoleDirectoryService roleDirectoryService) {
```
staubesv: The code does not depend on the RoleDirectoryService anymore, does it?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -62,6 +62,7 @@
 import org.opencastproject.security.api.Organization;
 import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.Permissions;
+import org.opencastproject.security.api.RoleDirectoryService;
```
staubesv: The code does not depend on the RoleDirectoryService anymore, does it?

gregorydlogan:
Tagging this as onhold since it depends on #864 which is also currently on hold.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/ToolsEndpoint.java
```diff
@@ -687,6 +688,9 @@ public Response editVideo(@PathParam("mediapackageid") final String mediaPackage
         } catch (NotFoundException e) {
           logger.warn("Workflow '{}' not found", workflowId);
           return R.badRequest("Workflow not found");
+        } catch (UnauthorizedException e) {
+          logger.warn("Workflow '{}' not accessible", workflowId);
```
KatrinIhler: I don't think we need to log this here, a user could flood the log otherwise by trying to access a workflow they're not authorized for.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
```diff
@@ -143,11 +143,30 @@
...
   /**
    * Removes all tags associated with this workflow definition
    */
   void clearTags();
 
+  /**
+   * Sets the organization associated with this workflow (or <code>null</code>, if it's global)
+   * @param organization the organization
+   */
+  void setOrganization(String organization);
```
KatrinIhler: Why do we need a set function for this but not for the roles?

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
```diff
@@ -269,7 +277,7 @@ public boolean containsTag(Collection<String> tags) {
    */
   @Override
   public String[] getTags() {
-    return tags.toArray(new String[tags.size()]);
+    return tags.toArray(new String[0]);
```
KatrinIhler, line 294: Why?

staubesv:
Hi @pmiddend, please note that this PR currently has merge conflicts.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
pmiddend: I sometimes let IntelliJ Idea's inspections loose on the Opencast source code. Here's the help text for this inspection:

> Call to 'toArray()' with pre-sized array argument 'new String[tags.size()]'
> Inspection info: There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
> This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

pmiddend:
@staubesv Thanks, fixed.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinition.java
pmiddend: Seems we don't. I have removed it.

staubesv:
@KatrinIhler I've removed the label "on hold" but don't forget about #864 needs to be finished first ;-)

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowDefinitionImpl.java
KatrinIhler: Ah, I see! You learn something new every day! ;)

KatrinIhler (commented):
I tested this successfully. Just two things:
- Maybe mention in the documentation that it doesn't work for included workflows?
- Maybe fail to load the workflow if a role doesn't exist?

staubesv:
Hi @KatrinIhler 

I don't think it matters whether the role does exist or not: If a user has the role required to access a workflow, he can access it. If not, he can't. 
If no user has a given role, the workflow can simply not be accessed using that given role. That might change when the "missing" role is added. This behaviour is consistent with the way Opencast deals with roles (e.g. ACLs) and seems to make sense.

Agree that it makes sense to mention in the documentation that roles of included workflows are ignored.

KatrinIhler:
@staubesv Alright, we'll leave it like this then.

@pmiddend Conflicts!

pmiddend:
@staubesv @KatrinIhler Fixed.

docs/guides/admin/docs/configuration/workflow.md
```diff
@@ -84,6 +88,10 @@ Start by naming the workflow and giving it a meaningful description:
   installations). If there are two workflows with the same id, the one corresponding to the user’s organization is
   always chosen. This pertains workflow dropdowns (for example, the “Add new event” dropdown) as well as workflows
   included in other workflows via the `include` workflow operation handler.
+* The `roles` define which user roles are allowed to see and start this workflow. If this is omitted or no roles are
+  specified, everyone can see and start the workflow (provided the `organization` constraints are satisfied). Also,
+  users with `ROLE_ADMIN` can see and start every workflow. Note that the workflows included in Opencast do not set
+  roles.
```
KatrinIhler: Maybe add that any user that has _one_ of the specified roles can start the workflow instead of needing to have _all_ the listed roles?

KatrinIhler:
Looks good otherwise, will merge as soon as that last comment is addressed.

docs/guides/admin/docs/configuration/workflow.md
pmiddend: Good point, fixed!

PR #876, opened 2019-05-07
MH-13534 Add Hourly Data Resolution For Statistics

krinnewitz:
This PR consists of two parts:

The first part allows for different sources for different resolutions with influx. This changes the handling of statistics configurations such that for each provider type, there can be a different configuration format. This makes the mechanism more flexible and reflects the actual variety of options that might come with different sources. This way, we also get rid of the "source" property of the configuration which used to be a compisite string. This didn't make too much sense since we were using JSON already. The new configuration format allows to specify different sources for one influx provider, so that one can use specific measuemrents for certain data resolutions. In practice you might want to use shorter retentions for finer-grained measurements. This change enables you to do so.

This makes it possible to implement the second part: Hourly data resolution for statistics. This adds a new data resolution "hourly" to support finer-grained charts.

Also, I have included two sample configs for an easy review. I will remove these configs before this PR gets merged.

### DIFF
--- a/docs/guides/admin/docs/configuration/admin-ui/statistics.md
+++ b/docs/guides/admin/docs/configuration/admin-ui/statistics.md
@@ -75,9 +75,9 @@ For each provider, the following properties have to be configured:
     - **`resourceIdName`**, e.g. `episodeId` tells Opencast that the InfluxDB tag identifying the resource type this
        provider refers to is named `episodeId`.
     - **`resolutions`** is a list of resolutions supported by this provider. Opencast allows the user to select a
-      _resolution_ with which the data is displayed. Valid values are `DAILY`, `WEEKLY`, `MONTHLY` and `YEARLY`. E.g.
-      when a chart shows data of two years, a `DAILY` resolution will lead to 2x365=730 values to be plotted while a
-      `MONTHLY` resolution would leave us with 24 values being plotted in the chart.
+      _resolution_ with which the data is displayed. Valid values are `HOURLY`, `DAILY`, `WEEKLY`, `MONTHLY` and
+      `YEARLY`. E.g. when a chart shows data of two years, a `DAILY` resolution will lead to 2x365=730 values to be
+      plotted while a `MONTHLY` resolution would leave us with 24 values being plotted in the chart.
 - **`type`** defines the structure of the data provided by this provider. Currently, only `timeseries` is supported.
 
 Here is an example json configuration for a provider which generates charts for episodes showing the number of views:
--- a/docs/guides/admin/docs/configuration/admin-ui/statistics.md
+++ b/docs/guides/admin/docs/configuration/admin-ui/statistics.md
@@ -65,23 +65,19 @@ For each provider, the following properties have to be configured:
 - **`description`** is the description to be displayed with the chart. This can be a translation key.
 - **`resourceType`** tells Opencast to which type of entity the chart refers to. Valid values are `EPISODE`, `SERIES`,
   and`ORGANIZATION`. This is used by Opencast to decide where to display the chart.
-- **`resolutions`** is a list of resolutions supported by this provider. Opencast allows the user to select a
-  _resolution_ with which the data is displayed. Valid values are `DAILY`, `WEEKLY`, `MONTHLY` and `YEARLY`. E.g. when a
-  chart shows data of two years, a `DAILY` resolution will lead to 2x365=730 values to be plotted while a `MONTHLY`
-  resolution would leave us with 24 values being plotted in the chart.
-- **`source`** is a composite string which holds information about your data schema and how to calculate the
-  actual data to be displayed in the Admin UI. For InfluxDB, the `source` string consists of 5 parts, separated by
-  colons (:). A valid `source` string would be `influx:infinite.impressions_daily:SUM:value:episodeId` which has the
-  following meaning:
-
-    1. `influx` tells Opencast that InfluxDB is the data source.
-    2. `infinite.impressions_daily` tells Opencast that your InfluxDB data retention policy is named `infinite` and your
-        InfluxDB measurement name is `impressions_daily`.
-    3. `SUM` tells Opencast that InfluxDB's `SUM()` function should be used to calculate the values to display in the
-        chart.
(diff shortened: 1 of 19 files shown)


modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/StatisticsUtil.java
```diff
@@ -67,6 +68,12 @@ private StatisticsUtil() {
           result.add(currentWeekStart); // No min, max here, because influx uses previous monday as bucket label
           localStart = localStart.plusDays(7).plusDays(daysOff).withHour(0).withMinute(0).withSecond(0).withNano(0);
           break;
+        case HOURLY:
+          final Instant currentHourStart = ZonedDateTime.of(localStart.withMinute(0).withSecond(0).withNano(0), zoneId)
+              .toInstant();
+          result.add(Ordering.natural().min(to, Ordering.natural().max(from, currentHourStart)));
+          localStart = localStart.plusHours(1).withMinute(0).withSecond(0).withNano(0);
+          break;
```
KatrinIhler, line 76: Do we need the min here? I assume that neither from nor currentHourStart can ever be > to, but I might be wrong.

modules/statistics-service-api/src/test/java/org/opencastproject/statistics/api/StatisticsUtilTest.java
```diff
@@ -49,6 +49,31 @@ private static Instant randomInstant() {
...
+      assertEquals(
+          LocalDateTime.ofInstant(
+              to,
+              ZoneOffset.UTC)
+              .withMinute(0)
+              .withSecond(0)
+              .withNano(0)
+              .toInstant(ZoneOffset.UTC),
+          buckets.get(buckets.size() - 1));
+    }
```
KatrinIhler, line 92: This test confuses me.
1. Why do we test with random instants instead of a set of predefined test cases?
2. Why do we only check the last bucket?

docs/guides/admin/docs/configuration/admin-ui/statistics.md
```diff
@@ -94,13 +90,18 @@ Here is an example json configuration for a provider which generates charts for
...
+    "aggregation": "SUM",
+    "aggregationVariable": "value",
+    "resourceIdName": "episodeId",
+    "resolutions": [
+      "DAILY",
+      "WEEKLY",
+      "MONTHLY",
+      "YEARLY"
+    ]
+  }],
```
KatrinIhler, line 104: Looks much better now! :)

KatrinIhler (commented):
My thoughts after looking the code over:

modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/StatisticsUtil.java
krinnewitz: Here we set `currentHourStart` to  an Instant with minute `0`. The idea of the min max is to not run out of the bounds defined by `from` and `to` in case `from` and/or `to` have minute != 0. E.g. in case of from=...T08:30:00Z we don't want to get the 08:00 Instant. However, this is more or less theoretical since the frontend should only provide full-hour Instants. (I am not sure what happens in the greatly feared timezone "America/St_Johns", though...)

modules/statistics-service-api/src/test/java/org/opencastproject/statistics/api/StatisticsUtilTest.java
krinnewitz: We also check the first bucket, but not the ones in between :) Using the random Instants was easier than sitting down an thinking about all possible edge cases. 

So the answer to both of your questions is: laziness (or "prioritization" in other words :)). I will spend some extra time to improve the tests.

krinnewitz: I have extended the tests a bit.

KatrinIhler: Alright, now I'm wondering why we need the randomised tests at all if we now have a nice set of predefined instants to test. I'm not too fond of random test cases because this could lead to tests failing only sometimes, which is terrible for debugging...

modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/StatisticsUtil.java
KatrinIhler: My thought process is something like this: We make sure that localStart is before localEnd in the loop, and currentHourStart can only ever become less by setting minutes, seconds etc to 0, not more, right? So while currentHourStart could become less than from, I can't imagine a case where currentHourStart could ever be bigger than to. Or am I missing something?

modules/statistics-service-api/src/test/java/org/opencastproject/statistics/api/StatisticsUtilTest.java
krinnewitz: Well, I wouldn't consider 3 date pairs a "nice set of predefined instants to test". The idea with the random tests is to log the input so you can debug it very easily. However, I just noticed that Opencast has set the log level to "WARN" for all tests running with maven so my logs don't get printed. Therefore, I added the from and to dates to the output which is printed in case a test fails.

I actually like randomized testdata since most of the time, people just test the cases they can imagine (which are mostly identical with the ones they have in mind when developing).

Though it still makes sense to have hand-picked test data as well to make sure that some important cases are always tested.

modules/statistics-service-api/src/main/java/org/opencastproject/statistics/api/StatisticsUtil.java
krinnewitz: Ah now I see it. You are right. `min` is not required, but `max` is. I removed the `min`.

KatrinIhler (approved):

wsmirnow (approved):
I'm fine with this PR. Thank you.

PR #879, opened 2019-05-09
MH-13535 statistics csv export

krinnewitz:
This patch adds "Download" buttons to the statistics graphs in the Admin UI. Pressing these buttons will trigger a CSV file download. The  content of the CSV file depends on the current settings of the graph (e.g. from, to, granularity). Basically, the data seen in the graph is exported to the CSV.
    
For series statistics data, the behavior is configurable: with the default settings, you just get the data you see (just like for episode and organization). With an advanced configuration, you get the data of all events of the series, instead. To enable this, you have to configure which statistics providers should be used to retrieve the events' statistics.

I have added example statistic provider configurations for an easy review. Before merging, I will remove them again.


etc/org.opencastproject.statistics.export.impl.StatisticsExportCSVImpl.cfg
```diff
@@ -0,0 +1,13 @@
...
+# data. So you get more detailed data when providing the mappings. However, if you don't provide a mapping for a certain
+# provider, you will be able to export the lower detailed series data. If you want the detailed data, Opencast needs to
+# know which statistics provider should be used on the episodes of the series you are exporting. That's why this mapping
+# is necessary.
+# Example: series.to.event.provider.mappings=\{\"series.views.sum.influx\":\"episode.views.sum.influx\",\"foo\":\"bar\"\}
+#
+# Default: not set
+#series.to.event.provider.mappings=
+
+series.to.event.provider.mappings=\{\"series.views.sum.influx\":\"episode.views.sum.influx\"\}
```
gregorydlogan: I'm assuming this is the bit that's being removed prior to merge?

gregorydlogan: Also, this requires the recent influx adaptor to be both installed and properly configured?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -318,6 +356,44 @@ public Response getProviderData(@FormParam("data") String data) {
...
 
+  @GET
+  @Path("export.csv")
+  @Produces(MediaType.TEXT_PLAIN)
+  @RestQuery(name = "getcsvdata", description = "Returns the statistical data for a specific provider and a specific resource as CSV.", returnDescription = "The statistical data as CSV", restParameters = {
+    @RestParameter(name = "providerId", isRequired = true, description = "The provider id", type = RestParameter.Type.TEXT),
+    @RestParameter(name = "resourceId", isRequired = true, description = "The resource id", type = RestParameter.Type.TEXT),
+    @RestParameter(name = "from", isRequired = true, description = "The from date in iso 8601 UTC notation", type = RestParameter.Type.TEXT),
+    @RestParameter(name = "to", isRequired = true, description = "The to date in iso 8601 UTC notation", type = RestParameter.Type.TEXT),
+    @RestParameter(name = "dataResolution", isRequired = true, description = "The data resolution", type = RestParameter.Type.TEXT)},
```
gregorydlogan: The data resolution in what?  Seconds?  Millis?  Without knowing what's in this method I don't know what my options are.

etc/org.opencastproject.statistics.export.impl.StatisticsExportCSVImpl.cfg
gregorydlogan: Also, while I'm not opposed to using json here, none of the other config files are json based, so I would suggest changing this out to match.

modules/statistics-export-csv/src/main/java/org/opencastproject/statistics/export/impl/StatisticsExportCSVImpl.java
```diff
@@ -0,0 +1,241 @@
...
+
+  @Override
+  public String getCSV(StatisticsProvider provider, String resourceId, Instant from, Instant to, DataResolution
+      dataResolution, AbstractSearchIndex index, ZoneId zoneId) throws SearchIndexException, UnauthorizedException,
+      NotFoundException {
+    if (!(provider instanceof TimeSeriesProvider)) {
+      throw new IllegalStateException("CSV export not supported for provider of type " + provider.getClass().getName());
+    }
+    final StringWriter stringWriter = new StringWriter();
+    try (CSVPrinter printer = CSVFormat.EXCEL.withHeader(header).print(stringWriter)) {
```
gregorydlogan: I haven't actually run this code yet, but this looks like we're spitting out the binary CSV format for Excel, correct?  Is that a hard requirement?  I, personally, hate getting not-actually-comma-separated data as a CSV, so I would be tempted to make this spit out true CSV data instead.

gregorydlogan (changes requested):
This PR looks good, although I haven't had a chance to actually run the code.  It is, however, missing documentation in the admin markdown docs.

etc/org.opencastproject.statistics.export.impl.StatisticsExportCSVImpl.cfg
krinnewitz: Ok, I changed it to not use JSON anymore and I also reverted the commit which adds the sample configs. If you want to test it, you can revert my revert and the configs will be back again :D

modules/statistics-export-csv/src/main/java/org/opencastproject/statistics/export/impl/StatisticsExportCSVImpl.java
krinnewitz: Here is a description of the different formats supported by apache commons csv: 

http://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html

I changed it to RFC4180 to avoid confusion but this doesn't make a difference in our case. The output is a plain text CSV file which can be used with excel or any other tool. However, since there is no real standard for CSV, Some tools may require the file to be edited before it can be imported.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
krinnewitz: I refined the documentation a bit. By the way, you are the first person I notice which actually reads these annotations :P

krinnewitz:
Thanks for the review @gregorydlogan. I think I have resolved your findings. I would be happy to hear back from you next week or so :)

etc/org.opencastproject.statistics.export.impl.StatisticsExportCSVImpl.cfg
```diff
@@ -0,0 +1,11 @@
+# A comma separated list of mappings which map from series provider IDs to episode provider IDs. The first part of each
+# mapping has to be a series provider ID. A colon marks the beginning of the second part of the mapping which has to be
+# a corresponding episode provider ID. If these mappings are defined, then, when exporting series statistics data, the
+# data of all events of the series will be exported, instead of the series data. So you get more detailed data when
+# providing the mappings. However, if you don't provide a mapping for a certain provider, you will be able to export the
```
KatrinIhler: "will only"?
I also feel like this description could be a bit shorter and on point.

modules/statistics-export-csv/src/main/java/org/opencastproject/statistics/export/api/StatisticsExportCSV.java
```diff
@@ -0,0 +1,66 @@
...
+import org.opencastproject.matterhorn.search.SearchIndexException;
+import org.opencastproject.security.api.UnauthorizedException;
+import org.opencastproject.statistics.api.DataResolution;
+import org.opencastproject.statistics.api.StatisticsProvider;
+import org.opencastproject.util.NotFoundException;
+
+import java.time.Instant;
+import java.time.ZoneId;
+
+public interface StatisticsExportCSV {
```
KatrinIhler: I feel like the fact that it's CSV data that is downloaded is a bit too ingrained in this PR (e.g. used in lots of names). If we want to change this later or make it configurable this is gonna mean lots of changes. I think we could focus more on the Statistics export, for example calling it StatisticsExportService or something, and less on the CSV aspect. That way we could add more methods for different file formats in the future. What do you think?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
```diff
@@ -155,24 +162,55 @@ public String getDataResolution() {
     private DataResolution dataResolution;
 
     ProviderQuery(RawProviderQuery query) {
```
KatrinIhler: Since this seems to grow, it's maybe time to move this to its own file?

modules/statistics-export-csv/pom.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
```
KatrinIhler: Would you mind adding the maven dependency plugin to new modules to check for unnecessary dependencies? 
To provide an example:
https://github.com/opencast/opencast/blob/7d1aeaad6ea4e3ad49d58148cad1c421d24a1806/modules/animate-impl/pom.xml#L88-L98

We have started to add that to some old modules and the goal would probably be to have it in all, so it makes sense to include it in any new modules.

gregorydlogan: I wonder if there's a way to make a check for this part of checkstyles?

etc/org.opencastproject.statistics.export.impl.StatisticsExportCSVImpl.cfg
gregorydlogan: Grain of salt: I'm not at all familiar with the influx related code

Without looking at the reverted commit, I wouldn't know what to put here.  Looking at the documentation, I'm still not sure what *should* be here.  A fully worked example (using the test data in the docs) would make sense.

krinnewitz: I have shortened the description.

@gregorydlogan The example in the config file should work with the example from the docs (although "foo:bar" is useless there).

modules/statistics-export-csv/src/main/java/org/opencastproject/statistics/export/api/StatisticsExportCSV.java
krinnewitz: Done

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/StatisticsEndpoint.java
krinnewitz: Done

modules/statistics-export-csv/pom.xml
krinnewitz: Done

krinnewitz:
thanks @KatrinIhler and @gregorydlogan. I have fixed your recent findings.

PR #880, opened 2019-05-09
MH-13498 Improve icons and wording in video editor

krinnewitz:
This patch improves the usability of the video editor by:

- removing unnecessary buttons
- improving the wording in some places
- removing confusing button icons
- replacing some icons with more suitable ones
- adding some button tooltips
- dynamically changing tooltips of show/hide buttons of segments

Before:
Screen Shot 2019-05-09 at 15 49 03

After:
Screen Shot 2019-05-09 at 15 50 19


modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -159,8 +159,8 @@
      "THEME_NOT_DELETED":    "The theme could not be deleted",
      "TASK_CREATED": "The task has been created",
      "TASK_NOT_CREATED": "The task could not be created",
-     "VIDEO_CUT_NOT_SAVED": "The editing information could not be saved",
-     "VIDEO_CUT_SAVED": "The editing information has been saved",
+     "VIDEO_CUT_NOT_SAVED": "The changes could not be saved",
+     "VIDEO_CUT_SAVED": "Changes saved",
```
KatrinIhler: I feel like this is inconsistent with each other and also with other translation keys above. Because of this I'd prefer "The changes were saved".

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
      },
      "ACTIONS": {
        "SPLIT": "Split",
-       "CUT":   "Cut",
-       "REPLAY_SEGMENT": "Replay segment",
-       "REPLAY_END_OF_SEGMENT": "End of segment",
+       "SPLIT_TOOLTIP": "Create a new trim mark at the current cursor position.",
+       "CUT":   "Hide/Show",
+       "CUT_TOOLTIP":   "Hide or show the selected part of the video.",
```
KatrinIhler, line 1537: "Hide/show" here, too?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
+       "REPLAY_SEGMENT": "Play Start of Segment",
+       "REPLAY_SEGMENT_TOOLTIP": "Play the first seconds of the current segment.",
```
KatrinIhler: Maybe "Play Beginning of Segment" and "Play current segment from the beginning"? (Since it's not just playing a few seconds.)

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
+       "REPLAY_END_OF_SEGMENT": "Play End of segment",
+       "REPLAY_END_OF_SEGMENT_TOOLTIP": "Play a few seconds at the end of the curent segment.",
```
KatrinIhler: current is missing an r.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
      },
      "ACTIONS": {
        "SPLIT": "Split",
-       "CUT":   "Cut",
-       "REPLAY_SEGMENT": "Replay segment",
-       "REPLAY_END_OF_SEGMENT": "End of segment",
+       "SPLIT_TOOLTIP": "Create a new trim mark at the current cursor position.",
```
KatrinIhler: Maybe "split mark" instead of "trim mark" since we don't necessarily trim? But I'm kinda unsure about this one, I don't know how prevalent "trim mark" is in video editing.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
+       "CUT":   "Hide/Show",
+       "CUT_TOOLTIP":   "Hide or show the selected part of the video.",
+       "REPLAY_SEGMENT": "Play Start of Segment",
+       "REPLAY_SEGMENT_TOOLTIP": "Play the first seconds of the current segment.",
+       "REPLAY_END_OF_SEGMENT": "Play End of segment",
```
KatrinIhler: Segment

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
+       "REPLAY_END_OF_SEGMENT_TOOLTIP": "Play a few seconds at the end of the curent segment.",
        "REPLAY_SEGMENT_WITH_PREROLL": "Check transition",
+       "REPLAY_SEGMENT_WITH_PREROLL_TOOLTIP": "Play a few seconds at the end of the previous segment and a few seconds at the start of the current segment to see the transition between these two segment.",
        "UNDO": "Undo",
        "REDO": "Redo",
-       "CLEAR_SEGMENTS": "Clear segments",
-       "CLEAR_SELECTED_SEGMENT": "Clear selected segment"
+       "CLEAR_SEGMENTS": "Clear All Trim Marks",
+       "CLEAR_SEGMENTS_TOOLTIP": "Clear all trim marks. This leaves you with just one segment spanning the whole video."
```
KatrinIhler: Maybe "split marks" here, too?
Also I think the second sentence in the tool tip is unnecessary. If you want to keep the info, though, I'd shorten it to "Clear all trim marks, leaving one segment" or something.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
-       "REPLAY_SEGMENT": "Replay segment",
-       "REPLAY_END_OF_SEGMENT": "End of segment",
+       "SPLIT_TOOLTIP": "Create a new trim mark at the current cursor position.",
+       "CUT":   "Hide/Show",
+       "CUT_TOOLTIP":   "Hide or show the selected part of the video.",
+       "REPLAY_SEGMENT": "Play Start of Segment",
+       "REPLAY_SEGMENT_TOOLTIP": "Play the first seconds of the current segment.",
+       "REPLAY_END_OF_SEGMENT": "Play End of segment",
+       "REPLAY_END_OF_SEGMENT_TOOLTIP": "Play a few seconds at the end of the curent segment.",
        "REPLAY_SEGMENT_WITH_PREROLL": "Check transition",
```
KatrinIhler: Transition.
Also maybe "Play Transition" for consistency?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1532,18 +1532,24 @@
...
+       "REPLAY_SEGMENT_WITH_PREROLL_TOOLTIP": "Play a few seconds at the end of the previous segment and a few seconds at the start of the current segment to see the transition between these two segment.",
```
KatrinIhler: Too long imo. Maybe "Play a few seconds containing the end of the previous and the beginning of the next segment."?

KatrinIhler (changes requested):
I guess one could argue about UI for hours and none of us are usability experts (as far as I know), but here are my 2 cents.

I'm also wondering: Do we really need the Split and Hide/Show button, especially the latter since there are already two other possibilities to hide/show a segment? I feel like they are quite unnecessary, but I'm also not a frequent user of the video editor.

Edit: One more thing: Since we don't show the x for the last segment in the timeline anymore, maybe also hide the x for segments that can't be deleted (which happens if there's only one visible segment) in the list instead of greying it out?

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: fixed

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: Since the tooltip is allowed to take up more space than the button, we decided to stick with "Hide or show" for the tooltip.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: Your proposal for the tooltip to read "Play current segment from the beginning" is a good way to explain what "Play Start of Segment" actually means. "Start" is consistent with the rest of the UI. So we decided to Change the tooltip, but keep the button label.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: fixed

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: Changed the wording. "Trim mark" is used no longer.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: fixed

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: Changed the wording. "Trim mark" is used no longer.

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: fixed

modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
krinnewitz: Fixed

staubesv:
@KatrinIhler 

> Edit: One more thing: Since we don't show the x for the last segment in the timeline anymore, maybe also hide the x for segments that can't be deleted (which happens if there's only one visible segment) in the list instead of greying it out?

OMG... just had a look at this with @krinnewitz and I am somewhat shocked about this behaviour: Why should the video editor want to prevent the user from "Hiding" or "Deleting" a segment?

I've created https://opencast.jira.com/browse/MH-13576 describing this as usability bug. Once the two open video editor pull requests are merged, we will most likely fix this with a separate pull request.

krinnewitz:
@KatrinIhler thanks for the review. 

We decided to keep the "hide/show" button as well as the "split" button in order to not have too controversial changes in this PR. However, we basically agree that these buttons could be removed. Maybe someone can create a separate PR for this.

I think all your findings should be fixed/answered.

KatrinIhler:
Just one thing:
The last segment in the timeline is not deletable, so no x is shown, however, the eye for hiding is shown even when no hiding is possible, it just doesn't do anything. We should choose one behavior here. Same goes for the list of segment: x is greyed out, the eye isn't, both don't do anything if it's the last unhidden segment. (Whether that's correct is another question, I think the UI should be consistent here even if we're going to change that behaviour later.)

krinnewitz:
@staubesv Do you have a preference here? Greying out the eye-icon is quite hard to implement as far as I remember. Should we just undo the visibility changes regarding the x-icon of the last segment?

staubesv:
@krinnewitz For now, just undo the x-icon hiding of the last segment. Later we can have a look at this when doing https://opencast.jira.com/browse/MH-13576 in a separate PR.

@KatrinIhler This will not grey out the x-icon or the eye-icon, though. It is then just as it has been before ;-)

KatrinIhler (approved):

KatrinIhler:
Whoops, approved a bit prematurely, there still is that one change missing. But sounds like a good plan, I will merge once it's done!

krinnewitz:
@KatrinIhler done.

staubesv:
@KatrinIhler I think the ball is on your side, isn't it?

PR #881, opened 2019-05-09
MH-13539, JavaScript Dependency Management

lkiesow:
Instead of keeping all admin interface JavaScript dependencies managed
by bower in our git repository, it makes much more sense to actually use
bower and trigger the installation just like we do that with npm modules.

*Note for the reviewer: It should suffice to take a look at commit f8c7bf6. The second one just removes the bower_components folder*


staubesv (dismissed):
The following error message appear in the output of mvn clean install:

`[ERROR] bower angular-translate                            extra-resolution Unnecessary resolution: angular-translate#2.18.1`

While that is not caused by this PR and won't break anything, it might be unexpected to other developers to see this error when building (after having this PR).

Could you please remove that extra resolution from bower.json?

staubesv (approved):

PR #884, opened 2019-05-10
MH-13542 Display global notifications as overlay

krinnewitz:
In Opencast's admin UI, notifications are used across various places.
This patch changes the look and feel of "global" notifications. By
"global" notifications, we mean those notifications, which are not
displayed inside a modal or wizard.

With this patch, global notifications appear as boxes in the top right
corner of the admin UI. This has some advantages:

- They are always visible, regardless of the scroll position within the
  page
- They don't take space and hence, do not move all subsequent html
  elements

localhost_9000_

### DIFF
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -281,6 +281,20 @@ prop.admin.shortcut.general.main_menu=m
 #
 #prop.admin.notification.duration.warning=5
 
+# Display position for global notifications
+#
+# Possible positions are:
+# - top-right
+# - top-left
+# - top-center
+# - bottom-right
+# - bottom-left
+# - bottom-center
+#
+# Default: bottom-right
+#
+#prop.admin.notification.position.global=bottom-right
+
 # If the preview mode in the video editor is enabled per default.
 #
 # Format: Boolean
--- a/modules/admin-ui/src/main/webapp/scripts/modules/configuration/partials/index.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/configuration/partials/index.html
@@ -13,7 +13,7 @@
 </section>
 
 <div class="main-view">
-  <div data-admin-ng-notifications="" class="global-notifications"></div>
+  <div data-admin-ng-notifications=""></div>
 
   <div class="controls-container">
     <div class="filters-container">
--- a/modules/admin-ui/src/main/webapp/scripts/modules/configuration/partials/index.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/configuration/partials/index.html
@@ -13,7 +13,7 @@
 </section>
 
 <div class="main-view">
-  <div data-admin-ng-notifications=""></div>
+  <div data-admin-ng-notifications="" class="global-notifications"></div>
 
   <div class="controls-container">
     <div class="filters-container">
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/index.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/index.html
@@ -22,7 +22,7 @@
 </section>
 
 <div class="main-view">
-  <div data-admin-ng-notifications="" class="global-notifications"></div>
+  <div data-admin-ng-notifications=""></div>
 
   <div class="controls-container">
     <div class="filters-container">
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/index.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/index.html
@@ -22,7 +22,7 @@
 </section>
 
 <div class="main-view">
-  <div data-admin-ng-notifications=""></div>
+  <div data-admin-ng-notifications="" class="global-notifications"></div>
 
   <div class="controls-container">
     <div class="filters-container">
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
@@ -28,8 +28,6 @@ angular.module('adminNg.controllers')
     var thumbnailErrorMessageId = null;
     var trackErrorMessageId = null;
 
-    var LOCAL_CONTEXT = 'video-tools';
-
     $scope.navigateTo = function (path) {
       $location.path(path).replace();
     };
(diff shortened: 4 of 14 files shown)


mtneug:
From the picture it seems that the notifications would block the buttons at the top of the page. Wouldn't it be better to place the notifications in either of the bottom corners where less important elements would be blocked? No strong opinion though.

staubesv:
@mtneug We also had that concern and the idea to mitigate the "issue" by moving the notifications to the bottom-right corner,  but our first testing indicated that there is not actually an issue. 
As it seems more common to have the notifications at the top-right corner, that motivated us to stay with top-right. I suggest to leave it that way and change it in case people report being annoying (when using the change in production).

karendolan:
2¢, notification overlays are as equally disruptive as moving the page's HTML in a dedicated area. The choice is that either a click goes to the wrong event row when the dedicated area notification moves the HTML. Or, the click target is completed covered over by the notification popup, work has to be halted to manually dismiss the overlay in order to get to the intended underlying click target. Personally, I prefer dedicated area notifications because they cause less interruption. However, the right corner notification overlay is annoyingly common in many applications, thus defacto standard.

mtneug:
@staubesv yes, top-right is probably most common, but I have seen other corners used as well (maybe not top left). The theodul player e.g. uses bottom right. Anyway, your suggestion seems reasonable to me.

Personally, I would prefer the overlay notifications.

KatrinIhler:
I'm kinda conflicted about this, too. Karen explained the problems with both the old and the new solution quite well in my opinion. We just had a short chat about this here and what we wondered about was: Would it be feasible to have a dedicated notification area that is normally hidden behind a button (think the Facebook bell or something)?
But the problem with that is that we use the notifications a lot for direct feedback to a user action - did my event upload successfully or not? Did the saving in the video editor work or not?
And then we wondered whether we even need notifications for that or whether this feedback couldn't be implemented in a more succinct way - think a green check mark or red cross next to a button or something like that. Do we even need text for this?
I don't really have a firm opinion yet, it's just something to think about.

staubesv:
> And then we wondered whether we even need notifications for that or whether this feedback couldn't be implemented in a more succinct way - think a green check mark or red cross next to a button or something like that. Do we even need text for this?

The problem is the backend: Because of the way the backend works (working with files, asynchronous search index updates), we cannot guarantee that the change the user makes is directly visible (user feedback). If it was, we would not need any notifications.
Think of event deletion: If the event would immediately disappear when the "Delete Window" closes, we would not need the notification. But waiting until the search index is updated and reload the data from the backend before closing the window takes too long. For other operations, performance is even worse, e.g. creating 100 scheduled events will create several hundreds files on disk which simply does take long.
IMO, that is very much the reason why we need the notifications as blocking the user (keeping windows open until we are 100% sure the changes are visible in the Admin UI) is annoying ("wait") and not saying anything about what (will) happen is confusing.

> Do we even need text for this?

I think we do: At least for the error notifications, it makes sense to say what went wrong (error, not authorized, ...). We could also improve the info notifications (e.g. 32 scheduled events created vs. event created).

While I agree with @karendolan that both the current solution and the notifications suggested in this PR have a similar problem (button moves around under mouse vs. button is hidden behind notification), the suggested solution results in a less "noisy" (things don't always move around) design so I think the notification overlay is better.

Anyway: It's best to build this PR and fiddle around it locally as you really not to "feel" the "look & feel". For my taste, "Look & feel" gets better.

JulianKniephoff:
As someone whose default zoom level in the browser is **at least** 200% at all times, I really don't like this style of notification. See here:

Screenshot at 2019-05-22 14-29-52

With the right combination of zooming and scrolling, these things can get quite annoying really quickly. Plus they remind me of the thousands of pop-up, -over and -under things presumably modern websites bombard us with these days: Subscribe to my newsletter, chat live with our support team, share our content on these 37 social networks; I was on some blog yesterday where with my standard zoom level the content area was, I kid you not, literally 3 pixels high!

I also thing that the readability of the notification text suffers from them being placed on top of random UI elements and/or other text. Sure, they have a box around them, but it still looks cluttered overall and that is never good for readability, I think.

As an alternative I find "notification center"-like designs really pleasant. Think macOS or Windows 10: There is a dedicated area for notifications where have enough space to put all the information they want, but they are normally hidden. A button, potentially with some sort of notification indicator, would slide the whole UI to the side to reveal that panel. A modal would be another (and probably more consistent) possibility in our case.

staubesv:
> As an alternative I find "notification center"-like designs really pleasant.

I don't think the "notification centers" are used to confirm user actions by anyone.

Anyway, I'll start a discussing on list to get more opinions...

krinnewitz:
Also, I would recommend to everybody to actually try this PR out (if not done yet). Just by looking at the screenshot you won't get an idea of what these new notifications feel like.

KatrinIhler:
@krinnewitz I did try it out. I'm still not quite happy with it, but if we decide to go with the overlay design, I think I'd prefer the bottom right corner of the top right.

staubesv:
Just started a thread on list about this topic and set the PR to "on hold".

https://groups.google.com/a/opencast.org/forum/#!topic/users/GuOMr12K70M

lkiesow:
Since this was brought up again:

Having used this in production for some time now I'm favoring the old implementation over the new one. I regularly ran into the problem of having a notification overlay over buttons which I wanted to interact with.

There is certainly a better solution compared to the old one but this does not feel like a step in the right direction. Hence, I'm in favor of declining this change.

lkiesow:
We've discussed this particular pull request with a larger audience at the conference in Münster last week. Long story short, the majority agreed that having the overlays but simply moving them to the bottom-right corner would be the preferred short-term solution for this.

- @krinnewitz, can you make that change?
- @karendolan, @JulianKniephoff, @KatrinIhler, would that be okay for you as well?

JulianKniephoff:
I unfortunately missed that discussion, so excuse me if I bring up things that were already discussed, but aren't the notifications even **more** likely to cover things up in the bottom right? If you don't fit the entire page on one screen, which is pretty much always the case for me, that virtually guarantees that they overlap the main table.

I think I personally prefer top-right to bottom-right, but I still like the way it was better than both of these alternatives. (Which is not to say that I **like** the way it was. :wink:)

karendolan:
@lkiesow @JulianKniephoff  For us, notification overlays have  less direct interference in the bottom-right corner than the upper-right corner where a heavy concentration of Opencast controls currently live. Users typically filter events of concern and sort them to the top of the event table. The bottom row of events is less important than the top row. 

IMHO, if the notification overlays go back to the upper-right corner, the Opencast controls that live there now, should migrate a bit left. Or, overlay notifications should swipe further into the middle of that big unused area in the top-middle of the Admin UI. Between the concentration of left-justified controls and the concentration of right-justified controls (on non-mobile monitors).
> Long story short, the majority agreed that having the overlays but simply moving them to the bottom-right corner would be the preferred short-term solution for this.

+1

krinnewitz:
I talked to Markus of SWITCH and he suggested to make the position configurable with bottom right being the default. I will implement it within the next few weeks.

krinnewitz:
Just made the notification position configurable. You can choose between:

- top-left
- top-right
- top-center
- bottom-left
- bottom-right (default)
- bottom-center

krinnewitz:
Maybe someone can remove the "on hold" label

KatrinIhler (approved):
I think this is something we can all live with :)

PR #886, opened 2019-05-10
MH-13544 Video editor shows incorrect notification

staubesv:
Steps to reproduce: 
1. Upload two events "Test1" and "Test2" and wait until processing is finished 
2. Start a workflow on event "Test 1" 
3. Open video editor for "Test 1" 
4. Close video editor for "Test 1" using the Close button 
5. Open video editor for "Test 2" 
  
 Actual Results: 
A notification in the video editor says that the event cannot be editing because it is being processed 
  
 Expected Results: 
The notification should not appear because the event "Test 2" is not being processed at this point of time. 

Workaround: 
Ignore. 

This pull requests also addresses:
- Disable workflow dropdown while a transaction or submission is active (the save button was already disabled)
- Don't reuse variable activeTransaction to indicate active submissions. That is not necessarily a bug but a clearer separation of concerns

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
@@ -169,16 +169,16 @@ angular.module('adminNg.controllers')
     $scope.player = {};
     $scope.video  = ToolsResource.get({ id: $scope.id, tool: 'editor' });
 
-    $scope.activeTransaction = false;
+    $scope.activeSubmission = false;
 
     $scope.submit = function () {
-      $scope.activeTransaction = true;
+      $scope.activeSubmission = true;
       $scope.video.thumbnail.loading = $scope.video.thumbnail && $scope.video.thumbnail.type &&
               ($scope.video.thumbnail.type === 'DEFAULT');
       // Remember $scope.video.workflow as $scope.video.$save will potentially overwrite this value
       var closeVideoEditor = $scope.video.workflow;
       $scope.video.$save({ id: $scope.id, tool: $scope.tab }, function (response) {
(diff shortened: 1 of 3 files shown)


modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/video/editController.js
```diff
@@ -163,6 +166,10 @@ angular.module('adminNg.controllers')
       }
     });
 
+    $scope.$on('$destroy', function () {
+      Notifications.removeAll(NOTIFICATION_CONTEXT);
```
staubesv, line 170: This fixes the bug described in the issue.

modules/admin-ui/src/main/webapp/scripts/modules/events/subresources/controllers/toolsController.js
```diff
@@ -169,16 +169,16 @@ angular.module('adminNg.controllers')
     $scope.player = {};
     $scope.video  = ToolsResource.get({ id: $scope.id, tool: 'editor' });
 
-    $scope.activeTransaction = false;
+    $scope.activeSubmission = false;
```
staubesv, line 172: Don't reuse the variable `activeTransaction` which is updated in 3 seconds polling intervals and indicates whether the event is being processed. This reuse is not causing a bug but makes the code harder to understand, e.g. how do "event being processed" and "submission ongoing" interfere with each other when both statuses are stored in the same variable?

Note: In 7.x, this has been done, too, using the variable name `activeRequest` which seems a bit unspecific to me ;-) Then again, 7.x is missing some uses of `activeRequest` in the tools template which 6.x did better using `activeTransaction`.

CGreweling (commented):
Works.

PR #889, opened 2019-05-10
MH-13547 Limit accepted file types when uploading assets

krinnewitz:
This patch allows the admin to limit the accepted file types for asset
uploads. For each asset type, the allowed file types can be configured.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
@@ -48,6 +48,7 @@ import org.opencastproject.index.service.catalog.adapter.MetadataUtils;
 import org.opencastproject.index.service.catalog.adapter.events.CommonEventCatalogUIAdapter;
 import org.opencastproject.index.service.catalog.adapter.series.CommonSeriesCatalogUIAdapter;
 import org.opencastproject.index.service.exception.IndexServiceException;
+import org.opencastproject.index.service.exception.ListProviderException;
 import org.opencastproject.index.service.impl.index.AbstractSearchIndex;
 import org.opencastproject.index.service.impl.index.event.Event;
 import org.opencastproject.index.service.impl.index.event.EventHttpServletRequest;
@@ -57,7 +58,9 @@ import org.opencastproject.index.service.impl.index.group.GroupIndexSchema;
 import org.opencastproject.index.service.impl.index.group.GroupSearchQuery;
 import org.opencastproject.index.service.impl.index.series.Series;
 import org.opencastproject.index.service.impl.index.series.SeriesSearchQuery;
+import org.opencastproject.index.service.resources.list.api.ListProvidersService;
 import org.opencastproject.index.service.resources.list.query.GroupsListQuery;
+import org.opencastproject.index.service.resources.list.query.ResourceListQueryImpl;
 import org.opencastproject.index.service.util.JSONUtils;
 import org.opencastproject.index.service.util.RestUtils;
 import org.opencastproject.ingest.api.IngestException;
@@ -125,6 +128,7 @@ import org.opencastproject.workspace.api.Workspace;
 import com.entwinemedia.fn.Fn2;
 import com.entwinemedia.fn.Stream;
 import com.entwinemedia.fn.data.Opt;
+import com.google.common.net.MediaType;
 
 import net.fortuna.ical4j.model.Period;
 import net.fortuna.ical4j.model.property.RRule;
@@ -151,6 +155,7 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -165,6 +170,7 @@ import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
 
@@ -202,6 +208,7 @@ public class IndexServiceImpl implements IndexService {
   private CaptureAgentStateService captureAgentStateService;
   private EventCommentService eventCommentService;
   private IngestService ingestService;
+  private ListProvidersService listProvidersService;
   private AssetManager assetManager;
   private SchedulerService schedulerService;
   private SecurityService securityService;
@@ -322,6 +329,16 @@ public class IndexServiceImpl implements IndexService {
    */
   public void setIngestService(IngestService ingestService) {
     this.ingestService = ingestService;
+  }
+
+  /**
+   * OSGi DI.
+   *
+   * @param listProvidersService
+   *          the service to set
+   */
+  public void setListProvidersService(ListProvidersService listProvidersService) {
+    this.listProvidersService = listProvidersService;
   }
 
   /**
@@ -497,6 +514,9 @@ public class IndexServiceImpl implements IndexService {
               }
             }
           } else {
+            // AngularJS file upload lib appends ".0" to field name, so we cut that off
+            fieldName = fieldName.substring(0, fieldName.lastIndexOf("."));
+            checkAcceptedTypes(fieldName, MediaType.parse(item.getContentType()));
             if ("presenter".equals(item.getFieldName())) {
               mp = ingestService.addTrack(item.openStream(), item.getName(), MediaPackageElements.PRESENTER_SOURCE, mp);
             } else if ("presentation".equals(item.getFieldName())) {
@@ -587,6 +607,9 @@ public class IndexServiceImpl implements IndexService {
             }
           }
         } else {
+          // AngularJS file upload lib appends ".0" to field name, so we cut that off
+          fieldName = fieldName.substring(0, fieldName.lastIndexOf("."));
+          checkAcceptedTypes(fieldName, MediaType.parse(item.getContentType()));
           if (item.getFieldName().toLowerCase().matches(attachmentRegex)) {
             assetList.add(item.getFieldName());
             // Add attachment with field name as temporary flavor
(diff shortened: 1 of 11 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2143,4 +2166,40 @@ public boolean isWorkflowActive(String workflowState) {
...
+          final List<String> accepts = Arrays.stream(assetUpload.getOrDefault("accept", "*/*").split(","))
+              .map(String::trim).collect(Collectors.toList());
+          for (String accept : accepts) {
+            if (accept.contains("/") && mediaType.is(MediaType.parse(accept))) {
+              return;
+            } else if (mediaType.subtype().equalsIgnoreCase(accept.replace(".", ""))) {
+              return;
+            }
+          }
+          throw new IllegalArgumentException("Provided file format " + mediaType.toString() + " not allowed.");
```
KatrinIhler: Instead of throwing an exception here, why don't we return a boolean indicating whether the upload is allowed? I feel like that would be more intuitive.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2143,4 +2166,40 @@ public boolean isWorkflowActive(String workflowState) {
             || WorkflowState.PAUSED.toString().equals(workflowState);
   }
 
+  private void checkAcceptedTypes(String assetUploadId, MediaType mediaType) {
```
KatrinIhler: I feel kinda bad about the IndexService becoming even longer than it is now... what would you think about moving this into its own class? Too weird?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2143,4 +2166,40 @@ public boolean isWorkflowActive(String workflowState) {
...
+    try {
+      final Collection<String> assetUploadJsons = listProvidersService.getList("eventUploadAssetOptions",
+          new ResourceListQueryImpl(),false).values();
+      for (String assetUploadJson: assetUploadJsons) {
+        if (!assetUploadJson.startsWith("{") || !assetUploadJson.endsWith("}")) {
+          // Ignore non-json-values
+          continue;
+        }
+        @SuppressWarnings("unchecked")
+        final Map<String, String> assetUpload = (Map<String, String>) parser.parse(assetUploadJson);
```
KatrinIhler: Any idea how fast it is to parse that to JSON every time for every asset?

KatrinIhler (changes requested):
Okay, so I just tested this by bypassing the supported files in the admin ui and uploading a png as a presenter, and what I get is an Internal Server Error because of the IllegalArgumentException. I think that this should be a Bad Request instead. Same goes for uploading assets to existing events.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: I agree that throwing the exceptions inside this method is a bit counter-intuitive. However, returning a boolean here would not be sufficient since I wouldn't be able to throw an excpetion with a precise message when evaluating the boolean. This method would have to return an enum instead which would then have to be evaluated in a big switch-case. And since we would have to do this in two places, I think that would be a lot of code overhead. So i would rather stick with throwing inside the method which performs the check.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: I just took a look and there already is `RequestUtils` and I think this method fits in there quite well. So I moved it there.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: No, but I don't think that this is significant compared to all the other long-running operations which take place when uploading an asset. We are likely to talk about a few milliseconds for the JSON parsing. When I moved it to `RequestUtils` I also declared a static `JSONParser`, so the parser is re-used now.

krinnewitz:
> Okay, so I just tested this by bypassing the supported files in the admin ui and uploading a png as a presenter, and what I get is an Internal Server Error because of the IllegalArgumentException. I think that this should be a Bad Request instead. Same goes for uploading assets to existing events.

Yep, good catch. There were big try-catches which caught `Exception` which shadowed all the `IllegalArgumentException`s. The code looks like this was unintended. So I changed it to just catch the relevant checked exceptions. Now, bad request is returned as intended.

Thanks for the review. I think all your findings are fixed/answered.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -551,7 +569,8 @@ public String createEvent(HttpServletRequest request) throws IndexServiceExcepti
       }
 
       return createEvent(metadataJson, mp);
-    } catch (Exception e) {
+    } catch (FileUploadException | UnauthorizedException | ParseException | IngestException | SchedulerException
+        | MediaPackageException | IOException | NotFoundException e) {
       logger.error("Unable to create event: {}", getStackTrace(e));
       throw new IndexServiceException(e.getMessage());
```
KatrinIhler: While we're at it, maybe add the other exception as a cause instead of only handing over the message?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -620,7 +642,7 @@ public String updateEventAssets(MediaPackage mp, HttpServletRequest request) thr
       }
 
       return startAddAssetWorkflow(metadataJson, mp);
-    } catch (Exception e) {
+    } catch (MediaPackageException | FileUploadException | IOException | IngestException e) {
       logger.error("Unable to create event: {}", getStackTrace(e));
       throw new IndexServiceException(e.getMessage());
```
KatrinIhler: Same here.

modules/index-service/src/main/java/org/opencastproject/index/service/util/RequestUtils.java
```diff
@@ -72,4 +82,40 @@ private RequestUtils() {
     return fieldMap;
   }
 
+  public static void checkAcceptedTypes(String assetUploadId, MediaType mediaType, ListProvidersService listProvidersService) {
```
KatrinIhler: Javadoc would be nice here.

modules/index-service/src/main/java/org/opencastproject/index/service/util/RequestUtils.java
```diff
@@ -72,4 +82,40 @@ private RequestUtils() {
...
+          final List<String> accepts = Arrays.stream(assetUpload.getOrDefault("accept", "*/*").split(","))
+              .map(String::trim).collect(Collectors.toList());
+          for (String accept : accepts) {
+            if (accept.contains("/") && mediaType.is(MediaType.parse(accept))) {
+              return;
+            } else if (mediaType.subtype().equalsIgnoreCase(accept.replace(".", ""))) {
+              return;
+            }
+          }
+          throw new IllegalArgumentException("Provided file format " + mediaType.toString() + " not allowed.");
```
KatrinIhler: I still think these are different scenarios. The method is supposed to check the file type, so simply returning true or false makes sense to me for that. The other possibilities are errors that occur, so it makes sense to throw an exception for them, although I'm not sure whether that should be an unchecked one...

KatrinIhler (changes requested):
Delivers Bad Request now as intended. :) Some more comments below.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: Done

modules/index-service/src/main/java/org/opencastproject/index/service/util/RequestUtils.java
krinnewitz: True. I changed the method to return a boolean (finally! :D).

The question if exceptions should be checked or unchecked is very controversial these days. Checked exceptions don't go well with java 8 functional programming features. Checked exceptions are safer. When I look at the codebase, I think opencast has way too many checked exceptions. 

I would prefer to use the IllegalArgumentException as it is already mapped to "bad Request" and we don't have add more boilerplate code, but I have no strong opinion here. If you insist, I will change it to a checked exception.

krinnewitz:
Thanks @KatrinIhler. I think I have resolved all your comments.

modules/index-service/src/main/java/org/opencastproject/index/service/util/RequestUtils.java
krinnewitz: Done

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
krinnewitz: Done

modules/index-service/src/main/java/org/opencastproject/index/service/util/RequestUtils.java
KatrinIhler: I just discussed this with Lars and be both would prefer a checked exception here, maybe something like UnsupportedAssetException or something.

My reasoning is the following:
Yes, checked exceptions conflict with functional programming, but we don't use that here and checked exceptions are still an important part of the functionality of Java. I think an IllegalArgumentException would make sense if it would result from a programming error (somebody uses the method in the wrong way), that can't be handled from within the code. However in this case we get user input and we have to deal with that being wrong. In the end, this exception absolutely needs to be caught if we don't want it to result in a server error, which doesn't make any sense. So yeah, checked exception is the way to go here imo. ;)

krinnewitz: Fair enough. Just added a checked exception.

PR #895, opened 2019-05-12
MH-13552, Update Tesseract Code

lkiesow:
*Coming back to one of the first modules I have been working on… 😺*

The tesseract module contains a lot of deprecated code as well as code
meant for ideas which have never been realized and only complicate and
slow down the module.

This patch updated the old code, replaces deprecated methods and cleans
up unused code while keeping the same functionality.


modules/textextractor-tesseract/src/main/java/org/opencastproject/textextractor/tesseract/TesseractTextExtractor.java
```diff
@@ -182,18 +180,17 @@ public TextFrame extract(File image) throws TextExtractorException {
...
    */
-  protected String getAnalysisOptions(File image, File outputFile) {
-    StringBuilder options = new StringBuilder();
-    options.append(image.getAbsolutePath());
-    options.append(" ");
-    options.append(outputFile.getAbsolutePath());
-    options.append(" ");
-    options.append(this.addOptions);
-    return options.toString();
+  private List<String> getAnalysisOptions(File image, File outputFile) {
```
staubesv: The method name does not fit its purpose anymore. Can you please adjust it to something like getTesseractComand?

staubesv (dismissed):
Just found a minor issue (see comment).

staubesv (approved):

PR #896, opened 2019-05-13
MH-13553, Fix Paella Track Selection

lkiesow:
While flavors published to engage are completely configurable, the
Paella player has parts of them hard-coded making it unable to copy with
any but some very specific configurations.

For example, videos produced and published by the fast workflow will not
play in Paella while they play without problem in the Theodul player.

This patch fixes the problem by removing the hard-coded parts of the
flavor based track selection.

---

*PS: I'll raise a separate pull request fixing the spelling of flavor*

### DIFF
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -44,6 +44,7 @@ Configuration changes
   controlled by Opencast and the routing through capture agents which existed for historical reasons was just an
   additional source for errors. If you rely on the old behavior, it can be configured in
   `etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg`.
+- The Paella player now respects all tracks published to engage instead of tracks with the flavor `*/delivery` only.
 
 API changes
 -----------
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -102,8 +102,9 @@ class OpencastToPaellaConverter {
     return source;
   }
 
-  getStreamFromFlavour(episode, mainFlavour) {
-    var currentStream = { sources:{}, preview: '', content: mainFlavour };
+  getStreamFromFlavour(episode, flavour) {
+    var mainFlavour = flavour.split('/')[0],
+        currentStream = { sources:{}, preview: '', content: mainFlavour };
 
     var tracks = episode.mediapackage.media.track;
     var attachments = episode.mediapackage.attachments.attachment;
@@ -112,7 +113,7 @@ class OpencastToPaellaConverter {
 
     // Read the tracks!!
     tracks.forEach((currentTrack) => {
-      if (currentTrack.type == `${mainFlavour}/delivery`) {
+      if (currentTrack.type == flavour) {
         var videoType = this.getVideoTypeFromTrack(currentTrack);
         if (videoType){
           if ( !(currentStream.sources[videoType]) || !(currentStream.sources[videoType] instanceof Array)){
@@ -180,11 +181,7 @@ class OpencastToPaellaConverter {
     var paellaStreams = [];
     var flavours = this.getContentToImport(episode);
     flavours.forEach((flavour) => {
-      var flavourSplit = flavour.split('/');
-      if (flavourSplit[1] == 'delivery') {
-        var s = this.getStreamFromFlavour(episode, flavourSplit[0]);
-        paellaStreams.push(s);
-      }
+      paellaStreams.push(this.getStreamFromFlavour(episode, flavour));
     });
     return paellaStreams;
   }
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -102,6 +102,12 @@ class OpencastToPaellaConverter {
     return source;
   }
 
+  /**
+   * Extract a stream identified by a given flavor from the media packages track list and try to find a corresponding
+   * image attachment for the selected track.
+   * @param episode  result structure from search service
+   * @param flavor   flavor used for track selection
+   */
   getStreamFromFlavour(episode, flavour) {
     var mainFlavour = flavour.split('/')[0],
         currentStream = { sources:{}, preview: '', content: mainFlavour };
@@ -181,7 +187,8 @@ class OpencastToPaellaConverter {
     var paellaStreams = [];
     var flavours = this.getContentToImport(episode);
     flavours.forEach((flavour) => {
-      paellaStreams.push(this.getStreamFromFlavour(episode, flavour));
+      var stream = this.getStreamFromFlavour(episode, flavour);
+      paellaStreams.push(stream);
     });
     return paellaStreams;
   }


staubesv (changes requested):
Hmm... this potentially breaks compatibility with existing publications depending on what flavors Adopters publish to engage, so it should at very least be documented in the upgrade notes, shouldn't it?

Looks good except of that.

lkiesow:
It will already be listed in the changelogs. Also, files which are published to the players are shown at other places already (e.g. other player, search service results, feeds, …) so if you want to hide them, there is something already wrong with your configuration.

staubesv:
@lkiesow I thought about Adopters that are already are using the Paella player which displayed */delivery so far even if other tracks have been published to "engage". 

In our case, for example, we publish both dual-stream and single-stream variants a single publication channel but the player will only show dual-stream if available. This change would break such a setup if an Adopter would something similar with the Opencast Paella Player (we use the standalone Paella Player and are therefore not affected).

A changelog entry "Fix Paella Track Selection" won't help much because depending on what you are doing, this could also just be labelled "Break the Paella Track Selection" in case your setup assumes the Player plays `*/delivery` tracks only (which it currently does) ;-)

From that point of view, it seems noteworthy to mention this change in the upgrade notices because it potentially could break an existing setup.

... as it doesn't break ours, I don't have a strong opinion, but still would recommend to mention this kind of changes as "breaking change".

> Also, files which are published to the players are shown at other places already (e.g. other player, search service results, feeds, …) so if you want to hide them, there is something already wrong with your configuration.

No, not necessarily: Maybe an Adopter did not use "the other player", the "feeds" but uses the search service results for other purposes. That seems a valid configuration to me.

Summary: The change is perfectly fine but I recommend to mention it as "Breaking Change" as I could break existing setups that depend on the current behaviour.

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -102,8 +102,9 @@ class OpencastToPaellaConverter {
     return source;
   }
 
-  getStreamFromFlavour(episode, mainFlavour) {
-    var currentStream = { sources:{}, preview: '', content: mainFlavour };
+  getStreamFromFlavour(episode, flavour) {
+    var mainFlavour = flavour.split('/')[0],
+        currentStream = { sources:{}, preview: '', content: mainFlavour };
```
karendolan, line 113: @lkiesow , I concur with @staubesv that this removes the possibility to publish different types of tracks to Engage for different types of readers/players. At line 107, the track's published subflavor is ignored and lost. The pull removes the self documenting code about filtering by "*/delivery" and retains assumptions about track flavors and what the "flavour.split('/')[0]" symbolizes. 

That "flavour.split('/')[0]" is more cryptic than before. It would help to add a comment regarding assumptions about flavors and that tracks of any published subflavor are grouped here by their main flavor. i.e. "presenter/abc", "presenter/xyz".

Ideally, at some future point, it'd be nice to configure the flavors that the Paella player accepts. If the configuration is missing or empty, take all subflavors and group them by their mainFlavor ;)

karendolan:
@lkiesow IMHO, it's the fast workflow that should be patched with a tag operation to flavor tracks to be published as */delivery before publishing.

lkiesow:
It's ridiculous to duplicate all of the video files just for publishing them and it doesn't make sense to be able to configure what to publish in the publish-engage operation if you hard-code the flavor here. Additionally, all other engage components work differently and do select arbitrary flavors if they exist. Finally, the fast workflow is a part of Opencast for longer than Paella is ;-P

staubesv:
@karendolan 

> [...] removes the possibility to publish different types of tracks to Engage for different types of readers/players

Yes, but hard-coding the flavor `*/delivery` is not a solution for that but rather just the "current state".

> Ideally, at some future point, it'd be nice to configure the flavors that the Paella player accepts.

That would be the solution (-> missing feature -> not scope of this PR ).

@lkiesow 

As said: The change is perfectly fine but I recommend to mention it as "Breaking Change" as I could break existing setups that depend on the current behaviour.

lkiesow:
I will add that and some additional comments to the code.

docs/guides/admin/docs/releasenotes.md
```diff
@@ -44,6 +44,7 @@ Configuration changes
   controlled by Opencast and the routing through capture agents which existed for historical reasons was just an
   additional source for errors. If you rely on the old behavior, it can be configured in
   `etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg`.
+- The Paella player now respects all tracks published to engage.
```
staubesv: Sorry, I don't want to be super-picky, but could you please append "instead of tracks with the flavor */delivery only" to make this _really_ clear?

lkiesow: done.

staubesv:
@karendolan For me, the PR is fine now. As you mentioned concerns: Do you agree with my _"Mis-using the poorly hard-coded current state ('implementation') for separation of tracks for various clients is not a feature but rather a bug. If someone needs more flexibility here, please contribute a feature that makes the flavors of the tracks to be used by the Paella player configurable"_-argument?

At the end of the day, using all tracks seems less arbitrary than just picking some */delivery.

karendolan:
@staubesv and @lkiesow, Ok

staubesv (approved):

staubesv:
@lkiesow Sorry for the long discussion, but I would be quite annoyed if I updated my Opencast 6.x to  7.x and my player just started to display some tracks it hadn't displayed before (without the upgrade/release notes mentioning anything related).
As we simply don't know whether there is an Adopter out there that would be affected, let's play safe.

miesgre:
@miesgre 

> I didn't tested, but I think this will have problems if the mediapackage contains tracks with the same mainflavor and different subflavor.  For example, If a mediapackage contains multiple `presenter/delivery` and a `presenter/preview`, all will be merged in a single paella stream labeled as `presenter`.

Yes, but presenter/preview is ("usually") not published to engage and won't therefore cause problems as it is not visible to Paella at all.

> Another think with this PR is that *all* tracks will be converted to paella streams. so all flavors will be played at time, and this will cause that the player consume more bandwith.

The assumption of this PR is that only tracks that are supposed to be played in Paella are published to Engage.

> For example, we at UPV have some tracks that are not played, and with old code those tracks does not generate paella streams, but with this PR, those tracks will generate paella streams :(

My concern as reviewer indeed was that some Adopters have that situation. That is why I so much insisted on the change being prominently mentioned in the upgrade guide.

> I think this is the better option. Maybe adding a section in the config.json file to configure which flavors import to paella, or how to import. For example, if the user is using and mobile, the institution can decide to server a side-by-side version or something similar

Absolutely, but this requires an Adopter that actually needs this to build and contribute the solution.

miesgre:
Wow, @staubesv... You, edited my comment instead to reply.. :P

> Yes, but presenter/preview is ("usually") not published to engage and won't therefore cause problems as it is not visible to Paella at all.

Yes, it's not usual to have presenter/preview + presenter/delivery... but Maybe the institution have various presenter/* for some reason. I don't like the idea to join all presenter/* in the same stream, but it can be ok if this change is documented.

 
> > Another think with this PR is that _all_ tracks will be converted to paella streams. so all flavors will be played at time, and this will cause that the player consume more bandwith.
> 
> The assumption of this PR is that only tracks that are supposed to be played in Paella are published to Engage.

Ok, so... At an institution I want to have some videos presenter/delivery, presentation/delivery that users can play, but also I want to have some other videos that should not be played by the player, but that can be downloaded by the users. If those other "downloadable tracks" are sopposed not to be published to the engage? How the users can download them? where I'm supposed to publish those files? 

> 
> > For example, we at UPV have some tracks that are not played, and with old code those tracks does not generate paella streams, but with this PR, those tracks will generate paella streams :(
> 
> My concern as reviewer indeed was that some Adopters have that situation. That is why I so much insisted on the change being prominently mentioned in the upgrade guide.

The problem here is that We publish (in addition to */delivery tracks) three more tracks that users can download, but are not played by the player.
So, with old code (with a dual video) two tracks are played at time (presenter/delivery and presentation/delivery), but with this PR 5 tracks are played with the considerable increase in bandwidth 

> 
> > I think this is the better option. Maybe adding a section in the config.json file to configure which flavors import to paella, or how to import. For example, if the user is using and mobile, the institution can decide to server a side-by-side version or something similar
> 
> Absolutely, but this requires an Adopter that actually needs this to build and contribute the solution.

Ok with that. But meanwhile....
The old code only plays the */delivery tracks. If the intention of this PR is that paella plays also the */preview, maybe a better idea is just add */preview to the tracks that can be played instead to add */*

staubesv:
@miesgre 

> Wow, @staubesv... You, edited my comment instead to reply.. :P

Sorry, that was not intended.

@lkiesow Hmm... I think we should revert this change. What do you think?

Btw. While */delivery is somewhat arbitrary, the same holds for all other media package elements identified by some hard-coded flavor (e.g. */player+preview).

karendolan:
@lkiesow @staubesv , If you continue to remove the common understanding of flavors, such as "*/player+preview", as is done here for the ~arbitrary  "*/delivery", it will increase flexibility, but decrease interoperability.

staubesv:
@karendolan Considering the flavor /delivery, there seems not to be a common understanding as the Theodul Player does not look for /delivery while Paella did. But I agree with the flexibility vs. interoperability argument.

lkiesow:
> If you continue to remove the common understanding of flavors […] as is done here for the ~arbitrary "_/delivery", it will increase flexibility, but decrease interoperability.

True but there is clearly no common understanding for `*/delivery` as you might have noticed by now. As far as I know, the first code that ever made this assumption in Opencast is from February last year. That's Paella. So you have over 9 years of being completely configurable and a lot of modules which handle it as if it is configurable versus one very recent module that assumes it to be hard-coded and which doesn't even work with Opencast's default configuration. Sorry, but that doesn't seem right.

Anyway, I've submitted a patch adding a configurable filter for tracks which should resolve this issue.

karendolan:
> > If you continue to remove the common understanding of flavors […] as is done here for the ~arbitrary "_/delivery", it will increase flexibility, but decrease interoperability.
> 
> True but there is clearly no common understanding for `*/delivery` as you might have noticed by now. As far as I know, the first code that ever made this assumption in Opencast is from February last year. That's Paella. So you have over 9 years of being .....

4 years

https://github.com/opencast/opencast/commit/664492e0d0e9d5b8e95118b3b4db8d6fedf42e39#diff-e5d98d1b17e9bd11ab81f69c7cda0553

miesgre:
@lkiesow 

You're right, Paella should play all the flavors or at least a configurable list of them.

I just indicated that this change could cause problems. In fact, it will in my institution. Maybe, it can also cause problems in other institutions and the release of 7.x is quite close...

I've been looking at theodul code and theodul filter the tracks by `Tag` and `format`
https://github.com/opencast/opencast/blob/db8f97a4d5ba21dfb89add6cadf572430cc4780b/modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js#L2194-L2196

Filter by tag could be a valid solution.

PR #898, opened 2019-05-14
MH-13555, Fix Spelling of Flavor

lkiesow:
We shouldn't use flavour except for the British translation, of course.


staubesv (approved):
Although it does not actually fix the spelling of "Flavor" but rather ensures consistent use of American English ;-)

PR #883, opened 2019-05-10
MH-13541 Remove inclusion of non-existent scripts in Admin UI

staubesv:
When starting the Admin UI with "grunt serve" or "grunt proxy", there are 404 NOT FOUND errors reported in the browser console.

Analysis:
The scripts referenced in index.html indeed to not exists.

PR #899, opened 2019-05-15
MH-13556, Fix Login Page

lkiesow:
Unfortunately, pull request #883 did remove a few necessary libraries as
well which completely broke the log-in page making it impossible for
anyone to log-in.

### DIFF
--- a/modules/admin-ui/src/main/webapp/login.html
+++ b/modules/admin-ui/src/main/webapp/login.html
@@ -84,6 +84,10 @@
     <!-- endbower -->
     <!-- endbuild -->
 
+    <!-- build:js(src/main/webapp) scripts/vendor.js -->
+    <script src="scripts/lib/chosen.js"></script>
+    <!-- endbuild -->
+
     <!-- build:js(src/main/webapp) scripts/scripts.js -->
     <script src="scripts/app.js"></script>
     <script src="scripts/shared/controllers/controllers.js"></script>


lkiesow:
That was possibly the fastest merge we ever had 🙀
Thanks!

modules/admin-ui/src/main/webapp/login.html
```diff
@@ -84,6 +84,10 @@
     <!-- endbower -->
     <!-- endbuild -->
 
+    <!-- build:js(src/main/webapp) scripts/vendor.js -->
+    <script src="scripts/lib/chosen.js"></script>
```
staubesv, line 88: @lkiesow & @gregorydlogan Sorry for breaking the login page... :-(

I've removed that because the included file does not actually exist. Do you know why this is required?

My bet is that this leads to the inclusion of scripts/vendor.js that is generated for the index page (even if there is no chosen.js). IIRC, there is a weird dependency between those pages (login/index) that makes such weird stuff required. But that is just a bet.

lkiesow: ` <!-- build:js(src/main/webapp) scripts/vendor.js -->` should include the file. I assumed that we need chosen since the error was chosen related but maybe we could actually remove the second line.

staubesv: @lkiesow As the file scripts/lib/chosen.js simply does not exist, I think what happens is the vendor.js is generated for index.html page and does contain all libraries that page requires. The weird dependency from login.html to index.html then is that the login page depends on vendor.js generated for index.html (which includes some dependencies also required by the login page) rather than including its own dependencies. 

Anyway, it does work ;-)

PR #900, opened 2019-05-16
MH-13558, Paella Track Filter

lkiesow:
To resolve the discussion about track filtering from pull request #896,
this patch adds a configurable track filter.

### DIFF
--- a/etc/ui-config/mh_default_org/paella/config.json
+++ b/etc/ui-config/mh_default_org/paella/config.json
@@ -119,6 +119,9 @@
     },
     "//**** For a list of available plugins and configuration, go to": "https://github.com/polimediaupv/paella/blob/master/doc/plugins.md",
     "list": {
+      "es.upv.paella.opencast.loader": {
+        "flavor": "*/*"
+      },
       "//**** Video profile plugins": "",
       "es.upv.paella.singleStreamProfilePlugin": {
           "enabled": true,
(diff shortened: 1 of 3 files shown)


etc/ui-config/mh_default_org/paella/config.json
```diff
@@ -119,6 +119,9 @@
     },
     "//**** For a list of available plugins and configuration, go to": "https://github.com/polimediaupv/paella/blob/master/doc/plugins.md",
     "list": {
+      "es.upv.paella.opencast.loader": {
+        "flavor": "*/*"
```
miesgre: Maybe it's better to have a list of accepted flavors

lkiesow: done.

etc/ui-config/mh_default_org/paella/config.json
```diff
@@ -119,6 +119,11 @@
     },
     "//**** For a list of available plugins and configuration, go to": "https://github.com/polimediaupv/paella/blob/master/doc/plugins.md",
     "list": {
+      "es.upv.paella.opencast.loader": {
+        "flavors": [
+          ["*", "*"]
```
miesgre, line 124: Why Do you choose this notation instead of "*/*" to be consistent with other workflow configuration files?
I think it is more readable 
```
flavors:["*/*"]
```
than
```
flavors:[["*", "*"]]
```

What do you think?

lkiesow: It's essentially the same and the array variant is less to parse.

staubesv:
@lkiesow Please don't forget to adjust the upgrade documentation ;-)

miesgre (approved):
It's fine to me

PR #901, opened 2019-05-16
MH-13559, Disable User Tracking By Default

lkiesow:
User tracking should be a deliberate choice and not enabled by default
since it may have legal implications.

### DIFF
--- a/etc/ui-config/mh_default_org/paella/config.json
+++ b/etc/ui-config/mh_default_org/paella/config.json
@@ -348,7 +348,7 @@
         "enabled": true
       },
       "es.upv.paella.opencast.userTrackingSaverPlugIn": {
-        "enabled": true
+        "enabled": false
       }
     }
   }


etc/ui-config/mh_default_org/paella/config.json
```diff
@@ -348,7 +348,7 @@
         "enabled": true
       },
       "es.upv.paella.opencast.userTrackingSaverPlugIn": {
-        "enabled": true
+        "enabled": false
```
miesgre, line 351: This is only indicating that the plugin is loaded, but the plugin checks that the user tracking is enabled using the rest endpoint `/usertracking/detailenabled` as documentation says.

https://github.com/opencast/opencast/blob/a6d60b86d71f6ed55f19d860198c80188315c9f5/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.userTrackingSaverPlugIn/mh_usertracking_saver.js#L25-L39

Paella is using the value of the `org.opencastproject.usertracking.detailedtrack` property of the `etc/org.opencastproject.usertracking.impl.UserTrackingServiceImpl.cfg` file.

Maybe it's better to disable it in `etc/org.opencastproject.usertracking.impl.UserTrackingServiceImpl.cfg` file.
https://github.com/opencast/opencast/blob/db8f97a4d5ba21dfb89add6cadf572430cc4780b/etc/org.opencastproject.usertracking.impl.UserTrackingServiceImpl.cfg#L1-L5

If it is disabled by default, please modify also the documentation:
https://github.com/opencast/opencast/blob/db8f97a4d5ba21dfb89add6cadf572430cc4780b/docs/guides/admin/docs/configuration/user-statistics.and.privacy.md#L19-L21

miesgre:
Please note that if this is disabled by default, the views count will not work (count will be 0).

smarquard:
The "legal implications" (in which jurisdictions?) is presumably about the fact that the user info (username and/or ip) is recorded, which is not necessary to establish view count.

So to avoid losing a feature (view count) or switching off a feature by default (new adopters may never discover it), it would be better to report playback activity anonymously rather than not report it at all, preserving the view count.

lkiesow:
It seems like IP's and usernames are not stored by default but only the views, so this should be fine. I'm closing this.

PR #902, opened 2019-05-17
MH-13560, Admin Role in Moodle User Provider

lkiesow:
The Moodle user provider assumed the username of the global admin user
to always be `admin` but the name is configurable.

### DIFF
--- a/modules/userdirectory-moodle/src/main/java/org/opencastproject/userdirectory/moodle/MoodleUserProviderInstance.java
+++ b/modules/userdirectory-moodle/src/main/java/org/opencastproject/userdirectory/moodle/MoodleUserProviderInstance.java
@@ -29,6 +29,7 @@ import org.opencastproject.security.api.JaxbUser;
 import org.opencastproject.security.api.Organization;
 import org.opencastproject.security.api.Role;
 import org.opencastproject.security.api.RoleProvider;
+import org.opencastproject.security.api.SecurityConstants;
 import org.opencastproject.security.api.User;
 import org.opencastproject.security.api.UserProvider;
 import org.opencastproject.userdirectory.moodle.MoodleWebService.CoreUserGetUserByFieldFilters;
@@ -39,6 +40,7 @@ import com.google.common.cache.LoadingCache;
 import com.google.common.util.concurrent.ExecutionError;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


modules/userdirectory-moodle/src/main/java/org/opencastproject/userdirectory/moodle/MoodleUserProviderInstance.java
```diff
@@ -142,26 +145,35 @@
...
-          String coursePattern, String userPattern, String groupPattern, boolean groupRoles, int cacheSize, int cacheExpiration) {
+          String coursePattern, String userPattern, String groupPattern, boolean groupRoles, int cacheSize,
+          int cacheExpiration, String adminUserName) {
     this.client = client;
     this.organization = organization;
     this.groupRoles = groupRoles;
     this.coursePattern = coursePattern;
     this.userPattern = userPattern;
     this.groupPattern = groupPattern;
+    this.ignoredUsernames = Arrays.asList("", SecurityConstants.GLOBAL_ANONYMOUS_USERNAME);
```
smarquard: It really doesn't make sense for a user directory provider to be called with either an empty string or the anonymous username, so this should be fixed higher in the stack than requiring the UDPs to check for these.

lkiesow: Can you be more specific? Where do you want to put these?

smarquard: I was thinking here:

https://github.com/opencast/opencast/blob/develop/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserAndRoleDirectoryServiceImpl.java#L306

lkiesow: But that's not the only place were `loadUser(…)` is used <strike>and more important than that, wouldn't that also block the real admin user?</strike>

---

EDIT: I just realized that you didn't talk about admin. But there are still other places which call `loadUser(…)` and it's a public interface anyone can use, so it's hard to hide.

smarquard: There's UserDirectoryService.loadUser() and UserProvider.loadUser(). We just need to catch all the places that call UserProvider.loadUser(). UserProviders register themselves with the UDS, so no other code really has any requirement to call a UserProvider directly (except Admin UI code and one or two other places call jpaUserAndRoleProvider directly).

The other place I can see to check this is https://github.com/opencast/opencast/blob/develop/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserAndRoleDirectoryServiceImpl.java#L280

Also somewhat related is this change to the Sakai provider that other providers would benefit from: https://github.com/opencast/opencast/commit/7f82bc9015812b4b27dc4a0be3869dfb5d7aa2f9#diff-60c9d09bd8177ec3e9278b428ef3cfb0 otherwise the providers can end up looking up full names like "Firstname Lastname" as a result of how the Admin UI is dealing with presenter metadata. I wonder if that's also something that should be caught higher up, if we consider it invalid to have a username with a space in it.

lkiesow: I gave it some thoughts and I am reluctant to make that change and will stay with the current patch for now. Reasoning:

1. As you noticed, it's easy to catch the major cases but hard to catch them all if you leave it to the caller. At any time, someone could use the user providers directly and forget about the special cases.
2. In most cases, the caller could not do all the filtering. For example, most user providers should still filter the admin user. Something only they can do. This means that you end up with filtering in multiple places for each user provider.
3. User providers should, in the end, be able to decide for themselves which usernames to handle and which they shouldn't. Some providers may even decide to decline much more than only these three.
4. Pulling it somewhere else is probably a riskier patch than this is and this is a bug-fix patch for an old version after all.

modules/userdirectory-moodle/src/main/java/org/opencastproject/userdirectory/moodle/MoodleUserProviderInstance.java
```diff
@@ -142,26 +145,35 @@
...
+    if (StringUtils.isNoneEmpty(adminUserName)) {
+      ignoredUsernames.add(adminUserName);
```
mtneug, line 178: This throws an `UnsupportedOperationException` as `Arrays#asList` creates an immutable list.

mtneug (changes requested):
Apart from the immutable list issue this change works as expected.

modules/userdirectory-moodle/src/main/java/org/opencastproject/userdirectory/moodle/MoodleUserProviderInstance.java
lkiesow: Good catch. Fixed.

PR #907, opened 2019-05-19
MH-13375 Handle empty-range errors correctly

smarquard:
This addresses server-side issues when dealing with conflict checks and event creation attempts for an empty range, where the recurrence rule and the start/end dates do not include any events.

For example, start date = end date and the recurrence rule is only for a different day of the week than the start and end date.

The REST endpoints now return appropriate responses: empty response for conflict check (because there are no conflicts because there are no events to conflict with), and bad request to attempt to create an event, because no event was created because there were no valid dates provided.

Previously both conflict and event endpoints returned 500 server errors, with stack traces in the logs.

The Admin UI behaviour is unchanged and not optimal, and that is left to another JIRA/PR to resolve. The event creation wizard will run through and complete finally with a UI warning that the event could not be created. It would be better to adjust the wizard to detect this case earlier.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -2024,6 +2024,9 @@ public abstract class AbstractEventEndpoint {
   public Response createNewEvent(@Context HttpServletRequest request) {
     try {
       String result = getIndexService().createEvent(request);
+      if (StringUtils.isEmpty(result)) {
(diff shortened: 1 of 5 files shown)


smarquard:
Sample scheduling parameter to test this with the external API POST /api/events endpoint:

{"agent_id":"devca","duration":"3300000","end":"2019-05-24T06:55:00Z","inputs":["default"],"rrule":"FREQ=WEEKLY;BYDAY=TH;BYHOUR=6;BYMINUTE=0","start":"2019-05-24T06:00:00Z"}

The admin-ng endpoint can be tested via the UI.

Note that duration must be provided as a string value which is another bug: https://opencast.jira.com/browse/MH-13566

staubesv (approved):

PR #910, opened 2019-05-20
MH-13566 Accept duration as either string or number in scheduling JSON

smarquard:
The implementation released in 6.0 requires a string duration, but
the docs say Integer, which is consistent with usage elsewhere.

So accept both formats, enforce presence of duration when rrule is
specified, and update the docs to show correct examples.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/util/SchedulingUtils.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/util/SchedulingUtils.java
@@ -60,6 +60,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.TimeZone;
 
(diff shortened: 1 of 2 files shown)


modules/external-api/src/main/java/org/opencastproject/external/util/SchedulingUtils.java
```diff
@@ -256,10 +256,20 @@ public static SchedulingInfo of(JSONObject json) {
...
       final String startDate = (String) json.get(JSON_KEY_START_DATE);
       final String endDate = (String) json.get(JSON_KEY_END_DATE);
-      final String durationString = (String) json.get(JSON_KEY_DURATION);
       final String agentId = (String) json.get(JSON_KEY_AGENT_ID);
       final JSONArray inputs = (JSONArray) json.get(JSON_KEY_INPUTS);
       final String rrule = (String) json.get(JSON_KEY_RRULE);
+
+      // Special handling because the original implementation required String but now we require long
+      final Object durationObj = json.get(JSON_KEY_DURATION);
+      String durationString = null;
```
lkiesow: Instead of checking for the object type and handling things differently based on the typüe, couldn't you just use

```java
final String durationString = String.valueOf(durationObj);
```
That should always work, regardless of the type being a String, a Long, a Float or whatever else.

lkiesow: EDIT: To handle `null` properly, this whould be better:

```java
… = Objects.toString(json.get(JSON_KEY_DURATION), null);
```

smarquard: That is indeed more elegant, thank you.

smarquard: also fixed a cosmetic double quote in the message.

lkiesow:
Looks okay → merged.
Though the overall state of the external API seems horrible and it easier to break something than it is not to.
But that's not the problem of this pull request ;-)

staubesv:
> Though the overall state of the external API seems horrible and it easier to break something than it is not to.

@lkiesow Just wondering: Why do you think the overall state is horrible?

lkiesow:
Because testing this patch I found multiple ways to break something generating long stack traces and internal server errors by providing just slightly modified inputs. The state seems to be that the API mostly does what is expected if you provide precisely the right input. But no one seems to have spared any remote thought about properly handling incorrect inputs… that's not what I would call a good state.

PR #911, opened 2019-05-20
MH-13568 Catch exception from overlapping RRule and return bad request

smarquard:
Clients can submit any RRule and duration to the external API POST /api/event endpoint.
If the RRule and duration specify overlapping events, an exception is thrown.

Here we catch this exception and return a bad request rather than allow it to be propagated
producing a 500 internal server error response.

The exception type is changed to IllegalArgumentException because it is an illegal argument,
and we can catch it specifically rather than converting all SchedulerExceptions to bad request
responses.


KatrinIhler (approved):
Look's good!

staubesv (approved):

PR #912, opened 2019-05-21
MH-13425 Feed Series Information Tab

ek926m:
This is a follow up of #845 . It contains a new tab with some feed Information


modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/serieController.js
```diff
@@ -22,11 +22,20 @@
...
+  'ResourcesListResource', 'UserRolesResource', 'Notifications', 'AuthService',
+  'StatisticsReusable','$http',
+  function ($scope, SeriesMetadataResource, SeriesEventsResource,
+    SeriesAccessResource, SeriesThemeResource, ResourcesListResource,
+    UserRolesResource, Notifications, AuthService, StatisticsReusable, $http) {
+
+    $http.get('/admin-ng/feeds/feeds')
+    .then( function(response) {
+      $scope.feedContent = response.data;
+    }, function(error) {
```
lkiesow: iirc, the new(er) AngularJS requires:

```suggestion
    }).catch(function(error) {
```

lkiesow (changes requested):
It seems the file `.factorypath` got committed by accident

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
```diff
@@ -0,0 +1,70 @@
...
+    super("org.opencastproject.feed.impl.FeedServiceImpl");
+  }
+
+  /** Logging facility */
+  private static Logger logger = LoggerFactory.getLogger(FeedEndpoint.class);
+
+  @GET
+  @Path("/feeds")
+  @Produces(MediaType.APPLICATION_JSON)
+  public String listFeedServices() {
```
gregorydlogan, line 54: This method just wraps a GET for the correct endpoint?  Why aren't you using internal service bindings instead of using this endpoint?

Edit: Ah, you're fetching the search feeds.  Then isn't the relative URL incorrect?  Shouldn't it be `/feeds/feeds`?

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/serieController.js
```diff
@@ -22,11 +22,20 @@
 
 // Controller for all single series screens.
 angular.module('adminNg.controllers')
-.controller('SerieCtrl', ['$scope', 'SeriesMetadataResource', 'SeriesEventsResource', 'SeriesAccessResource',
-  'SeriesThemeResource', 'ResourcesListResource', 'UserRolesResource', 'Notifications', 'AuthService',
-  'StatisticsReusable',
-  function ($scope, SeriesMetadataResource, SeriesEventsResource, SeriesAccessResource, SeriesThemeResource,
-    ResourcesListResource, UserRolesResource, Notifications, AuthService, StatisticsReusable) {
+.controller('SerieCtrl', ['$scope', 'SeriesMetadataResource',
+  'SeriesEventsResource', 'SeriesAccessResource', 'SeriesThemeResource',
```
gregorydlogan, line 26: As much as the previous formatting is horrible, I would avoid reformatting it...

gregorydlogan (changes requested):
I haven't actually had time to build this yet, but it looks good aside from a couple of questions.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
ek926m: I think it is already /feeds/feeds, it gets the base Path above @Path("/feeds) too.

http://localhost:8080/feeds/feeds

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/serieController.js
ek926m: I had to. The maxiumum length of 120 characters was reached in Eslint.

ek926m:
@gregorydlogan is the review done? could this be merged?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/FeedEndpoint.java
gregorydlogan: Generated paths in the tab are of the form localhost:8080/(event_video/aggregated) etc.  These all 404.  Manually adding admin-ng, or admin-ng/feeds in between the feed type and the port yields a 404 as well.  Possibly related to the issue I'm reporting in #845.

gregorydlogan:
Nope, this doesn't appear to be functional to me.  The series feeds tab has URLs like `http://localhost:8080/latest`, rather than `http://localhost:8080/feeds/rss/2/latest` or `http://localhost:8080/admin-ng/feeds/rss/2/latest`.

The tab's header is also not spaced right.  Looks like one of the ngIf elements is somehow still taking up space.

gregorydlogan:
@ek926m any idea when this will be moved forwards?

ek926m:
@gregorydlogan 

I removed the table header due to duplication.

now my tab view looks like this: is it spaced right?

Bildschirmfoto 2019-07-11 um 15 09 16

ek926m:
> Nope, this doesn't appear to be functional to me. The series feeds tab has URLs like `http://localhost:8080/latest`, rather than `http://localhost:8080/feeds/rss/2/latest` or `http://localhost:8080/admin-ng/feeds/rss/2/latest`.

what do you exactly mean?

gregorydlogan:
@ek926m See the space between `Branding` and `Feeds`?  It should be the same spacing as the space between `Zugriffsrechte` and `Branding`.

Regarding the functionality: The `Identifier` column contains a number of urls, all of which 404 by default.  This should be fixed in #845, then this PR would be rebased onto that branch, and then we should be good to go.

ek926m:
> @ek926m See the space between `Branding` and `Feeds`? It should be the same spacing as the space between `Zugriffsrechte` and `Branding`.
> 
> Regarding the functionality: The `Identifier` column contains a number of urls, all of which 404 by default. This should be fixed in #845, then this PR would be rebased onto that branch, and then we should be good to go.

> @ek926m See the space between `Branding` and `Feeds`? It should be the same spacing as the space between `Zugriffsrechte` and `Branding`.
> 
> Regarding the functionality: The `Identifier` column contains a number of urls, all of which 404 by default. This should be fixed in #845, then this PR would be rebased onto that branch, and then we should be good to go.

100x50 on all tab headers, seems right to me.

Bildschirmfoto 2019-07-18 um 13 53 10
Bildschirmfoto 2019-07-18 um 13 53 07

ek926m:
@gregorydlogan pls look at #1050, this is a updated one from this

PR #913, opened 2019-05-21
MH-13569 Change of PlayerRedirection variable from {{id}} to #{id}

mliradelc:
This PR is created to modify the part from the organization configuration file, where the event id variable for the player redirection is actually {{id}}, this definition can create problem with Provisioning and orchestration software like Ansible.

The new definition is #{id} to avoid this problem.

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -66,6 +66,8 @@ Configuration Changes
   option.
 - `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` has a new option `maintenance` which temporarily
   disables the scheduler if set to `true`.
+- `etc/org.opencastproject.organization-mh_default_org.cfg` the variable `{{id}}` 
+  from `prop.player` has changed to `#{id}`.
 
 Scheduler Migration
 -------------------
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -66,8 +66,7 @@ Configuration Changes
   option.
 - `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` has a new option `maintenance` which temporarily
   disables the scheduler if set to `true`.
-- `etc/org.opencastproject.organization-mh_default_org.cfg` the variable `{{id}}` 
-  from `prop.player` has changed to `#{id}`.
+
 
 Scheduler Migration
 -------------------
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -225,11 +225,11 @@ prop.adminui.user.external_role_display=false
 # Path to the default video player used by the engage interfaces and the LTI tools. Parameters for selecting the
 # specific videos will be appended automatically.  Common values include:
 #
-# - theodul player: /engage/theodul/ui/core.html?id={{id}}
-# - paella  player: /paella/ui/watch.html?id={{id}}
+# - theodul player: /engage/theodul/ui/core.html?id=#{id}
+# - paella  player: /paella/ui/watch.html?id=#{id}
 #
-# Default: /engage/theodul/ui/core.html?id={{id}}
-#prop.player=/engage/theodul/ui/core.html?id={{id}}
+# Default: /engage/theodul/ui/core.html?id=#{id}
+#prop.player=/engage/theodul/ui/core.html?id=#{id}
 
 # The default flavor of the master video (the video on the "left side" in the video display)
 prop.player.mastervideotype=presenter/delivery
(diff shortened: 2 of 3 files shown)


staubesv (changes requested):
Can you please remove the changes to all package-lock.json files?

docs/guides/admin/docs/releasenotes.md
```diff
@@ -51,6 +51,7 @@ Configuration changes
     - `etc/org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider.cfg`
     - `etc/org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg`
     - `etc/org.opencastproject.security.urlsigning.verifier.impl.UrlSigningVerifierImpl.cfg`
+- In the `org.opencastproject.organization-mh_default_org.cfg` file change the variable of the events ID in the player properties (prop.player) to avoid issues with Ansible or other type of deployment software.
```
staubesv: Maybe something like:

Due to [MH-13569](https://opencast.jira.com/browse/MH-13569) the value of the property `prop.player` in`org.opencastproject.organization-mh_default_org.cfg`needs to be adjusted 

Idea: You don't need to explain here why this has been done or how exactly it is done. Rather just indicate that the value prop.player has been adjusted.

staubesv:
Please also fix

`./admin/docs/releasenotes.md: 54: MD013/line-length Line length [Expected: 120; Actual: 210]`

docs/guides/admin/docs/releasenotes.md
```diff
@@ -51,6 +51,7 @@ Configuration changes
     - `etc/org.opencastproject.security.urlsigning.provider.impl.GenericUrlSigningProvider.cfg`
     - `etc/org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg`
     - `etc/org.opencastproject.security.urlsigning.verifier.impl.UrlSigningVerifierImpl.cfg`
+- Due [MH-13569](https://opencast.jira.com/browse/MH-13569), the variable in prop.player has been changed.
```
lkiesow: Can you remove this line? This is nothing that has changed between releases and is only a modification within the release. From a release point of view `{{id}}` never existed.

lkiesow (changes requested):
The upgrade guide has a section about the new player URL format and possible migration work which needs to be adapted.

docs/guides/admin/docs/releasenotes.md
mliradelc: No problem, but where I left the comment of the change?  Because people who is using scripts to install can get their installation fail if we don't notice about this.

lkiesow: You don't need to. The changelog is generated automatically.

mliradelc: I see, Done!

mliradelc:
> Can you please remove the changes to all package-lock.json files?

Done!

staubesv:
@lkiesow Seems we both assigned the review to ourselves ;-) As far as I see, the only open point now is:

> The upgrade guide has a section about the new player URL format and possible migration work which needs to be adapted.

Is that correct?

lkiesow:
yep: https://github.com/opencast/opencast/blob/7d1aeaad6ea4e3ad49d58148cad1c421d24a1806/docs/guides/admin/docs/upgrade.md#update-player-links

mliradelc:
> The upgrade guide has a section about the new player URL format and possible migration work which needs to be adapted.

I pushed an indication in the upgrade guide.

docs/guides/admin/docs/upgrade.md
```diff
@@ -66,6 +66,8 @@ Configuration Changes
   option.
 - `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` has a new option `maintenance` which temporarily
   disables the scheduler if set to `true`.
+- `etc/org.opencastproject.organization-mh_default_org.cfg` the variable `{{id}}` 
+  from `prop.player` has changed to `#{id}`.
```
lkiesow: As mentioned already, please remove this. Looking at releases, the `{{id}}` syntax never existed in **any** release. Hence, no update is required. <strike>However, there is the matter of upgrading from the old, hard-coded URLs for which this file does contain an upgrade section which needs to be adapted.</strike>

*EDIT: Realized that that section is still fine as it is, so just removing this should do.*

mliradelc: But what about the people who use the packages from version 7.0 if they upgrade to an upcoming version 7.1?

lkiesow: If I merge this now, the 7.0 package already has the format `#{id}`. No version ever used `{{id}}`.

mliradelc: Of course! You have right!, Ok I will delete that, even I am the release manager and we haven't release this version 😋!!

I will delete that line right now.

lkiesow (approved):
LGTM. @staubesv, any objections to merging this?

staubesv (approved):

staubesv:
@lkiesow No, fine for me.

docs/guides/admin/docs/upgrade.md
```diff
@@ -67,6 +67,7 @@ Configuration Changes
 - `etc/org.opencastproject.scheduler.impl.SchedulerServiceImpl.cfg` has a new option `maintenance` which temporarily
   disables the scheduler if set to `true`.
 
+
```
staubesv, line 70: @mliradelc & @lkiesow although it would be desirable to not add a new line to upgrade.md...

PR #915, opened 2019-05-22
MH-13574 Code reviews should ensure configuration changes to be documented

staubesv:
Undocumented changes to configuration files can cause problems when Adopters upgrade to newer versions of Opencast in case they use their own customized copies of configuration files.

To make the upgrade as simple as possible and avoid Adopters running into troubles, it would be desirable if the upgrade guide would contain a complete list of any configuration file change including references to relevant JIRA issues that document the reasons for the change.

### DIFF
--- a/docs/guides/developer/docs/reviewing-and-merging.md
+++ b/docs/guides/developer/docs/reviewing-and-merging.md
@@ -34,7 +34,8 @@ Reviewing Rules
 Reviewing a Pull Request
 ------------------------
 
-Our goals in doing code reviews for all pull requests are to ensure that:
+There is no list of things you are required to do as reviewer of a pull request. Our primary goals in doing reviews for
+all pull requests are to ensure that:
 
 * There are no bugs in the pull request
     * The feature works as advertised
@@ -46,6 +47,8 @@ Our goals in doing code reviews for all pull requests are to ensure that:
 * Adopters can easily upgrade to new versions
     * An upgrade path exists
     * The upgrade path is documented
+
+The exact review process heavily depends on the type, size and purpose of the pull request.
 
 Here are some things a reviewer should usually do:
 
@@ -59,3 +62,10 @@ Here are some things a reviewer should usually do:
   configuration. Such configuration changes need to be documented in the upgrade guide including a reference to
   the Jira ticket(s) that are relevant for the changes
 * Check the documentation
+
+Some changes require special attention:
+
+Folder                         | Description
+:------------------------------|------------
+etc/listproviders              | Changes here might need to be reflected in the static mockup data for the Admin UI facade found in modules/admin-ui/src/test/resources/app/admin-ng/resources
+modules/admin-ui/src/main/java | In case the interface of the Admin UI facade changes, those changes need to be also reflected in the static mockup data for the Admin UI facade found in modules/admin-ui/src/test/resources/app.
--- a/docs/guides/developer/docs/reviewing-and-merging.md
+++ b/docs/guides/developer/docs/reviewing-and-merging.md
@@ -34,8 +34,7 @@ Reviewing Rules
 Reviewing a Pull Request
 ------------------------
 
-There is no list of things you are required to do as reviewer of a pull request. Our primary goals in doing reviews for
-all pull requests are to ensure that:
+Our goals in doing code reviews for all pull requests are to ensure that:
 
 * There are no bugs in the pull request
     * The feature works as advertised
@@ -44,21 +43,19 @@ all pull requests are to ensure that:
     * [Licenses seem to be fine](license.md)
     * Licenses are added to the `NOTICES` file
 * Feature documentation is in place
-* An upgrade path exists (usually relevant for database changes)
-
-
-The exact review process heavily depends on the type, size and purpose of the pull request.
+* Adopters can easily upgrade to new versions
+    * An upgrade path exists
+    * The upgrade path is documented
 
 Here are some things a reviewer should usually do:
 
 * Check if code looks sensible (e.g. no nonsensical files, no obvious formatting errors, no large binary files, …)
 * Locally run Opencast and verify that the issue has been fixed (as described in the pull request and/or the Jira
   ticket) or the feature does what it is supposed
+* Check whether the database schema is modified. In case it is, a database migration scripts needs to be provided
+  and the upgrade guide needs to contain the information that a database migration is needed including a reference
+  to the Jira ticket(s) the cause the migration
+* Check whether configuration files have been changed in a way that potentially requires Adopters to adjust their
+  configuration. Such configuration changes need to be documented in the upgrade guide including a reference to
+  the Jira ticket(s) that are relevant for the changes
 * Check the documentation
-
-Some changes require special attention:
-
-Folder                         | Description
-:------------------------------|------------
-etc/listproviders              | Changes here might need to be reflected in the static mockup data for the Admin UI facade found in modules/admin-ui/src/test/resources/app/admin-ng/resources
-modules/admin-ui/src/main/java | In case the interface of the Admin UI facade changes, those changes need to be also reflected in the static mockup data for the Admin UI facade found in modules/admin-ui/src/test/resources/app.


docs/guides/developer/docs/reviewing-and-merging.md
```diff
@@ -44,21 +43,19 @@ all pull requests are to ensure that:
...
   ticket) or the feature does what it is supposed
+* Check whether the database schema is modified. In case it is, a database migration scripts needs to be provided
+  and the upgrade guide needs to contain the information that a database migration is needed including a reference
+  to the Jira ticket(s) the cause the migration
+* Check whether configuration files have been changed in a way that potentially requires Adopters to adjust their
+  configuration. Such configuration changes need to be documented in the upgrade guide including a reference to
+  the Jira ticket(s) that are relevant for the changes
 * Check the documentation
-
-Some changes require special attention:
```
staubesv: IIRC, it was me that added this a while ago but I suggest to remove it because a) nobody takes care and b) no directly relevant for Adopters

docs/guides/developer/docs/reviewing-and-merging.md
```diff
@@ -34,8 +34,7 @@ Reviewing Rules
 Reviewing a Pull Request
 ------------------------
 
-There is no list of things you are required to do as reviewer of a pull request. Our primary goals in doing reviews for
```
staubesv: Streamline: "There is no list of things you are required to do as reviewer of a pull request." really isn't helpful.

docs/guides/developer/docs/reviewing-and-merging.md
```diff
@@ -44,21 +43,19 @@ all pull requests are to ensure that:
     * [Licenses seem to be fine](license.md)
     * Licenses are added to the `NOTICES` file
 * Feature documentation is in place
-* An upgrade path exists (usually relevant for database changes)
-
-
-The exact review process heavily depends on the type, size and purpose of the pull request.
```
staubesv: Streamline: "The exact review process heavily depends on the type, size and purpose of the pull request." really just isn't of any help.

staubesv:
Proposal on list: https://groups.google.com/a/opencast.org/forum/#!topic/dev/-zcjMuRRgyU

docs/guides/developer/docs/reviewing-and-merging.md
lkiesow: The intent behind this sentence was to say that reviewers should first and foremost think about the change and assess if the change makes sense and will not harm (short or long-term) anyone which is not just checking a list. Additional, it also means that as a reviewer, you can be pragmatic and, if it makes sense in that situation, you can ignore all other listed items here. Being pragmatic is sometimes better than just sticking to rules.

docs/guides/developer/docs/reviewing-and-merging.md
lkiesow: Same here: The main purpose is to make the reviewer think about the review and not stupidly follow a list of rules. That's what automated tests are for ;-)

docs/guides/developer/docs/reviewing-and-merging.md
staubesv: Just undone changes that are not directly relevant to the proposal

docs/guides/developer/docs/reviewing-and-merging.md
staubesv: Just undone changes that are not directly relevant to the proposal.

docs/guides/developer/docs/reviewing-and-merging.md
```diff
@@ -44,21 +43,19 @@ all pull requests are to ensure that:
...
+    * The upgrade path is documented
 
 Here are some things a reviewer should usually do:
 
 * Check if code looks sensible (e.g. no nonsensical files, no obvious formatting errors, no large binary files, …)
 * Locally run Opencast and verify that the issue has been fixed (as described in the pull request and/or the Jira
   ticket) or the feature does what it is supposed
+* Check whether the database schema is modified. In case it is, a database migration scripts needs to be provided
+  and the upgrade guide needs to contain the information that a database migration is needed including a reference
+  to the Jira ticket(s) the cause the migration
```
lkiesow, line 60: -1 to just referencing Jira. It depends on the change but mostly, I would expect a description of the upgrade path, not a list of changes. That's part of the changelog. just removing the “including a reference to the Jira ticket(s) the cause the migration” should do.

staubesv: Let's discuss this on list ;-)

docs/guides/developer/docs/reviewing-and-merging.md
```diff
@@ -44,21 +43,19 @@ all pull requests are to ensure that:
...
+* Check whether configuration files have been changed in a way that potentially requires Adopters to adjust their
+  configuration. Such configuration changes need to be documented in the upgrade guide including a reference to
```
lkiesow, line 62: Right now, the documentation is mostly in the [release notes](https://docs.opencast.org/r/7.x/admin/releasenotes/#configuration-changes) which I think makes sense to keep the upgrade guide concise. Also, -1 to linking the Jira tickets. It might make sense but doesn't always, so just leave it out.

staubesv: ... but this information is required when upgrading, so why should it be in the release notes?

-> Discussion should happen on list

lkiesow: change logs and upgrade guides are sub-items of the release notes so
they are in there either way ;) …which makes sense since they are notes
about the release after all.

If you feel strongly about it, we can move it but historically, the
upgrade guide always contained a point which said more or less “compare
the old and new configuration” which is exactly what you need to do and
what you need to know and what I (and from comments on list I think
others as well) are doing. It's much faster than reading though
specific changes with many of them usually being irrelevant.

That's why my tendency would be to keep the detailed list of changes in
the release notes. People can read them if they want to and reference
them if they need to, but they don't clutter the upgrade guide which in
turn means that the important points are less likely to be overlooked.

But I do not have a strong feeling about this and we can put the notes
in the upgrade section if people prefer that. In the end, it's the
release manager's decision anyway how to exactly structure the release
notes and if to copy things to a specific place.

docs/guides/developer/docs/reviewing-and-merging.md
lkiesow: Suggested change:

> If the database schema is modified, check if a database migration script is provided and the upgrade guide contains migration documentation.

gregorydlogan: Very much agree with a descriptive fix, rather than linking to JIRA.  JIRA is where highly technical discussions can take place.  Adopters neither need to see it, nor in many cases even want to know about it.  They only want the quick-hit fix so they an get on with their lives.  I'm fine including a link as well, but it needs to be more in the vein of "for more information see MH-Whatever".

docs/guides/developer/docs/reviewing-and-merging.md
gregorydlogan: IMO, in a prod context, if you're not using a change management system for anything at scale then you're doing it wrong.  I'm sure we have adopters who *do* run everything by hand, but at some point they're going to feel the pain of that regardless of how well we document things.  Comparing their config to the new one is the only viable way for them to know that what they have is right for them.  As Lars said, our docs are going to list a ton of changes which are not relevant - how will they know what's relevant to them, since they're 'just' using the system rather than developing.  This becomes easier with staging environments, but even if they don't have that they'll still need to mark up the new config for their CM system.

staubesv: > IMO, in a prod context, if you're not using a change management system for anything at scale then you're doing it wrong. I'm sure we have adopters who do run everything by hand, but at some point they're going to feel the pain of that regardless of how well we document things

In our case we have an own copy of the full config which is in a separate GIT repository and managed/deployed using Ansible.
I am currently doing diffs between the configs to adjust them but consider this a quite error prone process: There are quite a large number of differences over the overall config between ours and the Opencast default configuration and it is easy to miss subtle differences that would need to be adapted.

> As Lars said, our docs are going to list a ton of changes which are not relevant

I don't think we would end-up a large list of irrelevant changes as there are not that many config changes then upgrading from one version to the next one. And the irrelevant changes also need to be looked at when going the "compare" way.

>  how will they know what's relevant to them, since they're 'just' using the system rather than developing.

The target audience of the upgrade guide is people that are performing the upgrades. Most likely they are also responsible for their Opencast configuration and hopefully know their configuration.

> This becomes easier with staging environments, but even if they don't have that they'll still need to mark up the new config for their CM system.

Imagine 6.x to 7.x had a complete list of configuration changes which are about 10 changes. Applying those to the CM is quite a lot less effort than doing a full diff on the full config which has several hundred differences which each need to be carefully looked at everytime.

docs/guides/developer/docs/reviewing-and-merging.md
staubesv: @gregorydlogan I very much thought of the "for more information see MH-XXXXX". That does make sense as it might sometimes not be clear how the value needs to be adjusted and the issue might be helpful to get more information about the change.
Note that the changelog is not providing this information as the issue titles might not always obviously relate to the configuration change in question.

docs/guides/developer/docs/reviewing-and-merging.md
lkiesow: > I don't think we would end-up a large list of irrelevant changes as there are not that many config changes then upgrading from one version to the next one.

I imagine that there are 10 changes and then suddenly wake up with the realization that it's far away from reality 😸

Between 6 and 7, we have changes in 60 configuration files. `custom.properties` alone has 10 changes in properties. Of course, I can read about all of those changes and then apply them manually but that is very tedious work and I wouldn't do that.

Instead, I would look at the differences which lets me quickly apply those changes and be done with it. Of course, I wouldn't need to apply the change if there is a sensible default, but that's not always guaranteed (although I like that to be the case) and it's hard to keep track of my actual changes if I do not sync with the changes.

Or, depending on how many changes I made in the first place, I might even go with the default configuration and just apply those few changes.

But that may very well just be a difference in upgrade/administration style resulting in you wanting everything and me wanting an abbreviated version in the docs with data changes only.

docs/guides/developer/docs/reviewing-and-merging.md
lkiesow: The values should always have proper documentation right at the key in the documentation file. For me, that's by far the most important part of any configuration change.

But note that I'm not saying that you may not link Jira if it makes sense or a pull request, a discussion on list or any other documentation. If that is helpful, please do! If it's not …don't.

staubesv:
@lkiesow The custom.properties changes from 6.x to 7.x would look like:

[CHANGED] **custom.properties** 
-  Moved _org.opencastproject.inbox.threads_ to file  etc/org.opencastproject.ingest.scanner.InboxScannerService-inbox.cfg ([MH-13365](https://opencast.jira.com/browse/MH-13365))
- Removed  _org.opencastproject.engage.paella.config.folder_ since Paella player configuration is now managed as user interface configuration ([MH-13459](https://opencast.jira.com/browse/MH-13459))
- Added _org.opencastproject.uiconfig.folder_ that specifies the folder containing the user interface configuration ([MH-13413](https://opencast.jira.com/browse/MH-13413))
- Removed _org.opencastproject.elasticsearch.config.dir_ as part of the elastic search configuration update ([MH-13399](https://opencast.jira.com/browse/MH-13399))
- Added  _org.opencastproject.elasticsearch.server.address_ & _org.opencastproject.elasticsearch.server.port_ because of added support for external elastic search servers ([MH-13253](https://opencast.jira.com/browse/MH-13253))
- Changed default value of _org.opencastproject.dataloader.csv_ ([MH-13324](https://opencast.jira.com/browse/MH-13324))
- Fix ActiveMQ default value for _activemq.broker.url_ ([MH-13331](https://opencast.jira.com/browse/MH-13331))

Changes can also be summarized, e.g.

[REMOVED] **/etc/index/\***
- Removed elastic search index configuration which is now included in a bundle (MH-ABCDE)


Note that this is incrementally built up when doing the changes, so not much work per change.


@lkiesow A good example of what I mean by getting more context/information by linking to the JIRA issue is the "...add support for external elastic search servers  ([MH-13253](https://opencast.jira.com/browse/MH-13253))".

Admittedly, it depends quite a lot of the quality of the description (surprisingly good in case of MH-13253).

docs/guides/developer/docs/reviewing-and-merging.md
staubesv: I'm not talking about the documentation of the configuration which I agree should be in the configuration file right were the configuration setting is located.

The reference to the ticket is supposed to provide more context about the configuration change in means of "What was changed in Opencast that caused the configuration to be changed" (not: how do I have to set a value) in case somebody want to dig deeper / have more information.

Always referring to the ticket simply avoids discussions whether a reference to the ticket makes sense for a given case. Any information additionally useful can be referenced from the ticket (rather than the upgrade guide). This includes pull requests, discussions on list and whatever might be relevant.

gregorydlogan: Referring to the ticket is great, requiring that an adopter read a ticket is not.  If the config and the markdown docs don't provide enough explanation of how a value *should* be set, then there needs to be additional work done on those files.  Putting the 'what' into the config and docs is required, the 'why' can be in both, but I'm betting most adopters are more interested in 'what' they should be setting rather than 'why'.

tl;dr: An adopter should never be dropped into JIRA.  It's enough of a mess as it is, and we've got docs which sometimes are not in sync with what's discussed in JIRA.  We don't need another source of truth for adopters.

staubesv: @gregorydlogan 

This example quite nails what I mean by providing more context of a change:

> - Added org.opencastproject.elasticsearch.server.address & org.opencastproject.elasticsearch.server.port because of added support for external elastic search servers([MH-13253](https://opencast.jira.com/browse/MH-13253))

Does not mean that an Adopter has to read it in means of documenting of the configuration (which is part of the configuration file), but it might be interesting for just additional information about the overall change.

gregorydlogan:
As long as the PR authors, and not the reviewers or RMs are responsible for writing the components of that, I guess I would be ok.  I still feel like this will not be as useful as you think it will be for most adopters, but it's also not that much additional work.  RMs would still be responsible for aggregating the files (living somewhere in the docs folder) into the final markdown-formatted list.

staubesv:
> RMs would still be responsible for aggregating the files (living somewhere in the docs folder) into the final markdown-formatted list.

Why? Authors of PRs can edit upgrade.md (or wherever the configuration changes list lives) directly.

lkiesow:
> ...add support for external elastic search servers ([MH-13253](https://opencast.jira.com/browse/MH-13253))

Providing context, even better would be:

> ...add support for external elastic search servers ([00329c30a8](https://github.com/opencast/opencast/commit/00329c30a8))

It will not only give you a proper explanation of the change but additionally the whole change itself so that you can (if you want or need to) dive deeper into the change.

Would I recommend to do this? In general, no. Because, in general, I don't think it's really helpful and it might lead to people just starting to link stuff and neglect a proper explanation. So, I still dislike this the idea. Having people think about what they are doing makes more sense to me.

staubesv:
> ...add support for external elastic search servers (00329c30a8)

The target audience of the upgrade guide is system administrators - not developers. Referring to code changes seems therefore inappropriate to me.

smarquard:
IMO, links to commits are not appropriate (wrong audience, and there could be later commits that change things). Links to JIRAs may be helpful (depends on the quality of the JIRA). A link to the documentation page for how to configure that feature is probably the most helpful.

staubesv:
I'm closing this as I'm not working on this anymore.

PR #917, opened 2019-05-23
MH-13578 Remove obviously unused classes

staubesv:
Basically, I've ran mvn de.is24.mavenplugins:deadcode4j-maven-plugin:find -Dmaven.test.skip=true (see https://github.com/Scout24/deadcode4j) to get a list of classes that tool thinks are not used.

The list mainly consisted of false positives, but going through the list, I could identify a number of classes that really seem not to be used. I've also searched the class names in *.xml files (think of OSGI-INF) as that is a way to reference classes.


lkiesow:
Will go through this list over the next days and double-check. Thanks for the patch, good idea to automatically check this.

PR #919, opened 2019-05-24
MH-13580 Fix dependencies for statistics-* and workflow-condition-parser

staubesv:
This work adds the maven-dependency-plugin to the modules statistics-* and workflow-condition-parser which have been recently added to Opencast. 
It then fixes the dependency issues reported by the plugin.

### DIFF
--- a/modules/statistics-service-remote/pom.xml
+++ b/modules/statistics-service-remote/pom.xml
@@ -16,6 +16,10 @@
   </properties>
   <dependencies>
     <dependency>
+      <groupId>com.googlecode.json-simple</groupId>
+      <artifactId>json-simple</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.opencastproject</groupId>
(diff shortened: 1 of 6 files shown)


modules/statistics-service-remote/pom.xml
```diff
@@ -26,36 +30,39 @@
...
-      <artifactId>jsr311-api</artifactId>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore-osgi</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.osgi</groupId>
-      <artifactId>org.osgi.core</artifactId>
+      <groupId>org.codehaus.jettison</groupId>
+      <artifactId>jettison</artifactId>
+      <version>1.3.1</version>
```
lkiesow, line 43: This should get central dependency management but I submitted pull request #924 for that.

PR #922, opened 2019-05-27
MH-13582 Support for exclusion pattern for URL signing

staubesv:
While stream security allows to protect files on Opencast distribution servers, there are situations where this is undesired. One often encountered use case is that people want to integrate a video into 3rd-party tools that don't support stream security (e.g. other video players like h5p).

By introducing an exclusion pattern, admins can configure Opencast to not sign specific URLs. Since the publication channel ID is part of the URL, this enables admins to disable stream security for specific publication channels, hence allowing unsafe publications.

### DIFF
--- a/docs/guides/admin/docs/configuration/stream-security.md
+++ b/docs/guides/admin/docs/configuration/stream-security.md
@@ -67,6 +67,12 @@ It is also possible to use one key for multiple URL prefixes:
     key.demoKeyThree.organization=mh_default_org
 
 
+A Java regular expression can be defined to identify URLs to be excluded from URL signing.
+Any URL that matches this anchored regex will not be signed.
+
+    exclude.url.pattern=.*/.*/unprotected/.*/.*
+
+
 Configuration of URL Signing Timeout Values
 -------------------------------------------
 
--- a/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
+++ b/modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
@@ -50,6 +50,9 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 
 public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, ManagedService {
   /** The prefix for key configuration keys */
@@ -63,6 +66,9 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
 
   /** Value indicating that the key can be used by any organization */
   public static final String ANY_ORGANIZATION = "*";
+
+  /** The configuration key used for the exlusion list */
+  public static final String EXCLUSION_PROPERTY_KEY = "exclude.url.pattern";
 
   /** The security service */
   protected SecurityService securityService;
@@ -96,6 +102,9 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
 
   /** A mapping of URL prefixes to keys used to lookup keys for a given URL. */
   private TreeMap<String, Key> urls = new TreeMap<>();
+
+  /** A regular expression pattern used to identify URLs that shall not be signed. Can be null */
+  private Pattern exclusionPattern;
 
   /**
    * @param securityService
@@ -149,6 +158,7 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
 
     // Collect configuration in a new map so we don't partially override the old one in case of error
     TreeMap<String, Key> urls = new TreeMap<>();
+    Pattern exclusionPattern = null;
 
     // Temporary list of key entries to simplify building up the keys
     Map<String, Key> keys = new HashMap<>();
@@ -192,6 +202,12 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
           default:
             throw new ConfigurationException(propertyKey, "Unknown attribute " + attribute + " for key " + id);
         }
+      } else if (EXCLUSION_PROPERTY_KEY.equals(propertyKey)) {
+        String propertyValue = Objects.toString(properties.get(propertyKey), "");
+        if (!StringUtils.isEmpty(propertyValue)) {
+          exclusionPattern = Pattern.compile(propertyValue);
+        }
+        getLogger().debug("Exclusion pattern: {}", propertyValue);
       }
     }
 
@@ -208,9 +224,20 @@ public abstract class AbstractUrlSigningProvider implements UrlSigningProvider, 
     }
 
     this.urls = urls;
+    this.exclusionPattern = exclusionPattern;
   }
 
-  private boolean isValidUrl(String url) {
+  private boolean isExcluded(String url) {
+    boolean isExcluded = false;
+    Pattern exclusionPattern = this.exclusionPattern;
+    if (exclusionPattern != null) {
+      Matcher matcher = exclusionPattern.matcher(url);
+      isExcluded = matcher.matches();
+    }
+    return isExcluded;
+  }
+
+  private boolean isValid(String url) {
     try {
       new URI(url);
       return true;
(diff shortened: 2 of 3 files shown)


staubesv:
@JulianKniephoff One more (should be the last one now)... ;-)

staubesv:
@JulianKniephoff Have you already had a chance to look at the PR?

JulianKniephoff:
@staubesv not yet, sorry. It's on my list for today, directly after going through my email backlog. :wink:

That might still mean after lunch, though, but I will get to it today, promised!

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -208,9 +224,20 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
     }
 
     this.urls = urls;
+    this.exclusionPattern = exclusionPattern;
+  }
+
+  private boolean isExcluded(String url) {
+    boolean isExcluded = false;
+    Pattern exclusionPattern = this.exclusionPattern;
```
JulianKniephoff, line 232: This assignment is unnecessary, right? :thinking:

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -208,9 +224,20 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
+    if (exclusionPattern != null) {
+      Matcher matcher = exclusionPattern.matcher(url);
```
JulianKniephoff, line 234: We use the `matcher` just once, so we might as well just write `exclusionPattern.matches(url)`

This method could basically be

```
return exlcusionPattern != null && exclusionPattern.matches(url);
```

But I could understand if people find that too terse. :wink:

JulianKniephoff (changes requested):
Found two minor things. Code looks fine otherwise and compiles on my end. Testing now.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: No, the assignment is required to make the code thread-safe:

```
    if (exclusionPattern != null) {
      // When the configuration changes, exclusionPattern might by null here
      Matcher matcher = exclusionPattern.matcher(url);
```

JulianKniephoff:
I just realized that the exclusion pattern is a "global" setting. Would it make sense to extend this to a "per key" basis?

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
staubesv: > We use the matcher just once, so we might as well just write exclusionPattern.matches(url)

No, we use a pre-compiled regex (`Pattern`)  which is not supported by the method `matches`.

> `return exlcusionPattern != null && exclusionPattern.matches(url);`

No, that code is not thread-safe.

JulianKniephoff: Nevermind, I misread the docs. It's too hot. :weary:

staubesv:
> I just realized that the exclusion pattern is a "global" setting. Would it make sense to extend this to a "per key" basis?

I don't see a need to have this configurable per key and as long as there is no need, I would generally avoid any work related ;-)

docs/guides/admin/docs/configuration/stream-security.md
```diff
@@ -67,6 +67,11 @@ It is also possible to use one key for multiple URL prefixes:
     key.demoKeyThree.organization=mh_default_org
 
 
+A Java regular expression can be defined to identify URLs which shall not be signed:
+
+    exclude.url.pattern=.*/.*/unprotected/.*/.*
+
+
```
JulianKniephoff, line 75: We might consider mentioning that this is an "anchored pattern", i.e. the entire URL has to match, not just a substring.

The pattern `foo` does not match `http://localhost:8080/foo`, as I thought it would during my testing.

JulianKniephoff (changes requested):

JulianKniephoff:
@staubesv re per key configuration: fair enough.

I tested this and it seems to work fine, but I added another comment regarding the documentation.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Why, what would modify `this.exclusionPattern` here? I think nothing could. The docs specifically state:

Instances of this class (`Pattern`) are immutable and are safe for use by multiple concurrent threads. Instances of the Matcher class are not safe for such use.

JulianKniephoff: Btw even if the pattern was not immutable, copying a reference to it would not make the code thread safe.

Am I missing something here? :thinking:

JulianKniephoff: Ooh yes I am, I take everything back, see my comment about the weather above. :sweat_smile:

staubesv: > Btw even if the pattern was not immutable, copying a reference to it would not make the code thread safe.

As far as I know, assigning references is an atomic operation in Java. Therefore, the code `return exclusionPattern != null && exclusionPattern.matcher(url)` is thread-safe for the local variable `exclusionPattern` but not for the instance variable `this.exclusionPattern` because that variable is set whenever the configuration file changes which could happen between any two statements.

docs/guides/admin/docs/configuration/stream-security.md
staubesv: Although "A Java regular expression..." implies that "foo" will not match "Hello foo".

Should I add some more documentation here?

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
JulianKniephoff: Yes, you are right. I was thinking you were trying to protect `this.exclusionPattern` from having its internal state mutated by another thread during this operation, which just creating another reference would not help you with. But that doesn't matter anyway, since `Pattern` is immutable. And your code correctly protects us from the reference itself changing under us between checking it and calling methods on it, I see that now.

staubesv: Admittedly, the case is highly unlikely and wouldn't cause much damage... but still, there is a difference between software that is almost correct and software that is 100% correct ;-)

docs/guides/admin/docs/configuration/stream-security.md
JulianKniephoff: In the strict sense of the word "regular expression" you are right of course, but there are enough implementations around that would disagree; even Java's `Matcher` offers `lookingAt` in addition to `matches` which would make the regex `foo` match the string `foobar`.

So while in theory "regex" implies "has to match the entire string", in practice, and even in this case,  that is not a property inherent to regexes, but is rather determined by its use, and I think I would like to see something in the docs mentioning how it is used here. :thinking:

staubesv: ok, will add some more words here.

staubesv: @JulianKniephoff Updated documentation. Better?

JulianKniephoff: Yes, thanks! :smile:

JulianKniephoff (approved):

PR #925, opened 2019-05-27
MH-13585, Media Module Dependency Management

lkiesow:
Instead of hiding old library versions into the source tree,
dependencies should be loaded via dependency management to make them
easy to upgrade or check against security vulnerabilities.

This patch adds dependency management for the JavaScript libraries,
updating some of them along the way. A few libraries still remain since
they are either not available through npm or since it wasn't entirely
sure if they have been modified.


lkiesow:
fixed merge conflicts

staubesv (approved):

PR #927, opened 2019-05-28
MH-13587, Metadata Transfer Operation

lkiesow:
It is sometimes useful to transfer metadata from one catalog or one
element to another metadata element. For example, one could transfer the
people involved in a series to the episode metadata.

This workflow operation allows us to transfer arbitrary elements to
arbitrary locations within the metadata catalogs.


karendolan:
@lkiesow the series WOH has copy-metadata attribute to copy series catalog metadata into an episode catalog. What other use cases does this feature provide? What is the example of a type of catalog would need metadata from the existing episode catalog?

lkiesow:
We use this to copy metadata from dublincore catalogs to custom metadata catalogs.
Also, the series operation has hard-coded which fields will be copied to which targets fields.

karendolan:
> We use this to copy metadata from dublincore catalogs to custom metadata catalogs.

I'm curious on use cases for the additional mediapackage catalogs. Are they used to simply conversion to custom 3rd party systems that don't understand DublinCore? i.e. constructing pre-reformated catalogs for instant consumption by other systems?

lkiesow:
Yep, that's more or less it. Some ETH systems need specific catalogs and we use this operation to avoid having their video managers transfer all the data manually.

staubesv (approved):

PR #929, opened 2019-05-28
MH-13589 Check dependencies at build time

staubesv:
The plugin maven-dependency-plugin can be used to check dependencies at build time.

The goal of this task is to add that plugin to as many modules as possible implying the correction of errors it detects.

Notes:
- This PR does not cover all modules
- This PR doesn't take much care of the order the dependencies appear but tries to fit in the respective files. I tend to go for alphabetical order but don't do this consistently as the files being edited are already quite inconsistent...
- This PR does not update NOTICES (it also does not add any dependency ;-)

### DIFF
--- a/modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestJobEndpoint.java
+++ b/modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestJobEndpoint.java
@@ -21,7 +21,6 @@
 
 package org.opencastproject.adminui.endpoint;
 
-import org.opencastproject.inspection.api.MediaInspectionService;
 import org.opencastproject.job.api.Job;
 import org.opencastproject.job.api.Job.Status;
 import org.opencastproject.job.api.JobImpl;
@@ -99,7 +98,7 @@ public class TestJobEndpoint extends JobEndpoint {
             "2014-06-05T09:16:00Z", "2014-06-05T09:16:00Z", "testuser1", "host3"));
     jobs.add(createJob(3, Status.RUNNING, WorkflowService.JOB_TYPE, "RESUME",
             "2014-06-05T09:11:11Z", "2014-06-05T09:11:11Z", "testuser2", "host3"));
-    jobs.add(createJob(4, Status.RUNNING, MediaInspectionService.JOB_TYPE, "Inspect",
+    jobs.add(createJob(4, Status.RUNNING, "org.opencastproject.inspection", "Inspect",
             "2014-06-05T09:16:00Z", "2014-06-05T09:16:00Z", "testuser1", "host2"));
     jobs.add(createJob(5, Status.RUNNING, "org.opencastproject.composer", "Encode",
             "2014-06-05T09:05:00Z", "2014-06-05T09:05:00Z", "testuser3", "host1"));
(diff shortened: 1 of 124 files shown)


lkiesow:
> This PR does not update NOTICES (it also does not add any dependency ;-)

iirc, the maven part of the NOTICES was auto-generated anyway…

If we have covered most modules, we can also activate this by default in the main `pom.xml`. Then you will only need to define exceptions.

staubesv:
> This PR does not update NOTICES (it also does not add any dependency ;-)
> 
> iirc, the maven part of the NOTICES was auto-generated anyway…

@gregorydlogan Not sure how important it is to have the dependencies in the NOTICES, but please note that the current state is "not that good" (quite a lot of dependencies not mentioned). We probably should either remove them or fix them.

staubesv:
> If we have covered most modules, we can also activate this by default in the main pom.xml. Then you will only need to define exceptions.

I haven't counted the exact number of modules not covered now, but I estimate it to approx. 20. Not sure about how to configure per-module infos (ingoreDeclaredDependencies) etc, so I would rather stay with the current solution.

But whatever we do, you don't need to do it now.

lkiesow:
My idea was to have the plugin enabled by default without any ignores. Modules can then overwrite this by applying their own configuration in exactly the same way you did it here. But you can drop it wherever no configuration is necessary and new modules are forced to deal with it ;-)

Another option would be to pre-configure the logger as an allowed dependency since it is the most common and it's probably less individual configuration. The downside is that people can still integrate it if unnecessarily.

staubesv:
I think we should move towards having it enabled for all modules (including modules that are added) to ensure we don't end up in the current state again ;-)
But that would require to make it work with all modules which I currently can't (I left the problematic ones out... would take more time).

I would not ignore any unused declared dependencies on globally.

staubesv:
@lkiesow Fixed build -> PR ready for review.

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestJobEndpoint.java
```diff
@@ -21,7 +21,6 @@
 
 package org.opencastproject.adminui.endpoint;
 
-import org.opencastproject.inspection.api.MediaInspectionService;
```
staubesv, line 24: This causes a dependency to the inspection service. Seems not worth it.

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/MessageReceiver.java
```diff
@@ -29,7 +29,7 @@
 public interface MessageReceiver extends MessageBrokerConnector {
 
   /**
-   * {@link java.util.concurrent.Future} for the next {@link Serializable} from an {@link javax.jms.ObjectMessage}.
```
staubesv, line 32: The link seems to imply a dependency which is not available anymore... OMG ;-)

gregorydlogan:
@staubesv NOTICES being a shambles is a known issue.  Keeping it up to date manually sucks, not sure how better to go about it.  One of those things which will likely never be up to date, but it is being tracked.

lkiesow (approved):

lkiesow:
Stats now don't look that bad any longer. Thanks for the work:

```sh
% grep -rl maven-dependency-plugin modules --include=pom.xml | wc -l
166
% grep -rL maven-dependency-plugin modules --include=pom.xml | wc -l
45
```

PR #930, opened 2019-05-29
MH-13591: Add `angular-wizard`-s `.gitignore`

JulianKniephoff:
When you build `develop` currently, `bower` installs all the dependencies now, which is great! One of those is `angular-wizard`, and in its subfolder of `bower_components`, there is a `.gitignore` file, that was previously (when we managed these folders in the repository) not tracked. So when you then switch to an older branch, that file stays, and when you then switch back **again** to `develop`, it is still there. In fact, it is the only file remaining in `bower_components/angular-wizard`, while the rest of the files of that module are removed correctly, since they were already tracked in previous branches. When you then try to build again, `bower` complains, because it does not recognize this directory structure.

The "fix" is to just track the offending file in previous branches as well. This is a bit of a hack, since I don't actually know whether the current version of that file (the one that `bower` installs) actually corresponds to the one that would have been installed in previous versions, but it doesn't really hurt either, as far as I can tell.

Note that the base branch here is `r/5.x`. This might create conflicts down the line, especially with `develop`. I'm happy to help with these, but the most important thing to know is that we want this file in all the current release branches (`r/5.x`, `r/6.x` and `r/7.x`), but not in `develop`.


gregorydlogan:
I'm going to merge this since Travis is happy.  The scheduler-impl tests are failing in buildbot, for no apparent reason.  Sigh.

PR #931, opened 2019-05-29
MH-13592, Assembly Configuration

lkiesow:
This patch tries to make Opencast's Karaf feature configuration more
flexible and easier to understand by grouping some bundles in dedicated
features instead of configuring them over and over again.

- All workflow operation modules got their own feature which is included
  in the admin distributions
- The transcoding modules got their own feature which is included in all
  of the worker modules.
- The cideo analysis modules got their own feature which is included in
  all of the worker modules.

Splitting up the transcoding modules which need a lot of performance and
the analysis which need less performance is done to make it easier to
create custom distributions for high-performance machines, …

These changes do not modify the current distributions.

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -149,11 +149,11 @@
 
   <feature name="opencast-allinone" version="${project.version}">
     <feature start-level="80">opencast-core</feature>
+    <feature start-level="80">opencast-services-processing-heavy-load</feature>
+    <feature start-level="80">opencast-services-processing-light-load</feature>
 
     <bundle start-level="82">mvn:org.opencastproject/opencast-admin-ui/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-admin-ui-index/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-animate-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-animate-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-animate-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-annotation-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-annotation-impl/${project.version}</bundle>
@@ -162,17 +162,11 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-asset-manager-storage-fs/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-asset-manager-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-authorization-manager/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-caption-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-caption-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-capture-admin-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-capture-admin-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-comments-workflowoperation/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-composer-ffmpeg/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-composer-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-composer-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-conductor/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-dictionary-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-dictionary-regexp/${project.version}</bundle>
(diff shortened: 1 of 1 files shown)


gregorydlogan:
This looks good to me.  I've restarted the failed Travis jobs, but they were failing on a network issue at their end.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -147,41 +147,105 @@
...
+    <bundle start-level="82">mvn:org.opencastproject/opencast-transcription-service-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-videoeditor-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-videosegmenter-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-waveform-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-workflow-workflowoperation/${project.version}</bundle>
+  </feature>
+
+  <!-- A collection of encoding services used on workers (these need a lot of performance) -->
+  <feature name="opencast-encoding" version="${project.version}">
+    <feature start-level="80">opencast-core</feature>
```
staubesv: Is it necessary to include opencast-core here? At the end of the day, any feature that is supposed to be a "runnable Opencast node" requires opencast-core and opencast-core is also included in all distributions (admin, worker, ...).

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -147,41 +147,105 @@
...
-    <bundle start-level="82">mvn:org.opencastproject/opencast-composer-ffmpeg/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-composer-service-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-composer-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-conductor/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-impl/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-cover-image-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-dictionary-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-dictionary-regexp/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-distribution-service-api/${project.version}</bundle>
```
staubesv: As there are workflow operation handlers in the opencast-worklow-operations feature, it might make sense to also create a feature opencast-distribution-services or similar to be consistent with the new structure of features.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -250,99 +301,61 @@
...
-    <bundle start-level="82">mvn:org.opencastproject/opencast-textanalyzer-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-textanalyzer-impl/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-textanalyzer-workflowoperation/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-textextractor-tesseract/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-themes/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-themes-workflowoperation/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-api/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-ffmpeg/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-transcription-service-api/${project.version}</bundle>
```
staubesv, line 247: Same here: As there are related workflow operations, this might be moved in one of the new features added by this PR.

staubesv (dismissed):
I do like the idea of having smaller feature packages for sake of readability which is the direction this PR goes.

Here some general ideas/feedback:

- Maybe we could add opencast-workflow-service-api, opencast-workflow-service-impl and opencast-workflow-condition-parser (others?) to the package opencast-workflow-operations and rename that package to something like opencast-processing-control (or opencast-workflow-service). Reasoning: the workflow operations need to be on the node that has those bundles anyway.
- I would prefer opencast-encoding-services over opencast-encoding
- opencast-analysis seems a somewhat arbitrary name choice as half of the operations actually don't do analysis. Also not really clear to me what that advantage of separating "heavy load" vs "low load" WOHs has... should be job load that controls the distribution of loads, shouldn't it. Please also note the opencast-inspection-service-ffmpeg is an analysis operation that is usually cheap. I would probably go for a single feature opencast-processing-services here (no strong opinion). In case you really want the separation based on load the WOHs cause, this might be directly indicated in the name of the feature (e.g. opencast-processing-services-high-load and opencast-processing-services-low-load). That makes the purpose explicit and avoids the need to come up with fancy names for the features containing unrelated "low load" WOHs  ;-)

A disadvantage the new structure has is the separation of closely related bundles (api, impl, remote, workflow operation) across different features. That is not necessarily a bad thing, but what I currently miss is a better handling of the "remotes" in means of: I can include a feature to include a number of bundles with api & impl, but then for each bundle need to place the corresponding remotes individually in the including feature. Maybe there is a way to end up with something like:

admin node:
Include: opencast-processing-control
Include: opencast-processing-services-remote

worker node:
Include: opencast-processing-services

Instead of the current state that seems to do:

admin node:
Include: Bundles required for processing (e.g. workflow-service-impl, ...?)
Include: opencast-workflow-operations
Include: All remotes of the bundles found in opencast-encoding/analysis

worker node:
Include: opencast-encoding
Include: opencast-analysis

lkiesow:
> * Maybe we could add opencast-workflow-service-api, opencast-workflow-service-impl and opencast-workflow-condition-parser …

Makes sense, will do if I don't drop this.

> * I would prefer opencast-encoding-services over opencast-encoding

I will change it to a general `opencast-services-*` prefix

> * Also not really clear to me what that advantage of separating "heavy load" vs "low load" WOHs has... should be job load that controls the distribution of loads, shouldn't it.

The purpose is quite different and has been discussed multiple times lately: If you have inhomogenous infrastructure you caneasily define you own distribution which  handle the havy load (transcoding, …) and put that on som bulky machines while offloading the less havy tasks to some other machines. The ETH practices this for a long time and and bringing their know-how to the community is the main goal behind this pull request.

> Please also note the opencast-inspection-service-ffmpeg is an analysis operation that is usually cheap. 

The transcoding operations heavily depend on this offloading this to other workers can increase your io load. It's good to keep them close to the transcoding operations.

> e.g. opencast-processing-services-high-load and opencast-processing-services-low-load

will go for something like that.
 
> A disadvantage the new structure has is the separation of closely related bundles (api, impl, remote, workflow operation) across different features. That is not necessarily a bad thing, but what I currently miss is a better handling of the "remotes" in means of: I can include a feature to include a number of bundles with api & impl, but then for each bundle need to place the corresponding remotes individually in the including feature.

Since the woh extraction was more an afterthough, I may drop that part for know if it's acceptance relies on restructuring the rest. Will see…

assemblies/karaf-features/src/main/feature/feature.xml
lkiesow: the idea here is to not modify the current distributions

assemblies/karaf-features/src/main/feature/feature.xml
lkiesow: This feature does actually depend on opencast-core so dependencies should be properly included.

smarquard:
I don't have any objections to this change, but I do think that the job load mechanism should be the primary way to allocate jobs efficiently amongst workers and if there are inefficiencies, we should address those through better load measurement and job allocation or queuing strategies.

If we expect adopters to start creating specialized distributions and figuring out what goes to what type of workers, we're increasing deployment complexity. Adopters shouldn't have to do complex things to make Opencast run well.

lkiesow:
@smarquard, this has *nothing* to do with job loads. I don't expect any usual Opencast user to build their own assemblies although I know of a few who do that for various reasons. This is about making modifications easier if you have to (if you add services as a developer or build your modified assemblies) not about making this mandatory.

staubesv:
@lkiesow I think what @smarquard and me mean is that if we had a better job scheduler that is able to dispatch jobs to workers in a way that makes this work well in case of workers that are very different considering their capabilities, there would be no need to build workers with "low-load"/"high load" services. This would likely take more efforts than just saying a worker has a maximum load and jobs have some loads, though.
Seem from that point of view, the idea of "low-load"/"high-load" services is a workaround for our job scheduler not being able to dispatch jobs to heterogeneous workers.

... but that kind of thinking is way out of scope of this PR.

KatrinIhler:
@lkiesow This branch has conflicts!

staubesv (approved):

PR #932, opened 2019-05-31
MH-13594, Fix waveform color

lkiesow:
This fixes the default color and a concurrency problem for generating
waveforms.

*untested*


lkiesow:
declining this in favor of #933

PR #933, opened 2019-05-31
MH-13593: Incorrect default waveform colors

gregorydlogan:
Resolving the white waveform image issue:

* This patch resolves an issue where the waveform workflow operation overwrote the service-wide defaults with its current settings.
* The waveform service's service wide setting was also (by default) blank, which leads to white images, which are invisible in the current UI.
* Set the default of black when the waveform WOH's color parameter was blank.

### DIFF
--- a/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
+++ b/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
@@ -44,6 +44,7 @@ import org.opencastproject.workspace.api.Workspace;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedService;
@@ -402,8 +403,10 @@ public class WaveformServiceImpl extends AbstractJobProducer implements Waveform
       filterBuilder.append(waveformFilterPre);
       filterBuilder.append(",");
     }
-    if (color != null) {
-      waveformColor =  StringUtils.split(color, "|");
+    String[] waveformOperationColors = ArrayUtils.clone(waveformColor);
+    //If the color was set, override the defaults
+    if (StringUtils.isNotBlank(color)) {
+      waveformOperationColors = StringUtils.split(color, "|");
     }
     filterBuilder.append("showwavespic=");
     filterBuilder.append("split_channels=");
@@ -415,7 +418,7 @@ public class WaveformServiceImpl extends AbstractJobProducer implements Waveform
     filterBuilder.append(":scale=");
     filterBuilder.append(waveformScale);
     filterBuilder.append(":colors=");
-    filterBuilder.append(StringUtils.join(Arrays.asList(waveformColor), "|"));
+    filterBuilder.append(StringUtils.join(Arrays.asList(waveformOperationColors), "|"));
     if (waveformFilterPost != null) {
       filterBuilder.append(",");
       filterBuilder.append(waveformFilterPost);
--- a/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
+++ b/modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
@@ -44,7 +44,6 @@ import org.opencastproject.workspace.api.Workspace;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
-import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
I see @lkiesow was a little faster than me with #932.  Whoever takes these please only merge one of them :)

modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
```diff
@@ -402,8 +403,10 @@ private String createWaveformFilter(Track track, int width, int height, String c
       filterBuilder.append(waveformFilterPre);
       filterBuilder.append(",");
     }
-    if (color != null) {
-      waveformColor =  StringUtils.split(color, "|");
+    String[] waveformOperationColors = ArrayUtils.clone(waveformColor);
```
lkiesow: There isn't any reason to copy the array and we don't even need to split the configuration (not that my patch did a better job here). Suggestion:

```java
if (StringUtils.isBlank(color) {
    color = StringUtils.join(Arrays.asList(waveformColor), "|");
}
…
filterBuilder.append(color);
```

modules/waveform-workflowoperation/src/main/java/org/opencastproject/workflow/handler/waveform/WaveformWorkflowOperationHandler.java
```diff
@@ -155,6 +158,10 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
     );
 
     String color = StringUtils.trimToNull(workflowInstance.getCurrentOperation().getConfiguration(COLOR_PROPERTY));
+    if (StringUtils.isBlank(color)) {
```
lkiesow: There isn't any real reason the set the default twice, is there?
Plus, we don't always additionally log that we are using the default in info level.

lkiesow (changes requested):
Should work.
Just some minor request for improvements.

mliradelc:
This is a coincidence, I made the same solution (Changed the color after parsing the job arguments if there was an empty string instead of a null one) last Tuesday, but I discarded because the All in one installation don't have this problem.

```java
           String color = arguments.get(5);
          if (("".equals(color))) {
           color = WAVEFORM_COLOR_CONFIG_KEY;
          }
```




For Me the problem is not in this part, it is a consequence of something in the way that there are sent or received the messages.


Anyways, the fix posted by @gregorydlogan and @lkiesow  works and creates a layer of redundancy,

modules/waveform-workflowoperation/src/main/java/org/opencastproject/workflow/handler/waveform/WaveformWorkflowOperationHandler.java
mliradelc: > There isn't any real reason the set the default twice, is there?

When I made tests last week I notice that the worker applied the default value at startup, but when it was called the function the `waveFormColor` variable was blank. I think is something with the scope.

lkiesow: It's because the remote implementation would send a blank value which then wouldn't meet the criteria of being `null` but the `isBlank` in the impl will catch that case.

modules/waveform-ffmpeg/src/main/java/org/opencastproject/waveform/ffmpeg/WaveformServiceImpl.java
gregorydlogan: Ah, good catch, definitely don't need the clone.  I was including the clone to make the operation additive (ie, you get the default, and whatever else you want) rather than absolute (you get the default, or what you say you want).  But that does change the API, so I'll revert that.

modules/waveform-workflowoperation/src/main/java/org/opencastproject/workflow/handler/waveform/WaveformWorkflowOperationHandler.java
gregorydlogan: Re: setting the default, the WOH's color instance variable is set to null if there is no setting provided in the XML, so yes, we do need to set the default here :)

gregorydlogan:
Should be ready for review again!

modules/waveform-workflowoperation/src/main/java/org/opencastproject/workflow/handler/waveform/WaveformWorkflowOperationHandler.java
lkiesow: > the WOH's color instance variable is set to `null` if there is no setting provided in the XML

…which should be okay since the service implementation now uses a default value if the incoming color is `null` or blank. If we catch empty values here, the impl wouldn't need a default and could just throw an exception instead to make it clear that it does expect proper values.

Setting the default at two places is confusing and may easily lead to error if we ever want to change the default.

gregorydlogan: Yeah, good point.  Ok, this has been removed.

PR #935, opened 2019-06-02
MH-13596, Drop Distribution “adminworker”

lkiesow:
As discussed on list, this drops the “adminworker” distribution from the
set of default distributions.


staubesv:
@lkiesow Not in scope of this PR, but it might make sense to consider removing the distribution adminpresentation, too.

staubesv (approved):

mtneug:
If I recall correctly, UTC is using `adminpresentation`. They extended the LTI tools with features that require access to APIs that only the `admin` distribution provides. Authenticating via LTI agains `presentation` only is therefore not enough.

lkiesow:
There are a few installations I know which are using `adminpresentation`. Just none I know of which are using `adminworker`.

staubesv:
@lkiesow & @mtneug Fine for me.

PR #938, opened 2019-06-03
MH-13599, Select well supported mime type by default

lkiesow:
By default, Opencast selects the first defined mime type for a given
file extension which leads to the publication of uncommon types in a few
cases.

This patch re-orders some type configurations to use the more common
type by default.

### DIFF
--- a/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml
+++ b/modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml
@@ -20,11 +20,6 @@
     <Extensions>json</Extensions>
   </MimeType>
   <MimeType>
-    <Type>application/ogg</Type>
-    <Description>Ogg file</Description>
-    <Extensions>ogg,oga,ogv,ogx</Extensions>
-  </MimeType>
-  <MimeType>
     <Type>application/pdf</Type>
     <Description>Portable Document Format</Description>
     <Extensions>pdf</Extensions>
@@ -115,22 +110,17 @@
     <Extensions>mid,midi,smf,kar</Extensions>
   </MimeType>
   <MimeType>
-    <Type>audio/mp3</Type>
-    <Description>MP3 audio</Description>
-    <Extensions>mp3,swa</Extensions>
-  </MimeType>
-  <MimeType>
-    <Type>audio/mp4</Type>
-    <Description>MPEG-4 media</Description>
-    <Extensions>mp4</Extensions>
-  </MimeType>
-  <MimeType>
     <Type>audio/mpeg</Type>
     <Description>MPEG audio</Description>
     <Extensions>m1a,mp2,mpa,m2a,mp3,swa</Extensions>
   </MimeType>
   <MimeType>
     <Type>audio/mpeg3</Type>
+    <Description>MP3 audio</Description>
+    <Extensions>mp3,swa</Extensions>
+  </MimeType>
+  <MimeType>
+    <Type>audio/mp3</Type>
     <Description>MP3 audio</Description>
     <Extensions>mp3,swa</Extensions>
   </MimeType>
@@ -387,7 +377,12 @@
   <MimeType>
     <Type>video/mp4</Type>
     <Description>MPEG-4 media</Description>
-    <Extensions>mp4,m4v</Extensions>
+    <Extensions>mp4,mp4v,m4v</Extensions>
+  </MimeType>
+  <MimeType>
+    <Type>audio/mp4</Type>
+    <Description>MPEG-4 media</Description>
+    <Extensions>mp4,mp4a,m4a</Extensions>
   </MimeType>
   <MimeType>
     <Type>video/mpeg</Type>
@@ -403,6 +398,11 @@
     <Type>video/ogg</Type>
     <Description>Ogg video</Description>
     <Extensions>ogg,ogv,ogx</Extensions>
+  </MimeType>
+  <MimeType>
+    <Type>application/ogg</Type>
+    <Description>Ogg file</Description>
+    <Extensions>ogg,oga,ogv,ogx</Extensions>
   </MimeType>
   <MimeType>
     <Type>video/quicktime</Type>


staubesv:
@lkiesow In what why does the order affect the determination of the mime type? First found top down or bottom up?

lkiesow:
The first one is selected. For example `audio/mpeg` is now used for mp3:

```
% grep -B2 'Extensions.*mp3' modules/common/src/main/resources/org/opencastproject/util/MimeTypes.xml
    <Type>audio/mpeg</Type>
    <Description>MPEG audio</Description>
    <Extensions>m1a,mp2,mpa,m2a,mp3,swa</Extensions>
--
    <Type>audio/mpeg3</Type>
    <Description>MP3 audio</Description>
    <Extensions>mp3,swa</Extensions>
--
    <Type>audio/mp3</Type>
    <Description>MP3 audio</Description>
    <Extensions>mp3,swa</Extensions>
--
    <Type>audio/x-mp3</Type>
    <Description>MP3 audio</Description>
    <Extensions>mp3,swa</Extensions>
--
    <Type>audio/x-mpeg3</Type>
    <Description>MP3 audio</Description>
    <Extensions>mp3,swa</Extensions>
```

Which [should be the correct mime type](https://stackoverflow.com/a/10688641).

The reasoning for this patch is that we have observed some problems with uncommon mime types and players/browsers.

staubesv:
> The first one is selected. For example audio/mpeg is now used for mp3:

So mp4 will end up as audio/mp4? Maybe take the chance and move video/mp4 upwards then as mp4 is more likely used with video?

lkiesow:
Good point though I tend to actually drop

```
    <Type>audio/mp4</Type>
    <Description>MPEG-4 media</Description>
    <Extensions>mp4</Extensions>
```

since the extension `mp4` really implies video and we have:

```
    <Type>video/mp4</Type>
    <Description>MPEG-4 media</Description>
    <Extensions>mp4,m4v</Extensions>
```

and audio is already covered here:

```
    <Type>audio/mpeg</Type>
    <Description>MPEG audio</Description>
    <Extensions>m1a,mp2,mpa,m2a,mp3,swa</Extensions>
```

…and…

```
    <Type>audio/m4a</Type>
    <Description>AAC audio</Description>
    <Extensions>m4a</Extensions>
```

Any thoughts?

staubesv:
> [...] since the extension mp4 really implies video [...]

No, it doesn't: MP4 containers can contain only audio streams in which case audio/mp4 would be the correct mime type. Seem from that point of view, it is more correct to just move audio/mp4 downwards as it is more likely that we deal with video/mp4 in our case.

The problem here is that the file extension simply does not provide enough information to determine the best fitting mime type. But that is not in scope of this PR.

As for this PR, I suggest to just move video/mp4 upwards.

lkiesow:
done.

staubesv (approved):

mliradelc:
Using Safari, the new combination doesn't allow some MP4 files to be uploaded, the file format that is:
is:

General
Complete name                            : /Users/mlira/Downloads/SampleVideo_1280x720_1mb.mp4
Format                                   : MPEG-4
Format profile                           : Base Media
Codec ID                                 : isom (isom/iso2/avc1/mp41)
File size                                : 1.01 MiB
Duration                                 : 5 s 312 ms
Overall bit rate mode                    : Variable
Overall bit rate                         : 1 590 kb/s
Encoded date                             : UTC 1970-01-01 00:00:00
Tagged date                              : UTC 2014-07-19 17:15:29
Writing application                      : Lavf53.24.2

Video
ID                                       : 1
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : Main@L3.1
Format settings                          : CABAC / 1 Ref Frames
Format settings, CABAC                   : Yes
Format settings, ReFrames                : 1 frame
Codec ID                                 : avc1
Codec ID/Info                            : Advanced Video Coding
Duration                                 : 5 s 280 ms
Bit rate                                 : 1 206 kb/s
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 25.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Scan type                                : Progressive
Bits/(Pixel*Frame)                       : 0.052
Stream size                              : 777 KiB (75%)
Encoded date                             : UTC 1970-01-01 00:00:00
Tagged date                              : UTC 1970-01-01 00:00:00
Codec configuration box                  : avcC

Audio
ID                                       : 2
Format                                   : AAC LC
Format/Info                              : Advanced Audio Codec Low Complexity
Codec ID                                 : mp4a-40-2
Duration                                 : 5 s 312 ms
Bit rate mode                            : Variable
Bit rate                                 : 384 kb/s
Maximum bit rate                         : 400 kb/s
Channel(s)                               : 6 channels
Channel layout                           : C L R Ls Rs LFE
Sampling rate                            : 48.0 kHz
Frame rate                               : 46.875 FPS (1024 SPF)
Compression mode                         : Lossy
Stream size                              : 250 KiB (24%)
Default                                  : Yes
Alternate group                          : 1
Encoded date                             : UTC 1970-01-01 00:00:00
Tagged date                              : UTC 1970-01-01 00:00:00

lkiesow:
I'm not quite sure in which way this patch has something to do with uploading files.

lkiesow:
@mliradelc the reason for your issue is more likely pull request #889

mliradelc:
> @mliradelc the reason for your issue is more likely pull request #889

That would be the case if all the browsers had the same behavior,  but this happens only with Safari. Chrome works.

staubesv:
@mliradelc I also tend to say that #889 causes this and Safari simply treats the mime type differently. To test this, maybe just try to set accept to `*/*` in https://github.com/opencast/opencast/pull/889/files#diff-965a7b694968b7cd16fb89cdcd15b2bbR43

I've also seen this while testing. There should be a patch available soon.

PR #943, opened 2019-06-05
MH-13604 color "blue" for links in the admin ui

CGreweling:
A user can not see directly which parts of the admin-ui are links or actions.
This patch adds color "blue" as used for Links Details/Edit for all links
in the admin ui.

Links_blue

### DIFF
--- a/modules/admin-ui/src/main/webapp/styles/base/_layout.scss
+++ b/modules/admin-ui/src/main/webapp/styles/base/_layout.scss
@@ -42,7 +42,7 @@ body {
 // Resets
 
 a {
-    color: $color-link;
+    color: #1d5888;
     text-decoration: none;
 
     &:hover {


staubesv:
@CGreweling No strong opinion, but some thoughts from my side: When we built the cross-link feature, we thought about whether we would want to indicate the links by using a different font color and concluded that we would not want to do so as this makes the table potentially "too colorful" and distracts the user from the actual content.
Haven't looked at this PR to see how it looks & feels with 20 rows all showing some columns with all rows "blue", but be aware that this is a trade-off between "fancy" colored tables vs. obvious indication of what can be clicked without the need to mouse-over.

gregorydlogan:
While I don't particularly care whether we colour the links or not, this particular shade of blue is almost invisible to me - it looks basically black, and thus doesn't really change me.  I also agree with @staubesv's comment about the colour tradeoff, which I get is not helpful when I'm also complaining that the colour isn't vivid enough :)

Maybe this is something that should be brought to the users or dev list?

uzk-rrzk:
I back ported this feature to our 7.x version and I like it. Often our users do not know that they can click on certain fields (for instance to perform a sort). The used color is descreet enough and clearly different to black.

PR #944, opened 2019-06-05
MH-13606 Don't archive smil on publication

KatrinIhler:
Don't overwrite the original smil with the one created on publication, since the later might have a different base than the former (e.g. longer tracks might have been hidden, changing the length of the video),
leading to strange effects in the video editor afterwards (e.g. cut segments appearing). Instead, recreate the second smil every time we publish.

### DIFF
--- a/etc/workflows/partial-publish-process-smil.xml
+++ b/etc/workflows/partial-publish-process-smil.xml
@@ -23,33 +23,11 @@
         <configuration key="source-flavors">*/work</configuration>
         <configuration key="preview-flavors">*/preview</configuration>
         <configuration key="smil-flavors">smil/cutting</configuration>
-        <configuration key="target-smil-flavor">smil/cutting</configuration>
+        <configuration key="target-smil-flavor">smil/publish</configuration>
         <configuration key="target-flavor-subtype">trimmed</configuration>
         <configuration key="interactive">false</configuration>
         <!-- skip the built in editor transcode to use process-smil-->
         <configuration key="skip-processing">true</configuration>
-      </configurations>
-    </operation>
-
-    <!-- Tag smil/cutting for later re-use in case of re-publishing of the event
-         Note that WOH editor will re-use target-smil-flavor instead of adding a new one -->
-
-    <operation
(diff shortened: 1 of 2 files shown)


karendolan:
@KatrinIhler can you clarify further on this pull. For example, which module creates the original smil file and how are the latest edits related to the current publication persisted and retrieved? Or are manual edits not persisted?

KatrinIhler:
@karendolan The original smil is created/overwritten whenever you either save in the video editor or start a workflow in the editor. It is based on the composite video shown in the editor. Later in the publication workflow, another smil is created from the first, but this one is based on the actual track(s). This can lead to weird effects if we open the video editor again after that, especially if we hide tracks before publishing (new select-tracks feature) and the tracks have different lengths.
Example: If I hide the longer track, the second smil is based only on the second track, which is shorter than the composite video, so if we archive this smil, the video editor will show a cut segment afterwards that wasn't there before.

karendolan:
@KatrinIhler I'm still confused on the big picture of the scenario. Are there 2 SMIL files generated for  a trimmed composite video: one specific to the ingested composite arrangement of video components (pre-manual-trim), and a secondary one that contains the manually requested trim points, such as which tracks are selected and which parts of the selected tracks are to be trimmed?

My underlying concern is that the manually selected trim points are still going to be archived, at some point, so they can be revisited if they need to be adjusted. Our site's use case is that the trimmer reviews the transcoded video with the applied trim points before giving students access. If the trim points were not applied correctly, the user can adjust the previous trim points and re-transcode. The last saved trim points can be re-used over the raw video to manually recreate the publication videos.

Would it be reasonable to import the pre-trim SMIL points into the editor and just add on to that SMIL when making manual trim adjustments, saving the SMIL to a new version, so there is always a SMIL with all the points and instructions relevant to the source ingest files?

KatrinIhler:
@karendolan The first smil file, based on the composite video shown in the video editor, already contains the manual trimming! The only difference is that the second smil is based on the tracks selected for publication, so the links to the video file will be different.

PR #948, opened 2019-06-07
MH-13515, add health-check endpoint

JamesUoM:
provides more detailed json response as defined by https://tools.ietf.org/id/draft-inadarei-api-health-check-03.html

_status_ | _notes_ | HTTP code | meaning
--- | --- | --- | ---
pass|n/a|200|All is OK
warn|service(s) in WARN state|200|Partially working service here
warn|services(s) in ERROR state|200|Look for service on another node
fail|maintenance|503|Node not available, try again later
fail|disabled|503|Node not available, try another node
fail|offline|503|Node not running, try another node


modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
```diff
@@ -218,6 +233,149 @@ public String getMyInfo() {
...
+  @Path("health")
+  @Produces("application/health+json")
+  @RestQuery(name = "health", description = "Opencast node health check. Implements this internet-draft health check api https://inadarei.github.io/rfc-healthcheck",
+          reponses = {
+            @RestResponse(responseCode = HttpServletResponse.SC_OK, description = "Node is running, check reponse for details"),
+            @RestResponse(responseCode = HttpServletResponse.SC_SERVICE_UNAVAILABLE, description = "Node is offline or unresponsive, check response for details")},
+          returnDescription = "Details of the Opencast node's health status")
+
+  public String getHealth(@Context HttpServletResponse response) {
+    /* Response implements https://inadarei.github.io/rfc-healthcheck
```
gregorydlogan, line 244: This, IMO, should be in the markdown docs somewhere, and linked to.

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
```diff
@@ -218,6 +233,149 @@ public String getMyInfo() {
...
+        } catch (ServiceRegistryException e) {
+          logger.error("Failed to get services: ", e);
+          status = HEALTH_CHECK_STATUS_FAIL;
+          notes.add("Internal health check error!");
+        }
+      }
+    } catch (ServiceRegistryException e) {
+      logger.error("Failed to get host registration: ", e);
+      status = HEALTH_CHECK_STATUS_FAIL;
+      notes.add("Internal health check error!");
```
gregorydlogan: Internal here is capitalized, where the earlier notes are not.

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
```diff
@@ -218,6 +233,149 @@ public String getMyInfo() {
...
+                break;
+              }
+              default:
+                break;
+            }
+          }
+        } catch (ServiceRegistryException e) {
+          logger.error("Failed to get services: ", e);
+          status = HEALTH_CHECK_STATUS_FAIL;
+          notes.add("Internal health check error!");
```
gregorydlogan: Internal here is capitalized, where the earlier notes are not.

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
```diff
@@ -218,6 +233,149 @@ public String getMyInfo() {
...
+    List<String> notes = new ArrayList<>();
+    List<Object> serviceStates = new ArrayList<>();
+    Map<String, Object> checks = new HashMap<>();
+
+    try {
+      HostRegistration host = serviceRegistry.getHostRegistration(hostname);
+
+      // check most severe conditions first
+      if (!host.isOnline()) {
+        // NOTE: This is not strictly possible as a node can't test if it's offline
```
gregorydlogan: I could see the case for extending this endpoint to be able to check on the health of *other* nodes from the current one...

gregorydlogan (changes requested):
Minor complaints

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
JamesUoM: As this is primarily for use by load balancers, that would not make sense initially. Maybe an option to check the cluster or separate endpoints e.g. /info/health/all,  /info/health/<host>

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
JamesUoM: surprised I didn't add it myself!

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
gregorydlogan: Yep exactly.  Future work, not really relevant here, but it's better to have the logic present so that it doesn't cause issues down the line :)

gregorydlogan:
@JamesUoM Build errors.

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/RuntimeInfo.java
JamesUoM: Without changing anything we could simply aggregate the responses from all the nodes by calling /info/health for each one.

Alternative just adding an additional checks object for "node:statuses" and use the observedValue to return pass|warn|fail. This can be added in the future without breaking the format.

For now I'll just look at breaking  the host analysis out into a separate function

gregorydlogan (approved):

PR #950, opened 2019-06-11
MH-13612, Tag elements retrieved from asset manager

lkiesow:
By now, it seems to be the general assumptions that elements stored in
and then retrieved from the asset manager have the tag `archive`.

While this is often the case as the `archive` tag is widely used to
flag elements for archival, this is by no means guaranteed and this
misconception has lead to multiple issues over the last years.

Multiple workarounds and partial fixes have been added for this, e.g.:

  https://github.com/opencast/opencast/blob/49fe433b42d6dd1ec59c7364d23ff1060f19040a/etc/workflows/fast.xml#L81-L83

…but this can still easily happen again.

This patch finally ensures that all elements which are retrieved from
the asset manager as part of a media package are guaranteed to have the
`archive` tag, finally bringing an end to workarounds and uncertainty.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
@@ -27,12 +27,14 @@ import org.opencastproject.assetmanager.api.Snapshot;
 import org.opencastproject.assetmanager.impl.SnapshotImpl;
 import org.opencastproject.assetmanager.impl.VersionImpl;
 import org.opencastproject.mediapackage.MediaPackage;
+import org.opencastproject.mediapackage.MediaPackageElement;
 import org.opencastproject.mediapackage.MediaPackageParser;
 
 import org.eclipse.persistence.annotations.CascadeOnDelete;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
 import java.util.Date;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


karendolan:
@lkiesow is it possible that there is an underlying bug in the schedule process causes an episode catalog to be snapshotted into the archive without ever intentionally tagging it to exist in there?
The "fix" in this pull is bypasses the intentionally configured tagging efforts in  the workflows. 

If it's difficult to tell now why things are in the archive that are not tagged to be there, it's going to be just as hard to figure out what things are there and now being tagged to be there that shouldn't be there.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
```diff
@@ -181,6 +183,13 @@ public boolean removeAsset(AssetDto asset) {
   }
 
   public Snapshot toSnapshot() {
+    MediaPackage mediaPackage = Conversions.toMediaPackage(mediaPackageXml);
+    // ensure elements are tagged `archive`
+    for (MediaPackageElement element: mediaPackage.getElements()) {
+      if (!Arrays.asList(element.getTags()).contains("archive")) {
+        element.addTag("archive");
+      }
```
karendolan, line 192: @lkiesow can you add an INFO or WARN debug line here to identify which archived elements were not previously tagged for archive. It will help debug how they got there in the first place.

lkiesow:
Adding some more logging is no problem.

---

> The "fix" in this pull is bypasses the intentionally configured tagging efforts in the workflows.

Is DCE deliberately archiving elements not tagged `archive` so you can use them for workflows started from the asset manager? I've never seen that before. Only ever people complaining why things aren't tagged `archive` even though they had been retrieved from the asset manager. Would this patch be a problem for you?

---

> @lkiesow is it possible that there is an underlying bug in the schedule process causes an episode catalog to be snapshotted into the archive without ever intentionally tagging it to exist in there?

If I understand correctly what you mean, that's no bug but a deliberate choice by Entwine back in the day when they designed the new scheduler. The scheduled event is stored in the asset manager so that arbitrary additional data (e.g. extended metadata catalogs) can be added to a scheduled event. So you never need to tag anything for that to be put into the asset manager.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
lkiesow: done.

PR #954, opened 2019-06-13
MH-13616, Fix bundle versions

lkiesow:
This patch adds a few missing build versions, fixing the display of the
build version in the admin interface. It also adds a simple check for
this, ensuring that new modules will not forget about this any longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,8 @@ jobs:
         - (! grep -rn ' $' modules assemblies pom.xml --include=pom.xml)
         - (! grep -rn '	' etc)
         - (! grep -rn ' $' etc)
-        # build number must be present in all but four modules
-        - grep -L '<Build-Number>${buildNumber}</Build-Number>' modules/*/pom.xml | wc -l | grep -q 4
+        # build number must be present in all modules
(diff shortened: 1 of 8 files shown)


staubesv (dismissed):
@lkiesow 

> build number must be present in all but four modules

OMG... could you please add information about which four modules don't require a build number and why? Or could we maybe just add the build number to those four modules (even if not required) to make the rule "build number must be present" (instead of "build number must be present in all but four modules").

lkiesow:
We cannot add this to non-OSGi-bundles like the `rest-test-environment`. So, no we cannot make this mandatory for all and it's mandatory for all Opencast bundles which are all but four modules. We could bring this down to two if we include it into the hello world modules but you will have a magic number. Nonetheless, it will be fairly obvious if you break this test because either you know what you are doing or you need to include the build number ;-P

staubesv:
But still, it might be worth to just state something like


```
# The build number must be present in all OSGI-bundles which currently are all except the following four modules:
# rest-test-environment, module2, module3, module4

```

Makes thing less magic for people as me who don't would understand why four modules don't need a build number.

What do you think?

staubesv (approved):

PR #956, opened 2019-06-14
MH-13617 Re-introduce Prepare AV

KatrinIhler:
The Select-Tracks WO replaced prepare-av at parts of the standard workflow where it doesn't make sense. This also archives the media created by prepare-av so they don't have to be re-created.

~This PR includes #952.~

### DIFF
--- a/etc/org.opencastproject.adminui.cfg
+++ b/etc/org.opencastproject.adminui.cfg
@@ -202,15 +202,15 @@
 #
 # The default workflows use presenter/source, so that's the recommended default value.
 #
-# Default: presenter/source
-#sourcetrack.left.flavor=presenter/source
+# Default: presenter/prepared
+sourcetrack.left.flavor=presenter/prepared
 
 # Flavor that identifies the right hand stream, for use in the "Tracks" and "Thumbnail" panels of the video editor.
 #
 # The default workflows use presentation/source, so that's the recommended default value.
 #
-# Default: presentation/source
-#sourcetrack.right.flavor=presentation/source
+# Default: presentation/prepared
+sourcetrack.right.flavor=presentation/prepared
 
 # Flavor subtype that identifies audio preview elements, for use in the
 # "Tracks" panel of the video editor.
--- a/etc/workflows/partial-publish.xml
+++ b/etc/workflows/partial-publish.xml
@@ -11,6 +11,17 @@
   <operations>
 
     <operation
+        id="select-tracks"
+        exception-handler-workflow="partial-error"
+        description="Preparing presenter (camera) audio and video work versions">
+      <configurations>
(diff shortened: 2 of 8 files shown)


etc/org.opencastproject.adminui.cfg
```diff
@@ -203,14 +203,14 @@
 # The default workflows use presenter/source, so that's the recommended default value.
 #
 # Default: presenter/source
```
gregorydlogan: Default no longer matches setting

etc/org.opencastproject.adminui.cfg
```diff
@@ -203,14 +203,14 @@
...
-#sourcetrack.left.flavor=presenter/source
+sourcetrack.left.flavor=presenter/prepared
 
 # Flavor that identifies the right hand stream, for use in the "Tracks" and "Thumbnail" panels of the video editor.
 #
 # The default workflows use presentation/source, so that's the recommended default value.
 #
 # Default: presentation/source
```
gregorydlogan: Default no longer matches setting

etc/workflows/partial-publish.xml
```diff
@@ -10,6 +10,17 @@
 
   <operations>
 
+    <operation
+        id="select-tracks"
+        exception-handler-workflow="partial-error"
+        description="Preparing presenter (camera) audio and video work versions">
+      <configurations>
+        <configuration key="source-flavor">*/prepared</configuration>
+        <configuration key="target-flavor">*/work</configuration>
```
gregorydlogan, line 19: So `/work` is used in this partial, and no where else?

gregorydlogan (changes requested):

etc/workflows/partial-publish.xml
KatrinIhler: True. I guess we could rename it, but I couldn't think of something better...

etc/org.opencastproject.adminui.cfg
KatrinIhler: Fixed.

etc/org.opencastproject.adminui.cfg
KatrinIhler: Thanks, fixed!

KatrinIhler:
@gregorydlogan Any more comments for this one?

etc/workflows/partial-publish.xml
gregorydlogan: Derp, I missed that this was used as source for other things in this partial.  I was expected it to be used in *other* partials...

gregorydlogan (approved):

gregorydlogan:
I'll wait for this report to finish, but looking at the error I'm pretty sure it's some weird transient with the admin ui (sigh).  Once that's done I'll merge this in.

PR #967, opened 2019-06-17
MH-13631 Introduce general CatalogUIAdapter

KatrinIhler:
Since a lot of code between the DCCatalogUIAdapters for series and
events is the same, create a general adapter as a parent class to
contain that code.

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/ConfigurableDCCatalogUIAdapter.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/ConfigurableDCCatalogUIAdapter.java
@@ -1,0 +1,181 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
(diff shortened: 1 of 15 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/ConfigurableDCCatalogUIAdapter.java
```diff
@@ -0,0 +1,181 @@
...
+   *          the configuration properties
+   * @throws ConfigurationException
+   *           if there is a configuration error
+   */
+  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+    config = CatalogUIAdapterConfiguration.loadFromDictionary(properties);
+    organization = getCfg(properties, CONF_ORGANIZATION_KEY);
+    flavor = MediaPackageElementFlavor.parseFlavor(getCfg(properties, CONF_FLAVOR_KEY));
+    title = getCfg(properties, CONF_TITLE_KEY);
+    dublinCoreProperties = DublinCoreMetadataUtil.getDublinCoreProperties(properties);
```
staubesv, line 86: From an operational perspective this should be logged as INFO as it changes the state of the application which is a relevant information if you want to understand what Opencast did/does based on its log files.

staubesv (approved):

PR #968, opened 2019-06-17
MH-13632, Resolution Based, Conditional Encoding

lkiesow:
This patch allows the `encode` workflow operation to extend the
configured encoding profile based on the resolution of the input video
file. This allows for a simple configuration of a parallel encoding
profile with support for high-resolution output, even if that necessary
only for parts of the processed recordings.


lkiesow:
@ts23, did you already find time to take a look at this?

ts23 (approved):

lkiesow:
Thanks for the review

PR #971, opened 2019-06-18
NOJIRA: Add `ALTER` to necessary MySQL permissions

JulianKniephoff:
I just tried to upgrade an Opencast installation from 6 to 7. Running the database migration script failed, though, because I didn't have the right to `ALTER` tables. I used the docs to create my database user, which did not list this permission bit.

However, looking through other migration scripts, I saw that they also already used `ALTER` statements, so maybe this omission is intentional?

### DIFF
--- a/docs/guides/admin/docs/configuration/database.md
+++ b/docs/guides/admin/docs/configuration/database.md
@@ -68,10 +68,19 @@ create a database called `opencast` by executing:
 
 Then create a user `opencast` with a password and grant it all necessary rights:
 
-    GRANT SELECT,INSERT,UPDATE,DELETE,CREATE,DROP,INDEX,TRIGGER,CREATE TEMPORARY TABLES,REFERENCES ON opencast.*
+    GRANT SELECT,INSERT,UPDATE,DELETE ON opencast.*
       TO 'opencast'@'localhost' IDENTIFIED BY 'opencast_password';
 
-You can choose another name for the user and database and should use a different password.
+The rights granted here are all that is needed to *run* Opencast. To execute the migration scripts
+used to initialize (see next section) and upgrade the database schema upon releases of new versions
+of Opencast, you need more. If you don't want to do this using the `root` user (which normally
+can do anything), but with a dedicated user called `admin` for the sake of the example,
+you should grant that user the following rights:
+
+    GRANT SELECT,INSERT,UPDATE,DELETE,CREATE,ALTER,DROP,INDEX,TRIGGER,CREATE TEMPORARY TABLES,REFERENCES ON opencast.*
+      TO 'admin'@'localhost' IDENTIFIED BY 'opencast_admin_password';
+
+You can choose other names for the users and the database, and you **should** use a different password.
 
 In a distributed system, apart from `'username'@'localhost'` (which would allow access from the local machine only),
 you should grant a external user access to the database by running the same command for a user like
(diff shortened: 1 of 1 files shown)


lkiesow:
I thought that to be intentional. These are the permissions you need to run Opencast. But I don't have a strong opinion on keeping it this way.

JulianKniephoff:
They are the permissions to run Opencast **and** the database initialization, right? So I would argue the permissions to run the migration scripts belong here as well. Alternatively, I could imagine a note in the upgrade guide about any additional permissions needed. Actually that might be preferable, because the necessary permissions might change from version to version. (Although at some point, we might just have everything in this list. ;D)

Further opinions?

gregorydlogan:
IMO not having ALTER present is a bug.  I don't see the usecase where *not* having alter to our own databases makes a ton of sense.

mtneug:
I have never noticed this since we have two different DB user: one for running Opencast and one for managing the DB itself. The last one has more privileges. If the DB is generated via the SQL scripts I can't think of a situation where Opencast should use `ALTER` during runtime.

lkiesow:
Same here. I usually use a different user for changes as well so that I notice something is wrong if Opencast itself tries to modify anything at runtime. But looking at the current grant statements, they also contain `CREATE` which should also not be necessary at runtime.

JulianKniephoff:
So maybe we should document **that** workflow? Or at least provide two grant statements, one for Opencast's runtime needs and one for initializing/migrating?

I mean sure, most of the time you can just use the `root` user for the latter, but in cases where a client provides a database to us at ELAN, for example, we don't always have that option, and it would be nice to have a place to see all the permissions you might need to request of them.

gregorydlogan:
+1 for @JulianKniephoff's comment.  Do you want to amend this PR to reflect that workflow?

Personally I have a single `opencast` user who has `ALTER`, but then I'm also not running in production - and production users are who the documentation targets.  @lkiesow and @mtneug have a better prod setup.

JulianKniephoff:
@gregorydlogan I'll adapt this PR in the next couple of days, hopefully. :slightly_smiling_face:

JulianKniephoff:
Alright, I did the thing. I hope I didn't get any rights wrong ... :sweat_smile:

docs/guides/admin/docs/configuration/database.md
```diff
@@ -68,10 +68,19 @@ create a database called `opencast` by executing:
 
 Then create a user `opencast` with a password and grant it all necessary rights:
 
-    GRANT SELECT,INSERT,UPDATE,DELETE,CREATE,DROP,INDEX,TRIGGER,CREATE TEMPORARY TABLES,REFERENCES ON opencast.*
+    GRANT SELECT,INSERT,UPDATE,DELETE ON opencast.*
```
lkiesow:
```suggestion
    GRANT SELECT,INSERT,UPDATE,DELETE,CREATE TEMPORARY TABLES ON opencast.*
```
`CREATE TEMPORARY TABLES` is a runtime requirement.

lkiesow (changes requested):

docs/guides/admin/docs/configuration/database.md
JulianKniephoff: This suggestion feature is cool and all, but where is the button to "amend last commit with suggestion" or something like that? x)

gregorydlogan:
This looks good to me.  @lkiesow?

lkiesow (approved):

PR #972, opened 2019-06-19
MH-13636 Improve setting values from dublin core catalog

KatrinIhler:
Don't use a switch-case to recreate each metadata field, instead use the copy constructor to avoid side effects.
Move switch-case for value setting to Metadatafield class where it can hopefully disappear in the future. Remove condition for duration from frontend since this can never happen, values set here are either a
default from a listprovider or from a dublincore catalog.

~~This PR also contains #967 and should be merged afterwards!~~

### DIFF
--- a/modules/admin-ui/src/test/resources/episode-catalog.properties
+++ b/modules/admin-ui/src/test/resources/episode-catalog.properties
@@ -52,16 +52,6 @@ property.startDate.required=false
 property.startDate.pattern=yyyy-MM-dd'T'HH:mm:ssX
 property.startDate.order=3
 
-# A Start Time property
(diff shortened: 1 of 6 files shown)


staubesv:
@KatrinIhler Merge conflicts...

KatrinIhler:
@staubesv Fixed!

gregorydlogan:
@KatrinIhler Conflicts, but it looks good to me otherwise.

KatrinIhler:
@gregorydlogan Fixed!

gregorydlogan:
Not sure what's up with Travis, but this builds fine for me and will be merged soon.

PR #978, opened 2019-06-21
Mh 13617 Duplicate encoding profiles for PrepareAV/SelectStreams

KatrinIhler:
PrepareAV and SelectStreams share a number of encoding profiles, including that for muxing audio and video. Considering that these WOH serve different purposes, an independent configuration could be desired. This PR allows this by duplicating the encoding profiles, which also allows a correct format to be set (work vs. prepared).

~~Be aware that his PR also contains #956!~~


gregorydlogan:
@KatrinIhler Looks good, conflicts.  Looking at the file, it also looks like the defaults are set and uncommented.  These should be commented out, but left default :)

KatrinIhler:
@gregorydlogan This was because the other included PR is now merged. Fixed!

PR #979, opened 2019-06-25
MH-13643: Update to Paella 6.1.4

miesgre:


lkiesow:
@miesgre, 6.1.5 has been released. Do you want to upgrade Paella straight to the new version? Or do you want to stay with 6.1.4? Also, is it really necessary to push the upgrade to our legacy version? Does it fix something important? Otherwise, I would tend to send the upgrade to 7.x. But I'm not very familiar with Paellas releases, so it may make sense?

miesgre:
@lkiesow I updated the PR with paella 6.1.5.
Paella 6.1.4 fixes some problems with the audio that is played. 

We detected a bug in our opencast installation (https://opencast.jira.com/browse/MH-13644). To fix it, paella need to be fixed and also the opencast adapter.
We split it in two PR: One to update the paella (#979) and another to fix the opencast adapter (#980)

Paella 6.1.5 solves some problems with the "es.upv.paella.breaksPlugin" plugin and it is not urgent as I think we are the only ones using that plugin.

If you have tracks with and without audio, I recommend you to apply this PR and also (#980). If all your tracks have audio, it is not urgent to upgrade.

PR #982, opened 2019-06-25
MH-13646, Delete scheduled events

lkiesow:
Deleting scheduled events does not work properly: The entry in the
Elasticsearch index still remains even after the event has been deleted
and disappears only when rebuilding the whole index.

The reason for this is that the check for an event being a scheduled
event has changed but when the scheduling information are removed, the
old field was modified so that Opencast would always find existing
scheduler information and never remove the event from the index.

*Work sponsored by SWITCH*


kristofkeppens:
@lkiesow just tested this patch and deleting events now works as expected ( via UI and via external api )! Thanks!

lkiesow:
Thanks for testing this, @flyapen  → merged

PR #984, opened 2019-06-27
MH-13648, Asset Manager Concurrecy Issue

lkiesow:
The asset managers filesystem storage backend may fail to create
directories if it tries to create the directory multiple time in
parallel.

The reason for this is an unnecessary, non-synchronized check if the
directory already exists before trying the create the directory
(`mkdirs()` already includes this check) and the incorrect assumption
that `mkdirs()` returning false means that the directory is not created.

### DIFF
--- a/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/AbstractFileSystemAssetStore.java
+++ b/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/AbstractFileSystemAssetStore.java
@@ -176,7 +176,11 @@ public abstract class AbstractFileSystemAssetStore implements AssetStore {
 
   /** Create this directory and all of its parents. */
   protected void mkDirs(File d) {
-    if (d != null && !d.exists() && !d.mkdirs()) {
+    if (d == null) {
+      return;
+    }
+    // mkdirs *may* not have succeeded if it returns false, so check if the directory exists in that case
+    if (!d.mkdirs() && !d.exists()) {
       final String msg = "Cannot create directory " + d;
       logger.error(msg);
       throw new AssetStoreException(msg);


staubesv (changes requested):
@lkiesow This probably should go to r/6.x, shouldn't it?

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/AbstractFileSystemAssetStore.java
```diff
@@ -176,7 +176,11 @@ private void mkParent(File f) {
 
   /** Create this directory and all of its parents. */
   protected void mkDirs(File d) {
-    if (d != null && !d.exists() && !d.mkdirs()) {
+    if (d == null) {
```
staubesv, line 179: Why not 

```
// mkdirs *may* not have succeeded if it returns false, so check if the directory exists in that case
if (d != null && !d.mkdirs() && !d.exists()) {

```
?

lkiesow: Because that was my first change and the reaction to that was that it's hard to understand.
I hoped that removing the somewhat unrelated null check would help a bit while we don't lose anything except for two lines ;-)

staubesv:
@gregorydlogan Could you please give me the role "Committer" on Buildbot... I can't restart the build  :-(

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/AbstractFileSystemAssetStore.java
staubesv: ... and a somewhat weird use of the if statement.

Anyway, while I personally don't like if statements that jump out of the method, I don't have a strong opinion about that, so fine for me.

gregorydlogan:
@staubesv, in this case that won't actually help.  I've restarted the build manually, and it's using the same magic revision number and base branch as the original PR.  I think @lkiesow either needs to push something new (ie, not a force push), or refile.  Looking at this PR, I would merge it as is, the logic is sound.

staubesv (approved):

PR #987, opened 2019-06-28
MH-13651 Don't call submit of SingleSelect twice

KatrinIhler:
Leave edit mode if the drop down menu was closed, but only submit data if a user hits Enter or clicks on a value.


lkiesow (approved):
A short background to this issue: We have seen this on multiple systems and this can have quite bad side effects like broken metadata or even a corrupt asset manager (*sigh*).

Tested this and it seems to solve the issue.

PR #989, opened 2019-07-02
MH-13653, Fully Migrate Scheduled Events

lkiesow:
The 7.0 scheduler migration only migrated scheduled events which have
not been recorded yet, leaving the metadata of all other events lying
vacant in the asset manager properties store.

While this has no effect on recording and processing, it does mean that
this data is unavailable can cannot be used for filtering, etc. In
particular, this effects the external API where scheduling information
should be present:

```
  /api/events/1c009e2d-347f-4343-b356-3c07689c702f?withscheduling=true&...

  {
    "identifier": "1c009e2d-347f-4343-b356-3c07689c702f",
    ...
    "scheduling": {
      "agent_id": "pyca",
      "start": "2019-06-28T08:08:00Z",
      "end": "2019-06-28T08:09:00Z"
    },
    ...
  }
```

and the admin interface where you can filter events for capture agent
ids. Both will not work after the migration as those data have not been
migrated.

*Note that there is an additional bug in the admin interface filtering.*

*Work sponsored by SWITCH*

### DIFF
--- a/modules/migration/src/main/java/org/opencastproject/migration/SchedulerMigrationService.java
+++ b/modules/migration/src/main/java/org/opencastproject/migration/SchedulerMigrationService.java
@@ -154,8 +154,7 @@ public class SchedulerMigrationService {
   private Stream<ARecord> getScheduledEvents() {
     final AQueryBuilder query = assetManager.createQuery();
     // query filter for organization could be helpful to split up big migrations
-    final Predicate predicate = withOrganization(query).and(withVersion(query)).and(withOwner(query))
-        .and(withProperties(query));
+    final Predicate predicate = withOrganization(query).and(withVersion(query)).and(withProperties(query));
     // select necessary properties when assembling query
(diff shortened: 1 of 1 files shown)


gregorydlogan:
@lkiesow to be clear, this patch enables searching in the API for scheduling info, and searching in the admin UI or CA ids?  Or does it disable those?  Either way, I'm surprised a change like that is accomplished by such a small changeset!

lkiesow:
The problem is the data migration from the old, ≤ 6.x scheduler which stored its data in the asset manager's property store to the new, 7.x one which has its own database table.  The data source here is the asset manager from which we want to get all scheduler data (we want to migrate all data after all). But the `query.owner().eq(SNAPSHOT_OWNER);` causes the migration routine to just look at events which the scheduler “owns” which is only true for upcoming recordings.

In other words, the database filter is incorrect and only selects half of the data we want to migrate.

The described features are affected by the missing data and will hence not work in 7.0 for migrated events (they haven't been migrated after all). The admin interface filter also suffers from the index rebuild though, so that this patch alone does not fix the issue. But you should at least see scheduler data for already recorded and then migrated events in the external API.

PR #991, opened 2019-07-02
MH-13655, Scheduler Message Logging

lkiesow:
This patch fixes a number of log statements in the message receiver for
scheduler messages, ensuring that exceptions are logged properly and
simplifying the overall logging.

This is a byproduct of the other scheduler fixes and has no real consequences except for logging.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/message/SchedulerMessageReceiverImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/message/SchedulerMessageReceiverImpl.java
@@ -68,145 +68,103 @@ public class SchedulerMessageReceiverImpl extends BaseMessageReceiverImpl<Schedu
     User user = getSecurityService().getUser();
     logger.debug("Received message of type {} for event {}", schedulerItem.getType(), mediaPackageId);
 
-    switch (schedulerItem.getType()) {
-      case UpdateCatalog:
-        // Load or create the corresponding recording event
-        try {
+    try {
+      switch (schedulerItem.getType()) {
+        case UpdateCatalog:
+          // Load or create the corresponding recording event
           event = getOrCreateEvent(mediaPackageId, organization, user, getSearchIndex());
           if (isBlank(event.getCreator()))
             event.setCreator(getSecurityService().getUser().getName());
           if (dc != null)
             EventIndexUtils.updateEvent(event, dc);
-        } catch (SearchIndexException e) {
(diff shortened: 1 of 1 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/message/SchedulerMessageReceiverImpl.java
```diff
@@ -78,92 +78,77 @@ private void executeSingle(final String mediaPackageId, final SchedulerItem sche
           if (dc != null)
             EventIndexUtils.updateEvent(event, dc);
         } catch (SearchIndexException e) {
-          logger.error("Error retrieving the recording event from the search index: {}", getStackTrace(e));
+          logger.error("Error retrieving the recording event from the search index", e);
           return;
```
karendolan: @lkiesow  since you have the mpId already, it'd be very helpful to add it to the error log so there's less digging for context when this log surfaces.

modules/index-service/src/main/java/org/opencastproject/index/service/message/SchedulerMessageReceiverImpl.java
```diff
@@ -78,92 +78,77 @@ private void executeSingle(final String mediaPackageId, final SchedulerItem sche
...
         }
 
         // Update series name if not already done
         try {
           EventIndexUtils.updateSeriesName(event, organization, user, getSearchIndex());
         } catch (SearchIndexException e) {
-          logger.error("Error updating the series name of the event to index: {}", getStackTrace(e));
+          logger.error("Error updating the series name of the event to index", e);
```
karendolan: Please add mpId to help give context for the log.

karendolan:
HI @lkiesow since the mpId and sometimes  also the User is known at the point of the logging the ERROR, but is no longer available at the point of reading the ERROR from the log, it'd be really helpful to add the context at the point it is known. Otherwise it's annoying to read the ERROR in the log saying "the event failed catastrophically" without knowing which event or who or any context at all to justify the ERROR even being logged.

lkiesow:
Makes sense. Will add it to all the error logs.

lkiesow:
@karendolan, done. Looking at this again, I noticed that you could actually simplify the logging quite a bit since it was basically the same catch block which was repeated over and over again. There is now one. In case an error occurs, you now get a message including the message type, the event id, and the whole stack trace.

PR #992, opened 2019-07-02
MH-13547 switch to compatible file type filter definitions

doofy:
Previously used filter definitions using wildcards like "video/*"
resulted in wildly different behaviour in various browsers.

This patch defines a fixed set of file type extensions which hopefully
make browsers behave somewhat uniform when showing a file selection
dialog.


doofy:
Please feel free to criticize the selection of file extensions I chose.

etc/listproviders/event.upload.asset.options.properties
```diff
@@ -32,15 +32,15 @@ list.name=eventUploadAssetOptions
...
-EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SEGMENTABLE={"id": "track_presentation","type":"track", "flavorType": "presentation","flavorSubType": "source", "multiple":false, "displayOrder":14, "accept": "video/*,audio/*"}
+# EVENTS.EVENTS.NEW.SOURCE.UPLOAD.MULTIPLE_PARTS={"id": "track_parts","type":"track", "flavorType": "multipart","flavorSubType": "part+source", "multiple":true, "displayOrder":11, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv,.flac,.m4a,.mp3,.ogg,.wav,.wma"}
+# EVENTS.EVENTS.NEW.SOURCE.UPLOAD.AUDIO_ONLY={"id": "track_audio","type":"track", "flavorType": "presenter-audio", "flavorSubType": "source", "multiple":false, "displayOrder":12, "accept": ".flac,.m4a,.mp3,.ogg,.wav,.wma"}
+EVENTS.EVENTS.NEW.SOURCE.UPLOAD.NON_SEGMENTABLE={"id": "track_presenter","type":"track", "flavorType": "presenter","flavorSubType": "source", "multiple":false, "displayOrder":13, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv"}
+EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SEGMENTABLE={"id": "track_presentation","type":"track", "flavorType": "presentation","flavorSubType": "source", "multiple":false, "displayOrder":14, "accept": ".flac,.m4a,.mp3,.ogg,.wav,.wma"}
```
lkiesow: If I'm not misinterpreting things, these two lines are controlling the two default upload flavors (`presenter` and `presentation`) which should both be video and audio or maybe even just video. But in any case, none of these is pure audio.

lkiesow (changes requested):

etc/listproviders/event.upload.asset.options.properties
doofy: You are correct. I do not know how I mixed this up. Good catch!

Fixed.

lkiesow (approved):

PR #993, opened 2019-07-02
MH-13656, Fix Scheduler Index Rebuild

lkiesow:
The new scheduler has two problems which causes index rebuilds to not
work properly:

1. The index is rebuild only from objects which have the scheduler set
   as owner in the asset manager. This is true only for upcoming
   recordings but not for those which have already been processed,
   causing technical metadata to not be present in the indexes which in
   turn causes features like filtering for a capture agent identifier in
   the admin interface to fail.

2. Updates are sent to all scheduler receivers. This means that if
   someone rebuilds the admin interface Elasticsearch index, the
   messages are sent to all other indexes as well and the data are
   received by, processed and stored in the External API index as well.
   When doing a full rebuild, in the end, this cases twice the amount of
   messages to be sent for scheduler updates (essentially building both
   indexes twice).

This patch addresses both issues, fixing issue 1 by using the
scheduler's own database backend as data source instead of the asset
manager and fixing issue 2 by (simply) sending the correct message type
intended for the specific index.

*Work sponsored by SWITCH*

### DIFF
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
@@ -121,10 +121,8 @@ import net.fortuna.ical4j.model.property.RRule;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.text.WordUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
-import org.osgi.framework.ServiceException;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedService;
 import org.osgi.service.component.ComponentContext;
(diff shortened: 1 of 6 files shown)


modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -1592,85 +1591,70 @@ private void sendRecordingUpdate(Recording recording) {
...
-  public void repopulate(final String indexName) {
+  public void repopulate(final String indexName) throws SchedulerServiceDatabaseException {
     notEmpty(indexName, "indexName");
 
-    final String destinationId = SchedulerItem.SCHEDULER_QUEUE_PREFIX + WordUtils.capitalize(indexName);
-    final Organization organization = new DefaultOrganization();
-    final User user = SecurityUtil.createSystemUser(systemUserName, organization);
-    SecurityUtil.runAs(securityService, organization, user, () -> {
-      int current = 0;
+    final String destinationId = SchedulerItem.SCHEDULER_QUEUE_PREFIX + indexName.substring(0, 1).toUpperCase()
```
gregorydlogan, line 1597: Doesn't substring() return characters?  What if we end up with multiple indexes with the same first letter under the scheduler queue prefix?

lkiesow: `.substring(0, 1)` returns a string containing the first character of the input string and `.substring(1)` returns the string except for the first letter. In other words, the input here is something like `adminui` and the required output we need is `Adminui`. This is just copied from the workflow index rebuild btb ;-)

PR #997, opened 2019-07-07
MH-13660, Update Development Process Documentation

lkiesow:
The development process documentation has grown old in some places and
does not completely reflect current practice any longer. Since the
document was written, the version numbers have changed, we do not use
Bitbucket any longer and we usually do not cut beta releases or release
candidates.

Additionally, to updating that, this patch includes a Markdown extension
to write inline Graphviz code for rendering vector graphics. We've
discussed adding something like this a while ago and this seemed a good
opportunity to get it done.

For a rendered version of the page including the graphics, visit:

- [data.lkiesow.io/…/development-process/#git-repository-branching-model](https://data.lkiesow.io/opencast/docs-developer-2019-07-07/development-process/#git-repository-branching-model)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ jobs:
       language: python
       python: "3.6"
       install:
-        - pip install mkdocs
+        - pip install mkdocs markdown_inline_graphviz_extension
         - (cd docs/guides && npm install)
       script:
         - (! grep -rn '	' modules assemblies pom.xml --include=pom.xml)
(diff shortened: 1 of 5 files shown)


lkiesow:
@gregorydlogan, buildbot does not install the necessary Markdown extension and I cannot add it since [the buildbot set-up scripts](https://github.com/opencast/opencast-project-infrastructure/tree/master/ansible-buildbot-cluster) seem to be hopelessly out of date. Can you fix that?

EDIT: It seems like I also have no SSH access to that machine.

gregorydlogan:
The Markdown issue is being fixed, per #998.  Will have more to say about this one once I've read it :)

lkiesow:
Please do. Btb, I deliberately kept in Jira for now since we haven't switched yet. I've mostly tried sticking to including old proposals that never made it in here plus to do some clarifications.

PR #998, opened 2019-07-07
MH-13661, Concat Operation Graphics

lkiesow:
The concat operation still uses an old, blurry PNG image instead of
proper vector graphics. This patch uses the Graphviz markdown extension
to replace the old picture.

*Publishing this as pull request instead of directly merging this since this will break the buildbot builds and I have no access to buildbot (neither machine nor configuration)*


gregorydlogan:
Fixing the image now, will address outlined shortcomings via list mail.

lkiesow:
Nice, thanks @gregorydlogan!

PR #1002, opened 2019-07-22
MH-13665S3 S3 compatibility - Endpoint configuration for Amazon S3 alternatives added

isabellakutger:
'withEndpointConfiguration' and 'withPathStyleAccessEnabled' added in AwsS3AssetStore.java. Now you can use Amazon S3 alternatives such as Netapp


etc/custom.properties
```diff
@@ -12,7 +12,7 @@
 # Also note that if this felix installation is proxied behind an Apache HTTPD reverse proxy, and communication is meant
 # to go through that proxy, then server.url should point to the proxy's port (usually 80).
 # Related options (like listening addresses) are located in etc/org.ops4j.pax.web.cfg
-org.opencastproject.server.url=http://localhost:8080
+org.opencastproject.server.url=https://nfldevocaio.rus.uni-stuttgart.de
```
gregorydlogan: This needs to be reverted :)

etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
```diff
@@ -18,3 +18,9 @@ org.opencastproject.assetmanager.aws.s3.enabled=false
 # The AWS secret access key to use for connecting to S3
 #org.opencastproject.assetmanager.aws.s3.secret.key=
 
+
+#org.opencastproject.assetmanager.aws.s3.endpoint=
```
gregorydlogan, line 22: These keys should have some explanation of what they do, and whether they're safe to leave alone if you're using S3 directly.

etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
```diff
@@ -34,3 +34,8 @@ org.opencastproject.distribution.aws.s3.distribution.enable=false
 # Default restore job load: 0.1
 #job.load.aws.s3.restore=0.1
 
+#org.opencastproject.distribution.aws.s3.endpoint=
```
gregorydlogan, line 38: These keys should have some explanation of what they do, and whether they're safe to leave alone if you're using S3 directly.

etc/workflows/offload.xml
```diff
@@ -42,11 +42,12 @@
     <!-- Offload the archived mediapackage -->
 
     <operation
-      id="move-to-remote"
+      id="move-storage"
       if="${offloadToAWSS3}"
       description="Offloading to AWS S3">
       <configurations>
         <configuration key="target-storage">aws-s3</configuration>
+        <configuration key="target-version">0</configuration>
```
gregorydlogan: This likely should not be included.  Is there a good reason to add this restriction?  Ideally this value would be provided by the user in a workflow settings pane, although I'm going to guess that's outside of scope here!

modules/asset-manager-storage-aws/pom.xml
```diff
@@ -46,7 +46,7 @@
     <dependency>
       <groupId>org.apache.servicemix.bundles</groupId>
       <artifactId>org.apache.servicemix.bundles.aws-java-sdk</artifactId>
-      <version>1.11.49_1</version>
+      <!--<version>1.11.49_1</version>-->
```
gregorydlogan: This line should be removed.

modules/admin-ui/package-lock.json
```diff
@@ -7667,4 +7667,4 @@
       "dev": true
     }
   }
-}
+}
```
gregorydlogan: This file should not be included, even if it automatically changes itself (Node bugs).

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -123,10 +137,8 @@ public void activate(final ComponentContext cc) throws IllegalStateException, IO
                 new BasicAWSCredentials(accessKeyIdOpt.get(), accessKeySecretOpt.get()));
 
       // Create AWS client.
-      s3 = AmazonS3ClientBuilder.standard()
-              .withRegion(regionName)
-              .withCredentials(provider)
-              .build();
+      s3 = AmazonS3ClientBuilder.standard().withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(endpoint
```
gregorydlogan: The older formatting was much easier to read, please reapply.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -211,8 +227,12 @@ public void activate(ComponentContext cc) {
         provider = new AWSStaticCredentialsProvider(
                 new BasicAWSCredentials(accessKeyIdOpt.get(), accessKeySecretOpt.get()));
 
-      // Create AWS client.
-      s3 = AmazonS3ClientBuilder.standard().withRegion(regionStr).withCredentials(provider).build();
+      // Create AWS client
+
+      s3 =
+              AmazonS3ClientBuilder.standard().withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(endpoint
```
gregorydlogan: This should have the formatting from the AssetManager module applied to it.

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerStorageMoveOperationHandler.java
```diff
@@ -67,19 +67,25 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
     //A missing version is ok, that just means to select all of them (which is represented as null)
     String targetVersion = StringUtils.trimToNull(operation.getConfiguration("target-version"));
     Version version = null;
-    if (null != targetVersion) {
+    if (null != targetVersion && !"*".equals(targetVersion)) {
```
gregorydlogan: I'm failing to see the usecase for the `*` flag here?  What does that get you that passing a null does not?

I don't necessarily oppose the flag's existence, but I feel like the function is well enough documented that adding such a flag is unnecessary.  Unnecessary code tends to get poorly tested, or missed entirely.  If we're going to keep it we also need to document its existence in the Markdown docs.

gregorydlogan (changes requested):

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/AssetManagerStorageMoveOperationHandler.java
isabellakutger: Thanks for reviewing:)
this flag was added, because we needed the use case that all versions of  mediapackage can be moved at once. According to the documentation, the `target-version` should simply be omitted and all versions of the mediapackage are moved to the cloud storage, but an error occurred while trying out. The reason was the lack of a `target-version` in the workflow operation. Therefore, the flag was introduced to indicate that all versions of a media package should be moved to the cloud storage.

gregorydlogan: Whoops, that's a mistake on my part.  I'll have a PR out shortly, and then you should be able to drop the `*` flag.

Edit: This PR is now out at #1021

gregorydlogan:
Blocked pending #1021, will revisit once that goes in.

gregorydlogan:
@isabellakutger if you can get this fixed up before Friday I should be able to get it into the mainline codebase.

isabellakutger:
@gregorydlogan sorry for not noticing your comment earlier, now the flag is fixed and everything should work

gregorydlogan:
@isabellakutger Code looks good, but there are still a few outstanding comments in `etc` to address.

isabellakutger:
> @isabellakutger Code looks good, but there are still a few outstanding comments in `etc` to address.

done:) The explanation of the keys in the config-files were already added. I don't know why there is no "outdated" flag on theses comments

PR #1005, opened 2019-07-22
MH-13668 Fix dropdown menus

KatrinIhler:
Currently dropdown menus get broken if the first thing you do is click outside of the menu so it closes. This is because of an error in custom code in chosen-jquery, where we try to reset the highlighting by searching for an element that no longer exists, since we use ng-if to remove the select when we leave edit mode.
This is solved by first checking if the element is defined.


KatrinIhler:
@gregorydlogan @slampunk Maybe you can take a look at this since you both worked on the custom code in chosen?

lkiesow:
Fixes the issue → merged.

PR #1014, opened 2019-08-01
MH-13676, Don't start opencast on a used port

ts23:
Stop opencast if it is started wanting to use ports that are used by other processes


lkiesow:
Hi @ts23, before I start complaining about anything and completely miss the point: What's your intention behind this? From what you said in the technical meeting, it sounded like failing fast on your dev system, not having to wait for Opencast to start up and check the log. Is that correct?

ts23:
Yes, verify the ports opencast wants to use  and bail out if they are used without hitting the logs.

lkiesow:
But this is meant as a quick check for a development environment, right?

ts23:
it doesn't harm when its in a live environment as opencast cant run anyway if the ports are used

assemblies/dist-allinone/src/main/resources/bin/setenv
```diff
@@ -38,3 +38,33 @@ export EXTRA_JAVA_OPTS="-Dorg.eclipse.jetty.server.Request.maxFormContentSize=15
...
+
+usehttp=`grep "org.osgi.service.http.enabled[ =]" etc/org.ops4j.pax.web.cfg`
+if [ ${usehttp##*=} = "true" ]; then 
+  httpport=`grep "org.osgi.service.http.port[ =]" etc/org.ops4j.pax.web.cfg`
+  if [ -x /bin/netstat ]; then
+    used=`/bin/netstat -lntp 2>/dev/null | grep -c ${httpport##*=} 2>/dev/null`
+    if [ $used = "1" ]; then
+      echo "\nHttp server port \e[1m${httpport##*=}\e[0m already in use \n"
+      echo "reconfigure \e[1morg.osgi.service.http.port\e[0m in \e[1metc/org.ops4j.pax.web.cfg\e[0m \n"
+      exit
```
mtneug: Maybe exit with an error code like `exit 1` (also for HTTPS)?

assemblies/dist-allinone/src/main/resources/bin/setenv
```diff
@@ -38,3 +38,33 @@ export EXTRA_JAVA_OPTS="-Dorg.eclipse.jetty.server.Request.maxFormContentSize=15
...
 export KARAF_NOROOT=true
+
+
+# -----------------------------------------------------------------------
+# check if http or https port is available to run opencast otherwise exit
+
+usehttp=`grep "org.osgi.service.http.enabled[ =]" etc/org.ops4j.pax.web.cfg`
+if [ ${usehttp##*=} = "true" ]; then 
+  httpport=`grep "org.osgi.service.http.port[ =]" etc/org.ops4j.pax.web.cfg`
+  if [ -x /bin/netstat ]; then
```
mtneug: Two remarks:

* Could this be located in another location (e.g. `/usr/bin`)? Probably better to check with `[ -x "$(command -v netstat)" ]` to use the `$PATH`.
* The Docker images currently do not include `netstat`. Not a problem there as the container interface should never have port conflicts, but can we assume that `netstat` is usually installed?

mtneug (changes requested):

assemblies/dist-allinone/src/main/resources/bin/setenv
ts23: If netstat is not installed we fall trough to the current behaviour, so we try to start opencast anyway

lkiesow:
If this is meant for production I'll start complaining quite a bit more ;-P

1. `netstat` is not installed on most systems we use. We would need to add it to the documentation as well as to the packages as a new dependency. Could we use the `/proc/` file system instead for checking the ports?
2. Never hard-code any paths. Whatever tool you call, it should be in the path.
3. The configuration location is dependent on `karaf.etc`. Not sure if the path is already set at this point in the `setenv`. We might need to check Karaf's startup scripts.
4. If `netstat` (or whatever we use in the end) is not present, we should just skip over the test
5. You are checking for some ports, not all. Opencast uses a few more (e.g. Elasticsearch)
6. Quotes are missing at some places. I suggest checking [the output of `shellcheck`](https://paste.fedoraproject.org/paste/xy7KJDsuUh8iGIzqAtIXrQ) for details.
7. It's somewhat weird to have this script in a file which is more or less part of the user configuration.

---

Thinking about this again (that's also why I asked earlier if this is meant for the development distribution) I was wondering if it would make more sense to simplify this to just check the HTTP port on the development distribution and skip over it in case anything doesn't work. That way we wouldn't need to be too strict and we could just check some defaults.

For production, we should be harder because it would be weird if the check works on one system and not the other.

So my suggestion would basically be to just add this to the development distribution and only fix points 2, 4 and 6

ts23:
1. I've just looked up using proc, and got to a way of doing this using awk and bc:

- can I assume awk and bc  on the systems?

2,6 - I'll fix those
4. is currently the case ... if $tool is not available the test will fall through to current behaviour

7.  I'd rather put it into "start_opencast.sh" but that seems to be controlled by karaf ... any suggestions?

assemblies/dist-allinone/src/main/resources/bin/setenv
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
lkiesow: This script is no executable and should not contain a shebang.
The same does for all other non-executables loaded vis `.`.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+# -----------------------------------------------------------------------
+# check if http or https port is available to run opencast otherwise exit
+
+if [ -d "${KARAF_ETC}" ]; then
+  SETENV_ETC="${KARAF_ETC}"
+else
+  SETENV_ETC="etc"
```
lkiesow: This is not correct since the scripts can be executed from anywhere
Karaf's scripts define it as `KARAF_ETC=${KARAF_BASE}/etc` …after loading `KARAF_BASE`.

assemblies/resources/bin/setenv
```diff
@@ -35,4 +36,9 @@
 # export KARAF_REDIRECT    # Enable/set the std/err redirection when using bin/start
 
 export EXTRA_JAVA_OPTS="-Dorg.eclipse.jetty.server.Request.maxFormContentSize=1500000 -Dfile.encoding=UTF-8"
+export JAVA_MAX_MEM=1G
+export JAVA_PERM_MEM=256M
```
lkiesow, line 39: Does this not make the file identical to the other ones? In that case, we could just drop the other files.

lkiesow (changes requested):
I realized that `setenv` is unfortunately also executed when we try to stop Opencast, causing this attempt to fail:

```
% ./bin/stop-opencast

Http server port 8080 already in use 
reconfigure org.osgi.service.http.port in etc/org.ops4j.pax.web.cfg
```

We could do something like:

```
PROGNAME=$(basename "$0")
```

…and only run the check if `PROGNAME` is `start-opencast`.

---

An alternative to `setenv` would be to modify the `start-opencast` or to not rename `karaf` to`start opencast` but call it from a custom script we call `start-opencast` which also does this check.

todo for that would be:

- remove the renaming of `bin/karaf` to `bin/start-opencast`
- add a custom `bin/start-opencast` whcih
    - executes the check
    - calls `bin/karaf`

assemblies/resources/bin/check_ports
lkiesow: maybe you could use the `bin/inc` to load all necessary Karaf variables.

assemblies/resources/bin/setenv
ts23: no it sources bin/check_ports in its last line

assemblies/resources/bin/check_ports
ts23: is bin inc executed?

assemblies/resources/bin/setenv
lkiesow: But the others also contain that line. I think you might have misunderstood me. Adding the `…MEM…` lines, the file `assemblies/resources/bin/setenv` is now identical to `assemblies/dist-develop/src/main/resources/bin/setenv` and `assemblies/dist-allinone/src/main/resources/bin/setenv`. Before these have been special cases which would override the default `setenv`. This does not make any sense any longer if now all files contain the same content.

assemblies/resources/bin/check_ports
lkiesow: It's not executed but included to load/set the Karaf variables for multiple Karaf files (of which we only use one):

```
bin/start-opencast:. "${DIRNAME}/inc"
```

assemblies/resources/bin/setenv
ts23: You are right they were in for some reason before

lkiesow: Generally, I wouldn't mind adding those settings to all distributions in which case we should just drop the *special-case*  files. We just shouldn't have several copies of the same identical file overwrite each other ;-)

ts23: just updated (it was in your version before ;-) )

assemblies/resources/bin/check_ports
ts23: should I move the check_ports to inc?

lkiesow: hm, need to check where `inc` is actually loaded but it might work to just extend the file in the same manner we do that with some other files already:

https://github.com/opencast/opencast/blob/18c4bb9c713f9f22002425f7f313101eb32f96f7/assemblies/pom.xml#L140-L146

assemblies/resources/bin/setenv
```diff
@@ -35,4 +35,9 @@
...
```
mtneug, line 39: Are `JAVA_PERM_MEM`, `JAVA_MAX_PERM_MEM`, and `KARAF_REDIRECT` still used? I can't find any references in any of the `bin` files.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,69 @@
...
+	if  [ -f "$i" ] && [ -r "$i" ]; then
+	   o=$(grep -c ":$hexPort " "$i")
+	   if [ "$o" = "1" ]; then
+		return 1;
+	   fi
+	fi
+    done
+  fi
+  return 0;
+}
```
mtneug: Could you convert tabs to spaces (or vice versa).

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,69 @@
...
+  SETENV_ETC="$1"
+else
+  if [ -d "${KARAF_BASE}" ]; then
+    SETENV_ETC="${KARAF_BASE}/etc"
+  else
+    SETENV_ETC="etc"
+  fi
+fi
+
+check_port() {
```
mtneug: I tried this function on an Ubuntu system and couldn't get it to reliable work:
```
$ uname -a
Linux [...] 4.15.0-54-generic #58-Ubuntu SMP Mon Jun 24 10:55:24 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ netstat -tulpn
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      870/systemd-resolve
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1204/sshd
tcp        0      0 0.0.0.0:6783            0.0.0.0:*               LISTEN      1596/weaver
tcp        0      0 127.0.0.1:6784          0.0.0.0:*               LISTEN      1596/weaver
tcp6       0      0 :::80                   :::*                    LISTEN      1278/dockerd
tcp6       0      0 :::22                   :::*                    LISTEN      1204/sshd
tcp6       0      0 :::443                  :::*                    LISTEN      1278/dockerd
tcp6       0      0 :::6782                 :::*                    LISTEN      1596/weaver
tcp6       0      0 :::2377                 :::*                    LISTEN      1278/dockerd
tcp6       0      0 :::7946                 :::*                    LISTEN      1278/dockerd
udp        0      0 0.0.0.0:4789            0.0.0.0:*                           -
udp        0      0 0.0.0.0:6783            0.0.0.0:*                           1596/weaver
udp        0      0 0.0.0.0:6784            0.0.0.0:*                           -
udp        0      0 127.0.0.53:53           0.0.0.0:*                           870/systemd-resolve
udp6       0      0 :::6784                 :::*                                -
udp6       0      0 :::7946                 :::*                                1278/dockerd
...

$ check_port 22    # -> 0
$ check_port 53    # -> 0
$ check_port 68    # -> 0
$ check_port 80    # -> 0
$ check_port 443   # -> 0
$ check_port 546   # -> 0
$ check_port 2377  # -> 0

$ check_port 4789  # -> 1
$ check_port 6782  # -> 1
$ check_port 6783  # -> 1
$ check_port 6784  # -> 1
$ check_port 7946  # -> 1
```

mtneug (changes requested):

assemblies/resources/bin/check_ports
ts23: Thanks ... proc stores the port numbers as 4 digit Hex, I'll fix this

assemblies/resources/bin/check_ports
ts23: will be spaces in the end

assemblies/resources/bin/setenv
mtneug: `JAVA_MAX_MEM` is used in `inc` and defaults to `512M`, so setting it to `1G` is probably better.

assemblies/resources/bin/check_ports
mtneug: This is working correctly now.

assemblies/resources/bin/check_ports
ts23: I'm checking if its run from `start-opencast` -  otherwise we have to gather all the variables from scratch as `inc` is run before  `setenv`

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,74 @@
...
+      SETENV_ETC="${KARAF_BASE}/etc"
+    else
+      SETENV_ETC="etc"
+    fi
+  fi
+
+  check_port() {
+    if [ "$1" ] ; then
+      hexPort=$(printf ":%04X " "$1")
+      for i in /proc/net/*; do
```
miesgre: This will not work on not linux machines, for example OSX (there is no /proc filesystem)

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,74 @@
...
+          fi
+        fi
+      done
+    fi
+    return 0;
+  }
+
+  usehttp=$(grep "org.osgi.service.http.enabled[ =]" "${SETENV_ETC}"/org.ops4j.pax.web.cfg)
+  if [ "${usehttp##*=}" = "true" ]; then
+    httpport=$(grep "org.osgi.service.http.port[ =]" "${SETENV_ETC}"/org.ops4j.pax.web.cfg)
```
miesgre, line 56: Please does not suppose that org.osgi.service.http.port is a number as karaf allow you to use other expressions

For example (to use the value of a environment variable): 
```
org.osgi.service.http.port = ${env:OC_HTTP_PORT}
```

Karaf doc: http://karaf.apache.org/manual/latest/#_files

miesgre (changes requested):

assemblies/resources/bin/check_ports
ts23: is there any way to find out from the system which ports are in use on a OSX?

ts23: can I rely on netstat or lsof being available on OSX?

mtneug: Apple ships both (`/usr/sbin/lsof`, `/usr/sbin/netstat`) although I'm not sure if you have to install the dev tools. I will check with a fresh macOS install.

mtneug: So a fresh install of macOS includes the tools in the mentioned locations.

assemblies/resources/bin/check_ports
ts23: is now checked and if it resolves to an environment variable containing a Port number this is used 
if it doesn't opencast tries to start as before and (if the port is in use) will fail as before

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+        if  [ -f "$i" ] && [ -r "$i" ]; then
+          o=$(grep -c "$hexPort" "$i")
+          if [ "$o" = "1" ]; then
+            return 1;
+          fi
+        fi
+      done
+    else
+        if [ -x "$(command -v netstat)" ]; then
+          o=$(netstat -anp tcp | grep -i "listen" | grep -c ".$port ")
```
mtneug: The dot in `grep -c ".$port "` will match anything on my system. So e.g. when I call `check_port 80` any port ending in 80 will be a match such as `9080`. Could you add the `-F` to `grep` that it searches with a fixed string?

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+    httpsport=$(grep "org.osgi.service.http.port.secure[ =]" ${SETENV_ETC}/org.ops4j.pax.web.cfg)
+    check_port "${httpsport##*=}"
+    if [ "$?" = "1" ]; then
+      printf "\nSecure http server port \e[1m%s\e[0m already in use \n" "${httpsport##*=}"
+      printf "reconfigure \e[1morg.osgi.service.http.port.secure\e[0m in \e[1m%s/org.ops4j.pax.web.cfg\e[0m \n" "${SETENV_ETC}"
+      exit 1
+    fi
+  fi
+
+fi
```
mtneug: Could you add a newline.

assemblies/resources/bin/check_ports
mtneug: Apart from the problem below, the function works on macOS.

mtneug (changes requested):

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/setenv
```diff
@@ -35,4 +35,9 @@
...
 export KARAF_NOROOT=true
+
+
+. bin/check_ports "$KARAF_ETC"
```
lkiesow: if you start the script from a different directory it now completely fails:

```
% ./build/opencast-dist-develop-8-SNAPSHOT/bin/start-opencast
…/opencast/build/opencast-dist-develop-8-SNAPSHOT/bin/setenv: line 43: bin/check_ports: No such file or directory
```

This include needs to be relative to `KARAF_HOME` or `KARAF_BASE` if I remember correctly… or relative to `start_opencast` I guess.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
lkiesow: This script is not executed but included. It shouldn't habe a shebang and should not be executable:

```
% ls -lh bin
total 40K
-rwxrwxr-x. 1 lars lars 3.1K Aug 26 21:19 check_ports
-rw-r--r--. 1 lars lars 9.6K Aug 26 21:24 inc
-rw-rw-r--. 1 lars lars 1.8K Aug 26 21:19 setenv
…
```

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+# -----------------------------------------------------------------------
+# check if http or https port is available to run opencast otherwise exit
+
+PROGNAME=$(basename "$0")
+
+if [ "$PROGNAME" = "start-opencast" ]; then
```
lkiesow: This check is not sufficient since you can pass arguments to the main script which make it do other things than to just start Opencast. E.g. consider Opencast is running and you execute:

```sh
% ./bin/start-opencast status

Http server port 8080 already in use 
reconfigure org.osgi.service.http.port in etc/org.ops4j.pax.web.cfg
```

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+
+  if [ -d "$1" ]; then
```
lkiesow: `$1` is either `$KARAF_ETC` or it's empty. You could check that directly without passing it as argument which may have some side-effects (or not work at all).

assemblies/resources/bin/setenv
lkiesow: I think overwriting existing arguments is bash specific and may not work in different environments. See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot

You actually do not need to pass this argument at all but since you also have access to `"$KARAF_ETC"` in `check_ports` if it is set here.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+    SETENV_ETC="$1"
+  else
+    if [ -d "${KARAF_BASE}" ]; then
+      SETENV_ETC="${KARAF_BASE}/etc"
+    else
+      SETENV_ETC="etc"
```
lkiesow: I think this also wouldn't work if executed from a different path.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+      b="${a##*:}"
+      port=$(eval echo "\$${b%%\}*}")
+      if [ "$port" -eq "$port" ] 2>/dev/null ; then
+        hexPort=$(printf ":%04X " "$port")
+      else
+        return 0;
+      fi
+    fi
+    if [ -d /proc/net ] ; then
+      for i in /proc/net/*; do
```
lkiesow: Opencast only binds to TCP ports, doesn't it? So this should be much simpler and faster:

```sh
if  grep -q "$hexPort" /proc/net/tcp*; then
  return 1
…
```

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+  else
+    if [ -d "${KARAF_BASE}" ]; then
+      SETENV_ETC="${KARAF_BASE}/etc"
+    else
+      SETENV_ETC="etc"
+    fi
+  fi
+
+  check_port() {
+    a="$1"
```
lkiesow: `a` isn't the best name for any variable…

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+    if [ "$a" -eq "$a" ] 2>/dev/null ; then
+      port="$a"
+      hexPort=$(printf ":%04X " "$a")
+    else
+      b="${a##*:}"
+      port=$(eval echo "\$${b%%\}*}")
```
lkiesow: I'm not quite sure what you are trying to do here but the `eval` here is dangerous since this can lead to arbitrary code execution based on the pax web configuration file.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+        if  [ -f "$i" ] && [ -r "$i" ]; then
+          o=$(grep -c "$hexPort" "$i")
+          if [ "$o" = "1" ]; then
+            return 1;
+          fi
+        fi
+      done
+    else
+        if [ -x "$(command -v netstat)" ]; then
+          o=$(netstat -anp tcp | grep -i "listen" | grep -c -F ".$port ")
```
lkiesow: `netstat -anp tcp` does not work on my system since the parameters do not all exist. But assuming that it would work, you would also want to use something like this instead of counting lines:

```sh
if netstat -anp tcp | grep -i "listen" | grep -q -F ".$port "; then
  return 1;
…
```

…there is no need to call an additional test since grep already does the test.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+          if [ "$o" -gt 0 ] ; then
+            return 1;
+          fi
+        else
+          return 0;
```
lkiesow: no need for `else` since `return 0` is called anyway.

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+    check_port "${httpport##*=}"
+    if [ "$?" = "1" ]; then
+      printf "\nHttp server port \e[1m%s\e[0m already in use \n" "${httpport##*=}"
+      printf "reconfigure \e[1morg.osgi.service.http.port\e[0m in \e[1m%s/org.ops4j.pax.web.cfg\e[0m \n" "${SETENV_ETC}"
+      exit 1
+    fi
+  fi
+
+  usehttps=$(grep "org.osgi.service.http.secure.enabled[ =]" ${SETENV_ETC}/org.ops4j.pax.web.cfg)
+  if [ "${usehttps##*=}" = "true" ]; then
```
lkiesow: The Pax Web interpretation of this value is case insensitive. Hence, it could also be `TrUe` and there could be spaces after the `=`. Maybe something like:

```sh
if grep -qi '^org.osgi.service.http.secure.enabled[ =]*true *$' "${SETENV_ETC}/org.ops4j.pax.web.cfg"; then
…
```

assemblies/resources/bin/check_ports
```diff
@@ -0,0 +1,94 @@
...
+            return 1;
+          fi
+        else
+          return 0;
+        fi
+    fi
+    return 0;
+  }
+
+  usehttp=$(grep "org.osgi.service.http.enabled[ =]" "${SETENV_ETC}"/org.ops4j.pax.web.cfg)
```
lkiesow: The Pax Web interpretation of this value is case insensitive. Hence, it could also be `TrUe` and there could be spaces after the `=`. Maybe something like:

```sh
if grep -qi '^org.osgi.service.http.enabled[ =]*true *$' "${SETENV_ETC}/org.ops4j.pax.web.cfg"; then
…
```

lkiesow (changes requested):
Sorry for finding more problems 🙈

assemblies/resources/bin/setenv
mtneug: While I'm all for POSIX compliant scripts, Opencast (or Karaf) requires `bash` for quite some time already (see shebang of `stop-opencast`). Not sure if this is intentional, but I think it used to be `sh`.

assemblies/resources/bin/check_ports
mtneug: Not sure if the arguments are macOS (or BSD) specific. The improved check works for me on macOS.

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/setenv
ts23: fixed

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/setenv
ts23: fixed the parameter passing, can't do anything about needing bash as i'm using bash variable mangling

assemblies/resources/bin/check_ports
ts23: removed eval, start_opencast will behave as before if port is not given as a number

assemblies/resources/bin/check_ports
ts23: these params were tested on macOS, as linux will use proc

assemblies/resources/bin/check_ports
ts23: removed

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/check_ports
ts23: fixed

assemblies/resources/bin/check_ports
ts23: fixed

lkiesow (approved):
Looks good now → merged.
Thanks for dealing with all my complaining, @ts23.

PR #1017, opened 2019-08-05
Securing Static Files

lkiesow:
> **tl;dr:** This patch adds a plug-in based authorization mechanism for
> delivering static files (e.g. videos) which, by default, requires no
> configuration and which adheres to Opencast's default access control
> mechanisms (users, roles, groups, and access control lists).

Opencast comes with a complex access management system with fine-grained
control over users, roles, groups of roles and access rights. But all of
this is blatantly ignored when it comes to serving static file content
like the video files. These are served by Opencast's static module which
is nothing but a file server.

Right now, these file's protection is the form of the URLs which
generally contain at least two UUIDs and is thus hard to guess
(comparable e.g. to the “Share via link” options in Google docs, …). The
problem of this is obviously that if an authorized user publishes the
direct URL to a file, all protection is lost.

To address this problem, Opencast has an implementation for a
token-based authorization system which can be configured to only allow
access with valid tokens which in turn are generated by a trusted entity
and which are valid only for a certain amount of time. The advantage of
this is that it can be used to have external systems authorize
downloads. The disadvantage is that it's hard to configure correctly,
that yet another authorization mechanism ties in the already existing
one and that there are problems with either granting or preventing
access in some edge cases. Can't we do better?

An intuitive solution for this problem would be that the file server
simply adheres to the access control lists set for a given media package
and only delivers files to users which are authorized to access this
content as part of their user rights and roles. So, let's do that.

The Problem
-----------

Opencast's static files server delivers files for all publication
channels. By default these are:

- /static/**api**/<media-package-id>/…
- /static/**engage-player**/<media-package-id>/…
- /static/**internal**/<media-package-id>/…
- /static/**oaipmh-default**/<media-package-id>/…

In other words, several components of Opencast are involved in this and
the module “static” does not know about them. It is not even guaranteed
that the form of URLs is always the same. The module could be used to
serve any path as long as the URL starts with `/static/`.

Solution
--------

To allow for flexibility, this patch allows arbitrary services to
optionally register themselves with “static”, providing an authorization
interface which contains:

- A list of URL patterns, the authorization plug-in feels responsible
for (e.g. the search service could offer responsibility for
`/static/engage-player/…`)

- A method checking authorization based on a given URL (e.g. the search
service would check access based on the media package identifier
contained in the URL)

Delivering files for URLs which are not allowed by anyone would be
forbidden.

This allows for intuitive use of the authorization mechanisms already in
place for parts of the system without the need to do any complicated,
additional configuration. For example, we publish access control lists
to the engage components anyway and they are used for checking access to
content in the search service based on the user session. Expecting the
same mechanism for content as well would be just natural.

Compatibility
-------------

Does this break the current usage? In most cases, it shouldn't since a
lot of people already rely on the access controls set by Opencast and
other users or systems shouldn't access any files right now.

Nevertheless, there might be the usual special case for which the system
should be easy to turn off, restoring the current behavior of Opencast.
The same can also be achieved by serving the static content directly via
an HTTP server. By default, however, the protection should be turned on
in the long run since this is like the more intuitive understanding when
it comes to protecting content.

An additional way to deal with custom downloads, e.g. for custom
publication channels, would be to provide a simple authorization plug-in
with configurable patterns which can be used to allow additional content
to be downloaded without any specific checks.

Roadmap
-------

This patch targets Opencast 9.2. To ensure a smooth upgrade path and
compatibility to previous 9.x releases, all checks are still off by
default. This means there is no difference running Opencast with or
without this patch in its default configuration.

But to meet users expectation when it comes to securing those files, the
mechanism should be turned on by default in Opencast 10.

For now, turning on the protection can be done in
`etc/org.opencastproject.fsresources.StaticResourceServlet.cfg`

_EDIT: Updated with latest patch_


modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -734,4 +758,41 @@ public void updated(@SuppressWarnings("rawtypes") Dictionary properties) throws
...
+    final User user = securityService.getUser();
+    if (user.hasRole(GLOBAL_ADMIN_ROLE)) {
+      logger.debug("Allow access for admin `{}`", user);
+      return true;
+    }
+
+    final Matcher m = staticFilePattern.matcher(path);
+    if (!m.matches()) {
+      return false;
+    }
```
karendolan, line 799: A DEBUG line in the failed pattern match and in the failed org match might save some sites some headaches.

mtneug:
How would stream security/token authorization be implemented in such a plug-in based solution? Does it run orthogonally to the plug-ins? One property of token authorization I find really desirable is that the user does not need to be authenticated to Opencast. With an ACL check required, this is of course lost. Is it possible to still support the old model?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
lkiesow: Please refrain from complaining about any minor things at the moment. This code is meant to show how an implementation of this proposal would look like. There is more missing and complaining about minor stuff right now will serve no purpose except for distracting from the main question of this idea being acceptable or not ;-P

lkiesow:
@mtneug, please read the compatibility section again ;-p

This does not disable any token authentication and you can (even in this example implementation) already allow users to make requests without being authenticated if you prefer that.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
lkiesow: Obviously, if you find a problem in the method's architecture, please complain!

mtneug:
@lkiesow I understood the "current usage" to speak about having no protection at all. Thanks for clearing that up 🙂.

karendolan:
> @mtneug, please read the compatibility section again ;-p
> 
> This does not disable any token authentication and you can (even in this example implementation) already allow users to make requests without being authenticated if you prefer that.

@lkiesow Just to verify, you mean that granting "Anonymous" READ access to the series or event associated to the mediapackage ACL will continue to allow unauthenticated users to access to the static content.

lkiesow:
Okay, then maybe a few more words about token-based authorization (i.e. stream security) in Opencast and the way that works: If anyone generates a token for an allowed URL pattern in Opencast, that token is checked as part of the requests filters before it reaches any specific service. That way, the authorization information needs to be contained in the token. No access to any additional information is possible and no information about users are contained within the token.

So, it's basically just: Access is granted or not. Since this happens before the request reaches any specific service, you can protect any part of Opencast using this method even if the part is otherwise completely unprotected.

But this also means that whoever generates these token is responsible for the security and needs to verify proper access rights, … in addition to whatever Opencast does plus it needs to be compatible with what Opencast does (e.g. if Opencast disallows access but the token allows access, you will still not have access and vice versa). This usually means that if you do things properly, you need to build an external service to handle authentication and authorization and you might want to disable other protections in Opencast.

There is also the additional problem that you need some error margins for the tokens which might cause problems in the current form, but I leave that for now.

Regarding authorization tokens, we had an internal debate if switching to something like [jwt.io](https://jwt.io) would be worth it since it's a standard, you have libraries to properly handle these tokens in many languages and switching away from the old token system, we could fix a number of shortcomings. But short-term we don't plan to work on this.

I hope I answered most questions related to this proposal. In case we want to continue this discussion, let's do that on list ;-)

lkiesow:
@karendolan, no. I'm referring to

> **Compatibility**
>
> […] there might be the usual special case for which the system
should be easy to turn off, restoring the current behavior of Opencast […]

Which you can already do, even in my example implementation:

https://github.com/opencast/opencast/blob/ed760ef1834890d4099174c267b94ea7a6d281fa/etc/org.opencastproject.fsresources.StaticResourceServlet.cfg#L1-L4

karendolan:
@lkiesow what I'm really asking is that it's ok to leave events with ACL Anonymous READ and a custom auth module, used by the Search Service, will still be invoked to authenticate and test authorization of the user with a 3rd party system. I think this is the case because it looks like the static service calls the Search Service to perform the authorization task.

> @karendolan, no. I'm referring to
> 
> > **Compatibility**
> > […] there might be the usual special case for which the system
> > should be easy to turn off, restoring the current behavior of Opencast […]
> 
> Which you can already do, even in my example implementation:
> 
> https://github.com/opencast/opencast/blob/ed760ef1834890d4099174c267b94ea7a6d281fa/etc/org.opencastproject.fsresources.StaticResourceServlet.cfg#L1-L4

lkiesow:
@karendolan, I don't quite understand you sentence

>  what I'm really asking is that it's ok to leave events with ACL Anonymous READ and a custom auth module, used by the Search Service, will still be invoked to authenticate and test authorization of the user with a 3rd party system. 

Can you re-word that question? Sorry.

etc/org.opencastproject.fsresources.SimpleConfigurableStaticFileAuthorization.cfg
```diff
@@ -0,0 +1,8 @@
+# Configuration to allow any URL patterns to be served by Opencast's internal static file server without any further
+# checks for authentication.
+#
+# An arbitrary amount of patterns may be added using property keys in the form of `allow.<id>.pattern`
```
gregorydlogan, line 4: I feel like a link to the Java regex syntax might be a good idea here?

gregorydlogan: Also, documentation about this file is missing from the markdown docs.  I know this is unfinished, just making a list of things which you can easily cross off :)

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -163,11 +173,25 @@
   /** The optional Mediapackage serializer */
   protected MediaPackageSerializer serializer = null;
 
+  private LoadingCache<Tuple<User, String>, Boolean> cache = null;
+
+  private static final Pattern staticFilePattern = Pattern.compile("^/([^/]+)/engage-player/([^/]+)/.*$");
```
gregorydlogan, line 178: I was going to complain about constant here, but we're screwed no matter what if this has to change (grep for engage-player, seriously).  Let's just leave it as is.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -734,4 +758,41 @@ public void updated(@SuppressWarnings("rawtypes") Dictionary properties) throws
...
+    if (!m.matches()) {
+      return false;
+    }
+    final String organizationId = m.group(1);
+    if (!securityService.getOrganization().getId().equals(organizationId)) {
+      return false;
+    }
+    final String mediaPackageId = m.group(2);
+
+    logger.debug("Check if user `{}` has access to media package `{}` using cache", user, mediaPackageId);
```
gregorydlogan: IMO, this message should include what is being returned as well.

modules/static/src/main/java/org/opencastproject/fsresources/SimpleConfigurableStaticFileAuthorization.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public List<Pattern> getProtectedUrlPattern() {
+    return patterns;
+  }
+
+  @Override
+  public boolean verifyUrlAccess(String path) {
+    return true;
```
gregorydlogan, line 85: Shouldn't this be `false`?

modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
```diff
@@ -92,7 +137,11 @@ public StaticResourceServlet() {
    */
   @Activate
   public void activate(ComponentContext cc) {
-    if (cc != null) {
+    if (cc == null) {
+      authRequired = true;
```
gregorydlogan: Shouldn't this be `false` to preserve previous behaviour?  If not, then we need a note in the release docs, likely in big bold font.

modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
```diff
@@ -101,28 +150,59 @@ public void activate(ComponentContext cc) {
...
+    }
+
+    if (authRequired && !isAuthorized(path)) {
+      resp.sendError(HttpServletResponse.SC_FORBIDDEN);
+      logger.debug("Not authorized");
+      return;
+    }
+
+    final File file = new File(distributionDirectory, path);
+    if (!file.isFile() || !file.canRead()) {
```
gregorydlogan: If it's not a file then return a 404, but if it *is* a file but we can't read it (I'm assuming IO error, or filesystem ACLs) then we ought to either return a 503 (IO Error) or a 503/403 (filesystem ACL), IMO.

modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
```diff
@@ -101,28 +150,59 @@ public void activate(ComponentContext cc) {
...
     logger.debug("Looking for static resource '{}'", req.getRequestURI());
     String path = req.getPathInfo();
     if (path == null || path.contains("..")) {
       resp.sendError(HttpServletResponse.SC_FORBIDDEN);
       return;
     }
 
+    if (authRequired && !isAuthorized(path)) {
+      resp.sendError(HttpServletResponse.SC_FORBIDDEN);
+      logger.debug("Not authorized");
```
gregorydlogan, line 185: Logging the path here would be helpful IMO.

gregorydlogan (approved):
I like the idea, and it looks (relatively) scalable.  My comments are likely to be things already on your internal todo list, but I figured it wouldn't hurt to include them all anyway :)

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
lkiesow: As it's marked `private static final`, it's also technically constant ;-P
But yes, this is hard-coded in every media package with engage publication.

modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
lkiesow: I realize that this is confusing but it's intentional to properly work with the [nginx auth_request](http://nginx.org/en/docs/http/ngx_http_auth_request_module.html) which requires a 2xx, 401 or 403 response code. I add adding an example and explanation for this to my todo list :-)

modules/static/src/main/java/org/opencastproject/fsresources/SimpleConfigurableStaticFileAuthorization.java
lkiesow: No. Returning `true` means that access is granted.
The idea for this simple plug-in is to grant access for all configured patterns.

modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
lkiesow: Yep, this needs to be in the release docs. This may even deserve a section in the migration notes. Nevertheless, this should be `true` since this is likely what people expect: If I use the admin interface to define that only `ROLE_XY` should have access, that's what I expect file access is checked for.

Except for a few special cases, most adopters I know indirectly expect this already so changing this shouldn't be a problem for them. New adopters certainly expect that. For the few *special cases*, it's easy enough to change.

lkiesow: Just went through a few systems in my head and I couldn't remember anyone for which it should be a problem. At least, I'm reasonably sure that it wouldn't be a problem for UOS, SWITCH, ETH, Graz, Wismar, Bochum, Vechta, Oldenburg, Halle, Hannover, Ulm, HS Osna, … I'm stopping here ;-)

lkiesow:
I've resolved some conversations. This does not mean that I'm ignoring those comments. I've added the important points to the todo list above and will go through the comments again when I'm starting to work on the final implementation. For now, this just makes this more readable.

karendolan:
> @karendolan, I don't quite understand you sentence
> 
> > what I'm really asking is that it's ok to leave events with ACL Anonymous READ and a custom auth module, used by the Search Service, will still be invoked to authenticate and test authorization of the user with a 3rd party system.
> 
> Can you re-word that question? Sorry.

@lkiesow Our Opencast published asset Search Service invokes a custom auth handler that verifies the user authenticity and authorization status with a 3rd party by mapping the requested Opencast asset to a real course and lecture and the real student's University policy access to assets from that course. Something like LDAP authentication checks, but more granular, and also does the authorization check. This works because it is only invoked by the Opencast Search Service when the asset has Anonymous READ access in Opencast. Without looking deeper into your pull, I am curious if our strategy will be fine with this static asset auth pull because the static auth checks also go through the search service, even for assets with Anonymous READ rights.

lkiesow:
@karendolan, I don't know your code so everything is just guesswork but I bet it will work. For once, the static file authorization in my current implementation already checks with the search service. If you have an additional check in there it should just be invoked as well and access should be granted based oin the usual Opencast rules plus your additional rules. In case that does not work, you will be able to  configure that the search service is not asked for additional files but they are just delivered in the same way they are delivered right now which means that things work exactly as they work right now and you should have no problem.

etc/org.opencastproject.fsresources.StaticResourceServlet.cfg
```diff
@@ -0,0 +1,4 @@
+# Configuration if the static file server shall just deliver files or if authentication shall be checked for each
```
Rillke: Maybe `Shall the static file server check authentication before delivering a file?`

modules/common/src/main/java/org/opencastproject/security/api/StaticFileAuthorization.java
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * @return List of pattern matchers
+   */
+  List<Pattern> getProtectedUrlPattern();
+
+  /**
+   * Check a given path for access.
+   *
+   * @param path Path to check
+   * @return if access if allowed
```
Rillke: if access i_s_ allowed?

modules/static/src/main/java/org/opencastproject/fsresources/SimpleConfigurableStaticFileAuthorization.java
```diff
@@ -0,0 +1,91 @@
...
+        Matcher matcher = ALLOW_PATTERN.matcher(key);
+        if (matcher.matches()) {
+          //final String id = matcher.group(1);
+          final String patternString = Objects.toString(cc.getProperties().get(key));
+          newPattern.add(Pattern.compile(patternString));
+        }
+      }
+    }
+    patterns = newPattern;
+    logger.debug("Updated allow patterns: {}", newPattern);
```
Rillke: `Updated allowed patterns: {}` may or may not sound more conscious.

modules/static/src/main/java/org/opencastproject/fsresources/SimpleConfigurableStaticFileAuthorization.java
```diff
@@ -0,0 +1,91 @@
...
+    logger.info("Started authentication handler for {}", patterns);
```
Rillke, line 75: Logging the same parameters as info. May or may not by intention.

lkiesow:
It's done. This is no draft any longer.

gregorydlogan:
On the pres node, at the end of the default workflow
```
2019-09-26T10:06:09,948 | WARN  | (DownloadDistributionServiceImpl$1:351) - Attempt to access distributed file http://deb9vm1/static/mh_default_org/internal/fd4714b0-135d-4b01-addc-dec98c61a11a/ab6812e8-8a41-4f4b-ba56-c2a612e03761/composite.mp4 returned code 403
2019-09-26T10:06:09,949 | WARN  | (DownloadDistributionServiceImpl:359) - Error distributing http://deb9vm0/files/mediapackage/fd4714b0-135d-4b01-addc-dec98c61a11a/ab6812e8-8a41-4f4b-ba56-c2a612e03761/composite.mp4
org.opencastproject.distribution.api.DistributionException: Unable to load distributed file http://deb9vm1/static/mh_default_org/internal/fd4714b0-135d-4b01-addc-dec98c61a11a/ab6812e8-8a41-4f4b-ba56-c2a612e03761/composite.mp4
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl$1.xrun(DownloadDistributionServiceImpl.java:352) ~[?:?]
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl$1.xrun(DownloadDistributionServiceImpl.java:347) ~[?:?]
        at org.opencastproject.util.data.Effect$X.run(Effect.java:63) ~[45:opencast-common:8.0.0.SNAPSHOT]
        at org.opencastproject.util.data.Effect.apply(Effect.java:35) ~[45:opencast-common:8.0.0.SNAPSHOT]
        at org.opencastproject.util.data.Effect$X.apply(Effect.java:59) ~[45:opencast-common:8.0.0.SNAPSHOT]
        at org.opencastproject.util.data.Either$2.fold(Either.java:336) ~[45:opencast-common:8.0.0.SNAPSHOT]
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.distributeElement(DownloadDistributionServiceImpl.java:347) [52:opencast-distribution-service-download:8.0.0.SNAPSHOT]
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.distributeElements(DownloadDistributionServiceImpl.java:244) [52:opencast-distribution-service-download:8.0.0.SNAPSHOT]
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.process(DownloadDistributionServiceImpl.java:545) [52:opencast-distribution-service-download:8.0.0.SNAPSHOT]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314) [45:opencast-common:8.0.0.SNAPSHOT]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:273) [45:opencast-common:8.0.0.SNAPSHOT]
        at java.util.concurrent.FutureTask.run(FutureTask.java:266) [?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:?]
        at java.lang.Thread.run(Thread.java:748) [?:?]
2019-09-26T10:06:09,995 | INFO  | (ServiceRegistryJpaImpl:2563) - State set to WARNING for current service org.opencastproject.distribution.download on host http://deb9vm1
2019-09-26T10:06:10,079 | ERROR | (AbstractJobProducer$JobRunner:344) - Error handling operation 'Distribute': org.opencastproject.serviceregistry.api.ServiceRegistryException: Error handling operation 'Distribute'
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.process(DownloadDistributionServiceImpl.java:561)
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314)
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:273)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.opencastproject.distribution.api.DistributionException: Unable to load distributed file http://deb9vm1/static/mh_default_org/internal/fd4714b0-135d-4b01-addc-dec98c61a11a/ab6812e8-8a41-4f4b-ba56-c2a612e03761/composite.mp4
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl$1.xrun(DownloadDistributionServiceImpl.java:352)
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl$1.xrun(DownloadDistributionServiceImpl.java:347)
        at org.opencastproject.util.data.Effect$X.run(Effect.java:63)
        at org.opencastproject.util.data.Effect.apply(Effect.java:35)
        at org.opencastproject.util.data.Effect$X.apply(Effect.java:59)
        at org.opencastproject.util.data.Either$2.fold(Either.java:336)
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.distributeElement(DownloadDistributionServiceImpl.java:347)
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.distributeElements(DownloadDistributionServiceImpl.java:244)
        at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.process(DownloadDistributionServiceImpl.java:545)
        ... 6 more
```

lkiesow:
Which server is `deb9vm0`? Presentation?

KatrinIhler:
@lkiesow Conflicts!

gregorydlogan:
Happy to give this a test whenever things are ready.  Please ping me.

lkiesow:
Will do.

lkiesow:
A short update to this: I've rebased this to the current state of `develop` and fixed the conflicts but because of the possibility that publication URLs may point to different servers, I will also add remote implementations for some of the static file authorization services. I will hopefully get that done until the end of the week.

@gregorydlogan, I'll ping you once I'm done and this is ready for review.

gregorydlogan:
Dropping an on-hold on this so I don't keep checking.  Please remove the label when this is ready to go.

gregorydlogan:
Reminder: As a feature this has until 2020-10-05 to be merged for it to make it in 9.0!

lkiesow:
Working on this again.
I will change this slightly to address the multi-node problem. Instead of going through API, index service, et cetera, I will create a asset manager database based static file authorization service. Not only can we deploy that one on all nodes, it is also more flexible and replaces both the admin ui and the external api based one. Hope to have this done by the end of the week.

lkiesow:
This should be done by now. Compared to the last review which was fine except for the problems on distributed systems, I've changed the way access is checked for the internal and api publications. Instead of checking the api and the admin interface services, I check directly with the asset manager which is the source for access control lists in both services anyway.

This not only simplifies the code but also allows for checks on any node with a database connection (which are all Opencast nodes). So no more access problems on the presentation node :tada:

docs/guides/admin/docs/upgrade.md
```diff
@@ -89,3 +90,15 @@ but to benefit from this mechanism, users have to reset their password.
 
 You can use the endpoint `/user-utils/users/md5.json` to find out which users are still using MD5-hashed passwords and
 suggest to them that they update their passwords.
+
+
+Static File Delivery
+--------------------
+
+Opencast 8 comes with a [completely new system for securing static file content](configuration/serving-static-files.md)
```
gregorydlogan: Opencast 9 :)

modules/asset-manager-static-file-authorization/src/main/java/org/opencastproject/assetmanager/auth/AssetManagerStaticFileAuthorization.java
```diff
@@ -0,0 +1,182 @@
...
+        .map(Role::getName)
+        .filter(roleFilter)
+        .map((role) -> role + " | read")
+        .collect(Collectors.toList());  // ["ROLE_XY | read", ...]
+
+    StringBuilder properties = new StringBuilder("property_name = ?");
+    for (int i = 1; i < roles.size(); i++) {
+      properties.append(" or property_name = ?");
+    }
+    String sql = "select count(1) from oc_assets_properties "
```
gregorydlogan, line 155: Why was this done with a string query rather than the object based stuff the rest of the codebase uses?

gregorydlogan (approved):
Conditionally approved, with one question (see review).  I'm happy to fix the OC 8 reference when merging, assuming no further commits are required.

modules/asset-manager-static-file-authorization/src/main/java/org/opencastproject/assetmanager/auth/AssetManagerStaticFileAuthorization.java
lkiesow: Basically two reasons:

1. This cannot depend directly on the asset manager since this runs on different nodes and using a remote for this would have made it way too slow. So this would have basically meant re-implementing the whole thing for a single query or splitting the whole asset manager apart to get a usable base.
2. This is so much faster than what the asset manager does. Speed doesn't really matter there, but here it does since we do real-time requests for checking access to a lot of assets instead of the occasional asset during workflows.
3. This is simple and easy to understand once you get over the somewhat weird `ROLE_XY | read` entries we cannot change since they are already baked into the asset manager.

docs/guides/admin/docs/upgrade.md
lkiesow: Missed that statement. It's actually not true → It's not active by default, so you don't need to upgrade. There is nothing breaking in this patch. Changing the default may come later. Will remove this section for now.

modules/asset-manager-static-file-authorization/src/main/java/org/opencastproject/assetmanager/auth/AssetManagerStaticFileAuthorization.java
gregorydlogan: Ok.  I'm going to add a note to this code, and the asset properties object, linking to this discussion.  FTR, the reason I'm leery of this is that it opens the door to other developers doing things like this - especially with queries which write to the DB.  We had quite the discussion way back in the pre-1.0 days with some developers pushing hard for hand crafted SQL, and other pushing back for stability and portability.

In this one case I can see a good, solid reason to do this, but without a similarly compelling reason (ie, needs a massive reimpl, plus perf considerations) I would be *strongly* against further raw queries like this.

lkiesow: I agree. I guess, what we mostly need to prevent against are module boundary violations. For example, a statistics service accessing the asset manager table to get data. That could easily become a mess when you change asset manager internals in the future. Writing would be even worst since you are then tinkering with the internal state of another service which could have very weird side effects.

I see this as something slightly different since it is still marked as an asset manager module. As such, accessing the internal state of the asset manager should not be completely unexpected and given the way this works, it should be safe.

But putting that into writing is probably a good idea. I will add a proper explanation as comment to this.

lkiesow: Added an explanation:

https://github.com/opencast/opencast/blob/d4086846f3cd8f27f3f5691157bf88fad7c7a884/modules/asset-manager-static-file-authorization/src/main/java/org/opencastproject/assetmanager/auth/AssetManagerStaticFileAuthorization.java#L135-L145

PR #1018, opened 2019-08-06
MH-13680: Add upload and delete LTI tools

pmiddend:
# Abstract #

This PR adds two new functionalities to the LTI module: 

- Uploading of new events (as a new tool).
- Deletion of single events from the `series` tool; this is optional and can be enabled via a tool parameter.

## Uploading ##

The upload tool currently doesn’t really compare to the “New Event” modal in the admin UI. It was designer for a specific use case and is kept simple. It presents the user merely with the *event title* and a *file input*. Internally, a workflow can be configured which is executed after ingestion.

The upload tool also contains a list of running jobs, so the user gets some feedback about the upload completion and isn’t put off by the video not instantly appearing in the series tool.

## Deletion ##

Events can be deleted by pressing the trashcan symbol in the series list. To make this as simple as possible, this PR is based on the “single step event deletion” PR, so the user doesn’t have to unpublish, wait, and then delete.

# The Code #

## New modules ##

The `lti` module in Opencast is very minimalistic (which was sufficient for now). To support uploading and deletion, we had to write some Java code which needed services that are not necessarily available on the node LTI runs on. Our customer would like to run LTI on the presentation node, for example. To add multi-node support, we added the following modules and functions:

- `lti-service-api` contains the `LtiService` interface, implemented by the “remote” and the “impl”
- `lti-service-impl` contains the implementation
- `lti-service-remote` contains the remote
- `lti` contains an additional endpoint which holds an `LtiService` reference and serves as a “trampoline” to either the remote or the impl.

While reviewing, do note that the endpoint in `lti` looks strikingly similar to the remote endpoint. That’s, unfortunately, unavoidable.

## The user provider ##

~~This PR also adds a user provider for otherwise “half-known” temporary LTI users. Please see~~

~~https://groups.google.com/a/opencast.org/forum/#!topic/dev/YX_uCIUxfLA~~

~~For an in-depth discussion about this user provider and why it’s there.~~

The user provider has been removed due to the customer not needing it anymore :D

# Attribution #

This work was sponsored by Uni Gent.


modules/userdirectory/src/main/java/org/opencastproject/userdirectory/LtiUserProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Override
+  public Iterator<User> getUsers() {
+    Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+    return Collections.emptyIterator();
+  }
+
+  @Override
+  public User loadUser(String userName) {
+    if (!userName.startsWith("lti:")) {
```
mtneug: As I have already [pointed out](https://groups.google.com/a/opencast.org/d/msg/dev/YX_uCIUxfLA/m1p50M_iBgAJ), LTI can be [configured](https://docs.opencast.org/r/7.x/admin/modules/ltimodule/#configure-lti-optional) to trust the LTI provider. In this case, temporary LTI user names **do not** start with `lti:`! Depending on the LMS, the user name could be an ID or the actual user name in the LMS, as e.g. in Moodle.

Would it be possible to give users a special LTI role and differentiate based on that role?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/LtiUserProvider.java
```diff
@@ -0,0 +1,85 @@
...
+  public static final String PROVIDER_NAME = "lti";
+
+  @Override
+  public String getName() {
+    return PROVIDER_NAME;
+  }
+
+  @Override
+  public Iterator<User> getUsers() {
+    Authentication auth = SecurityContextHolder.getContext().getAuthentication();
```
mtneug: `auth` is not used?

mtneug (changes requested):

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/LtiUserProvider.java
pmiddend: I understand what you're saying, and a solution emplying a user provider for the LMS, plus highly trusted users is far better than this `LtiUserProvider` thingy. We'll talk to our customer and if they agree, I'll delete this provider from the PR.
I'm not sure where you'd like to differentiate via a role, though. Can you elaborate?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/LtiUserProvider.java
pmiddend: Ah yes, leftover code. I'll fix it shortly. :)

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/LtiUserProvider.java
mtneug: Now that I think about it, this won't work. I thought of adding a special role during the execution of `LtiLaunchAuthenticationHandler`, but this won't help you in a user provider that only gives you a name 🤦‍♂.

etc/org.opencastproject.lti.service.impl.LtiServiceImpl.cfg
```diff
@@ -0,0 +1,6 @@
+# Which workflow to use when uploading videos
+workflow=fast
```
lkiesow, line 2: While great for testing, we probably don't want to go with fast by default.

modules/lti-service-api/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
```
lkiesow, line 72: Can you add the maven-dependency-plugin to this:

https://github.com/opencast/opencast/blob/6618829c9117e7b340b6c9e5c830d02593b992c9/modules/lti/pom.xml#L222-L225

modules/lti-service-api/src/main/java/org/opencastproject/lti/service/api/Job.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.lti.service.api;
+
+public final class Job {
```
lkiesow, line 23: Naming this class Job may be a bit confusing considering the use of Job in other parts of Opencast. Can we find an alternative name for this?

lkiesow: Plus, can we add a one-line docs explaining what this class does?

modules/lti-service-api/src/main/java/org/opencastproject/lti/service/api/LtiService.java
```diff
@@ -0,0 +1,35 @@
...
+package org.opencastproject.lti.service.api;
+
+import java.io.InputStream;
+import java.util.List;
+import java.util.Map;
+
+public interface LtiService {
+  String JOB_TYPE = "org.opencastproject.lti.service";
+
+  List<Job> listJobs(String seriesName, String seriesId);
```
lkiesow, line 30: Can you add JavaDocs to the interface methods?

modules/lti-service-impl/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
```
lkiesow, line 133: dependency plugin, same as above.

modules/lti-service-impl/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+            <Build-Number>${buildNumber}</Build-Number>
+            <Import-Package>
+              org.opencastproject.matterhorn.search.impl;version=${project.version},
+              org.opencastproject.rest;version=${project.version},
+              org.opencastproject.index.service.message,
+              org.opencastproject.metadata.dublincore;version=${project.version},
```
lkiesow, line 146: Just wondering why we need to specifically import all of this.

modules/lti-service-impl/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore-osgi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.googlecode.json-simple</groupId>
+      <artifactId>json-simple</artifactId>
```
lkiesow, line 113: You are already using GSON. Is there any reason to use two JSON libraries?

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+    logger.info("Activating LTI service Endpoint");
+
+    final Tuple<String, String> endpointUrl = getEndpointUrl(cc, OpencastConstants.EXTERNAL_API_URL_ORG_PROPERTY,
+            RestConstants.SERVICE_PATH_PROPERTY);
+    endpointBaseUrl = UrlSupport.concat(endpointUrl.getA(), endpointUrl.getB());
+    logger.debug("Configured service endpoint is {}", endpointBaseUrl);
+  }
+
+  @GET
+  @Path("/jobs")
```
lkiesow, line 98: You could add 
```java
@Produces(MediaType.APPLICATION_JSON)
```

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+    }).collect(Collectors.toList());
+    return Response.status(Status.OK).entity(new Gson().toJson(results, List.class)).build();
+  }
+
+  @POST
+  @Path("/")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  @RestQuery(name = "createevent", description = "Creates an event by sending metadata, access control list, processing instructions and files in a multipart request.", returnDescription = "", restParameters = {
+          @RestParameter(name = "acl", isRequired = false, description = "A collection of roles with their possible action", type = STRING),
+          @RestParameter(name = "metadata", description = "Event metadata as Form param", isRequired = false, type = STRING),
```
lkiesow, line 114: What do you mean by `as Form param`?

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+                  @RestResponse(description = "The request is invalid or inconsistent..", responseCode = HttpServletResponse.SC_BAD_REQUEST) })
+  public Response createNewEvent(@HeaderParam("Accept") String acceptHeader, @Context HttpServletRequest request) {
+    String seriesName = "";
+    String seriesId = "";
+    Map<String, String> metadata = new HashMap<>();
+    try {
+      for (FileItemIterator iter = new ServletFileUpload().getItemIterator(request); iter.hasNext();) {
+        final FileItemStream item = iter.next();
+        final String fieldName = item.getFieldName();
+        if ("hidden_series_name".equals(fieldName)) {
```
lkiesow, line 131: `hidden_series_name`?

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+      return eventMap;
+    }).collect(Collectors.toList());
+    return Response.status(Status.OK).entity(new Gson().toJson(results, List.class)).build();
+  }
+
+  @POST
+  @Path("/")
+  @Consumes(MediaType.MULTIPART_FORM_DATA)
+  @RestQuery(name = "createevent", description = "Creates an event by sending metadata, access control list, processing instructions and files in a multipart request.", returnDescription = "", restParameters = {
+          @RestParameter(name = "acl", isRequired = false, description = "A collection of roles with their possible action", type = STRING),
```
lkiesow, line 113: Is `acl` actually used? Seems like you are hard-coding the ACL later.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,165 @@
...
+  /** OSGi DI */
+  public void setService(LtiService service) {
+    this.service = service;
+  }
+
+  /** OSGi activation method */
+  void activate(ComponentContext cc) {
+    logger.info("Activating LTI service Endpoint");
+
+    final Tuple<String, String> endpointUrl = getEndpointUrl(cc, OpencastConstants.EXTERNAL_API_URL_ORG_PROPERTY,
```
lkiesow, line 91: All this is not used at all and does not belong here

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,285 @@
...
+              .filter(e -> ZonedDateTime.parse(e.getCreated()).compareTo(startOfDay) > 0)
+              .map(e -> new Job(e.getTitle(), e.getEventStatus()))
+              .collect(Collectors.toList());
+    } catch (SearchIndexException e) {
+      throw new RuntimeException("search index exception", e);
+    }
+  }
+
+  @Override
+  public String upload(final InputStream file, final String sourceName, String seriesId, final String seriesName,
```
lkiesow, line 158: I've got multiple issues with this method but mainly, it's a bad and weird idea to mix the ingest and index services here. You can do all of this using ingest. No need for the index service. That should make the code simpler and be less work for Opencast since it does not need to jump through several loops to finally end up at the ingest service anyway.

modules/lti-service-remote/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
```
lkiesow, line 81: dependency plugin

modules/lti/src/main/java/org/opencastproject/lti/LtiServiceGuiEndpoint.java
```diff
@@ -0,0 +1,162 @@
...
+  /** OSGi DI */
+  public void setService(LtiService service) {
+    this.service = service;
+  }
+
+  /** OSGi activation method */
+  void activate(ComponentContext cc) {
+    logger.info("Activating LTI service Endpoint");
+
+    final Tuple<String, String> endpointUrl = RestUtil.getEndpointUrl(cc, OpencastConstants.EXTERNAL_API_URL_ORG_PROPERTY,
```
lkiesow, line 88: again, this isn't used, is it?

modules/lti/src/main/java/org/opencastproject/lti/LtiServlet.java
```diff
@@ -200,9 +200,11 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
     // We must return a 200 for some OAuth client libraries to accept this as a valid response
 
     // The URL of the LTI tool. If no specific tool is passed we use the test tool
-    UriBuilder builder = null;
+    UriBuilder builder;
     try {
-      URI toolUri = new URI(StringUtils.trimToEmpty(req.getParameter(LTI_CUSTOM_TOOL)));
+      String toolUriStr = req.getParameter(LTI_CUSTOM_TOOL);
+      toolUriStr = toolUriStr.replace(" ", "%20");
```
lkiesow, line 206: if you need to replace anything here, don't you need to do proper URL encoding?

modules/lti/src/main/resources/tools/series/bin.svg
```diff
@@ -0,0 +1,42 @@
...
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Capa_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 width="459px" height="459px" viewBox="0 0 459 459" style="enable-background:new 0 0 459 459;" xml:space="preserve">
+<g>
+	<g id="delete">
+		<path d="M76.5,408c0,28.05,22.95,51,51,51h204c28.05,0,51-22.95,51-51V102h-306V408z M408,25.5h-89.25L293.25,0h-127.5l-25.5,25.5
+			H51v51h357V25.5z"/>
+	</g>
+</g>
+<g>
```
lkiesow, line 12: This is obviously unnecessary.
Same image just with aria labels and without all the rubbish:

```xml
<svg xmlns="http://www.w3.org/2000/svg"
aria-label="delete" role="img"
viewBox="0 0 459 459"><path
d="m76.5 408c0 28.05 22.95 51 51 51h204c28.05 0 51-22.95 51-51v-306h-306zm331.5-382.5h-89.25l-25.5-25.5h-127.5l-25.5 25.5h-89.25v51h357z"/></svg>
```

modules/lti/src/main/resources/tools/series/index.html
```diff
@@ -21,8 +21,8 @@
 <main></main>
```
lkiesow, line 21: New ui left, old ui right. I don't think the new one is supposed to look like this, is it?

Screenshot from 2019-09-27 00-54-31

docs/guides/admin/docs/modules/ltimodule.md
```diff
@@ -86,7 +90,11 @@ custom parameters to be defined separately in each place where an LTI tool is us
 custom parameters to be defined globally.
 
 - To show the media module, use `tool=engage/ui/`
-- To show all videos for a single series, use `tool=ltitools/series/index.html?series=SERIESID`
+- To show all videos for a single series, use `tool=ltitools/series/index.html?series=SERIESID`. You can also pass
+  `series_name` to select a series by name (the name has to be unique). If you want the ability to delete events
+  from the series (which will only be available for instructors), add `&deletion=true` to the URL.
+* To show an upload form, use `tool=ltitools/upload/index.html?series=SERIESID`. You can also pass
```
lkiesow, line 96: Please use `-`

modules/lti/src/main/resources/tools/series/series.js
```diff
@@ -60,49 +61,67 @@ function i18n(key) {
...
       var episode = results[i],
           i18ncreator = Mustache.render(i18n('CREATOR'), {creator: episode.dcCreator}),
-          template = $('#template-episode').html(),
+          template = isInstructor ? $('#template-episode-admin').html() : $('#template-episode-user').html(),
           tpldata = {
             player: player + episode.id,
+            uid: episode.id,
             title: episode.dcTitle,
             i18ncreator: i18ncreator,
+            hasDeletion: getParam('deletion') == 'true',
```
lkiesow, line 124: This parameter isn't documented

lkiesow (changes requested):
I didn't manage to get through today and I think there are a few more problems in here.
Will continue the review over the next days.
How urgent is it for you to get this into Opencast 8?

pmiddend:
@lkiesow Thanks for the thorough review so far. We've just discussed this plapadoo-internally and we don't think it's urgent to get this into 8. Please continue to review this PR, we will fix your findings and update it (and think of something more appropriate for the `upload()` function as you suggested). However, as we announced last week, we'll rewrite the CSS, HTML and Javascript portions of the PR to use React (and Bootstrap), so expect to see major changes instead of small-scale bugfixes.

modules/lti/src/main/resources/tools/series/series.js
pmiddend: Not sure _where_ you want it to be documented. It's described in `ltimodule.md`. Should I add a source code comment?

pmiddend:
This is now superseded by #1289

PR #1019, opened 2019-08-06
MH-12618, SMIL tests fail when doctype url can't be resolved

JamesUoM:
remove the doctype with the dtd url and it's flakey http protocol in the test smil file. live smil files have no doctype anyway


gregorydlogan:
I'm confident that the reports are failing for spurious reasons, but I've restarted the build anyway.  IMO should just merge as is.

PR #1022, opened 2019-08-07
MH-13681 Automatic publication of streaming URLs

lwillmann:
First version without the implementation of the LiveScheduleService

### DIFF
--- a/docs/guides/admin/docs/modules/liveschedule.md
+++ b/docs/guides/admin/docs/modules/liveschedule.md
@@ -74,6 +74,10 @@ live.mimeType=video/x-flv
 
 # The distribution service to use: download or aws.s3
 live.distributionService=download
+
+# A list of combinations with target flavor and resolution for which streaming URIs should be published.
+# Default is an empty list
+# live.publishStreaming=presenter/delivery:1920x540
 ```
 
 ### Step 2: Configure the capture agent
--- a/docs/guides/admin/docs/modules/liveschedule.md
+++ b/docs/guides/admin/docs/modules/liveschedule.md
@@ -76,8 +76,9 @@ live.mimeType=video/x-flv
 live.distributionService=download
 
 # A list of combinations with target flavor and resolution for which streaming URIs should be published.
-# Default is an empty list
-# live.publishStreaming=presenter/delivery:1920x540
+# For example: live.publishStreaming=presenter/delivery:1920x540
+# Default is not to publish streaming URIs
+# live.publishStreaming=
 ```
 
 ### Step 2: Configure the capture agent
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -116,6 +116,8 @@ org.opencastproject.download.url=${org.opencastproject.server.url}/static
 # The directory to store media, metadata, and attachments for download from the engage tool
 org.opencastproject.download.directory=${org.opencastproject.storage.dir}/downloads
 
+# The mime types of streaming formats for which the raw streaming URLs should be published and retracted automatically
+# org.opencastproject.publish.streaming.formats=application/x-mpegURL
 
 ######### DATABASE #########
 
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -116,8 +116,9 @@ org.opencastproject.download.url=${org.opencastproject.server.url}/static
 # The directory to store media, metadata, and attachments for download from the engage tool
 org.opencastproject.download.directory=${org.opencastproject.storage.dir}/downloads
 
-# The mime types of streaming formats for which the raw streaming URLs should be published and retracted automatically
-# org.opencastproject.publish.streaming.formats=application/x-mpegURL
+# The mime types of streaming formats for which raw streaming URLs will be published automatically under the channel:
+# 'engage-player'. Use the character: '*' if all configured streaming URLs should be published
+# org.opencastproject.publish.streaming.formats=*
 
 ######### DATABASE #########
 
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -117,8 +117,9 @@ org.opencastproject.download.url=${org.opencastproject.server.url}/static
 org.opencastproject.download.directory=${org.opencastproject.storage.dir}/downloads
 
 # The mime types of streaming formats for which raw streaming URLs will be published automatically under the channel:
-# 'engage-player'. Use the character: '*' if all configured streaming URLs should be published
-# org.opencastproject.publish.streaming.formats=*
+# 'engage-player'. Use the character: '*' if all configured streaming URLs should be published.
+# The publication of streaming URI's is switched off by default
+# org.opencastproject.publish.streaming.formats=
 
 ######### DATABASE #########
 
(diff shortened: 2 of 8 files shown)


gregorydlogan:
@lwillmann I've restarted the buildbot build.  If you notice things failing like this again ping me or Lars and we can restart the build.

lwillmann:
Thank you @gregorydlogan I will do so

lwillmann:
When publishing events with streaming (e.g., HLS), the raw streaming URLs should be added to  existing publication-elements.   

- The live scheduler automatically publishes at the beginning of an event and the streaming URLs are added to the "engage-live"-channel

- For VOD streaming, an option of the "publish-engage" operation should be added to generate streaming links under the "engage-player"-channel

Withdrawing published events already remove the publication-channels "engage-live" and "engage-play.

docs/guides/admin/docs/modules/liveschedule.md
```diff
@@ -74,6 +74,10 @@ live.mimeType=video/x-flv
 
 # The distribution service to use: download or aws.s3
 live.distributionService=download
+
+# A list of combinations with target flavor and resolution for which streaming URIs should be published.
+# Default is an empty list
+# live.publishStreaming=presenter/delivery:1920x540
```
gregorydlogan: Defaults should be listed, and match the commented out configuration

etc/custom.properties
```diff
@@ -116,6 +116,9 @@ org.opencastproject.download.url=${org.opencastproject.server.url}/static
 # The directory to store media, metadata, and attachments for download from the engage tool
 org.opencastproject.download.directory=${org.opencastproject.storage.dir}/downloads
 
+# The mime types of streaming formats for which raw streaming URLs will be published automatically under the channel:
+# 'engage-player'. Use the character: '*' if all configured streaming URLs should be published
+# org.opencastproject.publish.streaming.formats=*
```
gregorydlogan: Defaults should be listed, and match the commented out configuration

etc/listproviders/publication.channels.properties
```diff
@@ -41,4 +41,4 @@ api={"label":"EVENTS.EVENTS.DETAILS.PUBLICATIONS.EXTERNAL_API", "order":1}
 oaipmh-default={"label":"Default OAI-PMH Repository", "order":2}
 engage-player={"icon":"img/engage_2x.png", "order":3}
 youtube={"icon":"img/youtube_2x.png", "order":4}
-engage-player={"icon":"img/engage_2x.png", "order":5}
+engage-player={"icon":"img/engage_2x.png", "order":5}
```
gregorydlogan: No change -> Remove this from the history.

etc/org.opencastproject.liveschedule.impl.LiveScheduleServiceImpl.cfg
```diff
@@ -32,3 +32,7 @@ live.mimeType=video/x-flv
 
 # The distribution service to use: download or aws.s3
 live.distributionService=download
+
+# A list of combinations with target flavor and resolution for which streaming URIs should be published.
+# Default is an empty list
+# live.publishStreaming=presenter/delivery:1920x540
```
gregorydlogan: Defaults should be listed, and match the commented out configuration

gregorydlogan (changes requested):
This looks ok to me, aside from the config stuff.  Can someone with a live streaming capable setup test this?  @karendolan, @rute-santos perhaps?

lwillmann:
Hi Greg, 
thanks for the review. 
I hope my last commit has resolved all requested changes. Normally the publication of Streaming-URLs is switched off by not having a configuration at all.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -444,7 +452,8 @@ void setDuration(MediaPackage mp, DublinCoreCatalog dc) {
             mp.getDuration());
   }
 
-  void addLiveTracks(MediaPackage mp, String caName) throws LiveScheduleException {
+  Map<String, Track> addLiveTracks(MediaPackage mp, String caName) throws LiveScheduleException {
+    HashMap<String, Track> generatedTracks = new HashMap<String, Track>();
```
rute-santos, line 480: Shouldn't the live tracks be updated in the publication element in updateLiveEvent() too? The publication element was not updated there before because it did not change, but now that the live tracks are being added, I believe they can change and thus need to be updated.

gregorydlogan (approved):

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
lwillmann: Thanks for the review Rute. I've changed updateLiveEvent() to add newly generated Tracks. I'm not sure in which case the updateLiveEvent comes to the point that it adds new Tracks to the mediapackage ... can I be sure that the publication element exists? Do I have to remove existing tracks?

rute-santos: Hi @lwillmann,

I don't think new live tracks will be added when an update occurs. In my comment above, I meant tracks were now "added" to the publication element. The publication element did not have tracks before. Sorry for the confusion.

When a live update occurs, the tracks will be basically the same (same url, but different element ids). I would remove and add the elements to the publication element in an update just to make sure that everything is consistent.

The only case I can think of where the publication element would not be there is if there was an exception when it was first created...

Thanks!

lwillmann:
Hi @rute-santos ,
thanks for your advice. I hope the method no behaves correctly.
Kind regards,
Ludwig

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -355,6 +365,27 @@ boolean updateLiveEvent(MediaPackage previousMp, DublinCoreCatalog episodeDC) th
     return true;
   }
 
+  private void createOrUpdatePublicationTracks(Publication publication, Map<String, Track> generatedTracks) {
+    for (String publishedStreamingFormat : publishedStreamingFormats) {
+      Track track = generatedTracks.get(publishedStreamingFormat);
+      if (track != null) {
+        if (publication.getTracks().length > 0) {
+          publication.clearTracks();
+        }
```
rute-santos: It's possible to have more than one publishedStreamingFormat, correct? Then, the clearTracks() may delete a track added in the previous iteration of the loop... It would be best to delete the tracks before the loop.

lwillmann: Absolutely %-) Stupid me ... thanks again

rute-santos: Looks good now! Thanks!

lkiesow:
Since the Opencast 8 feature freeze is fast approaching, @rute-santos, do you mind reviewing and merging this once it is okay?

PR #1024, opened 2019-08-07
MH-13684, Do not include auth token in republished URLs

lkiesow:
Using Opencast's token based authorization system, users may require a
valid token for accessing resources like videos (this is configurable,
of course, but that would be a common use case). Access without a token
will always be denied in such a case, regardless of the users login
status.

To obtain the token, the search service will include then in search
results for users with valid access as part of the returned URLs. This
means that there is no need for any special handling of search results
if tokens are required in contrast to them not being required.

Unfortunately, this has the side-effect that for re-publication of
metadata, tokens will end up in the published data since they look just
like regular URLs, causing the URLs to become permanently invalid.

This patch ensures that signatures will only be attached to requests via
the REST interfaces, but not to requests made internally to the search
service, ensuring that no tokens get mangled in published URLs.

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
@@ -159,7 +159,7 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
       @PathParam("format")    String  format
       ) throws SearchException, UnauthorizedException {
 
-    SearchQuery query = new SearchQuery();
+    SearchQuery query = new SearchQuery().signURLs(true);
 
     // If id is specified, do a search based on id
     if (StringUtils.isNotBlank(id))
@@ -252,9 +252,13 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
     }
 
     SearchQuery search = new SearchQuery();
-    search.withId(id).withSeriesId(seriesId)
-            .withElementFlavors(flavorSet.toArray(new MediaPackageElementFlavor[flavorSet.size()]))
-            .withElementTags(tags).withLimit(limit).withOffset(offset);
+    search.withId(id)
+        .withSeriesId(seriesId)
+        .withElementFlavors(flavorSet.toArray(new MediaPackageElementFlavor[0]))
+        .withElementTags(tags)
+        .withLimit(limit)
+        .withOffset(offset)
+        .signURLs(true);
 
     if (StringUtils.isNotBlank(text)) {
       search.withText(text);
@@ -314,7 +318,7 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
   public Response getByLuceneQuery(@QueryParam("q") String q, @QueryParam("sort") String sort, @QueryParam("limit") int limit,
           @QueryParam("offset") int offset, @QueryParam("admin") boolean admin, @PathParam("format") String format)
           throws SearchException, UnauthorizedException {
-    SearchQuery query = new SearchQuery();
+    SearchQuery query = new SearchQuery().signURLs(true);
     if (!StringUtils.isBlank(q))
       query.withQuery(q);
 
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
@@ -37,12 +37,14 @@ import org.opencastproject.util.doc.rest.RestQuery;
 import org.opencastproject.util.doc.rest.RestResponse;
 import org.opencastproject.util.doc.rest.RestService;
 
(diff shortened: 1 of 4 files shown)


mtneug:
Do you think redirecting to 6.x makes sense?

lkiesow:
No, I don't think it does. It's a known bug for many years and it hasn't bothered anyone enough to fix this until now. It can't suddenly be that urgent to port it back to old versions.

lkiesow:
…and the fact that for a week now no one has picked up the review of this patch further shows that people do not think that this is important ;-P

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -147,7 +149,10 @@ public Response remove(@PathParam("id") String mediaPackageId) throws SearchExce
...
           @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN) }, reponses = { @RestResponse(description = "The request was processed succesfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json.")
+          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN),
+          @RestParameter(defaultValue = "true", description = "If results are to be signed", isRequired = false,
+              name = "sign", type = RestParameter.Type.BOOLEAN)
+    }, reponses = { @RestResponse(description = "The request was processed succesfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json.")
```
Rillke, line 155: `successfully`

Rillke: `The search results, formatted as XML or JSON.`

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -233,11 +240,14 @@ public Response getEpisodeAndSeriesById(
...
           @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN) }, reponses = { @RestResponse(description = "The request was processed succesfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json.")
+          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN),
+          @RestParameter(defaultValue = "true", description = "If results are to be signed", isRequired = false,
+              name = "sign", type = RestParameter.Type.BOOLEAN)
+  }, reponses = { @RestResponse(description = "The request was processed successfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json.")
```
Rillke, line 246: `The search results, formatted as XML or JSON.`

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -310,11 +326,16 @@ public Response getEpisode(@QueryParam("id") String id, @QueryParam("q") String
...
           @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN) }, reponses = { @RestResponse(description = "The request was processed succesfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json")
+          @RestParameter(defaultValue = "false", description = "Whether this is an administrative query", isRequired = false, name = "admin", type = RestParameter.Type.BOOLEAN),
+          @RestParameter(defaultValue = "true", description = "If results are to be signed", isRequired = false,
+              name = "sign", type = RestParameter.Type.BOOLEAN)
+    }, reponses = { @RestResponse(description = "The request was processed succesfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, expressed as xml or json")
```
Rillke, line 332: `successfully`

Rillke: `The search results, formatted as XML or JSON.`

PR #1025, opened 2019-08-08
MH-13685: Added dublincore, security to publish (merge) and replublish-oaimph

TurRil:
Added dublincore and security to publish (merge) and replublish-oaimph, and small ui updates to use mediapackage.start or dcCreated.

This fixes the bug where dcCreated is not filled in correctly and the ACL's are not merged in for events where a caption gets added from an attachment workflow.

### DIFF
--- a/etc/workflows/attach-watson-transcripts.xml
+++ b/etc/workflows/attach-watson-transcripts.xml
@@ -33,9 +33,21 @@
       exception-handler-workflow="partial-error"
       description="Distribute and publish to engage server">
       <configurations>
+        <configuration key="download-source-flavors">dublincore/*,security/*</configuration>
         <configuration key="download-source-tags">engage-download</configuration>
         <configuration key="strategy">merge</configuration>
         <configuration key="check-availability">true</configuration>
+      </configurations>
+    </operation>
+
+    <operation
+      id="republish-oaipmh"
+      exception-handler-workflow="partial-error"
+      description="Update recording metadata in default OAI-PMH repository">
+      <configurations>
+        <configuration key="source-flavors">dublincore/*,security/*</configuration>
+        <configuration key="download-source-tags">engage-download</configuration>
+        <configuration key="repository">default</configuration>
       </configurations>
     </operation>
 
--- a/etc/workflows/publish-uploaded-assets.xml
+++ b/etc/workflows/publish-uploaded-assets.xml
@@ -49,7 +49,8 @@
       exception-handler-workflow="partial-error"
       description="Update recording in Opencast Media Module">
       <configurations>
-        <configuration key="download-source-flavors">${download-source-flavors}</configuration>
+        <configuration key="download-source-flavors">${download-source-flavors},dublincore/*,security/*</configuration>
+        <configuration key="download-source-tags">engage-download</configuration>
         <configuration key="strategy">merge</configuration>
         <configuration key="check-availability">false</configuration>
       </configurations>
@@ -63,7 +64,8 @@
       exception-handler-workflow="partial-error"
       description="Update recording metadata in default OAI-PMH repository">
       <configurations>
-        <configuration key="source-flavors">${download-source-flavors}</configuration>
+       <configuration key="source-flavors">${download-source-flavors},dublincore/*,security/*</configuration>
+        <configuration key="download-source-tags">engage-download</configuration>
         <configuration key="repository">default</configuration>
       </configurations>
     </operation>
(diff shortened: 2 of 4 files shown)


gregorydlogan:
As annoying as it would be, this should be two issues.  That being said, this is filed.  Next time.

Regarding the ACL changes, these seem fine to me.  I don't see anything obvious that prevents a merge.
Regarding the dcCreated field, this is more problematic.  We discussed a bunch of dc and (specifically) datetime related things this morning.  @wsmirnow had things to say about this.  Would this cause issues with, or fix things you're looking at.  cc @karendolan

etc/workflows/attach-watson-transcripts.xml
```diff
@@ -33,12 +33,24 @@
       exception-handler-workflow="partial-error"
       description="Distribute and publish to engage server">
       <configurations>
+        <configuration key="download-source-flavors">dublincore/*,security/*</configuration>
```
wsmirnow, line 36: Why are you updating dublincore and security catalogs here? The `attach-watson-transcription` operation should not touch any metadata.

etc/workflows/attach-watson-transcripts.xml
```diff
@@ -33,12 +33,24 @@
...
         <configuration key="check-availability">true</configuration>
       </configurations>
     </operation>
 
+    <operation
+      id="republish-oaipmh"
+      exception-handler-workflow="partial-error"
+      description="Update recording metadata in default OAI-PMH repository">
+      <configurations>
+        <configuration key="source-flavors">dublincore/*,security/*</configuration>
```
wsmirnow, line 48: Same here… Removing this line should leave the metadata in the oai-pmh repository untouched but add/update all `download-source-tags` marked assets.

etc/workflows/publish-uploaded-assets.xml
```diff
@@ -49,7 +49,8 @@
       exception-handler-workflow="partial-error"
       description="Update recording in Opencast Media Module">
       <configurations>
-        <configuration key="download-source-flavors">${download-source-flavors}</configuration>
+        <configuration key="download-source-flavors">${download-source-flavors},dublincore/*,security/*</configuration>
```
wsmirnow, line 52: Updating already published metadata on asset upload isn't intuitive. Here you only want to add new asset to the publication.

etc/workflows/publish-uploaded-assets.xml
```diff
@@ -63,7 +64,8 @@
       exception-handler-workflow="partial-error"
       description="Update recording metadata in default OAI-PMH repository">
       <configurations>
-        <configuration key="source-flavors">${download-source-flavors}</configuration>
+       <configuration key="source-flavors">${download-source-flavors},dublincore/*,security/*</configuration>
```
wsmirnow, line 67: See above…

wsmirnow (changes requested):

wsmirnow:
> As annoying as it would be, this should be two issues. That being said, this is filed. Next time.
> 
> Regarding the ACL changes, these seem fine to me. I don't see anything obvious that prevents a merge.
> Regarding the dcCreated field, this is more problematic. We discussed a bunch of dc and (specifically) datetime related things this morning. @wsmirnow had things to say about this. Would this cause issues with, or fix things you're looking at. cc @karendolan

I've only touched the Admin UI. My PR #1079 does not conflict with this PR.

karendolan:
I notice that dcTitle is also not guaranteed to exist and has a fallback value, per the JavaScript. I wonder why the assumption that mediapackage "start" will exist if the dcCreated does not. How are @TurRil 's mediapackages getting compiled and published without some core metadata?    
> Regarding the dcCreated field, this is more problematic. We discussed a bunch of dc and (specifically) datetime related things this morning.

wsmirnow:
@TurRil, I would splitt this PR into two seperate PRs. The second part (theodul model start date fallback) is fine for me. But…

Why do you want to update dc and security catalogs on transcriptions add/update? Does the transcription operation touch any data of the dublincore catalog? The ACLs shouldn't be touched anyway.

KatrinIhler:
@TurRil Conflicts!

karendolan:
@TurRil I don't understand why this pull adds the dublincore and ACLS to be publish during an upload asset and attach caption workflow. Are you seeing those files being removed during your attach caption or upload asset workflows? Does your site have a process that modifies those files when captions or other attachments are added to a mediapackage?

Our site uses the upload asset and attach caption workflows every day in our modified Opencast v5/6/7/8x and have had no problems with existing published dublincore or ACLs.

gregorydlogan:
@TurRil any progress here?

KatrinIhler:
@TurRil What's the status here? If there can't be any work done on this in the foreseeable future, it might be better to close this PR and reopen at a later date.

KatrinIhler:
As per our recent discussion in the dev meeting and the proposal that followed, I'm going to close this PR since this has seen no updates within the past 1,5 years. This does not constitute a rejection, feel free to re-open once you have time to address the review findings. :)

PR #1026, opened 2019-08-13
NOJIRA: Adding Ansible script documentation

gregorydlogan:
This PR adds documentation for the ansible scripts.  Normally I would just push this in, but I wanted a second set of eyes to make sure they make sense both in terms of actions, and in terms of grammar.

### DIFF
--- a/docs/guides/admin/docs/installation/ansible.md
+++ b/docs/guides/admin/docs/installation/ansible.md
@@ -1,0 +1,45 @@
+Install from Repository via Ansible (Debian, and RedHat based distros)
+===========================================================================
+
+We provide [Ansible](https://www.ansible.com/) installation and configuration scripts which drastically reduce the time
+required to set up an Opencast installation, and help you manage the configuration across your entire cluster.  Whether
+you are installing a single machine, or dozens, these scripts should do the basic setup of your Opencast install.  These
+scripts use the package repository, so the first step is to register.
+
+Note: Examples in this document refer to `7.x` and `r/7.x`.  You should replace those with the version you want to install!
+
+Registration
+------------
+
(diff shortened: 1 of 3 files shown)


docs/guides/admin/docs/installation/ansible.md
```diff
@@ -0,0 +1,45 @@
...
+
+Ansible may be available from your distribution's packaging manager, or you may need to install it manually.  We
+recommend following the [Ansible installation documentation](https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html)
+to get the latest version.
+
+
+Script Setup
+------------
+
+The next step is getting the scripts themselves.  Go to the [list of branches](https://github.com/opencast/oc-config-management/branches)
```
lkiesow, line 36: I'm not entirely happy with this section but it should do and I don't really mind keeping it.

To use an installation method like this, you should really know what you are doing. More than when you do it manually. No one should just run these scripts. This means that we can expect people to have a certain technical skill level and it would just suffice to point them to the GitHub repo without explaining how to click a download button ;-)

lkiesow (approved):
As discussed, I added a patch to your docs. If you are fine with it, feel free to merge this.

gregorydlogan:
I get what you're saying. The main goal here is to make setting up a *cluster* easier. Anyone ought to be able to stand up a demo AIO, but I regularly screw up setting up a cluster manually. I like the patch, will merge next review round.

docs/guides/admin/docs/installation/ansible.md
gregorydlogan: I wasn't really sure how to do this part.  We need a better technical writer, but this will do for now.

PR #1030, opened 2019-08-21
MH-13691, Configure max open files

lkiesow:
This patch configures the maximum allowed number of open files for the
Opencast process since the default can be a bit low, especially on
larger workers where we have seen “too many open files” errors show up
multiple times.

Since Opencast usually runs on dedicated machines, this setting should
not harm anyone, but only prevent errors in these edge cases.

### DIFF
--- a/docs/scripts/service/opencast.service
+++ b/docs/scripts/service/opencast.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Opencast
+Documentation=https://docs.opencast.org
+
 After=local-fs.target
 After=network.target
 After=remote-fs.target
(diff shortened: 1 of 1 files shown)


mtneug:
> Since Opencast usually runs on dedicated machines…

What errors can occur if this is not the case?

lkiesow:
What we have seen happening multiple times now is a Java error saying something along the lines of “Cannot open file”, sometimes also when trying to launch a sub-process (e.g. FFmpeg) which results in either the node staying up but no further processing or in the node going randomly up and down. In both cases, Opencast did not recover on its own and in the second case it caused problems for the whole cluster since Opencast was trying to start jobs on those nodes only to notice that they went down but since the came immediately up again… well… `goto dispatching;`

lkiesow:
I mean, in theory, the idea for these kinds of limits is that a single process can not use up all your resources and by doing so block other important processes. If you run a mail server and a web server on a single machine you don't want your web server to take down your mail server under heavy load…

This type of problem shouldn't usually be a problem in the case of Opencast deployments.

PR #1031, opened 2019-08-26
MH-13692 Load all roles in Admin UI at once

KatrinIhler:
Load all roles at once in the admin UI instead of loading them in increments of 100 when the user is scrolling through the dropdown menu since this doesn't provide a real benefit and makes the search function unreliable: A role that isn't loaded yet cannot be searched for, so for search results to be accurate, one first has to scroll through all the roles...which kind of defeats its purpose.

_This work was sponsored by the ETH Zürich._

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js
@@ -25,21 +25,16 @@ angular.module('adminNg.controllers')
 .controller('EventCtrl', [
   '$scope', 'Notifications', 'EventTransactionResource', 'EventMetadataResource', 'EventAssetsResource',
   'EventAssetCatalogsResource', 'CommentResource', 'EventWorkflowsResource', 'EventWorkflowActionResource',
-  'EventWorkflowDetailsResource', 'ResourcesListResource', 'UserRolesResource', 'EventAccessResource',
+  'EventWorkflowDetailsResource', 'ResourcesListResource', 'EventAccessResource',
   'EventPublicationsResource', 'EventSchedulingResource','NewEventProcessingResource', 'CaptureAgentsResource',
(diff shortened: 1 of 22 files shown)


gregorydlogan:
So, the reason we don't load them all at the beginning is that for some role provider and user combinations (I'm talking @smarquard's installation here) there are many many roles.  It would not be sane to query them all at once, either in terms of cache memory, or in terms of user-side perf.  It *should* be automatically searching as the number of roles dwindles, so the closer you get to the right role the more likely it is to be in the role list.  If it's not doing that then there's some kind of regression IMO.  Loading all of the Opencast local roles might be ok, but loading all roles from all providers likely is not.

lkiesow:
@gregorydlogan, same concerns here but I will test this in the next days in an installation where we get all roles from external providers… and we have a few thousands of them. But for @smarquard this isn't a problem since they [never actually implemented the full roleprovider interface](https://github.com/opencast/opencast/blob/bed9fc591b334dfc637148bb201a8f516735294b/modules/userdirectory-sakai/src/main/java/org/opencastproject/userdirectory/sakai/SakaiUserProviderInstance.java#L603-L608). I'll let you know how tests go. Though, @KatrinIhler already tested this on a system with 20 000 JPA roles without problem.

lkiesow:
@gregorydlogan, btb. multiple other user and role provider followed the Sakai example and only half-implemented the interfaces instead of fixing the problem or the interfaces. That's why getting a reliable list of roles or users is actually broken on a lot of systems :-(

smarquard:
The design was presented some time ago, before it was even in develop:

https://www.slideshare.net/smarquard/opencast-valencia-2017-users-groups-roles-acls-and-providers 

(slide 12). There are some RoleProviders for which getRoles() makes sense, and others for which it does not (specifically where a potentially large number of roles are sourced from an authoritative external source, for which the performance is unknown).

PR #1038, opened 2019-08-29
MH-13699, Fix Workflow Index Rebuild ACL Handling

lkiesow:
The workflow index rebuild process tried to load and set the ACLs set
during the workflow execution by loading those as files from the
working file directory. This will nearly always fail for terminated
workflows since the ACLs do not exist any longer.

While this has no real consequence since the ACLs will be overwritten
later in the re-indexing process by ACLs from the asset manager anyway,
it still produces a huge amount of NullPointerException stack-traces
(one for each workflow ever run by the system), making it impossible
to spot and identify real errors in the process.

This patch prevents the process from trying to access the non-existing
files in the first place, not only preventing the errors, but also
speeding up the rebuilding process.

*This patch is based on pull request #1036*

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -2342,51 +2342,60 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
 
   @Override
   public void repopulate(final String indexName) throws ServiceRegistryException {
-    List<String> workflows =  serviceRegistry.getJobPayloads(Operation.START_WORKFLOW.toString());
+    final String startWorkflow = Operation.START_WORKFLOW.toString();
+    final int total = serviceRegistry.getJobCount(startWorkflow);
+    final int limit = 1000;
 
     final String destinationId = WorkflowItem.WORKFLOW_QUEUE_PREFIX + indexName.substring(0, 1).toUpperCase()
             + indexName.substring(1);
-    if (workflows.size() > 0) {
-      final int total = workflows.size();
+    if (total > 0) {
       logger.info("Populating index '{}' with {} workflows", indexName, total);
       final int responseInterval = (total < 100) ? 1 : (total / 100);
       int current = 0;
-      for (final String workflow : workflows) {
-        current += 1;
-        if (StringUtils.isEmpty(workflow)) {
-          logger.warn("Skipping restoring of workflow no {}: Payload is empty", current);
-          continue;
-        }
-        WorkflowInstance instance;
-        try {
-          instance = WorkflowParser.parseWorkflowInstance(workflow);
-        } catch (WorkflowParsingException e) {
-          logger.warn("Skipping restoring of workflow. Error parsing: {}", workflow, e);
-          continue;
-        }
-        Organization organization = null;
-        try {
-          organization = organizationDirectoryService.getOrganization(instance.getOrganizationId());
-        } catch (NotFoundException e) {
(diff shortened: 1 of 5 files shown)


modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2342,51 +2342,69 @@ public synchronized void cleanupWorkflowInstances(int buffer, WorkflowState stat
...
+      List<String> workflows;
+      do {
+        workflows = serviceRegistry.getJobPayloads(startWorkflow, limit, offset);
+        logger.debug("Got {} workflows for re-indexing", workflows.size());
+        offset += limit;
+
+        for (final String workflow : workflows) {
+          current += 1;
+          if (StringUtils.isEmpty(workflow)) {
+            logger.warn("Skipping restoring of workflow no {}: Payload is empty", current);
```
Rillke, line 2365: `Skip restore of workflow #{}: Payload is empty`? Looks clearer than `no`.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2342,51 +2342,69 @@ public synchronized void cleanupWorkflowInstances(int buffer, WorkflowState stat
...
+            continue;
+          }
+          WorkflowInstance instance;
+          try {
+            instance = WorkflowParser.parseWorkflowInstance(workflow);
+          } catch (WorkflowParsingException e) {
+            logger.warn("Skipping restoring of workflow. Error parsing: {}", workflow, e);
```
Rillke, line 2372: Parameter missing? `Skip restore of workflow {}. Error parsing: {}`

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2342,51 +2342,69 @@ public synchronized void cleanupWorkflowInstances(int buffer, WorkflowState stat
...
-                  messageSender.sendObjectMessage(destinationId, MessageSender.DestinationType.Queue,
-                          WorkflowItem.updateInstance(instance, dcXml, accessControlList));
-                });
-        if ((current % responseInterval == 0) || (current == total)) {
-          logger.info("Updating {} workflow index {}/{}: {} percent complete.", indexName, current, total,
-                  current * 100 / total);
+          // get metadata for index update
+          final String dcXml = getEpisodeDublinCoreXml(instance.getMediaPackage());
+
+          // get acl for active workflows.
```
Rillke, line 2386: `Get ACL for active workflows.`

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2342,51 +2342,69 @@ public synchronized void cleanupWorkflowInstances(int buffer, WorkflowState stat
...
+          // don't try this for terminated workflows since the ACLs are no longer in the working file repository and
```
Rillke, line 2387: `Don't try ...`

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
lkiesow: The logger will handle exceptions gracefully if passed as the last parameter. Using replacement parameters will cause the stack-trace being cut. That's usually a bad idea.

PR #1040, opened 2019-08-30
MH-13701 Interpret source-audio-name correctly for composite operation

KatrinIhler:
If source-audio-name is not set on a composite operation, it's set to null, however this becomes an empty string when using the remote service, resulting in a composite video with no audio. To avoid this, consider the empty string as a possible value on interpretation.

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -1693,8 +1693,8 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
       // handle audio
       boolean lowerAudio = lowerLaidOutElement.getElement().hasAudio();
       boolean upperAudio = upperLaidOutElement.get().getElement().hasAudio();
-      // if not specfied or "both", use both videos
-      if (audioSourceName != null && ! ComposerService.BOTH.equalsIgnoreCase(audioSourceName)) {
+      // if audio source name is "both" or unspecified, use audio of both videos, otherwise pick one
+      if (StringUtils.isNotBlank(audioSourceName) && ! ComposerService.BOTH.equalsIgnoreCase(audioSourceName)) {
         lowerAudio = lowerAudio & ComposerService.LOWER.equalsIgnoreCase(audioSourceName);
         upperAudio = upperAudio & ComposerService.UPPER.equalsIgnoreCase(audioSourceName);
       }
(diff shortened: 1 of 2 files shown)


Rillke (commented):
1. (De)-serialization of NULL should either work or be somehow documented not to work? This is a work-around for just one case?
2. Can we have unit tests to catch the serialization issue in general, or this specific one? (If this is beyond this patches quota, can you send a quote/estimate your unihalle contact?)
3.  Since the documentation of the CompositeWorkflowOperationHandler documents that "both" is the default, does it make sense to you if the CompositeWorkflowOperationHandler already applies these defaults (as it does for the background color, for example) (i.e. not deferring it to a nested service.)

gregorydlogan:
@Rillke 
1. This patch is handling *invalid* paramters.  If you're feeding `composite()` a null or blank value here you should either get the defaults, or an exception.
2. Not my pervue, but I would say this is technically possible.
3. The operation handler already enforces one of the three choices.  This does not prevent the method from being called directly (ie, hit the REST endpoint).

@KatrinIhler I'm not hugely familiar with the code here, but the comment above the line you've changed doesn't make a ton of sense.  `If it's not specified, or both` doesn't match the logic - it should read `... or *not* both`.  Or the logic is wrong.  But I'm not sure which it is :)

KatrinIhler:
@gregorydlogan The comment actually does make sense, but I've rephrased it to be a bit more obvious.

KatrinIhler:
@Rillke 
1. Apparently it's Apache that translates a null for a form parameter into an empty string. In my opinion this should by handled depending on the service.
2. I'm not exactly sure what you want me to do here, could you elaborate?
3. Greg is correct that this doesn't fix the case where the ComposerService is called directly with an empty string, but I've added the default to the WOH to comply with the documentation.

gregorydlogan:
@KatrinIhler comment is now clearer, but I remember this branch having more code.  Did you force-push a single commit by accident, or was that code included in another PR somewhere?

KatrinIhler:
@gregorydlogan Idk what happened there, but another force-push fixed it...

Rillke:
@KatrinIhler Let's phrase it differently: I assume that most developers use the All-In-One distribution for testing, while we are using the distributed distribution. Therefore I assume that it is likely we get more of these issues in future. Consequently I wish there would be no difference at all between remote services and the all-in-one, or if there are, developers are well aware of these.

KatrinIhler:
@Rillke Since this behaviour is out of our control, I don't see how we could fix this in a general way. Transforming each empty string back into a null could also lead to errors. Imo it's a better idea to make the services more robust since an empty string can also get in in other ways.

PR #1044, opened 2019-09-04
MH-13705: clean node, node_modules and bower_components folders 

miesgre:
When you do a `mvn clean`, some folders are not removed (node, node_modules, bower_components)

This PR removes node, node_modules and bower_components folders when running `mvn clean`

### DIFF
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -537,6 +537,34 @@
           </instructions>
         </configuration>
       </plugin>
+      <plugin>
+        <artifactId>maven-clean-plugin</artifactId>
+        <configuration>
+          <filesets>
+            <fileset>
+              <directory>node</directory>
+              <includes>
+                <include>**</include>
+              </includes>
+              <followSymlinks>false</followSymlinks>
+            </fileset>
+            <fileset>
+              <directory>node_modules</directory>
+              <includes>
+                <include>**</include>
+              </includes>
+              <followSymlinks>false</followSymlinks>
+            </fileset>
+            <fileset>
+              <directory>bower_components</directory>
+              <includes>
+                <include>**</include>
+              </includes>
+              <followSymlinks>false</followSymlinks>
+            </fileset>
+          </filesets>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/modules/engage-paella-player/pom.xml
+++ b/modules/engage-paella-player/pom.xml
@@ -201,6 +201,27 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <artifactId>maven-clean-plugin</artifactId>
+        <configuration>
+          <filesets>
+            <fileset>
+              <directory>node</directory>
+              <includes>
+                <include>**</include>
+              </includes>
+              <followSymlinks>false</followSymlinks>
+            </fileset>
+            <fileset>
+              <directory>node_modules</directory>
+              <includes>
+                <include>**</include>
+              </includes>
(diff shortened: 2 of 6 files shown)


JulianKniephoff (approved):
I tried various combinations of `mvn clean`, `mvn clean install` and just `mvn install`. Everything still seems to build and work fine and indeed, no traces of any `node`, `node_modules` or `bower_components` are left after `mvn clean`.

lkiesow:
`-1` to this unless there is a good explanation for why you want to do this (no explanation is given here at this pull request). This should significantly slow down build times for no apparent reason. `mvn clean` is not supposed to remove repository caches. E.g. it will also not clear the local Maven repository. Why should it remove the npm or Bower caches?

gregorydlogan:
I agree with @lkiesow.  The usecase you have here is addressed with `git clean -fdx`, and this change *drastically* slows down builds - or makes them impossible if you are relying on said cache to be present (ie, no network connectivity).

JulianKniephoff:
I can't say anything to the original use case of course, but to explain why I merged this so eagerly: I think a `mvn clean install` should be as clean an install as possible, semantically.

More practically, though, I have seen several weird glitches now from keeping these folders around between changing branches, ranging from the build failing to the admin UI not loading correctly.

I also don't think that `git clean` is the solution here; maybe I want to keep some untracked files. I think the build system should at least offer **some** command to remove **all** build artifacts. And in Maven-land, `clean` **is** that command, IMO, but maybe we could hide removing these files behind a profile? (Or hide keeping them, which would be my preference.)

miesgre:
The reason for this PR is the same as @JulianKniephoff explained. Some times a `mvn clean install` builds correctly, but then there are problems with the admin UI (some `js` libraries are not found). I found it annoying

I could remove the folders with a simple rm, but I also think a mvn clean install should be as clean an install as possible.

gregorydlogan:
I don't think I've ever experienced said issues @miesgre @JulianKniephoff.  Is that occurring when you're actively working on the admin-ui module, or randomly?  If this is happening regularly I would have expected to see more people complaining about this, which suggests development.  IMO, if you're working on one of the JS related modules, you should expect to have to clear things manually.  Throwing out the caches every time is throwing the baby out with the bathwater - it's excessive, and causes issues for others when building.

I agree that there should be a way to clear the caches, but I would *strongly* disagree that these artefacts are build artefacts.  They are dependencies (ie, static, unchanging build-over-build), not build artefacts (ie, changing depending on the code in the module).  IMO, if this is happening a lot then what we really have is an issue with Node, which has two solutions: Remove Node, or update Node to a non-broken version.

JulianKniephoff:
I've been thinking about the issue a bit more and have to admit that I actually agree with these folders not being build artifacts. :thinking: That being said, I would still like to be able to say something like `mvn clean -PnukeNode` to easily get rid of the issues I've mentioned.

As to their nature: It's hard to say when these happen, since it's only very sporadically. Also, there have been different kinds of issues over time. So yes, I agree of course that we should rather fix the root cause for things like this, but that turns out to be hard sometimes. :sweat_smile:

Anyway, how about -- as a kind of peace offering -- I cook up a version of this PR that can be activated with a profile, as alluded to above? Would that be okay with everyone?

gregorydlogan:
A profile sounds good to me - I don't hate the idea of easily nuking the node cache(s), I hate the idea of doing it every time.  But I also don't tend to work in the javascript that often, so `git clean -fdx` tends to work when I run into suspected Node issues :D

JulianKniephoff:
May I direct all y'all's attention towards #1048, then? :wink:

PR #1045, opened 2019-09-04
MH-13705, Revert "clean node, […] when running mvn clean"

lkiesow:
This reverts commit a4660c57cb0b27a5e51f087f1dc184197a0b3237 (#1044)

There was no good explanation for this change. This should significantly
slow down build times for no apparent reason. mvn clean is not supposed
to remove repository caches. E.g. it will also not clear the local Maven
repository. Why should it remove the npm or Bower caches?

### DIFF
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -537,34 +537,6 @@
           </instructions>
         </configuration>
       </plugin>
-      <plugin>
-        <artifactId>maven-clean-plugin</artifactId>
-        <configuration>
-          <filesets>
-            <fileset>
-              <directory>node</directory>
-              <includes>
-                <include>**</include>
-              </includes>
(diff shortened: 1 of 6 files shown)


miesgre:
-1 to be reverted.

The reason for #1044 is the same as @JulianKniephoff explained here: https://github.com/opencast/opencast/pull/1044#issuecomment-528199460. 
Some times a `mvn clean install` builds correctly, but then there are problems with the admin UI (some `js` libraries are not found). I found it annoying

I could remove the folders with a simple rm, but I also think a mvn clean install should be as clean an install as possible.

gregorydlogan:
I'm going to comment on #1044 since that's where the majority of the discussion lives.  Others with opinions please join in there.

gregorydlogan:
Resolved in favour of #1048.

PR #1046, opened 2019-09-05
MH-13692 Load all roles in Admin UI

KatrinIhler:
Load all roles at once in the admin UI instead of loading them in increments of 100 when the user is scrolling through the dropdown menu since this doesn't provide a real benefit and makes the search function unreliable: A role that isn't loaded yet cannot be searched for, so for search results to be accurate, one first has to scroll through all the roles...which kind of defeats its purpose.

20000 JPARoles take about 1.6 seconds in a local allinone.

New version of #1004.

This work was sponsored by the ETH Zürich.


staubesv:
@KatrinIhler Hmm.... IIRC, loading the roles incrementally was implemented by @gregorydlogan on behalf of @smarquard because UCT ran into performance issues because loading all roles at once from an external user directory (LDAP) with something like 100K users simply does not scale... 
In case that issue has somehow been resolved meanwhile, be sure to check whether https://github.com/opencast/opencast/blob/develop/modules/admin-ui/src/main/webapp/scripts/lib/angular-chosen.js can be replaced by the official angular-chosen library (our version contains customisations for the incremental loading you are about to remove). 

[Update] Oh, just saw https://github.com/opencast/opencast/pull/1031 so this comment might be obsolete and also realised you reverted to standard angular-chosen... Note, btw, that the customisation of angular-chosen is the only reason why angular-chosen is not managed by bower... but that is another issue not related to this PR  ;-)

smarquard:
It wasn't because we "ran into a performance issue" trying to load 100K roles (although undoubtedly we would have if we tried). By design, roles may be provided for a user from an external system, and the external system should not be required to provide every single role (for performance and other reasons, and that external systems may vary widely in performance characteristics).

For reference: 
https://github.com/opencast/opencast/blob/develop/modules/common/src/main/java/org/opencastproject/security/api/RoleProvider.java

There are 3 cases that should be handled properly in the UI:

1. The RoleProvider provides a complete list of roles in advance (getRoles() returns a non-empty iterator, and the UI gets the complete list from the admin-ng endpoint)

2. The RoleProvider does not provide a complete list (getRoles() returns empty iterator), but findRoles() may return one or more roles for a partial pattern (e.g. for the Sakai UDP, entering a GUID like becb2aa2-7ccd-4f77-932b-341e77fec14b would return available roles becb2aa2-7ccd-4f77-932b-341e77fec14b_Learner and becb2aa2-7ccd-4f77-932b-341e77fec14b_Instructor).

3. The RoleProvider does not provide a complete list (getRoles() returns empty), but findRoles() will validate that a specific string is a valid role by returning an appropriate response from findRoles(). This is used by the https://github.com/opencast/opencast/blob/develop/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/CustomRoleProvider.java to validate role patterns like SITEID_Learner in conjunction with LTI where SITEID is some pattern expressed as a regexp that validates the format of site ids used in moodle or another LMS (or could be used for some other purpose).

Note that the behaviour is different for Role.Target.ACL vs Role.Target.USER: https://github.com/opencast/opencast/blob/develop/modules/common/src/main/java/org/opencastproject/security/api/Role.java#L45

The "unknown number of possible roles" is really a Role.Target.ACL concern, specific to ACLs for Events and Series. For Role.Target.USER, I'd expect the set of roles to be known in advance.

From a UI point of view, I think the progressive loading is not a good approach and should be scrapped. If a role string is entered into the role input box, then the UI needs to call the admin-ng endpoints and pass the search string to get back the set of matching roles from the providers.

If the UI only permits pre-loaded roles to be entered, then that breaks both role providers like the Sakai UDP and the built-in CustomRoleProvider.

It may be worth allowing Role Providers to indicate the minimum valid search length and let that bubble up so the UI doesn't have to issue a REST query to search for possible role matches from the very first character. For example, for the Sakai roles, we wouldn't search for anything less than 36 chars (length of GUID) but that could vary by site.

gregorydlogan:
@KatrinIhler I haven't had a chance to look through this, but does it handle the above three cases gracefully?

If so, @smarquard, considering the upcoming feature freeze, would you merge as is and then polish going forward?  Or should this be left for OC 9?

This seems reasonable to me as is, although I have not yet looked through the code.

lkiesow:
@gregorydlogan, I don't think this pull request affects the provided cases but cases two and three are a clear violation of our [RoleProvider interface](https://github.com/opencast/opencast/blob/develop/modules/common/src/main/java/org/opencastproject/security/api/RoleProvider.java#L33) and we really shouldn't put any hacks somewhere in our code for things that misuse the interface. If the interface is problematic, I have no problem with changing it. But right now, it is what it is.

smarquard:
Created PR #1065 to update the description of the RoleProvider interface, to align it with the usage established by the LMS providers and the CustomRoleProvider (which go back > 2 years).

smarquard:
I will test this PR in a build with the Custom and Sakai role providers enabled to see how cases #2 and #3 are handled.

smarquard:
Test environment: add / uncomment these in etc/custom.properties (I changed these slightly from defaults for testing purposes):

```
# Optional custom roles which can used in ACLs and granted to users or groups (comma-separated list of role names)
org.opencastproject.security.custom.roles=ROLE_CUSTOM_ONE, ROLE_CUSTOM_TWO, ROLE_CUSTOM_THREE

# Optional custom roles pattern (regular expression). Roles matching this pattern can be added to ACLs.
org.opencastproject.security.custom.roles.pattern=^ext:[0-9a-f-]+_(Learner|Instructor)$
```

Short version: this PR currently fixes things for editing roles for users and groups, and breaks things editing roles for ACLs (event and series).

Longer version: the use-case for roles that are sourced dynamically from an external system is only relevant (so far) to ACLs (access to events and series): there are no known use-cases where people want to do that for Role.Target.USER ACLs (access to Opencast capabilities).

So for the edit user / group dialogs, it's acceptable to call the endpoint

    /admin-ng/resources/ROLES.json?filter=role_target:USER&limit=-1

once when the edit dialog is loaded. The only changes that might happen to the role list here are when new groups are created, or the capture agent list changes. To verify that this PR fixes the behaviour compared to develop.opencast.org:

- Edit a user
- Select Roles (don't scroll the select box all the way down)
- Type in "ROLE_UI_U"
- ROLE_UI_USERS_VIEW should appear and be selectable (with some other roles)

However, for the ACL dialogs (specify ACL when creating an event or series, or updating ACL when editing an event or series), the user must be able to enter a partial or complete role string, and the UI must query the current contents of the input box to see if it's a partial or complete match, e.g.:

/admin-ng/resources/ROLES.json?filter=role_name:ext:1_Learner,role_target:ACL&limit=100&offset=0

where the input string is "ext:1_Learner". In this case, the role is matched by the CustomRoleProvider, and so that role becomes selectable and can form part of an ACL. This is the same mechanism used by the external LMS providers except they will do a webservice call to validate rather than match against a regexp (some will do both for efficiency). Provides may return more than one role for a partial match, e.g. given a query SITEID, the Sakai provider will return possible matches SITEID_Learner and SITEID_Instructor in which case both should be offered and be selectable by the UI.

smarquard:
7.x example with CustomRoleProvider configured:

image

This is not as elegant as it should be as for some reason other non-matching roles still show up in the drop-down list including ones that shouldn't be there at all (ROLE_UI_*). So it's querying and offering the custom role correctly but the UI needs cosmetic fixing.

smarquard:
I would suggest targeting the user and group UI role selection fixes at 7.x because it's a UI bug that some roles don't show up for search until you've scrolled all the way through.

KatrinIhler:
@smarquard I rather wouldn't fix just some of the modals since this problem also exists for series and events and then that behaviour would be inconsistent.
Do I understand you correctly that my PR breaks case 2 and 3 for events and series?

smarquard:
@KatrinIhler, yes, cases 2 and 3 are broken by this PR. But the "search before scroll" problem does not affect ACL entry in events and series because those input fields already do progressive search as you type from the backend, so you'll always get a search result without having to scroll first.

On develop.stable.org, the population of the role list in the Event ACL is initially wrong, because it's calling ROLES.json only without filtering by ACL target. The Series ACL input is correct though.

gregorydlogan:
@KatrinIhler should this go into 8.x, or develop?

gregorydlogan:
Please ping me when this is ready to go.

gregorydlogan:
@KatrinIhler Is the force push meaning this is ready?

KatrinIhler:
@gregorydlogan Nope, not yet, but I'm working on it. It's mostly done, but there's a minor thing that's still bugging (ha!) me. Sorry for the confusion.

KatrinIhler:
Ready to review! * crosses fingers*

PR #1047, opened 2019-09-05
MH-13675: ServiceRegistry not updating database correctly when dispatching jobs

rute-santos:
I've seen 2 cases where the database is not correctly updated:

1) When the job cannot be dispatched and is put into the dispatchPriorityList: the job's processor service is never set to null in the database.

2) When the job changes status from QUEUED to DISPATCHING: the value DISPATCHING never makes it to the database.

Note that these are not critical because #1 will "fix itself" in one of the next runs of the dispatcher and #2 will "fix itself" by the worker when it accepts the job, but it has caused us at DCE some trouble because of our horizontal scaling mechanism that needs to have accurate info in the database :)

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
@@ -1010,7 +1010,8 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
     try {
       tx.begin();
       fromDb = em.find(JpaJob.class, job.getId());
-      originalJob = JpaJob.from(fromDb.toJob());
+      // MH-13675 This resets the job state in the 'job' variable back to the previous value so commenting it out
+      // originalJob = JpaJob.from(fromDb.toJob());
       if (fromDb == null) {
         throw new NoResultException();
       }
@@ -1022,7 +1023,6 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
       setJobUri(job);
       return job;
     } catch (PersistenceException e) {
-      dumpJobs(originalJob, fromDb);
       if (tx.isActive()) {
         tx.rollback();
       }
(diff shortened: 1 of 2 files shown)


modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
```diff
@@ -401,19 +408,20 @@ public void testDispatchingJobsHigherMaxLoad() throws Exception {
     properties.put("dispatchinterval", "1000");
     serviceRegistryJpaImpl.updated(properties);
     registerTestHostAndService();
-    Job testJob = serviceRegistryJpaImpl.createJob(TEST_HOST, TEST_SERVICE, TEST_OPERATION, null, null, true, null,
+    Job testJob = serviceRegistryJpaImpl.createJob(TEST_HOST, TEST_SERVICE_FAIRNESS, TEST_OPERATION, null, null, true,
+            null,
             10.0f);
```
rute-santos, line 413: @gregorydlogan, I changed this to use TEST_SERVICE_FAIRNESS because the mock http client always returns 503 for TEST_SERVICE - TEST_PATH.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -1010,7 +1010,8 @@ protected JpaJob updateInternal(EntityManager em, JpaJob job) throws Persistence
     try {
       tx.begin();
       fromDb = em.find(JpaJob.class, job.getId());
-      originalJob = JpaJob.from(fromDb.toJob());
+      // MH-13675 This resets the job state in the 'job' variable back to the previous value so commenting it out
```
gregorydlogan: I would just remove this entirely since all references to `originalJob` are now gone...  While you're at it, removing :1008 probably makes sense as well.

gregorydlogan (changes requested):
Looks good, minor quibble.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
rute-santos: Done! I had missed the originalJob declaration... Thanks!

gregorydlogan:
Odd, I wouldn't have thought the changes in the latest commit could break anything.  I'll restart the tests, but if you see another failure after this @rute-santos it might be something to check again.

PR #1048, opened 2019-09-06
Add an optional build step to clean easily clean the frontend caches

JulianKniephoff:
This is a compromise between #1044 and #1045. It adds a build profile to the top level POM, which removes all the front end related caches (`node_modules`, `node` and `bower_components` at the moment) from the source tree during Mavens `clean` lifecycle.

You can activate it like so:

```
mvn clean -PcleanFrontendCaches
```

### DIFF
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -537,34 +537,6 @@
           </instructions>
         </configuration>
       </plugin>
-      <plugin>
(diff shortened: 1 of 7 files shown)


gregorydlogan (approved):
This looks like a much better approach.

miesgre (approved):
It's ok to me

PR #1051, opened 2019-09-10
MH-13707: Watermark missing

gregorydlogan:
This changeset corrects some issues which prevented the watermark tool from working correctly.

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -1449,10 +1449,12 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
 
           Option<LaidOutElement<Attachment>> watermarkOption = Option.none();
           //If there is a watermark defined, it needs both args 8 and 9
-          if (arguments.size() > WATERMARK_INDEX) {
+          if (arguments.size() > WATERMARK_INDEX && arguments.size() <= WATERMARK_LAYOUT_INDEX) {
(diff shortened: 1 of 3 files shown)


uzk-rrzk:
Hi,

I tested your fix 
```
//If there is a watermark defined, it needs both args 8 and 9
if (arguments.size() > WATERMARK_INDEX && arguments.size() <= WATERMARK_LAYOUT_INDEX) {
```
without success, because 10 <= 9 will always fail.

I used 
`if (arguments.size() > 8)`
 instead of the old
` if (arguments.size() == 9)`
which works fine for me. Currently there are no other index after the watermark definition.

gregorydlogan:
Hi @uzk-rrzk, my bad, I miscounted the range.  I've also added a unit test.  This should work as expected now.

stefanosgeo:
Hi @gregorydlogan .

sorry for the late reply, due to Opencast DACH Summit didn't have time to test it. It's works for me also (multiple servers installation).

Thanks again for the fix.Great job!!!

Now the question... will it make it for 7.3? If yes, when 7.3 is going to be rolled out?

mliradelc:
I tested and I saw the code, is all ok from my part

> Hi @gregorydlogan .
> 
> sorry for the late reply, due to Opencast DACH Summit didn't have time to test it. It's works for me also (multiple servers installation).
> 
> Thanks again for the fix.Great job!!!
> 
> Now the question... will it make it for 7.3? If yes, when 7.3 is going to be rolled out?

I asked to @KatrinIhler and we agreed to release after the merge of this PR.

mliradelc (approved):
All ok from my part, tested and compiled with tests, all passed.

No objections about the code also.

gregorydlogan:
@mliradelc if you're happy then let's go ahead and merge this and then you can get 7.3 out the door.

PR #1054, opened 2019-09-11
MH-13710: Fix a bug in paella loader plugin when a track has no tags

miesgre:
Jira ticket: https://opencast.jira.com/browse/MH-13710

**Steps to reproduce:**
1. Process a mediapackage with a track without tags
2. Try to play with paella

**Actual Results:**
Paella fails to load the video

**Expected Results:**
Paella should play the video

### DIFF
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -143,8 +143,8 @@ class OpencastToPaellaConverter {
 
     var tracks = episode.mediapackage.media.track;
     var attachments = episode.mediapackage.attachments.attachment;
-    if (!(tracks instanceof Array)) { tracks = [tracks]; }
-    if (!(attachments instanceof Array)) { attachments = [attachments]; }
+    if (!(tracks instanceof Array)) { tracks = tracks ? [tracks] : []; }
+    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
 
     // Read the tracks!!
     tracks.forEach((currentTrack) => {
@@ -254,8 +254,8 @@ class OpencastToPaellaConverter {
 
     var attachments = episode.mediapackage.attachments.attachment;
     var catalogs = episode.mediapackage.metadata.catalog;
-    if (!(attachments instanceof Array)) { attachments = [attachments]; }
-    if (!(catalogs instanceof Array)) { catalogs = [catalogs]; }
+    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
+    if (!(catalogs instanceof Array)) { catalogs = catalogs ? [catalogs] : []; }
 
 
     // Read the attachments
@@ -338,7 +338,7 @@ class OpencastToPaellaConverter {
     var segments = [];
 
     var attachments = episode.mediapackage.attachments.attachment;
-    if (!(attachments instanceof Array)) { attachments = [attachments]; }
+    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
 
     // Read the attachments
     var opencastFrameList = {};
@@ -391,7 +391,7 @@ class OpencastToPaellaConverter {
     let otherPreview;
 
     var attachments = episode.mediapackage.attachments.attachment;
-    if (!(attachments instanceof Array)) { attachments = [attachments]; }
+    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
     attachments.forEach((currentAttachment) => {
       if (currentAttachment.type == 'presenter/player+preview') {
         presenterPreview = currentAttachment.url;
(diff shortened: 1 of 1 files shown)


modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -215,11 +215,17 @@ class OpencastToPaellaConverter {
...
+      let importT = false;
+      if (currentTrack.tags.tag) {
+        if (!(currentTrack.tags.tag instanceof Array)) {
+          currentTrack.tags.tag = [currentTrack.tags.tag];
+        }
+        importT = filterStream.tracks.tags.some(function(cTag) {
+          return currentTrack.tags.tag.some(function(t){
+            return ((cTag == '*') || (cTag == t));
+          });
         });
```
karendolan: @miesgre This solves the problem is an Opencast mediapackage TRACK element has no tags such as to "archive" the track or to publish it to "engage".... if the track was never tagged to be published to the search index. Where is Paella getting it from? I suspect this is related to the other recent "bug fix"  issue where error handing is needed for published events which have no creation date nor even title.

lkiesow: I think, I see where the confusion comes from: @karendolan, you don't need to tag a track to publish to engage. The `publish-engage` operation has always supported publication by flavor only. Tagging tracks is just a convenient way to mark things for publication without having to remember later on what needs publication.

miesgre: @karendolan Paella is getting the information from the search service. As @lkiesow there is no need to tag a track to upload to the engage. See the `download-source-flavors` key of the `publish-engage` operation.  But I found this problem on live events. When there is a live events, the tracks haven't tags.

lkiesow (commented):
Looking at the code, it seems that if `cTag == '*'`, tracks without tags would not be selected. Is that intentional? My expectation for configuring a global wildcard (`*`) would be that no filter applies. But I've also not looked into this very deeply.

miesgre:
> Looking at the code, it seems that if `cTag == '*'`, tracks without tags would not be selected. Is that intentional? My expectation for configuring a global wildcard (`*`) would be that no filter applies. But I've also not looked into this very deeply.

It's not intentional. I'm going to fix it.

miesgre:
@lkiesow Fixed. You can continue to review it

lkiesow:
There still seems to be a problem.
I get:

```
paella_player.js:4380 Uncaught (in promise) TypeError: Cannot read property 'currentTime' of null
    at paella_player.js:4380
```

I used the following workflow:

```xml
<definition xmlns="http://workflow.opencastproject.org">

  <id>import</id>
  <title>Import Mediapackage and Publish without reencode</title>
  <tags>
  </tags>
  <description>
  A minimal Workflow that only publishes a Mediapackage
  </description>

  <operations>

    <!-- Ingest elements from external working file repositories and file servers -->

    <operation
      id="ingest-download"
      description="Ingesting external elements">
      <configurations>
        <configuration key="delete-external">true</configuration>
      </configurations>
    </operation>

    <!-- Apply ACL from series to the mediapackage -->

    <operation
      id="series"
      fail-on-error="true"
      exception-handler-workflow="error"
      description="Applying access control entries from series">
      <configurations>
        <configuration key="apply-acl">true</configuration>
      </configurations>
    </operation>

    <!-- Inspect the media -->

    <operation
      id="inspect"
      fail-on-error="true"
      exception-handler-workflow="error"
      description="Inspecting audio and video streams">
      <configurations>
        <configuration key="overwrite">false</configuration>
        <configuration key="accept-no-media">false</configuration>
      </configurations>
    </operation>

    <!-- Publish to engage player -->

    <operation
      id="publish-engage"
      max-attempts="2"
      fail-on-error="true"
      exception-handler-workflow="error"
      description="Publishing to Engage">
      <configurations>
        <configuration key="download-source-flavors">dublincore/*,security/*,*/delivery</configuration>
        <configuration key="check-availability">true</configuration>
      </configurations>
    </operation>

    <!-- Archive the current state of the mediapackage -->

    <operation
      id="snapshot"
      fail-on-error="true"
      exception-handler-workflow="error"
      description="Archiving">
      <configurations>
        <configuration key="source-tags">archive</configuration>
      </configurations>
    </operation>

    <!-- Clean up the working file repository -->

    <operation
      id="cleanup"
      fail-on-error="false"
      description="Cleaning up">
      <configurations>
        <configuration key="delete-external">true</configuration>
      </configurations>
    </operation>

  </operations>

  <state-mappings>
    <state-mapping state="running">EVENTS.EVENTS.STATE_MAPPING.IMPORTING</state-mapping>
    <state-mapping state="failing">EVENTS.EVENTS.STATE_MAPPING.IMPORTING</state-mapping>
  </state-mappings>

</definition>
```

and ingested a video using:

```sh
curl -f -i --digest -u opencast_system_account:CHANGE_ME -H "X-Requested-Auth: Digest" http://localhost:8080/ingest/addMediaPackage/import -F 'flavor=presentation/delivery' -F 'BODY=@test.mp4' -F title="`uuidgen`" -F creator="`uuidgen`"
```

miesgre:
@lkiesow Fixed. You can continue the review

lkiesow:
Works. Merged.

PR #1056, opened 2019-09-11
NOISSUE: CAS security example is very out of date

gregorydlogan:
This patch fixes two issues with the existing CAS security example file:

1) It adds a number of missing paths which were added to the default file, but not to the example
2) It attempts to re-enable CAS, which appears to have been accidentally removed.

### DIFF
--- a/etc/security/security_sample_cas.xml-example
+++ b/etc/security/security_sample_cas.xml-example
@@ -311,7 +311,7 @@
     <!-- sec:x509 subject-principal-regex="CN=(.*?)," user-service-ref="userDetailsService" / -->
 
     <!-- Enable and configure the failure URL for form-based logins -->
-    <sec:form-login authentication-failure-url="/admin-ng/login.html?error" authentication-success-handler-ref="authSuccessHandler" />
+    <!-- <sec:form-login authentication-failure-url="/admin-ng/login.html?error" authentication-success-handler-ref="authSuccessHandler" / -->
 
     <!-- Authentication filter chain -->
     <sec:custom-filter position="BASIC_AUTH_FILTER" ref="authenticationFilters" />
--- a/etc/security/security_sample_cas.xml-example
+++ b/etc/security/security_sample_cas.xml-example
@@ -1,11 +1,25 @@
 <beans xmlns="http://www.springframework.org/schema/beans" xmlns:sec="http://www.springframework.org/schema/security"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:osgi="http://www.springframework.org/schema/osgi"
   xsi:schemaLocation="http://www.springframework.org/schema/beans
-       http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
+       http://www.springframework.org/schema/beans/spring-beans.xsd
        http://www.springframework.org/schema/osgi
        http://www.springframework.org/schema/osgi/spring-osgi.xsd
        http://www.springframework.org/schema/security
        http://www.springframework.org/schema/security/spring-security-3.1.xsd">
+
+  <!-- ######################################## -->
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Per discussion at https://groups.google.com/a/opencast.org/d/msg/users/NoZFBjxkzzg/aSa9t6UqBAAJ, this appears to be working for at least one user.

miesgre (commented):
It does not work for me. I get this error:
```
2019-09-16 11:14:03,154 | ERROR | (SpringSecurityConfigurationArtifactInstaller:102) - Unable to refresh spring security configuration file /opencast-miesgre/build/opencast-dist-develop-7-SNAPSHOT/etc/security/mh_default_org.xml: {}
org.springframework.beans.factory.parsing.BeanDefinitionParsingException: Configuration problem: Filter beans '<casFilter>' and '<org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter#0>' have the same 'order' value. When using custom filters, please make sure the positions do not conflict with default filters. Alternatively you can disable the default filters by removing the corresponding child elements from <http> and avoiding the use of <http auto-config='true'>.
Offending resource: OSGi resource[file:/opencast-miesgre/build/opencast-dist-develop-7-SNAPSHOT/etc/security/mh_default_org.xml|bnd.id=113|bnd.sym=opencast-kernel]
```

etc/security/security_sample_cas.xml-example
```diff
@@ -68,9 +310,21 @@
     <!-- Uncomment to enable x509 client certificates for identifying clients -->
     <!-- sec:x509 subject-principal-regex="CN=(.*?)," user-service-ref="userDetailsService" / -->
 
+    <!-- Enable and configure the failure URL for form-based logins -->
+    <sec:form-login authentication-failure-url="/admin-ng/login.html?error" authentication-success-handler-ref="authSuccessHandler" />
```
miesgre: Comment this line to make CAS work

miesgre (changes requested):
It does not work for me. I get this error:
```
2019-09-16 11:14:03,154 | ERROR | (SpringSecurityConfigurationArtifactInstaller:102) - Unable to refresh spring security configuration file /opencast-miesgre/build/opencast-dist-develop-7-SNAPSHOT/etc/security/mh_default_org.xml: {}
org.springframework.beans.factory.parsing.BeanDefinitionParsingException: Configuration problem: Filter beans '<casFilter>' and '<org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter#0>' have the same 'order' value. When using custom filters, please make sure the positions do not conflict with default filters. Alternatively you can disable the default filters by removing the corresponding child elements from <http> and avoiding the use of <http auto-config='true'>.
Offending resource: OSGi resource[file:/opencast-miesgre/build/opencast-dist-develop-7-SNAPSHOT/etc/security/mh_default_org.xml|bnd.id=113|bnd.sym=opencast-kernel]
```

gregorydlogan:
@miesgre this should be working now!

miesgre:
@gregorydlogan It works, but there are two issues:

1.  I get this warning. I don't know if it's important or not:
```
2019-09-17 12:04:53,826 | WARN  | (SingleSignOutHandler:147) - Front Channel single sign out redirects are disabled when the 'casServerUrlPrefix' value is not set.
```
2.  You need to configure a role provider to get roles, if not you get a bad credentials, and user is not logged in. This happens in the previous versions too (6.x, ...)

gregorydlogan:
@miesgre no idea, I don't have anything to test against, just attempting to revert the merge by sight.

Regarding the role provider, that's expected isn't it?

miesgre (approved):
@gregorydlogan I think it can be merged as is.
I can Log in and not noticed any problems. It works in the same way old releases works

PR #1057, opened 2019-09-15
MH-13712, Update ESLint

lkiesow:
This patch updates ESLint to its latest version.

### DIFF
--- a/modules/admin-ui/package-lock.json
+++ b/modules/admin-ui/package-lock.json
@@ -13,12 +13,12 @@
       }
     },
     "@babel/generator": {
-      "version": "7.5.5",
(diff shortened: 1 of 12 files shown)


mliradelc:
😮 I was trying this yesterday (The feature, not the PR itself). To use automatically with VScode I had to:

- move the file from `docs/checkstyle/eslintrc.js` to the project's root. 
- add a `.` before the filename to `.eslintrc.js`
- Then install `eslint-plugin-header` as a global packet `npm i eslint-plugin-header@latest --global`

With that steps it is possible to use ESlint automatically with VScode, but I don't know about other IDEs in how they look for the ESlint rules.

PR #1065, opened 2019-09-21
MH-13720 Remove the RoleProvider.getRoles() method

smarquard:
Remove the obsolete getRoles() method in the RoleProvider and RoleDirectoryService interfaces.

This method has not been used since the introduction of findRoles(...) which is used by both internal role providers and roleproviders for LMS integration (Sakai, Moodle, Brightspace) and the CustomRoleProvider (MH-12181).

Related:
MH-13692 Load all roles in Admin UI #1046


smarquard:
Not sure why the buildbot build failed, but it's unrelated to this commit.

lkiesow:
A change like this makes the interface pretty weird to develop against since you cannot rely at all on `getRoles()`. Also, the relationship between `getRoles()` and `findRoles()` is weird since from name and description, I would expect the latter to be equal to or a subset of the former. Having `getRoles()` and `findRoles("%", …)` return something entirely different would be really unexpected and may easily lead to problems. This is also my main reason for the objections: Attaching a number of spacial behaviors will only lead to people getting it wrong in the end. That's not good.

A better way, if you want to not implement a part of this interface, would be to split this apart and having an interface for providing user roles and one for listing roles.

If only `getRoles()` is problematic, shortly looking into the usage, it seems as if `getRoles()` isn't even really used except for easy to replace occurrences (e.g. a few times in `findRoles()` implementations). So maybe, we could even drop that part of the interface, getting rid of the problem for Sakai while still keeping a sane interface you can develop against.

smarquard:
It does seem possibly to just remove getRoles() from the RoleProvider interface. We could take this a little further by removing the method from RoleDirectoryService. It only seems called in that form from the REST endpoints /roles/ which I don't think are used anywhere (and could just findRoles).

smarquard:
This turned into more of a code cleanup (please squash on merge).

If there are other considerations regarding the Admin UI and the interaction with roles (the starting point of #1046) then I think we should tackle them in a separate issue or PR.

modules/security-aai/src/main/java/org/opencastproject/security/aai/ConfigurableLoginHandler.java
```diff
@@ -493,11 +493,7 @@ private JpaOrganization fromOrganization(Organization org) {
     }
   }
 
-  /**
-   * @see org.opencastproject.security.api.RoleProvider#getRoles()
-   */
-  @Override
-  public Iterator<Role> getRoles() {
+  private Iterator<Role> getRoles() {
```
lkiesow: This is used only by `findRoles()` and can easily be dissolved into that method.

lkiesow (approved):
This looks good. Thanks for tackling this. I've added a minor change which dissolves one of the leftover `getRoles()` methods. If you are happy with this as well, feel pree to merge this pull request at any time.

smarquard:
Thanks @lkiesow, looks good, so merged.

PR #1067, opened 2019-09-23
Publication Button show fix

ek926m:
This Pull Request closes the Publish Button Window by clicking anywhere. Additionally you can click the Escape Key to close the popup too.
image

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/touchController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/touchController.js
@@ -27,6 +27,12 @@ angular.module('adminNg.controllers')
       var popup = angular.element(document.getElementById(id));
       popup.toggle('is-hidden');
     };
+    $scope.clearPopup = function() {
+      var links = document.getElementsByClassName('js-popover');
+      for (var index = 0; index < links.length; index++) {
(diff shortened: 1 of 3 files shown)


modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/touchController.js
```diff
@@ -23,11 +23,36 @@
...
+        var links = document.getElementsByClassName('js-popover');
+        for (var index = 0; index < links.length; index++) {
+          links[index].style.display = 'none';
+        }
+      }
+    };
   }
-]);
+])
+.directive('escKey', function() {
```
gregorydlogan: I feel like the directive in this PR and #1073 should be moved into its own file, and then reused, rather than copy and pasted.

gregorydlogan (changes requested):

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/touchController.js
ek926m: @gregorydlogan I try to do a quick fix now

ek926m:
@gregorydlogan it should be fixed now.

Issue #1068, opened 2019-09-23
Service Registry Fails Tests on Buildbot


gregorydlogan:
**Describe the bug**
The service registry is increasingly failing the service registry unit tests with errors like this
```
[ERROR] Tests run: 12, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 19.391 s <<< FAILURE! - in org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest
[ERROR] testIgnoreHostsInPriorityList(org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest)  Time elapsed: 2.167 s  <<< FAILURE!
java.lang.AssertionError
	at org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest.testIgnoreHostsInPriorityList(ServiceRegistryJpaImplTest.java:344)
```

This appears to be related to the processing service registration, and thus is likely related to #1047.  I can't reproduce this at home, but it seems to be semi common on ci.opencast.org.

**To Reproduce**
1. Build Opencast develop (cur 8.x).  Has been sighted on earlier versions as well.

karendolan:
@gregorydlogan and @lkiesow I notice that Travis CI build throws the "Job with Id 1 cannot be deleted" during that test, and other similar sort of db errors occasionally on other tests. And then other times is doesn't. I also see this when I run the test locally. The assert error on line344 is testing "fromDb.setProcessorServiceRegistration(null);", that the persistence update was correct. Is it possible that the H2 db test framework is sometimes slower to respond, and sometimes faster? Is there a way to ensure that entity updates are completed before asserting on them, without setting a time delay?

karendolan:
I notice there's a later version of H2 db engine. I'll do a little testing to see if that helps.
https://www.h2database.com/html/main.html

Updating the h2 didn't help, I'm looking at the PersistenceUtil.newTestEntityManagerFactory("org.opencastproject.common").[EDIT] was the right package... now testing resetting H2 between test methods... looks Ok to me. 
The unit is randomly having a hard time dealing with the serviceregistry activate()'s "cleanUndispatchableJobs":  
```
...
org.h2.jdbc.JdbcSQLSyntaxErrorException: Table "OC_ORGANIZATION_PROPERTY" not found; SQL statement:
org.h2.jdbc.JdbcSQLSyntaxErrorException: Table "OC_ORGANIZATION_NODE" not found; SQL statement:
[...]

PR #1070, opened 2019-09-24
Temporarily Ignore Service Registry Test

lkiesow:
This patch makes Opencast's build system temporarily ignore a constantly
failing service registry test to avoid people loosing trust in the build
system.

Related to #1068

## Before you file, you pull request must...

* [ ] have a concise title.
* [ ] [closes an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] be against the correct branch (features can only go into develop).
* [ ] include migration scripts and documentation, if appropriate.
* [ ] pass automated testing.
* [ ] have a clean commit history.
* [ ] have proper commit message (title and body) for all commits.
* [ ] have appropriate tags applied.

### DIFF
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -58,6 +58,7 @@ import org.easymock.IAnswer;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Filter;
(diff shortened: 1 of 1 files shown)


smarquard:
I think it would be better to revert the commit that's causing the intermittent test failure.

karendolan:
> I think it would be better to revert the commit that's causing the intermittent test failure.

@smarquard that commit has a unit test that verifies the correct state of the database. Normally, the state is correct, but occasionally, the persistence returns an unexpected result during the unit test. I suspect it has to do with the H2 db speed, and possibly threading, between the unit test and the persistence state. IMHO reverting commits because the tests are not properly mocking them would void a lot of code. On the other hand, there may be a real underlying problem with the Opencast entity managers that this commit is actually revealing, but I hope not!

lkiesow:
There is an open issue for this and @gregorydlogan is working on that. But until it is fixed, the CI will fail again and again making people not trusting in it anymore. I've already restarted builds manually thrice today. That's the reason for this pull request and why I think it's worth merging this as fast as possible, unrelated to weather or not the actual issue lies in the code or the unit tests. Thoughts?

gregorydlogan:
I'm going to @ignore the test for now.  Hopefully I can get this fixed today, but having people not trust the build system isn't worth it right now.  Better to run without a frequently failing test than hold up development.

PR #1073, opened 2019-09-24
close esc function for new event and new series modals

ek926m:
This PR makes the modals for new event and new series closable by using the escape key.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js
@@ -985,17 +985,4 @@ angular.module('adminNg.controllers')
     });
   }
 
-])
-.directive('escKey', function() {
-  return function(scope, element, attrs) {
-    element.bind('keydown keypress', function(event) {
-      var keyCode = event.which || event.keyCode;
-      if (keyCode === 27) {
-        scope.$apply(function() {
-          scope.$eval(attrs.escKey);
-        });
-        event.preventDefault();
-      }
-    });
-  };
-});
+]);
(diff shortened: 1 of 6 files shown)


modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/eventController.js
```diff
@@ -984,4 +984,18 @@ angular.module('adminNg.controllers')
       cleanupScopeResources();
     });
   }
-]);
+
+])
+.directive('escKey', function() {
```
gregorydlogan: I feel like the directive in this PR and #1067 should be moved into its own file, and then reused, rather than copy and pasted.

modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/newSeriesController.js
```diff
@@ -135,4 +135,17 @@ angular.module('adminNg.controllers')
     });
 
     $scope.components = ResourcesListResource.get({ resource: 'components' });
-  }]);
+  }])
+  .directive('escKey', function() {
```
gregorydlogan: Same as above

gregorydlogan (changes requested):

KatrinIhler:
Why only for these modals? Won't this create an inconsistency with the others, or does this already work for the rest?

gregorydlogan:
@KatrinIhler brings up an even better point wrt moving this to a directive: It's much, much easier to share a directive around than copy and paste everywhere :)

lkiesow:
> Why only for these modals? Won't this create an inconsistency with the others, or does this already work for the rest?

It does already work for the rest. 

@gregorydlogan, @KatrinIhler: Please note that @ek926m does not work for us any longer and this was a last-minute fix to remove some inconsistencies. So it's either merging or declining this. Sry.

ek926m:
@lkiesow @KatrinIhler @gregorydlogan I try to do a quick fix now

ek926m:
@lkiesow @KatrinIhler @gregorydlogan it should be fixed now.

ek926m:
@gregorydlogan and yes, @lkiesow is right, I don't work for them anymore

PR #1080, opened 2019-09-25
Documentation reworked

marwyg:
The documentation of Opencast got a little rework.
- The new mkdocs theme, called “windmill”, allows more levels in the navigation, which leads to a compacter and better overview.
- The search field is showing results while typing.
- The theme itself got some css adjustments, so the design of the documentation looks more similar to the design of Opencast.
- Code snippets and blocks containing “etc/” are showing a little tooltip.
- Code blocks have a copy-to-clipboard button on the right side.
- Google Analytics code snippet on every page

opencast_docu


lkiesow:
@marwyg, can you add your new dependency to the Travis configuration so that build does not fail:

https://github.com/opencast/opencast/blob/d8c1bf1cda8a6c563dbc53e841c9cff8d0d47cec/.travis.yml#L68

@gregorydlogan will need to update the Buildbot config.

---

Also, can you add proper documentation to this pull request? Was is being changed? Why is it changed, … I have read your mail (and know anyway) but others may not have. Saying that it adds “some features” does not help.

gregorydlogan:
@lkiesow Buildbot should already have that dep.  It's complaining about the 'windmill' theme :)

Edit: Which is `mkdocs-windmill` when adjusting docs.

PR #1081, opened 2019-09-25
MH-13718 Add modal to edit metadata of multiple events

KatrinIhler:
This adds a new modal to the Start Tasks menu in which you can edit the metadata of all selected events at once. Currently only episode/dublincore metadata is supported, so no extended metadata.

In case there were problems getting the metadata for one or more events, the user will be shown what will wrong. They can decide to proceede with the rest of the events if they like.

Only selected metadata fields will be updated. Metadata fields can only be updated if this will result in a chance for at least one event. The metadata field also has to be valid (e.g. titles can't be empty). The fields cannot be set to different values for different events, but the opposite is possible: Setting a field to the same value that now contains different values. These fields are highlighted in blue.

If there were problems with updating events, this will be shown to the user afterwards so the user knows which events were updated successfully and which weren't.


karendolan:
Hi @KatrinIhler , if the user bulk edits metadata, can the user also bulk start the workflow to publish the edits? Or would the user have to reselect the rows and start a new task to publish the changes?

KatrinIhler:
@karendolan Currently this cannot start workflows. But it sounds like a nice idea for the future...

karendolan:
@Katrinlhler Do you know if there is a way to tell which events have changed metadata that has not been published? I ask because our users required that we add a custom button onto the event details metadata tab in the Admin UI, labelled "update everywhere", that is revealed when metadata for an event is changed in the tab. But, it disappears if you exit the event details and come back in. The button is just there to start a publish metadata workflow for the event. Our users protested about having to remember which event they edited to find it again and then to manually start a workflow. The bulk metadata update feature is useful. Our little custom button could be updated to bulk start publish metadata workflows. But, I'm curious about the strategy other sites use deal with this. Bulk editing metadata and then bulk finding which events were edited to bulk start a publish workflows sound harder than doing it for just one.

KatrinIhler:
@karendolan I'm not aware of an easy way to find events that have unpublished changes. One possibility for the batch-editing could be to not deselect the events when leaving the modal. However that only helps if you want to start the workflow right away...

mliradelc:
I'm Checking it 😄

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1181,6 +1182,122 @@ public Response getEventMetadata(@PathParam("eventId") String eventId) throws Ex
     return okJson(metadataList.toJSON());
   }
 
+  @POST  // use POST instead of GET because of a possibly long list of ids
```
mliradelc, line 1188: Instead of a Put, you will say??

mliradelc (changes requested):
First of all, is a very nice feature, very easy to use and its very handfull when you need to change some metadata from some events. I have some questions about the usability also, but I will leave that in another comment.

I just made a lot of tests of this new feature, but when I try with an unprivileged user (User only with access to the Admin UI and access to the capture agents) with events and series created by this user (That only have the role `ROLE_USER_{unprivileged_user_name}`) The new feature doesn't load and the Javascript console throws 403 error when trying to access:
`[Error] Failed to load resource: the server responded with a status of 403 (Access is denied) (metadata.json, line 0) http://<opencast-server>/admin-ng/event/events/metadata.json`

The user that created the series has the ACLs set to read and write in the selected events.

The list of roles that have the unprivileged user are:
ROLE_ADMIN_UI
ROLE_CAPTURE_AGENT_VBOXTEST
ROLE_UI_ACLS_CREATE
ROLE_UI_ACLS_DELETE
ROLE_UI_ACLS_EDIT
ROLE_UI_ACLS_VIEW
ROLE_UI_EVENTS_COUNTERS_VIEW
ROLE_UI_EVENTS_CREATE
ROLE_UI_EVENTS_DELETE
ROLE_UI_EVENTS_DETAILS_ACL_EDIT
ROLE_UI_EVENTS_DETAILS_ACL_VIEW
ROLE_UI_EVENTS_DETAILS_ASSETS_EDIT
ROLE_UI_EVENTS_DETAILS_ASSETS_VIEW
ROLE_UI_EVENTS_DETAILS_ATTACHMENTS_VIEW
ROLE_UI_EVENTS_DETAILS_COMMENTS_CREATE
ROLE_UI_EVENTS_DETAILS_COMMENTS_DELETE
ROLE_UI_EVENTS_DETAILS_COMMENTS_EDIT
ROLE_UI_EVENTS_DETAILS_COMMENTS_REPLY
ROLE_UI_EVENTS_DETAILS_COMMENTS_RESOLVE
ROLE_UI_EVENTS_DETAILS_COMMENTS_VIEW
ROLE_UI_EVENTS_DETAILS_MEDIA_VIEW
ROLE_UI_EVENTS_DETAILS_METADATA_EDIT
ROLE_UI_EVENTS_DETAILS_METADATA_VIEW
ROLE_UI_EVENTS_DETAILS_PUBLICATIONS_VIEW
ROLE_UI_EVENTS_DETAILS_SCHEDULING_EDIT
ROLE_UI_EVENTS_DETAILS_SCHEDULING_VIEW
ROLE_UI_EVENTS_DETAILS_STATISTICS_VIEW
ROLE_UI_EVENTS_DETAILS_VIEW
ROLE_UI_EVENTS_DETAILS_WORKFLOWS_DELETE
ROLE_UI_EVENTS_DETAILS_WORKFLOWS_EDIT
ROLE_UI_EVENTS_DETAILS_WORKFLOWS_VIEW
ROLE_UI_EVENTS_EDITOR_EDIT
ROLE_UI_EVENTS_EDITOR_VIEW
ROLE_UI_EVENTS_VIEW
ROLE_UI_GROUPS_CREATE
ROLE_UI_GROUPS_DELETE
ROLE_UI_GROUPS_EDIT
ROLE_UI_GROUPS_VIEW
ROLE_UI_JOBS_VIEW
ROLE_UI_LOCATIONS_DELETE
ROLE_UI_LOCATIONS_DETAILS_CAPABILITIES_VIEW, ROLE_UI_LOCATIONS_DETAILS_CONFIGURATION_VIEW, ROLE_UI_LOCATIONS_DETAILS_GENERAL_VIEW
ROLE_UI_LOCATIONS_DETAILS_VIEW
ROLE_UI_LOCATIONS_VIEW
ROLE_UI_NAV, ROLE_UI_NAV_CAPTURE_VIEW
ROLE_UI_NAV_CONFIGURATION_VIEW
ROLE_UI_NAV_ORGANIZATION_VIEW
ROLE_UI_NAV_RECORDINGS_VIEW
ROLE_UI_NAV_STATISTICS_VIEW
ROLE_UI_NAV_SYSTEMS_VIEW
ROLE_UI_SERIES_CREATE
ROLE_UI_SERIES_DELETE
ROLE_UI_SERIES_DETAILS_ACL_EDIT
ROLE_UI_SERIES_DETAILS_ACL_VIEW
ROLE_UI_SERIES_DETAILS_METADATA_EDIT
ROLE_UI_SERIES_DETAILS_METADATA_VIEW
ROLE_UI_SERIES_DETAILS_STATISTICS_VIEW
ROLE_UI_SERIES_DETAILS_THEMES_EDIT
ROLE_UI_SERIES_DETAILS_THEMES_VIEW
ROLE_UI_SERIES_DETAILS_VIEW
ROLE_UI_SERIES_VIEW
ROLE_UI_SERVERS_MAINTENANCE_EDIT
ROLE_UI_SERVERS_VIEW
ROLE_UI_SERVICES_STATUS_EDIT
ROLE_UI_SERVICES_VIEW, ROLE_UI_STATISTICS_ORGANIZATION_VIEW
ROLE_UI_TASKS_CREATE
ROLE_UI_THEMES_CREATE
ROLE_UI_THEMES_DELETE
ROLE_UI_THEMES_DETAILS_VIEW
ROLE_UI_THEMES_EDIT
ROLE_UI_THEMES_VIEW
ROLE_UI_USERS_CREATE
ROLE_UI_USERS_DELETE
ROLE_UI_USERS_EDIT
ROLE_UI_USERS_VIEW


Maybe I need another role so It can load the tool de metadata.

mliradelc:
About the usability:

When I tried with the admin, no problems with this new feature, but I have some doubts about some options inside.

Why allow a bulk Change of the Title and start date? 

The first one I tried and make all the selected events with the same title, making it confusing to differentiate each event from another. Meanwhile, if you change the start date for several events in the same location will create a schedule conflict.

Now also there another thing and can create confusion, **the Location field**. If you read it, it has the name of the capture Agent, so you can think that if you change the location **WILL CHANGE** the capture agent, but this not happens. To change the location in reality you have to change it in the Schedule tab instead.

The location issue is not an issue from this PR per se, it comes from how was designed the UI and the confusion that occurs between "Location" and "Capture agent" in two separate tabs. I recommend to look this issue in other PR. For this PR I suggest to remove the field "Location".

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
KatrinIhler: Nope, this would usually be a get request actually. Put is farther below. :)

mliradelc: Ahh ok I understand, I'm not accustomed to requesting data with a POST, but I understand the inherent limitations you have with the GET method in this type of requests.

mliradelc (commented):
Another thing that I forgot, there is no documentation to notice this new feature and how to use it.

KatrinIhler:
> Another thing that I forgot, there is no documentation to notice this new feature and how to use it.

I thought about where to put this documentation. IMO it only makes sense in the user guide, and I'm not sure how much sense it makes to add this while more important parts of that guide are terribly out of date/non-existent.

KatrinIhler:
> About the usability:
> 
> When I tried with the admin, no problems with this new feature, but I have some doubts about some options inside.
> 
> Why allow a bulk Change of the Title and start date?
> 
> The first one I tried and make all the selected events with the same title, making it confusing to differentiate each event from another. Meanwhile, if you change the start date for several events in the same location will create a schedule conflict.
> 
> Now also there another thing and can create confusion, **the Location field**. If you read it, it has the name of the capture Agent, so you can think that if you change the location **WILL CHANGE** the capture agent, but this not happens. To change the location in reality you have to change it in the Schedule tab instead.
> 
> The location issue is not an issue from this PR per se, it comes from how was designed the UI and the confusion that occurs between "Location" and "Capture agent" in two separate tabs. I recommend to look this issue in other PR. For this PR I suggest to remove the field "Location".

It's true that matching titles are a bit confusing, but since the events are identified by their ID, they are possible, so should be possible here, too. Also matching start dates don't lead to scheduling conflicts if you have non-scheduled events or different locations. Regarding locations: The event details show this, too, so I'd rather be consistent here.
(In general the metadata fields here aren't selected in any way, they're simply all metadata fields that events can have that aren't read-only.)

KatrinIhler:
> I just made a lot of tests of this new feature, but when I try with an unprivileged user (User only with access to the Admin UI and access to the capture agents) with events and series created by this user (That only have the role `ROLE_USER_{unprivileged_user_name}`) The new feature doesn't load and the Javascript console throws 403 error when trying to access:
> `[Error] Failed to load resource: the server responded with a status of 403 (Access is denied) (metadata.json, line 0) http://<opencast-server>/admin-ng/event/events/metadata.json`

You're right, thank you for catching this! I pushed a fix, so this should work now. :)

KatrinIhler:
@mliradelc Actually you can't create scheduling conflicts with this modal since the start date that is updated is the bibliographic one, not the technical one.

mliradelc:
> @mliradelc Actually you can't create scheduling conflicts with this modal since the start date that is updated is the bibliographic one, not the technical one.

You have right because is the metadata and not the scheduling data!. Now with the Location and the start-date fields can convert in a UX catastrophe. because they will not change the real "scheduling" parameters.

KatrinIhler:
@mliradelc But this is not a problem specific to this PR, since this is the same way in the event details.

lkiesow:
What's the status of this? Is there anything missing or any problem not addressed?

@mliradelc, will you take the official review (if so, please assign yourself to this pull request). If not, @wsmirnow, can you review this? You are testing (and by now using?) this at ETH anyway.

wsmirnow:
I would grab this PR if @mliradelc or someone else not do.

mliradelc:
Sorry the last time I misread the comment about the fix for the unprivileged user, I was waiting for the fix. But the fix it was ready. I will try it.

mliradelc:
All is ok, now works with unprivileged users. The issue with the bibliographical Date and location vs technical data is being discussed in another pull request (#1233 ).

Merged

mliradelc (approved):

PR #1082, opened 2019-09-25
Docs readme extended.

marwyg:
Docs readme extended by one command for the new upcoming documentation theme.

### DIFF
--- a/docs/guides/readme.md
+++ b/docs/guides/readme.md
@@ -20,6 +20,7 @@ Requires python virtual environment.
 % . ./venv/bin/activate
 % pip install mkdocs
 % pip install markdown_inline_graphviz_extension
+% pip install mkdocs-windmill
 % cd developer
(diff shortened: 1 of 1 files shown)


lkiesow:
This is part of pull request #1082 

@marwyg, don't worry about it too much now but you can simply extend pull requests by pushing additional commits to the same branch. This does not mean that you should push everything you do into one pull request (you absolutely shoudn't! :D) but don't split it up if it's part of one change.

gregorydlogan:
@lkiesow recursive link to PR -> my stack has overflowed ;)  I'm guessing this is supposed to be part of #1080 ?  If so, @marwyg please repoint to `r/6.x`

lkiesow:
He he mentioned today that this docs don't exist at all in 6.x which makes it hard to add there. I guess it should be fine like this.

PR #1088, opened 2019-09-26
MH-13476 Filter capture agent roles for ACLs

smarquard:
This is a repeat of commit f1e5a17457e6b9b1a6578491f4ac7298c687e845 for PR #818
which was merged into 6.x and later 7.x and develop, but subsequently got lost in
7.x and develop.


gregorydlogan:
@wsmirnow you did the previous merge, this look right to you still?

wsmirnow:
Yes, this  patch still make sense.

PR #1092, opened 2019-09-26
MH-13517 Set an absolute limit on solr query size

smarquard:
This imposes an absolute limit on the new of rows that can be returned for a solr search query. See MH-13517 for details.


karendolan:
@smarquard can you post the URL for that ticket? I can't find it in Opencast Jira and not sure where else to look.

smarquard:
https://opencast.jira.com/browse/MH-13517 but you'll probably need to be logged into JIRA as a user in the committers group.

gregorydlogan:
Does it make sense to fix the endpoint logic as well?

smarquard:
It does, yes, but this is a local change which I'm pushing upstream, and we haven't done further work to fix the endpoint logic. This seems like a good fix to have in place anyway, in case some other code or endpoint passed through a query without limit.

PR #1095, opened 2019-09-27
Add a new optional date_expected column to the transcription job table

smarquard:
This enables transcription service implementations to record the date/time
that a transcription job is expected to be completed.

This is to support the forthcoming Nibity API implementation.


lkiesow (changes requested):
This is missing an upgrade script

smarquard:
This table is new in 8.x so a conversion is only required if this is merged after 8.x. It would be preferable to get this into 8.x though.

The table was introduced by:

MH-13265, Automatic caption using Google speech to text api #823

smarquard:
Oh right I see it should go into the 7-8 script. Will do that.

PR #1102, opened 2019-09-29
Clean-up Fast Testing Workflow

lkiesow:
This patch cleans up the fast testing workflow, removing previously
implemented hacks and unnecessary steps while simplifying the user
interface.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/etc/workflows/fast.xml
+++ b/etc/workflows/fast.xml
@@ -17,37 +17,8 @@
   <configuration_panel>
     <![CDATA[
       <div id="workflow-configuration">
-        <fieldset>
-          <legend>Add a comment that the recording needs:</legend>
-          <ul>
-            <li>
-              <input id="comment" name="comment" type="checkbox" class="configField" value="true" />
-              <label for="comment">Review / Cutting</label>
-            </li>
-          </ul>
-        </fieldset>
-        <fieldset>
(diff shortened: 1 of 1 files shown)


etc/workflows/fast.xml
```diff
@@ -311,7 +220,7 @@
       exception-handler-workflow="partial-error"
       description="Archiving">
       <configurations>
-        <configuration key="source-tags">archive</configuration>
+        <configuration key="source-flavors">*/source,dublincore/*,security/*</configuration>
       </configurations>
```
karendolan, line 224: Are none of the generated encodes or images archived? Or, are working files snapshot'd along the way except for the original input files (which would be odd logic)? Does this final snpashot override the previous temporary working snapshots that contained encodes and images?

lkiesow: Encoded material is not and was never being archived by this worklfow. The archive tag was previously also added only to these specific files and the only reason it was used at all was that other workflows were expecting it to exist despite being optional. That problem has been fixed in the meantime since now every element retrieved from the asset manager automatically gets the `archive` tag. Hence, it was no longer necessary to use it here which means one less operation to run.

PR #1103, opened 2019-09-30
Fix PostreSQL Support

lkiesow:
Fix PostreSQL Support

This patch fixes a few issues with Opencast's PostgreSQL compatibility,
upgrading the Postgres driver and fixing a few SQL statements.

Note that the OAI-PMH code still contains a hack with makes
auto-generated tables for OAI-PMH not work with MariaDB, MySQL or
PostgreSQL:

https://github.com/opencast/opencast/blob/8938bc896d837c6687345b3f674fc15321f012ec/modules/oaipmh-persistence/src/main/java/org/opencastproject/oaipmh/persistence/OaiPmhEntity.java#L87-L88

But simply removing this hack will break our internal H2 database right
now so I'm leaving it like this.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
+++ b/modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
@@ -66,7 +66,7 @@ import javax.persistence.UniqueConstraint;
                 + "WHERE job.processorServiceRegistration.online=true and job.processorServiceRegistration.active=true and job.processorServiceRegistration.hostRegistration.maintenanceMode=false "
                 + "AND job.status in :statuses "
                 + "AND job.creatorServiceRegistration.serviceType != :workflow_type "
-                + "GROUP BY job.processorServiceRegistration.hostRegistration.baseUrl, job.status"),
+                + "GROUP BY job.processorServiceRegistration.hostRegistration.baseUrl, job.status, job.processorServiceRegistration.hostRegistration.maxLoad"),
         @NamedQuery(name = "ServiceRegistration.getRegistration", query = "SELECT r from ServiceRegistration r "
                 + "where r.hostRegistration.baseUrl = :host and r.serviceType = :serviceType"),
         @NamedQuery(name = "ServiceRegistration.getAll", query = "SELECT rh FROM ServiceRegistration rh WHERE rh.hostRegistration.active = true"),
--- a/modules/db/pom.xml
+++ b/modules/db/pom.xml
@@ -66,8 +66,13 @@
       <artifactId>mysql-connector-java</artifactId>
     </dependency>
     <dependency>
-      <groupId>postgresql</groupId>
+      <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>net.java.dev.jna</groupId>
+      <artifactId>jna</artifactId>
+      <version>4.1.0</version>
     </dependency>
   </dependencies>
   <build>
@@ -86,6 +91,7 @@
               !org.apache.lucene.*,
               !org.jboss.resource.adapter.jdbc.*,
               !org.hibernate.*,
+              !com.sun.jna.platform.win32,
               org.w3c.dom;version=0,
               *
             </Import-Package>
@@ -95,6 +101,7 @@
               h2;inline=true,
               mchange-commons-java;inline=true,
               mysql-connector-java;inline=true,
+              jna,
               postgresql;inline=true
             </Embed-Dependency>
             <Export-Package>
--- a/pom.xml
+++ b/pom.xml
@@ -1324,9 +1324,9 @@
         <version>5.1.43</version>
       </dependency>
       <dependency>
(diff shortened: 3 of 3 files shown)


gregorydlogan:
Does #1105 resolve the broken H2/hack thing?

modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
```diff
@@ -66,7 +66,7 @@
                 + "WHERE job.processorServiceRegistration.online=true and job.processorServiceRegistration.active=true and job.processorServiceRegistration.hostRegistration.maintenanceMode=false "
                 + "AND job.status in :statuses "
                 + "AND job.creatorServiceRegistration.serviceType != :workflow_type "
-                + "GROUP BY job.processorServiceRegistration.hostRegistration.baseUrl, job.status"),
+                + "GROUP BY job.processorServiceRegistration.hostRegistration.baseUrl, job.status, job.processorServiceRegistration.hostRegistration.maxLoad"),
```
gregorydlogan, line 69: This seems unrelated to fixing Postgres?

lkiesow: It's related. PostgreSQL will not allow thew previous query …and if you think about it, it does make sense that having multiple values for `maxload` would be problematic in the query. I guess MariaDB is just ignoring additional values.

lkiesow:
> Does #1105 resolve the broken H2/hack thing?

No. OAI-PMH is still broken for auto-generated databases in PostegreSQL and MariaDB and this isn't something you can easily change. I decided to ignore that since there are multiple other related problems with OAI-PMH as well which need to be fixed at some point anyway.

modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
smarquard: mysql will also not allow that query (before the change) in certain modes: https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html

gregorydlogan: Huh, I suppose it's technically possible that such a value could exist, but our business logic precludes that from actually happening.  Interesting that one driver objects.

gregorydlogan:
I'm find with merging this as is, but don't have a Postgres install to test against.  Can someone with Postgres confirm that this works as expected?

lkiesow:
@gregorydlogan, would it help if I write you a short PostgreSQL guide? If yes, for Docker? For Debian? For Fedora?

gregorydlogan:
@lkiesow a docker guide would be great.

lkiesow:
@gregorydlogan, sorry. Forgot about this. What you want to do is basically:

#### Start PostgreSQL using docker, podman, …

    docker run --name postgres -p 5432:5432 \
      -e POSTGRES_PASSWORD=opencast \
      -e POSTGRES_USER=opencast \
      -e POSTGRES_DB=opencast -d postgres

#### Configure Opencast

```diff
diff --git a/etc/custom.properties b/etc/custom.properties
index e8d7944179..ab716ac6ed 100644
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -131,16 +131,16 @@ org.opencastproject.db.ddl.generation=true
 # db.vendor can be any of the values listed at under the "eclipselink.target-database" section of
 # http://www.eclipse.org/eclipselink/documentation/2.4/jpa/extensions/p_target_database.htm#target-database
 # Common values include MySQL, PostgreSQL.
-#org.opencastproject.db.vendor=MySQL
+org.opencastproject.db.vendor=PostgreSQL
 
 # Opencast comes with the jdbc drivers for MySQL (com.mysql.jdbc.Driver) and PostgreSQL (org.postgresql.Driver). To add
 # other jdbcDrivers to the Opencast runtime, rebuild the db module with your desired drivers.
-#org.opencastproject.db.jdbc.driver=com.mysql.jdbc.Driver
+org.opencastproject.db.jdbc.driver=org.postgresql.Driver
 
 # The jdbc connection url, username, and password
-#org.opencastproject.db.jdbc.url=jdbc:mysql://localhost/opencast
-#org.opencastproject.db.jdbc.user=opencast
-#org.opencastproject.db.jdbc.pass=dbpassword
+org.opencastproject.db.jdbc.url=jdbc:postgresql://127.0.0.1/opencast
+org.opencastproject.db.jdbc.user=opencast
+org.opencastproject.db.jdbc.pass=opencast
 
 # The jdbc connection pool properties. See http://www.mchange.com/projects/c3p0/#basic_pool_configuration
 # max.idle.time should be lower than the database server idle connection timeout duration (wait_timeout for mysql)
```

When testing, remember that OAI-PMH will fail right now.

KatrinIhler (approved):
I just tested this and apart from OAIPMH this works fine.

PR #1104, opened 2019-09-30
MH-13557, Login Response for JavaScript

lkiesow:
All JavaScript code letting user log-in to Opencast are using some kind
of ugly hack right now. For example, the engage interface is evaluating
if the response HTML contains a particular string.

For JavaScript, it would be much easier, better and more reliable if we
could be a proper HTTP status and JSON response.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/AuthenticationSuccessHandler.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/AuthenticationSuccessHandler.java
@@ -32,7 +32,6 @@ import org.springframework.security.core.Authentication;
 import java.io.IOException;
 import java.util.Map;
 
-import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
With this patch, what happens when the authentication is unsuccessful? I would probably still be redirected back to the login form, right? Is this what we want? :thinking:

modules/kernel/src/main/java/org/opencastproject/kernel/security/AuthenticationSuccessHandler.java
```diff
@@ -63,7 +62,12 @@
    */
   @Override
   public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response,
-          Authentication authentication) throws IOException, ServletException {
+          Authentication authentication) throws IOException {
+
+    if ("application/json".equalsIgnoreCase(request.getHeader("Accept"))) {
```
gregorydlogan, line 67: Perhaps I'm unclear on what exactly the Accept header can contain, but can't it also be a list?  What if it's 'application/json,application/xml'?

gregorydlogan:
> With this patch, what happens when the authentication is unsuccessful? I would probably still be redirected back to the login form, right? Is this what we want?

This seems reasonable to me.  What other behaviour would you expect here?

modules/kernel/src/main/java/org/opencastproject/kernel/security/AuthenticationSuccessHandler.java
gregorydlogan: Doesn't this also mean that *any* successful request with an accept for json gets redirected to `info/me.json`, or am I misreading what this class does?

lkiesow: But right now only if it's specifically `application/json`. If it's JSON or a web page, they will get the web page

smarquard: It doesn't seem totally intuitive to me to provide a redirect to an application/json request rather than say a 200 and a json response in the body like {"status":"success"}.

PR #1105, opened 2019-09-30
Use JPA to auto-generate SQL schema

lkiesow:
This patch shows how the JPA code can be changed to make Opencast
automatically generate a database schema on the same level as the
manually created DDL scripts do. This patch does not yet make the
full change but is limited to `oc_search` as an example.

The main differences we see and need to overcome are:

1. The ordering is different
2. Indexes are missing in the generated version
3. Constraints are missing in the generated version
4. Field types and sizes are different
5. Fields may be named slightly different

Ordering
--------

Unfortunately, it seems like [hibernate always sorts columns
alphabetically](https://stackoverflow.com/a/1298327) and there is no way
to change that. Thus, the only option would be to re-order the database
tables previously generated by Opencast's DDL script. That leads to a
long database migration.

However, the good thing about this is that, in theory, the ordering does
not matter to Opencast. But you never know for which tool it later does.

Generate Indexes
----------------

As this code shows, indexes can easily be generated using a few
additional JPA annotations.

Constraints
-----------

Again, this can easily be archived using JPA annotations although this
might be a little more complex and involve slight changes in Opencast's
Java code since the relations need to be modeled in Java as well.

Field Types
-----------

While it's a bit harder to get an exact type, exact types usually do not
matter and it is easy to ensure a fitting type is used by using the
correct Java type and annotations.

Field Names
-----------

In a few cases, the names in Opencast's code and in the DDL scripts are
slightly different (e.g. upper- vs lowercase letters being used). For
most databases this should not matter, but it's also easily fixed.

*Note that while this changes only parts of the JPA code it still
improves the overall situation (albeit slightly) and is worth merging.
Also note that this does not require any database migration since it
only brings the auto-generated SQL closer to the DDL scripts.*

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/search-service-impl/pom.xml
+++ b/modules/search-service-impl/pom.xml
@@ -45,6 +45,12 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-common-jpa-impl</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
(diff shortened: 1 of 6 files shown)


gregorydlogan:
While you're at it, do you want to address https://opencast.jira.com/browse/MH-12733 too?

lkiesow:
Maybe at some point. But this change is unrelated to that.

PR #1110, opened 2019-10-02
Build failed on captions-impl tests for non english OS 

rrolf:
Fixed problem that captions-impl Google Speech test failed when locale was not using "." as a delimiter for double values (i.e. on german systems), as the parsing of the time failed. Adjusted time parsing for IBM Watson accordingly, although it did not fail in the tests.


rrolf:
To test this you can use 

`export LANG=de_DE.utf8; mvn clean install -Pdev`

lkiesow (approved):

PR #1118, opened 2019-10-05
Removes String.format calls in logs

### LINKED ISSUE #1113: Remove superfluous calls to format() during logging
Per https://opencast.jira.com/browse/MH-11915

In many places in the codebase, we find statements like:
```
logger.info(format("Distributing %s elements to %s@%s", elementIds.size(), channelId, distributionChannel));
```
The call to the method format() is superfluous, since:
```
logger.info("Distributing {} elements to {}@{}", elementIds.size(), channelId, distributionChannel);
```
... does just the same.

The goal of this issue is to go through the codebase and remove those superfluous calls of format().

AdamMoas:
- Removes calls to String.format
- Adds missing message parameters to logger.warn and logger.error

Resolves: #1113

### DIFF
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -336,10 +336,10 @@ public class SeriesRestService {
     try {
       dc = this.dcService.load(new ByteArrayInputStream(series.getBytes("UTF-8")));
     } catch (UnsupportedEncodingException e1) {
-      logger.error("Could not deserialize dublin core catalog: {}", e1.getMessage(), e1);
+      logger.error("Could not deserialize dublin core catalog: {}", e1);
       throw new WebApplicationException(INTERNAL_SERVER_ERROR);
     } catch (IOException e1) {
-      logger.warn("Could not deserialize dublin core catalog: {}", e1.getMessage(), e1);
+      logger.warn("Could not deserialize dublin core catalog: {}", e1);
       return Response.status(BAD_REQUEST).build();
     }
     try {
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -336,10 +336,10 @@ public class SeriesRestService {
     try {
       dc = this.dcService.load(new ByteArrayInputStream(series.getBytes("UTF-8")));
     } catch (UnsupportedEncodingException e1) {
-      logger.error("Could not deserialize dublin core catalog: {}", e1);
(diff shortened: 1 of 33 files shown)


smarquard:
Build failure from unused imports, e.g.

[ERROR] /builds/Pull_Request_Reports/build/modules/common/src/main/java/org/opencastproject/util/persistencefn/PersistenceUtil.java:26:15: Unused import - java.lang.String.format. [UnusedImports]

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -336,10 +336,10 @@ public Response addOrUpdateSeries(@FormParam("series") String series, @FormParam
     try {
       dc = this.dcService.load(new ByteArrayInputStream(series.getBytes("UTF-8")));
     } catch (UnsupportedEncodingException e1) {
-      logger.error("Could not deserialize dublin core catalog: {}", e1);
+      logger.error("Could not deserialize dublin core catalog: {}", e1.getMessage(), e1);
```
smarquard: Think it's simpler and equivalent just to do:

      logger.error("Could not deserialize dublin core catalog", e1);

AdamMoas: It looked as though the logger.error() wasn't picking up the throwable in the message, but I'll revert that and remove the unused imports. Thanks!

smarquard: I mean the original is wrong in some way: either the intention is to log only the message in which case it should have been:

      logger.error("Could not deserialize dublin core catalog: {}", e1.getMessage());

or the intention was to log a full stack trace in which case it should have been:

      logger.error("Could not deserialize dublin core catalog", e1);

lkiesow: For an error, please use the latter. An error indicated that something is actually wrong. Hence, having a stack trace is really helpful.

lkiesow:
@AdamMoas, looks good and I can merge it like this if you want me to. However, there are [a number of format statements left](https://paste.fedoraproject.org/paste/uibJZCNWQ8rZQi~RopXOMQ). Do you want to take a look at them as well?

AdamMoas:
Oh, interesting. My search didn't find those somehow. I'll work on it. Thanks!

AdamMoas:
Hi, @lkiesow updating the logger in `AbstractAssetManagerPerformanceTest` removes the double's significant figures, but I've also noticed this test has been `@Ignore` since 2017, should this be removed at some point?

lkiesow:
Don't worry about it too much. If most usage of `format()` in the logger is gone that's already nice since people then don't start copying that thinking they have to use it. I'll review and (if possible) merge this patch later today.

lkiesow:
Merged. Thanks again for the patch.

PR #1120, opened 2019-10-05
Updates Service Registry dispatch interval property name and time unit

### LINKED ISSUE #1117: Service Registry configuration is inconsistent
Per https://opencast.jira.com/browse/MH-10535

etc/services/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.properties has these two settings:

dispatchinterval=0
heartbeat.interval=0

For consistency, they should be similarly named, i.e.

dispatch.interval=0
heartbeat.interval=0

and both specified in the same units (seconds) - currently dispatchinterval is in milliseconds, and heartbeat.interval in seconds.

cartucciam:
Changing dispatch interval property name and time unit to be consistent with the other settings (as per #1117).

Closes #1117


lkiesow (changes requested):
Hi @cartucciam, thanks for the patch. Can you please add an explanation of what you are doing and reasoning why you are doing this to the pull request's description?

Also, changing the dispatch interval's property name can easily cause problems if people miss this when upgrading. This does not necessarily mean that we cannot make that change but we should be pretty clear in the upgrade documentation about this.

cartucciam:
Hi @lkiesow, that's a good point and makes me wondering if this change is really needed. Can you point me at the correct upgrade documentation?

lkiesow:
I guess [docs/guides/admin/docs/upgrade.md](https://github.com/opencast/opencast/blob/develop/docs/guides/admin/docs/upgrade.md) would be the right place to announce a change like this.

etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
```diff
@@ -7,7 +7,7 @@ max.attempts=1
 # a mimimum value of 1s is enforced due to performance reasons. Set to 0 to disable dispatching from this service
 # registry.
 #Service registry dispatching should be set to 0 on everything but admin or allinone
-#dispatchinterval=5000
+#dispatch.interval=5
```
smarquard, line 10: Needs the reference above this changed: "# The interval in milliseconds"

smarquard: Also I wonder if there are actually any performance implications that need this to be so slow, or if we could change the default to 1s.

gregorydlogan: A sufficiently heavily loaded system can take longer than 1 second to do a full round of dispatching :(

smarquard: Yes but that doesn't matter because the setting is the interval between dispatch runs:

https://github.com/opencast/opencast/blob/develop/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java#L836

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html#scheduleWithFixedDelay(java.lang.Runnable,%20long,%20long,%20java.util.concurrent.TimeUnit)

We don't need to resolve that question for this PR but I don't think it needs to be that slow.

cartucciam: Changed to seconds

docs/guides/admin/docs/upgrade.md
```diff
@@ -4,6 +4,10 @@ Upgrading Opencast from 8.x to 9.x
 This guide describes how to upgrade Opencast 8.x to 9.x. In case you need information about how to upgrade older
 versions of Opencast, please refer to [older release notes](https://docs.opencast.org).
 
+Configuration changes
+----------------
+1. The dispatch interval property is now called _dispatch.interval_ and expects seconds instead of milliseconds (see [etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg](/etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg)).
+
```
cartucciam: @lkiesow is this enough to warn users of the breaking change?

lkiesow: Should do although I've removed the link since it would get invalid once published in our documentation. Finally, the release managers of Opencast 9 need to vet this document anyway. The most important part is that it is in there so it is not forgotten.

lkiesow:
Thanks for helping out with this → merged.

PR #1121, opened 2019-10-06
Remove unnecessary ExceptionUtils.getStackTrace #1119

### LINKED ISSUE #1119: Remove unnecessary ExceptionUtils.getStackTrace
At a lot of places `ExceptionUtils.getStackTrace(…)` is used for logging which is completely unnecessary since Opencast's logger is already able to handle errors on its while being more efficient. 

So instead of

```java
logger.error("Some error happened somewhere: {}", ExceptionUtils.getStackTrace(e));
```

…the code should be just:

```java
logger.error("Some error happened somewhere", e);
```

Note that no `{}` may be used in the logger for the exception and the exception is the last parameter in the logger call.

alexmario74:
Resolves #1119


gregorydlogan:
@alexmario74 conflicts.

lkiesow:
Looks good. Thanks, @alexmario74!
Merged.

PR #1122, opened 2019-10-07
fixes #1069 workflow tab shows scheduling info instead of workflows 

### LINKED ISSUE #1069: Event Details - Workflows fails to show workflow information for scheduled events.
**Describe the bug**
When going to event details ( trying this with failed events ) , workflows you should get an overview of the workflows and possible errors with the workflows. This works as intented for uploaded mediafiles but for scheduled events you get the workflow configuration view on that page, instead of the applied workflows and their status.

**Environment Information**
 - MacOS
 - Chrome
 - Opencast 7.2

kristofkeppens:
Directly use the event status instead of a separate logic to determine if an event is in scheduled state. This separate logic wasn't working in our case, showing the scheduled information instead of the running workflows in a lot of our events.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1513,7 +1513,7 @@ public abstract class AbstractEventEndpoint {
       return notFound("Cannot find an event with id '%s'.", id);
 
     try {
-      if (optEvent.get().getEventStatus().contains("SCHEDULE")) {
+      if (optEvent.get().getEventStatus().equals("EVENTS.EVENTS.STATUS.SCHEDULED")) {
         List<Field> fields = new ArrayList<Field>();
         Map<String, String> workflowConfig = getSchedulerService().getWorkflowConfig(id);
         for (Entry<String, String> entry : workflowConfig.entrySet()) {
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1513,7 +1513,7 @@ public abstract class AbstractEventEndpoint {
       return notFound("Cannot find an event with id '%s'.", id);
 
     try {
-      if (optEvent.get().isScheduledEvent() && !optEvent.get().hasRecordingStarted()) {
+      if (optEvent.get().getEventStatus().contains("SCHEDULE")) {
         List<Field> fields = new ArrayList<Field>();
         Map<String, String> workflowConfig = getSchedulerService().getWorkflowConfig(id);
         for (Entry<String, String> entry : workflowConfig.entrySet()) {


karendolan:
@flyapen checking that event status contains text that includes the sequence "schedule" is risky. I  don't know how much complexity it would introduce, have you considered promoting the (frequently hardcoded) text string "EVENTS.EVENTS.STATUS.SCHEDULED" into a first class Event status member in the event class:
https://github.com/opencast/opencast/blob/bdfbe174732ee1603f6277be0cec2d3843e4df79/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/Event.java#L88-L111

If it can be promoted to a true Event Status ScheduleState entity. That state will be more stably referenced, instead of as a potential REGEX match in a theoretical event status.

At one point, the community was considering the idea of queuing workflows... "EVENTS.EVENTS.STATUS.PROCESSING_SCHEDULED" ;)

karendolan:
I have a general scenario question: If the WorkflowCleanupScanner cleans old successful workflows that are 1 year old, does an event that had been scheduled, captured, ingested, and processed over 1 year ago register as an event in "scheduled" state?

1. Clean up old successful workflows:
https://github.com/opencast/opencast/blob/workflow/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowCleanupScanner.java#L174-L183

2. Event's source of eventStatus: updateEventStatus() https://github.com/opencast/opencast/blob/bdfbe174732ee1603f6277be0cec2d3843e4df79/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/Event.java#L870-L887

3.  From Event.java lines 906-909
```
  public boolean isScheduledEvent() {
    /* Only scheduled events have a capture ID assigned */
    return StringUtils.isNotBlank(getAgentId());
```

kristofkeppens:
@karendolan I did run into something similar a couple of days ago. I manually deleted all the workflows associated to an event because it was acting weird, and it resulted in that event being displayed under scheduled events. I was however an event that wasn't published yet, so that might have an impact on things.

karendolan:
> @karendolan I did run into something similar a couple of days ago. I manually deleted all the workflows associated to an event because it was acting weird, and it resulted in that event being displayed under scheduled events. I was however an event that wasn't published yet, so that might have an impact on things.

@flyapen the priority sequence of the current iteration of event status code seems to be (1) does the event have an associated workflow? Else (2), has event ever been associated with a capture agent? Else (3), the event must have been processed at some time in the past.

So I don't think publish comes into it. Also, it doesn't make sense for publish state to factored in unless the Asset Manager is no longer a player for assessing event state.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1513,7 +1513,7 @@ public Response getEventWorkflows(@PathParam("eventId") String id)
       return notFound("Cannot find an event with id '%s'.", id);
 
     try {
-      if (optEvent.get().isScheduledEvent() && !optEvent.get().hasRecordingStarted()) {
+      if (optEvent.get().getEventStatus().contains("SCHEDULE")) {
```
staubesv: I think what you want here is...

`if (optEvent.get().getEventStatus().equals("EVENTS.EVENTS.STATUS.SCHEDULED")) {`

... although I am somewhat shocked to see a string being used in https://github.com/opencast/opencast/blob/develop/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/Event.java#L882.

Obviously, the best practice would be to define a string constant and use the constant instead of copies of string values... but that is another issue.

karendolan: > Obviously, the best practice would be to define a string constant and use the constant instead of copies of string values... but that is another issue.

I agree with @staubesv: remove the reference to that weird hard coded string "EVENTS.EVENTS.STATUS.SCHEDULED" so making it into a constant does not become part of this pull.

gregorydlogan:
@flyapen are you going to have time to address the review comments here?

PR #1128, opened 2019-10-08
Load series ACL-list step by step

lwillmann:
If a series contains a large number of ACLs (more than 300) and one wants to edit the ACL-list or any other metadata of the series the UI is no longer responding. Especially if you are using the Chrome browser.

The problem is related to the way AngularJS tracks changes in the displayed items. The long list creates a very high number of so-called watchers that are polled regularly.
I have made some changes in the code to reduce the number of watchers.
In addition, initially only the first 50 ACL entries are loaded, by clicking on a link it is possible to reload 50 additional ACL entries until all ACL entries are loaded.


lkiesow:
@lwillmann, this seems to be based on the known to be broken version of Paella causing the builds to fail. Can you rebase this onto the current HEAD of develop?

lwillmann:
@lkiesow, done ;-)

KatrinIhler:
This actually breaks the loading of more roles: When the roles list is updated by the javascript, the UI doesn't change accordingly since the watcher is disabled. I'm gonna fix this in my PR, but we might want to reconsider if this use case (model change in the background making a UI update necessary) affects any of the other collections for which the watchers were disabled.

PR #1130, opened 2019-10-08
Updated com.fasterxml.jackson from version 2.9.9 to 2.10.0.

marwyg:
Build, Tests and Login worked.


lkiesow:
Rebased this since this was based on the known to be broken version of paella which broke the tests.

PR #1085, opened 2019-09-25
MH-12821 (#1008): Better crop detect test

gregorydlogan:
This change fixes the logic of a unit test, and throws errors when errors happen rather than relying on ffmpeg's return code

PR #1008, opened 2019-07-24
MH-12821 Crop service

lwillmann:
The pull request is based on the closed pull request #178 . 

The workflow operation excutes ffmpeg to detect and remove black bars from videos.

I've taken @isabellakutger work and refactored some parts of her code and added a few Tests.

PR #1151, opened 2019-10-15
MH-12821 (#1008): Better crop detect test #1085

gregorydlogan:
This change fixes the logic of a unit test, and throws errors when errors happen rather than relying on ffmpeg's return code


KatrinIhler (approved):
Looks reasonable! :)

PR #1153, opened 2019-10-18
Keep generated SMIL for partial tracks

lwillmann:
If one uses the IngestService and adds so called "partial"-tracks to a media package, the ingest service generates a SMIL-Catalog for these partial tracks. Unfortunately this generated SMIL-Catalog is removed if the option "skip.catalogs.for.existing.events" is set to true, which is the default. This pull request fixes this behaviour.

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -1379,7 +1379,9 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
     // drop catalogs sent by the capture agent in favor of Opencast's own metadata
     if (skipCatalogs) {
       for (MediaPackageElement element : mp.getCatalogs()) {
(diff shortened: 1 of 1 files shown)


lkiesow (approved):

lkiesow:
@lwillmann, looks good. Just wondering if we don't want this in `r/8.x` since it's a bug fix.

lwillmann:
@lkiesow yes, I think it would be good to have this in `r/8.x`. Is there something I have to do?

lkiesow:
Yes, you would probably have to rebase your branch, push this again and then update this pull request to point towards `r/8.x` (“Edit” right of the pull requests title).

For the rebase, it's more or less:

```
git fetch upstream  ← assuming `upstream` is the remote pointing to the community repo
git checkout keep-generated-smil
git rebase -i upstream/r/8.x  ← remove all commits except yours
git push --force-with-lease origin keep-generated-smil
```

We can take a short look at this together on Tuesday if you want

lwillmann:
Ok, thanks for the guidance @lkiesow. Rebased to `r/8.x`

PR #1154, opened 2019-10-18
Show users with same mail address and name

KatrinIhler:
Always use the username as key when sending users lists to the admin ui since only that is supposed to be unique. If it's the other way around, users with the same name and mail address cannot be added to a group that a similar user is already in.

### DIFF
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -168,11 +168,11 @@ prop.adminui.user.external_role_display=false
 # Name of the list to provide a user listing as used in the group editor of the admin ui.
 #
 # Values:
-# - USERS.INVERSE will list users by name (e.g. "John Sample")
-# - USERS.INVERSE.WITH.EMAIL will list users by name and email (e.g. "John Sample <j.sample@opencast.org>")
-# - USERS.INVERSE.WITH.USERNAME will list users by name and username (e.g. "John Sample (jsample)")
-# Default: USERS.INVERSE.WITH.USERNAME
-#prop.adminui.user.listname=USERS.INVERSE.WITH.USERNAME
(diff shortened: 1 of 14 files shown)


lkiesow:
@lwillmann, can you review this patch?

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -168,11 +168,11 @@ prop.adminui.user.external_role_display=false
...
-# - USERS.INVERSE will list users by name (e.g. "John Sample")
-# - USERS.INVERSE.WITH.EMAIL will list users by name and email (e.g. "John Sample <j.sample@opencast.org>")
-# - USERS.INVERSE.WITH.USERNAME will list users by name and username (e.g. "John Sample (jsample)")
-# Default: USERS.INVERSE.WITH.USERNAME
-#prop.adminui.user.listname=USERS.INVERSE.WITH.USERNAME
+# - USERS.NAME will list users by name (e.g. "John Sample")
+# - USERS.NAME.AND.EMAIL will list users by name and email (e.g. "John Sample <j.sample@opencast.org>")
+# - USERS.NAME.AND.USERNAME will list users by name and username (e.g. "John Sample (jsample)")
+# Default: USERS.NAME.AND.USERNAME
+prop.adminui.user.listname=USERS.NAME.AND.EMAIL
```
lwillmann: Please don't change the default configuration

lwillmann (changes requested):
Looks good for me, except the mentioned change in the configuration.

etc/org.opencastproject.organization-mh_default_org.cfg
KatrinIhler: Whoops, this wasn't intended. Fixed!

lwillmann (approved):

PR #1161, opened 2019-10-23
fix #1158, add config properties to prevent XSS attacks on session co…

### LINKED ISSUE #1158: Set https cookies with Secure and HttpOnly
Per https://opencast.jira.com/browse/MH-8659

**Description**
Cookie is set without flags e.g. 

JSESSIONID=hwh3algw3xyz;Path=/ 

Expected Results: 

If the session is https, cookie should set Secure and HttpOnly flags, e.g.: 

JSESSIONID=hwh3algw3xyz;Path=/; Secure; HttpOnly 

See https://www.owasp.org/index.php/Session_Management#Protect_Session_Identifiers and https://www.owasp.org/index.php/HttpOnly 

Via @JamesUoM 
```
Under Karaf with PAX web you can add following to etc/org.ops4j.pax.web.cfg:

org.ops4j.pax.web.session.cookie.httpOnly=true
org.ops4j.pax.web.session.cookie.secure=true

I don't know though if the secure flag requires jetty to be using HTTPS itself rather than setting behind a HTTPS proxy
```

JamesUoM:
Set httpOnly session cookie flag by default
Add property to encrypt cookie if using https

This fixes #1158

### DIFF
--- a/etc/org.ops4j.pax.web.cfg
+++ b/etc/org.ops4j.pax.web.cfg
@@ -41,3 +41,10 @@ org.ops4j.pax.web.ssl.keypassword=password
 # Custom jetty config file. Uncomment this if necessary to configure the jetty http connector idleTimeout value (MH-12329)
 # Adjust the settings in the file below to match the listening address and port settings above.
 #org.ops4j.pax.web.config.file=${karaf.etc}/jetty-opencast.xml
+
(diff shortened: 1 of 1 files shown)


lkiesow:
Looks good. Merged.

PR #1163, opened 2019-10-24
Improve embed code generation of Theodul player to create a fully responsive embed code fragment

DZenker:
Embed code generation improvement: IFRAME integrated into two nested DIV tags that defines a relative size of the embedded element, making it fully responsive to size changes of the browser window or portrait presentation modes on smartphones and preventing truncation of the embedded video.

In Theodul engage player, 5 different relative embed sizes (in percent) can be selected.


lkiesow:
Works as expected → merged.

PR #1167, opened 2019-10-27
Migrate IBM Watson transcription to shared persistence

### LINKED ISSUE #1166: Update the IBM Watson support to use the shared transcription persistence
The Google Speech support introduced a shared transcription persistence service.

We can reduce some code duplication by updating the IBM Watson transcription module to use this shared persistence service rather than its own.

smarquard:
Update the IBM Watson transcription service to use the shared persistence module.

There should be no functional change here, so the only requirement is that the PR doesn't break anything.

During a 7.x to 8.x migration, the SQL migration script migrates the old Watson transcription job data to the new shared job table, and drops the Watson-specific table.

Resolves #1166


modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
```diff
@@ -872,16 +875,32 @@ public void run() {
...
       try {
         // Find jobs that are in progress and jobs that had transcription complete i.e. got the callback
-        List<TranscriptionJobControl> jobs = database.findByStatus(TranscriptionJobControl.Status.Progress.name(),
+
+        long providerId;
+        TranscriptionProviderControl providerInfo = database.findIdByProvider(PROVIDER);
+        if (providerInfo != null) {
+          providerId = providerInfo.getId();
+        } else {
+          logger.debug("No provider entry for {}", PROVIDER);
```
gregorydlogan: I feel like this should be louder than debug.  If the dispatcher in the watson service can't find its own registration, something is wrong.

modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
```diff
@@ -872,16 +875,32 @@ public void run() {
...
+          return;
+        }
+
+        List<TranscriptionJobControl> jobs = database.findByStatus(TranscriptionJobControl.Status.InProgress.name(),
                 TranscriptionJobControl.Status.TranscriptionComplete.name());
 
         for (TranscriptionJobControl j : jobs) {
+
+          // Don't process jobs for other services
+          if (j.getProviderId() != providerId) {
```
gregorydlogan, line 894: While i'm sure there should not be *that* many jobs, I think having the query itself do this filtering probably makes more sense.  Perhaps in a follow on PR.

modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
smarquard: Updated to warn. It should never happen in practice.

modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
smarquard: Agreed (for a later PR), at which point Watson, Google Speech and Nibity modules can all be updated to match.

smarquard:
It would be preferable to get this into 8.0 release, otherwise the db changes have to be re-done for 9.x.

karendolan (approved):
It looks good. The changes are localized to IBM Watson and transcription persistence. The tests pass and are reasonable. The upgrade script is reasonable.

PR #1171, opened 2019-10-29
Removed wrong comma in .json example

nuokh:
Let's make examples valid again :-)


smarquard:
:-) Thank you for the contribution!

PR #1179, opened 2019-10-30
Make wowza configuration tenant-specific

KatrinIhler:
This PR moves a couple streaming configuration keys around to make it tenant-specific, allowing it to configure streaming for multi-tenant installations.
It also removes the "adaptive" from the class names, modules and so on to make naming more consistent.
Additionally, as RTMP is rarely used for on-demand streaming nowadays, the support was removed since this makes the code much easier to understand.

If you've got a Wowza installation, it would be great if you could test this!


etc/custom.properties
```diff
@@ -85,41 +85,24 @@ org.opencastproject.staticfiles.rootdir=${org.opencastproject.storage.dir}/stati
...
-# Server URL for Adaptive Streaming, usually the same server like RTMP streaming with a different protocol
-#org.opencastproject.adaptive-streaming.url=http(s)://<wowza-server>/<wowza-application>
+org.opencastproject.streaming.directory=${org.opencastproject.storage.dir}/streams
 
-# Port for Adaptive Streaming (for Wowza usually 1935)
-#org.opencastproject.adaptive-streaming.port=<port_number>
+# The mime types of streaming formats for which raw streaming URLs will be published automatically under the channel:
+# 'engage-player'. Use the character: '*' if all configured streaming URLs should be published.
+# The publication of streaming URI's is switched off by default
+org.opencastproject.publish.streaming.formats=*
```
lkiesow: If this should be the new default value, let's make it an actual default:

```suggestion
# Default: *
#org.opencastproject.publish.streaming.formats=*
```

etc/org.opencastproject.distribution.streaming.wowza.WowzaStreamingDistributionService.cfg
```diff
@@ -1,13 +1,21 @@
...
 
 # Configure the order in which the different video qualities will be written in the
 # SMIL files. The sorting criteria is the video bitrate.
 # Possible values are "ascending" or "descending".
 # Default: ascending
-#org.opencastproject.adaptive-streaming.smil.order=ascending
+#org.opencastproject.wowza.smil.order=ascending
+
+# Tenant specific Server URL for Streaming with Wowza
+# In case of multitenancy, add more configuration keys looking like this: org.opencastproject.<tenant>.wowza.url
```
lkiesow: What's the default? Is the default still empty and the service deactivated if this is empty/blank?

etc/org.opencastproject.distribution.streaming.wowza.WowzaStreamingDistributionService.cfg
```diff
@@ -1,13 +1,21 @@
...
+#org.opencastproject.mh_default_org.wowza.url=http(s)://<wowza-server>/<wowza-application>
+
+# Tenant specific Port for Wowza, usually 1935
```
lkiesow:
```suggestion
# Tenant specific port for Wowza, usually 1935
```

etc/org.opencastproject.distribution.streaming.wowza.WowzaStreamingDistributionService.cfg
```diff
@@ -1,13 +1,21 @@
...
+# In case of multitenancy, add more configuration keys looking like this: org.opencastproject.<tenant>.wowza.port
```
lkiesow: Is there a default?

modules/distribution-service-api/src/main/java/org/opencastproject/distribution/api/StreamingDistributionService.java
```diff
@@ -34,6 +34,8 @@
  */
 public interface StreamingDistributionService extends DistributionService {
 
+  boolean publishToStreaming();
```
lkiesow, line 42: What is this method for? Short JavaDoc? ;-p

modules/distribution-service-streaming-wowza/src/main/java/org/opencastproject/distribution/streaming/wowza/WowzaStreamingDistributionService.java
```diff
@@ -218,42 +191,28 @@
...
-      } catch (IllegalArgumentException e) {
-        logger.info("Adaptive streaming URL was not defined in the configuration file");
-      }
+    // get configuration
+    if (properties != null && bundleContext != null) {
 
-      if (adaptiveStreamingUri == null && streamingUri == null) {
-        throw new ComponentException("Streaming URL and adaptive streaming URL are undefined.");
-      }
+      streamingUrls = new HashMap<>();
```
lkiesow: 1. this should be a ConcurrentHashMap
2. You shouldn't set it first and then fill it, but fill it  and only at the end update the whole configuration at once

modules/distribution-service-streaming-wowza/src/main/java/org/opencastproject/distribution/streaming/wowza/WowzaStreamingDistributionService.java
```diff
@@ -276,49 +235,76 @@ public void activate(BundleContext bundleContext) {
...
+            }
+
+            streamingUrls.put(orgId, tenantStreamingUrl);
+            logger.info("Wowza Streaming URL for tenant {} set to \"{}\"", orgId, tenantStreamingUrl);
+          } catch (URISyntaxException e) {
+            throw new ComponentException(
+                    String.format("Wowza Streaming URL %s of tenant %s could not be parsed", streamingUrl, orgId), e);
+          }
+        } else {
+          logger.warn("Wowza Streaming URL is undefined for tenant {}", orgId);
```
lkiesow: Wouldn't that be a valid configuration?
This is basically warning people that they aren't using streaming which doesn't make sense, does it?
I suggest dropping this to debug level.

lkiesow (changes requested):

mliradelc:
Great work with this PR!! I'm starting to testing it right now, I need only to finish the configuration from the wowza site to put it to work.

I have a question, since opencast can sign the wowza url, in theory should work the `org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg` without issues right? because you can set a key per url.

mliradelc (changes requested):
This part in the documentation:

```markdown
4. Enter a name for the new application. You **must** use the same application name you have configured
in `$KARAF/etc/custom.properties` (for instance: *opencast-engage*) in `$KARAF/etc/custom.properties` (for instance: *opencast-engage*)
```

This is from the previous version and doesn't explain very well (Says name, but not what specific variable to change)

~~With also this in mind, Does each Wowza server needs to use the same application name?~~ 
Ex:
wowza.server.com/opencast-engage
wowza.server2.com/opencast-engage
wowza.server3.com/opencast-engage
... 

~~Or it can be different application names also??~~


Nevermind I figure out myself, the names correspond to the names of the applications you put in the URL.

modules/distribution-service-api/src/main/java/org/opencastproject/distribution/api/StreamingDistributionService.java
KatrinIhler: Whoops, forgot.

modules/distribution-service-streaming-wowza/src/main/java/org/opencastproject/distribution/streaming/wowza/WowzaStreamingDistributionService.java
KatrinIhler: Will do.

modules/distribution-service-streaming-wowza/src/main/java/org/opencastproject/distribution/streaming/wowza/WowzaStreamingDistributionService.java
KatrinIhler: Done.

etc/org.opencastproject.distribution.streaming.wowza.WowzaStreamingDistributionService.cfg
KatrinIhler: Yep, I'll add that to the comment.

KatrinIhler:
> Great work with this PR!! I'm starting to testing it right now, I need only to finish the configuration from the wowza site to put it to work.
> 
> I have a question, since opencast can sign the wowza url, in theory should work the `org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg` without issues right? because you can set a key per url.

Yeah, I think that still works, you'll need to set a key per application.

KatrinIhler:
@lkiesow I think I addressed everything. Thanks for taking a look at it! :)

mliradelc:
Tested. Worked without any issues. Wowza player is able to play from different streams.

I didn't test the streaming security

mliradelc (approved):

KatrinIhler:
@mliradelc Since you're a committer, you can merge this if you're fine with it.

PR #1218, opened 2019-11-07
[Documentation] Added path hint to upgrade.md

Solfur:
Added a hint to the upgrade.md to clearify the replacement of the three dots with the actual path.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -84,7 +84,8 @@ But old publications still reference the players directly in the admin interface
 pose any immediate problem unless you actually want to switch players in which case you would need to republish the
 material once to update the publication links.
 
-Alternatively, you can rewrite the old links all at once without re-publication using the following method:
+Alternatively, you can rewrite the old links all at once without re-publication using the following method:  
+*Hint: replace the '...' with the actual path to your installation-storage*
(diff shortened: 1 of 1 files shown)


docs/guides/admin/docs/upgrade.md
```diff
@@ -84,7 +84,8 @@ But old publications still reference the players directly in the admin interface
 pose any immediate problem unless you actually want to switch players in which case you would need to republish the
 material once to update the publication links.
 
-Alternatively, you can rewrite the old links all at once without re-publication using the following method:
+Alternatively, you can rewrite the old links all at once without re-publication using the following method:  
+*Hint: replace the '...' with the actual path to your installation-storage*
```
KatrinIhler, line 88: I think it's kind of self-explanatory, but in any case I'd prefer "Alternatively, you can rewrite the old links all at once without re-publication using the following method ( replace the '...' with the actual path to your installation):"

Solfur: #self-explanatory hmm it wasn't for me 
'..' is the previous folder, so i thought maybe this is some sed speciality i missed. Took me a while to figure that out - as those three dots look like real code

PR #1227, opened 2019-11-19
Delete option of series for the SearchService

CGreweling:
This commit implements the deletion option of a series for the search service.

If all episodes of a series a retracted from the search service,
the empty series still appears at the search service.

Therse was no way to delete a Series from the searchservice.

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
@@ -33,6 +33,7 @@ import org.opencastproject.metadata.dublincore.DublinCoreValue;
 import org.opencastproject.rest.AbstractJobProducerEndpoint;
 import org.opencastproject.search.api.SearchException;
 import org.opencastproject.search.api.SearchQuery;
+import org.opencastproject.search.api.SearchResult;
 import org.opencastproject.search.api.SearchResultImpl;
 import org.opencastproject.search.impl.SearchServiceImpl;
 import org.opencastproject.security.api.UnauthorizedException;
@@ -182,6 +183,50 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
     } catch (Exception e) {
       logger.info("Unable to remove mediapackage {} from search index: {}", mediaPackageId, e.getMessage());
       return Response.serverError().build();
+    }
+  }
+
+  @DELETE
+  @Path("/seriesId/{seriesid}")
+  @Produces(MediaType.APPLICATION_XML)
+  @RestQuery(
+      name = "removeSeries",
+      description = "Removes a series from the search index, if there are no episodes left connected to the series.",
+      pathParameters = {
+          @RestParameter(
+              description = "The series ID to remove from the search index.",
+              isRequired = true,
+              name = "seriesid",
+              type = RestParameter.Type.STRING
+          )
+      },
+      responses = {
+          @RestResponse(description = "The removing job.", responseCode = HttpServletResponse.SC_OK),
+          @RestResponse(
+              description = "There has been an internal error and the series could not be deleted",
+              responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+            )
+      },
+      returnDescription = "The job receipt")
+  public Response removeSeries(@PathParam("seriesid") String seriesId) throws SearchException {
+    try {
+      SearchQuery searchQuery = new SearchQuery();
+      searchQuery.withText(seriesId);
+      SearchResult searchResult = searchService.getForAdministrativeRead(searchQuery);
+      if (searchResult.size() > 0) {
+        logger.error("Can not delete series: '{}', it still contains episodes.", seriesId);
+        return Response.status(Response.Status.FORBIDDEN)
+            .entity("Can not delete series it still contains episodes.")
+            .build();
+      }
+
+      Job job = searchService.deleteSeries(seriesId);
+      return Response.ok(new JaxbJob(job)).build();
+    } catch (Exception e) {
+      logger.info(e.getMessage());
+      return Response.status(Response.Status.INTERNAL_SERVER_ERROR)
+          .entity("Unable to start job delete Series.")
+          .build();
     }
   }
 
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
@@ -133,7 +133,7 @@ public final class SearchServiceImpl extends AbstractJobProducer implements Sear
 
   /** List of available operations on jobs */
   private enum Operation {
-    Add, Delete
+    Add, Delete, DeleteSeries
   };
 
   /** Solr server */
@@ -428,6 +428,20 @@ public final class SearchServiceImpl extends AbstractJobProducer implements Sear
   }
 
   /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.search.api.SearchService#delete(java.lang.String)
+   */
+  public Job deleteSeries(String seriesId) throws SearchException, UnauthorizedException, NotFoundException {
+    try {
+      return serviceRegistry.createJob(
+          JOB_TYPE, Operation.DeleteSeries.toString(), Arrays.asList(seriesId), deleteJobLoad);
+    } catch (ServiceRegistryException e) {
+      throw new SearchException(e);
+    }
+  }
+
+  /**
    * Immediately removes the given mediapackage from the search service.
    *
    * @param mediaPackageId
@@ -480,6 +494,40 @@ public final class SearchServiceImpl extends AbstractJobProducer implements Sear
       return success;
     } catch (SolrServerException | SearchServiceDatabaseException e) {
       logger.info("Could not delete media package with id {} from search index", mediaPackageId);
+      throw new SearchException(e);
+    }
+  }
+
+  /**
+   * Immediately removes the given series from the search service.
+   *
+   * @param seriesId
+   *          the series
+   * @return <code>true</code> if the series was deleted
+   * @throws SearchException
+   */
+  public boolean deleteSeriesSynchronously(String seriesId) throws SearchException {
+    SearchResult result;
+    try {
+      SearchQuery searchQuery = new SearchQuery();
+      searchQuery.withId(seriesId);
+      searchQuery.includeSeries(true);
+
+      result = solrRequester.getForWrite(searchQuery);
+      if (result.getItems().length == 0) {
+        logger.warn(
+                "Can not delete series {}, which is not available for the current user to delete from the search index."
+                    + "",
+                seriesId);
+        return false;
+      }
+      logger.info("Removing series {} from search index", seriesId);
+
+      Date now = new Date();
+      //only delete from searchindex, there is no series Element in the Database
+      return indexManager.delete(seriesId, now);
+    } catch (SolrServerException e) {
+      logger.info("Could not delete series with id {} from search index", seriesId);
       throw new SearchException(e);
     }
   }
@@ -646,6 +694,10 @@ public final class SearchServiceImpl extends AbstractJobProducer implements Sear
         case Delete:
           String mediapackageId = arguments.get(0);
           boolean deleted = deleteSynchronously(mediapackageId);
+          return Boolean.toString(deleted);
+        case DeleteSeries:
+          String seriesId = arguments.get(0);
+          deleted = deleteSeriesSynchronously(seriesId);
           return Boolean.toString(deleted);
         default:
           throw new IllegalStateException("Don't know how to handle operation '" + operation + "'");
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
@@ -203,7 +203,6 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
     }
   }
 
-
   /**
    * {@inheritDoc}
    *
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
@@ -214,6 +214,49 @@ public class SolrIndexManager {
   }
 
   /**
+   * Removes the entry with the given <code>id</code> from the database.
+   *
+   * @param seriesid
+   *          identifier of the series or episode to delete
+   * @param deletionDate
+   *          the deletion date
+   * @throws SolrServerException
+   *           if an errors occurs while talking to solr
+   */
+  public boolean deleteSeries(String seriesid, Date deletionDate) throws SolrServerException {
+    try {
+      QueryResponse solrResponse = null;
+      try {
+        SolrQuery query = new SolrQuery(Schema.ID + ":" + ClientUtils.escapeQueryChars(seriesid) + " AND -"
+                + Schema.OC_DELETED + ":[* TO *]");
+        solrResponse = solrServer.query(query);
+      } catch (Exception e1) {
+        throw new SolrServerException(e1);
+      }
+
+      if (solrResponse.getResults().size() == 0) {
+        logger.warn("Trying to delete non-existing Series {} from the search index", seriesid);
+        return false;
+      }
+
+      // Use all existing fields
+      SolrDocument doc = solrResponse.getResults().get(0);
+      SolrInputDocument inputDocument = new SolrInputDocument();
+      for (String field : doc.getFieldNames()) {
+        inputDocument.setField(field, doc.get(field));
+      }
+
+      // Set the oc_deleted field to the current date, then update
+      Schema.setOcDeleted(inputDocument, deletionDate);
+      solrServer.add(inputDocument);
+      solrServer.commit();
+      return true;
+    } catch (IOException e) {
+      throw new SolrServerException(e);
+    }
+  }
+
+  /**
    * Removes the entry with the given <code>id</code> from the database. The entry can either be a series or an episode.
    *
    * @param id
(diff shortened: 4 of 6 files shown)


karendolan:
@CGreweling What is the use case of this pull request? If it is used directly there are orphaned published artifacts and orphaned publication links in the Asset Managed mediapackage.  Is this endpoint intended to be used by a workflow? Or emergency index cleanup of unwanted series to prevent having to rebuild the search index?

CGreweling:
@karendolan  it the first step. 
As it is now A series never gets deleted from the search index. 
This is not good, if you have a Mediaportal, which is based on the search service. 
We have a lot of Series shown there, which have been deleted a long time ago.
And as is now, there is no way, to delete a series. This Patch gives the opportunity to do so.
It just the same way implemented as the delete Episode REST.

The next Step would be to use it out of the retract Workflow. 
For Example: If there is no Epiosode in this series, delete it from the Search Index,

gregorydlogan:
I'm wondering if we should fix the future issue (ie, orphaned episodes) by changing delete series to actually delete the series, and its episodes.  Is there a usecase where you would *not* want to delete the contents of the series as well as the series itself?

My naive reading of the description in this ticket had me assuming that the implementation would remove the episodes as well if there were any, however that does not look to be the case.  IMO this is a bug waiting to bite someone.

JamesUoM:
I don't think this we should add API calls that can break things.
The REST method (at least) should check that there are no episodes belonging to the series before removing it.

CGreweling:
In my opinion the Rest-Api for the searchservice is just a api for admins.
But, i understand your point @JamesUoM. Will add that case soon.

@gregorydlogan okay may be should try write a bit more next time. 
Would be james solution be enough for you?
I think if i remove a Series at the Rest-API i don't want to remove all containing Episodes. That should be done out of the admin ui.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -173,6 +173,47 @@ public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws
...
+            throw new UnauthorizedException(currentUser + " is not authorized to delete series " + seriesId);
+
+          searchEntity.setDeletionDate(deletionDate);
+          em.merge(searchEntity);
+        }
+        tx.commit();
+      } catch (NotFoundException e) {
+        throw e;
+      } catch (Exception e) {
+        logger.error("Could not delete Series {}: {}", seriesId, e.getMessage());
```
lkiesow:
```suggestion
```

You are re-throwing the exception anyway.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -443,6 +456,51 @@ public boolean deleteSynchronously(String mediaPackageId) throws SearchException
...
+   *           if the series did not exist
+   */
+  public boolean deleteSeriesSynchronously(String seriesId) throws SearchException, UnauthorizedException,
+          NotFoundException {
+    SearchResult result;
+    try {
+      result = solrRequester.getForWrite(new SearchQuery().withSeriesId(seriesId));
+      if (result.getItems().length == 0) {
+        logger.warn(
+                "Can not delete Series {}, which is not available for the current user to delete from the search index.",
```
lkiesow:
```suggestion
                "Cannot delete series {}, which is not available for the current user to delete from the search index.",
```

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -480,4 +521,21 @@ private SearchEntity getSearchEntity(String id, EntityManager em) {
...
+   * @param id
+   *          the series identifier
+   * @param em
+   *          an open entity manager
+   * @return the search entity, or null if not found
+   */
+  private SearchEntity getSeriesSearchEntity(String id, EntityManager em) {
+    Query q = em.createNamedQuery("Search.findBySeriesId").setParameter("seriesId", id);
+    try {
+      return (SearchEntity) q.getSingleResult();
```
lkiesow: This will return a single result. The database table stores events. A series can have multiple events.Just retrieving – and later marking as deleted – the first event of a series is a bit weird.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -213,6 +213,49 @@ public void clear() throws SolrServerException {
...
+      try {
+        SolrQuery query = new SolrQuery(Schema.DC_IS_PART_OF + ":" + ClientUtils.escapeQueryChars(seriesid) + " AND -"
+                + Schema.OC_DELETED + ":[* TO *]");
+        solrResponse = solrServer.query(query);
+      } catch (Exception e1) {
+        throw new SolrServerException(e1);
+      }
+
+      if (solrResponse.getResults().size() == 0) {
+        logger.warn("Trying to delete non-existing media package {} from the search index", seriesid);
```
lkiesow: This confuses media package id and series id.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -213,6 +213,49 @@ public void clear() throws SolrServerException {
...
+        return false;
+      }
+
+      // Use all existing fields
+      SolrDocument doc = solrResponse.getResults().get(0);
```
lkiesow, line 243: I think the document you get back here represents an event and you throw away all but the first event. I don't think this will work für series with multiple events, will it?

lkiesow (changes requested):
Removing a series is not necessarily a bad idea since it's nice to be able to fix a problem with an index and to do that without rebuilding the whole thing.

Nevertheless, I don't think this does what it's advertised to do. From what I see [the search service Solr index stores only events](https://github.com/opencast/opencast/blob/a5ef8620a777bb8db9ae56c972c6a826ba20b561/modules/search-service-impl/src/main/resources/solr/conf/schema.xml#L230) and you will actually delete – or rather mark as deleted – events which belong to the specified series. Well, at least the first one since you handle only that.

The [same thing is true for the representation in the database](https://github.com/opencast/opencast/blob/a5ef8620a777bb8db9ae56c972c6a826ba20b561/docs/scripts/ddl/mysql5.sql#L237) which again represents events and for which you delete only the first.

I'm also wondering if there is a specific reason to decouple the work from the REST interface and do it asynchronously instead of synchronously.  There should be no performance issue here and the decoupling means that users will get no feedback if the deletion was successful.

gregorydlogan:
@CGreweling Yes, James' solution would be just fine with me.

@lkiesow Thanks for the links.  I can see how this could get bogged down doing the delete, but that would likely be an edge case.  IMO doing the delete synchronously makes sense.  Or add a flag so it can do either :)

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
CGreweling: A Series has its own id. It will not return multiple events here if you search for an id from a series with multiple events.
then the quuery should serch for the field is part of. 
I hope it is implemented like this if not.. urgh

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
CGreweling: i can take a look at this. 
As it is now this case could not be, because the rest-endpoint does not allow to delete a series wich still contains episodes. So if it comes to this code, there is only one series(?episode?) Element left.

CGreweling: Yes good catch @lkiesow  now it searches for the right id.

CGreweling:
@lkiesow after long time, i solved the merge-conflcts and rebased it againts develop.
As discussed last Friday.
I Solved the issues here. So would be nice if you can take a look again.
The Rest-Api delteseries now checks if there is an episode left in the series, if there is none, deletion gets started.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -173,6 +173,45 @@ public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws
...
+          Organization currentOrg = securityService.getOrganization();
+          if (!AccessControlUtil.isAuthorized(acl, currentUser, currentOrg, WRITE.toString()))
+            throw new UnauthorizedException(currentUser + " is not authorized to delete series " + seriesId);
+
+          searchEntity.setDeletionDate(deletionDate);
+          em.merge(searchEntity);
+        }
+        tx.commit();
+      } catch (Exception e) {
+        logger.error("Could not delete series {}: {}", seriesId, e.getMessage());
```
lkiesow: Proper error logging:
```suggestion
        logger.error("Could not delete series {}", seriesId, e);
```

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
lkiesow: > A Series has its own id. It will not return multiple events here if you search for an id from a series with multiple events.

Yes, it will. The query you use is:

```java
@NamedQuery(name = "Search.findBySeriesId", query = "SELECT s FROM SearchEntity s WHERE s.seriesId=:seriesId and "
        + "s.deletionDate is null"),
```

The database structure is:

```
MariaDB [opencast]> desc oc_search;
+-------------------+--------------+------+-----+---------+-------+
| Field             | Type         | Null | Key | Default | Extra |
+-------------------+--------------+------+-----+---------+-------+
| id                | varchar(128) | NO   | PRI | NULL    |       |
| series_id         | varchar(128) | YES  | MUL | NULL    |       |
| organization      | varchar(128) | YES  | MUL | NULL    |       |
| deletion_date     | datetime     | YES  |     | NULL    |       |
| access_control    | longtext     | YES  |     | NULL    |       |
| mediapackage_xml  | longtext     | YES  |     | NULL    |       |
| modification_date | datetime     | YES  |     | NULL    |       |
+-------------------+--------------+------+-----+---------+-------+
7 rows in set (0.001 sec)
```

So, you are looking for all non-deleted entries in `oc_search` with a given `series_id`.
A result from the same production database:

```
MariaDB [opencast]> select count(*) from oc_search where series_id = 'c1ddd210-dcc1-446b-b649-151d1b2f2143' and deletion_date is null;
+----------+
| count(*) |
+----------+
|       17 |
+----------+
1 row in set (0.001 sec)
```

So, you get 17 results.
The reason for that is that the `oc_search` database table holds events only. Series are only represented through the existence of events. You cannot have a series without events. You are just retrieving the first event of the series and delete that one. This won't work if there are two non-deleted events in that series.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -443,6 +456,55 @@ public boolean deleteSynchronously(String mediaPackageId) throws SearchException
...
+        logger.info("Removed series {} from search persistence", seriesId);
+      } catch (NotFoundException e) {
+        logger.info("Could not find series with id {} in persistence, but will try remove it from index, anyway.",
+                seriesId);
+      } catch (SearchServiceDatabaseException e) {
+        logger.error("Could not delete series with id {} from persistence storage", seriesId);
+        throw new SearchException(e);
+      }
+
+      return indexManager.delete(seriesId, now);
```
lkiesow, line 528: While you did delete events before in the database, you only delete the series in the index, potentially leaving database and index out of sync. The index does have a series entry btb.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -146,6 +147,30 @@ public Response remove(@PathParam("id") String mediaPackageId) throws SearchExce
...
+    try {
+      SearchQuery searchQuery = new SearchQuery();
+      searchQuery.withText(seriesId);
+      SearchResult searchResult = searchService.getForAdministrativeRead(searchQuery);
+      if (searchResult.size() > 0) {
+        logger.error("Can not delete series: '{}', it still contains episodes.", seriesId);
+        return Response.serverError().build();
+      }
+
+      Job job = searchService.deleteSeries(seriesId);
```
lkiesow, line 223: I'm curious, why use a `Job` and do this asynchronously? Shouldn't the delete operation be fast enough? Like this, users will not get any errors.

lkiesow (dismissed):
Unfortunately, there are still some errors, potentially leaving the database and index out of sync.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
CGreweling: As mentioned, its just implemented as the rest.
As communicated, i would close this diescussion.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
CGreweling: As we talked some days ago.
There is a condition now in the REST, so if there is no event left in the series; only then the deletion would take place.
I remove, the deletion from  the database, because there is now series object in the database.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
```diff
@@ -161,6 +161,7 @@ AccessControlList getAccessControlList(String mediaPackageId) throws NotFoundExc
   void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws SearchServiceDatabaseException,
           NotFoundException;
 
+
```
KatrinIhler: I assume this change was unintended? :D

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -213,6 +213,49 @@ public void clear() throws SolrServerException {
...
+   * Removes the entry with the given <code>id</code> from the database.
+   *
+   * @param seriesid
+   *          identifier of the series or episode to delete
+   * @param deletionDate
+   *          the deletion date
+   * @throws SolrServerException
+   *           if an errors occurs while talking to solr
+   */
+  public boolean deleteSeries(String seriesid, Date deletionDate) throws SolrServerException {
```
KatrinIhler, line 226: That's basically the same method as the one below, isn't it? I think it would be better to put that into a private method called deleteElement() or something, and just use it in the public method instead of duplicating the code.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -374,6 +374,8 @@ public Job add(MediaPackage mediaPackage) throws SearchException, MediaPackageEx
    *          the media package
    * @throws SearchException
    *           if the media package cannot be added to the search index
+   * @throws MediaPackageException
```
KatrinIhler: Does it really?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -484,6 +500,46 @@ public boolean deleteSynchronously(final String mediaPackageId) throws SearchExc
...
+   *          the series
+   * @return <code>true</code> if the series was deleted
+   * @throws SearchException
+   *           if deletion failed
+   * @throws UnauthorizedException
+   *           if the user did not have access to the series
+   * @throws NotFoundException
+   *           if the series did not exist
+   */
+  public boolean deleteSeriesSynchronously(String seriesId) throws SearchException, UnauthorizedException,
```
KatrinIhler: Some of those exceptions are not actually thrown by this code, e.g. Unauthorized & NotFound

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
KatrinIhler: So I assume this change is actually no longer necessary?

KatrinIhler (commented):
Didn't test this yet, but here are some of my thoughts already.

KatrinIhler:
I've got another question - do I understand this correctly that this only allows for a manual removal of a series via rest endpoint? Wouldn't it be nicer if the series was automatically removed when the last episode is removed?

KatrinIhler (changes requested):
I just tested this and if a series still contains episodes, the user gets a server error without any info to what they did wrong - that's not correct IMO. This should be a client error, maybe Forbidden? And it should contain something like "Series still containing episodes cannot be deleted."

KatrinIhler:
Okay, finally I can confirm that deleting via the rest endpoint works indeed as intended. :D

KatrinIhler:
So to get this straight: If the server error is changed to a client error and the unused code in the SearchServiceDatabase is removed, I'm going to be okay with the rest. Just in case @CGreweling still wants to work on this today. @gregorydlogan  Maybe you can take a look before you cut OC 10?

gregorydlogan:
This won't make `r/10.x` today, but IMO would be fine to merge in later.  @CGreweling don't forget to update the target when this is ready.

CGreweling (commented):
@KatrinIhler I fixed your findings as discussed. I hope i got everything finally.

KatrinIhler:
Christian doesn't really need this in OC 10, so I'm gonna merge this into develop.

KatrinIhler (approved):

PR #1231, opened 2019-11-21
Add option to enable ExecuteMany WOH to set workflow properties

smarquard:
This adds the ability for the ExecuteMany workflow operation handler to set workflow properties as output rather than creating a new track.

This is similar to the support for workflow properties for ExecuteOnce introduced in MH-12286.


smarquard:
Closing as this is incorporated in #1234

PR #1233, opened 2019-11-22
Change bibliographicdate if technicaldate is changed.

CGreweling:
Since 7.3 at the metadata Tab the Biographical date is shown.
This date is only changed when you edit the technical Date out of the Adminui(Scheduling-Tab).
But not within the Rest-APIs.

This Patch fixes this behaviour. The bibliographical date is always updated if the scheduling technical date is edited through the scheduler-service.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -169,9 +169,7 @@ import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.time.Instant;
-import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


karendolan:
@CGreweling can you add/update the  junit to verify that changing the capture agent (spacial) without changing the start and end time updates the episode catalog Ok?

gregorydlogan:
@CGreweling Unclear if this patch should resolve this, but testing a merge of this and #1122 exposed an issue where the scheduling data is inconsistent when changed from the admin ui's metadata page.  

If I change the start and/or duration on the metadata page, it persists correctly, but does not(!) update the schedule.  The scheduling tab in the admin ui retains the original data (ie, matching the scheduling data retrieved from the calendar endpoint), while the metadata tab retains the changed data (ie, matching the changes which were just made).

CGreweling:
@gregorydlogan Here begins the wired metadata things. I don't know if i unterstand your point correctly.
I was told, that the biographical Date (Schown on the matadata TAB), could be different to the techincal Date (Shown in the scheduling TAB). Exept the Scheduling Date is changed, than the bibiographical Date should be the same as the technical Date.
So if anybody can give me the right way, of handling these dates. I would add this to this patch. :)

karendolan:
@gregorydlogan & @CGreweling  Metadata start and technical start can be different AFTER ingest without impacting processes. But before ingest, if metadata/bibliographic start does not reflect the technical start, live published events are negatively impacted. The student might miss the technical start of the live event.

Our site prevents the user from changing the metadata's tab's start date and forces them to use the Scheduler tab to change the start date, in the one location, which change both dates.The one in the scheduler tab and the one in the metadata tab.

gregorydlogan:
Since this is not changing the behaviour within the UI, but rather fixing a mismatch between what the UI does, and what the REST endpoints do, I'm going to merge this as is.

PR #1234, opened 2019-11-22
MH-12391 Add audio and video stream selectors for tracks to ExecuteMany WOH

smarquard:
ExecuteMany WOH can limit element selection to tags and flavor, but in some cases you can have two presenter/source tracks, one video and one audio.

Add an option to restrict element selection to video and/or audio.

This branch also includes PR #1231: adds the ability for the ExecuteMany workflow operation handler to set workflow properties as output rather than creating a new track.

This is similar to the support for workflow properties for ExecuteOnce introduced in MH-12286.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/execute-many-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/execute-many-woh.md
@@ -1,8 +1,9 @@
 Execute Many Workflow Operation
 -------------------------------
 
-This operation handler filters a set of MediaPackageElements that match certain input conditions and runs a command on
-each of them.
+This operation handler filters a set of MediaPackageElements that match certain input conditions
+and runs a command on each of them. The command may be used to create a new mediapackage element,
+or to add configuration properties to the running workflow.
 
 To run a command only once for the whole mediapackage, use the [Execute Once](execute-once-woh.md) operation.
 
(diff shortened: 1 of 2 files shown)


docs/guides/admin/docs/workflowoperationhandlers/execute-many-woh.md
```diff
@@ -19,15 +20,27 @@ in [Execute Service: Parameter Substitution](../modules/execute.md#parameter-sub
...
+If `set-workflow-properties` is true, the command should write a plain-text properties file to the
+location specified by #{out} in the key-value format supported by the [Java Properties](http://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-)
+class, for example:
+
+````
+key1=value1
+key2=value2
+````
+
 ### Operation Example
```
KatrinIhler: Examples?

KatrinIhler (changes requested):
Took me a while to figure out how to test this, but it works. Just one minor thing below.

smarquard:
Thanks for the review @KatrinIhler. Any ideas about how to make it easier to test or for new adopters to use the operations? Maybe we should include some template scripts or something as examples.

smarquard:
Please squash on merge.

KatrinIhler:
@smarquard Yes, an example script that goes with the workflow operation examples sounds like a good idea!

PR #1235, opened 2019-11-22
Update selected components to use OSGI annotations

darolfes:
This pullrequest updates a set of selected components to use OSGI annotations instead of component-definition-xmls


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
```diff
@@ -248,30 +263,37 @@ public AQueryBuilder createQuery() {
...
   public void setEntityManagerFactory(EntityManagerFactory emf) {
     this.emf = emf;
   }
 
+  @Reference(name = "securityService")
   public void setSecurityService(SecurityService securityService) {
     this.secSvc = securityService;
   }
 
+  @Reference(name = "authSvc")
```
lkiesow, line 276: Just wondering, are these specific – and not necessarily well chosen – names actually used somewhere or would this work the same with just a bare `@Reference`?

lkiesow (dismissed):
- Tests fail
    - There are some tabs in `pom.xml` files (see Travis CI)
    - JavaDocs build fails when running `mvn -B -V -Dmaven.repo.local=/builder/m2 cobertura:cobertura site site:stage -Daggregate=true -Dcheckstyle.skip=true '-P "none,!frontend"'`
- Is this still unfinished? Why a draft pull request?
- Plus a question about the reference names (See below).

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/OsgiAssetManager.java
darolfes: I think the names aren't used and a bare `@Reference` would work the same, but I'm not sure, so I just copied them.

lkiesow: Fair enough. We can clean that out later.

darolfes:
fixed ci

gregorydlogan:
@darolfes Offhand this looks good, thanks for the hard work!  I see a number of descriptors still remaining, is this unit of work done?  Or should we wait for you to remove the rest of them?

lkiesow:
> Or should we wait for you to remove the rest of them?

Don't wait :-)

PR #1246, opened 2019-11-25
Remove default storage_id setting from asset Manager

### LINKED ISSUE #1244: Inconsistent default value for `storage_id`
@gregorydlogan, you set the `storage_id` in the ddl script to a default value which is not guaranteed if you use any other system of database set-up:

https://github.com/opencast/opencast/blob/a4019af192858a0dccda1ee89a1dae3752f632c2/docs/scripts/ddl/mysql5.sql#L382

Is this actually required? Opencast seems to work just fine without is (e.g. when running with the internal database). Can you take a look at this?

---

Same here:

https://github.com/opencast/opencast/blob/a4019af192858a0dccda1ee89a1dae3752f632c2/docs/scripts/ddl/mysql5.sql#L364

gregorydlogan:
Setting the default with JPA proved to be difficult/impossible to do in a
cross platform way, and since the AM makes assumptions regarding everything
being in local-storage by default anyway, I'm removing the default from the
DDL.

Fixes #1244

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/docs/scripts/ddl/mysql5.sql
+++ b/docs/scripts/ddl/mysql5.sql
@@ -360,7 +360,7 @@ CREATE TABLE oc_assets_snapshot (
   organization_id VARCHAR(128) NOT NULL,
   owner VARCHAR(256) NOT NULL,
   version BIGINT NOT NULL,
-  storage_id VARCHAR(256) NOT NULL DEFAULT 'local-filesystem',
+  storage_id VARCHAR(256) NOT NULL,
   --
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Open question: Should we add a migration script to unset the default?

gregorydlogan:
@lkiesow ideally this makes 8.0, so can we bump this up in priority?

gregorydlogan:
Not entirely sure why the crowdin deploy job a) is firing at all, and b) is failing, but it's nothing I've changed...

PR #1252, opened 2019-11-27
Make JPA Generated Database Match Script Generated Database

lkiesow:
This patch is the next step in making Opencast's JPA code generate a
proper and performant database schema. With this patch applied, column
sizes, names, indexes, and unique constraints should match the ones
generated by the ddl scripts.

This should already give you a well-performing database which is
auto-generated by Opencast using either MariaDB or PostgreSQL.

That said, a few foreign key constraints could still be improved, but
that is handled separately.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/persistence/UserSettingDto.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/persistence/UserSettingDto.java
@@ -29,6 +29,8 @@ import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
+import javax.persistence.Index;
+import javax.persistence.Lob;
 import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
 import javax.persistence.Table;
@@ -36,13 +38,16 @@ import javax.persistence.UniqueConstraint;
 
 /** Entity object for user settings. */
 @Entity(name = "UserSettings")
-@Table(name = "oc_user_settings", uniqueConstraints = { @UniqueConstraint(columnNames = { "id"}) })
+@Table(name = "oc_user_settings", indexes = {
+    @Index(name = "IX_oc_user_setting_organization", columnList = "organization")
+  }, uniqueConstraints = {
+    @UniqueConstraint(columnNames = { "username", "organization" }) })
 @NamedQueries({
-        @NamedQuery(name = "UserSettings.countByUserName", query = "SELECT COUNT(us) FROM UserSettings us WHERE us.username = :username AND us.organization = :org"),
-        @NamedQuery(name = "UserSettings.findByIdAndUsernameAndOrg", query = "SELECT us FROM UserSettings us WHERE us.id = :id AND us.username = :username AND us.organization = :org"),
-        @NamedQuery(name = "UserSettings.findByUserName", query = "SELECT us FROM UserSettings us WHERE us.username = :username AND us.organization = :org"),
-        @NamedQuery(name = "UserSettings.findByKey", query = "SELECT us FROM UserSettings us WHERE us.key = :key AND us.username = :username AND us.organization = :org"),
-        @NamedQuery(name = "UserSettings.clear", query = "DELETE FROM UserSettings us WHERE us.organization = :org") })
+  @NamedQuery(name = "UserSettings.countByUserName", query = "SELECT COUNT(us) FROM UserSettings us WHERE us.username = :username AND us.organization = :org"),
+  @NamedQuery(name = "UserSettings.findByIdAndUsernameAndOrg", query = "SELECT us FROM UserSettings us WHERE us.id = :id AND us.username = :username AND us.organization = :org"),
+  @NamedQuery(name = "UserSettings.findByUserName", query = "SELECT us FROM UserSettings us WHERE us.username = :username AND us.organization = :org"),
+  @NamedQuery(name = "UserSettings.findByKey", query = "SELECT us FROM UserSettings us WHERE us.key = :key AND us.username = :username AND us.organization = :org"),
+  @NamedQuery(name = "UserSettings.clear", query = "DELETE FROM UserSettings us WHERE us.organization = :org") })
 public class UserSettingDto {
   @Id
   @GeneratedValue
@@ -52,10 +57,11 @@ public class UserSettingDto {
   @Column(name = "setting_key", nullable = false)
   private String key;
 
+  @Lob
   @Column(name = "setting_value", nullable = false)
   private String value;
 
-  @Column(name = "username", nullable = false)
+  @Column(name = "username", nullable = false, length = 128)
   private String username;
 
   @Column(name = "organization", nullable = false)
--- a/modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
+++ b/modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
@@ -41,6 +41,7 @@ import javax.persistence.ElementCollection;
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.IdClass;
+import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.Lob;
 import javax.persistence.NamedQueries;
@@ -48,6 +49,7 @@ import javax.persistence.NamedQuery;
 import javax.persistence.PostLoad;
 import javax.persistence.Table;
 import javax.persistence.Transient;
+import javax.persistence.UniqueConstraint;
 
 /**
  * An in-memory construct to represent the state of a capture agent, and when it was last heard from.
@@ -96,10 +98,15 @@ public class AgentImpl implements Agent {
 
   /** The roles allowed to schedule this agent */
   @ElementCollection
-  @Column(name = "role")
-  @CollectionTable(name = "oc_capture_agent_role", joinColumns = { @JoinColumn(name = "id", referencedColumnName = "id"),
-          @JoinColumn(name = "organization", referencedColumnName = "organization") })
-  protected Set<String> schedulerRoles = new HashSet<String>();
+  @Column(name = "role", nullable = false)
+  @CollectionTable(name = "oc_capture_agent_role", indexes = {
+      @Index(name = "IX_oc_capture_agent_role_pk", columnList = "id, organization")
+    }, uniqueConstraints = {
+      @UniqueConstraint(name = "UNQ_capture_agent_role_id_org_role", columnNames = {"id", "organization", "role"})
+    }, joinColumns = {
+      @JoinColumn(name = "id", referencedColumnName = "id", nullable = false),
+      @JoinColumn(name = "organization", referencedColumnName = "organization", nullable = false) })
+  protected Set<String> schedulerRoles = new HashSet<>();
 
   /**
    * The capabilities the agent has Capabilities are the devices this agent can record from, with a friendly name
--- a/modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
+++ b/modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
@@ -50,6 +50,7 @@ import javax.persistence.Entity;
 import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
+import javax.persistence.Index;
 import javax.persistence.JoinColumn;
 import javax.persistence.Lob;
 import javax.persistence.ManyToOne;
@@ -70,7 +71,17 @@ import javax.persistence.Version;
  */
 @Entity(name = "Job")
 @Access(AccessType.FIELD)
-@Table(name = "oc_job")
+@Table(name = "oc_job", indexes = {
+    @Index(name = "IX_oc_job_parent", columnList = ("parent")),
+    @Index(name = "IX_oc_job_root", columnList = ("root")),
+    @Index(name = "IX_oc_job_creator_service", columnList = ("creator_service")),
+    @Index(name = "IX_oc_job_processor_service", columnList = ("processor_service")),
+    @Index(name = "IX_oc_job_status", columnList = ("status")),
+    @Index(name = "IX_oc_job_date_created", columnList = ("date_created")),
+    @Index(name = "IX_oc_job_date_completed", columnList = ("date_completed")),
+    @Index(name = "IX_oc_job_dispatchable", columnList = ("dispatchable")),
+    @Index(name = "IX_oc_job_operation", columnList = ("operation")),
+    @Index(name = "IX_oc_job_statistics", columnList = ("processor_service, status, queue_time, run_time")) })
 @NamedQueries({
         @NamedQuery(name = "Job", query = "SELECT j FROM Job j "
                 + "where j.status = :status and j.creatorServiceRegistration.serviceType = :serviceType "
@@ -146,7 +157,6 @@ public class JpaJob {
   @Column(name = "creator", nullable = false, length = 65535)
   private String creator;
 
-  @Lob
   @Column(name = "organization", nullable = false, length = 128)
   private String organization;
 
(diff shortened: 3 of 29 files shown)


gregorydlogan:
This looks good, does it make sense to remove the parts that *are* autogenerated from the ddl scripts, and leave just the parts that aren't?

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
```diff
@@ -96,10 +98,15 @@
 
   /** The roles allowed to schedule this agent */
   @ElementCollection
-  @Column(name = "role")
-  @CollectionTable(name = "oc_capture_agent_role", joinColumns = { @JoinColumn(name = "id", referencedColumnName = "id"),
-          @JoinColumn(name = "organization", referencedColumnName = "organization") })
-  protected Set<String> schedulerRoles = new HashSet<String>();
+  @Column(name = "role", nullable = false)
```
lwillmann, line 101: Role could be NULL in the DB script.

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
```diff
@@ -96,10 +98,15 @@
...
+  @CollectionTable(name = "oc_capture_agent_role", indexes = {
+      @Index(name = "IX_oc_capture_agent_role_pk", columnList = "id, organization")
+    }, uniqueConstraints = {
+      @UniqueConstraint(name = "UNQ_capture_agent_role_id_org_role", columnNames = {"id", "organization", "role"})
+    }, joinColumns = {
+      @JoinColumn(name = "id", referencedColumnName = "id", nullable = false),
+      @JoinColumn(name = "organization", referencedColumnName = "organization", nullable = false) })
```
lwillmann, line 108: - This generates a combined foreign key on oc_capture_agent_state:

` CONSTRAINT `FK_oc_capture_agent_role_organization` FOREIGN KEY (`organization`, `id`) REFERENCES `oc_capture_agent_state` (`organization`, `id`)`
In the script there is only a foreign key constraint to `oc_organization`. 

- The foreign key constraint from `oc_capture_agent_state` to `oc_organization` is missing

modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
```diff
@@ -52,8 +52,13 @@
  */
 @Entity(name = "ServiceRegistration")
 @Access(AccessType.FIELD)
-@Table(name = "oc_service_registration", uniqueConstraints = @UniqueConstraint(columnNames = { "host_registration",
-        "service_type" }))
+@Table(name = "oc_service_registration", indexes = {
```
lwillmann, line 55: For the fields: `error_state_trigger` and `warning_state_trigger` JPA only regenerates `INTEGER` fields where the script generates `BIGINT`

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
```diff
@@ -198,7 +208,7 @@ public Date getDateStarted() {
   @Column(name = "dispatchable")
   private boolean dispatchable;
 
-  @Column(name = "job_load")
+  @Column(name = "job_load", nullable = false)
```
lwillmann, line 211: Default value `1.0` missing

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
```diff
@@ -70,7 +71,17 @@
  */
 @Entity(name = "Job")
 @Access(AccessType.FIELD)
-@Table(name = "oc_job")
+@Table(name = "oc_job", indexes = {
```
lwillmann, line 74: In the script the field  `dispatchable` has a default of `1`

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserActionImpl.java
```diff
@@ -53,28 +54,33 @@
  */
 @Entity(name = "UserAction")
 @Access(AccessType.FIELD)
-@Table(name = "oc_user_action")
+@Table(name = "oc_user_action", indexes = {
+    @Index(name = "IX_oc_user_action_created", columnList = "created"),
+    @Index(name = "IX_oc_user_action_inpoint", columnList = "inpoint"),
+    @Index(name = "IX_oc_user_action_outpoint", columnList = "outpoint"),
+    @Index(name = "IX_oc_user_action_mediapackage_id", columnList = "mediapackage"),
```
lwillmann, line 61: The field `mediapackage` is generated as `LONGTEXT` instead of `VARCHAR(128)`
Therefore the generation of the index throws the following exception:
`Internal Exception: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: BLOB/TEXT column 'mediapackage' used in key specification without a key length
Error Code: 1170`

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
```diff
@@ -40,9 +41,11 @@
 
 @Entity(name = "UserSession")
 @Access(AccessType.FIELD)
-@Table(name = "oc_user_session")
+@Table(name = "oc_user_session", indexes = {
+    @Index(name = "IX_oc_user_session_user_id", columnList = "user_id")
```
lwillmann, line 44: The field `user_id` is generated as `LONGTEXT` instead of `VARCHAR(128)`
Therefore the generation of the index throws the following exception:
`Internal Exception: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: BLOB/TEXT column 'user_id' used in key specification without a key length Error Code: 1170`

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
```diff
@@ -54,17 +57,17 @@
...
 
   @Lob
-  @Column(name = "user_id", length = 255)
+  @Column(name = "user_id")
   @XmlElement(name = "userId")
   private String userId;
 
   @Lob
-  @Column(name = "user_ip", length = 255)
+  @Column(name = "user_ip")
```
lwillmann, line 62: The field `user_ip` is generated as `LONGTEXT` instead of `VARCHAR(255)`

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
```diff
@@ -54,17 +57,17 @@
...
   @XmlElement(name = "userIp")
   private String userIp;
 
   @Lob
-  @Column(name = "user_agent", length = 255)
+  @Column(name = "user_agent")
```
lwillmann, line 66: The field `user_agent` is generated as `LONGTEXT` instead of  `VARCHAR(255)`

modules/common-jpa-impl/src/main/java/org/opencastproject/security/impl/jpa/JpaGroup.java
```diff
@@ -81,13 +81,18 @@
   private String role;
 
   @ElementCollection
-  @CollectionTable(name = "oc_group_member", joinColumns = { @JoinColumn(name = "group_id") })
+  @CollectionTable(name = "oc_group_member", joinColumns = {
+      @JoinColumn(name = "group_id", nullable = false) })
```
lwillmann, line 85: This generates a foreign key constraint to `oc_group` which seems to be correct but is not in the db script

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/persistence/UserSettingDto.java
```diff
@@ -29,20 +29,25 @@
...
+import javax.persistence.Lob;
 import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 import javax.persistence.UniqueConstraint;
 
 /** Entity object for user settings. */
 @Entity(name = "UserSettings")
-@Table(name = "oc_user_settings", uniqueConstraints = { @UniqueConstraint(columnNames = { "id"}) })
+@Table(name = "oc_user_settings", indexes = {
```
lwillmann, line 41: I didn't find a way to generate this table.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesEntity.java
```diff
@@ -79,20 +79,21 @@
...
   protected Map<String, String> properties;
 
   @ElementCollection
-  @MapKeyColumn(name = "type")
+  @MapKeyColumn(name = "type", length = 128, nullable = false)
   @Column(name = "data")
   @CollectionTable(name = "oc_series_elements", joinColumns = {
-          @JoinColumn(name = "series", referencedColumnName = "id"),
-          @JoinColumn(name = "organization", referencedColumnName = "organization") })
+      @JoinColumn(name = "series", referencedColumnName = "id", nullable = false),
```
lwillmann, line 100: No primary key `(series, organization, type)` is generated for this table.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesEntity.java
```diff
@@ -79,20 +79,21 @@
...
+  @Lob
   @ElementCollection(targetClass = String.class)
-  @MapKeyColumn(name = "name")
-  @Column(name = "value")
+  @MapKeyColumn(name = "name", nullable = false)
+  @Column(name = "value", length = 65535)
   @CollectionTable(name = "oc_series_property", joinColumns = {
-          @JoinColumn(name = "series", referencedColumnName = "id"),
-          @JoinColumn(name = "organization", referencedColumnName = "organization") })
+          @JoinColumn(name = "series", referencedColumnName = "id", nullable = false),
```
lwillmann, line 90: No primary key `(organization, series, name)` is generated for this table.

lwillmann (commented):
I've compared the changes with the 'mysql5.sql' database creation script (btw what does the '5' stand for?). It might be a good idea to mention in the documentation that one needs a database user with admin rights to create the tables 'on the fly'. For relations between tables the 'on delete cascase' rule is not represented inside the database, but as far as I understood in the JPA code ... maybe this should be documented either. I've marked all items that are different between the script and the code.. However I feel not competent enough to say if the script or the generated code is correct. 

The creation of database tables for OAI-PMH does not work , as there is a H2 specific syntax in the column definition for `modification_date` in the class `OaiPmhEntity`. Thus I could not test the creation of tables concerning OAI-PMH related tables.

Large textfieds are often generated as `LONGTEXT` instead of `TEXT` or `MEDIUMTEXT` I guess this is ok, thus I have only commented on the column definitions where `VARCHAR` is converted to `LONGTEXT`.

I did not mention the places where foreign key constraints are missing, just the places where they are different than the script.

I did not test the generation of the tables:
```
- oc_aws_asset_mapping,

- oc_transcription_service_provider

- oc_transcription_service_job
```
because of the rather complicated setup

gregorydlogan:
@lkiesow are you wanting to address @lwillmann's comments, or do we merge this as is and address in a later PR?

gregorydlogan:
@lkiesow ping

gregorydlogan:
Dropping an on-hold on this so I don't keep checking.  Please remove the label when this is ready to go.

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
lkiesow: Changing this makes sense since this specifies “the roles allowed to schedule this agent”. The role being `null` would be an error.

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/AgentImpl.java
lkiesow: _Comment to myself: Next time, respond to reviews earlier. That makes it easier to remember what you did._ Nevertheless, the new constraint makes sense since we are joining on the id and organization.

modules/common-jpa-impl/src/main/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImpl.java
lkiesow: That's fine since it corresponds to what we are using in the code.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
lkiesow: Generating that reliably for multiple databases is unfortunately not possible but `nullable = false` plus setting a default value should be nerly identical. Fixed that.

modules/common-jpa-impl/src/main/java/org/opencastproject/job/jpa/JpaJob.java
lkiesow: Same as with `job_load`. Fixed.

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserActionImpl.java
lkiesow: Fixed.

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
lkiesow: Fixed.

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
lkiesow: Fixed.

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/persistence/UserSettingDto.java
lkiesow: That's because the code for this got removed from the admin interface. I don't think there is a way of accessing this apart maybe from the REST api. This probably should be removed…

modules/usertracking-impl/src/main/java/org/opencastproject/usertracking/impl/UserSessionImpl.java
lkiesow: Fixed

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesEntity.java
lkiesow: Unfortunately, that's not possible:

> `@ElementCollection` cannot take a primary key, because an Embeddable types cannot have an identifier. You can add an `@OrderColumn` to optimize the generates SQL statements.

But we can add a unique constraint archiving the same thing. Did that.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesEntity.java
lkiesow: Same here.

lkiesow:
@gregorydlogan, back to you.
I've addressed all of Ludwigs findings.
I've tested this with both PostgreSQL and MariaDB.

To get proper _official_ PostreSQL support, I will add some additional documentation once this is merged.

For testing, these might be handy:

- https://data.lkiesow.io/opencast/custom.properties.postgres
- https://data.lkiesow.io/opencast/postgres-docker-compose.yml
- https://data.lkiesow.io/opencast/custom.properties.mariadb
- https://data.lkiesow.io/opencast/mariadb-docker-compose.yml

gregorydlogan:
@lkiesow the intention here is that we can drop the ddls entirely, correct?  If so:

a) We need to change the docs to reflect that
b) We need to think about how to handle permissions.  In testing I couldn't figure out why things weren't working - none of the tables were being created.  Turns out the runtime `opencast` database user does not have `CREATE TABLE` perms, per the existing docs.  Offhand I'm not sure what else it will need (still working on getting it started), but IIRC there was a reason we didn't want the runtime user to have those perms...

lkiesow:
That's the longtime goal. The intention of this pull request is just to make the autogeneration match the scripts reasonably well. Dropping the scripts will need additional work and definitely more documentation.

gregorydlogan:
Looks good to me.

PR #1254, opened 2019-11-28
Map internal service host URLs to tenant-specific URLs

KatrinIhler:
Add a configuration option in the tenant configuration to map the internal host URLs to tenant-specific ones in the services/available endpoint. This is necessary when using tenant-specific capture agent
users.

_This work was sponsored by the University of Wien._


modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
```diff
@@ -192,14 +190,28 @@ public void updated(String pid, Dictionary properties) throws ConfigurationExcep
...
+        String tenantSpecificHost = StringUtils.trimToNull((String) properties.get(key));
+        serverUrls.add(tenantSpecificHost);
+      }
+
       orgProperties.put(key.substring(ORG_PROPERTY_PREFIX.length()), (String) properties.get(key));
     }
 
+    if (serverUrls.isEmpty()) {
+      logger.debug("No server configured for organization " + name + ", setting default");
+      serverUrls.add(DEFAULT_SERVER);
```
lkiesow: Just wondering what happens if no server is set?

modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
```diff
@@ -84,7 +84,10 @@
   public static final String ORG_NAME_KEY = "name";
 
   /** The managed property that specifies the organization server name */
-  public static final String ORG_SERVER_KEY = "server";
+  public static final String ORG_SERVER_KEY = "prop.org.opencastproject.host.";
```
lkiesow:
```suggestion
  public static final String ORG_SERVER_PREFIX = "prop.org.opencastproject.host.";
```

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
```diff
@@ -291,13 +301,31 @@ public Response setMaintenanceMode(@FormParam("host") String host, @FormParam("m
           @RestResponse(responseCode = SC_OK, description = "Returned the available services."),
           @RestResponse(responseCode = SC_BAD_REQUEST, description = "No service type specified, bad request.") })
   public Response getAvailableServicesAsXml(@QueryParam("serviceType") String serviceType) {
+
     if (isBlank(serviceType))
       throw new WebApplicationException(Response.status(Status.BAD_REQUEST).entity("Service type must be specified")
               .build());
+
+    Organization currentOrg = securityService.getOrganization();
+    Map<String, String> properties = currentOrg.getProperties();
```
lkiesow: No strong opinion on this:

```suggestion
    Map<String, String> properties = securityService.getOrganization().getProperties();
```

modules/serviceregistry/src/main/resources/OSGI-INF/serviceregistry-endpoint.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lkiesow: Isn't this unnecessary due to the OSGi annotations?

lkiesow (changes requested):
This generally looks good although I've added a few minor comments.
Additional to those, there is a minor merge conflict.

One thing I've wondering though is if we need to be able to configure additional organization hostnames. For example, an admin server could be available via `oc-admin.uos.de` and `oc-admin.uni-osnabrueck.de`. We could add an additional setting for that. Though I'm also not sure if we need to.

modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
KatrinIhler: Sounds good.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
KatrinIhler: Yep.

modules/serviceregistry/src/main/resources/OSGI-INF/serviceregistry-endpoint.xml
KatrinIhler: I don't really know. Most, if not all, our current files contain this, and it's also part of examples in the official OSGI docs, so I'm gonna leave it in for now.

modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
KatrinIhler: We fall back to localhost, which is what happened before (just in a different way). And there won't be any mapping, since it's not configured.

KatrinIhler:
@lkiesow PR is updated.

Re: Multiple Mappings:
I don't think we need to do that. The mapping is only so that the capture agents can talk to the services in a tenant-specific way, whether it maps to uos.de or uni-osnabrueck.de shouldn't matter in that case. And if the internal (non-tenant-specific) URL is also different, you can just configure both of them.Though I think that might lead to some interesting behavior of the OrganizationDirectoryService...

modules/serviceregistry/src/main/resources/OSGI-INF/serviceregistry-endpoint.xml
lkiesow: I actually meant this as a rhetorical question. You can use annotations or specify the OSGi components in this XML file which then needs to be referenced in the bundle plug-in configuration in the `pom.xml` where this is missing. This is also why, for example, [the hello world bundle has no OSGi XML files](https://github.com/opencast/opencast/tree/b19b51be54fee1fca70503f2cd886a0a6e87e0a7/modules/hello-world-impl/src). Using both makes no sense and will only end in confusion when the next developer tries to change e.g. bindings in here and it does not work.

modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
lkiesow: I meant, do we need to set something at all?

lkiesow:
> The mapping is only so that the capture agents can talk to the services in a tenant-specific way, […]

The mapping is for capture agents and also for identifying all URLs which belong to a specific organization, isn't it?

KatrinIhler:
> > The mapping is only so that the capture agents can talk to the services in a tenant-specific way, […]
> 
> The mapping is for capture agents and also for identifying all URLs which belong to a specific organization, isn't it?

Good point.
But then you could do something like
prop.org.opencastproject.host.oc-admin.uos.de=tenant1.oc-admin.uos.de
prop.org.opencastproject.host.oc-admin.uni-osnabrueck.de=tenant1.oc-admin.uni-osnabrueck.de

modules/serviceregistry/src/main/resources/OSGI-INF/serviceregistry-endpoint.xml
KatrinIhler: I just realized I completely misunderstood your comment. Never mind, you're right, gonna remove that.

modules/kernel/src/main/java/org/opencastproject/kernel/security/OrganizationDirectoryServiceImpl.java
KatrinIhler: I traced it back to the point where it's used, and it's falling back to the default organization there anyway, so I'm gonna remove this.

KatrinIhler:
@lkiesow Ready for review again! :)

lkiesow:
I actually meant adding more tenant URLs but since they wouldn't need any mapping that should work as well by doing something like:

     prop.org.opencastproject.host.oc-admin.uos.de=tenant1.oc-admin.uos.de
     prop.org.opencastproject.host.also-tenant1.oc-admin.uos.de=also-tenant1.oc-admin.uos.de

PR #1267, opened 2019-12-03
Log the Ids of items being indexed

smarquard:
This adds some more specific logging to make it clearer what items are being added to the index to assist in troubleshooting reindex issues.

We could implement toString() for each of these classes, but this is a simple enough change.


CGreweling:
Looks good.

PR #1272, opened 2019-12-04
Log progress of solr search reindex

smarquard:
This logs progress when reindexing the solr search index, which can be quite large and take a long time.

Similar to how the series solr index logs progress.

Also log the mediapackage ID for any item that can't be added to the index.

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
@@ -176,6 +176,27 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
   /**
    * {@inheritDoc}
    *
+   * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#countMediaPackages()
+   */
+  @Override
+  public int countMediaPackages() throws SearchServiceDatabaseException {
+    EntityManager em = emf.createEntityManager();
+    Query query = em.createNamedQuery("Search.getCount");
+    try {
+      Long total = (Long) query.getSingleResult();
+      return total.intValue();
+    } catch (Exception e) {
+      logger.error("Could not find number of mediapackages", e);
+      throw new SearchServiceDatabaseException(e);
+    } finally {
+      em.close();
+    }
+  }
+
+
+  /**
+   * {@inheritDoc}
+   *
    * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#getAllMediaPackages()
    */
   @Override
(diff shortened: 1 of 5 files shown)


smarquard:
Logs progress like:

````
2019-12-04T15:55:11,511 | INFO  | (SearchServiceImpl:513) - Starting population of search index from 62734 items in database
2019-12-04T15:57:22,382 | INFO  | (SearchServiceImpl:551) - Indexing search 100/62734 (0 percent done)
2019-12-04T15:57:45,840 | INFO  | (SearchServiceImpl:551) - Indexing search 200/62734 (0 percent done)
...
2019-12-04T15:59:32,857 | INFO  | (SearchServiceImpl:551) - Indexing search 700/62734 (1 percent done)
````

File: modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java [hunk 1]
```diff
@@ -173,6 +173,27 @@ public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws
...
+   *
+   * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#countMediaPackages()
+   */
+  @Override
+  public int countMediaPackages() throws SearchServiceDatabaseException {
+    EntityManager em = emf.createEntityManager();
+    Query query = em.createNamedQuery("Search.getCount");
+    try {
+      Long total = (Long) query.getSingleResult();
+      return total.intValue();
```
lkiesow, line 187: Why cast this to an `int`. This really shouldn't break anything given that it's unlikely that you have more than `MAX_INT` items in the database, but still, why not just stay with `long`?

lkiesow (commented):
Generally looks good and works. Just one minor question.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
smarquard: It's just copied from a similar implementation for series: https://github.com/opencast/opencast/blob/develop/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java#L535 

No objection to leaving it as long. Should I change it?

lkiesow: Let's put it this way: Do you want to quickly update this? If not, then I'll hit merge :-)

smarquard: Merge :-) Hopefully these modules that use solr indexes will get refactored anyway to use elasticsearch indexes in due course, so this code won't be around for ever.

PR #1276, opened 2019-12-06
Don't add the internal publication of the original event twice

lwillmann:
When one duplicates an event, the duplicated event contains two internal publications. One of them references the original event. (see attached screenshot)
doubleInteral
I guess this is relevant for the releases 8.x and 7.x  either

### DIFF
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
@@ -207,6 +207,8 @@ public class DuplicateEventWorkflowOperationHandler extends AbstractWorkflowOper
     for (MediaPackageElement e : mediaPackage.getElements()) {
       if (e instanceof Publication && InternalPublicationChannel.CHANNEL_ID.equals(((Publication) e).getChannel())) {
         internalPublications.add((Publication) e);
+        // Remove publications since they are cloned in another way than the other elements
+        elements.remove(e);
       }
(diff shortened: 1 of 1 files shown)


lkiesow:
If this is relevant for 7.x, can you re-direct your pull request towards `r/7.x`?

lwillmann:
@lkiesow Not sure what to with the failed Travis build

gregorydlogan:
@lwillmann you should be able to restart the failed job. I've restarted it for now, but the error looks to me like something that would happen again.

lwillmann:
@gregorydlogan thanks for the restart. It happened again as predicted. Is there something I can do to fix this?

gregorydlogan:
Take a look at the failed job, right [here](https://travis-ci.org/opencast/opencast/jobs/623089342#L241).  You can also get to it via Details on the Travis failure.

lkiesow:
and the issue basically seems to be that Travis uses the build instructions from `develop`. Likely because the patch was pushed when this pull request was still pointing at `develop`. My guess would be that you can just touch (`git commit --amend --no-edit`) and push it again to fix this. Or Greg can ignore the test ;-P

lkiesow:
Looks like this worked.

gregorydlogan:
@lwillmann You've succeeded in keeping the duplicate from keeping the original publication, but in testing against r/7.x I'm seeing the duplicate with *no* publications at all :)

lwillmann:
@gregorydlogan Hmmm ... I've tested it again and for me it seems to work. There is only one internal publication as you can see in the screenshot.
SingleInternal
External publications of a duplicated event have never been published automatically as described in the documentation of the WOH: `If the original event was already published, the duplicate won't be published right away. The user will have to publish it manually when he is done editing it.`

PR #1283, opened 2019-12-11
Fix hourly statistics export

krinnewitz:
This patch just adds a missing switch-case to avoid an exception being
thrown when exporting hourly statistics.


krinnewitz:
Note: this should also be backported to r/8.x. 

For r/8.x, the file to patch is located under `modules/statistics-export-service/src/main/java/org/opencastproject/statistics/export/impl/StatisticsExportServiceImpl.java`

wsmirnow:
Looks good for me

PR #1284, opened 2019-12-11
Refactor Metadata classes

pmiddend:
# Abstract #

This PR refactors parts of the `dublincore` module, specifically classes regarding Metadata, such as:

- `MetadataField`
- `MetadataCollection`
- `MetadataList`

The aim was to remove special cases and duplication, as well as improving code safety.

# Measures taken #

## Move JSON reading/writing to utility class  ##

Part of the reason the aforemetioned classes are so “cluttered” is because reading and writing to JSON is baked into them, using methods `toJSON` and `fromJSON`). There are multiple issues with that:

- JSON object field name constants (a la `JSON_KEY_READONLY`) are scattered over the three classes, and serialization isn’t really consistent (some `fromJSON` methods took `String`, other a deserialized `JObject`, for example).
- Changing the JSON serialization to something other than the Entwine library is really hard.
- The `Metadata*` classes are a layer above Dublincore catalogs. As such, serializing to *JSON* isn’t really an inherent property. If anything, one would have expected `toXML` and `fromXML`.

## Move all Metadata* classes to dublincore  ##

For some reason, while `MetadataCollection` is situated in the `dublincore` module, but `AbstractMetadataCollection` is located in the `index-service` module. Despite the fact that it doesn’t depend on anything inside `index-service`. We moved all metadata classes to `dublincore`.

`DublinCoreMetadataCollection` *does* depend on the `index-service` losely. But we refactored that away, see below.

## Remove MetadataCollection inheritance ##

The inheritance hierarchy currently looks like this: `interface MetadataCollection` → `abstract class AbstractMetadataCollection` → `class DublinCoreMetadataCollection`. Note that there is exactly one implementation of the abstract class and interface, save for some test classes.

We flattened all of this and left `DublinCoreMetadataCollection`. This actually improved readability and safety a bit, because, for example, we can now simply use a copy constructor to copy the collection, instead of the `getCopy` method from the interface.

Previously, `DublinCoreMetadataCollection` called functions from the index service. This was moved to a utility class inside `index-service` (inheritance wasn’t really the best way to implement this, anyway).

## Simplify MetadataList ##

Currently, `MetadataList` inherits from `Iterable<Entry<String, Tuple<String, MetadataCollection>>>`, which is a mouthful. We removed that `implements` line and added a getter for the corresponding `Map`. Also, since `Tuple` only has obscure `getA` and `getB` methods (and we’re dealing with two otherwise unnamed `Strings`), we replaced it with a small inner class.

We also removed the usage of Entwine’s `Stream` class.

## Simplify MetadataField ##

`MetadataField` is a particular beast. Here are some of its shortcomings:

- It is a generic class, but the generic parameter is very rarely used (mostly we’re dealing with `MetadataField<?>`) and leads to “unchecked call” warnings. 
- It holds its “JSON type” as a field, rather than inferring it from its `type`.
- It uses the nonstandard `Opt` for basically all of its properties, and has confusing `setFoo` methods that deal with `null`.
- Its constructor confusingly doesn’t take *all* of its fields as parameter. It leaves out the `value` field, for example, leading to code like `f = new MetadataField<>(…); f.setValue(x);`
- It has tons of `create*` methods that sometimes call themselves.
- It uses two functions per data tytpe to serialize and deserialize from and to JSON, and has huge `create*` factory methods.

This was refactored in the following way:

- We removed `Opt`, using `null` values instead.
- `MetadataField` now only has a `type`, the JSON type is inferred.
- `MetadataField` only has two constructors: one copy constructor and one constructor taking *all* parameters.
- There is just one create method left, one which creates the field from a `Map<String, String>`, which is the most important one.
- Serialization to/from JSON is now found in the `MetadataJson` utility class.

# Attribution #

This work is sponsored by the University of Vienna as part of the 2019/2020 crowdfunding campaign.


lkiesow:
@pmiddend, merging your other pull request seems to have caused some merge conflicts. Can you take a look at them?

pmiddend:
@lkiesow Thanks for the reminder, I've fixed the conflicts.

gregorydlogan:
This looks ok to me.  @KatrinIhler, do you have any issues with merging this as is?

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
```diff
@@ -195,12 +153,8 @@ public MetadataField(final MetadataField<A> other) {
    *          be none. This is also possible to use the collectionId parameter for that.
    * @param collectionID
    *          The id of the limit list of possible value that should be get through the resource endpoint.
-   * @param valueToJSON
-   *          Function to format the metadata field value to a JSON value.
-   * @param jsonToValue
-   *          Function to parse the JSON value of the metadata field.
    * @throws IllegalArgumentException
-   *           if the id, label, type, valueToJSON or/and jsonToValue parameters is/are null
+   *           if the id, label, type, valueToJSON parameters is/are null
```
KatrinIhler: "valueToJSON" can go as well, can't it?

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataField.java
```diff
@@ -606,62 +603,6 @@ public static MetadataField createMetadataField(final String inputID, final Opt<
...
-      case ORDERED_TEXT:
-      case TEXT_LONG:
-        ((MetadataField<String>)metadataField).setValue(Iterables.getLast(filteredValues), false);
-        break;
-      default:
-        throw new IllegalArgumentException("Unknown metadata type! " + metadataField.getType());
-    }
-    return metadataField;
-  }
-
```
KatrinIhler, line 1066: I think this was an alright place for it to be, and I don't appreciate a lot of utility classes because it can be hard to find the relevant functionality.

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/ConfigurableDCCatalogUIAdapter.java
```diff
@@ -116,9 +117,9 @@ public DublinCoreMetadataCollection getRawFields() {
         String defaultKey = getCollectionDefault(metadataField, listProvidersService); // check for default
 
         if (StringUtils.isNotBlank(defaultKey)) {
-          rawFields.addField(new MetadataField(metadataField), defaultKey, listProvidersService);
+          MetadataCollectionUtils.addField(rawFields, new MetadataField(metadataField), defaultKey, listProvidersService);
         } else {
-          rawFields.addEmptyField(new MetadataField(metadataField), listProvidersService);
+          MetadataCollectionUtils.addEmptyField(rawFields, new MetadataField(metadataField), listProvidersService);
```
KatrinIhler: I think this is the only change I really don't like. Why put it into a utility class instead of making it a part of the Metadata Collection where I would expect it?

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/MetadataCollectionUtils.java
```diff
@@ -0,0 +1,163 @@
...
+    if (!filteredValues.isEmpty()) {
+      setValueFromDCCatalog(filteredValues, metadataField);
+    }
+
+    metadataField.setIsTranslatable(getCollectionIsTranslatable(metadataField, listProvidersService));
+    metadataField.setCollection(getCollection(metadataField, listProvidersService));
+
+    collection.addField(metadataField);
+  }
+}
```
KatrinIhler: See above.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1138,7 +1138,7 @@ public Response getEventMetadata(@PathParam("eventId") String eventId) throws Ex
     if (getOnlySeriesWithWriteAccessEventModal()) {
       MetadataField<?> seriesField = metadataCollection.getOutputFields().get(DublinCore.PROPERTY_IS_PART_OF.getLocalName());
       Map<String, String> seriesWithWriteAccess = getSeriesService().getUserSeriesByAccess(true);
-      seriesField.setCollection(Opt.some(seriesWithWriteAccess));
+      seriesField.setCollection(seriesWithWriteAccess);
```
KatrinIhler, line 1148: No more Opt! Hallelujah!

modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/event/EventHttpServletRequest.java
```diff
@@ -512,18 +512,16 @@ protected static MetadataList deserializeMetadataList(
    *          The metadata list created from the json request to create a new event
    */
   private static void setStartDateAndTimeIfUnset(MetadataList metadataList) {
-    Opt<MetadataCollection> optCommonEventCollection = metadataList
+    final MetadataCollection optCommonEventCollection = metadataList
```
KatrinIhler: Maybe rename this since it's no longer an Opt?

modules/index-service/src/test/java/org/opencastproject/index/service/catalog/adapter/MetadataFieldTest.java
```diff
@@ -111,69 +125,173 @@ public void testSetDifferentValues()  throws Exception {
...
+            required,
+            null,
+            null,
+            MetadataField.Type.DATE,
+            null,
+            null,
+            null,
+            null,
+            null,
+            StringUtils.isNotBlank(datePattern) ? datePattern : null,
```
KatrinIhler, line 246: Can't we just handle a blank pattern gracefully within the constructor instead of checking outside of it like this?

modules/index-service/src/test/java/org/opencastproject/index/service/catalog/adapter/MetadataFieldTest.java
```diff
@@ -111,69 +125,173 @@ public void testSetDifferentValues()  throws Exception {
...
+            null,
+            null,
+            MetadataField.Type.TEXT,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null);
```
KatrinIhler, line 270: I feel like spacing it out like this when half of it is null anyway isn't really that helpful, but I don't have a hard opinion on this.

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/DublinCoreMetadataUtil.java
```diff
@@ -287,29 +287,23 @@ static void setDuration(DublinCoreCatalog dc, MetadataField<?> field, EName enam
...
+  public static Map<String, MetadataField> getDublinCoreProperties(Dictionary<String, ?> configProperties) {
 
-    Map<String,Map<String, String>> allProperties = new HashMap();
+    final Map<String,Map<String, String>> allProperties = new HashMap<>();
 
-    for (Object configObject : Collections.list(configProperties.keys())) {
-      String property = configObject.toString();
+    for (String configObject : Collections.list(configProperties.keys())) {
+      String propertyNameOpt = getDublinCorePropertyName(configObject);
+      String propertyKeyOpt = getDublinCorePropertyKey(configObject);
```
KatrinIhler: These aren't Opts anymore, so they should probably be renamed.

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/MetadataList.java
```diff
@@ -45,7 +41,25 @@ public String getValue() {
 
   }
 
-  private final Map<String, Tuple<String, MetadataCollection>> metadataList = new HashMap<>();
+  public static final class TitledCollection {
```
KatrinIhler: I'd like "TitledMetadataCollection" as a name better, since it's more accurate.

KatrinIhler (changes requested):
Sooo I finally took a look at this. Sorry for taking so long. The gist of it my review is:
I really think this is gonna be an improvement to before, and I'm really happy about a lot of the changes.
Most of my comments are minor things except for MetadataCollectionUtils, which I really don't like. We had a lot of utility classes for metadata handling before and it can be really tough to find the right functionality, which sometimes led to things being implemented twice, and so on.

Edit: I also tested this and couldn't find any buggy behavior.

pmiddend:
@KatrinIhler I tried changing the code, removing the `MetadataCollectionUtil` class, thinking to myself: "Yeah, why _did_ I change this indeed?". 
Turns out that it's really a dependency issue. `DublinCoreMetadataCollection` is in the `dublincore` module (where it _should_ be, I think). All functions in the utility class depend on types from the `index-service` (notably `ListProvidersService`), though. All except `setValueFromDCCatalog`, which is just a helper function for those other functions, though.
Now I could _maybe_ move the `DublinCoreMetadataCollection` to the `index-service`, I suppose, but I was rather delighted to have this inheritance hierarchy and POJOs in different maven modules sorted out. The other possibility would be to move the functions to the `ListProvidersService` interface, but they're really not part of the "core" functionality of a "list provider", or are they?

KatrinIhler:
@pmiddend Okay, I see the problem. I agree that this really isn't core functionality of a list provider.
I struggled with something similar recently and decided to move some functionality out of the index service module. But since this isn't finished yet, there isn't a pull request for this. Maybe this would be the way to go here, to have a third module that both the index service and the dublin core module could depend on?

pmiddend:
And this module would contain just the utility class for now? It seems a _little_ cleaner, but it's on the verge of being overkill, at least for this PR, don't you think?

KatrinIhler:
@pmiddend No, I meant dissolving the utility class / put the stuff back where they came from and then pull the classes from the index-service module that are causing the dependency issues. 
I think this is really the crux of the issue: The index service module contains a lot of stuff the index service itself depends on, but other classes depend on them as well (see DublinCoreMetadataCollection), so they just get put in there even though that's not where they belong.

pmiddend:
@KatrinIhler Okay, this is tricky and I want us both to be on the same page. So let's talk about what we have and how we connect it:

- `MetadataField`: A property with a type, a name, and an optional list provider, though represented just as a string here
- `DublinCoreMetadataCollection`: Basically a “dumb” container for `MetadataFields` with getters/setters
- `ListProvidersService`: Something that resolves a list provider name to a list

So at its core, `ListProviders` and `MetadataFields` are connected. Ideally, we’d might even a *reference* to a list provider inside `MetadataField`, not just the name, though I’m not sure if that causes problems.

`ListProvidersService` isn’t directly connected to the index service (it doesn’t import anything from it. What we could do (and what I think you suggest): 

- Move at least the interface `ListProvidersService` to a module separate from both `index-service` and `dublincore` (the `Impl` doesn’t seem to depend on `index-service` as well, so we can move that, too). 
- Let `index-service` and `dublincore` depend on this new module and move the list provider utility functions into `MetadataField` (and remove the utility class, of course).

This keeps the inheritance hierarchy flat (which I like), removes the utility class (which you like) and shrinks the `index-service` module a bit (which everyone likes).

Objections?

KatrinIhler:
@pmiddend YES, this is exactly what I mean. I'd move both the ListProvidersService and its Impl.

pmiddend:
@KatrinIhler I did what we discussed and: 1. moved some classes and interfaces to a new module, 2. removed the utility class. This worked pretty well, and I included some of the `Impl` classes in the new module.

gregorydlogan:
@KatrinIhler Any progress here?

KatrinIhler:
@gregorydlogan Same as the other one - Corona happened and I completely forgot about this. ;) It's on my to-do-list for this month.

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/events/CommonEventCatalogUIAdapter.java
```diff
@@ -79,12 +78,12 @@ public Catalog storeFields(MediaPackage mediaPackage, MetadataCollection abstrac
...
-    if (series.getValue().isSome() && series.isUpdated()) {
-      if (isNotBlank(series.getValue().get().toString())) {
-        mediaPackage.setSeries(series.getValue().get().toString());
+    MetadataField series = abstractMetadata.getOutputFields().get(DublinCore.PROPERTY_IS_PART_OF.getLocalName());
+    if (series.getValue() != null && series.isUpdated()) {
+      if (isNotBlank(series.getValue().toString())) {
+        mediaPackage.setSeries(series.getValue().toString());
         final Opt<String> seriesTitle = getSeriesTitle(series);
-        if (seriesTitle.isSome())
+        if (seriesTitle != null)
```
KatrinIhler, line 86: Potential bug? This doesn't handle the case that Opt is none!
Maybe it would be better to throw the Opt out of that method entirely.

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreMetadataCollection.java
```diff
@@ -0,0 +1,280 @@
...
+
+  public void updateStringField(final MetadataField current, final String value) {
+    if (current.getValue() != null && !(current.getValue() instanceof String)) {
+      throw new IllegalArgumentException("Unable to update a field to a different type than String with this method!");
+    }
+    removeField(current);
+    final MetadataField field = new MetadataField(current);
+    field.setValue(value);
+    addField(field);
+  }
```
KatrinIhler, line 269: This looks weird. Why do we need to do this at all? Maybe this could be removed at a later date.

KatrinIhler (changes requested):
Alright, I went through all the changes again in the last couple of days (because I couldn't remember a lot from my first review), and all in all, this is looking pretty good. My tests worked fine, too. I laid out some thoughts below, but the **TLDR** is:

IMO there's a potential bug in CommonEventCatalogUIAdapter (see comment below). That should probably be fixed. Everything else is nitpicking that you can just ignore, I just wanted somewhere to put it. ;)

#### Some thoughts on style
- I don't like spreading out the method calls with one parameter per line, but I guess that's a question of style.
- I'm not quite sure if I like MetadataJSON - on the one hand it's one of those pesky utility classes that can get confusing, but on the other hand MetadataField is now quite a lot shorter - I guess you just can't win here.

#### For the future
- I was considering if it makes sense to extend MetadataField for the different subtypes - this way the constructor calls could be a bit shorter.
- Can't we move the specific list providers from the index module into the new list providers service module (or another one)? Don't see why they should be part of the index service, but they might be woven in more tightly than I expect.

lkiesow:
This has been replaced by #1857

PR #1287, opened 2019-12-11
Add option to remove running workflows, fix restdocs for delete requests

KatrinIhler:
This PR does two things:

- Send parameter as query for delete requests in rest docs
    
    Form parameters for DELETE requests are currently not used and are not even
    possible because we can't set them in the remote implementations in the
    backend. However the rest docs currently use form parameters for everything
    but GET, resulting in some endpoints with query parameters not working
    correctly. This commit makes the rest docs send query parameters for
    DELETE requests.

- Add a force parameter to remove running workflows
    
    If a workflow gets stuck, the affected event can't be deleted. In that
    case, it can make sense to delete a workflow even though it's still
    running.

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowService.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowService.java
@@ -233,6 +233,25 @@ public interface WorkflowService {
           UnauthorizedException, WorkflowStateException;
 
   /**
+   * Permanently removes a workflow instance. Option to remove a workflow instance regardless of its status.
+   *
+   * @param workflowInstanceId
+   *          the workflow instance identifier
+   * @param force
+   *          remove the workflow instance no matter the status
+   * @throws WorkflowDatabaseException
+   *           if there is a problem writing to the database
+   * @throws NotFoundException
+   *           if no workflow instance with the given identifier could be found
+   * @throws UnauthorizedException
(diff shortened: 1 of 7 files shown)


modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowService.java
```diff
@@ -229,7 +231,7 @@ WorkflowInstance start(WorkflowDefinition workflowDefinition, MediaPackage media
    * @throws WorkflowStateException
    *           if the workflow instance is in a disallowed state
    */
-  void remove(long workflowInstanceId) throws WorkflowDatabaseException, WorkflowParsingException, NotFoundException,
+  void remove(long workflowInstanceId, boolean force) throws WorkflowDatabaseException, WorkflowParsingException, NotFoundException,
```
lkiesow: Instead of having to define the default (`false`) over and over again, would it make sense to have ` void remove(long workflowInstanceId)` and ` void remove(long workflowInstanceId, boolean force)` instead, with the former defined as

```java
 void remove(long workflowInstanceId … {
  remove(workflowInstanceId, false);
}
```

KatrinIhler: You're right, done! I also extended one test a bit.

lkiesow (approved):

PR #1288, opened 2019-12-12
Java 11 Compatibility

krinnewitz:
This pull request consists of two commits. The first one externalizes and updates Elasticsearch. The second one makes Opencast compatible with JDK11. This work is part of the 2019 Crowd Funding Campaign. Special thanks go to Harvard DCE for funding this goal.

## Externalize & update Elasticsearch

This patch upgrades to the Elasticsearch High Level REST Client 7.4
which needs an Elasticsearch 7 node to work with.
Since embedding Elasticsearch is no longer supported, the embedded node
is removed. Elasticsearch 7 does no longer allow to store different
document types in the same index. Instead, one index per document type
is required. So the code was changed to create four "sub indices", one
for each document type. This means, that for the admin UI index, there
are actually 4 Elasticsearch indices now. The same applies for the
external API index.

So from now on, running an external Elasticsearch node is mandatory in
order to be able to run Opencast.

One small test case was removed as it would now require a running
external Elasticsearch node during execution time.

## Make Opencast compatible with JDK11

This patch improves compatibility with newer JDKs. Namely, it is now
possible to build and run Opencast with JDK11. For JDK12 and JDK13,
building Opencast is possible as well, while running Opencast is not
possible with these JDKs, yet.

Since we want to stay compatible with JDK1.8 for now, language features
of newer JDKs cannot be used at the moment. Once we decide to drop
JDK1.8 support, new language features can be used without any further
preparations, except increasing the `jdk.version` in the main POM.

The following table shows the current state of Opencast compatibility
with some JDKs.

| JDK  | Build | Run | Language Features |
|------|-------|-----|-------------------|
|JDK1.8|  YES  | YES |        YES        |
|JDK11 |  YES  | YES |        NO         |
|JDK13 |  YES  | NO  |        NO         |

Compatibility with JDK 9, 10, and 12 was not tested, but for JDK 9 and
10, full (build _and_ runtime) compatibility is much likely given.

To also achieve runtime compatibility with JDK12 and JDK13, Karaf 4.3
is needed, which is not yet released. Once the new Karaf version is
available, enabling JDK12 and JDK13 runtime compatibility should be
fairly easy.

In order to achieve runtime compatibility with JDK11, a little quirk was
necessary: `jakarta.xml.bind-api-2.3.3.jar` had to be packaged with
Opencast as this version is required for JDK11. Version 2.3.3 of this
library is already released, but not yet uploaded to Maven Central
because it is part of the upcoming Java EE release, which was not yet
published. Once this is published, the packaged jar can most likely be
replaced with a Maven reference. Once Karaf comes with a Java EE version
including version 2.3.3 of jakarta.xml.bind-api or higher, it should no
longer be necessary to manually manage this dependency at all.

To enable compatibility with newer Java versions, lots of dependencies
were updated. Also, Java EE is currently in the process of renaming all
artifacts named `javax.something` to `jakarta.something-else` due to
legal issues with the term `javax`. Hence, a lot of pom files had to be
touched as upgrading a dependency also meant to rename it in many cases.

After upgrading all the dependencies, a few issues came up which are
worth being mentioned:

- JPA was complaining about duplicate column names in some queries. So I
  had to extend 3 `@JoinColumn` annotations.
- With Java 9, breaking changes to locales were made which caused some
  date formatting code parts to break. This was fixed by setting the
  JDK's `java.locale.providers` property to `COMPAT`.
- The `Lazy` class caused strange behavior where class files of classes
  using `Lazy` were compiled differently by different JDKs, even though
  both JDKs were compiling for the same byte version. This caused test
  failures. Since `Lazy` could be easily replaced and seemed to not
  provide any advantages, it was removed.
- Since Jetty had to be updated from version 1.x to version 2.x, some
  major changes to the `rest-test-environment` bundle were necessary.

The file `system.properties.append` used to configure XML processing
related factories to use internal JDK classes. This was changed to use
JDK-independent classes to improve compatibility with future JDKs.

Upgrading the ActiveMQ client library was not necessary, but was done
anyway.

Since a lot of dependencies were changed, the Maven part of `NOTICES`
was re-generated.


assemblies/pom.xml
```diff
@@ -151,6 +148,8 @@
                   <!-- Special configuration for development -->
                   <replaceregexp file="target/classes/package.xml" match="tar\.gz" replace="dir" byline="true"/>
                   <replaceregexp file="target/classes/package.xml" match="baseDirectory&gt;.*&lt;/baseDirectory&gt;" replace="includeBaseDirectory&gt;false&lt;/includeBaseDirectory&gt;" byline="true"/>
+                  <!-- To support java11+, we need at least jakarta.xml.bind-api:2.3.3. So we delete the older version to avoid conflicts. Once karaf comes with version 2.3.3+, this is no longer necessary. -->
+                  <delete file="target/assembly/lib/jdk9plus/jakarta.xml.bind-api-2.3.2.jar" quiet="true" failonerror="false"/>
```
lkiesow, line 152: Just wondering where this file comes from since this line seems to be very likely to break in the future.

docs/guides/admin/docs/upgrade.md
```diff
@@ -9,14 +10,73 @@ Configuration Changes
...
+
+Install and configure a standalone Elasticsearch node
+-----------------------------------------------------
+
+In the past, Opencast came with its own integrated Elasticsearch node. However, recent versions of Elasticsearch no longer
+support to be embedded in applications. Since the Elasticsearch client was updated to version 7, Opencast now requires an
+external Elasticsearch node of the same version to be present. This means, that all Opencast adopters now have to run
+Elasticsearch.
+
+Please check [the documentation](https://docs.opencast.org/develop/admin/#modules/searchindex/elasticsearch/)
```
lkiesow, line 41: Don't use a link to `develop`. Link the markdown file instead:
```suggestion
Please check [the documentation](modules/searchindex/elasticsearch.md)
```

docs/guides/admin/docs/upgrade.md
```diff
@@ -9,14 +10,73 @@ Configuration Changes
...
 --------------
-
 1. Stop your current Opencast instance
 2. Replace Opencast with the new version
 3. Back-up Opencast files and database (optional)
 4. Upgrade the database
-5. Remove search index data folder
-6. Start Opencast
-7. Rebuild Admin UI and External API index
+5. [Install and configure a standlone Elasticsearch node](#install-and-confifure-a-standalone-elasticsearch-node)
```
lkiesow, line 27: This is not only necessary for upgrades but also for every installation and should hence be part of the installation guides and/or the [basic configuration documentation](http://docs.opencast.org/develop/admin/#configuration/basic/)

lkiesow (changes requested):
I didn't go through all the code so if you did hide something funny in here… you likely get it in ;-P

Overall, this works as expected. I've built this with Java 11 and deployed this with Java 11. I need to do an additional test with Java 8, but not today and maybe I'm asking someone else to do that so we have two people running a black-box test against this.

The only thing I've really found was missing is the documentation for new users which tells them that they need to install Elasticsearch which should be easy enough to add.

lkiesow:
Forgot to mention that I've also not yet tested this merged into the current HEAD of `develop´. Will do soon.

rrolf:
Ups, I did not mean to merge

krinnewitz:
Should I create a follow-up PR which adds the changes suggested by @lkiesow ?

assemblies/pom.xml
krinnewitz: This file is downloaded by the karaf assembly maven plugin. When we update karaf, we should also update the assembly plugin. This lib will then most likely be updated as well.

lkiesow:
Since I had no major complaints, let's just do that and spare us any reverts, …
So, what's missing is basically just a documentation update.

PR #1289, opened 2019-12-12
Improve LTI: add create event and edit event, improve series tool

pmiddend:
# Abstract #

This PR rewrites the whole LTI interface from mustache.js to ReactJS (using Typescript as the language of choice, instead of Javascript). It also embiggens LTI by adding/changing the following:

- Uploading of new events (as a new tool)
- Editing the metadata of existing events from the `series` tool; this is optional and can be enabled via a tool parameter
- Deletion of single events from the `series` tool; this is optional and can be enabled via a tool parameter.
- Copying an event from one series to another (this is a special wish from our customer, University of Gent).

## Uploading/Editing ##

The upload tool currently doesn’t really compare to the “New Event” modal in the admin UI. It was designer for a specific use case and is kept simple. It presents the user merely with the *event title*, a *file input* and some assorted metadata fields: presenter(s), license and language. Internally, a workflow can be configured which is executed after ingestion. Uploading a captions file is supported as well.

The upload tool also contains a list of running jobs, so the user gets some feedback about the upload completion and isn’t put off by the video not instantly appearing in the series tool.

The edit tool is almost the same, but it doesn’t have any upload features. It also has a “copy to different series” dropdown which triggers a new workflow (see below).

## Deletion ##

Events can be deleted by pressing the trashcan symbol in the series list. To make this as simple as possible, this PR uses the “single step event deletion”, so the user doesn’t have to unpublish, wait, and then delete.

# The Code #

## New modules ##

The `lti` module in Opencast is very minimalistic (which was sufficient for now). To support uploading and deletion, we had to write some Java code which needed services that are not necessarily available on the node LTI runs on. Our customer would like to run LTI on the presentation node, for example. To add multi-node support, we added the following modules and functions:

- `lti-service-api` contains the `LtiService` interface, implemented by the “remote” and the “impl”
- `lti-service-impl` contains the implementation
- `lti-service-remote` contains the remote
- `lti` contains an additional endpoint which holds an `LtiService` reference and serves as a “trampoline” to either the remote or the impl.

While reviewing, do note that the endpoint in `lti` looks strikingly similar to the remote endpoint. That’s, unfortunately, unavoidable.

## The “copy to series” workflow ##

To implement the “copy to different series” functionality, we did three two things:

- Add a new workflow “copy-event-to-series”, which just triggers the duplicate-WOH
- Modify the “duplicate” workflow operation handler so it can set a new series while copying the event.
- Listen for the copy workflow’s completion, in order to immediately trigger the publish workflow on the new event.

## Translating LTI ##

Since we’re retrieving the event’s metadata list, which includes translation keys from the Admin UI, we though it’d be wise to simply copy the Admin UI’s translation files and add a new “LTI” section specifically for LTI translation keys. This, of course, means that LTI is sort of dependent on the Admin UI, but the copying seemed more “robust” than copy&pasting translations.

# Attribution #

This work was sponsored by Uni Gent.


smarquard:
Bonus points for using the rarely used but fabulous word "embiggens" 👍 💯

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -245,7 +254,7 @@ public Response getEpisodeAndSeriesById(
               name = "sign", type = RestParameter.Type.BOOLEAN)
   }, reponses = { @RestResponse(description = "The request was processed successfully.", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "The search results, formatted as xml or json.")
   public Response getEpisode(@QueryParam("id") String id, @QueryParam("q") String text,
-          @QueryParam("sid") String seriesId, @QueryParam("sort") String sort, @QueryParam("tag") String[] tags, @QueryParam("flavor") String[] flavors,
+          @QueryParam("sid") String seriesId, @QueryParam("sname") String seriesName, @QueryParam("sort") String sort, @QueryParam("tag") String[] tags, @QueryParam("flavor") String[] flavors,
           @QueryParam("limit") int limit, @QueryParam("offset") int offset, @QueryParam("admin") boolean admin,
```
karendolan, line 258: That new sname param should to be added to the REST docs annotation describing the params for this method. It would be helpful to clarify in the REST doc annotation description that the used of "sname" param here assumes a unique series titles. i.e. not something like "Intro to Science 101" duplicated across different terms (like our site does).

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -261,6 +270,36 @@ public Response getEpisode(@QueryParam("id") String id, @QueryParam("q") String
...
+    if (seriesName != null && seriesId != null) {
+      return Response.status(Response.Status.BAD_REQUEST).entity("invalid request, both 'sid' and 'sname' specified")
+              .build();
+    }
+
+    if (seriesName != null) {
+      DublinCoreCatalogList result;
+      try {
+        result = seriesService.getSeries(new SeriesQuery().setSeriesTitle(seriesName));
+      } catch (SeriesException e) {
```
karendolan, line 286: In the past, we found it dangerous to tie the Search Service REST endpoint, that usually deploys to engage, to a service that usually lives on the Admin node. It caused the Engage player to be tightly dependent on the Admin node being up and accessible from the Engage node.

modules/search-service-impl/src/main/resources/OSGI-INF/search-service-rest.xml
```diff
@@ -15,4 +15,6 @@
              cardinality="1..1" policy="static" bind="setSearchService"/>
   <reference name="serviceregistry" interface="org.opencastproject.serviceregistry.api.ServiceRegistry"
              cardinality="1..1" policy="static" bind="setServiceRegistry"/>
+  <reference name="seriesservice" interface="org.opencastproject.series.api.SeriesService"
+             bind="setSeriesService"/>
```
karendolan: The REST endpoint assumes the series service exists and that this cardinality is 1..1. IMHO, it's human friendly to clarify that requirement in this doc.

HOWEVER, as I mentioned in early comment, tying the Engage Search service to the Admin's Series service, for sites that use the SeriesServiceRemote on Engage, can/will cause service interruptions for students watching lectures that was not expected. It would be more site friendly to make this dependency dynamic and optional and change the REST code to return an unimplemented 501, or similar if the sname is passed in. 

NOTE that the seriesId param uses the Search Service series path. If a series has no events published to engage, the seriesId query returns the no-results response. It's odd for the sname to return different results than seriesId. The reason seriesId uses the Search Service is to prevent the possible binding of the endpoint  on Engage (that students require to access and watch published lectures) to the Admin node's Series service impl.

Alternatively, the documentation for this can strongly recommend using external indices so no remote services are required (the service impls deployed on all remote nodes using the common external index and common external database).

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -155,6 +205,9 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
     final String configuredSourceFlavors = trimToEmpty(operation.getConfiguration(SOURCE_FLAVORS_PROPERTY));
     final String configuredSourceTags = trimToEmpty(operation.getConfiguration(SOURCE_TAGS_PROPERTY));
     final String configuredTargetTags = trimToEmpty(operation.getConfiguration(TARGET_TAGS_PROPERTY));
+    final boolean noSuffix = Boolean.parseBoolean(trimToEmpty(operation.getConfiguration(NO_SUFFIX)));
+    final String seriesIdStr = operation.getConfiguration(SET_SERIES_ID);
+    final String seriesId = seriesIdStr == null || seriesIdStr.startsWith("$") ? "" : seriesIdStr;
```
karendolan: I'm curious why the set-series-id value in the operation configuration might start with a "$"? It seems like a really odd edge case.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -229,11 +301,21 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+        }
         // Clone the media package (without its elements)
-        newMp = copyMediaPackage(mediaPackage, i + 1, copyNumberPrefix);
+        newMp = copyMediaPackage(mediaPackage, series, newMpId, noSuffix, i + 1, copyNumberPrefix);
+
+        if (series != null) {
+          newMp = ingestService
+                  .addCatalog(new ByteArrayInputStream(series.dc.toXmlString().getBytes(StandardCharsets.UTF_8)),
+                          UUID.randomUUID().toString() + ".xml", MediaPackageElements.SERIES, newMp);
+        }
```
karendolan, line 345: How is the new series ACL associated to this mediapackage?

karendolan:
HI @pmiddend, what is the reason to keep the lti module intact and not to roll it into lti-service-api and lti-service-impl? Is the purpose of the lti-service modules to act as mediators between the lti module and Opencast services (i.e. job creation)? If so, would it have been possible for the lti module to have used the external API for executing jobs and services from Opencast?

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
pmiddend: Agreed, fixed that one.

modules/search-service-impl/src/main/resources/OSGI-INF/search-service-rest.xml
pmiddend: First of all, thanks for this feedback "from the trenches". I hadn't thought of the problems of tying the two services together.

First, regarding the response in case the series doesn't exist. You're right, with a series ID instead of a name, this gives an empty result. I have now adapted the function to do that.

Second, regarding the dynamic dependency: I have now declared the series service `dynamic` and of cardinality `0..1`. Also, the method returns 501 as you suggested (I thought about 503 instead, but couldn't make up my mind), so you don't pay a price if you're not using the feature. Since I'm on OSGi god, I'm hoping this is enough. The code seems to be working.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
pmiddend: Good catch. I remember this code making sense when I set up the new WOH, but I've now tested it without this code and it still works (I also removed parameters from the `duplicate-event` workflow).

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
pmiddend: It...isn't. :D 
Again, good catch. I added code to copy the series ACL now.

pmiddend:
> HI @pmiddend, what is the reason to keep the lti module intact and not to roll it into lti-service-api and lti-service-impl? Is the purpose of the lti-service modules to act as mediators between the lti module and Opencast services (i.e. job creation)? If so, would it have been possible for the lti module to have used the external API for executing jobs and services from Opencast?

With regards to the LTI Maven modules, you’re right, `lti-service-api`, `lti-service-remote` and `lti-service-impl` mediate between the LTI module and the rest of Opencast.

We could have merged `lti-service-api` and `lti`, I think. Though I personally like the distinction between an LTI GUI module and a module specifying its abstract interface (similar to the current Admin UI separation PR).

I gave the External API some thought and in retrospect, I think we might have used it to build the LTI interface. However, we would have to have adapted it for the series names, and I’m not sure how the External API authentication plays along with LTI’s own mechanisms.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
```diff
@@ -0,0 +1,224 @@
...
+          return Response.ok().build();
+        }
+      }
+      if (eventId == null) {
+        return Response.status(Status.BAD_REQUEST).entity("No file given").build();
+      }
+      service.upsertEvent(null, captions, eventId, seriesId, metadataJson);
+      return Response.ok().build();
+    } catch (FileUploadException | IOException e) {
+      return Response.status(Status.INTERNAL_SERVER_ERROR).entity("error while uploading").build();
```
CGreweling, line 144: Uppercase
Error

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -232,6 +243,7 @@ public Response getEpisodeAndSeriesById(
           @RestParameter(description = "The ID of the single episode to be returned, if it exists.", isRequired = false, name = "id", type = RestParameter.Type.STRING),
           @RestParameter(description = "Any episode that matches this free-text query.", isRequired = false, name = "q", type = RestParameter.Type.STRING),
           @RestParameter(description = "Any episode that belongs to specified series id.", isRequired = false, name = "sid", type = RestParameter.Type.STRING),
+          @RestParameter(description = "Any episode that belongs to specified series name (note that the specified series name must be unique).", isRequired = false, name = "sname", type = RestParameter.Type.STRING),
           // @RestParameter(defaultValue = "false", description =
           // "Whether to include this series episodes. This can be used in combination with \"id\" or \"q\".",
           // isRequired = false, name = "episodes", type = RestParameter.Type.STRING),
```
CGreweling: Why is this uncommented? Maybe delete these 3 lines?

modules/lti/src/main/java/org/opencastproject/lti/LtiServlet.java
```diff
@@ -200,9 +202,21 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws S
     // We must return a 200 for some OAuth client libraries to accept this as a valid response
 
     // The URL of the LTI tool. If no specific tool is passed we use the test tool
-    UriBuilder builder = null;
+    UriBuilder builder;
     try {
-      URI toolUri = new URI(StringUtils.trimToEmpty(req.getParameter(LTI_CUSTOM_TOOL)));
+      logger.info(
+              "LTIPREFIX: Custom tool, raw parameter, before rewrite: {}",
+              StringUtils.trimToEmpty(req.getParameter(LTI_CUSTOM_TOOL)));
```
CGreweling: This should be logger.debug i think. Otherwise the logs would be full of every Userrequest through LTI.

CGreweling:
Good work. Nice to see the LTI-Tool with these features.

Some small hints.
Is it possible to configure the upload tracks. By default it is here in this tool presenter and caption.
The most common case i have seen is presenter only or presenter +  presentation.

If you edit an Event the confirm-button should be save, or confirm. Not edit.

CGreweling (commented):
Good work. Nice to see the LTI-Tool with these features.

Some small hints.
Is it possible to configure the upload tracks. By default it is here in this tool presenter and caption.
The most common case i have seen is presenter only or presenter + presentation.

If you edit an Event the confirm-button should be save, or confirm. Not edit.

pmiddend:
@CGreweling Thanks for your suggestions. I have changed the code accordingly.

As to the the upload tracks, I realize this PR is pretty customer specific. You can only specify the presenter currently, not the presentation track. Ideally, you would have a kind of copy of the "New event" modal in LTI, with the possibility to hide certain fields.
We tried to make it easy to extend the LTI interface, but sadly, I cannot estimate how long it would take to add the ability to add additional tracks, and plapadoo doesn't really have the time to do extensive work on this PR anymore. :(

karendolan:
@pmiddend It's unintuitive to leave the "lti" module name as-is, while adding the lti-api, lti-impl, lti-remove module names. Does it reflect the "lti" module's new role if it was named "lti-endpoint"? (or, ugg,  lti-external :p)

[EDIT - just saw that the names are distinguished from lti as lti-service-XYZ!]  lti-service-api, lti-service-impl, etc.

CGreweling:
@pmiddend  Okay. I am fine with it as it is now.
One last question, is the LTI-Module checking for the appropiate LTI USER, like 838823_Instructor.
If you want to change, edit, view an Event. 
Or is always the standard User "OAUTH_USER"?
Sorry i was not able to test it with different users.

pmiddend:
@CGreweling I've tested this quite often with multiple users and it "just works" apparently.
@karendolan Technically, I guess `lti` could be called `lti-frontend`. Although it does contain Java code, too, so I'm not terribly eager to change the name.

karendolan:
> @CGreweling I've tested this quite often with multiple users and it "just works" apparently.
> @karendolan Technically, I guess `lti` could be called `lti-frontend`. Although it does contain Java code, too, so I'm not terribly eager to change the name.

I am wondering what the technical role of the lti module is at this point. If it is the external API to the internal LTI service.

CGreweling (approved):
I am fine with this pr.

gregorydlogan:
@pmiddend conflicts

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -15,6 +15,49 @@
...
+     "EDIT_TITLE": "Edit event",
+     "NO_OPTION_SELECTED": "No option selected",
+     "EDITING": "Saving…",
+     "SELECT_OPTION": "Select option",
+     "JOB_TITLE": "Title",
+     "JOB_STATUS": "Status",
+     "UPLOAD_SUCCESS": "Upload succeded!",
+     "UPLOAD_FAILURE": "Upload failed!",
+     "UPLOAD_FAILURE_DESCRIPTION": "Please try again later.",
+     "EDIT_SUCCESS": "Edit succeeded, republishing begins!",
```
gregorydlogan: Edit succeeded, beginning republishing.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -15,6 +15,49 @@
    "SELECT_NO_OPTION_SELECTED": "No option selected",
    "SELECT_NO_OPTIONS_AVAILABLE": "No options available",
    "YES":          "Yes",
+   "LTI": {
```
gregorydlogan, line 19: I *think* the rest of these translations avoided exclamation points, but I'm not sure.  If so, please remove them from the new keys.

etc/workflows/copy-to-event-to-series.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>copy-event-to-series</id>
```
lkiesow: Filename should probably not be `etc/workflows/copy-to-event-to-series.xml ` but `etc/workflows/copy-event-to-series.xml `

modules/lti-service-impl/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+<!-- Gives weird error in opencast-search
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+      </plugin>
+-->
```
lkiesow: You can exclude dependencies but this shouldn't just stay commented out in here.

modules/lti-service-api/src/main/java/org/opencastproject/lti/service/api/LtiService.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Upload a new event or update existing event's metadata
+   * @param file File to upload
+   * @param captions Subtitles file
+   * @param eventId ID of the event (can be <code>null</code> for new events)
+   * @param seriesId ID of the series
+   * @param metadataJson Metadata for the event as JSON string
+   */
+  void upsertEvent(
```
lkiesow, line 49: Typo? `uploadEvent`?

modules/lti-service-impl/src/main/resources/OSGI-INF/lti-service-impl.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lkiesow, line 1: Can we use OSGi annotations instead?

modules/lti-service-remote/src/main/resources/OSGI-INF/lti-service-remote-impl.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lkiesow, line 1: same here

modules/lti/public/index.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      name="description"
+      content="Web site created using create-react-app"
```
lkiesow:
```suggestion
      content="Website created using create-react-app"
```

lkiesow (commented):
Just a few things I found when scrolling through the code.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+          final String captions,
+          final String eventId,
+          final String seriesId,
+          final String metadataJson) throws UnauthorizedException, NotFoundException {
+    if (eventId != null) {
+      updateEvent(eventId, metadataJson);
+      return;
+    }
+    if (workflow == null || workflowConfiguration == null) {
+      throw new RuntimeException("no workflow configured, cannot upload");
```
gregorydlogan: Capitalization here.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+      if (mp == null) {
+        throw new RuntimeException("Unable to create media package for event");
+      }
+      if (captions != null) {
+        final MediaPackageElementFlavor captionsFlavor = new MediaPackageElementFlavor("vtt+en", "captions");
+        final MediaPackageElementBuilder elementBuilder = MediaPackageElementBuilderFactory.newInstance().newElementBuilder();
+        final MediaPackageElement captionsMpe = elementBuilder
+                .newElement(MediaPackageElement.Type.Attachment, captionsFlavor);
+        captionsMpe.setMimeType(mimeType("text", "vtt"));
+        captionsMpe.addTag("lang:en");
```
gregorydlogan, line 248: Are these reasonable assumptions?

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+      replaceField(collection, "isPartOf", seriesId);
+      adapter.storeFields(mp, collection);
+
+      final AccessControlList accessControlList = new AccessControlList(
+              new AccessControlEntry("ROLE_ADMIN", "write", true),
+              new AccessControlEntry("ROLE_ADMIN", "read", true),
+              new AccessControlEntry("ROLE_OAUTH_USER", "write", true),
+              new AccessControlEntry("ROLE_OAUTH_USER", "read", true));
+      this.authorizationService.setAcl(mp, AclScope.Episode, accessControlList);
+      mp = ingestService.addTrack(file.getStream(), file.getSourceName(), MediaPackageElements.PRESENTER_SOURCE, mp);
```
gregorydlogan, line 275: This looks to me like you're adding the uploaded track as `presenter/source`, with a hardcoded ACL.  Both of those seem like things which ought to come in via parameters?  Or am I completely out to lunch?

lkiesow:
Merged into the current `develop`, I get a build failure:

```
[INFO] Building Opencast :: lti-service-impl 9-SNAPSHOT               [129/226]                                                                                                                                                               
[INFO] -------------------------------[ bundle ]-------------------------------                                                                                                                                                               
[WARNING] The POM for org.opencastproject:opencast-asset-manager-util:jar:9-SNAPSHOT is missing, no dependency information available                                                                                                          
[INFO] ------------------------------------------------------------------------                                                                                                                                                               
[INFO] Reactor Summary: 
```

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+  public void copyEventToSeries(final String eventId, final String seriesId) {
+    final String workflowId = COPY_EVENT_TO_SERIES_WORKFLOW;
+    try {
+      final WorkflowDefinition wfd = workflowService.getWorkflowDefinitionById(workflowId);
+      final Workflows workflows = new Workflows(assetManager, workflowService);
+      final ConfiguredWorkflow workflow = workflow(wfd, createCopyWorkflowConfig(seriesId, UUID.randomUUID().toString()));
+      final List<WorkflowInstance> workflowInstances = workflows
+              .applyWorkflowToLatestVersion(Collections.singleton(eventId), workflow).toList();
+      if (workflowInstances.isEmpty()) {
+        throw new RuntimeException(String.format("couldn't start workflow '%s' for event %s", workflowId, eventId));
```
gregorydlogan: Caps

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+      }
+    } catch (WorkflowDatabaseException | NotFoundException e) {
+      logger.error("unable to get workflow definition {}", workflowId, e);
```
gregorydlogan: Caps

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  private EventCatalogUIAdapter getEventCatalogUIAdapter() {
+    final MediaPackageElementFlavor flavor = new MediaPackageElementFlavor("dublincore", "episode");
+    final EventCatalogUIAdapter adapter = catalogUIAdapters.stream().filter(e -> e.getFlavor().equals(flavor)).findAny()
+            .orElse(null);
+    if (adapter == null) {
+      throw new RuntimeException("no adapter found");
```
gregorydlogan, line 317: Caps

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+  }
+
+  private void republishMetadata(final String eventId) {
+    final String workflowId = "republish-metadata";
+    try {
+      final WorkflowDefinition wfd = workflowService.getWorkflowDefinitionById(workflowId);
+      final Workflows workflows = new Workflows(assetManager, workflowService);
+      final ConfiguredWorkflow workflow = workflow(wfd, Collections.emptyMap());
+      if (workflows.applyWorkflowToLatestVersion(Collections.singleton(eventId), workflow).isEmpty()) {
+        throw new RuntimeException(String.format("couldn't start workflow '%s' for event %s", workflowId, eventId));
```
gregorydlogan, line 344: Caps

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+      }
+    } catch (WorkflowDatabaseException | NotFoundException e) {
+      logger.error("unable to get workflow definition {}", workflowId, e);
```
gregorydlogan: Caps

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -0,0 +1,497 @@
...
+  public String getEventMetadata(final String eventId) throws NotFoundException, UnauthorizedException {
+    final Opt<Event> optEvent;
+    try {
+      optEvent = indexService.getEvent(eventId, searchIndex);
+    } catch (SearchIndexException e) {
+      throw new RuntimeException(e);
+    }
+
+    if (optEvent.isNone())
+      throw new NotFoundException("cannot find event with id '" + eventId + "'");
```
gregorydlogan, line 362: Caps.

Going to stop commenting on this now :)

modules/lti-service-remote/src/main/java/org/opencastproject/lti/endpoint/LtiServiceRemoteImpl.java
```diff
@@ -0,0 +1,178 @@
...
+      entity.addPart(file.getSourceName(), new InputStreamBody(file.getStream(), file.getSourceName()));
+    }
+    final HttpPost post = new HttpPost("/");
+    post.setEntity(entity.build());
+    closeConnection(safeGetResponse(post));
+  }
+
+  @Override
+  public void copyEventToSeries(final String eventId, final String seriesId) {
+    final HttpPost get = new HttpPost("/" + eventId + "/copy?seriesId=" + seriesId);
```
gregorydlogan: Variable name!

modules/lti-service-remote/src/main/java/org/opencastproject/lti/endpoint/LtiServiceRemoteImpl.java
```diff
@@ -0,0 +1,178 @@
...
+    final HttpResponse response = getResponse(get);
```
gregorydlogan: This isn't using `safeGetResponse`.  Also, doesn't `getReponse` check for a 200, whereas the copy endpoint returns a 204?

modules/lti/copy-translations.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
gregorydlogan, line 1: Accidental add?

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
gregorydlogan: This looks to me like an out of date comment, but just in case: this is bound to the impl/remote, not the rest endpoint.

gregorydlogan (changes requested):
Haven't actually used the UI yet, this is just looking at the cod.e

etc/workflows/copy-to-event-to-series.xml
pmiddend: Thanks, fixed.

modules/lti-service-api/src/main/java/org/opencastproject/lti/service/api/LtiService.java
pmiddend: Not really, the function inserts or updates, thus upserts. It's got a Wiktionary entry, even: https://en.wiktionary.org/wiki/upsert :)

modules/lti/public/index.html
pmiddend: Okay...more of a PR for the `create-react-app` people, but...fixed that locally here :)

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
pmiddend: Okay, sounds good. Fixed.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
pmiddend: Just to be sure, you mean messages shouldn't end in a "!" in Opencast? Some apparently do in the `lang-en_US.json`.

modules/lti/copy-translations.sh
pmiddend: Not sure what you mean by that? The file is necessary and new.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
pmiddend: That was basically the question I was asking myself as well. At the time of writing, the whole subtitle stuff was not very well documented, so I tried stuff out until paella/theodul actually displayed the subtitles.

pmiddend:
@gregorydlogan I have hopefully fixed all capitalization issues! :)

I've also fixed the build and pom issues.

Note that I'm not working full-time on Opencast anymore, so there is way less time I can put into this PR. Sorry for that...Maybe I can look into the annotation stuff "later".

modules/lti-service-impl/pom.xml
pmiddend: You're right, fixed.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
gregorydlogan: Ok, then let's leave these as is.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
gregorydlogan: Considering Plapadoo's current state, I would be tempted to file this as an issue once this PR is in.  This will come back and bite someone if we don't fix it at some point.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
gregorydlogan: Similar to the caption data questions, this should likely be filed as an issue as well.

modules/lti/copy-translations.sh
gregorydlogan: I hadn't caught that it was used, nevermind :)

gregorydlogan:
@lkiesow I'm fine merging this as is, especially considering there are additional OSGI related changes which will need to go in.  Thoughts?

gregorydlogan (approved):

gregorydlogan:
@pmiddend Conflicts, then I'll merge this in.

kristofkeppens:
@gregorydlogan I've contacted @pmiddend to see if he is still willing to fix this so we can finally see this merged into opencast.

gregorydlogan:
@pmiddend @flyapen Any news?  Conflicts look minor to me.

krinnewitz:
Hey guys, I hope you are all well. I am sorry - I know it took us ages to respond. We have all been busy working in our new jobs and moving to different cities. I just rebased Philipps branch and fixed the conflicts and compile problems. Unfortunately, I didn't have enough time to actually run Opencast and test if everything still works.

gregorydlogan:
@krinnewitz no worries.  At this point we seemed satisfied with this, and the current Travis issue is a known issue in the service registry tests - I've restarted the build.  I'm off from about now until Thursday, so assuming it works I would encourage @lkiesow (or someone he delegates) to merge it before more conflicts crop up.  If not, I'll try to get to it Thursday.

gregorydlogan:
Thanks for the work guys.  I've merge this in as is, if there's major breakage we'll have to chase it down.

PR #1291, opened 2019-12-13
Clean up unused code and ignored tests

pmiddend:
As the title suggests, this PR cleans up some unused code and also removes tests that have `@Ignore` on them.

It was done by applying the `unused declaration` inspection to the whole codebase and then hand-picking. Manual work was needed because simply removing all of it would have broken OSGi (because IntelliJ doesn't know about `activate` and such being actually used) and because not all unused code is _bad_ per se. Stuff like `*Query` classes with `with` methods, for example, and getters in general, I left as is (mostly).

I hope I didn't remove _too_ much. Please correct me if I'm wrong. I'd rather not remove useful code, but _rotten_ code just hinders sensible refactorings and makes the code more "static" than it needs to be.


lkiesow:
Note that this has a few conflicts.

pmiddend:
Conflicts are fixed now.

lkiesow:
Skimmed the code and did some production tests. Everything seems to work as expected → merged.

PR #1293, opened 2019-12-16
Actually update event workflow via API

KatrinIhler:
The docs allow setting processing info when updating an event via POST
to api/events/, but the processing info is never actually used. Now this
will be used to update scheduled events if they aren't recorded yet.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -455,6 +455,41 @@ public class EventsEndpoint implements ManagedService {
           indexService.updateEventAcl(eventId, eventHttpServletRequest.getAcl().get(), externalIndex);
         }
 
+        if (eventHttpServletRequest.getProcessing().isSome()) {
+
+          if (!event.isScheduledEvent() || event.hasRecordingStarted()) {
+            return RestUtil.R.badRequest("Processing can't be updated for events that are already uploaded.");
+          }
+          JSONObject processing = eventHttpServletRequest.getProcessing().get();
+
+          String workflowId = (String) processing.get("workflow");
+          if (workflowId == null)
+            throw new IllegalArgumentException("No workflow template in metadata");
+
+          Map<String, String> configuration = new HashMap<>();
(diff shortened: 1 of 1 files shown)


modules/kernel/src/main/java/org/opencastproject/kernel/security/RemoteUserAndOrganizationFilter.java
```diff
@@ -164,7 +164,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 
           if (!originalUser.hasRole(GLOBAL_ADMIN_ROLE)) {
             // if the original user did not have system privileges, the target user must not gain those, either.
-            for (String systemRole : GLOBAL_SYSTEM_ROLES) {
+            for (String systemRole : GLOBAL_SYSTEM_ROLES) { // This IS WRONG
```
lkiesow: If this is wrong… why is this here?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -455,6 +455,41 @@ public Response updateEventMetadata(@HeaderParam("Accept") String acceptHeader,
           indexService.updateEventAcl(eventId, eventHttpServletRequest.getAcl().get(), externalIndex);
         }
 
+        if (eventHttpServletRequest.getProcessing().isSome()) {
+          if (event.isScheduledEvent() && !event.hasRecordingStarted()) {
```
lkiesow: Avoid unnecessary long conditions:

```suggestion
          if (!event.isScheduledEvent() || event.hasRecordingStarted()) {
            return RestUtil.R.badRequest("Processing can't be updated for events that are already uploaded.");
          }
```

modules/kernel/src/main/java/org/opencastproject/kernel/security/RemoteUserAndOrganizationFilter.java
KatrinIhler: Because... I forgot to remove this. Whoops.

KatrinIhler:
@lkiesow Fixed!

PR #1294, opened 2019-12-17
Remove ffmpeg banner by default

gregorydlogan:
Adding -hide_banner to the ffmpeg commandline by default.  This removes a large chunk of text that ffmpeg spits out by default, which tends to clutter our log files with garbage.

Happy to rebase this to 8.x, or even 7.x if people think it's appropriate.

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncoderEngine.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncoderEngine.java
@@ -342,6 +342,7 @@ public class EncoderEngine implements AutoCloseable {
     command.add(binary);
     command.add("-nostdin");
     command.add("-nostats");
+    command.add("-hide_banner");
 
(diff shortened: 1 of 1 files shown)


lkiesow:
I'm wondering about the use-case for this. Do you see any specific large data this would hide? Because this also hides some useful data for debugging and I would have expected the logging selection to be handled in `handleEncoderOutput(…)`:

https://github.com/opencast/opencast/blob/1f421f1b5cb2dec6d77770eee81e46a34538b60a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncoderEngine.java#L428-L464

gregorydlogan:
My motivation here was strictly to clean up the logging output since (99% of the time anyway) error reports come in without this information anyway.  I'm not opposed to moving the logic to this method instead.  I'll take another look at some point, likely after the holidays.

gregorydlogan:
Considering that I'm unwilling to spend enough time to catch all of the potential usecases, I'm going to just decline this :)

PR #1296, opened 2019-12-17
Read published files direct from filesystem if possible

smarquard:
This addresses a performance issue when building the solr search index with a large number of published events. The event metadata files and catalogs (e.g. dublincore.xml, segments.xml, slidetext.xml) are loaded into the workspace from http requests which is relatively slow, especially if Opencast is configured with https.

With a large number of events (60K+), this change reduced indexing time from > 6 hours to just over 1 hour on a test system.

This uses a similar implementation to the technique used to read asset manager files directly, though the location of DistributionPathUtils in asset-manager-utils is probably questionable and I'm open to suggestions there.

### DIFF
--- a/modules/asset-manager-util/src/main/java/org/opencastproject/assetmanager/util/DistributionPathUtils.java
+++ b/modules/asset-manager-util/src/main/java/org/opencastproject/assetmanager/util/DistributionPathUtils.java
@@ -1,0 +1,127 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.opencastproject.assetmanager.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.osgi.framework.BundleContext;
+import org.osgi.service.component.ComponentContext;
+import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


modules/asset-manager-util/src/main/java/org/opencastproject/assetmanager/util/DistributionPathUtils.java
```diff
@@ -0,0 +1,127 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.net.URI;
+import java.nio.file.Paths;
+
+/**
+ *
+ */
+public final class DistributionPathUtils {
```
gregorydlogan, line 37: What's the difference between `AssetPathUtils` and this class?

smarquard: AssetPathUtils provides direct access to files (if possible) stored by the Asset Manager. DistributionPathUtils provides direct access to files (if possible) published to the engage distribution channel.

gregorydlogan:
I don't see anything wrong with this, but I also don't have any prod instances.  Waiting on @lkiesow's thoughts (or someone at Elan) since you've got a few large installs to take care of.

gregorydlogan (approved):

lkiesow:
I can probably take a look at this until end of day, Monday.

lkiesow:
This looks quite reasonable and depending on the set-up, could probably speed up index rebuilds quite a bit. My only complaint is that the distribution utils should not be part of the assetmanager utils bundle. That's weird. We should either put it into a separate bundle or rename the existing bundle to something which fits both utility classes.

smarquard:
I agree that the location in assetmanager utils bundle is weird. Do you have a preferred way to resolve this (i.e. separate bundle, or rename the bundle)?

lkiesow:
I really don't mind either way.
Maybe rename this to `opencast-shared-filesystem-utils`?
Should be easier than splitting this up

gregorydlogan:
+1 to @lkiesow's comment.  While you're at it, maybe move `AssetPathUtils` as well?

lkiesow:
yep, I meant just rename the asset-manager-utils module.

gregorydlogan:
@smarquard This is ready to go aside from that rename as far as I'm concerned.

smarquard:
I'm not sure I will be able to return to this any time soon given local #COVID19 priorities. If anyone else would like to push changes and get this ready to merge please go ahead.

lkiesow:
I've asked @marwyg to quickly finish this so that we can hopefully include this in 8.3

gregorydlogan:
Closing this in favour of #1483, which completes the work.

Issue #1219, opened 2019-11-07
Preview extraction fails on short videos


lkiesow:
When ingesting a video with a length shorter than 5s, the extraction of a preview image will fail due to the configuration of the default workflow.

lkiesow:
```
2020-01-05 02:12:18,761 | ERROR | (AbstractJobProducer$JobRunner:344) - Error handling operation 'Image': org.opencastproject.serviceregistry.api.ServiceRegistryException: Error handling operation 'Image'                                  
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1530)                                                                                                                                       
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314)                                                                                                                                       
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:273)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.opencastproject.composer.api.EncoderException: An image could not be extracted from the track http://localhost:8080/files/mediapackage/9bf975ce-67ad-406a-9193-9e0807c13d20/f7544c6d-69c2-4943-85e5-9b66432b0a5c/0_5s.mp4 with 
id f7544c6d-69c2-4943-85e5-9b66432b0a5c because the extraction time (1.0 second(s)) is outside of the track's duration (0.535 second(s))
        at org.opencastproject.composer.impl.ComposerServiceImpl.image(ComposerServiceImpl.java:1121)
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1422)
        ... 6 more

```

PR #1327, opened 2020-01-06
Video Segemntation On Short Videos

lkiesow:
Our video segmentation works extremely bad on short videos, e.g. trying
to divide a 0.5 seconds video into 100 segments, causing the segments to
be shorter than a single frame which then caused problems with the
slide-preview extraction.

Assuming that a segmentation on extremely short videos is useless at
best and causes workflow failures or services in error states at worst,
this patch will make Opencast skip the segmentation on videos shorter
than 15 seconds.

This is related to #1219 and #1326

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/videosegmenter-workflowoperation/src/main/java/org/opencastproject/workflow/handler/videosegmenter/VideoSegmenterWorkflowOperationHandler.java
+++ b/modules/videosegmenter-workflowoperation/src/main/java/org/opencastproject/workflow/handler/videosegmenter/VideoSegmenterWorkflowOperationHandler.java
@@ -45,7 +45,6 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Iterator;
 import java.util.List;
 
 /**
@@ -61,6 +60,9 @@ public class VideoSegmenterWorkflowOperationHandler extends AbstractWorkflowOper
 
   /** Name of the configuration key that specifies the flavor of the track to analyze */
   private static final String PROP_TARGET_TAGS = "target-tags";
+
+  /** Minimum video length in seconds for video segmentation to run */
+  private static final int MIN_VIDEO_LENGTH = 30000;
 
   /** The composer service */
   private VideoSegmenterService videosegmenter = null;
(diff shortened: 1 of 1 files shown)


mtneug:
Maybe make this configurable? I can imagine that it doesn't make sense for even longer videos (1-5 min.).

lkiesow:
Not doing that. It's nothing anyone will really configure ever if there is a sensible default.
I can increase this to maybe 30sec though?

mtneug:
Fair enough, though I would have misused this safety measure to skip videos of up to 5 min. where I don't think segmentation would be beneficial 😄.

Just as a safety measure 30 sec. or even 15 sec. seems fine to me. What I don't know, however, is if we should somehow consider [`absoluteMin`](https://github.com/opencast/opencast/blob/f940467be0dc3fcb52bcfed6c7aa99f92bb5514d/etc/org.opencastproject.videosegmenter.ffmpeg.VideoSegmenterServiceImpl.cfg#L38). In theory, you could set it to a very high value such that any number of seconds at given fps would lead to an error.

modules/videosegmenter-workflowoperation/src/main/java/org/opencastproject/workflow/handler/videosegmenter/VideoSegmenterWorkflowOperationHandler.java
```diff
@@ -110,6 +110,12 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, Jo
     }
     Track track = candidates.get(0);
 
+    // Skip operation if media is shorter than 15s since we won't generate a sensible segmentation on such a short
+    // video anyway.
+    if (track.getDuration() != null && track.getDuration() < 15000) {
```
gregorydlogan: 15 seconds is fine to me, although I somewhat agree with Matthias in that sub 5 minute videos don't really benefit from segmentation.

My complaint here whoever, is somewhat simpler: Magic numbers.  IMO this should be a constant properly defined so it shows up in the javadocs.

gregorydlogan (changes requested):

lkiesow:
- added variable for time
- set time to 30sec

PR #1347, opened 2020-01-10
LDAP Configuration

lkiesow:
This patch simplifies the LDAP configuration by doing the long-discussed
and anticipated move from the external, old and out-of-date LDAP sample
security configuration into the main security configuration file.

It also updates the LDAP configuration guide to reflect this move as
well as doing a general overhaul since it seemed like the guide still
contained relics from our wiki days.

---

A fully functional test configuration using sysforums public test server following this guide can be found at:

- https://data.lkiesow.io/opencast/ldapconfig/mh_default_org.xml
- https://data.lkiesow.io/opencast/ldapconfig/org.opencastproject.userdirectory.ldap-forumsys.cfg

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


CGreweling:
Hey @lkiesow i am not sure if it is really necessary. Bu i think the "ROLE_USER" is also needed, to be added.

lkiesow:
@CGreweling, you are talking about the linked test configuration, not the patch, right?
For that, I think it depends on what you want to do.

etc/security/mh_default_org.xml
```diff
@@ -507,6 +507,56 @@
...
+            <constructor-arg index="0" value="ou=GroupName,dc=my-institution,dc=country" />
+            < ! - - Filter to located valid users. Use {0} as a placeholder for the login name - - >
+            <constructor-arg index="1" value="(uid={0})" />
+            <constructor-arg ref="contextSource" />
+          </bean>
+        </property>
+        - - >
+      </bean>
+    </constructor-arg>
+    < ! - - Defines how the user attributes are converted to authorities (roles) - - >
```
mliradelc, line 555: The `< ! - -  .....  - - - >` Comment indicators needs to be changed to `<!-- ..... -->` because the previous generates a read error from Opencast.


Maybe is because I had to uncomment some lines because this kind of file doesn't support nested comments. But anyways I think is more foolproof have all the comments not nested.

mliradelc:
@lkiesow We tried in our new opencast 9 and we couldn't make it work, because It doesn't allow the users to enter the admin-UI, always redirects to the presenter, and to the presenter in the same node, creating a 404 error instead to go to another node.

The file is almost the same from the previous one so it should work.

lkiesow:
If you configure it for your local LDAP, the security configuration file you end up with should be the same one you are using with Opencast 8 right now. To make things easier, you can also use the pre-configured examples which should just work.

This patch does not modify the code at all., so it cannot create any problems which haven't been there before :-)

mliradelc (commented):
> This patch does not modify the code at all., so it cannot create any problems which haven't been there before :-)

You are right! I checked with our config file so It should be another thing in between.  Logs in perfect to the Presenter mode, hence the ldap config works.

etc/security/mh_default_org.xml
mliradelc: @lkiesow Please check my point with the comments, what do you think? Besides that detail is ok for me to merge.

lkiesow: Unfortunately, we cannot just change that since you cannot use an XML comment within an XML comment. The notation of adding spaces – which need to be removed when removing the outer comment – is what the previous LDAP configuration was already using and I did not have a better idea than that:

https://github.com/opencast/opencast/blob/71480e193b927ecf4397bdeffc1fecd99f625e86/etc/security/security_sample_ldap.xml-example#L519-L522

mliradelc (commented):
Never mind All Ok

PR #1359, opened 2020-01-16
Fix workflow dropdown in start task

### LINKED ISSUE #1357: "Start Task" shows the wrong workflow options
When selecting a workflow in the "Start Task" dialog, the options for the workflow **below** the selected one are shown:

image

image

This happens in 8.0 on stable.opencast.org, but not on legacy.opencast.org. I can't test develop.opencast.org right now, because I only see translation keys and no events, there. :wink: 

Also it does not happen consistently, unfortunately, but during my brief testing right now it happened more times than not.

KatrinIhler:
When selecting a workflow in start task, new event or event details, the workflow being picked would
often not match the description and configuration panel shown. This is
fixed by using track by workflow id.

This fixes #1357


mliradelc:
Ok, will look into that :)

mliradelc (approved):
Code Tested, works without hassle! 

Good job!

lkiesow:
thanks for the review.

PR #112, opened 2018-03-02
MH-12709: Replacing broken conflict detection logic

gregorydlogan:
The previous check searched for events which started or ended in a pair of windows.  This missed several edge cases, and was thus replaced.

*Work sponsored by UCT*

PR #1370, opened 2020-01-20
Fix Conflict Detection

lkiesow:
This patch fixes Opencast's conflict detection for scheduled events
which would marks two events (9am to 10am) and (10am to 11am) as
conflicting which is a very exexpected behavior.

Worse, even disjoind end/start times could cause conflicts with no sane
reason why they would. E.g. (8:00:00 - 9:00:00) and (9:00:10 to 9:30:00)
would be marked as conflicting.

This patchfixes this erratic behavior, returning to the iCal definition
for start dates, end dates and conflicts we have always used.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/scheduler-api/src/main/java/org/opencastproject/scheduler/api/Util.java
+++ b/modules/scheduler-api/src/main/java/org/opencastproject/scheduler/api/Util.java
@@ -45,7 +45,7 @@ import java.util.TimeZone;
 
 public final class Util {
   /** The minimum separation between one event ending and the next starting */
-  public static final int EVENT_MINIMUM_SEPARATION_MILLISECONDS = 60 * 1000;
+  public static final int EVENT_MINIMUM_SEPARATION_MILLISECONDS = 0;
 
   private static final Logger logger = LoggerFactory.getLogger(Util.class);
   private static final TimeZoneRegistry registry = TimeZoneRegistryFactory.getInstance().createRegistry();
--- a/modules/scheduler-impl/src/test/java/org/opencastproject/scheduler/impl/SchedulerServiceImplTest.java
+++ b/modules/scheduler-impl/src/test/java/org/opencastproject/scheduler/impl/SchedulerServiceImplTest.java
@@ -1138,14 +1138,6 @@ public class SchedulerServiceImplTest {
       //Event A contains event B entirely
       conflicts = schedSvc.findConflictingEvents("Device A", new Date(currentTime + hours(23)), new Date(currentTime + hours(26)));
       assertEquals(1, conflicts.size());
-
-      //Event A ends with less than one minute before event B starts
-      conflicts = schedSvc.findConflictingEvents("Device A", new Date(currentTime + hours(23)), new Date(currentTime + hours(24) - seconds(1)));
-      assertEquals(1, conflicts.size());
-
-      //Event A begins than one minute after event B ends
-      conflicts = schedSvc.findConflictingEvents("Device A", new Date(currentTime + hours(25) + seconds(1)), new Date(currentTime + hours(27)));
-      assertEquals(1, conflicts.size());
     }
   }
 


smarquard:
It doesn't make sense to just set EVENT_MINIMUM_SEPARATION_MILLISECONDS = 0 because that's a concept used in the code e.g. in eventWithinMinimumSeparation() so setting it to 0 defeats the whole purpose.

Functionally, I think many capture agents are going to have difficulty dealing with consecutive recordings with no interval between them. So this seems like a breaking change as far as CAs are concerned. If Opencast permits this, then CAs need to understand to stop one recording early or start the next one late if there's time required (as there typically is) to shutdown a recording pipeline, move files around,  and start a new one.

At least this should be configurable with a safe default, and production sites can set it to 0 at their own risk.

lkiesow:
This is actually a regression bug since it was definitely possible to configure this before. There is also no documentation for capture agents anywhere which guarantees capture agents any pause between any recordings, so it should not break any reasonable implementation. As a capture agent author, I can also say that this is no problem for pyCA.

Setting this to 0 was the minimal change for me to fix this bug. But I can make this configurable if you would like me to.

smarquard:
Firstly, I'm happy with the change if it's configurable. I think the default should be the same as present (60s) but don't mind too strongly about that. If the default is not the same, it could be a breaking change for some sites and CAs, and should be listed in the changelog.

For reference, I checked where this was introduced and it was part of a set of changes to the scheduler code in MH-12709 PR
https://github.com/opencast/opencast/pull/112

The change was clearly intentional and reviewed, so I don't think it's appropriate to call this a regression especially as it's been like since 4.2 it seems. I appreciate you have a different view of what correct behaviour is but "as it's always been" is now quite a few releases ago.

gregorydlogan:
Back in the day there was definitely a 1 minute minimum separation between recordings.  This was in the code, and tested, but got lost by the way side.  It was a byproduct of the hardware at the time not being able to cycle over to a new capture in (reliably) less than 30 secondss.

gregorydlogan:
Looking back at the logs, the minimum appears to have been added in 4.x as part of https://opencast.jira.com/browse/MH-12709, with 31d6d16b04d.  Presumably there was a good reason for that.  Whether it is still valid today is unclear.

If we're going to remove that separation, let's get rid of the variable entirely.  No sense in keeping dead vars around.  If we're not, then let's document it.  IMO, I would document that one minute turnaround and be done with it.

gregorydlogan:
I'm going to close this.  If you've still got a customer complaining about this we can reopen and continue discussion.

Rillke:
>  If you've still got a customer complaining about this we can reopen and continue discussion.

Though it's not high-priority. 60s is a lot, according to our lecturers  (that's 16 times 2 times 1 minute = about 1/3 lecture over a semester). If the buffer time would be either configurable, or removed, they'd be happy. Currently I have to recompile Opencast in order to get rid of this and end a discussion which took  place amongst 4 well-paid people for 20 minutes.

Our Stud.IP plugin takes care of the scheduling and has a configurable buffer time built-in.

Rillke:
[Here's the full reasoning for that.](https://groups.google.com/a/opencast.org/g/users/c/FvAaw_5TUoA/m/3Fv98pvABQAJ) In case it's rejected, it's also not too bad for us, as we could either continue to compile a custom version of opencast, have a configurable buffer time, or tell our room service that for technical reason, they have to change their current practice.

gregorydlogan:
So, now that the original Matterhorn source is posted, I can finally go digging properly.  The code I'm remembering was added by https://github.com/opencast/matterhorn/commit/9f59cf8f4a84d050da18658bb6cae454aef8d805.  I can't find any enforcement in the admin UI, but the code is truly spaghetti (there's a reason it was replaced!) and I distinctly remember the UI not accepting sub-minute turnaround times.

That being said, the reason for adding this is *12 years* old.  I'm not saying it has to stay, since it's not technically part of the CA API, but hardware has come a long way.  If we're seeing a customer need to make this configurable, or go away, I'm fine with it.  Personally, I would remove it entirely and leave enforcement of any kind of conflict detection to customer business rules.  In large part, this is because we don't know what hardware you have - your admin staff need to not schedule things which won't work with your hardware.

tl;dr: Fine, remove these checks.  But make sure there's something in the release notes to make it clear that your mileage may vary if you schedule things to start and stop at the same time.

geichelberger:
Can this PR be part of OC 12?

lkiesow:
Switched the branch.
Now the only thing left is for someone to review this.

geichelberger (approved):
LGTM had to rebase it locally before testing, but that should not make any difference.

gregorydlogan:
@lkiesow Can you ensure there is a note in the release notes for this?

PR #1371, opened 2020-01-21
Split AdminUI in Java and JavaScript parts

isabellakutger:
This pull request splits the admin ui into two parts:
- the Java backend (remains at admin-ui)
- the JavaScript frontend (moved to admin-ui-frontend)

Purpose of this split is to facilitate migrating from AngularJS (1.x) to Angular (8+) in a new module in parallel to the current admin-ui.


mliradelc:
Hi, 
First of all, thank you for your work!   😄

~~Now I don't understand very clear what makes this PR (Besides the title of course). Can you please explain what about is this PR?~~ and Squash and merge it to have a better view of the changed files.

lkiesow:
Don't request a dependency exception for this. Since this has no Java code, the proper plugin configuration should be no more than:

https://github.com/opencast/opencast/blob/71480e193b927ecf4397bdeffc1fecd99f625e86/modules/animate-api/pom.xml#L27-L30

lkiesow:
Generally looks good. Just a few very minor issue:

---

It seems like you moved one file you shouldn't have, causing exceptions on start-up:

```
2020-01-23 21:09:19,590 | ERROR | (R6LogServiceLogger:50) - bundle opencast-admin-ui:9.0.0.SNAPSHOT (48)[org.opencastproject.adminui.userdirectory.UIRolesRoleProvider(260)] : The activate method has thrown an exception                    
java.lang.NullPointerException: null                                                                                                                                                                                                          
        at java.io.Reader.<init>(Reader.java:78) ~[?:1.8.0_232]                                                                                                                                                                               
        at java.io.InputStreamReader.<init>(InputStreamReader.java:113) ~[?:1.8.0_232]                                                                                                                                                        
        at org.apache.commons.io.IOUtils.readLines(IOUtils.java:1415) ~[?:?]                                                                                                                                                                  
        at org.apache.commons.io.IOUtils.readLines(IOUtils.java:1440) ~[?:?]                                                                                                                                                                  
        at org.opencastproject.adminui.userdirectory.UIRolesRoleProvider.activate(UIRolesRoleProvider.java:81) ~[?:?]                                                                                                                         
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_232]
        …
```

The reason should probably be 

```
R100    modules/admin-ui/src/main/webapp/scripts/shared/services/wizards/new-group/roles.js     modules/admin-ui-frontend/app/scripts/shared/services/wizards/new-group/roles.js

```

---

It seems like you missed moving a few files:

```
modules/admin-ui/.checkstyle
modules/admin-ui/.editorconfig
modules/admin-ui/.gitattributes
modules/admin-ui/.gitignore
```

Of these and of a few you already moved, I think some may be deleted:

```
modules/admin-ui-frontend/.editorconfig
modules/admin-ui-frontend/.gitattributes
modules/admin-ui-frontend/.jscsrc
modules/admin-ui-frontend/.jshintignore
```

lkiesow:
Good catch, I missed the translation files.
But you also need to adjust their configuration in the `.crowdin.yaml`.
And this also has a minor merge conflict.

lkiesow:
@pascalseeland, @isabellakutger, can you take a look at the conflicts?
Ping me (or request a review) once you are done and you think this is ready to be merged.

pascalseeland:
@lkiesow I updated the pull requests accordingly. I think every time dependabot updates the package.json we do get those as conflicts.

lkiesow:
`modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json` is still missing in the Crowdin configuration at `.crowdin.yaml` and `modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/` should not be unnecessary to include in the admin-ui JAR.

pascalseeland:
I did have a look at the files in modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/. Those are as far as I can see only needed to generate the data provided by the LanguageServiceEndpoint in the admin-ui. From a first analysis, the data provided by this endpoint could also be statically provided as fixed file in admin-ui-frontend and rendering the Endpoint unnecessary. The functionally of detecting the best available language could than be implemented in the languageservice.js of the admin-ui-frontend. In case additional languages would be added, it would necessary to manually add the required information.  I would howevery accept that task and look for a solution that does not require this patching moving forwards to the new admin-ui-frontend

lkiesow:
The backend does not actually use the translations, just the file names to check which translations are available. My suggestion is hence to just place an empty file int here instead:

```sh
cd modules/admin-ui/
for i in src/main/resources/public/org/opencastproject/adminui/languages/lang-*.json; do echo '{}' > "$i"; done
```

Though we might possibly want to change that to something like `{"dummy":"dummy"}` depending on if Crowdin can handle empty files or not (something we will see only after this has been merged.)

In the Crowdin translation, I would then suggest having:

```yaml
   -
    source: '/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json'
    translation: '/modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/lang-%locale_with_underscore%.json'

   -
    source: '/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json'
    translation: '/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-%locale_with_underscore%.json'
```

…to ensure both folders get updated when new languages are added or old ones are removed.

pascalseeland:
Sorry, I did not see your comment and edited mine above. So here again my suggestion:
I did have a look at the files in modules/admin-ui/src/main/resources/public/org/opencastproject/adminui/languages/. Those are as far as I can see only needed to generate the data provided by the LanguageServiceEndpoint in the admin-ui. From a first analysis, the data provided by this endpoint could also be statically provided as fixed file in admin-ui-frontend and rendering the Endpoint unnecessary. The functionally of detecting the best available language could than be implemented in the languageservice.js of the admin-ui-frontend. In case additional languages would be added, it would necessary to manually add the required information. I would howevery accept that task and look for a solution that does not require this patching moving forwards to the new admin-ui-frontend

lkiesow:
@isabellakutger, what's the current state? Is this ready for review?

pascalseeland:
We opted for not adding the extractions of the formatting keys from the translation files. Instead, we opted for adding a check in Grunt to verify that all languages are registered in the languages.json

lkiesow:
I've found a few more unused files in the admin-ng module but I've simply created a new pull request for those. @isabellakutger or @pascalseeland, does one of you have the time to do a quick review on that pull request (#1413)?

gregorydlogan:
This appears to have broken my build somehow, filed as #1418.

PR #1391, opened 2020-02-05
Documentation: Text Extraction Configuration

Rillke:
- --psm is just -psm on Debian with Tesseract 3.x
- --user-words path must not be put in quotation marks (The way ProcessBuilder is used by the Impl. does not allow this)

### Your pull request should…

* [x] have a concise title
* ~[ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists~
* [x] pass automated testing
* ~[ ] have appropriate tags applied~


lkiesow:
The command line was meant as an example. To not have to explain that different versions have different options (`-psm` vs `--psm`), my suggestion for this documentation is to remove `--psm` entirely. The used `--psm 3` is the default option anyway. What do you think?

lkiesow:
Merging this like this for now. It does improve the documentation after all. Will try the make the suggested change if I have some free time later.

Issue #1384, opened 2020-02-01
WebM-Videos from Opencast Studio will not be processed with schedule-and-upload workflow


rrolf:
**Describe the bug**
Use Opencast Studio and upload the video to an Opencast with the schedule-and-upload workflow selected. 
The processing will fail in the "image" operation with the following error:

```
2020-02-01 13:28:21,643 | ERROR | (WorkflowOperationWorker:141) - Workflow operation 'operation:'image', position:17, state:'FAILED'' failed
org.opencastproject.workflow.api.WorkflowOperationException: Track https://oc-test.virtuos.uni-osnabrueck.de/files/mediapackage/6269562a-8cc0-4190-aea5-c5b8137e653e/f76dea5f-2ad4-4969-bca7-d20222e1cda1/presenter_video___recording__1_.webm cannot tell its duration
	at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler$6.apply(ImageWorkflowOperationHandler.java:484) ~[?:?]
	at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler$6.apply(ImageWorkflowOperationHandler.java:481) ~[?:?]
	at com.entwinemedia.fn.StreamOp$35$1.apply(StreamOp.java:483) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.StreamOp$Iterate.hasNext(StreamOp.java:668) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.data.ListBuilderUtils.fillAndFinish(ListBuilderUtils.java:112) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.data.ListBuilderUtils.createNew(ListBuilderUtils.java:64) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.data.ImmutableLooseListBuilder.mk(ImmutableLooseListBuilder.java:62) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.Stream.toList(Stream.java:400) ~[27:functional:1.4.2]
	at com.entwinemedia.fn.Stream.toList(Stream.java:394) ~[27:functional:1.4.2]
	at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler.configure(ImageWorkflowOperationHandler.java:487) ~[?:?]
	at org.opencastproject.workflow.handler.composer.ImageWorkflowOperationHandler.start(ImageWorkflowOperationHandler.java:142) ~[?:?]
[...]

PR #1395, opened 2020-02-08
Fix Image Extraction At Position Zero

lkiesow:
This patch fixes the image extraction at a position of zero seconds
which would fail due to the trim filter not including the first frame.

This is related to #1384

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


gregorydlogan:
I've been fighting this with our old demo media, as well as the default media from the helper scripts, thanks for fixing this.

PR #1396, opened 2020-02-09
Image Extraction Without Stream Duration

lkiesow:
Trying to extract an image from a video with no duration and/or frame
count information contained in the video container will cause problems
like NullPointerExceptions at multiple places in the `image` workflow
operation.

This patch updates the code to gracefully handle such scenarios without
just breaking down. This also slightly modified the operation's behavior
in that image extraction from streams with no duration information now
works exactly like image extraction from streams with a single frame:
We just extract the first frame we know does exist.

This is related to #1384

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


gregorydlogan:
I've been fighting this with our old demo media, as well as the default media from the helper scripts, thanks for fixing this.

PR #1409, opened 2020-02-18
Asset Upload Title

lkiesow:
Configuring custom asset uploads without rebuilding Opencast is nearly
impossible since the title for the upload options is always taken from
the translation files which are build into Opencast.

This minimal patch allows the title to also be specified as part of the
upload configuration, allowing for configuration-only specifications of
new upload options.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/docs/guides/admin/docs/configuration/admin-ui/asset-upload.md
+++ b/docs/guides/admin/docs/configuration/admin-ui/asset-upload.md
@@ -23,13 +23,21 @@ listprovider properties file:
 
 Two source types are enabled by default for use in the Admin UI.
 
-    EVENTS.EVENTS.NEW.SOURCE.UPLOAD.NON_SEGMENTABLE={
-       "id":"track_presenter", "type":"track", "flavorType":"presenter",
-       "flavorSubType":"source", "multiple":false, "displayOrder": 1}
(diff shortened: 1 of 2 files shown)


karendolan:
@lkiesow I don't understand where the "title" should be set. How is that param compatible with the displayFallback and displayOverride params?
https://github.com/opencast/opencast/pull/1409/files#diff-32a5eb0192d18c4d224c6b9731c8e8e7R179-R184

lkiesow:
@karendolan, I might just have overlooked `displayOverride`. That option might make this change obsolete. I will take another look at this… which might take me until next week.

karendolan:
> @karendolan, I might just have overlooked `displayOverride`. That option might make this change obsolete. I will take another look at this… which might take me until next week.

The "displayFallback" set of params are the one that will add text in the admin UI if there is no translation. Once translation is available, the translation overrides that param. The "displayOverride" overrides existing translation.

The issue appears to be that the custom params table is too far down on the docs page to be seen when skimming for information. The sample example, higher up on  the page, references a subset of the possible params.

IMHO, inserting an anchor link to the custom param table higher at the top of the docs page (where you had added the "title" reference) would remove a lot of confusion.

PR #1421, opened 2020-02-19
Remove compose in favor of encode

lkiesow:
By now, the compose workflow operation is doing an exact subset of the
encode workflow operation which universally can be used as a drop-in
replacement.

In addition to the functionality of compose, encode supports parallel
encoding as well as the adaptive encoding profiles but these are
additional features and using the same configuration as compose, encode
works identical. It even uses mostly the same code internally.

So, there is no real reason to keep compose around any longer. That is
why this patch drops it.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


pascalseeland (approved):
Removing unneeded old code and simplyfing config for users is good.

lkiesow:
@pascalseeland, did you test this? In that case, I could merge this.

pascalseeland:
@lkiesow Yes, I did a checkout of the PR and it did encode my testvideos without issues.

gregorydlogan:
Signing off on this given @lkiesow and @pascalseeland's results.

PR #1425, opened 2020-02-21
Don't duplicate publications

KatrinIhler:
When duplicating events via the duplicate workflow, we shouldn't duplicate publications since they then point to the same videos, leading to problems when one of the events is retracted.

This is NOT reproducible with the current configuration of the duplicate workflow of the community, since we only set source flavors to */* and publications don't have a flavor. If you also set the source-tags to "archive" though, the publications will be selected and duplicated.


karendolan:
Hi @KatrinIhler, a side question, what is the use case for the duplicate workflow? Would someone want to duplicate archived resources from one mediapackage to another in order to re-edit the media and publish the new resources to that new mediapackage id?

lkiesow:
Imagine a conference recording where you recorded a whole block and want to cut out the specific talks later on. You can just import but not publoish the recording, then duplicate it, and finally cut out the talks and publish them separately,

PR #1426, opened 2020-02-21
Autodetect browsers for admin ui tests, fix phantomJS openssl issue

### LINKED ISSUE #1424: opencast-admin-ui(-frontend) does not build on Debian 10
**Describe the bug**
Attempting to build the admin UI on Debian 10 displayed errors at PhantomJS startup similar to the following
```
[INFO] 140164897451648:error:25066067:DSO support routines:DLFCN_LOAD:could not load the shared library:dso_dlfcn.c:185:filename(libssl_conf.so): libssl_conf.so: cannot open shared object file: No such file or directory 
```

This can be resolved by removing the `openssl_conf = default_conf` line from `/etc/ssl/openssl.cnf`, or starting the build with `OPENSSL_CONF=asdf mvn install...`

**Environment Information**
 - Debian 10, but not Debian 9, or Ubuntu 16, or 18

gregorydlogan:
This PR introduces autodetection for the admin ui unit tests.  If no browsers are detected it falls back to PhantomJS, but if a real browser is installed it prefers that.  If multiple browsers are installed it tests with *all* of them.

This PR also resolves the Debian 10 OpenSSL issue, where the config file breaks PhantomJS.  It works around this by setting the config file location to blank, which causes file load warnings/errors, but does not affect the build.

Fixes #1424

### DIFF
--- a/docs/guides/developer/docs/installation/source-linux.md
+++ b/docs/guides/developer/docs/installation/source-linux.md
@@ -25,6 +25,7 @@ Required:
     ffmpeg >= 3.2.4
     maven >= 3.1
     python >= 2.6, < 3.0
+    firefox/chrome/some other major browser
     unzip
     gcc-c++
     tar
--- a/modules/admin-ui/Gruntfile.js
+++ b/modules/admin-ui/Gruntfile.js
@@ -1,6 +1,10 @@
 // Generated on 2016-03-07 using generator-angular 0.15.1
 'use strict';
 
+// Prevents Debian 10 (at least) from erroring out if PhantomJS is being used
+// Appears to be related to something in the config file, but we can't exactly ship our own config
+process.env.OPENSSL_CONF='';
+
 // # Globbing
 // for performance reasons we're only matching one level down:
 // 'test/spec/{,*/}*.js'
@@ -454,8 +458,7 @@ module.exports = function (grunt) {
       },
       coverage: {
         singleRun : true,
-        reporters : ['dots', 'coverage'],
-        browsers  : ['PhantomJS']
+        reporters : ['dots', 'coverage']
       }
     }
   });
--- a/modules/admin-ui/Gruntfile.js
+++ b/modules/admin-ui/Gruntfile.js
@@ -1,10 +1,6 @@
 // Generated on 2016-03-07 using generator-angular 0.15.1
 'use strict';
 
-// Prevents Debian 10 (at least) from erroring out if PhantomJS is being used
-// Appears to be related to something in the config file, but we can't exactly ship our own config
-process.env.OPENSSL_CONF='';
-
 // # Globbing
 // for performance reasons we're only matching one level down:
 // 'test/spec/{,*/}*.js'
--- a/modules/admin-ui/package-lock.json
+++ b/modules/admin-ui/package-lock.json
@@ -1422,12 +1422,6 @@
         "is-arrayish": "^0.2.1"
       }
     },
-    "es6-promise": {
-      "version": "4.2.6",
-      "resolved": "https://registry.npmjs.org/es6-promise/-/es6-promise-4.2.6.tgz",
-      "integrity": "sha512-aRVgGdnmW2OiySVPUC9e6m+plolMAJKjZnQlCwNSuK5yQ0JN61DZSO1X1Ufd1foqWRAlig0rhduTCHe7sVtK5Q==",
-      "dev": true
-    },
     "escape-html": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/escape-html/-/escape-html-1.0.3.tgz",
@@ -3554,16 +3548,6 @@
         }
       }
     },
-    "hasha": {
-      "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/hasha/-/hasha-2.2.0.tgz",
-      "integrity": "sha1-eNfL/B5tZjA/55g3NlmEUXsvbuE=",
-      "dev": true,
-      "requires": {
-        "is-stream": "^1.0.1",
-        "pinkie-promise": "^2.0.0"
-      }
-    },
     "he": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/he/-/he-1.2.0.tgz",
@@ -4137,12 +4121,6 @@
         "is-unc-path": "^1.0.0"
       }
     },
-    "is-stream": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/is-stream/-/is-stream-1.1.0.tgz",
-      "integrity": "sha1-EtSj3U5o4Lec6428hBc66A2RykQ=",
-      "dev": true
-    },
     "is-typedarray": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/is-typedarray/-/is-typedarray-1.0.0.tgz",
@@ -4515,36 +4493,11 @@
       "integrity": "sha1-ENjIz6pBNvHIp22RpMvO7evsSjE=",
       "dev": true
     },
-    "karma-phantomjs-launcher": {
-      "version": "1.0.4",
-      "resolved": "https://registry.npmjs.org/karma-phantomjs-launcher/-/karma-phantomjs-launcher-1.0.4.tgz",
-      "integrity": "sha1-0jyjSAG9qYY60xjju0vUBisTrNI=",
-      "dev": true,
-      "requires": {
-        "lodash": "^4.0.1",
-        "phantomjs-prebuilt": "^2.1.7"
-      }
-    },
-    "kew": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/kew/-/kew-0.7.0.tgz",
-      "integrity": "sha1-edk9LTM2PW/dKXCzNdkUGtWR15s=",
-      "dev": true
-    },
     "kind-of": {
       "version": "6.0.2",
       "resolved": "https://registry.npmjs.org/kind-of/-/kind-of-6.0.2.tgz",
       "integrity": "sha512-s5kLOcnH0XqDO+FvuaLX8DDjZ18CGFk7VygH40QoKPUQhW4e2rvM0rwUq0t8IQDOwYSeLK01U90OjzBTme2QqA==",
       "dev": true
-    },
-    "klaw": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/klaw/-/klaw-1.3.1.tgz",
-      "integrity": "sha1-QIhDO0azsbolnXh4XY6W9zugJDk=",
-      "dev": true,
-      "requires": {
-        "graceful-fs": "^4.1.9"
-      }
     },
     "lazy.js": {
       "version": "0.4.3",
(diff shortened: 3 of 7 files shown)


modules/admin-ui/src/test/resources/karma.conf.js
```diff
@@ -52,14 +52,41 @@ module.exports = function (config) {
...
+            // here you can edit the list of browsers used by karma
+            postDetection: function(availableBrowsers) {
+                /* Karma configuration with custom launchers
+                customLaunchers: {
+                    IE9: {
+                        base: 'IE',
+                        'x-ua-compatible': 'IE=EmulateIE9'
+                    }
+                }
+                */
```
lkiesow: This can be removed, right?

modules/admin-ui/src/test/resources/karma.conf.js
```diff
@@ -52,14 +52,41 @@ module.exports = function (config) {
 
         autoWatch : true,
 
-        frameworks: ['jasmine'],
+        frameworks: ['detectBrowsers', 'jasmine'],
 
-        browsers : ['PhantomJS'],
+        detectBrowsers: {
+            enabled: true,
+            usePhantomJS: true,
```
lkiesow: Since [PhantomJS is long since abanfoned](https://github.com/ariya/phantomjs/issues/15344), do we want to continue using it? For a lot of the other test dependencies (ffprobe, sox, synfig, …) we just don't run the tests if the dependency is not present and we could do that here as well. That would also make the build system a bit simpler since we don't need to check for special cases and would make things way easier on systems like Alpine Linux.

That said, I don't have a very stromg opinion here and if you want to keep it, I won't start a fight ;-P

modules/admin-ui/src/test/resources/karma.conf.js
gregorydlogan: Sure, I left it in so that if we need to define one at some point we would have a template, but I can remove it.

modules/admin-ui/src/test/resources/karma.conf.js
gregorydlogan: What does removing Phantom do for Alpine?

I don't really have a strong feeling about this either way, but here's my reasoning:

If, somehow, Opencast is being built on a system without a browser installed we want to be able to at least have some sense of whether things are working as expected.  That's where Phantom comes in.  It's not ideal, which is also why it's excluded if anything better is available, but it allows automated build images built without a browser to continue working.

That being said, it's not that big a deal to include Firefox as a build dep.

lkiesow: > What does removing Phantom do for Alpine?

By default, Opencast does not build on systems like Apine since Opencast downloads the PhantomJS binary which requires the glibc. Since most Linux distributions use glibc, Opencast builds on most Linux systems. Unfortunately, most are not all. Apline, for instance, uses the [musl libc](https://musl.libc.org/) since it is quite a bit smaller. So, building Opencast on such systems is very hard due to parts like PhantomJS.

> If, somehow, Opencast is being built on a system without a browser installed we want to be able to at least have some sense of whether things are working as expected.

The recent discussion about build systems at the Dev'n'Hack camp made me realize that these tests mainly have two target audiences:

1. **CI systems** which, by now, are certainly our main defense against accidentally pushing broken stuff into the project – and the CI systems only work because of the tests. Everyone has to pass these unless it's local code which not a lot of people still have and which is their own responsibility.

2. The team of **core developers** needs a local testing environment since they are making a lot of changes to the project.

In both cases, having a browser isn't hard. In case of a new developer not having a browser on their machine – unlikely but could happen – this does not really matter since worst case, patches would fail in the CI.

Finally, looking at users, no tests do not matter at all for several reasons:

- Hardly any users build their own code today. They use packages, containers, tarballs, … meaning that they do not run these tests.
- People do not build on the systems they deploy on. We discourage that… which makes sense.
- If tests for release versions fail for someone, we usually advise them to disable tests. After all, we know that the tests have passed on our systems so it's likely a glitch in the tests.

That is why I tend to go with simpler is better, preventing us from having to find out that things failed for someone just because they did not have r.g. Firefox installed and we didn't realize that this failed on a certain system since we are testing with browsers.

Reasons explained, I still leave the final to you ;-)

modules/admin-ui/src/test/resources/karma.conf.js
lkiesow: Let's remove it. I'm pretty sure we don't want a custom IE launcher at any point ;-)

modules/admin-ui/src/test/resources/karma.conf.js
gregorydlogan: Fair enough.  This will probably break Buildbot, and possibly Travis.  Let's see.

lkiesow:
Some findings and thoughts:

- This might or might not [cause build failures on systems with additional browsers like a Mac with Safari](https://github.com/litixsoft/karma-detect-browsers/issues/34). I need to test this tomorrow when I have access to a Mac.
- This seems to cause the build to hang without any warning or error if there is no browser installed. That's not good since it is really confusing and could cause larger problems in automated build environments. Better would be:
    - Best case, the tests would be skipped if no browser is present
    - Acceptable would be an instant build error stating that you need a browser
    - Still acceptable (I know I complained about this before but it's still better than the current case) would be the PhantomJS fallback.

gregorydlogan:
@lkiesow Can't seem to stop the tests at the point where the browsers are being detected, so instead it just fails the build instantly.

rrolf:
So it unfortunately does not build on a Mac:

[ERROR] node-pre-gyp WARN Using request for node-pre-gyp https download 
...
[ERROR] Browserslist: caniuse-lite is outdated. Please run next command `npm update`
...
[INFO] 06 03 2020 17:28:08.195:ERROR [launcher]: Cannot load browser "Safari": it is not registered! Perhaps you are missing some plugin?
[INFO] 06 03 2020 17:28:08.195:ERROR [karma-server]: [Error: Found 1 load error
[INFO]   at Server.webServer.listen (/Users/rrolf/opencast/browser-mac/modules/admin-ui/node_modules/karma/lib/server.js:182:27)
[INFO]   at Object.onceWrapper (events.js:286:20)
[INFO]   at Server.emit (events.js:203:15)
[INFO]   at emitListeningNT (net.js:1313:10)
[INFO]   at process._tickCallback (internal/process/next_tick.js:63:19)
[INFO] ]
[INFO] 06 03 2020 17:28:08.205:INFO [launcher]: Starting browser ChromeHeadless
[INFO] 06 03 2020 17:28:08.548:INFO [launcher]: Starting browser FirefoxHeadless
[INFO] 
[INFO] 
 

Full stack-trace:
https://pastebin.com/rjjPPE18

gregorydlogan:
Ok, so I see two options here:

- Whitelist the browsers we have plugins set up for
- Attempt to broadly support more browsers (ie, add the Safari plugin, etc)

I don't like the idea of whitelisting, but broadly supporting browsers sounds to me like a recipe for uncommon-but-detected browsers causing errors...

gregorydlogan:
In the meantime, @rrolf this should work with the latest changes.

lkiesow:
But then, this will still break if you have Opera installed. Adding just one more browser isn't really a solution. The issue I linked suggested a solution. If that does not work, you could add a whitelist which would make sure to only use supported browsers.

gregorydlogan:
As suggested, removed the plugin list.

gregorydlogan:
I get that this isn't high priority, but this causes every one of my builds to fail, at least until I remember to add the OPENSSL_CONF env variable.  Can we get this in?

modules/admin-ui/Gruntfile.js
```diff
@@ -1,6 +1,10 @@
 // Generated on 2016-03-07 using generator-angular 0.15.1
 'use strict';
 
+// Prevents Debian 10 (at least) from erroring out if PhantomJS is being used
+// Appears to be related to something in the config file, but we can't exactly ship our own config
+process.env.OPENSSL_CONF='';
+
```
KatrinIhler: We're no longer using PhantomJS, right? So this can be removed?

docs/guides/developer/docs/installation/source-linux.md
```diff
@@ -25,6 +25,7 @@ Required:
     ffmpeg >= 3.2.4
     maven >= 3.1
     python >= 2.6, < 3.0
+    firefox/chrome/some other major browser
```
KatrinIhler, line 28: So we don't know exactly which browsers work and which don't, right?

modules/admin-ui/package.json
```diff
@@ -36,11 +36,11 @@
     "karma": "^4.1.0",
     "karma-chrome-launcher": "^2.2.0",
     "karma-coverage": "^1.1.2",
+    "karma-detect-browsers": "^2.0",
     "karma-firefox-launcher": "^1.1.0",
     "karma-jasmine": "^1.1.2",
     "karma-ng-html2js-preprocessor": "^1.0.0",
-    "karma-phantomjs-launcher": "^1.0.4",
-    "phantomjs-prebuilt": "^2.1.16",
+    "karma-safari-launcher": "^1.0.0",
```
KatrinIhler, line 43: Since the plugin list was removed, does this still have to be here?

modules/admin-ui/package-lock.json
```diff
@@ -4506,20 +4493,10 @@
...
-      "requires": {
-        "lodash": "^4.0.1",
-        "phantomjs-prebuilt": "^2.1.7"
-      }
-    },
-    "kew": {
-      "version": "0.7.0",
-      "resolved": "https://registry.npmjs.org/kew/-/kew-0.7.0.tgz",
-      "integrity": "sha1-edk9LTM2PW/dKXCzNdkUGtWR15s=",
+    "karma-safari-launcher": {
```
KatrinIhler, line 4496: Same question as below.

KatrinIhler (changes requested):
So, I'm a bit confused. For which browsers do the tests run now? Does Karma decide this itself?

I currently have Firefox and Chromium installed and it works for me. I'm also making Julian test it. ;) But we currently don't have a mac at home... maybe I can bother someone else as well.

docs/guides/developer/docs/installation/source-linux.md
gregorydlogan: It tries to detect all of them, but I'm hedging in the docs to keep maintenance low :)

modules/admin-ui/package.json
gregorydlogan: Yes, otherwise safari doesn't work.

This autodetection has a nasty side effect: if it can't find anything then it just sits forever. We need to cast a wide net, otherwise someone with a default Mac install will just be stuck.

gregorydlogan:
Karma decides for itself based on what it detects.  This should in theory work as long as you have a major browser installed. If you only have something we don't search for then things won't work :)

modules/admin-ui/Gruntfile.js
gregorydlogan: Fixed.

modules/admin-ui/package.json
KatrinIhler: But this will result in a build error in the end, right? Did I understand that correctly? Because being stuck forever would suck.

gregorydlogan: Yep, line 78 in karma.conf throws an error if nothing is detected.

KatrinIhler (approved):
Alright, this looks fine now, let's merge!

KatrinIhler:
@gregorydlogan Seems like buildbot is trying to run the admin ui tests with firefox, but it's not working.  Is this something you can fix?

gregorydlogan:
I have this working locally, but it's repeatedly breaking on ci.  Travis seems happy.  Let's try merging this and if things break horribly then we can roll it back out.  Not quite sure what to make of the issues on ci :/

KatrinIhler:
@gregorydlogan Alright, let's do this!

Issue #1429, opened 2020-02-24
Modify build-up & tear-downs in ServiceRegistryJpaImplTest so Travis CI won't choke


karendolan:
**Describe the bug**
When Travis CI attempts to run resource intensive junit tests, such as in ServiceRegistryJpaImplTest, the tests get Assert errors when they should not get Assert errors. This only happens in the Travis CI environment and appears to be a limitation in the Travis CI environment.

**To Reproduce**
1. Run the ServiceRegistryJpaImplTest junit tests a few hundred times in a non-Travis CI system
2. Never see an assert error from ServiceRegistryJpaImplTest
3. Run the ServiceRegistryJpaImplTest junit tests 3-4 times in Travis CI
4. See Assert error in various places
(NOTE: the "ERROR ...Job with Id 1 cannot be deleted"  is normal and not part of the issue with the junits. That line can be ignored)

**Environment Information**
 - Opencast github Travis CI

Possible solution:  The fails appear to be race conditions associated to the build up and tearing down of test resources around each test in that class. It might help if the core test resources are built at test class load and kept in tact between each test, only cleaned between tests, but not torn down between test. Tear down all resources when the class unloads. 

Related issue: #1281

karendolan:
Also related to #1068

karendolan:
FYI -  My concern is that valid junit tests are being actively removed from the code base because Travis CI occasionally says something like  1 + 1 ==  3 , not 2. The "temporarily" ignored tests are being stripped from the code base to "clean up the code". Yet, the code that the junits were testing is still left in the code, just in an untested, unprotected, state.

lkiesow:
Sorry to tell you that these issues are not at all limited to builds on Travis CI. I've been regularly restarting builds on Greg's buildbot cluster since they fail with the same issue and I have seen this locally as well though far less often. Generally, it seems like the more powerful the machine you are building on, the less likely it seems that the issue will pop up.

I agree that having tests is nice but if they happen to fail on a regular basis – which the ones we got rid of clearly did – they are hurting the project far more than the tests help. I have seen people ignore failed builds a few times already since “it must be a Travis/buildbot issue” while it was not and thus broken code got in. That's not good. It makes the tests untrustworthy and additionally causes work since (mostly Greg or I) need to restart the tests.

All in all, I'm happy about more tests. But they must be rock-solid and not cause any false build errors.

gregorydlogan:
[...]

PR #1430, opened 2020-02-24
#1429 rewrite ServiceRegistryJpaImplTest to reduce TravisCI failures

karendolan:
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

This pull reduces setup and teardown overhead of the ServiceRegistryJpaImplTest junit tests. The goal is to remove the Travis CI proclivity towards random assert errors that cannot be duplicated in any other testing environment.

### DIFF
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -416,7 +416,7 @@ public class ServiceRegistryJpaImplTest {
   private void assertHostloads(Job j, Float a, Float b, Float c) throws Exception {
     // launch the  dispatcher and wait a little longer for dispatch to complete before getting job
     launchDispatcherOnce(false);
-    Thread.sleep(3 * DISPATCH_START_DELAY);
+    Thread.sleep(JOB_BARRIER_TIMEOUT);
     Job k = serviceRegistryJpaImpl.getJob(j.getId());
     k.setStatus(Status.RUNNING);
     serviceRegistryJpaImpl.updateJob(k);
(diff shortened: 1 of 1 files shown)


karendolan:
Hi @lkiesow and @gregorydlogan , so far, I only got error on the second run: [ERROR]   DownloadDistributionServiceImplTest.testRetractByDistributed:288 » NullPointer. 
Do you have a list of test classes that often have issues succeeding in Travis CI? I'm going to keep doing Travis CI build runs to see what errors arise.

Hmm... The  POSIX pull was concurrently running the same mvn build test when that error above occurred... I vaguely recall from last summer that the Travis CI errors are more likely if there is a lot of builds going on at once.

karendolan:
Travis CI build run results...
1. no errors(no concurrent builds)
2. The DownloadDistributionServiceImplTest.testRetractByDistributed (concurrent pull was also in build state). No error on the reinstated ServiceRegistryJpaImplTest test.
3. no errors (no concurrent builds)
4. ...

karendolan:
Ok, the issue was a junit test race condition with the normal job dispatch loop. This pull is updated with a launchDispatchOnce() to ensure the junit tests can control when the dispatch happens so the test can properly assert what expects the state to be after a dispatch (and not the odd state of jobs when they are in the middle of a dispatch process).

Because disptach execution is now controlled in the tests. The big sleep delay in the fairness test can be safely reduced to speed the tests.

karendolan:
"java.lang.AssertionError: Host load is incorrect, should be 1.000000, is 0.000000" will make that sleeper 3x vs 2x, also it doesn't need the hearbeat to be launched with the 1 threaded scheduler....

modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
```diff
@@ -252,69 +334,89 @@ public void testCancelUndispatchablesOrphanedByActivatingNode() throws Exception
...
+    } finally {
+      logger.debug("end testIgnoreHostsInPriorityList");
     }
   }
 
   private void assertHostloads(Job j, Float a, Float b, Float c) throws Exception {
-    Thread.sleep(1100); //1100 is 100ms more than the minimum job dispatch interval.  Setting this lower causes race conditions.
+    // launch the  dispatcher and wait a little longer for dispatch to complete before getting job
+    launchDispatcherOnce(false);
+    Thread.sleep(3 * DISPATCH_START_DELAY);
```
karendolan: Maybe safer and more consistent to use JOB_BARRIER_TIMEOUT here... it'll make the test take longer, but less likely for TravisCI to choke.

Issue #1432, opened 2020-02-25
Series ACLs not propagating to individual events


NgoniChoga:
**Describe the bug**
When using the default option in the configuration (https://docs.opencast.org/r/8.x/admin/#configuration/acl/#merge-mode-roles), pressing the "Update event permissions" in the UI is not updating the ACLs of individual events in the series. 

**To Reproduce**
1. Go to a series with events.
2. Go to series details and update the series Access Policy.
3. Click the "Update event permissions" after changing the series Access Policy.
4. Go to event details for any event in the series.
5. Look at event Access policy.

**Environment Information**
 - Browser: Chrome
 - Opencast version with minor release: 8.x
 - Also tested in stable.opencast.org

PR #1433, opened 2020-02-25
Series ACLs not propagating to individual events

NgoniChoga:
### Series ACLs not propagating to individual events

* [ ] [Issue](https://github.com/opencast/opencast/issues/1432)

### DIFF
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
@@ -276,7 +276,7 @@ public class SeriesServiceImpl extends AbstractIndexProducer implements SeriesSe
     if (accessControl == null) {
       throw new IllegalArgumentException("ACL parameter must not be null");
     }
-    if (needsUpdate(seriesId, accessControl)) {
+    if (needsUpdate(seriesId, accessControl) || overrideEpisodeAcl) {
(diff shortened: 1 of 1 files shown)


lkiesow:
Can you describe in a few more details what the problem is and what you are trying to solve with this patch? The default behavior is that a series ACL will not override an episode ACL if one is set.

NgoniChoga:
The issue is described [here](https://github.com/opencast/opencast/issues/1432) in a bit more detail. The default behaviour is, if the "Update event permissions" button is pressed, the series ACL should propagate down to the individual events.

smarquard:
The issue with the "Update event permissions" button is that if you're in the Series ACL tab and update an ACL rule, it gets saved and updated immediately. Then when you click the "Update event permissions" button, needsUpdate(seriesId, accessControl) returns false because the ACL hasn't changed between the POST and what's stored, so nothing happens. Thus clicking the button has no effect.

lkiesow:
Looks good and fixes the problem. Thanks for the patch, @NgoniChoga

Rillke:
@lkiesow @smarquard Can this bugfix be merged into legacy r/7.x, too, please?

gregorydlogan:
@Rillke that question would bet better addressed to @KatrinIhler or @mliradelc as release managers most likely.

KatrinIhler:
@Rillke Since it was just one line, I backported this to r/7.x directly instead of opening another pullrequest. But be aware that Opencast 8 is getting much more attention currently, so it might be awhile until the next 7 Release.

PR #1441, opened 2020-03-03
Remove databasemigration for Opencast 7

CGreweling:
The Database Migration Module was build for the new Scheduler for migration from Opencast 6 to 7.
There is no need for this for Opencast > 7.
It is even not started if the Opencastversion is  not 7.


gregorydlogan:
@CGreweling shouldn't this go into r/8.x?

CGreweling:
@gregorydlogan as i wrote this pr, i was not sure if removing a whole module would be to much for a legacy release. 
But as i read your comment here, Yes! ... :)  I will do a rebase to 8.x.

PR #1454, opened 2020-03-12
Fix HLS on iOS

### LINKED ISSUE #1452: HLS on iOS broken
**Describe the bug**
iOS devices don't play HLS videos with PaellaPlayer. It doesn't matter which browser you use. When you click a video the load animation starts but never ends.

**To Reproduce**

1. Configure HLS as the first streaming protocol.
2. Either:
a. Take an actual iOS device.
b. Use any other OS and switch the browser's user agent to iOS:
`Mozilla/5.0 (iPhone; CPU iPhone OS 13_2_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.3 Mobile/15E148 Safari/604.1`
3. Try to watch a video.
4. Inspect it and notice that you see something like this:
```
<video id=...>
    <source src=...m3u8" type="undefined">
</video>
```

**Environment Information**
 - OS: iOS 10, iOS 13
 - Browser: Safari, Chrome, Firefox
 - Opencast: 8.1

miesgre:
### Your pull request should…

* [x] have a concise title
* [x] closes #1452
* [ ] pass automated testing

### DIFF
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -134,7 +134,7 @@ class OpencastToPaellaConverter {
       isLiveStream: (track.live === true)
     };
 
-    if(track.video) {
+    if(track.mimetype != 'audio/m4a') {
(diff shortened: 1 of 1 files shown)


miesgre:
@t-schroeder I do not have an iOS device to test it, but this should fix the bug. Can you test it please?

PR #1473, opened 2020-03-18
Move OC Studio configuration to `etc/ui-config` and update Studio

LukasKalbertodt:
That way, the configuration can still be modified. Before, it was hardcoded on the module's jar file.

This only makes sense after https://github.com/elan-ev/opencast-studio/pull/488 has landed and the Studio version included here is updated. That's why this is a draft PR.

CC @lkiesow @darolfes


gregorydlogan:
@LukasKalbertodt Looks like the blocker here is done.  This ready to go?

LukasKalbertodt:
@gregorydlogan Thanks for the ping. I just discussed with Lars and we would like to wait with merging this PR until after a new Studio version was released and included in Opencast. That's because without a new Studio version, this PR breaks things. And since we want to update the Studio version on (probably) Tuesday anyway, we figured we can wait until then instead of updating it twice.

I hope you're OK with that!

gregorydlogan:
Fine with me, will wait for this to not be a draft anymore.

PR #1476, opened 2020-03-18
Fix Studio Redirect discarding GET Parameters

### LINKED ISSUE #1470: GET parameters to Opencast Studio get discarded due to redirect
**To Reproduce**
1. Log into stable.opencast.org
2. Visit URL `https://stable.opencast.org/studio?foo=bar`
3. Observe how you get redirected to `https://stable.opencast.org/studio` (i.e. GET parameters get discarded)

This is a 302 redirect; not a JS redirect. So it's an issue with the Studio module. The same happens with `/studio/?foo=bar`. 

However, it does *not really* happen with `/studio/index.html?foo=bar`. In the latter case, the GET parameters disappear, but this time due to a bug in Studio itself (which is [already fixed](https://github.com/elan-ev/opencast-studio/pull/482)).

**Environment Information**
 - Firefox & Chrome
 - Opencast 8.2

CC @lkiesow @darolfes

darolfes:
This Pullrequest fixes #1470.

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/rest/StaticResource.java
+++ b/modules/common/src/main/java/org/opencastproject/rest/StaticResource.java
@@ -124,6 +124,8 @@ public class StaticResource extends HttpServlet {
       } else {
         redirectPath = alias + "/" + welcomeFile;
       }
+      String queryString = req.getQueryString();
+      redirectPath += queryString != null ? "?" + queryString : "";
       logger.debug("redirecting {} to {}", path, redirectPath);
       resp.sendRedirect(redirectPath);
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
I tested locally and it works for me! Thanks!

For anyone testing as well: as described in #1470, there is still a bug in the studio code that is used in Opencast right now. That code also removes GET parameters. However, the difference is that with this PR, Studio at least once sees the GET parameters and you can see them in the web console output under "current settings".

lkiesow (approved):
Looks reasonable

PR #1479, opened 2020-03-20
Fix Theodul quality selection

mtneug:
The Theodul quality selection has a couple of bugs that are fixed in this PR, but let's describe our usage scenario first. We create delivery versions based on the source resolution. This differs between `presenter` and `presentation` (for CA recordings always; for LMS uploads possibly). In a normal recording we thus have:

* `presenter`: `720p-quality`, `480p-quality`
* `presentation`: `1080p-quality`, `720p-quality`, `480p-quality`

Next to some generic bugs, Theodul does not correctly handle this particular case well. When selecting `1080p` as quality, I would expect the player chooses `720p` for `presenter` (the closest available resolution). In addition, Theodul can handle non-standard tags only partially i.e not during initialization where it defaults to `medium-quality` (or `low-quality` for mobile).

In individual commits, I patched the following:

* General bug in track filtering (string argument instead of an array)
* Fix for non-standard quality tags
* Fix for setting the initial quality twice and thus possible the incorrect one
* Fix for the described scenario above by considering the resolution and choosing the closest one to the selected quality level

Front-end development is not my area so be extra careful when reviewing my JavaScript skills 😄.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/engage-theodul-core/src/main/resources/ui/js/engage/models/engage.js
+++ b/modules/engage-theodul-core/src/main/resources/ui/js/engage/models/engage.js
@@ -58,24 +58,8 @@ define(['jquery', 'backbone', 'bowser', 'basil', 'engage/models/pluginInfo', 'en
         }
         // set quality
         if (this.get('urlParameters').quality) {
-          switch (this.get('urlParameters').quality) {
-          case 'low':
-            this.set('quality', 'low');
-            break;
-          case 'medium':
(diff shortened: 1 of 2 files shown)


mtneug:
I could have introduced a bug for videos with no quality tags. I'm investigating.

mtneug:
Should be fixed.

PR #1485, opened 2020-03-24
Custom Roles for LTI users

CGreweling:
With this patch it is possible to translate a LTI-Role to several
Opencast Roles.
With this it is possible for example, to show adminui parts like the editor to
LTI users.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
+++ b/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
@@ -113,10 +113,10 @@ public class LtiLaunchAuthenticationHandler implements OAuthAuthenticationHandle
   /** The user reference provider */
   private UserReferenceProvider userReferenceProvider = null;
 
-  /** If the Role exists add CUSTOM_ROLES to this user **/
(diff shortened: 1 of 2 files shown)


modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
```diff
@@ -113,6 +113,16 @@
   /** The user reference provider */
   private UserReferenceProvider userReferenceProvider = null;
 
+  /** If the Role exists add CUSTOM_ROLES to this user **/
+  private static final String CUSTOM_ROLE_NAME = "lti.custom_role_name";
+
+  /** If the Role exists add CUSTOM_ROLES to this user **/
+  private static final String CUSTOM_ROLES = "lti.custom_roles";
+
+  private String customRoleName = "";
```
gregorydlogan, line 122: These two new variables need comments.

CGreweling: Right good catch. Hope this is better.

PR #1488, opened 2020-03-24
Make encoding profiles support odd widths/heights

### LINKED ISSUE #1486: Error when encoding video with odd pixel count in height
**Describe the bug**
When I upload a video with an odd height, e.g. 1063px it is successfully encoded to 360p, 480p and 720p, but encoding to 1080p fails with:

> \[libx264 @ 0x64b4140\] height not divisible by 2 (1890x1063)
> Error initializing output stream 0:0 -- Error while opening encoder for output stream #0:0 - maybe incorrect parameters such as bit_rate, rate, width or heigh

This causes the entire workflow to fail.

**To Reproduce**
Upload a video with an odd height using the default "schedule-and-upload" workflow and select at least one encoding profile that would produce a video with height larger than that of the original video.

**Environment Information**
 - OS: CentOS 7
 - Opencast: 8.1, 8.2

t-schroeder:
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

Fixes #1486

### DIFF
--- a/etc/encoding/adaptive-streaming-movies.properties
+++ b/etc/encoding/adaptive-streaming-movies.properties
@@ -29,7 +29,7 @@ profile.adaptive-2160p.http.suffix = -preview.mp4
 profile.adaptive-2160p.http.ffmpeg.command = -i #{in.video.path} \
   -c:v libx264 -crf 23 -maxrate 10000k -bufsize 10000k -profile:v high -level 4.0 -pix_fmt yuv420p \
   -x264opts keyint=25:min-keyint=25:no-scenecut -movflags +faststart \
-  -filter:v scale='-2:min(ih\\,2160)',fps=fps=25 \
+  -filter:v crop='floor(iw/2)*2:floor(ih/2)*2',scale='-2:min(ih\\,2160)',fps=fps=25 \
   -c:a aac -b:a 96k -ac 1 \
   -f mp4 #{out.dir}/#{out.name}#{out.suffix}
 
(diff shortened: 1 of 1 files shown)


etc/encoding/adaptive-streaming-movies.properties
```diff
@@ -40,7 +40,7 @@ profile.adaptive-1080p.http.suffix = -preview.mp4
 profile.adaptive-1080p.http.ffmpeg.command = -i #{in.video.path} \
   -c:v libx264 -crf 23 -maxrate 4000k -bufsize 8000k -profile:v high -level 4.0 -pix_fmt yuv420p \
   -x264opts keyint=25:min-keyint=25:no-scenecut -movflags +faststart \
-  -filter:v scale='-2:min(ih\\,1080)',fps=fps=25 \
+  -filter:v crop='floor(iw/2)*2:floor(ih/2)*2',scale='-2:min(ih\\,1080)',fps=fps=25 \
```
lkiesow, line 43: Wouldn't `force_original_aspect_ratio=decrease` and `force_divisible_by=2` make way more sense than chaining two scale filters? I know that the latter option is somewhat new but the release of 9.x is also quite some time off.

Essentially:
```suggestion
  -filter:v scale=1920:1080:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
```

t-schroeder: (Your edit should probably say `scale=`, not `crop=`.)

I tried to use `force_divisible_by` but on my CentOS 7 machine neither the package version of ffmpeg nor the statically linked one from https://johnvansickle.com/ffmpeg/ support it. So I couldn't test that yet.

Technically the new profile uses a crop and a scale filter, not two scale filters. I compared execution times with vs without the crop on a video with even width/height. Videos with odd widht or height can't be encoded without the the original profile, so I had nothing to compare it against.

Without crop:
```
real    2m3.469s
user    3m38.199s
sys     0m1.443s
```

With crop:
```
real    1m56.107s
user    3m34.956s
sys     0m1.481s
```

So the crop operation doesn't seem to make a difference performance-wise. Probably because nothing is cropped when using a video with an even width and height.

lkiesow: Just wondering: Do you want to get this into 8.x (maybe even 8.3)? If yes, then rebase this against `r/8.x` and we can leave it like this. The feature I mentioned is too new for that and this could be counted as a bug fix. Note that 8.3 should be released this morning (CET).

---

If this is targeted at `develop`:

It's not in the stable but it is in the master build of johnvansickle.com/ffmpeg and it's also in the release which is in Opencast's package repo.

With the target release for this probably ending up to be in December, that's more than enough time for that feature to settle and I just think that that filter chain makes more sense ;-)

t-schroeder: Ok. I rebased this PR against `r/8.x` and made a new one with your suggested change for `develop`: #1490

PR #1489, opened 2020-03-25
Workaround early lti session timeout

darolfes:
This increases the lti session timeout so Studio can upload even after a longer period of recording.


lkiesow (approved):

lkiesow:
This generally updates the session timeout which was previously set to 10min. If you are active, you might not notice this and if you clicked remember-me (activated by default) you never notice this but if you are logged in using  LTI, that's not an option. Being logged out after 4h seemed a reasonable choice to us.

PR #1490, opened 2020-03-26
Make encoding profiles support odd widths and heights develop

### LINKED ISSUE #1486: Error when encoding video with odd pixel count in height

t-schroeder:
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

Fixes #1486

### DIFF
--- a/etc/encoding/adaptive-streaming-movies.properties
+++ b/etc/encoding/adaptive-streaming-movies.properties
@@ -29,7 +29,7 @@ profile.adaptive-2160p.http.suffix = -preview.mp4
 profile.adaptive-2160p.http.ffmpeg.command = -i #{in.video.path} \
   -c:v libx264 -crf 23 -maxrate 10000k -bufsize 10000k -profile:v high -level 4.0 -pix_fmt yuv420p \
   -x264opts keyint=25:min-keyint=25:no-scenecut -movflags +faststart \
-  -filter:v scale='-2:min(ih\\,2160)',fps=fps=25 \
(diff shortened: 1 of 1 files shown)


t-schroeder:
@lkiesow I noticed the encoding profiles `profile.studio.adaptive-parallel.http.ffmpeg.*` use these scale filters which do handle odd widths/heights as well:

```
scale=w=1920:h=trunc(1920/dar/2)*2,setsar=1,fps=25
```

Should I change this to use the new scale filter options as well?

lkiesow:
Too late. But it made it into 8.3
Thanks for the patch.
For 9.x, we have time for improved profiles.
And there will be a workflow update anyway

lkiesow:
ups, missed that this is an additional pull request.
Everything that is in `r/x` will go into `develop` anyway.
So, this wouldn't have been necessary.
But it also does not hurt

PR #1506, opened 2020-04-06
Amend Studio documentation with some details about the LTI config

JulianKniephoff:
This documents one way to give the LTI users access to Opencast Studio, that I just had to research myself. :wink: 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/docs/guides/admin/docs/modules/studio.md
+++ b/docs/guides/admin/docs/modules/studio.md
@@ -14,6 +14,23 @@ The path `/studio` is accessible by users with the role `ROLE_ADMIN` or `ROLE_ST
 
 The preferred way to let your users access Studio is via LTI. Remember to configure your LTI users to have the role `ROLE_STUDIO` so that they can access Studio and all APIs used by Studio.
 
+This can be done, for example, by configuring the `LtiLaunchAuthenticationHandler` in `etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg`. At the bottom of this file you see the two options `lti.custom_role_name` and `lti.custom_roles`. The former should be set to an LTI role that is supposed to have access to Studio, and the latter is a comma-separated list of Opencast roles that users with this LTI role should get. This list has to at least include `ROLE_STUDIO`.
+
+A minimal configuration giving LTI-`Instructor`-s access to Opencast Studio could look like this:
+
+```
+lti.custom_role_name=Instructor
+lti.custom_roles=ROLE_STUDIO
+```
+
+You will also have to enable persisting LTI users in the same file to avoid access management related failures during ingests initiated by them:
+
+```
+lti.create_jpa_user_reference=true
+```
+
+Note that with this configuration, every LTI user is persisted in the database with their ID and all the roles assigned to them! They are stored in the user references table (`oc_user_ref`) as opposed to the "normal" users table (`oc_user`), though.
+
 
 ## Configuring Studio
 


lkiesow:
Wouldn't it make more sense to link the LTI documentation about this?
(I'm assuming that one exists)

JulianKniephoff:
Oh it exists, but it does not explain anything useful regarding Opencast Studio. The only mention of the necessary file is [this](https://docs.opencast.org/r/8.x/admin/#modules/ltimodule/#configure-lti).

docs/guides/admin/docs/modules/studio.md
```diff
@@ -14,6 +14,17 @@ The path `/studio` is accessible by users with the role `ROLE_ADMIN` or `ROLE_ST
 
 The preferred way to let your users access Studio is via LTI. Remember to configure your LTI users to have the role `ROLE_STUDIO` so that they can access Studio and all APIs used by Studio.
 
+This can be done for example by configuring the `LtiLaunchAuthenticationHandler` in `etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg`. At the bottom of this file you see the two options `lti.custom_role_name` and `lti.custom_roles`. The former should be set to an LTI role that you want to give access to Studio to, and the latter is a comma-separated list of Opencast roles that users with this LTI role should get. This list has to at least include `ROLE_STUDIO`.
```
LukasKalbertodt: I would add commas around "for example" in the first sentence. 

"access to Studio to" -> "Studio access to"?

docs/guides/admin/docs/modules/studio.md
```diff
@@ -14,6 +14,17 @@ The path `/studio` is accessible by users with the role `ROLE_ADMIN` or `ROLE_ST
...
+
+A minimal configuration giving LTI-`Instructor`-s access to Opencast Studio could look like this:
+
+```
+lti.custom_role_name=Instructor
+lti.custom_roles=ROLE_STUDIO
+```
+
+Note that with this configuration, every LTI user is persisted in the database with their ID and all the roles assigned to them!
```
LukasKalbertodt: Maybe mention that those users are not persisted in the normal user table, but merely in the user references table?

LukasKalbertodt (commented):
I can't say anything about the correctness of this documentation. Apart from that and the inline comments, this looks good to me.

docs/guides/admin/docs/modules/studio.md
JulianKniephoff: > I would add commas around "for example" in the first sentence.
> 
Apparently you are right. x)

> "access to Studio to" -> "Studio access to"?

I don't like that, but reading it again, I don't like mine anymore either. :sweat_smile: 

How about "an LTI role that you want to be able to access Studio"?

docs/guides/admin/docs/modules/studio.md
JulianKniephoff: Mentioned.

docs/guides/admin/docs/modules/studio.md
LukasKalbertodt: "an LTI role that you want to be able to access Studio" -> mhh, not completely happy about that. "a role that is supposed to have access to studio"? But yeah, nit picking at this point. Don't block on this mini concern.

JulianKniephoff: I can live with that. Changed.

JulianKniephoff:
Just to be clear: Since this is a doc change, I could've just merged this, but @LukasKalbertodt wanted to at least see the text before, thus the PR.

However, now that we're here, maybe someone could check the technical validity of what I wrote here, since I'm not 100% sure about it, either. So maybe someone, maybe even from ELAN (again, this being a docs change), could take a look. Maybe @wsmirnow whom I originally got the intel from. (Thanks and sorry Waldemar. :D)

LukasKalbertodt:
Note for people porting this PR to `develop` at some point: the part about `lti.create_jpa_user_reference=true` can be removed as this is the default value for 9.x

gregorydlogan:
Since this is going into develop, perhaps we should make the change @LukasKalbertodt outlined, and then verify that it's right (@wsmirnow?).  It looks and sounds right to me, but I don't have any kind of LTI bits to test against...

wsmirnow:
Why is this PR against `develop`? This would also apply to `r/8.x`.

docs/guides/admin/docs/modules/studio.md
```diff
@@ -14,6 +14,23 @@ The path `/studio` is accessible by users with the role `ROLE_ADMIN` or `ROLE_ST
...
+This can be done, for example, by configuring the `LtiLaunchAuthenticationHandler` in `etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg`. At the bottom of this file you see the two options `lti.custom_role_name` and `lti.custom_roles`. The former should be set to an LTI role that is supposed to have access to Studio, and the latter is a comma-separated list of Opencast roles that users with this LTI role should get. This list has to at least include `ROLE_STUDIO`.
+
+A minimal configuration giving LTI-`Instructor`-s access to Opencast Studio could look like this:
+
+```
+lti.custom_role_name=Instructor
+lti.custom_roles=ROLE_STUDIO
+```
+
+You will also have to enable persisting LTI users in the same file to avoid access management related failures during ingests initiated by them:
```
wsmirnow, line 26: …during processing ingested videos
```suggestion
You will also have to enable persisting LTI users in the same file to avoid access management related failures during processing ingested video files:
```

wsmirnow (changes requested):

docs/guides/admin/docs/modules/studio.md
wsmirnow: The ingest will work fine without persisting LTI user reference, because the LTI session exists and the `ROLE_STUDIO` is applied. But to start the workflow, the user will be queried outside of the LTI Session by Opencast. This will fail, if the user (reference) isn't found.

wsmirnow: The rest is fine for me.

gregorydlogan:
If this applies to 8.x let's target it there then :)

docs/guides/admin/docs/modules/studio.md
lkiesow: This section is simply not true. You don't have to enable LTI user references. You just have to make sure a user with the stored identifier exists which sounds similar but is something entirely different and Opencast has several different means of archiving that. For instance, you could use the Sakai user provider, or the Moodle user provider, or the Canvas user provider, or LDAP, … More than one of these ways are in active use.

I would drop this whole pull request and even go so far as to drop

> The preferred way to let your users access Studio is via LTI. Remember to configure your LTI users to have the role `ROLE_STUDIO` so that they can access Studio and all APIs used by Studio.

…and replace it with something like…

> Make sure users have the necessary access rights for the ingest service, e.g. by giving them `ROLE_STUDIO` and don't use temporary users since otherwise, workflows may fail due to missing user privileges. When using LTI, you could do this by enabling user persistence in `etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg`.

JulianKniephoff:
Since there seem to be quite a few open questions and/or strong opinions that **I personally** don't feel qualified to answer/evaluate, I'll actually follow @lkiesow's suggestion and just drop this PR. If anyone wants to pick this up, feel free to use what is already here, of course. :smile:

PR #1507, opened 2020-04-06
Add a list of download buttons for each episode to the LTI series list

Arnei:
Download button dropdown on the LTI series list

Adds a dropdown to each episode in a series. Clicking an entry in the dropdown 
will download the corresponding video file. @wsmirnow 
This pull request is intended as a quick change to r/8.x and will very likely be overwritten in r/9.x.

Expected result:
DropDownOpen

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/etc/ui-config/mh_default_org/ltitools/series.css
+++ b/etc/ui-config/mh_default_org/ltitools/series.css
@@ -23,6 +23,13 @@ div.loading {
   color: #666;
 }
 
+div.episode-container {
+  display: flex;
+  background-color: #eee;
+  align-items: center | stretch;
+  margin: 15px;
+}
+
 a.episode {
   display: block;
   text-decoration: none;
@@ -30,6 +37,7 @@ a.episode {
   background-color: #eee;
   padding: 5px;
   margin: 15px;
+  flex-grow: 1;
 }
 
 a.episode:hover {
@@ -45,32 +53,59 @@ a.episode div.preview {
   width: 200px;
 }
 
-div.episode-container {
-  display: flex;
-  background-color: #eee;
-  padding: 5px;
-  margin: 15px;
-}
-
-div.download {
-  display: flex;
-  background-color: red;
-  flex-direction: column;
-  justify-content: center;
-  margin-left: auto;
-  margin-right: 5px;
-}
-
-div.download button.downloadButton {
-  width: 100%;
-  margin-top: 1px;
-  margin-bottom: 5px;
-}
-
-
 a.episode div.desc h2 {
   margin: 0;
   margin-bottom: 5px;
+}
+
+/* Style The Dropdown Button */
+.dropbtn {
+  background-color: #eee;
+  color: black;
+  padding: 16px;
+  font-size: 16px;
+  border: 2px solid black;
+  border-radius: 2px;
+  cursor: pointer;
+  margin-top: 5px;
+  margin-right: 5px;
+}
+
+/* The container <div> - needed to position the dropdown content */
+.dropdown {
+  position: relative;
+  display: inline-block;
+}
+
+/* Dropdown Content (Hidden by Default) */
+.dropdown-content {
+  display: none;
+  position: absolute;
+  background-color: #f9f9f9;
+  min-width: 160px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+}
+
+/* Links inside the dropdown */
+.dropdown-content a {
+  color: black;
+  padding: 12px 16px;
+  text-decoration: none;
+  display: block;
+}
+
+/* Change color of dropdown links on hover */
+.dropdown-content a:hover {background-color: #f1f1f1}
+
+/* Show the dropdown menu on hover */
+.dropdown:hover .dropdown-content {
+  display: block;
+}
+
+/* Change the background color of the dropdown button when the dropdown content is shown */
+.dropdown:hover .dropbtn {
+  background-color: #fafafa;
 }
 
 footer  {
--- a/etc/ui-config/mh_default_org/ltitools/series.css
+++ b/etc/ui-config/mh_default_org/ltitools/series.css
@@ -59,7 +59,7 @@ a.episode div.desc h2 {
 }
 
 /* Style The Dropdown Button */
-.dropbtn {
+.drop-button {
   background-color: #eee;
   color: black;
   padding: 16px;
@@ -67,10 +67,7 @@ a.episode div.desc h2 {
   border: 2px solid black;
   border-radius: 2px;
   cursor: pointer;
-  margin-top: 10px;
-  margin-right: 10px;
-  margin-left: 10px;
-  margin-bottom: 10px;
+  margin: 10px;
 }
 
 /* The container <div> - needed to position the dropdown content */
@@ -98,7 +95,9 @@ a.episode div.desc h2 {
 }
 
 /* Change color of dropdown links on hover */
-.dropdown-content a:hover {background-color: #f1f1f1}
+.dropdown-content a:hover {
+  background-color: #f1f1f1
+}
 
 /* Show the dropdown menu on hover */
 .dropdown:hover .dropdown-content {
--- a/etc/ui-config/mh_default_org/ltitools/series.css
+++ b/etc/ui-config/mh_default_org/ltitools/series.css
@@ -25,8 +25,8 @@ div.loading {
 
 div.episode-container {
   display: flex;
+  flex-wrap: wrap;
   background-color: #eee;
-  align-items: center | stretch;
   margin: 15px;
 }
 
@@ -69,6 +69,8 @@ a.episode div.desc h2 {
   cursor: pointer;
   margin-top: 5px;
   margin-right: 5px;
+  margin-left: 10px;
+  margin-bottom: 10px;
 }
 
 /* The container <div> - needed to position the dropdown content */
--- a/etc/ui-config/mh_default_org/ltitools/series.css
+++ b/etc/ui-config/mh_default_org/ltitools/series.css
@@ -67,8 +67,8 @@ a.episode div.desc h2 {
   border: 2px solid black;
   border-radius: 2px;
   cursor: pointer;
-  margin-top: 5px;
-  margin-right: 5px;
+  margin-top: 10px;
+  margin-right: 10px;
   margin-left: 10px;
   margin-bottom: 10px;
 }
@@ -84,7 +84,7 @@ a.episode div.desc h2 {
   display: none;
   position: absolute;
   background-color: #f9f9f9;
-  min-width: 160px;
+  min-width: 200px;
   box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
   z-index: 1;
 }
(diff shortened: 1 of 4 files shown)


lkiesow:
Note that this likely conflicts with #1289

mtneug:
Not all institutions want to make it so easy to download the video files. Can this be made configurable?

gregorydlogan:
This is also pointed at `develop`, which would be `9.x`.  You likely want to point this at `r/8.x`.

modules/runtime-info-ui/src/main/resources/ui/index.html
```diff
@@ -79,7 +79,7 @@ <h2>Community Resources</h2>
           <ul class="links">
             <li><a href="https://groups.google.com/a/opencast.org/forum/#!myforums">Mailing Lists (Google
                 Groups)</a></li>
-            <li><a href="http://webchat.freenode.net/?channels=opencast">IRC-Channel (Webinterface)</a></li>
+            <li><a href="http://webchat.freenode.net/?channels=opencast">IRC-Channel (Webinterfafa face)</a></li>
```
gregorydlogan: This should be reverted

lkiesow:
> This is also pointed at `develop`, which would be `9.x`. You likely want to point this at `r/8.x`.

Not sure if we really want to bring this in before #1289? This should at least go by the release managers (@karendolan, @rrolf). We are not only putting a feature in 8.x this time but deliberately breaking another pull request (we do that, correct? I'm actually just assuming that.). But then… it's not my pull request so I care less then I would otherwise ;-P

gregorydlogan:
IMO #1289 takes priority since its larger and been around too long. I have it built, was going to look at it in the next couple of days.

I was only commenting on the branch target since the description mentions 8.x. I don't really care where this goes, or if it goes in at all :)

etc/ui-config/mh_default_org/ltitools/series.css
```diff
@@ -50,6 +58,58 @@ a.episode div.desc h2 {
...
+  color: black;
+  padding: 16px;
+  font-size: 16px;
+  border: 2px solid black;
+  border-radius: 2px;
+  cursor: pointer;
+  margin-top: 10px;
+  margin-right: 10px;
+  margin-left: 10px;
+  margin-bottom: 10px;
```
LukasKalbertodt: Why not just `margin: 10px`?

etc/ui-config/mh_default_org/ltitools/series.css
```diff
@@ -50,6 +58,58 @@ a.episode div.desc h2 {
   margin-bottom: 5px;
 }
 
+/* Style The Dropdown Button */
+.dropbtn {
```
LukasKalbertodt: I personally don't think that shortening words like "button" to "btn" is useful at all. Why not simply `drop-button`?

etc/ui-config/mh_default_org/ltitools/series.css
```diff
@@ -50,6 +58,58 @@ a.episode div.desc h2 {
...
+/* Links inside the dropdown */
+.dropdown-content a {
+  color: black;
+  padding: 12px 16px;
+  text-decoration: none;
+  display: block;
+}
+
+/* Change color of dropdown links on hover */
+.dropdown-content a:hover {background-color: #f1f1f1}
```
LukasKalbertodt: This should be 3 lines. I don't think we gain anything by squashing this into one.

modules/lti/src/main/resources/tools/series/index.html
```diff
@@ -23,14 +23,27 @@
...
-    <div class=desc>
-      <h2>{{title}}</h2>
-      {{i18ncreator}}<br />
-      {{created}}
+  <div class=episode-container>
+    <a class=episode href={{player}}>
+      <div class=preview><img alt=Preview src={{image}}></div>
+      <div class=desc>
+        <h2>{{title}}</h2>
+        {{i18ncreator}}<br>
```
LukasKalbertodt: Here the `<br />` changed to `<br>`. I don't think it's problematic, I'm just not sure if it's intended.

modules/lti/src/main/resources/tools/series/index.html
```diff
@@ -23,14 +23,27 @@
...
+        {{created}}
+      </div>
+    </a>
+    <div class=dropdown>
+      <button class=dropbtn>Direct Download</button>
+      <div class=dropdown-content>
+      {{#download}}
+        <a href={{path-to-file}} download="">
```
LukasKalbertodt: Isn't there some useful filename we could add to `download=`? If not, I think you can simply remove the `=""` part (`<a ... download>`)

modules/lti/src/main/resources/tools/series/index.html
```diff
@@ -23,14 +23,27 @@
...
+        {{{download-button-name}}}
```
LukasKalbertodt: I think this line should be indented one additional level?

modules/lti/src/main/resources/tools/series/index.html
```diff
@@ -23,14 +23,27 @@
...
+        </a>
+        <br>
```
LukasKalbertodt: Is this line break really necessary? The `<a>` elements are already `display: block`, right? If it's about spacing between the links, I'd probably rather solve this via `margin-bottom` on the `<a>` elements. 

Additionally, it seems that this is creating an empty space after the last element which is undesirable. You can fix this in CSS like this:

```css
.dropdown-content a:not(:last-child) {
    margin-bottom: ...;
}
```

modules/lti/src/main/resources/tools/series/series.js
```diff
@@ -68,6 +68,12 @@ function getSeries() {
   return '';
 }
 
+function capitalize(s) {
+  if (typeof s !== 'string')
+    return '';
```
LukasKalbertodt: I'm not sure about the current policy but I remember the community discussing this on the mailing list. I *think* the outcome was that we want to enforce adding braces `{ }` in these places. I personally absolutely think that braces here should be required.

modules/lti/src/main/resources/tools/series/series.js
```diff
@@ -115,6 +121,32 @@ function loadPage(page) {
...
       }
 
+      // Download buttons
+      // Create a button for each track that contains a link to a mp4 or webm
+      var buttonData = [];
+      var tracks = episode.mediapackage.media.track;
+      tracks = Array.isArray(tracks) ? tracks : [tracks];
+      for (var k = 0; k < tracks.length; k++)
+      {
+        if( tracks[k].url.endsWith('mp4') || tracks[0].url.endsWith('webm')) {
```
LukasKalbertodt: Spacing is inconsistent. From looking at the rest of this file, this should probably be

```js
        if (tracks[k].url.endsWith('mp4') || tracks[0].url.endsWith('webm')) {
```

Also: why only mp4 and webm? 

Also also: `tracks[0]`... is that intended or should it be `tracks[k]`? If the latter, the `for` loop should be

```js
for (const track of tracks) {
```

modules/lti/src/main/resources/tools/series/series.js
```diff
@@ -115,6 +121,32 @@ function loadPage(page) {
...
+
+          // Calculate filesize
+          var fileSize = (tracks[k].video.bitrate / 8 / 1000 / 1000)      // Converts to byte, kilobyte, megabyte /s
+            * (tracks[k].video.framecount / tracks[k].video.framerate);
```
LukasKalbertodt: Is there no way to get access to the actual file size? This seems like a ... hacky approximation?

modules/lti/src/main/resources/tools/series/series.js
```diff
@@ -115,6 +121,32 @@ function loadPage(page) {
...
+
+          // Collect output to mustache
+          buttonData.push ({
+            'download-button-name' : capitalize( tracks[k].type.split('/')[0] )
+            + '<br /> ('
+            + tracks[k].video.resolution
+            + ', '
+            + fileSize.toFixed(2) + 'MB'
+            + ')',
```
LukasKalbertodt: I don't think this is how it should be done. Rather than adding HTML in this string and then using `{{{ x }}}` to not escape the string, you should use two strings (`download-button-name` and `download-description` or something like that). Using `{{{ }}}` in Mustache should be avoided if possible.

LukasKalbertodt (changes requested):
I added a couple of line comments. I tested locally and it seems to work for me. If you have any questions about my comments, just ask!

modules/lti/src/main/resources/tools/series/series.js
lkiesow: This also disregards the audio.
If the size is not in the JSON, it probably makes sense to not display it.
And I know that I originally suggested to show the size :D

modules/lti/src/main/resources/tools/series/series.js
Arnei: Checking whether the url of a track points to a mp4 or webm file was recommended as a way to check whether the track is for a downloadable video or not.

And indeed, it should be `tracks[k]`.

modules/lti/src/main/resources/tools/series/series.js
Arnei: It may seem that way because it is ^^'. The JSON containing the track information does not specify a file size, so I thought I'd use the other information at hand. Not sure if I can get the file size from the url itself, something to look into.

modules/lti/src/main/resources/tools/series/index.html
Arnei: I suppose you could name the file similarly to the download button name, but I do not know whether this is desirable or not.

modules/lti/src/main/resources/tools/series/series.js
LukasKalbertodt: If webm and mp4 is the only thing produces by Opencast, then that's fine. I am just not sure about that.

modules/lti/src/main/resources/tools/series/index.html
LukasKalbertodt: It looks like the URL contains a somewhat useful filename. So I guess we can just use `download` without `=""`.

modules/lti/src/main/resources/tools/series/series.js
Arnei: So apparently you can just _use a HTTP HEAD request and then look at Content-Length response header. ( [https://bitexperts.com/Question/Detail/3316/determine-file-size-in-javascript-without-downloading-a-file](url) )

Unsure whether it is worth it to send a bunch of web requests, since the hacky approximation seems to arrive at fairly similar numbers.

LukasKalbertodt: > If the size is not in the JSON, it probably makes sense to not display it.

I guess I agree with that. Doesn't seem like a critical feature, so just remove the file size?

Arnei: Huh, I figured the bitrate would include the audio. 
But if showing file size is not that important anyway, I'll cut it.

Arnei:
The pull request is too specific to be merged into any main branch.

PR #1539, opened 2020-04-23
Fix Karma Safari test in mac OS 

mliradelc:
Follow up of #1426, This fix changes the karma plugin to make the tests in safari. The issue with the former plugin was because of the security policies safari asks to open the file `redirect.html`, if you don't make this action the build fails.

Screenshot 2020-04-23 at 08 43 27

Using the `karma-safari-applescript-launcher` allows to comunicate with safari directly. The first time it will only ask in OSX if allows the terminal to run it. later builds it will not ask again.

### Your pull request should…

* [x] have a concise title

* [x] pass automated testing

### DIFF
--- a/modules/admin-ui/package.json
+++ b/modules/admin-ui/package.json
@@ -40,7 +40,7 @@
     "karma-firefox-launcher": "^1.1.0",
     "karma-jasmine": "^1.1.2",
     "karma-ng-html2js-preprocessor": "^1.0.0",
-    "karma-safari-launcher": "^1.0.0",
+    "karma-safari-applescript-launcher": "^0.1.0",
     "request": "^2.88.0",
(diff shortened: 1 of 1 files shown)


KatrinIhler:
This is working for me, but I mean, I don't have a Mac, so... I bugged some of my colleagues, so I'll wait what they have to say.

gregorydlogan:
Works for me on my Debian box.  Still worried about the buildbot stuff, but it works here so I don't think this is a blocker.

KatrinIhler (approved):
Rüdiger tested this with his Mac and says it works for him.
I'm still not happy that you have to confirm when you run the tests for the first time, but I guess it's better than before...

gregorydlogan:
There's also not a ton we can do about it unfortunately.

PR #1542, opened 2020-04-23
Allow Admin UI Users to Access /

lkiesow:
Users with limited access rights but ROLE_ADMIN_UI can log in and
access the admin interface but will get an error when they get back to
the main domain since they have no access while already being logged in.

This patch allows anonymous access to / which will then redirect users
to the admin interface where they get access if they have the

This fixes #1345

### Your pull request should…

* [x] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] be against the correct branch (features can only go into develop)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated testing
* [x] have a clean commit history
* [x] have proper commit messages (title and body) for all commits
* [x] have appropriate tags applied

PR #1543, opened 2020-04-23
403 Logout Option

lkiesow:
In case you are logged in but do not have the necessary roles to access
e.g. the admin interface, users will end up on an error page with no
sane way to log out.

This can be pretty problematic if users are unknowingly authenticated
via LTI and then miss the necessary groups.

This patch adds a very simple user interface to convey the necessary
information to users and provide them with an easy way to log out again.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -16,6 +16,8 @@
   <sec:http create-session="ifRequired" servlet-api-provision="true" realm="Opencast"
     entry-point-ref="opencastEntryPoint">
 
+    <sec:access-denied-handler error-page="/403.html"/>
+
     <!-- ################ -->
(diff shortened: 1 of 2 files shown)


mliradelc:
Can it be backported back to 8.0?

Two months ago looking I was looking at how to do this, looking where was the original 403 page to modify and the solution it was to add that line to the sec file + an Html file in a module.

mliradelc:
Works perfect!! great job
Screenshot 2020-04-24 at 11 50 18

lkiesow:
@mliradelc, this and #1542 is actually meant to go into `r/8.x`. I've updated the target branch. Sorry. can you review this and #1542 and merge it if it works as expected?

PR #1131, opened 2019-10-08
Issue1123 TEMP FIX for Paella Player Build error

karendolan:
* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] be against the correct branch (features can only go into develop)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated testing
* [x] have a clean commit history
* [x] have proper commit messages (title and body) for all commits
* [x] have appropriate tags applied

This pull is TEMP FIX for the CI error with engage-paella-player. If the Paella player repo accepts the npm-shrinkwrap.json the this pull can be adjusted to just keep the npm ci changes.
Ref: https://github.com/polimediaupv/paella/pull/448
Closes Issue https://github.com/opencast/opencast/issues/1123
Originating trigger: https://github.com/olivernn/lunr.js/issues/420

PR #1547, opened 2020-04-24
Download paella source code from github instead of using npm + paella update to 6.4.3

miesgre:
This PR changes the way opencast downloads the paella source from npm to download the tar from github.

The problem with npm is that the package-lock.json file is not fetched and during build other versions can be used that can cause problems (see #1131, polimediaupv/paella#534)

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/engage-paella-player/gulpfile.js
+++ b/modules/engage-paella-player/gulpfile.js
@@ -29,7 +29,7 @@ var source = require('vinyl-source-stream');
 var gunzip = require('gulp-gunzip');
 var untar = require('gulp-untar');
 
-var PAELLA_VERSION = '6.3.2';
+var PAELLA_VERSION = '6.4.3';
 
 var buildPath = 'target/gulp',
     paellaSrc = 'src/main/paella-opencast',
--- a/modules/engage-paella-player/gulpfile.js
+++ b/modules/engage-paella-player/gulpfile.js
@@ -24,24 +24,34 @@
 'use strict';
 var gulp = require('gulp');
 var spawn = require('child_process').spawn;
-var mergeStream = require('merge-stream');
+var request = require('request');
+var source = require('vinyl-source-stream');
+var gunzip = require('gulp-gunzip');
+var untar = require('gulp-untar');
 
+var PAELLA_VERSION = '6.3.2';
 
 var buildPath = 'target/gulp',
-    paellaSrc = 'src/main/paella-opencast';
+    paellaSrc = 'src/main/paella-opencast',
+    paellaBuildPath = buildPath + '/paella-' + PAELLA_VERSION;
 
 
-gulp.task('paella-opencast:prepare:source', function(){
-  var s1 = gulp.src('node_modules/paellaplayer/**').pipe(gulp.dest(buildPath + '/paella'));
-  var s2 = gulp.src(paellaSrc + '/plugins/**').pipe(gulp.dest(buildPath + '/paella/plugins'));
-
-  return mergeStream(s1,s2);
+gulp.task('paella-opencast:download:paella', function(){
+  return request('https://github.com/polimediaupv/paella/archive/' + PAELLA_VERSION + '.tar.gz')
+  .pipe(source(PAELLA_VERSION + '.tar.gz'))
+  .pipe(gunzip())
+  .pipe(untar())
+  .pipe(gulp.dest(buildPath));
 });
 
 
+gulp.task('paella-opencast:prepare:source', gulp.series('paella-opencast:download:paella', function(){
+  return gulp.src(paellaSrc + '/plugins/**').pipe(gulp.dest(paellaBuildPath + '/plugins'));
+}));
+
 
 gulp.task('paella-opencast:prepare', gulp.series('paella-opencast:prepare:source', function(cb){
-  var cmd_npm = spawn('npm', ['install'], {cwd: buildPath + '/paella', stdio: 'inherit'});
+  var cmd_npm = spawn('npm', ['ci'], {cwd: paellaBuildPath });
   cmd_npm.on('close', function (code) {
     cb(code);
   });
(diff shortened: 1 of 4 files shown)


miesgre:
@karendolan Can you review this?

karendolan:
@miesgre why is npm-shrinkwrap bad to publish to npm?

miesgre:
> @miesgre why is npm-shrinkwrap bad to publish to npm?

No. But paella source code is not in npm repository. We were using npm but just to download the code from github as you can see is package.json
```
"paellaplayer": "github:polimediaupv/paella#6.3.2"
```

modules/engage-paella-player/gulpfile.js
```diff
@@ -24,39 +24,49 @@
 'use strict';
 var gulp = require('gulp');
 var spawn = require('child_process').spawn;
-var mergeStream = require('merge-stream');
+var request = require('request');
+var source = require('vinyl-source-stream');
+var gunzip = require('gulp-gunzip');
+var untar = require('gulp-untar');
 
+var PAELLA_VERSION = '6.3.2';
```
karendolan: @miesgre  This pull pushes the version of JQuery, and other dependencies, back for people who have already built on this version of Paella. I recommend making this pull part part of the next Paella upgrade. Then the dependencies are as current, or more current, then the  current version of the adopter's Paella dependencies.
https://github.com/polimediaupv/paella/pull/533

miesgre: In this PR I only change the way paella source code is downloaded
The idea is to update paella player to 6.4.2 once this PR was accepted, but I can include it in this commit if you want.

karendolan: I prefer the Paella update to coincide with this change because the update will be associated to a snapshot with the latest dependencies. Then there would not be the chance that an upgrade Opencast would give users older Paella dependencies.

miesgre: I updated this PR and to update paella to latest version 6.4.3 with newer dependencies (including jQuery 3.5.1)

karendolan:
Hi @miesgre,

An artifact of this pull is that the paella_player.js classes are compiled differently. It might not matter. But, the old style is easier to read and follow. Here's an example segment of before and after from the start of the Paella Mouse Manager in the compiled paella_player.js:
```
 // Paella Mouse Manager
 ///////////////////////////////////////////////////////
-(() => {
-  class MouseManager {
-    get targetObject() {
-      return this._targetObject;
-    }
-
-    set targetObject(t) {
-      this._targetObject = t;
-    }
-
-    constructor() {
-      paella.events.bind('mouseup', event => this.up(event));
-      paella.events.bind('mousemove', event => this.move(event));
-      paella.events.bind('mouseover', event => this.over(event));
-    }
+(function () {
+  var MouseManager = /*#__PURE__*/function () {
+    _createClass(MouseManager, [{
+      key: "targetObject",
+      get: function get() {
+        return this._targetObject;
+      },
+      set: function set(t) {
+        this._targetObject = t;
+      }
+    }]);
 
-    down(targetObject, event) {
-      this.targetObject = targetObject;
+    function MouseManager() {
+      var _this = this;
 
-      if (this.targetObject && this.targetObject.down) {
-        this.targetObject.down(event, event.pageX, event.pageY);
-        event.cancelBubble = true;
-      }
+      _classCallCheck(this, MouseManager);
 
-      return false;
+      paella.events.bind('mouseup', function (event) {
+        return _this.up(event);
+      });
+      paella.events.bind('mousemove', function (event) {
+        return _this.move(event);
+      });
+      paella.events.bind('mouseover', function (event) {
+        return _this.over(event);
+      });
     }
```

miesgre:
> Hi @miesgre,
> 
> An artifact of this pull is that the paella_player.js classes are compiled differently. It might not matter. But, the old style is easier to read and follow. 

I'm not sure, but maybe a babel update produce this code? Anyway, the player now uses paella_player_es2015.js and this file is more readable

https://github.com/opencast/opencast/blob/afca8595fe6d45844f026fff44b600c086c454c6/modules/engage-paella-player/src/main/paella-opencast/ui/watch.html#L14

karendolan:
> > Hi @miesgre,
> > An artifact of this pull is that the paella_player.js classes are compiled differently. 
> 
> I'm not sure, but maybe a babel update produce this code? Anyway, the player now uses paella_player_es2015.js and this file is more readable
> 

@miesgre  how do you feel about updating most/all of Paella 6.4.3's package-lock to be recent non-major release versions of library dependencies? The concern is that people have been building Paella from  this Opencast module with the more recent version's of libraries for  a while now, and that going backwards in the libraries might cause issues. [EDIT] Going  forward might cause more issues too, since the updated lock file would not have been fully tested in standalone. 

The change from existing build libs to the new build libs:
[EDIT] I just ran  "npm run build" with those newer babel's in the standalone 6.4.3 and got the same paella_player.js build, so the babel version wasn't the issue with the build from Opencast. 

```
     "@babel/generator": {
+      "version": "7.8.7",
-      "version": "7.9.6",

     "@babel/helper-function-name": {
+      "version": "7.8.3",
-      "version": "7.9.5",

     "@babel/helper-replace-supers": {
+      "version": "7.8.6",
-      "version": "7.9.6",

     "@babel/traverse": {
+      "version": "7.8.6",
-      "version": "7.9.6",

     "@babel/types": {
+      "version": "7.8.7",
-      "version": "7.9.6",

```

miesgre:
> @miesgre how do you feel about updating most/all of Paella 6.4.3's package-lock to be recent non-major release versions of library dependencies?

@karendolan I don't understand you. I updated the paella dependencies to latest version in 6.4.3. see https://github.com/polimediaupv/paella/commit/27a041897e1a4fd07e4addc49e5f2aafd3c2315a

Do you want to downgrade the build deps to the same as 6.4.2? I'm not sure what are you proposing

karendolan:
>[...] @karendolan I don't understand you.[...]

Hi @miesgre, I had been confused why the paella_player.js and the paella_player_es2015.js are the same in past Opencast builds, but they are different in this pull. I thought it might be a lib version issue, but, MYSTERY SOLVED!  It's the hidden ".babelrc" file that does not get copied out of the node_modules to the Opencast build directory prior to  building the Paella player. Adding the file to the Paella build results in the old-style babel paella_player. Removing it gives the new style.  The new style  paella_player.js  has been part of the Opencast Paella player from the beginning because the ".babelrc" file has never been copied into the player build directory (!). 

In summary, it's good that the player files now reference paella_player_es2015.js, because Opencast has been using that build in its babel'less build of paella_player.js since it was first introduced in Paella. Pulling down  the tar.gz version of Paella ensures that it was built with  the ".babelrc"  file, so Opencast will always get the correct intention of the build from now on. No more worry about hidden build dependency files.

I'm satisfied that this pull resolves the Paella build issues from the past and provides consistency for the future.

Issue #1541, opened 2020-04-23
Adding write access parameter to Events endpoint and Series endpoint


otti-ssystems:
There have to be the possibility to pass a write access parameter to the Events endpoint and the Series Endpoint.
If passed, the Events endpoint and the Series Endpoint should only response results with permission "write access".

**Environment Information**
- opencast 8.3

Ticket is related to :
https://github.com/opencast/opencast/pull/1063
and
https://github.com/opencast/opencast/pull/209

**Implementation**

EventsEndpoint.java

- Adding the parameter "onlyWithWriteAccess" to the Events Endpoint getEvents(...)
- when parameter is passed as true we are querying with Permissions.Action.WRITE

_**Attribut**_
isRequired = false 
description = "Whether only to get the events to which we have write access." 
type = Type.BOOLEAN

- additional function deserializeMetadataList for deserialize and mapping the catalogs json to a list

further :

SeriesEndpoint.java

- Adding the parameter "onlyWithWriteAccess" to the Series Endpoint getSeriesList(...)
- when parameter is passed as true we are querying with Permissions.Action.WRITE

Merge Request in Progress...

I think you will categorize as feature?
So request should go into develop or should i offer the request to 8.3 / 8.x?

Greetz

Otti

gregorydlogan:
This, to me, sounds like a minor feature.  @KatrinIhler, @mliradelc thoughts?

KatrinIhler:
What endpoints are we talking about here? If it's the api, this would probably also require a new api version.
Either way, yeah, I think that's a feature and should go into develop.

otti-ssystems:
Hi.
Yes, we are influencing the endpoints of the external api.
I already created the pull request locally, but i need a branch with the reference number to commit.

Can you please tell me how to do it, then you can have a look at the request.

THX a lot

gregorydlogan:
@otti-ssystems if you have the code, follow along at https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork.  File it against `opencast/opencast`'s `develop` branch.

otti-ssystems:
@gregorydlogan thank you. Thats what i was looking for

LukasKalbertodt:
~~#1541~~ #1548 seems to close this issue?

[...]

PR #1548, opened 2020-04-28
#1541 adding write access parameter to events and series endpoint

otti-ssystems:
## Pull Request for adding wrtite access paramater to Series and Events endpoint in the external api.

Issue : https://github.com/opencast/opencast/issues/1541

We are adding the **write access** params in the Method : 

`getEvents(...)`

and 

` getSeriesList(...)`

if parameter is not set the response will include all events/series, even these with no write access.


mliradelc:
This commit is incomplete!
- Needs to update the documentation about the API
- The API version should be updated like in  #1615, in this case, you can use that the bump to version 1.5.0 is linked to the commit #1615 to avoid repeat code :smile:

PR #1550, opened 2020-04-29
Be more lenient towards service failures

lkiesow:
With much more user content like we see right now, it's much more likely
that a failed operation does not mean a broken service or a broken
service configuration.

That is why this pull request increases the number of times a job may
fail on a service before that service is sent into an error state.

Note that services in a warning state will be sanitized automatically
when operations succeed on the same service. Hence, if a service fails
three times but then succeeds, the service will be in a normal state.

The amount of allowed failures chosen is just a guess. A single
operation failing can cause multiple jobs fail simultaneously. That's
why 10 did not seem too high.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
+++ b/etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
@@ -1,7 +1,7 @@
 # Configuration for the service registry
 
 # Number of failed jobs on a service before to set it in error state
-max.attempts=1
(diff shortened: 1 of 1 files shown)


gregorydlogan:
This seems reasonable to me, especially in light of it being easily configurable.  I'd say we should document this somewhere, but I imagine it's already documented - just not something you run into when you're doing the naive initial setup.

gregorydlogan:
Ignore the buildbot error, that's buildbot bug.

PR #1568, opened 2020-05-06
Fix URL to Wowza documentation

EmTeedee:
URL to Wowza documentation about adaptive streaming has changed.


lkiesow:
Thanks for the patch

PR #1578, opened 2020-05-12
Add cutting to the default Studio workflow

JulianKniephoff:
The next Opencast Studio release, slated to be included in Opencast 8.4, will likely support a very minimal editing interface. See elan-ev/opencast-studio#454. This needs to be supported by the Opencast workflow used to process the Studio uploads, though, since the plan at the moment is to just submit a SMIL file together with the video, as opposed to cutting the video client side.

This PR adapts the standard studio workflow accordingly.

### DIFF
--- a/etc/workflows/partial-transcode-studio-tracks.xml
+++ b/etc/workflows/partial-transcode-studio-tracks.xml
@@ -17,6 +17,20 @@
       </configurations>
     </operation>
 
+    <!-- Cut the videos -->
+
+    <operation id="editor"
+      exception-handler-workflow="partial-error"
+      description="Cut the recording according to the edit decision list">
+      <configurations>
+        <configuration key="source-flavors">*/source+transcoded</configuration>
+        <configuration key="smil-flavors">smil/cutting</configuration>
+        <configuration key="target-smil-flavor">smil/cutting</configuration>
+        <configuration key="target-flavor-subtype">source+trimmed</configuration>
+        <configuration key="skip-if-not-trimmed">true</configuration>
+      </configurations>
+    </operation>
+
     <operation
       id="tag"
       exception-handler-workflow="partial-error"
@@ -31,9 +45,9 @@
     <operation
       id="tag"
       exception-handler-workflow="partial-error"
-      description="Tagging transcoded presenter video as source and for archive">
+      description="Tagging transcoded and trimmed presenter video as source and for archive">
       <configurations>
-        <configuration key="source-flavors">presenter/source+transcoded</configuration>
+        <configuration key="source-flavors">presenter/source+trimmed</configuration>
         <configuration key="target-flavor">presenter/source</configuration>
         <configuration key="target-tags">archive</configuration>
       </configurations>
@@ -53,9 +67,9 @@
     <operation
       id="tag"
       exception-handler-workflow="partial-error"
-      description="Tagging transcoded presentation video as source and for archive">
+      description="Tagging transcoded and trimmed presentation video as source and for archive">
       <configurations>
-        <configuration key="source-flavors">presentation/source+transcoded</configuration>
+        <configuration key="source-flavors">presentation/source+trimmed</configuration>
         <configuration key="target-flavor">presentation/source</configuration>
         <configuration key="target-tags">archive</configuration>
       </configurations>


etc/workflows/studio-upload.xml
```diff
@@ -202,14 +202,37 @@
...
+
+    <operation id="editor"
+      exception-handler-workflow="partial-error"
+      description="Cut the recording according to the edit decision list">
+      <configurations>
+        <configuration key="source-flavors">*/prepared</configuration>
+        <configuration key="smil-flavors">smil/tmp</configuration>
+        <configuration key="target-smil-flavor">smil/tmp</configuration>
+        <configuration key="target-flavor-subtype">trimmed</configuration>
+        <configuration key="skip-if-not-trimmed">true</configuration>
```
wsmirnow: In case the smil file contains only one segment over the whole video length, the operation will be "skipped" but it does not copy the source track and set the target flavor/tags on it. The next operation will fail (or be skipped), because there are no tracks with `*/trimmed` flavor. Be carefully with this option.

etc/workflows/studio-upload.xml
```diff
@@ -142,7 +142,7 @@
         exception-handler-workflow="partial-error"
         description="Tagging source material for archival">
       <configurations>
-        <configuration key="source-flavors">dublincore/*,security/*,*/source</configuration>
+        <configuration key="source-flavors">dublincore/*,security/*,smil/*,*/source</configuration>
```
wsmirnow: In case of Studio video upload, you want to trimm the video and archive the trimmed result, without the uploaded smil file. Also the `editor`-trimm operation should be moved to the `partial-transcode-studio-tracks` workflow, where it can do the magic stuff before the first snapshot operation.

wsmirnow (changes requested):

etc/workflows/studio-upload.xml
JulianKniephoff: I tested it and it worked. And from my understanding of the code of the `editor`-WOH, that is expected: Before the operation is actually skipped, the media package elements should be cloned, and the clones be given the correct flavors.

However, I admit that I **was** unsure about this option, so if anyone has evidence for it not working or for my understanding to be wrong, it would be greatly appreciated. :smile:

etc/workflows/studio-upload.xml
JulianKniephoff: I'm not sure I follow. If I put the `editor` operation in `partial-transcode-studio-tracks`, that happens before `prepare-av`, but it operates on the media generated by `prepare-av`. I mean, maybe it doesn't have to, I'm not sure, but that's the way it is in the standard workflow, where I stole this from. :sweat_smile: 

And what's the reasoning behind archiving the trimmed version without the cutting information as opposed to all of the source material, like the standard workflow does? :thinking:

wsmirnow: The idea behind the `partial-transcode-studio-tracks` is not to loose the source material at any state of the workflow but archive the transcoded version of the tracks without waste space with original video files. 

1. mark source material for archive (but do not take a snapshot)
1. transcode all tracks
1. switch the flavors and archive tags
1. take snapshot and go on with regular processing

If any operation from 1. or 2. fail, the error handler workflow will archive the original source files and you will be save. 
Ideally you want to replace the transcode operation (2.) with the editor. If it does not work (for reasons), put the editor operation in between 2. and 3. Do not forget to remove the archive tag from smil file before taking snapshot (4.)

As other workflows expects the `*/prepared` version, this must be generated and archived. Otherwise the following workflows will fail. In most cases all tracks does have an audio stream and the prepared version is same as source. AssetManager keep track of it and make a hardlink on filesystem.

JulianKniephoff: Okay, I think I understand now. However, I'm not sure I agree with your premise, that we actually want to archive the trimmed version, and just discard the original videos.

That's not what we do for other ways of getting data into Opencast, either. When a CA uploads a video and someone cuts it, we then also don't just throw away the uncut version.

On the other hand when Studio finally learns to cut the video itself client side, we would also not have the uncut version in Opencast, right? Or would we still upload the uncut version plus a SMIL file and only use the client side trimming for the download version? :thinking: 

Can we get a third (or even more for all I care :wink:) opinion on this?

wsmirnow: The initial trimming is initiated by the user, that creates the video file with studio and decide to trim it. It has nothing to do with the editor in the admin UI. We should respect it.

LukasKalbertodt: Right now, I would slightly tend to keep the original file as well (i.e. uncut), as this is unlikely to be significantly larger than the cut one (Studio users will very likely only remove tens of seconds at most). However, as mentioned, as soon as Studio actually does the cutting itself, Opencast does not have the uncut version anyway. And I would certainly oppose deliberately uploading the uncut version from Studio. 

Soooooo... "I don't care" is what I want to say :D I think it doesn't make a large difference and the situation will change in some time anyway.

JulianKniephoff:
@wsmirnow I hope I addressed all of your main concerns. Re: Avoiding extra encoding: I don't think it's possible with `editor`, because the encoding profile it uses is global config, not an option to the workflow operation. **Maybe** it could work with `process-smil`, but tbh I would tend to not go through the whole testing cycle with that operation again, only for this hopefully rather temporary solution.

etc/workflows/studio-upload.xml
gregorydlogan: Personally I'm very much in the camp where the complete set of original files is kept, and trimmed copies are treated as any other intermediate file.  We want a single source of truth, and discarding the original after trimming it seems wrong to me.

That being said, those saying that the user trimmed the video prior to ingesting it, and that we should respect that do have a point.  What if the user is trimming off something that's not allowed to be kept (say, privacy reasons, or licensing).  Do our adopters run a risk by potentially having that footage in their archive, whether they know they have it or not?  If the trim in Studio was truly done client side (this would be a nice feature), then we wouldn't get the data at all - and we would be ok with that.  So why do we feel like we get to keep it when the presentr thinks that we're not even getting it in the first place.  Normally I would say that the presenter is out of luck, but *especially* in light of the current COVID-19 situation we have to assume that these recordings may be being done in suboptimal places.  Kids could be run through the frame, someone could spill their coffee, all kinds of things which would not otherwise occur in a classroom or office setting.

As much as I prefer in all other cases to keep the original source media and create an intermediate trimmed copy, I think _in this case_ we need to respect user choice.  Perhaps we should document that this is not what is done in other cases in the studio docs, so that institutions can make the choice to keep everything if they really feel a need for it.  It will also (hopefully) help us in the future when someone eventually digs this up and asks why we did it differently :D

gregorydlogan:
@wsmirnow are you happy with this in its current state?

gregorydlogan:
@JulianKniephoff cutting should be enabled by default correct?  Or does it still need to be manually enabled?

lkiesow:
> @JulianKniephoff cutting should be enabled by default correct? Or does it still need to be manually enabled?

Not sure if this is a misconception. This is for compatibility with the cutting tool in Opencast Studio. This is not directly related to the editor in Opencast:

Screenshot from 2020-05-14 21-59-21

lkiesow:
@gregorydlogan, answering for Waldemar since we discussed this internally in a meeting, yes, that is what he wanted.

gregorydlogan:
> > @JulianKniephoff cutting should be enabled by default correct? Or does it still need to be manually enabled?
> 
> Not sure if this is a misconception. This is for compatibility with the cutting tool in Opencast Studio. This is not directly related to the editor in Opencast:

Misconception I think.  I get that Studio now does a soft edit, but last I checked the resulting workflow from a Studio upload was broken for editing on the Opencast end.  Either this patch or an intermediate one has fixed that - thus the question of whether we should be enabling server-side editing in the workflow by default, or not.

lkiesow:
> […] thus the question of whether we should be enabling server-side editing in the workflow by default […]

I wouldn't. Since people can review videos in Studio, most probably do not need cutting. With the addition of the new editor, it's almost certainly so.

It's also not really soft-cutting. Studio sends the SMIL file with the cutting information to Opencast where it is then automatically cut. This patch ensures that the cutting is done. The Studio update then should contain the user interface for cutting.

gregorydlogan:
Ok, I'm good with that.

PR #1580, opened 2020-05-12
TagWorkflowOperationHandler now allows wildcards in target flavor

### LINKED ISSUE #1513: Featurerequest: tag WOH should handle *-flavors as target flavor
Currently the [tag WOH](https://docs.opencast.org/develop/admin/#workflowoperationhandlers/tag-woh/) can handle multiple source flavors with the `source-flavors` option by setting `*` as flavor type, subtype or separate it by a collon. But you can not use the same sheme in `target-flavor` option. I would like to have an `target-flavors` option, that can handle the `*`-flavors.  Let me give you an example:

```
<operation
      id="tag"
      description="Tagging source videos for processing">
      <configurations>
        <configuration key="source-flavors">*/source</configuration>
        <configuration key="target-flavors">*/work</configuration>
        <configuration key="copy">true</configuration>
      </configurations>
    </operation>
```

marwyg:
This pull request should close #1513.

As described in the feature request, the `TagWorkflowOperationHandler` should now allow target flavors starting with `*/...`. When using this wildcard, all MediaPackages passing the operation should keep their flavor type and just change the subtype.

Example:
Defined target flavor: `*/foobar`

Mediapackge 1 flavor: `presentation/source`
after passing the `TagWorkflowOperationHandler`: `presentation/source` -> `presentation/foobar`

Mediapackge 2 flavor: `somethingelse/source`
after passing the `TagWorkflowOperationHandler`: `somethingelse/source` -> `somethingelse/foobar`

### DIFF
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
@@ -127,8 +127,24 @@ public class TagWorkflowOperationHandler extends AbstractWorkflowOperationHandle
         element.setIdentifier(null);
         element.setURI(e.getURI()); // use the same URI as the original
       }
-      if (configuredTargetFlavor != null)
-        element.setFlavor(MediaPackageElementFlavor.parseFlavor(configuredTargetFlavor));
+
+      if (configuredTargetFlavor != null) {
+
+        MediaPackageElementFlavor targetFlavor = MediaPackageElementFlavor.parseFlavor(configuredTargetFlavor);
+        String targetFlavorType = targetFlavor.getType();
+        String targetFlavorSubtype = targetFlavor.getSubtype();
+
+        if (MediaPackageElementFlavor.WILDCARD.equals(targetFlavorType)) {
+          targetFlavorType = element.getFlavor().getType();
+        }
+
+        if (MediaPackageElementFlavor.WILDCARD.equals(targetFlavorSubtype)) {
+          targetFlavorSubtype = element.getFlavor().getSubtype();
+        }
+
+        String targetFlavorStr = targetFlavorType + MediaPackageElementFlavor.SEPARATOR + targetFlavorSubtype;
+        element.setFlavor(MediaPackageElementFlavor.parseFlavor(targetFlavorStr));
+      }
 
       if (overrideTags.size() > 0) {
         element.clearTags();
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageElementFlavor.java
@@ -52,7 +52,7 @@ public class MediaPackageElementFlavor implements Cloneable, Comparable<MediaPac
   /**
    * Character that separates both parts of a flavor
    */
-  private static final String SEPARATOR = "/";
+  public static final String SEPARATOR = "/";
 
   /**
    * String representation of type
--- a/modules/workflow-workflowoperation/src/test/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandlerTest.java
+++ b/modules/workflow-workflowoperation/src/test/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandlerTest.java
@@ -98,4 +98,70 @@ public class TagWorkflowOperationHandlerTest {
     Assert.assertEquals("tag3", track.getTags()[1]);
   }
 
+  @Test
+  public void testTargetFlavourWithTypeWildcard() throws Exception {
+    WorkflowInstanceImpl instance = new WorkflowInstanceImpl();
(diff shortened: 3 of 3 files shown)


wsmirnow:
Hi @marwyg, thanks. What is about `dublincore/*` as target flavor? I know, this is an edge case but we are here to improve the tag WOH and make it more intuitive to use.

gregorydlogan:
@wsmirnow as a *target* flavor?  Or do you mean source?

@marwyg looks good to me, happy to merge.

gregorydlogan (approved):

wsmirnow:
I mean the target flavor.

gregorydlogan:
I can see the usecase, but it is very much an edge case IMO.  I'll let you guys discuss internally regarding time spent (or not) adding that.

wsmirnow:
Please do not merge this PR, we will improve this patch to handle wildcards in flavor type and sub type.

gregorydlogan:
Marking this as work-in-progress, please unlabel when ready

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
```diff
@@ -119,6 +119,20 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
 
+    boolean wildcardInTargetFlavorType = false;
+    boolean wildcardInTargetFlavorSubtype = false;
+    if (configuredTargetFlavor != null) {
+      String[] targetFlavorTypes = configuredTargetFlavor.split("/");
+      if (targetFlavorTypes.length != 2) {
+        String errorMsg = String.format("The configured target flavor has a wrong format. "
+                + "The configured target flavor: '%s'.", configuredTargetFlavor);
+        logger.error(errorMsg);
+        throw new RuntimeException(errorMsg);
```
gregorydlogan: This should be `WorkflowOperationException`

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
```diff
@@ -119,6 +119,20 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+      }
+      wildcardInTargetFlavorType = "*".equals(targetFlavorTypes[0]);
```
gregorydlogan: Ideally this should be a class level constant.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
```diff
@@ -127,8 +141,21 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
         element.setURI(e.getURI()); // use the same URI as the original
       }
-      if (configuredTargetFlavor != null)
-        element.setFlavor(MediaPackageElementFlavor.parseFlavor(configuredTargetFlavor));
+
+      if (configuredTargetFlavor != null) {
+        String targetFlavor = configuredTargetFlavor;
+
+        if (wildcardInTargetFlavorType) {
+          targetFlavor = targetFlavor.replaceFirst("\\*", element.getFlavor().getType());
```
gregorydlogan: This call escapes the *, the following one does not.  They should be the same, right?

gregorydlogan (changes requested):

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/TagWorkflowOperationHandler.java
lkiesow: Don't log an error when you throw an exception anyway. The result will just be that you see the same error twice in the logs. Also, I like short messages ;-)

```suggestion
        final String errorMsg = String.format("The configured target flavor `%s` has a wrong format.", configuredTargetFlavor);
        throw new RuntimeException(errorMsg);
```

lkiesow (commented):
I didn't properly read through all the code but wouldn't it be easier to do something like:


```java
flavor = MediaPackageElementFlavor.parseFlavor(flavorStr == "*" ? "*/*" : flavorStr)
…
if (flavor().getType() == "*") {
  targetFlavor = new MediaPackageElementFlavor(element.getFlavor().getType(), flavor.getSubType());
}
```

Seems easier than all the string handling plus the boolean flags.

marwyg:
@gregorydlogan 

The code improvement from Lars doesn't work like he suggested I think, but he gave me an idea. The whole boolean part above the loop were redundant, because the whole "checking if the flavor string is correct" already exists in `MediaPackageElementFlavor.parseFlavor()`. So I removed the block and just parse the configuredTargetFlavor with this method. I think the code is much more readable this way now. I am also using the constants, like you suggested. The tests still work as it seems. What do you think?

gregorydlogan:
LGTM.

Issue #1588, opened 2020-05-19
Audio does not play in Theodul player


Rillke:
**Describe the bug**
When publishing an mp3, m4a and ogg/opus file to Engage, the Theodul player does not play it, instead a warning is shown:

> Error: The audio codec is not supported by this browser

![image](https://user-images.githubusercontent.com/2311611/82325865-34729200-99dc-11ea-8308-48dd6519d9d3.png)

However, this is not true. All the audio files just play fine in the browser

**To Reproduce**
1. Have an audio-only workflow
2. Throw in an audio file, publish to engage
3. Open the publication in Theodul player

**Environment Information**
 - OS: All
 - Browser 76.0.1 (64-bit); Chromium Version 81.0.4044.138 (Official Build)
 - Opencast version with minor release 7.7

**Debug findings**
It  turns out that the message is unconditionally shown if the supposed audio element does not fire the `canplay` event. The issue here is that the event listener is bound to a div and not the audio element ;-)

PR #1589, opened 2020-05-19
Engage: Theodul: Audio: Refer to correct items

Rillke:
- The audioPlayer_id is supposed to be an HTML5 audio elem
- The audioPlayer is supposed to be a player object, not an elem
- After this patch is applied, one can at least press the play button
  and the audio plays if the codec is supported.
  Still no timeline support or seeking though.
  The play button will be still grey.

Resolves: #1588
Does not resolve: Video.js 4.x issues with `<audio>` element (seeking and much
more broken)

### Your pull request should…

* [x] have a concise title
* - include migration scripts and documentation, if appropriate
* [x] pass automated testing
* - have appropriate tags applied


gregorydlogan:
This seems like a good patch to me, although I would prefer to see the play button start working.  I'm not terribly familiar with Theodul's guts, what's keeping that from happening in this case?

lkiesow:
Definitely improves the current situation → merged.

PR #1594, opened 2020-05-22
Fix Calculation of Aspect Ratio

lkiesow:
This patch fixes the calculation of the non-existent aspect ratio of
audio files in the Theodul player.

To test this, you can use the following workflow to directly publish an audio file:

- https://data.lkiesow.io/opencast/publish-audio.xml

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
+++ b/modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
@@ -969,7 +969,7 @@ define(['require', 'jquery', 'underscore', 'backbone', 'basil', 'bowser', 'engag
   }
 
   Engage.on(plugin.events.aspectRatioSet.getName(), function (param) {
-    if (param === undefined) {
+    if (param === undefined && aspectRatio) {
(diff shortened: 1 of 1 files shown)


modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
```diff
@@ -969,7 +969,7 @@ define(['require', 'jquery', 'underscore', 'backbone', 'basil', 'bowser', 'engag
   }
 
   Engage.on(plugin.events.aspectRatioSet.getName(), function (param) {
-    if (param === undefined) {
+    if (param === undefined && aspectRatio && aspectRatio[2]) {
       Engage.trigger(plugin.events.aspectRatioSet.getName(), [aspectRatio[1], aspectRatio[2], (aspectRatio[1] / aspectRatio[2]) * 100]);
```
karendolan, line 973: This conditional check on aspectRatio looks useful in that it makes sure aspectRatio exists before attempting to extract attribute 1 and 2 from  it. The additional "aspectRatio[2]" check does not appear to be useful, because it throws an exception, as before, when aspectRatio does have an attribute number 2  (zero is never used on this object?).

lkiesow: updated

PR #1595, opened 2020-05-22
Gracefully Handle Missing Timeline Preview

lkiesow:
This patch ensures that if no timeline preview exists, e.g. because it's
an audio-only track, loading the timeline will not result in an ugly
exception.

To test this, you can use the following workflow to directly publish an audio file:

- https://data.lkiesow.io/opencast/publish-audio.xml

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


karendolan:
@lkiesow I did not try with your workflow, but it is clear that the code tests that the preview exists before attempting  to access it now.

PR #1607, opened 2020-05-27
Shibboleth dynamic login handler

jchssystems:
This feature allows to add Shibboleth users in a dynamic way from configuration.

- Configuration is located in **mh-default-org.xml**.
- Shibboleth attributes are mapped using SpEL.
- It uses a new class DynamicLoginHander instead of ConfigurableLoginHandler.

### DIFF
--- a/modules/security-aai/pom.xml
+++ b/modules/security-aai/pom.xml
@@ -45,6 +45,17 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.easymock</groupId>
+      <artifactId>easymock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <version>3.1.4.RELEASE</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
     </dependency>
@@ -62,6 +73,29 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>analyze</id>
+            <goals>
+              <goal>analyze-only</goal>
+            </goals>
+            <configuration>
+              <failOnWarning>true</failOnWarning>
+              <ignoredUsedUndeclaredDependencies>
+                <ignoredUsedUndeclaredDependency>junit:junit</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>javax.annotation:javax.annotation-api</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>org.springframework:spring-expression</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>org.springframework:spring-core</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>org.springframework.security:spring-security-core</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>org.springframework:spring-beans</ignoredUsedUndeclaredDependency>
+                <ignoredUsedUndeclaredDependency>org.hamcrest:hamcrest</ignoredUsedUndeclaredDependency>
+              </ignoredUsedUndeclaredDependencies>
+              <ignoredUnusedDeclaredDependencies>
+                <ignoredUnusedDeclaredDependency>org.easymock:easymock</ignoredUnusedDeclaredDependency>
+              </ignoredUnusedDeclaredDependencies>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.apache.felix</groupId>
--- a/modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
+++ b/modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
@@ -1,0 +1,375 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.security.aai;
+
+import org.opencastproject.security.aai.api.AttributeMapper;
+import org.opencastproject.security.api.Group;
+import org.opencastproject.security.api.JaxbOrganization;
+import org.opencastproject.security.api.JaxbRole;
+import org.opencastproject.security.api.Organization;
+import org.opencastproject.security.api.Role;
+import org.opencastproject.security.api.SecurityService;
+import org.opencastproject.security.api.UnauthorizedException;
+import org.opencastproject.security.api.User;
+import org.opencastproject.security.api.UserProvider;
+import org.opencastproject.security.impl.jpa.JpaGroup;
+import org.opencastproject.security.impl.jpa.JpaOrganization;
+import org.opencastproject.security.impl.jpa.JpaRole;
+import org.opencastproject.security.impl.jpa.JpaUserReference;
+import org.opencastproject.security.shibboleth.ShibbolethLoginHandler;
+import org.opencastproject.userdirectory.ConflictException;
+import org.opencastproject.userdirectory.api.GroupRoleProvider;
+import org.opencastproject.userdirectory.api.UserReferenceProvider;
+import org.opencastproject.util.NotFoundException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.service.cm.ManagedService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.Dictionary;
+import java.util.HashSet;
+import java.util.Hashtable;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Dynamic login with Shibboleth data through SpEL mappings
+ */
+public class DynamicLoginHandler implements ShibbolethLoginHandler, GroupRoleProvider, InitializingBean {
+
+  /** The logging facility */
+  private static final Logger logger = LoggerFactory.getLogger(DynamicLoginHandler.class);
+
+  /** The user reference provider */
+  private UserReferenceProvider userReferenceProvider = null;
+
+  /** The security service */
+  private SecurityService securityService = null;
+
+  /** The security service */
+  private AttributeMapper attributeMapper = null;
+
+  public DynamicLoginHandler() {
+    BundleContext bundleContext = FrameworkUtil.getBundle(this.getClass()).getBundleContext();
+    registerAsManagedService(bundleContext);
+  }
+
+  protected DynamicLoginHandler(BundleContext bundleContext) {
+    registerAsManagedService(bundleContext);
+  }
+
+  private void registerAsManagedService(BundleContext bundleContext) {
+    Dictionary<String, String> properties = new Hashtable<String, String>();
+    properties.put("service.pid", this.getClass().getName());
+    bundleContext.registerService(ManagedService.class.getName(), this, properties);
+  }
+
+  /**
+   * Handle a new user login.
+   *
+   * @param id
+   *          The identity of the user, ideally the Shibboleth persistent unique identifier
+   * @param request
+   *          The request, for accessing any other Shibboleth variables
+   */
+  @Override
+  public void newUserLogin(String id, HttpServletRequest request) {
+    String name = extractName(request);
+    String email = extractEmail(request);
+    Date loginDate = new Date();
+    JpaOrganization organization = fromOrganization(securityService.getOrganization());
+
+    // Compile the list of roles
+    Set<JpaRole> roles = extractRoles(id, request);
+
+    // Create the user reference
+    JpaUserReference userReference = new JpaUserReference(id, name, email, MECH_SHIBBOLETH, loginDate, organization,
+            roles);
+
+    logger.debug("Shibboleth user '{}' logged in for the first time", id);
+    userReferenceProvider.addUserReference(userReference, MECH_SHIBBOLETH);
+  }
+
+  /**
+   * Handle an existing user login.
+   *
+   * @param id
+   *          The identity of the user, ideally the Shibboleth persistent unique identifier
+   * @param request
+   *          The request, for accessing any other Shibboleth variables
+   */
+  @Override
+  public void existingUserLogin(String id, HttpServletRequest request) {
+    Organization organization = securityService.getOrganization();
+
+    // Load the user reference
+    JpaUserReference userReference = userReferenceProvider.findUserReference(id, organization.getId());
+    if (userReference == null) {
+      //Triggers creation of user reference
+      //Possible problem: if there is user (not reference) with that id, we will get conflicts
+      throw new UsernameNotFoundException("User reference '" + id + "' was not found");
+    }
+
+    // Update the reference
+    userReference.setName(extractName(request));
+    userReference.setEmail(extractEmail(request));
+    userReference.setLastLogin(new Date());
+    Set<JpaRole> roles = extractRoles(id, request);
+    userReference.setRoles(roles);
+
+    logger.debug("Shibboleth user '{}' logged in", id);
+    userReferenceProvider.updateUserReference(userReference);
+  }
+
+  /**
+   * Sets the security service.
+   *
+   * @param securityService
+   *          the security service
+   */
+  public void setSecurityService(SecurityService securityService) {
+    this.securityService = securityService;
+  }
+
+  /**
+   * Sets the user reference provider.
+   *
+   * @param userReferenceProvider
+   *          the user reference provider
+   */
+  public void setUserReferenceProvider(UserReferenceProvider userReferenceProvider) {
+    this.userReferenceProvider = userReferenceProvider;
+  }
+
+  /**
+   * Extracts the name from the request.
+   *
+   * @param request
+   *          the request
+   * @return the name
+   */
+  private String extractName(HttpServletRequest request) {
+    String displayName = extractDisplayName(request);
+    if (StringUtils.isNotBlank(displayName)) {
+      return displayName;
+    }
+    return null;
+  }
+
+  /**
+   * Extracts the e-mail from the request.
+   *
+   * @param request
+   *          the request
+   * @return the e-mail address
+   */
+  private String extractEmail(HttpServletRequest request) {
+    List<String> mailAdresses = attributeMapper.getMappedAttributes(request, "mail");
+
+    if (mailAdresses.size() == 0) {
+      return null;
+    }
+
+    String mailValue = mailAdresses.get(0);
+    String mail = StringUtils.isBlank(mailValue) ? ""
+            : new String(mailValue.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8);
+    return mail;
+  }
+
+  /**
+   * Extracts the e-mail from the request.
+   *
+   * @param request
+   *          the request
+   * @return the e-mail address
+   */
+  private String extractDisplayName(HttpServletRequest request) {
+    List<String> displayNames = attributeMapper.getMappedAttributes(request, "displayName");
+
+    if (displayNames.size() == 0) {
+      return null;
+    }
+
+    String displayNameValue = displayNames.get(0);
+    String displayName = StringUtils.isBlank(displayNameValue) ? ""
+              : new String(displayNameValue.getBytes(StandardCharsets.ISO_8859_1),
+                    StandardCharsets.UTF_8);
+    return displayName;
+  }
+
+  /**
+   * Extracts the roles from the request.
+   *
+   * @param request
+   *          the request
+   * @return the roles
+   */
+  private Set<JpaRole> extractRoles(String id, HttpServletRequest request) {
+    List<String> aaiRoles = attributeMapper.getMappedAttributes(request, "roles");
+    JpaOrganization organization = fromOrganization(securityService.getOrganization());
+    Set<JpaRole> roles = new HashSet<JpaRole>();
+    if (aaiRoles != null) {
+      for (String aaiRole : aaiRoles) {
+        roles.add(new JpaRole(aaiRole, organization));
+      }
+    }
+
+    return roles;
+  }
+
+  /**
+   * Creates a JpaOrganization from an organization
+   *
+   * @param org
+   *          the organization
+   */
+  private JpaOrganization fromOrganization(Organization org) {
+    if (org instanceof JpaOrganization) {
+      return (JpaOrganization) org;
+    }
+
+    return new JpaOrganization(org.getId(), org.getName(), org.getServers(), org.getAdminRole(),
+        org.getAnonymousRole(), org.getProperties());
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.userdirectory.api.GroupRoleProvider#getRoles()
+   */
+  @Override
+  public Iterator<Role> getRoles() {
+    JaxbOrganization organization = JaxbOrganization.fromOrganization(securityService.getOrganization());
+    HashSet<Role> roles = new HashSet<Role>();
+    roles.add(new JaxbRole(organization.getAnonymousRole(), organization));
+    roles.addAll(securityService.getUser().getRoles());
+    return roles.iterator();
+  }
+
+  /**
+   * @see org.opencastproject.security.api.RoleProvider#getRolesForUser(String)
+   */
+  @Override
+  public List<Role> getRolesForUser(String userName) {
+      ArrayList<Role> roles = new ArrayList<Role>();
+      User user = userReferenceProvider.loadUser(userName);
+      if (user != null)
+        roles.addAll(user.getRoles());
+      return roles;
+  }
+
+  /**
+   * @see org.opencastproject.security.api.RoleProvider#getOrganization()
+   */
+  @Override
+  public String getOrganization() {
+    return UserProvider.ALL_ORGANIZATIONS;
+  }
+
+  /**
+   * @see org.opencastproject.security.api.RoleProvider#findRoles(String, Role.Target, int, int)
+   */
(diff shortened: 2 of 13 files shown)


modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
```diff
@@ -0,0 +1,186 @@
...
+
+    List<String> expressions = attributeMap.get(mappingId);
+
+    if (expressions == null) {
+      logger.error("No mapping for \"" + mappingId
+          + "\" specified. Did you forget to configure a <util:map id=\""
+          + mappingId + "\" ...?");
+      throw new IllegalArgumentException("No mapping for \"" + mappingId
+          + "\" specified. Did you forget to configure a <util:map id=\""
+          + mappingId + "\" ...?");
```
lkiesow: Wouldn't the exception end up in the log anyway and make the complete error message unnecessary?

modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    for (String expression : expressions) {
+      Expression exp = null;
+      try {
+        exp = parser.parseExpression(expression);
+        String res = (String) exp.getValue(sourceAttributes);
+        logger.debug("Mapping " + exp.getExpressionString() + " -> " + res);
```
lkiesow: This kind of change applies to a number of log messages in this patch:

```suggestion
        logger.debug("Mapping {} to {}", exp.getExpressionString(), res);
```

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -275,6 +303,10 @@ public void updateGroupMembershipFromRoles(String userName, String orgId, List<S
     // List of the user's groups
     List<JpaGroup> membership = UserDirectoryPersistenceUtil.findGroupsByUser(userName, orgId, emf);
     for (JpaGroup group : membership) {
+      if (prefix != null && !"".equals(prefix) && !group.getRole().startsWith(prefix)) {
```
lkiesow: Maybe something like:
```suggestion
      if (StringUtils.isNotBlank(prefix) && !group.getRole().startsWith(prefix)) {
```

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -303,7 +335,8 @@ public void updateGroupMembershipFromRoles(String userName, String orgId, List<S
             logger.warn("Cannot add user {} to group {} - no group found with that role", userName, rolename);
           }
         } catch (UnauthorizedException e) {
-          logger.warn("Unable to add user {} to group {} - unauthorized", userName, group.getRole());
+            logger.warn("Unable to add user {} to group {} - unauthorized: {}", userName, group.getRole(),e.getMessage(),e);
+            e.printStackTrace();
```
lkiesow:
```suggestion
            logger.warn("Unauthorized to add user {} to group {}", userName, group.getRole(), e);
```

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaUserReferenceProvider.java
```diff
@@ -159,11 +177,14 @@ public String toString() {
    */
   @Override
   public List<Role> getRolesForUser(String userName) {
-    ArrayList<Role> roles = new ArrayList<Role>();
-    User user = loadUser(userName);
-    if (user != null)
-      roles.addAll(user.getRoles());
-    return roles;
+    if (roleProvider == null) {
```
lkiesow: Easier to swap this (less nesting):
```suggestion
    if (roleProvider != null) {
      return roleProvider.getRolesForUser(userName);
    }
```

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/api/GroupRoleProvider.java
```diff
@@ -0,0 +1,97 @@
...
+import org.opencastproject.security.api.RoleProvider;
+import org.opencastproject.security.api.UnauthorizedException;
+import org.opencastproject.security.impl.jpa.JpaGroup;
+import org.opencastproject.userdirectory.ConflictException;
+import org.opencastproject.util.NotFoundException;
+
+import java.util.Iterator;
+import java.util.List;
+
+public interface GroupRoleProvider extends GroupProvider, RoleProvider {
```
lkiesow, line 35: _Comment to myself_: Not entirely sure what the GroupRoleProvider is for. Need to investigate…

modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @param sourceAttributes
+   *    Key is attribute name, value a list of split AAI attribute values
+   * @param mappingId
+   *    The mapping list to use
+   * @return
+   */
+  public List<String> getMappedAttributes(
+      Map<String, List<String>> sourceAttributes, String mappingId) {
+    Set<String> mappedAttributes = new LinkedHashSet<String>();
```
lkiesow, line 78: Just wondering what the expected behavior is when an attribute appears twice?

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * This configurable implementation of the ShibbolethLoginHandler uses the UserReferenceProvider interface to create
+ * and update Opencast reference users provided and authenticated by an external identity provider.
+ * Note that this configurable implementation aims at requiring the minimum number of Shibboleth attributes
+ * to make Opencast work with most Shibboleth-based Authentication and Authorization Infrastractures (AAI).
+ */
+//public class DynamicLoginHandler implements ShibbolethLoginHandler, RoleProvider, ManagedService, InitializingBean {
```
lkiesow: Should this be deleted?

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+
+  private void registerAsManagedService(BundleContext bundleContext) {
+    Dictionary<String, String> properties = new Hashtable<String, String>();
+    properties.put("service.pid", this.getClass().getName());
+    bundleContext.registerService(ManagedService.class.getName(), this, properties);
+  }
+
+  @Override
+  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+    // TODO Auto-generated method stub
```
lkiesow: Why register a managed service if you don't handle any configuration?

modules/security-aai/pom.xml
```diff
@@ -59,10 +70,12 @@
   </dependencies>
   <build>
     <plugins>
+      <!--
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-dependency-plugin</artifactId>
       </plugin>
+      -->
```
lkiesow: Add exceptions if necessary but don't just uncomment this
```suggestion
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-dependency-plugin</artifactId>
      </plugin>
```

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+  /**
+   * Extracts the name from the request.
+   *
+   * @param request
+   *          the request
+   * @return the name
+   */
+  private String extractName(HttpServletRequest request) {
+    String displayName = extractDisplayName(request);
+    if (displayName != null && !"".equals(displayName)) {
```
lkiesow:
```suggestion
    if (StringUtils.isNotBlank(displayName)) {
```

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+      return displayName;
+    }
+    List<String> givenNames = attributeMapper.getMappedAttributes(request, "givenName");
+    String givenName = "";
+    if (givenNames.size() > 0) {
+      String givenNameValue = givenNames.get(0);
+      givenName = StringUtils.isBlank(givenNameValue) ? ""
+              : new String(givenNameValue.getBytes(StandardCharsets.ISO_8859_1),
+                      StandardCharsets.UTF_8);
```
lkiesow: Is the input guaranteed to be ISO_8859_1?

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+    List<String> mailAdresses = attributeMapper.getMappedAttributes(request, "mail");
+    if (mailAdresses.size() > 0) {
+      String mailValue = mailAdresses.get(0);
+      String mail = StringUtils.isBlank(mailValue) ? ""
+              : new String(mailValue.getBytes(StandardCharsets.ISO_8859_1),
+                      StandardCharsets.UTF_8);
+      return mail;
+    } else {
+      return null;
+    }
```
lkiesow:
```suggestion
    }
    return null;
```

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
lkiesow: And if `givenNameValue` is blank wouldn't `new String(givenNameValue.getBytes(StandardCharsets.ISO_8859_1),                      StandardCharsets.UTF_8);` still be a blank string?

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+    Set<JpaRole> roles = new HashSet<JpaRole>();
+    if (aaiRoles != null) {
+      for (String aaiRole : aaiRoles) {
+        roles.add(new JpaRole(aaiRole, organization));
+      }
+    }
+    roles.add(new JpaRole(organization.getAnonymousRole(), organization));
+    if (StringUtils.equals(id, bootstrapUserId)) {
+      roles.add(new JpaRole(GLOBAL_ADMIN_ROLE, organization));
+    }
```
lkiesow: What's the rule for assigning the global admin role here?

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+   * @param org
+   *          the organization
+   */
+  private JpaOrganization fromOrganization(Organization org) {
+    if (org instanceof JpaOrganization) {
+      return (JpaOrganization) org;
+    } else {
+      return new JpaOrganization(org.getId(), org.getName(), org.getServers(), org.getAdminRole(),
+           org.getAnonymousRole(), org.getProperties());
+    }
```
lkiesow:
```suggestion
    }
    return new JpaOrganization(org.getId(), org.getName(), org.getServers(), org.getAdminRole(),
         org.getAnonymousRole(), org.getProperties());
```

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+  public void createGroup(String name, String description, String roles, String users)
+         throws IllegalArgumentException, UnauthorizedException, ConflictException {
+    // TODO Auto-generated method stub
+  }
+
+  @Override
+  public void updateGroup(String groupId, String name, String description, String roles, String users)
+         throws NotFoundException, UnauthorizedException {
+    // TODO Auto-generated method stub
+  }
```
lkiesow: There are a lot of to-dos

lkiesow (changes requested):
I haven't actually compiled or even tested this but just skimmed through the code. I didn't find any critical issues, just a bunch of minor comments so far…

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
staubesv: Hi @lkiesow, long time no see... As I originally wrote this code, I can give some comments here.

To answer your question: https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg#L12

@jchssystems Assuming that the DynamicLoginHanlder replaces the ConfigurableLoginHandler, the bootstrap user logic is obsolete as you can configure a bootstrap user directly in the security configuration.
But the following changes seem missing to me:
- Delete ConfigurableLoginHandler.java
- Delete the configuration file of the ConfigurableLoginHandler (https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg)

Also, please properly adjust the documentation (https://github.com/opencast/opencast/blob/develop/docs/guides/admin/docs/configuration/security.aai.md). 

- Step 1 and 2 need to be revised quite a bit
- Don't forget to mention which are the mandatory headers that need to be configured in the "new way" in the security configuration to make Opencast work at all and which ones are optional
- Please explain how bootstrapping works (even if the example in the security configuration makes this obvious to me)

@JulianKniephoff This change should be mentioned in https://docs.opencast.org/develop/admin/#upgrade/#_top (Upgrading 8.x to 9.x) as it is not obvious that adaptors which use Shibboleth need to adapt their configuration quite a bit.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.security.aai;
+
+import static org.opencastproject.security.api.SecurityConstants.GLOBAL_ADMIN_ROLE;
```
staubesv: Obsolete if removing bootstrap user

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+ * This configurable implementation of the ShibbolethLoginHandler uses the UserReferenceProvider interface to create
+ * and update Opencast reference users provided and authenticated by an external identity provider.
+ * Note that this configurable implementation aims at requiring the minimum number of Shibboleth attributes
+ * to make Opencast work with most Shibboleth-based Authentication and Authorization Infrastractures (AAI).
+ */
+//public class DynamicLoginHandler implements ShibbolethLoginHandler, RoleProvider, ManagedService, InitializingBean {
+public class DynamicLoginHandler implements ShibbolethLoginHandler, GroupRoleProvider, ManagedService, InitializingBean {
+
+  /** Default value of the configuration property CFG_ROLE_FEDERATION_KEY */
+  private static final String CFG_ROLE_FEDERATION_DEFAULT = "ROLE_AAI_USER";
```
staubesv: Leftover from old code

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+  private UserReferenceProvider userReferenceProvider = null;
+
+  /** The security service */
+  private SecurityService securityService = null;
+
+  /** The security service */
+  private AttributeMapper attributeMapper = null;
+
+  /** The ID of the bootstrap user if configured */
+  private String bootstrapUserId = null;
```
staubesv: Obsolete when removing boot strap user

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+
+  /** Role assigned to all Shibboleth authenticated users */
+  private String roleFederationMember = CFG_ROLE_FEDERATION_DEFAULT;
```
staubesv: Leftover from old version

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+
+  /*
+   * It is the bundle matterhorn-kernel what will need to instantiate the ConfigurableLoginHandler
```
staubesv: DynamicLoginHandler

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
JulianKniephoff: @staubesv good to hear from you again, and thanks for the heads up. :smile: Will do when/if this is merged.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
hsssystems: Wenn givenNameValue is blank, givenName is Blank (""), otherwise not. Same code in ConfigurableLoginHandler and for the other attributes. Should be OK.

The whole building of the displayName is only a fallback since it's also possible to configure the building of the displayName in SPeL configuration...

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
hsssystems: > What's the rule for assigning the global admin role here?

Backward compatibility. However it is also possible to assign this ROLE in the SpEL configuration, so we will remove `bootstrapUserId`

hsssystems: > * Delete ConfigurableLoginHandler.java
> 
> * Delete the configuration file of the ConfigurableLoginHandler (https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.security.aai.ConfigurableLoginHandler.cfg)

It is still possible to use the old ConfigurableLoginHandler. Our intention was a new / additional LoginHandler, not to replace the old one.

modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
hsssystems: The First one will be taken. We could add a test...

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -81,7 +82,7 @@
   immediate = true,
   service = { RoleProvider.class, JpaGroupRoleProvider.class }
 )
-public class JpaGroupRoleProvider extends AbstractIndexProducer implements RoleProvider, GroupProvider {
+public class JpaGroupRoleProvider extends AbstractIndexProducer implements RoleProvider, GroupProvider, GroupRoleProvider {
```
hsssystems: Isn't it sufficient to implement only GroupRoleProvider?

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/api/GroupRoleProvider.java
hsssystems: With the interface we avoid cglib proxies.. that can cause weired problems with spring..

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
staubesv: @jchssystems As far as I understand, the DynamicLoginHandler offers a superset of the functionality of the ConfigureableLoginHandler. Why should Opencast include both the ConfigurableLoginHandler and the DynamicLoginHandler? That is just more code with no added value...

hsssystems: @staubesv we did not want to break existing configurations for backward compatibility, not every adopter needs this functionality. My proposal: we continue with both handlers, first writing a dedicated documentation for the DynamicLoginHandler, fixing all issues and finally make a new PR to remove the ConfigurableLoginHandler

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
```diff
@@ -0,0 +1,432 @@
...
+    List<String> aaiRoles = attributeMapper.getMappedAttributes(request, "roles");
+    JpaOrganization organization = fromOrganization(securityService.getOrganization());
+    Set<JpaRole> roles = new HashSet<JpaRole>();
+    if (aaiRoles != null) {
+      for (String aaiRole : aaiRoles) {
+        roles.add(new JpaRole(aaiRole, organization));
+      }
+    }
+    roles.add(new JpaRole(organization.getAnonymousRole(), organization));
+    if (StringUtils.equals(id, bootstrapUserId)) {
```
staubesv: @jchssystems bootstrapUserId is loaded from the configuration file in the original ConfigurableLoginHandler. As you don't it in DynamicLoginHandler, bootstrapUserId will always be null. But since you can configure a bootstrap user in the security configuration, it is safe to just omit those lines here.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
staubesv: @jchssystems No strong opinion, but you could expect adopters to change the configuration. Note, however, that "fixing all issues" should be done before merging ;-)

modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
jchssystems: Fix

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: It is not necesary

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: bootstrapUserId was removed.
A new independent documentation will be made when all the code be fixed.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: bootstrapUserId was removed.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: CFG_ROLE_FEDERATION_DEFAULT was also removed

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: It seems that ISO_8859_1 is guaranteed:
https://issues.shibboleth.net/jira/projects/SSPCPP/issues/SSPCPP-2?filter=allopenissues
but the code here is unnecesary

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: In my opinion you should not force to use one or another login handler (or a third in the future).

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: CFG_ROLE_FEDERATION_DEFAULT was removed

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: bootstrapUserId was also removed

modules/security-aai/src/main/java/org/opencastproject/security/aai/api/AttributeMapper.java
jchssystems: We have a Test for a duplicate attribute

modules/security-aai/pom.xml
jchssystems: Exceptions were added.

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: Fix

modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java
jchssystems: Removed

jchssystems:
Please, review tthe code again. Some changes were made.
I'm still not sure if both logins should exist or if the new one will replace the old one.
And about the documentation, should I have a totally independent page for the new login handler? when the code is ok, fix the documentation.

gregorydlogan:
@staubesv any chance you can take another pass through this?  I'm unfamiliar with any other major adopter using Shibboleth.

hsssystems:
> @staubesv any chance you can take another pass through this? I'm unfamiliar with any other major adopter using Shibboleth.

Yesterday we made another integration test on a OC9 test installation with this branch and it worked. However, we found some issues:

- we need to adapt the the documentation (example configuration) and the unit test
- Updating the displayName or group memberships (only removing users from groups) needs several logins. Do we need some index / cache invalidation?
- The GroupRoleProvider interface needs refactoring

otti-ssystems:
Hi.

I adjusted the structure of the interfaces.
DynamicLoginHandler is not extending GroupRoleProvider anymore. 
Instead we are using now AAIRoleProvider which contains getRoles, so 
i haven't had to touch RoleProvider.

I also ajusted the Tests, because the mapping of givenName and sn is not 
needed in this context.

Travis is not able to build because of a weird Exception.

https://travis-ci.com/github/opencast/opencast/jobs/366125028

jchssystems:
@lkiesow  @gregorydlogan
I have fixed the problem that was there some time ago and it compiles well.  The code was also refactored. Could you please check it again? So we can close the issue as soon as possible.
Thanks in advance.

otti-ssystems:
Hi Guys.
Sorry but i dont get it.

Is there still something to be done or is this ready to merge?

gregorydlogan:
Hi @jchssystems @otti-ssystems we don't get notifications when things change, so unless someone manually checks we don't know that you'd pushed changes.  If you don't see something happen within a week or so please ping the people who have been active in the ticket, or pipe up on the dev list.

I have no way to test this unfortunately, so I"m somewhat hamstrung in terms of getting it merged.  I'm assuming you're writing this for a client, can they verify that it's working in prod?

hsssystems:
Hi Greg,

the handler is in production on several clients! However, we will deploy again an Opencast 9 with the handler and test it another time systematically (we did already twice). Is that OK for you? 

Our QA guys will test it, not the programmers, so it should be safe :-)

gregorydlogan:
Ok, sounds fair. Let's merge this then.

PR #1615, opened 2020-05-29
Aditive Filter for Api/events endpoint

mliradelc:
This PR modifies the filter to be additive. That means that you can ask for different types of events in the same query that are mutually exclusive

For Example:

/api/events/?=filter=series:\<series-id-1\>&filter=series:\<series-id-2\>

If this is accepted I can make this enhancement for the others endpoints too

* [x] have a concise title
* [x] pass automated testing

### DIFF
--- a/docs/guides/developer/docs/api/events-api.md
+++ b/docs/guides/developer/docs/api/events-api.md
@@ -5,11 +5,11 @@
 
 Returns a list of events.
 
-The following query string parameters are supported to filter, sort and pagingate the returned list:
+The following query string parameters are supported to filter, sort and paginate the returned list:
 
 Query String Parameter |Type                         | Description
 :----------------------|:----------------------------|:-----------
-`filter`               | [`string`](types.md#basic)  | A comma-separated list of filters to limit the results with (see [Filtering](usage.md#filtering)). See the below table for the list of available filters
+`filter`               | [`string`](types.md#basic)  | A comma-separated list of filters to limit the results with (see [Filtering](usage.md#filtering)). See the below table for the list of available filters. Version 1.5.0 and newer support comma separated use of the `filter` keyword, creating a logical OR.
 `sort`                 | [`string`](types.md#basic)  | A comma-separated list of sort criteria (see [Sorting](usage.md#sorting)).  See the below table for the list of available sort criteria
 `limit`                | [`integer`](types.md#basic) | The maximum number of results to return (see [Pagination](usage.md#pagination))
 `offset`               | [`integer`](types.md#basic) | The index of the first result to return (see [Pagination](usage.md#pagination))
@@ -35,7 +35,7 @@ Filter Name       | Description
 `is_part_of`      | Events based upon which series they are a part of. Use the series identifier here (version 1.1.0 or higher)
 `source`          | Filter events whose source match this value (version 1.1.0 or higher)
 `agent_id`        | Filter events based on the capture agent id (version 1.1.0 or higher)
-`start`           | Filter events based on start date (version 1.1.0 or higer)
+`start`           | Filter events based on start date (version 1.1.0 or higher)
 `technical_start` | Filter events based on the technical start date (version 1.1.0 or higher)
 
 Note:
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -326,6 +326,10 @@ public class EventsEndpoint implements ManagedService {
     configuredMetadataFields = DublinCoreMetadataUtil.getDublinCoreProperties(properties);
   }
 
+  public static <T> boolean isNullOrEmpty(List<String> list) {
+    return list == null || list.isEmpty();
+  }
+
   @GET
   @Path("{eventId}")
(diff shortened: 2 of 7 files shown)


lkiesow:
Just wondering: Do we still care about bumping the API version when modifying the API?

gregorydlogan:
IMO yes.  Just because what we have is suboptimal does not mean that someone who is integrating with it does not care about API stability.

mliradelc:
I made the changes, now is the version 1.5.0 for the API.

mliradelc:
@gregorydlogan I'm having issues with the tests:
```
[ERROR]   ApiMediaTypeTest.testDefaultVersionAndFormat:32 expected:<v1.4.0> but was:<v1.5.0>
[ERROR]   BaseEndpointTest.testGetApiInfo:64 expected:<v1.[4].0> but was:<v1.[5].0>
[ERROR]   BaseEndpointTest.testGetVersion:125 expected:<v1.[4].0> but was:<v1.[5].0>
[ERROR]   BaseEndpointTest.testGetVersionDefault:141 expected:<v1.[4].0> but was:<v1.[5].0>
```
Should I change the tests to v1.5.0 or should be accepted because it is a lower minor version?

gregorydlogan:
Looking at the test names, I would expect 1.5.0.  They're looking at the *current* version, as well as the default, both of which should be 1.5.0 if we're bumping it.

mliradelc:
Ok, I understand, I will update the tests. I thought that it was 1.4.0 or greater.

mliradelc:
Ok, is ready, tests also in check.

docs/guides/developer/docs/api/base-api.md
```diff
@@ -11,7 +11,7 @@ Returns key characteristics of the External API such as the API base URL and the
 
 __Response__
 
-`200 (OK)`: The api information is returned as a JSON object containing the following fields:
```
gregorydlogan: While the original response descriptions were in a passive voice, the switch to active voice doesn't quite work.  This kind of thing is quite common in technical writing.

docs/guides/developer/docs/api/events-api.md
```diff
@@ -9,7 +9,7 @@ The following query string parameters are supported to filter, sort and pagingat
 
 Query String Parameter |Type                         | Description
 :----------------------|:----------------------------|:-----------
-`filter`               | [`string`](types.md#basic)  | A comma-separated list of filters to limit the results with (see [Filtering](usage.md#filtering)). See the below table for the list of available filters
+`filter`               | [`string`](types.md#basic)  | A comma-separated list of filters to limit the results with (see [Filtering](usage.md#filtering)). See the below table for the list of available filters. From version 1.5.0 is possible to repeat this filter in the query to add other results
```
gregorydlogan: Version 1.5.0 and newer support comma separated use of the `filter` keyword, creating a logical OR.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -642,177 +646,265 @@ public Response getEvents(@HeaderParam("Accept") String acceptHeader, @QueryPara
...
+      if (!requestedVersion.isSmallerThan(ApiVersion.VERSION_1_5_0)) {
+        filter = filter.subList(0,1);
+      }
+      for (String filterPart : filter) {
+        // Parse the filters
+
+        for (String f : filterPart.split(",")) {
+          String[] filterTuple = f.split(":");
+          if (filterTuple.length < 2) {
+            logger.info("No value for filter {} in filters list: {}", filterTuple[0], filter);
```
gregorydlogan, line 664: I would move this to debug since your server admin probably doesn't care what random junk their users are dropping into the filter :)

gregorydlogan (changes requested):

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -642,177 +646,265 @@ public Response getEvents(@HeaderParam("Accept") String acceptHeader, @QueryPara
...
-              Tuple<Date, Date> fromAndToCreationRange = RestUtils.getFromAndToDateRange(value);
-              query.withTechnicalStartFrom(fromAndToCreationRange.getA());
-              query.withTechnicalStartTo(fromAndToCreationRange.getB());
-            } catch (Exception e) {
-              return RestUtil.R
-                      .badRequest(String.format("Filter 'technical_start' could not be parsed: %s", e.getMessage()));
+          }
+        }
 
+        if (optSort.isSome()) {
```
gregorydlogan, line 744: I would move the sorting logic into a helper function rather than duplicating code.

docs/guides/developer/docs/api/base-api.md
mliradelc: Maybe it is because of my background with UK English but I will roll back that. I trust in a native speaker more :smile:

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
mliradelc: You are right!. How I didn't see that! 

It was bothering me a lot have to duplicate the code. Creates a new point of failure. I will create the helper function to do that.

gregorydlogan: Sounds good.  Please ping me when this is done!

PR #1622, opened 2020-06-05
Fix LTI Without Persistence

lkiesow:
This patch fixes LTI which was completely broken if persisting user
references were not enabled and partly broken if it was enabled but the
user was not yet persisted.

This problem was caused by pull request #1532 which this pull request
complete reverts.

In addition, this pull request fixes the problem #1532 was originally
supposed to fix: On concurrent LTI requests it could happen that
Opencast would try to create the same user multiple times in parallel.
This patch fixes the problem by using an optimistic approach,
deliberately allowing an insert to fail, if an entry already exists.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] be against the correct branch (features can only go into develop)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated testing
* [x] have a clean commit history
* [x] have proper commit messages (title and body) for all commits
* [x] have appropriate tags applied

PR #1532, opened 2020-04-20
Fix race condition when creating or updating user references from LTI

hsssystems:
When a LTI client submits very fast calls to Opencast for the same user it sometimes comes to bad race conditions:

Example for two concurrent requests:

```
-04-19T15:17:02,756 | DEBUG | (LtiLaunchAuthenticationHandler:254) - xcampus is a trusted key
2020-04-19T15:17:02,756 | DEBUG | (LtiLaunchAuthenticationHandler:254) - xcampus is a trusted key
2020-04-19T15:17:02,756 | DEBUG | (LtiLaunchAuthenticationHandler:278) - LTI user id is : s324216@uni-xxxx.de
2020-04-19T15:17:02,756 | DEBUG | (LtiLaunchAuthenticationHandler:278) - LTI user id is : s324216@uni-xxxx.de
2020-04-19T15:17:02,758 | DEBUG | (LtiLaunchAuthenticationHandler:384) - Adding group: ROLE_GROUP_LEARNER
2020-04-19T15:17:02,758 | DEBUG | (LtiLaunchAuthenticationHandler:384) - Adding group: ROLE_GROUP_LEARNER
2020-04-19T15:17:02,758 | DEBUG | (LtiLaunchAuthenticationHandler:394) - Adding role: 13671_Lear
```

Inserts and updates run simultanously resulting in conflicts:

```
2020-04-19T15:17:02,936 | DEBUG | (LtiProcessingFilter:49) - Skip resetting the authentication
2020-04-19T15:17:02,937 | WARN  | (HttpChannel:585) - /lti
javax.persistence.RollbackException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.3.v20180807-4
be1041): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry '13
671_Learner-mh_default_org' for key 'UNQ_oc_role'
Error Code: 1062
Call: INSERT INTO oc_role (id, description, name, organization) VALUES (?, ?, ?, ?)
        bind => [4 parameters bound]
Query: InsertObjectQuery(13671_Learner:mh_default_org)
        at org.eclipse.persistence.internal.jpa.transaction.EntityTransactionImpl.commit(EntityTransactionImpl.java:161) ~[?:?]
[...]

PR #1633, opened 2020-06-08
Using ConcurrentHashMap for synchronizing LTI user login

otti-ssystems:
This Pull Request is correcting 

https://github.com/opencast/opencast/pull/1532 
and 
https://github.com/opencast/opencast/pull/1622

Now we simple using a ConcurrentHashMap to synchronize the user authenticated by LTI.
We save the username in the map and removing it in the finally block after user is added or updated.
The fix is related to https://github.com/opencast/opencast/pull/1532 and the mentioned exceptions.

The code of the patches above is not needed anymore.

### DIFF
--- a/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
+++ b/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
@@ -52,10 +52,10 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.Dictionary;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-
-import javax.persistence.RollbackException;
+import java.util.concurrent.ConcurrentHashMap;
 import javax.servlet.http.HttpServletRequest;
 
 /**
@@ -136,6 +136,9 @@ public class LtiLaunchAuthenticationHandler implements OAuthAuthenticationHandle
 
   /** Set of usernames that should not authenticated as themselves even if the OAuth consumer keys is trusted */
   private Set<String> usernameBlacklist = new HashSet<>();
+
+  /** concurrent attemtps */
+  private Map<String, Boolean> attempts = new ConcurrentHashMap<>(128);
 
   /** Determines whether a JpaUserReference should be created on lti login */
   private boolean createJpaUserReference = false;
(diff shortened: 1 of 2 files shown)


modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
```diff
@@ -315,37 +317,36 @@ public Authentication createAuthentication(HttpServletRequest request, ConsumerA
...
-      JpaUserReference jpaUserReference = userReferenceProvider.findUserReference(username, organization.getId());
-
-      Set<JpaRole> jpaRoles = new HashSet<>();
-      for (GrantedAuthority authority : userAuthorities) {
-        jpaRoles.add(new JpaRole(authority.getAuthority(), organization));
-      }
-
-      Date loginDate = new Date();
+      if (attempts.putIfAbsent(username, Boolean.TRUE) != null) {
+        logger.warn("Concurrent access of user {}. Igoring.", username);
```
lkiesow: This now becomes expected behavior and is not something we need to warn users about, is it?

```suggestion
        logger.debug("Concurrent access of user {}. Igoring.", username);
```

modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
```diff
@@ -137,6 +137,9 @@
   /** Set of usernames that should not authenticated as themselves even if the OAuth consumer keys is trusted */
   private Set<String> usernameBlacklist = new HashSet<>();
 
+  /** concurrent attemtps */
+  private Map<String, Boolean> attempts = new ConcurrentHashMap<>(128);
```
lkiesow: Purely nitpicking but it's not that we count attempts so the choice for the name seems a bit odd.
;aybe:

```suggestion
  private Map<String, Boolean> activePersistence = new ConcurrentHashMap<>(128);
```

lkiesow (commented):
I did not try this so far but the mechanism seems valid and should work well.
Can you fix the build problem and maybe take a look at the two suggestions?

lkiesow:
Hopefully, this should now build properly.

lkiesow (approved):
Tested this and works as expected. @wsmirnow or @gregorydlogan, do you want to test this as well? For me, it's fine like this and can be merged at any time.

wsmirnow (approved):
It's fine for me.

gregorydlogan:
I haven't had a chance to test, but it looks good to me too.

lkiesow:
Okay, everyone approves 🎉
Then, if no one complains until EOD tomorrow I'll merge this.

lkiesow:
No one else complains → merging this

PR #1635, opened 2020-06-09
Admin UI embedding code

jchssystems:
This feature adds a button in the action column of the events, which allows to open a modal window. This window shows the html code to reproduce the event in a frame in five possible sizes. This code can be quickly copied to the clipboard.


modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
```diff
@@ -0,0 +1,114 @@
...
+    color: #466DC3;
+  }
+  </style>
+
+  <script>
+    var engageUrl = "<SERVER_URL>";
+
+    $(document).ready(function() {
+      var reply = $.ajax({
+        url: "/api/info/organization/properties",
```
hsssystems: isn't there a simpler way to get the engage URL? An user needs at least one of these roles (or a new role ROLE_UI_EVENTS_EMBEDDING_CODE) to access the api

`    <sec:intercept-url pattern="/api/info/organization/properties" method="GET" access="ROLE_ADMIN, ROLE_API" />`

modules/admin-ui-frontend/app/scripts/modules/events/partials/eventActionsCell.html
```diff
@@ -62,3 +62,9 @@
    with-role="ROLE_UI_EVENTS_DETAILS_ASSETS_VIEW">
 </a>
 
+<a ng-click="row.embeddingCode()"
+   class="embeddingcode"
+   title="{{ 'EVENTS.EVENTS.TABLE.TOOLTIP.EMBEDDING_CODE' | translate }}">
```
hsssystems: We should display this only for users with related roles, something like the following

```
<a ng-click="row.embeddingCode()"
class="embeddingcode"
title="{{ 'EVENTS.EVENTS.TABLE.TOOLTIP.EMBEDDING_CODE' | translate }}"
with-role="ROLE_UI_EVENTS_EMBEDDING_CODE">
</a>
```

hsssystems (changes requested):

modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
lkiesow: Plus, `/api` does not necessarily exist on the same server.
I'm not sure if the information is in the admin interface endpoints, you should get it from `/info/me.json` though. The Media Molule link in the top right corner of the admin interface uses the same information, maybe take a look at where it gets the information from.

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -146,7 +146,7 @@ prop.org.opencastproject.admin.mediamodule.url=${prop.org.opencastproject.engage
 # Value: <a complete url with scheme and port>
 # Default: ${org.opencastproject.server.url}
 #
-#prop.org.opencastproject.engage.ui.url=http://localhost:8080
+prop.org.opencastproject.engage.ui.url=http://localhost:8080
```
lkiesow: This should be commented out by default.

modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
```diff
@@ -0,0 +1,114 @@
+<section ng-show="open" ng-keyup="keyUp($event)" tabindex="1" class="modal modal-animation ng-hide" id="embedding-code">
+  <header>
+    <a class="fa fa-times close-modal" ng-click="close()"></a>
+    <h2 translate="CONFIRMATIONS.ACTIONS.SHOW.EMBEDDING_CODE"><!-- Confirm --></h2>
+  </header>
+
+  <style>
```
lkiesow: Avoid inline styles. Use the sass files in the admin interface.

jchssystems:
Check the solution to get the engage.ui.url property and how the role is managed to make the service call. We've tried to give as few permissions as possible.

modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
```diff
@@ -0,0 +1,85 @@
...
+            if (embedSizeButtons[j].id == value) {
+              embedSizeButtons[j].classList.add("embedSizeButtonSelected");
+            } else {
+              embedSizeButtons[j].classList.remove("embedSizeButtonSelected");
+            }
+          }
+          var id = $("#id_video").html();
+          var url = engageUrl + "/play/" + id;
+          var size = value.trim().split("x");
+          $("#social_embed-textarea")[0].value = '<iframe allowfullscreen src="' + url + '" style="border:0px #FFFFFF none;" name="Paella Player" scrolling="no" frameborder="0" marginheight="0px" marginwidth="0px" width="' + size[0] + '" height="' + size[1] + '"></iframe>';
```
gregorydlogan: Paella is the default, but not the guaranteed player.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
```diff
@@ -194,6 +194,27 @@ public Response getOrganizationProperties() {
     return RestUtil.R.ok(MediaType.APPLICATION_JSON_TYPE, serializer.toJson(obj(props)));
   }
 
+  @GET
```
gregorydlogan, line 197: Being that this is in the external API it would require a new version.  Luckily, OC 9 already has that done, but it's something to keep in mind for next time.

gregorydlogan (changes requested):
Minor nitpick, looks good otherwise

modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
jchssystems: @gregorydlogan
Thank you for the review. The name attribute is now only **Player** instead of **Paella Player.** You mean that? The url doesn't change, right?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/BaseEndpoint.java
jchssystems: Ok, thanks

modules/admin-ui-frontend/app/scripts/shared/partials/modals/embedding-code.html
gregorydlogan: Correct, `/play`/ is a magic endpoint which does the redirection to the correct player.

gregorydlogan (approved):

PR #1636, opened 2020-06-10
Support Serverless HLS

phoebem:
This is a rough pull request for serverless HLS. 
It has been tested with ffmpeg 4.2.3, but it should work with 4.1.x. 
The demo workflow is fast-HLS.xml.  
I changed paella to ignore mp4s if there is a master playlist, which may not be the correct behavior if you want to allow users to play mp4s as well as HLS.


mliradelc:
Hi Phoebe!
I'm reviewing your code. great work you have done!! :+1:  It has also unit tests!! :smile: 

I have some questions, from what you presented on Gent last winter, HLS serverless capability only needs to be added like an encoding profile.  What is the reason to create 3 new WOHs? Can you add in the PR description a little abstract of each?

Also, I saw that are some changes to the editor to be compatible with HLS, I though that HLS is only for publishing because it is transport stream and not is suitable to work with for editing tasks. What let you make that modification?

phoebem:
Hi Max,
	Thanks for reviewing my code. I hope that there are enough comments. Please don't hesitate to ask me questions.
We run highly specialized and optimized workflows here, so it may look a bit different from the default.
First of all, there is only one new woh - SanitizeAdaptiveWorkflowOperationHandler - It is useful only when an HLS file is uploaded (from archive) and republished without re-encoding. It is something that we do to republish old lectures.
The other 2 (MultiEncode and ProcessSmil) have been in the repo for a few releases and are modified to use the new EncodingProfile type.
We use them to encode directly from source to delivery and they share the same underlying code. The difference is that ProcessSmil uses the smil file to do a trim as well. They both call the same code to create a huge ffmpeg command to do all the encoding.
The default workflow would first run the editor UI and encode the "*/source" into "*/trimmed", and then after some processing, segmentation, slides, etc, there is another encode into "*/delivery".
Our wf takes the output from the editor - the smil file with the trim points - and save it in the mediapackage. Then ProcessSmil uses the smil file and the encoding profiles in its configuration to encode the source into the delivery formats. The delivery files are tagged with profile names and we use the highest quality one to do segmentation and generate slides. 
We do this to save one encode ("*/source" -> "*/trimmed") which can take ~90m for a 2 hr lecture.
To use it to generate HLS just for delivery, you can always use the same wf, but use MultiEncode to encode from "*/trimmed" to "*/delivery" at the end.

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -258,6 +277,13 @@ class OpencastToPaellaConverter {
         }
       }
     });
+    // HLS-VOD Where there's a master HLS index, remove all non-master HLS sources
+    if (hasAdaptiveMasterTrack && currentStream.sources.hls) {
+      var filteredHls = currentStream.sources.hls.filter(track => track.master);
+      currentStream.sources.hls = filteredHls;
+      /****  NOTE: This is a hack to force it to play HLS as a demo, probably wrong, Please fix ****/
+      delete currentStream.sources.mp4;
+    }
```
karendolan, line 285: @phoebem That delete line is not in DCE's deploy. Is it needed here because the version of OC develop Paella prefers mp4 to hls source?

miesgre: @phoebem, @karendolan In the default configuration mp4 take priority over HLS. You can change this in the `config.json`. The order in which the loaders appear in the list defines the priority.

https://github.com/opencast/opencast/blob/develop/etc/ui-config/mh_default_org/paella/config.json#L31-L76

mliradelc:
Tests Made/planned
- [x]  local machine
- [x] With no privileged user
- [ ] Multi-node installation

I will test in our local develop multi-node scheme, after that I think is ok to merge :)

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
karendolan: > @phoebem, @karendolan In the default configuration mp4 take priority over HLS. You can change this in the `config.json`. The order in which the loaders appear in the list defines the priority.
> 
@miesgre thanks! If we reverse the order in the config to put HLSVideoFactory before Html5VideoFactory, is it true that browsers that do no support HLS, will still fall back to the mp4 sources?

miesgre: Yes, Paella uses the first method the browser supports. We are using it in our installation.

gregorydlogan:
Marking this as on-hold pending results.

mliradelc:
Hi, I want to notice that I'm now able to make the last test with a cluster, I have a problems with the search endpoint in the presenter node but I think are unrelated to this commit. If I have a correct result I will merge it.

mliradelc (commented):
All ok

PR #1641, opened 2020-06-16
Capture Agent Authentication Method

lkiesow:
This patch documents and defines two authentication methods capture
agents can use. Historically, HTTP Digest was used by capture agents
while by now, HTTP Basic makes more sense in a lot of cases.

Technically, the method to use was never specified and this does not
change at all how the capture agent API works. Still, there was always
the implicit assumption that capture agents would use HTTP Digest. This
assumption was mainly for historical reasons.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


mtneug:
LGTM

gregorydlogan (approved):
+1 from me

lkiesow:
No objections here or on list → merging

PR #1657, opened 2020-06-23
Fixed the video outline in the theodul player

marwyg:
This PR should fix the uneven outline of videos in the theodul player. Before, the outline on the right side would often be thicker as the other ones.

### DIFF
--- a/modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
+++ b/modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
@@ -1422,11 +1422,6 @@ define(['require', 'jquery', 'underscore', 'backbone', 'basil', 'bowser', 'engag
       $engageVideoId.css('max-width', minWidth + 'px');
     }
     $engageVideoId.css('min-height', minVideoAreaHeight + 'px');
-    if (maxVideoAreaHeight > minVideoAreaHeight) {
-      $engageVideoId.css('max-height', maxVideoAreaHeight + 'px');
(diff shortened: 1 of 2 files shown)


modules/engage-theodul-plugin-video-videojs/src/main/resources/static/main.js
```diff
@@ -1422,11 +1422,6 @@ define(['require', 'jquery', 'underscore', 'backbone', 'basil', 'bowser', 'engag
       $engageVideoId.css('max-width', minWidth + 'px');
     }
     $engageVideoId.css('min-height', minVideoAreaHeight + 'px');
-    if (maxVideoAreaHeight > minVideoAreaHeight) {
-      $engageVideoId.css('max-height', maxVideoAreaHeight + 'px');
-    } else {
-      $engageVideoId.css('max-height', minVideoAreaHeight + 'px');
-    }
```
lkiesow, line 1429: I'm not very familiar with the code but this does not seem to be related to the issue you describe. This controls the overall height, doesn't it? Should this be removed?

marwyg: I removed this part, because it cut off a small part of the video area on the bottom side. After removing this, the video sizes were still correct. I tested different video resolutions and ratios, single and dual videos and used Firefox and Chrome. It seems to work so far.

KatrinIhler:
I tested this with one and two video sources and Martin tested with three, it looks right each time.

PR #1663, opened 2020-06-25
Enable publish-configure to publish to streaming

KatrinIhler:
Enable publish-configure and retract-configure workflow operations to be able to publish and retract streaming elements as well. The documentation and the workflows were amended accordingly, but the default behavior is unchanged. Streaming elements will only be published if the operation is configured for it and streaming is enabled in the streaming service. Streaming elements are only retracted if the operation is configured for it. Publish configure will now return "skipped" if no elements (neither download nor streaming) were published.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/publish-configure-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/publish-configure-woh.md
@@ -2,26 +2,31 @@
 
 ## Description
 
-The `ConfigurablePublishWorkflowOperationHandler` will distribute the given elements and create a publication element for
-it. By default it will retract all publications before publishing anew.
+The `ConfigurablePublishWorkflowOperationHandler` will distribute the given elements and create a publication element 
+for them. By default it will retract all published download elements before publishing anew.
 
 ## Parameter Table
 
-These are the keys that are configured through the workflow definition. At least one media package element must match
-the supplied `source-flavors` or `source-tags` or else the operation will not know what to publish. The `channel-id` and
-`url-pattern` are also mandatory.
+These are the keys that can be configured for this operation in the workflow definition. At least one media package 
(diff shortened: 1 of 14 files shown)


LukasKalbertodt:
If I'm understanding this correctly, this is a breaking change, right? The attributes `source-flavors` and `source-tags` don't work anymore and were replaced by `download-source-flavors` and `download-source-tags`, correct? I ran into this today when updating to Opencast 9.1. While this PR is mentioned in the release notes and changelog, it just includes the title of this PR. Further, it is listed under "features". So it was not apparent to me at all that this is a breaking change and that I have to potentially update my workflows. 

Was this breakage intended or an accident? If it was intended, shouldn't it be mentioned more prominently in the release notes or the upgrade guide?

KatrinIhler:
@LukasKalbertodt Yeah, you're right, this should've been in the release notes or the upgrade guide, mea culpa! Though I'm still unsure which of these would be a good spot. I guess this would've been major enough to get a section in the upgrade guide, but in general I'm still a bit fuzzy in regards to the correct workflow here. Maybe I'll bring it up in the Technical Meeting.

As for now, do you want to create a short PR to mention this change or should I?

LukasKalbertodt:
> I'm still a bit fuzzy in regards to the correct workflow here. Maybe I'll bring it up in the Technical Meeting.

I have no idea (either) when it comes to that. But yeah, communicating these breaking changes is probably pretty important for a project like Opencast. Bringing it up in the technical meeting seems like a good idea! Curious to see what comes out of it :)

> As for now, do you want to create a short PR to mention this change or should I?

You know more about the context of this change, so I guess you can describe it better than I could.

KatrinIhler:
@LukasKalbertodt I added a point to the upgrade docs, see #2328 :)

LukasKalbertodt:
Thanks ^_^

PR #1686, opened 2020-07-08
Workflow Operation "CutMarksToSmil"

Arnei:
This operation parses a JSON containing cut marks into a SMIL that can be used by the VideoEditorWorkflowOperation. It does this by attributing the given times to the tracks in the given presentation and presenter flavors. This is useful in case you have cutmarks that need to be applied to files that are created during a workflow.

Use-case example: Sending raw recording data from a web-conferencing application to Opencast, where Opencast should first concatenate the data and then cut the concatenated video file by user-defined cut marks.

An example workflow of how this can be used can be found in the [bbb-upload.xml in this repo](https://github.com/elan-ev/opencast-bigbluebutton-integration/tree/master/post-archive).

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,66 @@
...
+[VideoEditorWorkflowOperation](editor-woh.md). It does this by attributing the given times to the tracks in the 
+given presentation and presenter flavors. 
+
+## Parameter Table
+
+|configuration keys         |example                |description                                                    |
+|------------------         |-----------            |---------------------------------------------------------------|
+|source-presenter-flavor    |`presenter/prepared`   |The flavor of the presenter video track. Must contain exactly one file.                   |
+|source-presentation-flavor |`presentation/prepared`|The flavor of the presentation video track. Must contain exactly one file.                  |
+|source-smil-flavor         |`smil/times`           |The flavor of the JSON. Must contain exactly one file.|
```
ebbertd: Naming this configuration key `source-smil-flavor` is confusing. Based on this name I would expect a smil file, which is xml. However, you expect a JSON with the cut marks from BBB. Therefore I suggest to name it `source-json-flavor` and to amend the description with the information that these are the cut marks from BBB.

Arnei: Yup, that makes sense.
I would like to keep BBB out of the description, because I wouldn't want people to think that this WHO can only be used with BBB.

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,66 @@
...
+This operation parses a JSON containing cut marks into a SMIL that can be used by the 
+[VideoEditorWorkflowOperation](editor-woh.md). It does this by attributing the given times to the tracks in the 
+given presentation and presenter flavors. 
+
+## Parameter Table
+
+|configuration keys         |example                |description                                                    |
+|------------------         |-----------            |---------------------------------------------------------------|
+|source-presenter-flavor    |`presenter/prepared`   |The flavor of the presenter video track. Must contain exactly one file.                   |
+|source-presentation-flavor |`presentation/prepared`|The flavor of the presentation video track. Must contain exactly one file.                  |
```
lkiesow: Can we not hard-code presentation and presenter but just use `source-flavors` where you can define a list of flavors in the form `a/x,b/x` or a wildcard like `*/source`? That should be negligently harder to do and makes this far more flexible.

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,66 @@
...
+|source-json-flavor         |`smil/times`           |The flavor of the JSON. Must contain exactly one file.|
+|target-smil-flavor         |`smil/cutmarks`        |The flavor of the resulting SMIL.|
+
+## JSON Format
+*begin* marks the start of a segment, *duration* its duration. Times are in milliseconds.
```
lkiesow:
```suggestion
The JSON structure specifies all segments which should be kept after cutting.
The property `begin` marks the start of a segmen while `duration` its duration.
The values are specified in milliseconds.
```

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,66 @@
...
+            description="Process ingested cutmarks by applying them to current tracks"
+            fail-on-error="true"
+            exception-handler-workflow="partial-error">
+          <configurations>
+            <configuration key="source-presenter-flavor">presenter/prepared</configuration>
+            <configuration key="source-presentation-flavor">presentation/prepared</configuration>
+            <configuration key="source-json-flavor">smil/times</configuration>
+            <configuration key="target-smil-flavor">smil/cutmarks</configuration>
+          </configurations>
+        </operation>
```
lkiesow: - Keep the example as simple as possible
- Directly generate `smil/cutting`
- Use an XML code block

```suggestion
```xml
<operation
    id="cut-marks-to-smil"
    description="Process ingested cutmarks by applying them to current tracks">
  <configurations>
    <configuration key="source-presenter-flavor">presenter/prepared</configuration>
    <configuration key="source-presentation-flavor">presentation/prepared</configuration>
    <configuration key="source-json-flavor">smil/times</configuration>
    <configuration key="target-smil-flavor">smil/cutting</configuration>
  </configurations>
</operation>
```
```

File: docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md [hunk 5]
```diff
@@ -0,0 +1,66 @@
...
+            exception-handler-workflow="error"
+            description="Waiting for user to review / video edit recording">
+          <configurations>
+            <configuration key="source-flavors">*/prepared</configuration>
+            <configuration key="skipped-flavors">*/prepared</configuration>
+            <configuration key="smil-flavors">smil/cutting</configuration>
+            <configuration key="target-smil-flavor">smil/cutting</configuration>
+            <configuration key="target-flavor-subtype">trimmed</configuration>
+          </configurations>
+        </operation>
```
lkiesow: Not sure if we need half a workflow here
```suggestion
```

etc/workflows/bbb-upload.xml
```diff
@@ -0,0 +1,291 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<definition xmlns="http://workflow.opencastproject.org">
+
+  <id>bbb-upload</id>
+  <title>BigBlueButton Upload</title>
+  <tags>
+  </tags>
+  <displayOrder>1000</displayOrder>
+  <description>
+    Knock! Knock! Who's there? Candice. Candice who? Candice door open, or am I stuck out here?
```
lkiesow: This probably belongs not in here but in the BBB integration documentation at https://github.com/elan-ev/opencast-bigbluebutton-integration/tree/master/post-archive

Though with no or a sensible description ;p

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+    try {
+      return cutMarksToSmil(workflowInstance.getMediaPackage(), workflowInstance.getCurrentOperation(), elementsToClean);
+    } catch (Exception e) {
+      throw new WorkflowOperationException(e);
+    } finally {
+      for (MediaPackageElement elem : elementsToClean) {
+        try {
+          workspace.delete(elem.getURI());
+        } catch (Exception e) {
+          logger.warn("Unable to delete element {}: {}", elem, e);
```
lkiesow:
```suggestion
          logger.warn("Unable to delete element {}", elem, e);
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+    final MediaPackageElementFlavor presentationFlavor = parseTargetFlavor(
+            getConfig(operation, SOURCE_PRESENTATION_FLAVOR), "presentation");
+    final MediaPackageElementFlavor targetSmilFlavor = MediaPackageElementFlavor.parseFlavor(
+            getConfig(operation, TARGET_SMIL_FLAVOR));
+
+    // Is there a catalog?
+    Catalog[] catalogs = mediaPackage.getCatalogs(jsonFlavor);
+    if (catalogs.length != 1) {
+      logger.warn("Number of catalogs in the source flavor does not equal one. Skipping...");
+      return skip(mediaPackage);
```
lkiesow: I tend to think we should rather fail here.
Otherwise, we might accidentally publish uncut material without anyone notiving

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+
+    // If the catalog was empty, give up
+    if (cutmarks.length < 1) {
+      logger.warn("Source JSON did not contain any timestamps! Skipping...");
+      return skip(mediaPackage);
+    }
+
+    // Check parsing results
+    for (Times entry : cutmarks) {
+      logger.info("Entry begin {}, Entry duration {}", entry.begin, entry.duration);
```
lkiesow:
```suggestion
      logger.debug("Entry begin {}, Entry duration {}", entry.begin, entry.duration);
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+        smilResponse = smilService
+                .addClips(smilResponse.getSmil(),
+                        par.getId(),
+                        new Track[] { presenterTrack, presentationTrack },
+                        mark.begin,
+                        mark.duration);
+      }
+
+      Smil smil = smilResponse.getSmil();
+      logger.info("Done Adding tracks");
```
lkiesow:
```suggestion
      logger.info("Done adding tracks");
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+      InputStream is = null;
+      try {
```
lkiesow: Maybe use Java's [try-with-resources statement](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.workflow.api.AbstractWorkflowOperationHandler#skip(org.opencastproject.workflow.api.WorkflowInstance,
+   *      JobContext)
+   */
+  private WorkflowOperationResult skip(MediaPackage mediaPackage) {
+    final WorkflowOperationResult result = createResult(mediaPackage, WorkflowOperationResult.Action.CONTINUE);
+    logger.debug("Partial import operation completed");
+    return result;
+  }
```
lkiesow: Why re-define _skip_ as _continue_? That seems wrong.
I dont' actually see the purpose of this method at all.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+
+  private String getMediaPackageElementPath(MediaPackageElement mpe) throws WorkflowOperationException {
+    File mediaFile;
+    try {
+      mediaFile = workspace.get(mpe.getURI());
+    } catch (NotFoundException e) {
+      throw new WorkflowOperationException(
+              "Error finding the media file in the workspace", e);
+    } catch (IOException e) {
```
lkiesow:
```suggestion
    } catch (NotFoundException | IOException e) {
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+      throw new WorkflowOperationException(
+              "Error reading the media file in the workspace", e);
+    }
+
+    String filePath = mediaFile.getAbsolutePath();
+    return filePath;
```
lkiesow:
```suggestion
    return mediaFile.getAbsolutePath();
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,326 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.workflow.handler.composer;
+
+import static com.entwinemedia.fn.Stream.$;
```
lkiesow: We passed a rule that we won't accept new code using the Entwine Functional Library unless it really can't be avoided. For this pull request, you can easily use other means to archive the same goal.
Please avoid it.

lkiesow (changes requested):

etc/workflows/bbb-upload.xml
Arnei: Agreed, the workflow was not intended as part of the merge request, but for potential testers to have a workflow at the ready. I'll remove it then and replace it with a link to the repo you mentioned.

...and add an actual description.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
Arnei: The choice to skip instead of fail if no (or too many) catalogs are provided was made to be consistent with the behaviour of the VideoEditor WHO, which also does not fail if no cataloge is provided. Therefore I'd rather not fail here, although maybe it would be better to fail if there >1 cataloges.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
Arnei: Understood, thanks for the heads up.

lkiesow:
The build failure is caused by misformatting:

> [ERROR] /home/travis/build/opencast/opencast/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java:268: Line matches the illegal pattern 'Line has trailing spaces.'. [Regexp]

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
Arnei: I changed it to support a list of flavors (`a/x,b/x`). Does not support wildcards, because I could not figure out how to derive a list of flavors from them. Which I would like to do to check for the amount of tracks in each flavor, because the WHO does not know how it should handle multiple tracks per flavor.

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,44 @@
+CutMarksToSmilWorkflowOperationHandler
```
lkiesow: This is missing from the [list of workflow operation handlers](https://docs.opencast.org/develop/admin/#workflowoperationhandlers/#default-workflow-operations)

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,44 @@
...
+The JSON structure specifies all segments which should be kept after cutting.
+The property `begin` marks the start of a segment while `duration` its duration.
+The values are specified in milliseconds.
+
+    [
+      {
+        "begin": 1672,
+        "duration": 7199
+      }
+    ]
```
lkiesow:
```suggestion
```json
[
  {
    "begin": 1672,
    "duration": 7199
  }
]
```
```

File: modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java [hunk 18]
```diff
@@ -0,0 +1,319 @@
...
+      flavors.add(MediaPackageElementFlavor.parseFlavor(flavorName));
+
+    // Target tags
+    String targetTagsOption = StringUtils.trimToNull(operation.getConfiguration(TARGET_TAGS));
+    List<String> targetTags = asList(targetTagsOption);
+
+    // Is there a catalog?
+    Catalog[] catalogs = mediaPackage.getCatalogs(jsonFlavor);
+    if (catalogs.length != 1) {
+      logger.warn("Number of catalogs in the source flavor does not equal one. Skipping...");
```
lkiesow: Skipping this if there is none may be fine but skipping it if there are multiple catalogs is weird and unexpected. I would expect that they are both converted or that there is an error.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      return result;
+    }
+
+    // Parse JSON
+    Times[] cutmarks;
+    Catalog jsonWithTimes = catalogs[0];
+    try (BufferedReader bufferedReader = new BufferedReader(new FileReader(getMediaPackageElementPath(jsonWithTimes)))) {
+      cutmarks = gson.fromJson(bufferedReader, Times[].class);
+    } catch (Exception e) {
+      throw new WorkflowOperationException("Could not read JSON: " + e);
```
lkiesow: Use the `cause` argument of exception constructors:
```suggestion
      throw new WorkflowOperationException("Could not read JSON", e);
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      logger.warn("Source JSON did not contain any timestamps! Skipping...");
+      final WorkflowOperationResult result = createResult(mediaPackage, WorkflowOperationResult.Action.SKIP);
+      return result;
+    }
+
+    // Check parsing results
+    for (Times entry : cutmarks) {
+      logger.info("Entry begin {}, Entry duration {}", entry.begin, entry.duration);
+      if (entry.begin < 0 || entry.duration < 0) {
+        throw new WorkflowOperationException("Times cannot be negative!");
```
lkiesow:
```suggestion
        throw new WorkflowOperationException("Times may not be negative.");
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+
+    // Get video tracks
+    logger.info("Get Tracks from Mediapackage");
```
lkiesow:
```suggestion
    logger.info("Get tracks from media package");
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+
+    // Check for number of videos to avoid any issues
+    // Possible TODO: Can we handle more than one track per flavor?
+    if (videos.isEmpty())
+      return result;
+    if (videos.size() > 1) {
+      for (Track track : videos) {
+        logger.info("Track {} in flavor {}", track, flavor);
```
lkiesow: A flavor is no collection. It's an attribute:
```suggestion
        logger.info("Track {} with flavor {}", track, flavor);
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      }
+      throw new WorkflowOperationException("The number of videos in each flavor cannot be more than one.");
```
lkiesow: Is there a reason why there can be only one video with a given flavor in this operation?

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,319 @@
...
+      final WorkflowOperationResult result = createResult(mediaPackage, WorkflowOperationResult.Action.SKIP);
+      return result;
+    }
+
+    // Create the new SMIL document
+    Smil smil;
+    try {
+      SmilResponse smilResponse = smilService.createNewSmil(mediaPackage);
+
+      logger.info("Start Adding tracks");
```
lkiesow:
```suggestion
      logger.info("Start adding tracks");
```

lkiesow (dismissed):

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
Arnei: Right, that makes sense. I'll have it throw an error in that case for now, since processing multiple catalogs seems non-trivial (e.g. what if there are overlapping cutmarks).

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
Arnei: Iirc the reason for that was that the behaviour for multiple tracks in the same flavor is undefined in the editor. So this basically acts as an early error check.

lkiesow: That sounds like a bug in the editor. For some operations, you actually need multiple tracks with identical flavors. Letting it break here and kind of building more bugs doesn't sound right ;-)

gregorydlogan:
@Arnei You're hoping to get this into the codebase correct?  It's ready for review, or are you waiting on other things?

Arnei:
@gregorydlogan Yes, I do hope to get this into the codebase and it is ready for review.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/CutMarksToSmilWorkflowOperationHandler.java
Arnei: That does make sense. I just did some testing on the matter and having multiple videos in one flavor seems to work just fine. Applied your suggest change with commit 6e9702c .

lkiesow (commented):
As discussed earlier, let's put this into a separate module since it does not really have to do anything with the composer. That's something I can prepare. This will hopefully also allow us to tidy up a bit.

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
```diff
@@ -0,0 +1,57 @@
...
+
+```xml
+<operation
+    id="cut-marks-to-smil"
+    description="Process ingested cutmarks by applying them to current tracks"
+    fail-on-error="true"
+    exception-handler-workflow="partial-error">
+  <configurations>
+    <configuration key="source-media-flavors">presenter/prepared,presentation/prepared</configuration>
+    <configuration key="source-json-flavor">json/times</configuration>
```
lkiesow: I'm wondering if something like `cut-marks/json` wouldn't be a better default name.

modules/smil-workflowoperation/src/main/java/org/opencastproject/workflow/handler/smil/CutMarksToSmilWorkflowOperationHandler.java
```diff
@@ -0,0 +1,303 @@
...
+
+    // Get only videos
+    ArrayList<Track> videos = new ArrayList<>();
+    for (Track video : tracks) {
+      if (video.hasVideo()) {
+        videos.add((video));
+      } else {
+        logger.debug("Skipping track {} since it does not seem to have a vide track.", video);
+      }
+    }
```
lkiesow: I'm wondering if there is a specific reason to require video streams to be present. This would mean that we cannot cut audio and also that we need to make sure that we run `inspect` on the media package before we run this operation.

modules/smil-workflowoperation/src/main/resources/OSGI-INF/operations/cut-marks-to-smil.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.1.0"
+               name="org.opencastproject.workflow.handler.smil.CutMarksToSmilWorkflowOperationHandler"
+               immediate="true">
+  <implementation class="org.opencastproject.workflow.handler.smil.CutMarksToSmilWorkflowOperationHandler"/>
+  <property name="service.description" value="Cut Marks To Smil Operation Handler"/>
+  <property name="workflow.operation" value="cut-marks-to-smil"/>
```
lkiesow, line 7: If anyone else is wondering: It would be nice to not specify this in XML but unfortunately, some methods are contained in the parent class which means we would probably need to change all the operation handlers at once.

lkiesow (commented):
As discussed, I've moved this to a separate module and fixed a few additional minor issues (see second commit). There are two remaining open questions I have for you, @Arnei. Apart from that, I'm happy with merging this.

For testing this, here are some test scripts (make sure you have a vide.mp4 in the script folder):
https://data.lkiesow.io/opencast/cut-marks-test-data/

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
Arnei: "cut-marks" instead of "times" seems more specific and therefore clearer to understand, so +1 for that. As for the order, I put the file type first to be consistent with naming of SMIL catalogue flavors (e.g. `smil/source+partial`), but I got no hard opinion on this.

modules/smil-workflowoperation/src/main/java/org/opencastproject/workflow/handler/smil/CutMarksToSmilWorkflowOperationHandler.java
Arnei: I honestly do not remember why only video tracks were allowed (should've commented on that, arg). I'll try and see if I can't figure it out.

Arnei:
> As discussed, I've moved this to a separate module and fixed a few additional minor issues (see second commit). 

Looks good to me.

modules/smil-workflowoperation/src/main/java/org/opencastproject/workflow/handler/smil/CutMarksToSmilWorkflowOperationHandler.java
Arnei: Whatever the reason was that I originally implemented this, it seems to have vanished.  So I suppose this code snippet could be removed.

lkiesow: Removed the filter. Now it's just the track selector.

docs/guides/admin/docs/workflowoperationhandlers/cut-marks-to-smil-woh.md
lkiesow: I decided to try `cut-marks/json`. The `smil/*` is a good argument, but I don't think that smil is really meant as a file type here since it's something specific and not a general purpose file type like JSON. That would make the main type otherwise very unspecific and a wildcard like `json/*` could catch all kinds of things.

 At the same time, we also have other favors with file format as sub-type:
 
- `security/xacml(+…)`
- `captions/google-speech-json`
- `captions/ibm-watson-json`

In the end, it's just the documentation, and you can pick whatever you want anyway. Take a look, and we can still change it if you want to. I'm not that fixed on this. I will push the update shortly.

lkiesow:
@Arnei, can you check a final time if you are happy with the current state? I will merge this if you are.

Arnei:
> @Arnei, can you check a final time if you are happy with the current state? I will merge this if you are.

Looks good to me and seems to be in working order.

lkiesow:
Yay, complains now about the new checkstyle rules… fixing. Then merging…

PR #1700, opened 2020-07-17
Document bundle:watch

lkiesow:
This patch adds a short video clip documenting how to use `bundle:watch`
for more effective development.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/docs/guides/developer/docs/development-environment.md
+++ b/docs/guides/developer/docs/development-environment.md
@@ -36,11 +36,15 @@ be as follows:
 * Watch how Karaf automatically redeploys the changed jars from your local Maven repository. You can verify that
   everything went smoothly by checking the log with `log:tail`.
 
+To see this technique in action, you can watch the following short video:
+
+* [Opencast development: Watch and reload modules](https://asciinema.org/a/348132)
+
 The updated bundles are only available in the currently running Karaf instance. To create a Opencast version that has
 this changes permanently, you have to run `mvn clean install` in the the assemblies directory again. Your current
 instance will be deleted by the new assembly!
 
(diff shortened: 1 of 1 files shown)


gregorydlogan:
I'm ok with this, although I have my reservations.

What if the service goes away?  Or starts restricting views, or other such nonsense?  Nothing worse than broken docs. 
 Personally, I find videos for things like this painfully slow to gain any information from, but at the same time documenting this specific bit would be fairly complex to do well.  It might be useful to run this past someone who is *not* familiar with bundle:watch and see what their reaction is.

lkiesow:
> What if the service goes away?

That's something which I don't think is a real concern since we have the same problem for all parts of our documentation and asciinema is around for quite a while. But we could also easily [include asciinema recordings directly in the docs](https://github.com/asciinema/asciinema-player/blob/master/README.md#self-hosting-quick-start) if we wanted to due to their small size.

The question is more if we want to have videos. I agree that I nearly always prefer very short documentation in the form of texts. But that's already in the docs and people have asked about this over and over again. This is far too slow for experienced developers but beginners may actually get it. Hopefully :D

Let's see what others think about this.

gregorydlogan:
Self hosting the videos, or at least retaining the possibility of doing so should asciinema suddenly disappear is enough IMO.  We don't need to burn a ton of time on this, especially if they've been around for a long time.

I totally agree with this kind of things being more useful for newer (in the sense of unfamiliarity with OC) developers.  I'm an old curmudgeon who dislikes these newfangled moving picture things in preference to text based systems, but I've also been around for perhaps too long.  If this helps new developers get up to speed with OSGi then I'm all for it.

PR #1711, opened 2020-07-21
Fixes issue #1642: Drop-down menu does not disappear

### LINKED ISSUE #1642: Menu does not disappear after launching „Start Task“ dialog
After clicking actions → start task, the menu stays open and atop the dialog which is clearly a usability bug

Screenshot from 2020-06-20 16-43-34

mheyen:
The drop-down-menu in the adminUI now disappears after launching „Start Task“ dialog. 
This also works for:
- The language menu, 
- if you select ? → Keyboard shortcuts, 
- and the other links in that menu.

Added control for click event in modules/admin-ui/src/main/webapp/scripts/shared/directives/oldDropdownDirective.js 

This fixes #1642

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/oldDropdownDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/oldDropdownDirective.js
@@ -28,9 +28,12 @@ angular.module('adminNg.directives')
     },
     link: function ($scope, element) {
       element.on('click', function (event) {
+        var is_active = angular.element(this).hasClass('active');
         angular.element('[old-admin-ng-dropdown]').removeClass('active');
         event.stopPropagation();
-        angular.element(this).toggleClass('active');
+        if (!is_active) {
+          angular.element(this).toggleClass('active');
+        }
       });
 
       $scope.$on('$destroy', function () {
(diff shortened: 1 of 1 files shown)


modules/admin-ui/src/main/webapp/scripts/shared/directives/oldDropdownDirective.js
```diff
@@ -28,9 +28,12 @@ angular.module('adminNg.directives')
     },
     link: function ($scope, element) {
       element.on('click', function (event) {
+        var is_active = angular.element(this).hasClass('active');
         angular.element('[old-admin-ng-dropdown]').removeClass('active');
```
lkiesow: Instead of re-activating, it might be simpler to spare `this` from deactivation instead?
```suggestion
        angular.element('[old-admin-ng-dropdown]').not(this).removeClass('active');
```

lkiesow (changes requested):
Seems to work. It does break the tests though.
This might be an issue with the tests or a sign that something is wrong in some edge cases.
We need to make sure the tests run through.

I've also suggested an alternative which might be a bit easier though I don't have a strong opinion on this.

lkiesow (approved):

PR #1720, opened 2020-07-28
Corrected title of user delete button

mheyen:
The user delete button showed "Open user details" on hover. 
Changed this to "Delete user".
Changed the title in modules/admin-ui/src/main/webapp/scripts/modules/users/partials/userActionsCell.html

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/modules/users/partials/userActionsCell.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/users/partials/userActionsCell.html
@@ -9,7 +9,7 @@
    data-callback="table.delete"
    data-object="row"
    class="remove"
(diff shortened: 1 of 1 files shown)


lkiesow (approved):
Seems reasonable.
And since the label already exists in `r/8.x`, we don't need any new translations for this:

```sh
% gron src/main/resources/public/org/opencastproject/adminui/languages/lang-en_US.json | grep USERS.USERS.TABLE.TOOLTIP.DELETE
json.USERS.USERS.TABLE.TOOLTIP.DELETE = "Delete user";
```

lkiesow:
Commit 88b261283613914b4fad1a886e990e60f29ad586 probably does not belong in this pull request but should be a separate one.

PR #1756, opened 2020-08-10
Change markdown CI checks to bash

gregorydlogan:
This PR moves the markdown CI checks to a bash script rather than doing it with Travis.  This means that the script can be reused with Buildbot while also not breaking existing Travis compatibility.

This branch was developed on develop (9.x) which has a few additional checks, so I've added a commit to the tip here that should be reverted upon forward merge of 8.x.

### DIFF
--- a/.style-and-markdown-build.sh
+++ b/.style-and-markdown-build.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+set -ue
+
+ret=0
+
+cd "$(dirname $(realpath -s $0))"
+
+if grep -rn $'\t' modules assemblies pom.xml --include=pom.xml; then
+  echo "Tabs found!"
+  ret=1
+fi
+if grep -rn ' $' modules assemblies pom.xml --include=pom.xml; then
+  echo "Trailing spaces found!"
+  ret=1
+fi
+if grep -rn $'\t' etc; then
+  echo "Tabs found in config files!"
+  ret=1
+fi
+if grep -rn ' $' etc; then
+  echo "Trailing spaces found in config files!"
+  ret=1
+fi
+
+
+# build number must be present in all modules
+if ! grep -L '<Build-Number>${buildNumber}</Build-Number>' modules/*/pom.xml | wc -l | grep -q '^0$'; then
+  echo "Build number is missing from a module!"
+  ret=1
+fi
+
+# maven-dependency-plugin should be active for all new modules
+grep -L maven-dependency-plugin modules/*/pom.xml > maven-dependency-plugin.list
+if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
+  ret=1
+fi
+
+cd docs/guides
+
+for docs in admin developer; do
+  cd $docs
+  echo "Markdown doc $docs build log:"
+  mkdocs build 2>&1 | tee mkdocs.log
+  if grep \
+       -e 'WARNING.*Documentation file' \
+       -e 'pages exist in the docs directory' \
+       -e 'is not found in the documentation files' \
+        mkdocs.log;
+    then
+      echo "$docs did not build correctly!"
+      ret=1
+  fi
+  cd ..
+done
+
+if ! npm test; then
+  echo "npm test failed!"
+  ret=1
+fi
+
+exit $ret
--- a/.style-and-markdown-build.sh
+++ b/.style-and-markdown-build.sh
@@ -31,14 +31,14 @@ if ! grep -L '<Build-Number>${buildNumber}</Build-Number>' modules/*/pom.xml | w
 fi
 
 # maven-dependency-plugin should be active for all new modules
-grep -L maven-dependency-plugin modules/*/pom.xml > maven-dependency-plugin.list
-if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
-  ret=1
-fi
+#grep -L maven-dependency-plugin modules/*/pom.xml > maven-dependency-plugin.list
+#if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
+#  ret=1
+#fi
 
 cd docs/guides
 
-for docs in admin developer; do
+for docs in admin developer user; do
   cd $docs
   echo "Markdown doc $docs build log:"
   mkdocs build 2>&1 | tee mkdocs.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,31 +68,7 @@ jobs:
         - pip install mkdocs markdown_inline_graphviz_extension mkdocs-windmill
         - (cd docs/guides && npm install)
       script:
-        - (! grep -rn '	' modules assemblies pom.xml --include=pom.xml)
-        - (! grep -rn ' $' modules assemblies pom.xml --include=pom.xml)
-        - (! grep -rn '	' etc)
-        - (! grep -rn ' $' etc)
-        # build number must be present in all modules
-        - grep -L '<Build-Number>${buildNumber}</Build-Number>' modules/*/pom.xml | wc -l | grep -q '^0$'
-        - >
-          ( cd docs/guides/admin &&
-            mkdocs build > mkdocs.log 2>&1 &&
-            ! grep 'WARNING.*Documentation file' mkdocs.log &&
-            ! grep 'pages exist in the docs directory' mkdocs.log &&
-            cat mkdocs.log )
-        - >
-          ( cd docs/guides/user &&
-            mkdocs build > mkdocs.log 2>&1 &&
-            ! grep 'WARNING.*Documentation file' mkdocs.log &&
-            ! grep 'pages exist in the docs directory' mkdocs.log &&
-            cat mkdocs.log )
-        - >
-          ( cd docs/guides/developer &&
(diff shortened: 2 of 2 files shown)


.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+
+# build number must be present in all modules
+if ! grep -L '<Build-Number>${buildNumber}</Build-Number>' modules/*/pom.xml | wc -l | grep -q '^0$'; then
+  echo "Build number is missing from a module!"
+  ret=1
+fi
+
+# maven-dependency-plugin should be active for all now modules
```
lkiesow:
```suggestion
# maven-dependency-plugin should be active for all new modules
```

gregorydlogan:
Marking this as on hold until I'm sure the Travis output is right.  As it stands right now, the new script *isn't actually being run* due to a syntax error, and Travis isn't alerting about it :D

gregorydlogan:
This appears to be right now @lkiesow

.travis.yml
```diff
@@ -66,33 +66,9 @@ jobs:
       python: "3.6"
       install:
         - pip install mkdocs markdown_inline_graphviz_extension mkdocs-windmill
-        - (cd docs/guides && npm install)
+        - (cd docs/guides && npm install && cd ../..)
```
lkiesow: Why the additional `cd`? This should have no effect whatsoever since you are executing the code in a subshell, should it?

```sh
% (cd ..; pwd); pwd
/home
/home/lars
% (cd ..; pwd; cd ..); pwd
/home
/home/lars
```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
```
lkiesow, line 2: Don't just continue if errors happen e.g. when changing directory:
```suggestion

set -eu

```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
...
+#grep -L maven-dependency-plugin modules/*/pom.xml > maven-dependency-plugin.list
+#if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
+#  rev=1
+#fi
+
+cd docs/guides
+
+for docs in admin developer user; do
+  pushd . > /dev/null
+  cd $docs
```
lkiesow: `pushd` stores the original directory, making the additional `cd` somewhat weird.
Using this would work the same:
```sh
pushd "$docs"
```

But since you are not keeping multiple locations, you can also just use the simpler `cd; …; cd -` which also does not print making the pipe unnecessary.

```suggestion
  cd "$docs"
```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
...
+       -e 'pages exist in the docs directory' \
+       -e 'is not found in the documentation files' \
+        mkdocs.log;
+    then
+      echo "$docs did not build correctly!"
+      ret=1
+  fi
+  echo "Markdown doc $docs build log:"
+  cat mkdocs.log
+  popd > /dev/null
```
lkiesow:
```suggestion
  cd -
```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
...
+#if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
+#  rev=1
+#fi
+
+cd docs/guides
+
+for docs in admin developer user; do
+  pushd . > /dev/null
+  cd $docs
+  mkdocs build > mkdocs.log 2>&1
```
lkiesow: Only with the additional `set -e` will the script fail if `mkdocs` fails. Right now, actually being able to build the docs is not checked like this. Once `set -e` is included, that should be fine.

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+ret=0
+
```
lkiesow, line 6: Not strictly necessary but maybe add a
```suggestion
cd "$(dirname $(realpath -s $0))"
```
That makes it work when it's run from different locations.

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,60 @@
...
+       -e 'WARNING.*Documentation file' \
+       -e 'pages exist in the docs directory' \
+       -e 'is not found in the documentation files' \
+        mkdocs.log;
+    then
+      echo "$docs did not build correctly!"
+      ret=1
+  fi
+  echo "Markdown doc $docs build log:"
+  cat mkdocs.log
```
lkiesow: Is it intentional that `mkdocs.log` is always printed? Before it was only printed in case there was an error. Both are fine, I just want to make sure it's intentional.

If you want to print it regardless, you can use `tee` above:
```sh
mkdocs build 2>&1 | tee mkdocs.log
```

lkiesow (changes requested):

.travis.yml
gregorydlogan: Derp.  That's a leftover from an initial revision which had the checks in separate files.

.style-and-markdown-build.sh
gregorydlogan: IIRC, when I was testing this some of the useful logging output was ending up to `stderr` and *not* in the log, hence the redirect.

.style-and-markdown-build.sh
gregorydlogan: I thought it handy?  I ran into issues getting markdown to fail during testing, so I wanted to be sure the log contained *everything* and that my checks were at fault :)

.style-and-markdown-build.sh
lkiesow: Not sure if you missed this. But without this, the script will run successfully with a return value of `0` even if the `mkdocs` command fails.

gregorydlogan: It shouldn't, that's what the `ret` flag is for.  I'd like it to run all the tests, and then fail if any of them have failed.

But now that you mention it, there's a typo in the commented out code :D

lkiesow: > It shouldn't, that's what the `ret` flag is for. I'd like it to run all the tests, and then fail if any of them have failed.

Yes it will and the `ret` will not help you a bit.
Let me give you one example (there are lots of other things which can go wrong. Don't try to catch them yourself):

```
% bash .style-and-markdown-build.sh
Markdown doc admin build log:
.style-and-markdown-build.sh: line 43: mkdocs: command not found
Markdown doc developer build log:
.style-and-markdown-build.sh: line 43: mkdocs: command not found
Markdown doc user build log:
.style-and-markdown-build.sh: line 43: mkdocs: command not found

> docs@ test /home/lars/dev/opencast/docs/guides
> markdownlint *.md ./admin/**/*.md ./developer/**/*.md ./user/**/*.md ../../*.md

% echo $?
0
```

As you can see, `mkdocs` clearly failed since the command did not exist but your script returned success.
Your script currently works fundamentally different than the tests did before since you are just looking at the specific tests but knowingly ignore all errors happening with the commands itself. The `set -e` will change that. It has no effect on the `ret` values and their evaluation you built.

gregorydlogan: Good point, added.

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+set -e
```
lkiesow: Is there a particular reason to not set `nounset`? You resolved my suggestion for adding it without comment. I don't think it will have a huge effect here but it also will not hurt and is a safer choice when people potentially modify the script.

```suggestion
set -ue
```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,63 @@
...
+# maven-dependency-plugin should be active for all new modules
+#grep -L maven-dependency-plugin modules/*/pom.xml > maven-dependency-plugin.list
+#if ! diff -q maven-dependency-plugin.list docs/checkstyle/maven-dependency-plugin.exceptions; then
+#  ret=1
+#fi
+
+cd docs/guides
+
+for docs in admin developer user; do
+  pushd . > /dev/null
```
lkiesow: This has no function anymore
```suggestion
```

.style-and-markdown-build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if grep -rn $'\t' modules assemblies pom.xml --include=pom.xml; then
+  echo "Tabs found!"
+  ret=1
+fi
+if grep -rn ' $' modules assemblies pom.xml --include=pom.xml; then
+  echo "Trailing spaces found!"
+  ret=1
+fi
+if grep -rn '	' etc; then
```
lkiesow: Just to keep to one style:
```suggestion
if grep -rn $'\t' etc; then
```

lkiesow (changes requested):
Generally looks good. There are a few leftovers from the previous review.

.style-and-markdown-build.sh
gregorydlogan: Nope, no reason, just forgot to add it :D

lkiesow (approved):

lkiesow:
@gregorydlogan, do you want to do the merge back to develop?

PR #1760, opened 2020-08-11
User interface to sort by number of publications

lkiesow:
This patch adds the user interface to allow for sorting events by the
number of publications. This allows for example to easily find events
without any publications.

This patch will only work once #1758 is merged to fix the broken
back-end.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


gregorydlogan:
To be clear, this is a binary field right?  It's either got publications, or it doesn't?  "Sort by number of publications" implies to me that I can sort by actual number of publications (ie, default wf is 3, fast is 1).

lkiesow:
publications is actually a nested field, basically storing an array of complex objects. That's why the current sorting did not work. The fix in #1758 fixes the error by using the number of publications for dorting. So, it's actually the number. It's not binary.

PR #1763, opened 2020-08-12
Fix for issue 1280:  Notification of Newer Opencast Version in Admin UI

### LINKED ISSUE #1280: Notification of Newer Opencast Version in Admin UI
There is a notification area in the Admin UI behind the bell. 

It would be nice to create a new notification, if a newer Opencast version exists then the version that its currently installed. Maybe even a link to the release docs would be nice.

The current version should be known as it is shown on the bottom of the admin UI.

New versions can be found in the Github tag list.

mheyen:
This patch adds a notification in the admin UI (behind the bell) if there is a newer version of Opencast available.
The notification shows:
- if there is a new minor release (e.g. 8.4 -> 8.5)
- if there is a new major release (e.g. 7.4 -> 8.1)
- if your version is no longer supported (older than two major releases)

Added service in restServiceMonitor.js :
- compares your version to the latest version and adds notification accordingly.
- adds link in the notification to https://docs.opencast.org/r/<LATEST VERSION>/admin/ to enable the user to update if neccesary.

Added new endpoint at VersionEndpoint.java :
- retrieves latest version from https://api.github.com/repos/opencast/opencast/releases/latest .
- updates every hour.
- provides version via /admin-ng/oc-version/version.json .

This fixes #1280

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
@@ -61,7 +61,7 @@ angular.module('adminNg.services')
           var my_version = response_my_version.data.version;
           var latest_version = response_latest_version.data;
           services.service[LATEST_VERSION_NAME].docs_url =
-            'https://docs.opencast.org/r/' + latest_version[0] + '.x/admin/';
+            'https://docs.opencast.org/r/' + parseInt(latest_version) + '.x/admin/';
 
           if (parseFloat(my_version) >= parseFloat(latest_version)
              || (my_version[0] == latest_version[0] && my_version.endsWith('SNAPSHOT'))) {
--- a/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
@@ -69,7 +69,7 @@ angular.module('adminNg.services')
           } else if (my_version[0] == latest_version[0]) {
             services.service[LATEST_VERSION_NAME].status = 'There is a minor update available.';
             services.service[LATEST_VERSION_NAME].error = true;
-          } else if (parseInt(latest_version[0]) - parseInt(my_version[0]) <= 2) {
+          } else if (parseInt(latest_version[0]) - parseInt(my_version[0]) < 2) {
             Monitoring.setError(LATEST_VERSION_NAME, 'There is a major update available.');
           } else {
             Monitoring.setError(LATEST_VERSION_NAME,
--- a/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
@@ -63,7 +63,8 @@ angular.module('adminNg.services')
           services.service[LATEST_VERSION_NAME].docs_url =
             'https://docs.opencast.org/r/' + latest_version[0] + '.x/admin/';
 
-          if (parseFloat(my_version) >= parseFloat(latest_version) || my_version.endsWith('SNAPSHOT')) {
+          if (parseFloat(my_version) >= parseFloat(latest_version)
+             || (my_version[0] == latest_version[0] && my_version.endsWith('SNAPSHOT'))) {
             services.service[LATEST_VERSION_NAME].status = OK;
             services.service[LATEST_VERSION_NAME].error = false;
           } else if (my_version[0] == latest_version[0]) {
--- a/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
@@ -29,6 +29,7 @@ angular.module('adminNg.services')
     numErr: 0
   };
 
+  var LATEST_VERSION_NAME = 'Latest Version';
   var AMQ_NAME = 'ActiveMQ';
   var STATES_NAME = 'Service States';
   var BACKEND_NAME = 'Backend Services';
@@ -36,6 +37,8 @@ angular.module('adminNg.services')
   var OK = 'OK';
   var SERVICES_FRAGMENT = '/systems/services';
   var SERVICE_NAME_ATTRIBUTE = 'service-name';
+  var LATEST_VERSION_PATH = 'oc-version/version.json';
+  var MY_VERSION_PATH = '/sysinfo/bundles/version?prefix=opencast';
 
   Monitoring.run = function() {
     //Clear existing data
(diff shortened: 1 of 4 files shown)


modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
```diff
@@ -45,6 +48,42 @@ angular.module('adminNg.services')
...
+          services.service[LATEST_VERSION_NAME].docs_url =
+            'https://docs.opencast.org/r/' + latest_version[0] + '.x/admin/';
+
+          if (parseFloat(my_version) >= parseFloat(latest_version) || my_version.endsWith('SNAPSHOT')) {
+            services.service[LATEST_VERSION_NAME].status = OK;
+            services.service[LATEST_VERSION_NAME].error = false;
+          } else if (my_version[0] == latest_version[0]) {
+            services.service[LATEST_VERSION_NAME].status = 'There is a minor update available.';
+            services.service[LATEST_VERSION_NAME].error = true;
+          } else if (parseInt(latest_version[0]) - parseInt(my_version[0]) <= 2) {
```
lkiesow:
```suggestion
          } else if (parseInt(latest_version[0]) - parseInt(my_version[0]) < 2) {
```

Currently, 8.6 is the latest version and 8.x and 7.x is supported.
If someone has a 6.4 then `8 - 6 = 2` which should yield the `…is no longer supported. Please update.`

modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
```diff
@@ -45,6 +48,42 @@ angular.module('adminNg.services')
...
+        Monitoring.populateService(LATEST_VERSION_NAME);
+
+        if (response_latest_version.status === 200 && response_my_version.status === 200
+        && response_my_version.data.consistent) {
+          var my_version = response_my_version.data.version;
+          var latest_version = response_latest_version.data;
+          services.service[LATEST_VERSION_NAME].docs_url =
+            'https://docs.opencast.org/r/' + latest_version[0] + '.x/admin/';
+
+          if (parseFloat(my_version) >= parseFloat(latest_version) || my_version.endsWith('SNAPSHOT')) {
```
lkiesow:
```suggestion
          if (parseFloat(my_version) >= parseFloat(latest_version)
             || (my_version[0] == latest_version[0] && my_version.endsWith('SNAPSHOT'))) {
```

This could also be `4.0.0-SNAPSHOT` which could be identified as very old.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/VersionEndpoint.java
```diff
@@ -0,0 +1,113 @@
...
+    try {
+      HttpClient client = HttpClientBuilder.create().build();
+      HttpGet request = new HttpGet(url);
+      HttpResponse response = client.execute(request);
+      String responseString = IOUtils.toString(response.getEntity().getContent(), "utf-8");
+      Map data = gson.fromJson(responseString, Map.class);
+      version = (String) data.get("tag_name");
+      lastUpdated = System.currentTimeMillis() / 1000L;
+    } catch (Exception e) {
+      logger.error("Error while parsing the version: " + e.toString());
```
lkiesow: I would make this a warning since it's not something an admin would (or most likely even could) fix.
And use the logger's functionality to handle exceptions:

```suggestion
      logger.error("Error while parsing the Opencast version from GitHub", e);
```

lkiesow (changes requested):
Mostly nitpicking…

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/VersionEndpoint.java
mheyen: How do I make this a warning? 
Does the logger.error() function handle the Exception in a way such that it will create a warning?

lkiesow: `logger.warn(…)`

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/VersionEndpoint.java
```diff
@@ -0,0 +1,113 @@
...
+  protected void activate() {
+    logger.info("Activate the Admin ui - Latest version endpoint");
+  }
+
+  @GET
+  @Path("version.json")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RestQuery(name = "latestversion", description = "Returns the latest Opencast version",
+          returnDescription = "Returns the latest Opencast version retrieved from GitHub",
+          reponses = { @RestResponse(responseCode = SC_OK, description = "The latest Opencast version.") })
```
gregorydlogan: This should be `re**s**ponses`, especially since I'm looking at #1766 too :)

modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
```diff
@@ -58,11 +98,7 @@ angular.module('adminNg.services')
         services.service[AMQ_NAME].error = true;
       }
     }).catch(function(err) {
-      Monitoring.populateService(AMQ_NAME);
```
gregorydlogan, line 61: What was the reasoning behind removing this?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/VersionEndpoint.java
```diff
@@ -0,0 +1,113 @@
...
+  public String getVersion() {
+    if (System.currentTimeMillis() / 1000L - lastUpdated >= 3600) {
```
gregorydlogan, line 91: Probably only need to check this once, either here, or in `updateVersion`

lkiesow: It does make sense to check this twice. If you check it just in `updateVersion` you have basically a single-threaded web server for this endpoint since that function is synchronized. If you have it just here, you can have multiple updates triggered at once which could cause GitHub temporarily blocking you.

gregorydlogan: Ah, I see what you're saying.  I'm not sure that avoiding the single thread matters, but you're correct in that this approach will scale better.

modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
mheyen: Monitoring.setError() already calls the Monitoring.populateService() function. 
Therefore I thought it would not be necessary to call it twice. 
Basically it was just to remove redundant code.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/VersionEndpoint.java
mheyen: That is absolutely correct ;)

modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
```diff
@@ -45,6 +48,43 @@ angular.module('adminNg.services')
...
+    $http.get(MY_VERSION_PATH).then(function(response_my_version) {
+      $http.get(LATEST_VERSION_PATH).then(function(response_latest_version) {
+        Monitoring.populateService(LATEST_VERSION_NAME);
+
+        if (response_latest_version.status === 200 && response_my_version.status === 200
+        && response_my_version.data.consistent) {
+          var my_version = response_my_version.data.version;
+          var latest_version = response_latest_version.data;
+          services.service[LATEST_VERSION_NAME].docs_url =
+            'https://docs.opencast.org/r/' + latest_version[0] + '.x/admin/';
```
gregorydlogan: Testing this, I patched out the endpoint to return "10.0" as the version.  This line generates links to `https://docs.opencast.org/r/1.x/admin/` rather than `https://docs.opencast.org/r/10.x/admin/`

Otherwise looks good and works as expected!

gregorydlogan (changes requested):

modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js
mheyen: You are right. I patched this and it should work now. Thanks!

PR #1768, opened 2020-08-14
Remove Empty Test Classes

lkiesow:
This patch removes a number of test classes which all contain no actual
tests and which clearly copied from each other. If someone actually
wants to write a test, restoring this is trivial so there is no value in
keeping this around.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


mtneug (approved):
LGTM

Note that at least Sakai and Moodle had ignored tests which were removed in #1291. So no, this isn't just copy-paste. The tests were marked as ignored because the LMS was not mocked. I found it helpful to activate the tests during development.

PR #1788, opened 2020-08-19
Properly parse boolean values

lkiesow:
This patch switches to using BooleanUtils for parsing boolean values
from configuration files instead of doing a string comparison. This
means that a string like `True` will not become the boolean value
`false` which is pretty confusing.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


lkiesow:
@mtneug, any complaints about this?

PR #1791, opened 2020-08-20
Add with acl option to series api

KatrinIhler:
This adds the option `withACL` to api/series/ and api/series/{seriesid} so you can additionally receive the ACL(s) of the series from the external API index. This way an external application doesn't have to query the ACL of each series separately. This is very similar to the same option in the events API.

And because this is an API change, it also introduces a new version of the API.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
@@ -55,7 +55,8 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0,
+            ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
 @RestService(
     name = "externalapicaptureagents",
     title = "External API Capture Agents Service",
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
@@ -55,7 +55,7 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0 })
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
 @RestService(
     name = "externalapicaptureagents",
     title = "External API Capture Agents Service",
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
@@ -124,8 +124,10 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
-@RestService(name = "externalapiseries", title = "External API Series Service", notes = {}, abstractText = "Provides resources and operations related to the series")
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0,
+            ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
+@RestService(name = "externalapiseries", title = "External API Series Service", notes = {},
+             abstractText = "Provides resources and operations related to the series")
 public class SeriesEndpoint {
 
   private static final int CREATED_BY_UI_ORDER = 9;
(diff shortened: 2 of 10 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -344,15 +359,45 @@ public JValue apply(SearchResultItem<Series> a) {
...
+   *
+   * @param series
+   *          The {@link Series} to get the ACL from.
+   * @return The {@link AccessControlList} stored in the {@link Series}
+   */
+  private static AccessControlList getAclFromSeries(Series series) {
+    AccessControlList activeAcl = new AccessControlList();
+    try {
+      if (series.getAccessPolicy() != null)
+        activeAcl = AccessControlParser.parseAcl(series.getAccessPolicy());
```
LukasKalbertodt: I remember a mailing list discussion about brace style and I *think* the vast majority of responses were in favor of Greg's proposal to make braces mandatory.

```suggestion
      if (series.getAccessPolicy() != null) {
        activeAcl = AccessControlParser.parseAcl(series.getAccessPolicy());
      }
```

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -344,15 +359,45 @@ public JValue apply(SearchResultItem<Series> a) {
...
+    } catch (Exception e) {
+      logger.error("Unable to parse access policy", e);
+    }
+    return activeAcl;
```
LukasKalbertodt: Is "log error and continue with dummy value" a valid error handling style in Opencast? I would assume the API should respond with 5xx if parsing the ACL fails. But I am not sure, as I have no experience on how the API code usually handles these things.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
```diff
@@ -55,7 +55,7 @@
 import javax.ws.rs.core.Response;
 
 @Path("/")
-@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0 })
+@Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0, ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
```
LukasKalbertodt: Opencast doesn't have to be any enforced line length limit, but maybe it's time to break this line? This also applies to other places in this PR.

LukasKalbertodt (changes requested):
This PR is already quite old, so I thought I would try to review it. Please note that I never worked in the API code and are thus not really qualified to approve this PR. But maybe some of my comments are useful anyway.

I tested this patch locally and it seems to work, as far as I can tell. I left a few minor inline comments of things I noticed. Additionally, in your PR description you mention "`withACL`", but the code uses `withacl`. Taking a look at similar parameters, e.g. `onlyWithWriteAccess`, it seems like camel case is used, so the API parameter should probably be renamed to `withAcl` or `withACL`.

(Tangential note: I couldn't help but notice that this PR is a perfect motivational example for why one would like to use GraphQL :D)

lkiesow:
@LukasKalbertodt, let me know when this is ready to be merged.

KatrinIhler:
@LukasKalbertodt 

Regarding the parameter name: Unfortunately the API is not at all consistent here. This is basically a copy-paste of the same code from the events API, where the parameter is also called "withacl". So I think it's slightly less confusing to keep it that way here. (I'll correct the commit message so that it matches though.)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
KatrinIhler: You're right. I simply copied this over from the EventsEndpoint, but I'll add some braces (I like braces!)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
KatrinIhler: I'll break the lines containing the API versions. The rest I'll leave as is for now, otherwise I'll still be here tomorrow. ;)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
KatrinIhler: This is a tricky one. In general, this is probably not a good idea. However, if asking for a list of a series fails with 500 only because one ACL is broken, that's also gonna suck. I guess it would be possible not to include an ACL if it can't be parsed. However, then the series API and the event API will behave differently there. And changing the latter is not something I want to spend time on right now. So maybe this is okay since it falls back to the default ACL, which is "allow nothing". Opinions?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/CaptureAgentsEndpoint.java
LukasKalbertodt: Looks good now. Apparently I can't "resolve" these PR comments. Feel free to resolve this one yourself.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
LukasKalbertodt: > However, then the series API and the event API will behave differently there.

That's a shame. When was that part of the API added? Would it still be possible to change it without breaking backwards compatibility? 

I do think that using the default ACL as fallback is not a good idea. For software using the API, this can easily lead to strange errors down the line. Even if these programs would explicitly check for an empty ACL and interpret it as "ACL cannot be parsed", then we have no way of actually communicating an empty ACL. (Not sure if that can ever happen.) 

In my opinion, if parsing the ACL should never happen (as in: if it fails, there is a bug somewhere else in Opencast), then the API should return 5xx in that case. If parsing failure is somewhat expected and does not imply a bug in Opencast, then the API should explicitly return `acl: null` or something like that to clearly indicate broken ACL. 

In this week alone, strange and unexpected behaviors of the API were brought up twice in our internal meetings. Everyone just sighs, maybe laughs, but then carries on. If the Opencast community just keeps adding to the pile of strange API things, the situation is only gonna get worse. I am obviously not blaming you by the way; it's just the community's stance on this issue that I kind of criticize. 

So I personally would not merge this PR, but fix the behavior as stated above and if somehow possible also fix the behavior of the event API to make them consistent. *That said*, this is not for me to decide. Maybe the benefits of having this API in 9.x outweighs the disadvantages of the strangeness. Maybe this can still be fixed after the feature freeze. And I probably just lack experience in this community to know what's best.

LukasKalbertodt:
> Regarding the parameter name: [...]

Unfortunate, but I can live with that.

So I tested and reviewed this PR again and everything looks fine to me... except the error handling discussion that is (see above). I guess now someone else has to decide how to proceed.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
KatrinIhler: I understands your frustration with the API in general and swallowing errors specifically , though I think this case isn't quite as bad. At least this way, Opencast offers a reasonable fallback (deny everything) instead of leaving every external system to figure this out for themselves.
The point of changing behaviour of existing API endpoints is that this can (per our rules) only be done for new API versions, so you would have to add a condition to the events API that handles broken ACLs differently for different versions. This is a general problem which stems from the trade-off between a stable API and improving things. Not sure if this case is worth it.

KatrinIhler:
@lkiesow I think your opinion is demanded here. ;) (I'm out for the rest of the week now.)

Arnei (commented):
I'm at least as unqualified as Lukas, but I can at least confirm that this PR seems to be working as advertised.

lkiesow:
Two unqualified developers = one qualified developer? Maybe.

PR #1814, opened 2020-09-03
Add a 'defaultValue' to getComponentContextProperty

Gao-Jun:
Usually, properties were loaded from cfg files.
Some of them may have default values or allow empty.
Yet, the code to set default value if redundant.

This PR can make it easy to load such properties.

Here's an example, to property 'endpoint' actually is optional for AWS S3 distruibution.
To handle it properly, the code may be
```Java
String endpoint = "";
try {
  endpoint = OsgiUtil.getComponentContextProperty(cc, "endpoint");
} catch (RuntimeException e) {
  endpoint = "";
}
```
With the new method, this can be easily written as
```Java
String endpoint = OsgiUtil.getComponentContextProperty(cc, "endpoint", "");
```

### Your pull request should…

* [x] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/util/OsgiUtil.java
+++ b/modules/common/src/main/java/org/opencastproject/util/OsgiUtil.java
@@ -39,6 +39,7 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.servlet.Servlet;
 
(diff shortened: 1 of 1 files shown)


lkiesow:
Thanks for the patch.
Maybe even simpler:
```java
Objects.toString(properties.getProperty(KEY), DEFAULT);
```

Gao-Jun:
@lkiesow 
You are right. This is simpler. I updated the PR.

Gao-Jun:
Anything I should do to make this PR to be merged?
SinceI have some other PRs depend on this PR merged.

gregorydlogan:
@gao-jun just wait, we'll get there. Our merge rate isn't where I would like it to be, but we'll get to it. If you have other PRs dependent on this one I would file them anyway and make a note in their descriptions that they depend on this PR. We see this not infrequently so we check before we merge.

PR #1848, opened 2020-09-15
Fix capture agent dropdown menus

### LINKED ISSUE #1845: Wrong capture agents selected in the admin UI for new events
**Describe the bug**

When creating a new event in the admin UI, the capture agent dropdown in the scheduling tab appears to be broken:

- Sometimes the capture agent you pick out isn't the one that's actually set.
- Sometimes a capture agent can't be picked, instead there's an error in the console: "Uncaught TypeError: this.form_field.options[a.options_index] is undefined". This appears to be the case if you pick the last capture agent in the list.

Both errors only occur if you change your mind _after_ you've already picked a capture agent once; the first time works just fine.

This seems to be a similar error as #1357 and can most likely be fixed in a similar way.

**To reproduce**
- Add a new event via the admin UI
- Choose "Schedule Single" or "Schedule Multiple" in the source tab
- In the scheduling tab, change capture agent selection more than once (!) via the dropdown menu
- Watch the conflicts requests in the network tab of your developer tools to see which capture agent is _actually_ chosen
- Or watch the console for errors
- You need to have at least two capture agents available to reproduce this bug

**Not affected**
- The scheduling tab in Event Details is not affected, most likely because the dropdown menu seems to be reset after each selection.
- Edit Scheduled Events is likewise not affected - though there is another bug that makes this tricky to test (see #1847).

**Environment Information**
 - OS: Ubuntu 20.04
- Browser: Firefox
 - Only 8.x and develop are affected, this bug does _not_ exist in 7.x

KatrinIhler:
This fixes #1845. Without this fix, the capture agent dropdown menu in the scheduling tab of the New Event dialog is broken: 
If you change your selection, either a different capture agent is picked than the one you actually selected (which is... very bad, because you might not notice this at all until much later) or the selection fails with an error. (For more details on how to reproduce this see the issue description.)

I also edited the ng-options of the capture agent dropdown menus in Event Details and Edit Scheduled Events, even though these weren't affected, to keep things consistent. It doesn't do any harm and is imo easier to understand.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/edit-events-modal.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/edit-events-modal.html
@@ -186,7 +186,7 @@
                             data-width="'200px'"
                             ng-change="roomChanged();"
                             ng-model="scheduling[wd].location"
-                            ng-options="value.name for (id, value) in captureAgents | filter:hasAgentAccess"
+                            ng-options="ca.name for ca in captureAgents | filter:hasAgentAccess track by ca.id"
                             placeholder-text-single="'{{ 'EVENTS.EVENTS.DETAILS.SOURCE.PLACEHOLDER.LOCATION' | translate }}'"
                             />
                   </td>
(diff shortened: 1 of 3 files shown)


KatrinIhler:
@gregorydlogan Since this is a potentially nasty bug that can lead to recordings happening in the wrong room, could you take a look at this?

PR #1857, opened 2020-09-18
Refactor Metadata classes (Updated version)

KatrinIhler:
This is an updated version of #1284. Unfortunately I wasn't able to push to plapadoo's fork to update that pull request directly, even though Github claims I should be able to, and a test with Julian's fork was successful. So instead I created this one. I'll close the other once this one is merged.

I reviewed Plapadoo's pullrequest and was content with everything except for two things, which this pull request addresses:
- CommonEventCatalogUIAdapter didn't consider the possibility that the returned value of SeriesTitle() could be Opt.none(). I removed the Opt from the method and always used null instead to be consistent.
- The LTI-service-module didn't compile, since the changes introduced by #1289 used methods that, after the refactoring of #1284, no longer existed in that form. I refactored the LTIServiceImpl accordingly.

One spot in the LTIServiceImpl was quite tricky, since it initially used a MetadataList just to fill a MetadataCollection from JSON. I instead decided to extract the relevant part out of the JSON and fill the MetadataCollection directly.

I tested the LTI stuff (first time trying out LTI, yaay) and everything seems to be fine, but I'd feel better if someone else took a look at my changes as well. @gregorydlogan Can you do that?


gregorydlogan:
Last time I touched LTI bits @lkiesow got quite cross with me 😂. I'll try and take a look later this week.

KatrinIhler:
@gregorydlogan That's why I wanted someone else to take a look, to share responsibility. ;)
But these changes are going into develop, so we also have some time to fix stuff if they break anything (which they shouldn't, but still...)

gregorydlogan:
Looks good to me.  Merging.

PR #1861, opened 2020-09-22
Update fast.xml to work with republish workflows (update for MH-13617)

karendolan:
This pull fixes the workflow fast.xml to archive audio fixed prepared flavored tracks expected by republish workflows as of commit "MH-13617 Reintroduce Prepare AV #956".

Currently, republish workflows applied to archived events from fast.xml fail. Republish workflows expect all archives from upload/ingest workflows to contain tracks flavored "*/prepared" with fixed audio. This pull adds  fixed audio */prepared tracks to fast.xml, as they have been applied to other upload/ingest workflows.

This is a fix only for future fast.xml archives, not for existing archives that do not contain "*/prepared" tracks.

### DIFF
--- a/etc/workflows/fast.xml
+++ b/etc/workflows/fast.xml
@@ -70,6 +70,44 @@
       </configurations>
     </operation>
 
+    <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
+    <!-- Prepare the work media                                            -->
+    <!--                                                                   -->
+    <!-- Ensure the work media is in a format that allows Opencast to do   -->
+    <!-- its work. This includes potentially rewriting the container and   -->
+    <!-- making sure that the audio track is part of each video track.     -->
+    <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
+
+    <!-- Prepare presenter track -->
+
+    <operation
+      id="prepare-av"
+      exception-handler-workflow="partial-error"
+      description="Preparing presenter (camera) audio and video work versions">
+      <configurations>
+        <configuration key="source-flavor">presenter/source</configuration>
+        <configuration key="target-flavor">presenter/prepared</configuration>
+        <configuration key="target-tags">archive</configuration>
+        <configuration key="rewrite">false</configuration>
+        <configuration key="audio-muxing-source-flavors">*/?,*/*</configuration>
+      </configurations>
+    </operation>
+
+    <!-- Prepare presentation track -->
+
+    <operation
+      id="prepare-av"
+      exception-handler-workflow="partial-error"
+      description="Preparing presentation (screen) audio and video work version">
+      <configurations>
+        <configuration key="source-flavor">presentation/source</configuration>
+        <configuration key="target-flavor">presentation/prepared</configuration>
+        <configuration key="rewrite">false</configuration>
+        <configuration key="target-tags">archive</configuration>
+        <configuration key="audio-muxing-source-flavors">*/?,*/*</configuration>
+      </configurations>
+    </operation>
+
     <!-- encode video -->
 
     <operation
@@ -78,7 +116,7 @@
       exception-handler-workflow="partial-error"
       description="Encoding video">
       <configurations>
-        <configuration key="source-flavor">*/source</configuration>
+        <configuration key="source-flavor">*/prepared</configuration>
         <configuration key="target-flavor">*/preview</configuration>
         <configuration key="target-tags">engage-download,engage-streaming,rss,atom</configuration>
         <configuration key="encoding-profile">fast.http</configuration>
@@ -220,7 +258,7 @@
       exception-handler-workflow="partial-error"
       description="Archiving">
       <configurations>
-        <configuration key="source-flavors">*/source,dublincore/*,security/*</configuration>
+        <configuration key="source-flavors">*/source,*/prepared,dublincore/*,security/*</configuration>
       </configurations>
     </operation>
 


karendolan:
I can update this pull to include a "*/source" fallback for old archives that don't have the "*/prepared" to workflow "etc/workflows/partial-publish.xml". Or, I could put fix in a different pull.

DZenker:
Hi Karen,

as discussed in the User group, ALL published events are afffected by this problem, even if they had been processed with the upload-and-schedule WF. I've also tried to fix this with an additional "prepare-av" WO in publish.xml, but this did not solve the problem. 
I could fix this with an additional retag operation just after the configuration section in publish.xml:

	_

	<operation
	  id="tag"
	  max-attempts="1"
	  fail-on-error="false"
	  exception-handler-workflow="error"
	  description="Removing prepared tag from Presentation track">
	  <configurations>
		<configuration key="source-flavors">presentation/prepared</configuration>
		<configuration key="target-flavor">-presentation/prepared</configuration>
		<configuration key="copy">false</configuration>
	  </configurations>
	</operation>

	<operation
	  id="tag"
	  max-attempts="1"
	  fail-on-error="false"
	  exception-handler-workflow="error"
	  description="Removing prepared tag from Presenter track">
	  <configurations>
		<configuration key="source-flavors">presenter/prepared</configuration>
		<configuration key="target-flavor">-presenter/prepared</configuration>
		<configuration key="copy">false</configuration>
	  </configurations>
	</operation>

    <operation
      id="include"
      if="NOT ${flagNormalizeAudio} AND NOT ${flagFixVideo}"
      description="Prepare the recording for further processing">
      <configurations>
        <configuration key="workflow-id">partial-work</configuration>
      </configurations>
    </operation>_

Now it's working in our installation.

karendolan:
> I could fix this with an additional retag operation just after the configuration section in publish.xml:
>  [untagging of the prepared tracks]
> <operation
>   id="include"
>   if="NOT ${flagNormalizeAudio} AND NOT ${flagFixVideo}"
>   description="Prepare the recording for further processing">
>   <configurations>
>     <configuration key="workflow-id">partial-work</configuration>
>   </configurations>
> </operation>_
> ```
> 
> Now it's working in our installation.

@DZenker where are the ${flagNormalizeAudio}  and ${flagFixVideo} coming from? I ask because the prepared tagging is related to fixing the audio. The "partial-work" workflow in r/8.x relies on "*/prepared" tagging to tag the tracks to "*/work" to perform editing and image extraction. That error about  the missing "*/work" tag is related to https://github.com/opencast/opencast/blob/develop/etc/workflows/partial-publish.xml#L13-L22.

The current r/8.x error with fast.xml is that it does not produce tracks with "*/prepared".
I cannot see, in the current r/8.x, how untagging "*/prepared" avoids that missing "*/work" tags, because the work tags are created from prepared tagged tracks. Are you sure you publish items are getting worked?

karendolan:
@DZenker I tested with the "Process upon upload and schedule" workflow in stable.opencast.org and a "publish" workflow after and it succeeded. It succeeded because that vanilla workflow archives tracks with the prepared tags. I think your workflow contains customizations that are in conflict with the default workflows.

DZenker:
Hi Karen,

you're completely right: these are indeed two problems showing the same symptom: on stable.opencast.org, only events published with _fast.xml_ are affected, despite in our installation where also the schedule-and-publish events are affected if they are edited (cut). I wasn't aware that all events on stable.opencast.org have been processed with the fast WF, and I should have to verify this by uploading and processing with a complete _schedule-and-upload_ WF. **Sorry for the confusion!**

I could also fix this in our installation with the additional retag operation described, now everything is working as expected.  

Concerning the conditionals: yes, these are some customizations in our WFs to fix some problems even if the videos already have been published, e.g. to normalize the audio if the audio level is too low - frequently, lecturers realize this after they have uploaded their recordings, as they don't check this before the upload.

Greetings,
Dietmar

gregorydlogan:
This works as advertised for me.  @lkiesow you self requested a review, waiting on you.

JulianKniephoff:
For the sake of a better overview over which PRs are already taken on this Review-Friday(TM), I'm going to assign this to you, @gregorydlogan. Your message above indicated that you tested it and would merge it, pending the review by @lkiesow, so no one else needs to look at this right now. (Although anyone is of course invited to!)

Feel free to unassign and yell at me if this is wrong. ;)

lkiesow (dismissed):
I'm not that happy with this:

- We defeat the goal of the fast testing workflow if we add more and more unnecessary operations again. For the purpose of processing and publication, the added operations here have basically no meaning except additional work and are thus slowing down the workflow. In the past, we have removed stuff from this over and over again since things ended up in here.
- The purpose of the two new operations is basically just to switch flavors/copy the track. We could do that in one operation if necessary. I think using `tag` should work.
- Generally, we shouldn't need to keep a copy of the same flavor around just because another workflow wants that.

While the 3rd option may be hard, at least using `tag` deserves a try.

gregorydlogan:
The majority of my testing is async (ie, start a workflow, go do something else for half an hour) so speed doesn't matter so much to me, and I don't have a baseline for how fast `fast` *should* be.

Let's explore what @lkiesow is suggesting, since the point of `fast` is to be... well... fast!

gregorydlogan:
@karendolan are you going to have time to look at this more, or should someone else take over?

karendolan:
@lkiesow , I'm confused about your suggestion. Are you saying I should be fixing the republish workflow to do the audio preparation, rather than including it in the fast.xml? Just doing a tag will not fix the fast.xml to work with the other Opencast workflows.

karendolan:
In other words, is the republish workflow broken because it can't work with the fast.xml or is fast.xml broken because it can't work with republish. IMHO, it's easier to make fast.xml compatible with the other Opencast workflows rather than adding a check to fix audio on the republish workflow.

karendolan:
> I'm not that happy with this:
> 
> * We defeat the goal of the fast testing workflow if we add more and more unnecessary operations again. For the purpose of processing and publication, the added operations here have basically no meaning except additional work and are thus slowing down the workflow. In the past, we have removed stuff from this over and over again since things ended up in here.
> * The purpose of the two new operations is basically just to switch flavors/copy the track. We could do that in one operation if necessary. I think using `tag` should work.
> * Generally, we shouldn't need to keep a copy of the same flavor around just because another workflow wants that.
> 
> While the 3rd option may be hard, at least using `tag` deserves a try.

@lkiesow can you help me out on how to fix the audio with tag? I only know how to add tags with tag.

lkiesow:
Sorry for leaving this unattended for so long :see_no_evil: 

You are right, `fast.xml` currently does not merge separately ingested audio and video tracks. I overlooked that you saw that as an additional issue. I made that as a deliberate choice since you cannot easily get such media into opencast apart from ingesting that manually via API. Opencast, by default, does not provide a user interface for such ingests for quite some time and thus it felt like unnecessary work slowing down the workflow to handle those cases.

Of course, you can still ingest such media and even configure the user interface to provide that option but in that case you could also configure your workflows differently.

Not sure though if you have a valid use-case for this. In that case, we can include this. But for me, the costs (loss of speed) did not out weight the benefits of handling such edge cases.

---

Regarding the issue that `fast` doesn't work nicely with `republish`, while we could do a few things which would be faster and simpler and produce less data than than `prepare-av`_but_ would not affect the audio (basically options for if we decide that it is okay to not support that).

#### Use `tag` to switch flavors

A very simple way to switch to `*/prepared` would be to use a single `tag` operation to switch the flavors in the `fast` workflow. This would effectively fix the problem of not having `*/prepared`. Since this would modify the flavor, it's relatively fast and means we have less data to archive. We would need to do that directly before the `snapshot` operation to allow other operations to use `*/source`. Switching using the `tag` operation would look like this:

```xml
  <operation                                                                                                           
      id="tag"                                                                                                         
      description="Duplicating media">                                                                                 
    <configurations>                                                                                                   
      <configuration key="source-flavors">*/source</configuration>                                                     
      <configuration key="target-flavor">*/prepared</configuration>                                                    
    </configurations>                                                                                                  
  </operation> 
```

The downside of using `tag` is that we don't archive `*/source`. 


####  Moving things to `publish.xml`

We could even move all of this tothe publish workflow which would have the benefit of having no additional operations in `fast` and working on already processed material. The idea here would be to move `source` to `prepared` if no `prepared` media exists. Something we can easily chack. The workflow would look like this:

```diff
diff --git a/etc/workflows/publish.xml b/etc/workflows/publish.xml
index 9fb414395d..0e3b9dc97a 100644
--- a/etc/workflows/publish.xml
+++ b/etc/workflows/publish.xml
@@ -144,6 +144,29 @@
       </configurations>
     </operation>
 
+    <operation
+        id="analyze-tracks"
+        fail-on-error="true"
+        exception-handler-workflow="partial-error"
+        description="Analyze tracks in media package and set control variables">
+      <configurations>
+        <configuration key="source-flavor">*/prepared</configuration>
+      </configurations>
+    </operation>
+
+
+    <operation
+        if="NOT (${presentation_source_media} OR ${presenter_source_media})"
+        id="tag"
+        fail-on-error="true"
+        exception-handler-workflow="partial-error"
+        description="Preparing media">
+      <configurations>
+        <configuration key="source-flavors">*/source</configuration>
+        <configuration key="target-flavor">*/prepared</configuration>
+      </configurations>
+    </operation>
+
     <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
     <!-- Publish                                                           -->
     <!--
```

This would be my suggestion, but still, there are some downsides to this as well:

- You have some additional operation in `publish`. That hopefully does not matter much though since the workflow is awefully slow anyway.
- You may still only archive `prepared` in certain circumstances.
- This only checks `presenter` and `presentation` while at least `fast` could handle more flavors (I don't think `schedule-and-upload` can and there is also no easy way to upload multiple flavors)

But the benefits are:
- This works regardless of the workflow which was used for processing before
- This fixes the issue for already processed media
- This puts no burden on `fast`


####  `prepare-av` in `publish.xml`

Finally, we could also do the same thing in `publish.xml` with `prepare-av` if we want to partly keep the support for separate audio files. You would need to publish using `publish.xml` though. I haven't actually tested this but see no reason why this would not work. But if we don't specifically want to support separate audio, there also should be no reason to use this.

```diff
diff --git a/etc/workflows/publish.xml b/etc/workflows/publish.xml
index 9fb414395d..d8900499f2 100644
--- a/etc/workflows/publish.xml
+++ b/etc/workflows/publish.xml
@@ -144,6 +144,46 @@
       </configurations>
     </operation>
 
+    <operation
+        id="analyze-tracks"
+        fail-on-error="true"
+        exception-handler-workflow="partial-error"
+        description="Analyze tracks in media package and set control variables">
+      <configurations>
+        <configuration key="source-flavor">*/prepared</configuration>
+      </configurations>
+    </operation>
+
+    <operation
+        if="NOT (${presentation_source_media} OR ${presenter_source_media})"
+        id="prepare-av"
+        exception-handler-workflow="partial-error"
+        description="Preparing presenter (camera) audio and video work versions">
+      <configurations>
+        <configuration key="source-flavor">presenter/source</configuration>
+        <configuration key="target-flavor">presenter/prepared</configuration>
+        <configuration key="target-tags">archive</configuration>
+        <configuration key="rewrite">false</configuration>
+        <configuration key="audio-muxing-source-flavors">*/?,*/*</configuration>
+      </configurations>
+    </operation>
+
+    <!-- Prepare presentation track -->
+
+    <operation
+        if="NOT (${presentation_source_media} OR ${presenter_source_media})"
+        id="prepare-av"
+        exception-handler-workflow="partial-error"
+        description="Preparing presentation (screen) audio and video work version">
+      <configurations>
+        <configuration key="source-flavor">presentation/source</configuration>
+        <configuration key="target-flavor">presentation/prepared</configuration>
+        <configuration key="rewrite">false</configuration>
+        <configuration key="target-tags">archive</configuration>
+        <configuration key="audio-muxing-source-flavors">*/?,*/*</configuration>
+      </configurations>
+    </operation>
+
     <!-- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -->
     <!-- Publish                                                           -->
     <!--                                                                   -->
```

Let me know what you think.

lkiesow:
I just realized that this patch would also more or less limit the usability of fast to two stream recordings while it did support an arbitrary amount of streams before. Again, this is an edge-case we don't need to support but it may be worth mentioning.

In any case, the first decision I think we have to make is if we care about supporting separate audio streams in the `fast` workflow.

gregorydlogan:
Personally I would either use `tag` to switch flavors, but then I see `fast` as a *testing* workflow, rather than something one might use in production.  I'm also leery of introducing dependence on the default workflow into fast since then we need to keep testing it with `fast`...

karendolan:
Ok, so for the purposes of fixing the bug of people misunderstanding the fast workflow for production and needing a fast workflow for production. Can I just clone the fast workflow to "fast-production" with the bug fix so it can be used in production?

I will also update this pull to emphasize that the "fast" testing workflow is a one-use and nothing archived from it can be reused.

gregorydlogan:
That sounds reasonable to me.  Perhaps `minimal` rather than `fast-production`?  Minimal being the minimum set of things which Opencast *needs* for reasonable prod use.

karendolan:
> That sounds reasonable to me. Perhaps `minimal` rather than `fast-production`? Minimal being the minimum set of things which Opencast _needs_ for reasonable prod use.

Sounds good, I'll rename the old "fast" to "minimal" with all documentation and notes that it's archive cannot be used for republishing or anything but downloads. I'll rename the fixed fast to "fast-republishable"

gregorydlogan:
`fast` is used in a number of scripts around the community. Renaming the fixed version and leaving the broken one as fast would be better imo.

gregorydlogan:
@karendolan Ruediger is planning on a release towards the end of the month, can we get this done before then?

karendolan:
[EDIT] upon re-read of all the comments on this pull. It sounds like just retagging "source" files as "prepared" works to make "fast" archived recordings compatible with other Opencast workflows?
[Third EDIT] I suspect removing the source tag and only archiving with prepared tag will cause havoc with peoples existing use of the fast workflow. I'll submit the textual clarification pull

karendolan:
replaced with https://github.com/opencast/opencast/pull/2734

PR #1865, opened 2020-09-22
Create h2-migration.md

orbruno:
First version of h2 to MariaDB/MySQL migration tutorial

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


lkiesow:
I don't think this is maintainable. This script relies on internal database structure which changes all the time. There was a recent database change which may already mean that this does not work any longer as expected. If we want to keep something like this around, we need something agnostic to the specific database structure right now.

Closing this for now. Let's re-open it once we have a solution.

PR #1874, opened 2020-09-26
PostgreSQL and auto-generated databases

lkiesow:
As promised in the proposal “JPA, PostgreSQL and DDL-Scripts” (passed on
Oct 29, 2019) we worked on replacing the current set of MariaDB based
DDL scripts with a proper auto-generated database. These changes are
now merged into the Opencast codebase so that we do not need the DDL
scripts any longer.

This patch brings these changes to an end, removing the DDL scripts and
updating the documentation to reflect the auto-generated database and
the newly added – for now, still marked as experimental – support for
PostgreSQL.

OAI-PMH with it's weird – and somewhat problematic – database usage
still needs a manual set-up of one table. This has been added to the
documentation but is only necessary if you actually configure and use
OAI-PMH which is not the case in a default installation.

To avoid confusion: We still need upgrade scripts. This is only about
the initial database setup.

---

While this is basically just documentation and the topic has already passed as proposal, it would be great if someone else could read through this before it gets merged.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing


docs/guides/admin/docs/configuration/database.md
```diff
@@ -104,49 +103,80 @@ Finally, leave the client and restart the database server to enable the new user
...
+  mp_id VARCHAR(128) NOT NULL,
+  organization VARCHAR(128) NOT NULL,
+  repo_id VARCHAR(255) NOT NULL,
+  series_id VARCHAR(128),
+  deleted tinyint(1) DEFAULT '0',
+  modification_date DATETIME DEFAULT NULL,
+  mediapackage_xml TEXT(65535) NOT NULL,
+  PRIMARY KEY (mp_id, repo_id, organization),
+  CONSTRAINT UNQ_oc_oaipmh UNIQUE (modification_date)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
lwillmann, line 174: Does not work with postgresql ... ENGINE creates a syntax error

docs/guides/admin/docs/configuration/database.md
```diff
@@ -104,49 +103,80 @@ Finally, leave the client and restart the database server to enable the new user
...
-% mysql -u root -p opencast
 ```
-
-Run the ddl script:
-
-```
-mysql> source mysql5.sql;
+sudo -u postgres psql
+postgres=# create database opencast;
+postgres=# create user opencast with encrypted password 'opencast_password';
```
lwillmann, line 118: Maybe it is helpful to add a comment that by default PostgreSQL uses IDENT-based authentication, thus username and password authorization has to be configured.

lwillmann (commented):
The configuration file etc/custom.properties still includes the comment:
" ... It is recommended that you run the ddl script for your db vendor (see docs/scripts/ddl/)"
The line  should be removed.

docs/guides/admin/docs/configuration/database.md
lkiesow: Depends on your installation type. E.g. containers use password by default. But I'll add a comment

lkiesow:
Docs are updated.

lwillmann (commented):
Looks good to me

Issue #1224, opened 2019-11-18
 Coverimage WOH appendXml not appending all values


stefanosgeo:
**Describe the bug**
A clear and concise description of what you see:
If there are 2 presenters in an event and a coverimage WHO is used, then the generated .png using  coverimage.xsl have only the first value

**To Reproduce**
1. Create an event with 2 presenters "prsenter1" and "presenter2" or 2 contributors
2. use  coverimage WHO like this
```xml
 <operation

      id="cover-image"

      fail-on-error="true"

      exception-handler-workflow="partial-error"

      description="Create a cover image">

      <configurations>

        <configuration key="stylesheet">file://${karaf.etc}/branding/coverimage.xsl</configuration>

        <configuration key="width">1920</configuration>

        <configuration key="height">1080</configuration>

        <configuration key="posterimage-flavor">presenter/player+preview</configuration>

        <configuration key="target-flavor">image/cover</configuration>

        <configuration key="target-tags">branding</configuration>

     </configurations>

    </operation>
```

3. Append the metadata value in branding/coverimage.xsl
        <tspan class="creators" x="730" y="810">
          <xsl:value-of select="opencast:XsltHelper.split(metadata/creators, 63, 1, false)" />
        </tspan>
        <tspan class="creators" dy="5%" x="730">
          <xsl:value-of select="opencast:XsltHelper.split(metadata/creators, 63, 2, false)" />
        </tspan>
        <tspan class="creators" dy="5%" x="730">
          <xsl:value-of select="opencast:XsltHelper.split(metadata/creators, 63, 3, true)" />
        </tspan>

4. appendXML should be append both values. By bug only one is displayed

**Environment Information**
 - OS: Ubuntu 18.04
 - Chrome, FF
 - Opencast version 7.3

stefanosgeo:
This maybe has to do, IIRC ,with the dublincore metadata structure.

In previous version has the creators term in dublincore comma seperated
like this

<dcterms:creator>user1, user2 </dcterms:creator>

but with either 6.x or 7.x every values has it's own xml element

<dcterms:creator>user1</dcterms:creator>
<dcterms:creator>user2</dcterms:creator>

The same happen with the metadata/contributors strings

Arnei:
Fixed by #1877

PR #1877, opened 2020-09-28
Extended CoverImageWOH to be able to use extended and series metadata

### LINKED ISSUE #1787: The cover-image WOH can only use episode metadata
**Describe the bug**
The [cover-image WOH](https://docs.opencast.org/r/8.x/admin/#workflowoperationhandlers/coverimage-woh/) can only use episode metadata. It should be possible to access the [series metadata as well as extended metadata](https://docs.opencast.org/r/8.x/admin/#configuration/metadata/).

**Environment Information**
* Opencast 8

jduehring:
This PR should fix #1787 . The user can now specify the flavor of the episode and series in the workflow using the parameters `episodeFlavor` and `seriesFlavor`. If not specified `dublincore/episode` and `dublincore/series` will be used as default values. Depending on what flavor was specified, the metadata of the episode gets extracted and built into xml format **including** a field for the metadata of the corresponding series (if there is any). I did not change the names of the other fields, so there shouldn't come up any conflicts with older workflows. 
Like before all fields can be accessed in the coverimage.xsl and therefore can be included in the cover image.  

* [ ] have a concise title
* [ ] pass automated testing


ebbertd:
Please also update the documentation: https://github.com/opencast/opencast/blob/develop/docs/guides/admin/docs/workflowoperationhandlers/coverimage-woh.md

modules/cover-image-workflowoperation/src/main/java/org/opencastproject/workflow/handler/coverimage/CoverImageWorkflowOperationHandlerBase.java
```diff
@@ -302,57 +304,87 @@ protected String loadXsl(WorkflowOperationInstance operation) throws WorkflowOpe
     }
   }
 
-  protected String getMetadataXml(MediaPackage mp) {
-    StaticMetadata metadata = getStaticMetadataService().getMetadata(mp);
+  protected String getMetadataXml(MediaPackage mp, WorkflowOperationInstance operation) {
+    //get specified episode/series flavor
+    String configuredEpisodeFlavor = StringUtils.trimToEmpty(operation.getConfiguration(EPISODE_FLAVOR));
```
lkiesow: This is easier than to check manually for empty values:
```suggestion
    final String configuredEpisodeFlavor = Objects.toString(StringUtils.trimToNull(operation.getConfiguration(EPISODE_FLAVOR)), "dublincore/episode);
```

modules/cover-image-workflowoperation/src/main/java/org/opencastproject/workflow/handler/coverimage/CoverImageWorkflowOperationHandlerBase.java
```diff
@@ -302,57 +304,87 @@ protected String loadXsl(WorkflowOperationInstance operation) throws WorkflowOpe
...
+      configuredEpisodeFlavor = "dublincore/episode";
+    }
+    if (("").equals(configuredSeriesFlavor)) {
+      configuredSeriesFlavor = "dublincore/series";
+    }
+
+    String episodeFlavorType = configuredEpisodeFlavor.split("/")[0];
+    String episodeFlavorSubtype = configuredEpisodeFlavor.split("/")[1];
+    String seriesFlavorType = configuredSeriesFlavor.split("/")[0];
+    String seriesFlavorSubtype = configuredSeriesFlavor.split("/")[1];
```
lkiesow: Better use

```java
MediaPackageElementFlavor.parseFlavor(flavorStr);
```
You can then access the flavor and subflavor via the flavor object's methods.
And it assures that it's a valid flavor or directly use it since you want a flavor object down below anyway.

modules/cover-image-workflowoperation/src/main/java/org/opencastproject/workflow/handler/coverimage/CoverImageWorkflowOperationHandlerBase.java
```diff
@@ -302,57 +304,87 @@ protected String loadXsl(WorkflowOperationInstance operation) throws WorkflowOpe
...
+
+    //Get episode metadata-catalog
+    Catalog[] catalogs = mp.
+            getCatalogs(new MediaPackageElementFlavor(episodeFlavorType, StringUtils.lowerCase(episodeFlavorSubtype)));
+
+   //load metadata-catalog
```
lkiesow: indentation

modules/cover-image-workflowoperation/src/main/java/org/opencastproject/workflow/handler/coverimage/CoverImageWorkflowOperationHandlerBase.java
```diff
@@ -302,57 +304,87 @@ protected String loadXsl(WorkflowOperationInstance operation) throws WorkflowOpe
...
-    for (Date[] period : metadata.getTemporalPeriod()) {
-      if (period[0] != null) {
-        appendXml(xml, "start", new SimpleDateFormat("YYYY-MM-dd'T'HH:mm:ss").format(period[0]));
-      }
-      if (period[1] != null) {
-        appendXml(xml, "end", new SimpleDateFormat("YYYY-MM-dd'T'HH:mm:ss").format(period[1]));
+    for (Map.Entry<EName, List<DublinCoreValue>> entry : data.entrySet()) {
+      String currentKey = entry.getKey().getLocalName();
+      switch(currentKey) {
+        case "creator": appendXml(xml, "creators", entry.getValue().get(0).getValue()); break;
```
lkiesow: All these fields can exist not at all or multiple times. Will this work if they do not exist? Onbly using the first is probably okay but should be documented.

wsmirnow: Also the order of the list isn't fixed. I guess, we have to provide all values as list.

jduehring: @lkiesow If they do not exist, they are not beeing added to the xml, so it should be fine. There were no issues during my tests with non existing fields. I also added a note in the documentation.

@wsmirnow You mean the order of the fields in the resulting xml should stay intact?

lkiesow: iirc, the order isn't guaranteed in an XML document anyway.

wsmirnow: I would prefer the same order. It may be technically not possible. If you put only the first value of the list, you may end up with different values each call of the handler.

lkiesow: Now I'm confused as well :D
The code should be deterministic, shouldn't it?
So you would get the same. You just don't get all of them.

@jduehring, what happens if you iterate e.g. over multiple creator fields and call `appendXml` every time?

jduehring: Not sure how to test this. What would be the easiest way to get an episode containing metadata with multiple creator fields? Should this happen when I add multiple creators via the admin-ui? So for each creator one creator field?
Currently there seems to be a bug so that this does not happen. (Probably the reason for #1224 as well)
In theory my code should currently add  the same field multiple times, which is also not ideal, i guess.

lkiesow: Sorry for the late reply :see_no_evil: 

The following cURL command will ingest media with two titles (which will mostly be ignored by Opencast code and just the first is usually being used) and multiple creators (which Opencast can handle just fine… mostly):

```sh
curl -i -u admin:opencast \
  https://develop.opencast.org/ingest/addMediaPackage \
  -F title="Title 1" \
  -F title="Title 2" \
  -F creator="Creator 1" \
  -F creator="Creator 2"
```

lkiesow:
@jduehring, what's the current state of this. Is it ready for another review?

gregorydlogan:
@jduehring Is this ready for further review?

jduehring:
Sorry for the late reply. Ready for review @lkiesow @gregorydlogan . 

I added a method, which is now able to include multiple creators, publishers etc. ,if neccessary. So #1224 should be fixed as well.

Issue #1893, opened 2020-09-30
Enforce Maven Dependency Check


lkiesow:
Most of Opencast's modules use the maven-dependency-pligin to ensure dependencies are specified correctly:

https://github.com/opencast/opencast/blob/a67ece488617d23101e16357ab5da0c69105644b/modules/animate-impl/pom.xml#L88-L98

Some modules still have an exception though because no one found the time to fix them until now:

https://github.com/opencast/opencast/blob/a67ece488617d23101e16357ab5da0c69105644b/docs/checkstyle/maven-dependency-plugin.exceptions#L1-L17

You can check which dependencies are problematic by running `mvn dependency:analyze` in the module directory (you need one full Opencast build before you can do that). Note that you may need to specify exceptions since for things which are dynamically loaded at runtime (like the logger in the example above).

---

List of missing modules:

- [x] `modules/admin-ui/pom.xml` (#1904)
- [x] `modules/asset-manager-impl/pom.xml` (#2033)
- [x] `modules/asset-manager-storage-aws/pom.xml` (#2050)
- [x] `modules/common-jpa-impl/pom.xml` (#2053)
- [x] `modules/common/pom.xml` (#2053)
- [x] `modules/cover-image-impl/pom.xml` (#2053)
- [x] `modules/db/pom.xml`  (#1957)
- [x] `modules/distribution-service-aws-s3/pom.xml` (#2065)
- [x] `modules/external-api/pom.xml` (#2065)
- [x] `modules/index-service/pom.xml` (#2065)
- [x] `modules/oaipmh-persistence/pom.xml`  (#2065)
- [x] `modules/publication-service-youtube-v3/pom.xml` (#1904)
- [x] `modules/scheduler-impl/pom.xml`  (#2065)
- [x] `modules/search/pom.xml` (#1904)
- [x] `modules/search-service-impl/pom.xml`
- [x] `modules/security-shibboleth/pom.xml`  (#2065)
- [x] `modules/solr/pom.xml` 

---

For anyone working on this: Don't hesitate to submit pull requests fixing only some modules.

SijmenHuizenga:
I will give this a shot. Do you prefer a pr for each module individually or rather many together?

gregorydlogan:
Fewer PRs would be better IMO.

lkiesow:
[...]

PR #1904, opened 2020-10-03
Enforce Maven Dependency Checks on some more modules

SijmenHuizenga:
Fixed problems reported by the maven dependency plugin in the following modules:
* publication-service-youtube-v3
* admin-ui
* search

Tested the changes using the method described by @lkiesow in the comments below.

* [x] have a concise title
* [x] this pr works towards closing #1893 
* [x] pass automated testing

### DIFF
--- a/NOTICES
+++ b/NOTICES
@@ -674,15 +674,19 @@ admin-ui
   com.google.code.gson                   gson                                             Apache 2.0
   com.googlecode.json-simple             json-simple                                      The Apache Software License, Version 2.0
   commons-io                             commons-io                                       Apache License, Version 2.0
+  commons-fileupload                     commons-fileupload                               Apache License, Version 2.0
   jakarta.ws.rs                          jakarta.ws.rs-api                                EPL 2.0-GPL2 w/ CPE
   jakarta.xml.bind                       jakarta.xml.bind-api                             Eclipse Distribution License - v 1.0
   javax.servlet                          javax.servlet-api                                CDDL + GPLv2 with classpath exception
   joda-time                              joda-time                                        Apache 2
   org.apache.commons                     commons-lang3                                    Apache License, Version 2.0
+  org.apache.httpcomponents              httpcore-osgi                                    Apache License, Version 2.0
   org.apache.httpcomponents              httpclient-osgi                                  Apache License, Version 2.0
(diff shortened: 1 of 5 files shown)


lkiesow:
>  Test! (I have to figure out how to do that)

For a quick test, the following should be sufficient:

```bash
# build Opencast (-Pdev will give you a distribution which is slightly easier to test – no tarballs to unpack, …)
% mvn clean install -Pdev
% cd docs/scripts/devel-dependency-containers/
# the next command will give you all necessary runtime dependencies (search index & message broker)
# alternatively, use  docker-compose up -d  if you prefer Docker
% podman-compose up -d
% cd ../../../build/opencast-dist-develop-9-SNAPSHOT/
% ./bin/start-opencast
# wait a few seconds until the logs have calmed down
# Go to http://localhost:8080
# Use “add event” to upload a video
# If it processes successfully, everything should be fine
```

We can run a few more tests as part of the review if necessary

SijmenHuizenga:
Thanks for the instructions @lkiesow! The `publication-service-youtube-v3` is finished. Compilation takes ages on my laptop, so testing and going back to fix problems is quite slow. I will do the `admin-ui` module next.

lkiesow:
> Compilation takes ages on my laptop, so testing and going back to fix problems is quite slow.

Then here is another tip: If you have built it already once and just modified one module, you can just compile that module ans assembly the distribution again. Basically:

```
cd modules/admin-ui
mvn clean install -Pdev
cd ../../assemblies/
mvn clean install -Pdev
```

If you don't need the unit tests, you can also speed things up by adding `-DskipTests` to your `mvn` command.
There are a number of variation on this which work better or worse in different scenarios but this is the basic one and already speeds things up quite a bit ;-)

Edit: Though it seems like the unit tests on the `admin-ui` module may now fail due to removed dependencies.
Not sure which specifically but my guess would be that the tests use an internal h2 database for testing.

SijmenHuizenga:
@lkiesow Thanks a lot :heart: This is some next-level maven configuration ;) Good thing we have travis to point out these failing tests, trying to fix them atm

lkiesow:
Patch looks hood → merged.
Thanks again for the help.

Issue #1930, opened 2020-10-07
The copy-event-to-series workflow id is hard coded.


ebbertd:
**Describe the bug**
In the new LTI tools the `copy-event-to-series` workflow id is hardcoded and hence not configurable. However, every workflow id and workflow should be configurable.

**To Reproduce**
1. Change the id of the `copy-event-to-series` workflow
2. Try to copy a event to a new series in the LTI tools
3. The `copy-event-to-series` workflow will not be found

**Environment Information**
  - Opencast 9

PR #1931, opened 2020-10-07
Make copy-event-to-series workflow id configurable

ebbertd:
This makes copy-event-to-series workflow id for the new LTI tools configurable. 
This fixes issue #1930 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -105,7 +105,6 @@ public class LtiServiceImpl implements LtiService, ManagedService {
   private static final Logger logger = LoggerFactory.getLogger(LtiServiceImpl.class);
 
   private static final Gson gson = new Gson();
-  private static final String COPY_EVENT_TO_SERIES_WORKFLOW = "copy-event-to-series";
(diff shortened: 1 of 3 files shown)


modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -493,6 +493,12 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
       } else {
         this.retractWorkflowId = retractWorkflowId;
       }
+      final String copyWorkflowId = (String) properties.get("copy-workflow-id");
+      if (copyWorkflowId == null) {
+        this.copyWorkflowId = "copy-event-to-series";
+      } else {
+        this.copyWorkflowId = copyWorkflowId;
+      }
```
lkiesow: Java's [`Objects.toString(Object o, String nullDefault`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#toString-java.lang.Object-java.lang.String-) is a very convenient way of doing exactly what you did:

```suggestion
      this.copyWorkflowId = Objects.toString(properties.get("copy-workflow-id"), "copy-event-to-series");
```

ebbertd: Thanks! I changed that.

PR #1936, opened 2020-10-07
Fixed lexer grammar to parse numerics in scientific notation

phoebem:
Workflow condition parser throws exception when a number is written in scientific notation. This is a fix in the lex grammar to include it.


modules/workflow-condition-parser/src/main/antlr4/org/opencastproject/workflow/conditionparser/antlr/WorkflowCondition.g4
```diff
@@ -10,7 +10,7 @@ atom : '(' relationOperand ')' | NUMBER | STRING | BOOL ;
 AND: 'AND' ;
 OR: 'OR' ;
 NOT: 'NOT' ;
-NUMBER: '-'? [0-9]+ ('.' [0-9]+)? ;
+NUMBER: '-'? [0-9]+ ('.' [0-9]+)? ( [E|e] [+\-]? [0-9]+ )? ;
```
lkiesow: If you actually want to parse what Java allows you to parse, you might also want to allow a `+` sign in front of the first number:
```suggestion
NUMBER: [+\-]? [0-9]+ ('.' [0-9]+)? ( [E|e] [+\-]? [0-9]+ )? ;
```

lkiesow (approved):
Seems reasonable. Does make the lexer more complex and I don't really see a use case but I also don't really mind :D

Let me know if you want to add my suggestion or not.

phoebem:
Thanks Lars,

 I like your suggestion. It makes it even more general and I updated my PR.

The reason we needed this was because I added bitrate to the analyze_tracks wf so that we can tune the encoding profiles according to width, height and bitrate.
When the bitrate was higher, I ended up with something like 1.0675E7 and the workflow choked.
This is a bit of a niche feature and I don’t think anyone would care about this upstream, so it is not in the PR, 
but the definition of number should include scientific notation, so I decided to make a pull against it.

Phoebe.

lkiesow:
Ups, seems like my suggestion broke a test:
https://github.com/opencast/opencast/blob/70bacbfaa9c25f0f070543bff6eaca2472499cc6/modules/workflow-condition-parser/src/test/java/org/opencastproject/workflow/conditionparser/WorkflowConditionInterpreterTest.java#L86-L89

Though to be fair, something like `1-1` was probably already broken before

phoebem:
Hi Lars,

As you said, it was broken before my changes because it didn’t deal with negative numbers.
I spent a little time on it and conclude that to fix it properly is a bit more work than I have time right now.
I don’t know if anyone is using this module to parse math expressions, 
can we do one of the following:

1) Put in the documentation that you need to use spaces between numbers and numerical operators, add spaces to the tests.
2) Disallow negative numbers and remove the sign before number in the grammar.


Phoebe.

lkiesow:
Hi Phoebe, I think both ways will work. After all, I don't think that it's actually used. At least not much.
I don't really have a favorite.
I f I had to pick, I would probably choose the first option.
But feel free to decide on your own.

KatrinIhler:
@phoebem Sounds like this PR is just waiting on a minor fix, any idea when that is going to be ready? :)

KatrinIhler:
@phoebem Since this hasn't been worked on for 6 months, I'm gonna close this for now. Feel free to re-open once it's ready to be reviewed again! :)

PR #1954, opened 2020-10-12
Fix Paella Player assuming track is audio-only when it's actually video and audio

felix-itz:
As of now, Opencast/Paella assumes the stream's type to be the same type the last track is.
However, the track-order returned by Opencast appears to depend on various things while processing and
publishing/indexing the video so the order (audio or video last) is hard to determine.

This PR makes Paella Player assume the stream is a video stream as soon as one video track is present.

For reference, see https://groups.google.com/a/opencast.org/g/users/c/e7hP9wANC1w/m/RAwjBBF0AAAJ

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing

### DIFF
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -205,7 +205,7 @@ class OpencastToPaellaConverter {
           if (currentTrack.video) {
             currentStream.type = 'video';
           }
-          else if (currentTrack.audio) {
+          else if (currentTrack.audio && currentStream.type !== 'video') {
             currentStream.type = 'audio';
           }
         }


lkiesow:
Seems reasonable. Is this relevant for 9.x (or maybe even 8.x) as well?

felix-itz:
> Seems reasonable. Is this relevant for 9.x (or maybe even 8.x) as well?

We are using that in Opencast 8 to prevent playback a playback issue with Paella in Stud.IP:
> Error: No video streams found. Paella Player requires at least one video stream."

lkiesow:
I assume that a “it makes sense if this gets fixed in 8.x” :D
Then, can you update this pull request to go into the `r/8.x` branch?
If there is a problem, let me know. I can also quickly do that for you if you want.

felix-itz:
> If there is a problem, let me know. I can also quickly do that for you if you want.

Ooops, let me remove those commits we do not require from the PR.

felix-itz:
> Then, can you update this pull request to go into the r/8.x branch?

Looks clean now, from my perspective, although the process involved more steps than I expected.

lkiesow:
The easiest way from my experience is to run `git rebase -i r/8.x; git push --force-with-lease <remote> <branch>`.
Will test this in a moment.

PR #1963, opened 2020-10-14
Non-unique-files fix for Videogrid WOH

Arnei:
There are several problems with multiple instances of the videogrid operation running at the same time, e.g. causing event B to have the same output video as event A or output videos appearing too short. This PR attempts to fix these problems by improving on the passing and naming of files.

- Now passes track objects instead of absolute track paths to the service, so that the service can copy files into its own workspace.
- Deletes copied files when done.
- Gives output files in service unique names (by adding the current job id to the file name).

To test this pull request:
- An example workflow could look like: [test-video-grid.zip](https://github.com/opencast/opencast/files/5582178/test-video-grid.zip)
- The videogrid operation requires a SMIL file. Other than writing one yourself, the (probably) easiest way to generate one is to use the `/ingest/addPartialTrack` API, which automatically generates one for you. (I.e. create an MP with `/ingest/createMediaPackage`, add tracks to it with `/ingest/AddPartialTrack` and then start a workflow on it with `/ingest/ingest/test-video-grid`)

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing


gregorydlogan:
Do you have an example workflow here we can test with?

Arnei:
> Do you have an example workflow here we can test with?

I've updated the first comment with an example workflow.

PR #1964, opened 2020-10-14
Docs: When the sidebar is hidden, the navigation links are disabled now

### LINKED ISSUE #1899: Docs: Navigation Links still active when Sidebar is hidden
When the navigation bar has been hidden using the 
image 
button top right, all links in the sidebar on the lefthand side of the page are still clickable and active, even though they are not shown.

marwyg:
Fixes #1899

When clicking on the "hide" button in the top right corner, the toc pane on the left side also becomes hidden by adding the "hidden" class.


Arnei (approved):
Working as advertised.

PR #1968, opened 2020-10-15
Allow addMediaPackage to Accept ACLs

lkiesow:
This patch allows the single request ingest to accept access control
lists:

    curl http://localhost:8080/ingest/addMediaPackage/fast \
      -u admin:opencast \
      -F title=test \
      -F acl='{"acl": {"ace": [
          {"allow": true,"role": "ROLE_ADMIN","action": "read"},
          {"allow": true,"role": "ROLE_ADMIN","action": "write"},
          {"allow": true,"role": "ROLE_USER","action": "read"}]}}'

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
@@ -22,7 +22,9 @@
 package org.opencastproject.ingest.endpoint;
 
 import static org.apache.commons.lang3.StringUtils.trimToNull;
+import static org.opencastproject.mediapackage.MediaPackageElements.XACML_POLICY_EPISODE;
 
+import org.opencastproject.authorization.xacml.XACMLUtils;
 import org.opencastproject.capture.CaptureParameters;
 import org.opencastproject.ingest.api.IngestException;
 import org.opencastproject.ingest.api.IngestService;
@@ -45,6 +47,9 @@ import org.opencastproject.metadata.dublincore.DublinCores;
 import org.opencastproject.rest.AbstractJobProducerEndpoint;
 import org.opencastproject.scheduler.api.SchedulerConflictException;
 import org.opencastproject.scheduler.api.SchedulerException;
+import org.opencastproject.security.api.AccessControlList;
+import org.opencastproject.security.api.AccessControlParser;
+import org.opencastproject.security.api.AccessControlParsingException;
 import org.opencastproject.security.api.TrustedHttpClient;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.serviceregistry.api.ServiceRegistry;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
Boy this would have been handy a couple of weeks ago when I was trying to figure out how to set ACLs at ingest time :D

gregorydlogan:
Any chance this can be extended to `/ingest/ingest` too?

lkiesow:
I don't think it makes sense to have it in `/ingest/ingest` since that's expecting to use multiple ingest, and we would just start building it into a single-request ingest – basically, exactly what `addMediaPackage` already is.

Right now, you would add an ACL like this:

https://github.com/opencast/helper-scripts/blob/3512ea2582fd27b67cb8913ce8c89030b7ec217f/ingest/ingest-with-acl.sh#L87-L90

We already have a specific endpoint for adding Dublin Core catalogs:

https://github.com/opencast/helper-scripts/blob/3512ea2582fd27b67cb8913ce8c89030b7ec217f/ingest/ingest-with-acl.sh#L82-L85

So, maybe we could introduce `/ingest/addAclAttachment` as a simple form of adding access control lists?
In any case, I think that's out of the scope for this particular pull request.
It was basically just some old code I had lying around for a while :D

gregorydlogan:
Fair enough.  Definitely out of scope if we're touching other endpoints.  I'll hopefully be able to take a look at this shortly!

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -796,6 +804,19 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
               InputStream is = new ByteArrayInputStream(value.getBytes("UTF-8"));
               ingestService.addCatalog(is, fileName, MediaPackageElements.SERIES, mp);
 
+              // Add ACL in JSON, XML or XACML format
+            } else if ("acl".equals(fieldName)) {
+              InputStream is = new ByteArrayInputStream(value.getBytes(StandardCharsets.UTF_8));
```
CGreweling, line 809: maybe a variable, longer then two chars would be nice? "inputstream"

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -796,6 +804,19 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
...
+              AccessControlList acl;
+              try {
+                acl = AccessControlParser.parseAcl(is);
+                is = new ByteArrayInputStream(XACMLUtils.getXacml(mp, acl).getBytes(StandardCharsets.UTF_8));
+              } catch (AccessControlParsingException e) {
+                // Couldn't parse this → already XACML. Why again are we using three different formats?
+                is.reset();
```
CGreweling, line 816: logger.debug would be nice here. instead of a comentar in the code

CGreweling (commented):
looks good, just some elegance hints from me.

PR #1971, opened 2020-10-16
Add I18n support for text 'General' in 2 modals

Gao-Jun:
Add I18n support for text 'General' in 2 modals.

The translation key is already in l10n json file, maybe someone forgot to change the html.

### Your pull request should…

* [x] have a concise title
* [x] pass automated testing

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/partials/modals/edit-events-modal.html
+++ b/modules/admin-ui-frontend/app/scripts/shared/partials/modals/edit-events-modal.html
@@ -4,7 +4,7 @@
     <h2 translate="BULK_ACTIONS.EDIT_EVENTS.CAPTION"><!-- Template --></h2>
   </header>
   <wizard edit-mode="false" name="editEventsWz" on-finish="submit()" template="shared/partials/wizardNav.html">
-  <wz-step wz-title="General" canexit="editEventsForm.generalForm.$valid" wz-disabled="!editEventsForm.generalForm.$valid" ng-form="generalForm" novalidate>
+  <wz-step wz-title="{{ 'BULK_ACTIONS.EDIT_EVENTS.GENERAL.CAPTION' | translate }}" canexit="editEventsForm.generalForm.$valid" wz-disabled="!editEventsForm.generalForm.$valid" ng-form="generalForm" novalidate>
     <!-- First Step: General -->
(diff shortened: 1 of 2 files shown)


lkiesow:
Looks like this would probably make sense in 9.0. Can you make this a pull request against `r/9.x` instead to include it there?

Gao-Jun:
PR updated.
Would you please give me a guideline of which branch should my PR based?

gregorydlogan:
In general you target the oldest supported branch. Right now that's 7.x, but it's reasonable to only target 8.x now since 9.x is due in December.

The exception is features, which have to go into develop.

lkiesow:
I tend to disagree. Don't ever target the oldest supported branch unless it's really important.
Bringing more stuff into ancient versions increases the chance to break something because:
- older versions will not be tested as thoroughly
- you have a higher chance to break something while merging through the other versions
    - patches get merged `r/7.x` → `r/8.x` → `r/9.x` → `develop` starting with the oldest supported branch
- you cause more headache for the people doing the merge

So, while you may bring fixes to older versions, think about where you think it will help. I guess, most of the time, the question , in which version you want to have the fix will suffice.

For example, I was asking about `r/9.x` here since, if I remember correctly, this feature was recently introduced and it's not yet in 8.x. But bringing this to `r/9.x` means that it will be in the 9.0 release in December.

In addition to that:
- You may not modify translation keys of already released versions
- Features are only allowed in `develop` which will then automatically become the next version

Gao-Jun:
Thank you for your advice.

@lkiesow
These two modals are used at least in r/8.x, since I can confirm it on stable.opencast.org

Should I rebase my branch to r/8.x ?

lkiesow:
If it's in there and this change does not require any translation
changes, feel free to do so.

Gao-Jun:
Well, my mistake.
Only one of these file is in r/8.x
So, I think it's better to keep this change in r/9.x .
Sorry for the mistake.

gregorydlogan:
@Gao-Jun are you going to repoint this to r/9.x?

Gao-Jun:
Yes, I made a new PR #2043 
Should I close this PR?

lkiesow:
Yes, we can close this. Will review the other one shortly.
Thanks.

PR #1980, opened 2020-10-19
Harmonize defaults of serviceregistry with config file

KatrinIhler:
This is a relatively minor change that is supposed to harmonize the defaults set in the service registry with the values currently set in the config file.


lkiesow (approved):

karendolan:
I would call this pull a bug fix vs feature.

mliradelc:
I don't have strong opinions against this, is very well explained and can be changed. Also, have up to 10 fails until changing the state it seems a very good number.

gregorydlogan (approved):

PR #2016, opened 2020-10-28
Set the RFC 822 "Date" header field when sending an email

### LINKED ISSUE #2010: Emails sent by Opencast are missing the "Date" header
**Describe the bug**
When Opencast sends a mail as part of a workflow, e.g. https://docs.opencast.org/r/8.x/admin/#workflowoperationhandlers/send-email-woh/ , the email is missing the "Date" header. If the email gateway is not adding missing headers automatically, the email client may have problems correctly sorting/displaying the email.

**To Reproduce**
1. Create a workflow
2. Send a mail as part of the workflow
3. Outgoing mail does not have the "Date" header, but your mail gateway may add the header automatically. (E.g. Postfix before 2.6 added the headers automatically, Postfix 2.6 and later only adds the missing headers, if the "always_add_missing_headers" flag is "yes" and the sending client is matching in "local_header_rewrite_clients".)

**Environment Information**
 - OS: CentOS 7
 - Opencast version: 8.5

**Possible fix**
I think it should be enough to call `message.setSentDate(new Date())` in the `send` method in `modules/kernel/src/main/java/org/opencastproject/kernel/mail/BaseSmtpService.java`:

```java
  public void send(MimeMessage message) throws MessagingException {
    message.setSentDate(new Date());
    if (!productionMode) {
      logger.debug("Skipping sending of message {} due to test mode", message);
      return;
    }
    Transport t = getSession().getTransport(mailTransport);
    [...]
  }
```

See also https://javaee.github.io/javamail/docs/api/javax/mail/internet/MimeMessage.html#setSentDate-java.util.Date-

_EDIT(Lars): Fixed misspelled Opencast ;-)_

Snakel:
Should fix #2010

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/mail/BaseSmtpService.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/mail/BaseSmtpService.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Date;
 import java.util.Properties;
 
 import javax.mail.MessagingException;
@@ -243,6 +244,7 @@ public class BaseSmtpService {
    *           if sending the message failed
    */
   public void send(MimeMessage message) throws MessagingException {
+    message.setSentDate(new Date());
     if (!productionMode) {
       logger.debug("Skipping sending of message {} due to test mode", message);
       return;


lkiesow (approved):
Looks good and works as expected.
This seems like a valid bug fix for 8.x.
Do you want to make this a pull request against `r/8.x`?

_Partly related: It seems like sending mails is broken in `develop` (and likely `r/9.x`)_

Snakel:
Should I
* change this pull request to go against `r/8.x` or 
* open a new additional one for `r/8.x` and leave this one at `develop`?

lkiesow:
As long as old versions are supported, whatever you get into older versions will automatically be merged into newer versions. So, if you bring it into `r/8.x`, it will be merged into `r/9.x` and then into `develop`. So, changing this pull request is fine. If you have problems rebasing this, ping me and I can help.

Snakel:
I will open a new pull request for r/8.x

lkiesow:
That works as well though you can update pull requests to change their base nrahc.
Will take a look at the new one later.

PR #2026, opened 2020-11-01
Fix Job Dispatching Test

lkiesow:
For a long time parts of the ServiceRegistryJpaImplTest have been
failing occasionally, but on a regular basis:

    [ERROR] Tests run: 11, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.336 s <<< FAILURE! - in org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest
    [ERROR] testDispatchingJobsHigherMaxLoad(org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest)  Time elapsed: 1.45 s  <<< FAILURE!
    java.lang.AssertionError: expected:<http://thirdhost:8080> but was:<null>
	at org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImplTest.testDispatchingJobsHigherMaxLoad(ServiceRegistryJpaImplTest.java:473)

The issue seems to be more likely to pop up on slower machines or if
there is a high load on the machine when running the particular test.

The problem with the assumption that test makes is that it creates a job
and then assumes it is dispatched to a host in a particular amount of
time.  Unfortunately, there is no guarantee for the dispatching to
happen in a given time. This means that from time to time the test will
fail.

One kind of solution would be to slow the test down, making failures
less likely. Still, this does not really fix the underlying false
assumption that it's guaranteed to happen and would also overall slow
down builds which is not nice.

That is why this fix removes the final check instead. That means that
the part where it will test that it jumps out of the job barrier
remains, but checking that it is on a certain host went away.

If that's something we need, we should build a proper test for that
eventually. But for now, not having the test will cause less harm.
That's good.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated testing

### DIFF
--- a/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
+++ b/modules/serviceregistry/src/test/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImplTest.java
@@ -22,6 +22,7 @@
 package org.opencastproject.serviceregistry.impl;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 
 import org.opencastproject.job.api.Job;
 import org.opencastproject.job.api.Job.Status;
@@ -461,17 +462,7 @@ public class ServiceRegistryJpaImplTest {
             10.0f);
     JobBarrier barrier = new JobBarrier(null, serviceRegistryJpaImpl, testJob);
     launchDispatcherOnce(false);
-    try {
-      barrier.waitForJobs(JOB_BARRIER_TIMEOUT);
-      // We should never successfully complete the job, so if we get here then something is wrong
-      Assert.fail("Did not receive a timeout exception");
-    } catch (Exception e) {
-      testJob = serviceRegistryJpaImpl.getJob(testJob.getId());
-      // Some explanation here: If the load exceeds the global maximum node load (ie, jobLoad > all individual node max
-      // loads), then we dispatch to the biggest, even if it's not going to normally accept the job.  That node may still
-      // reject the job, but that's AbstractJobProducer's job, not the service registry's
-      Assert.assertEquals(TEST_HOST_THIRD, testJob.getProcessingHost());
-    }
+    assertThrows(IllegalStateException.class, () -> barrier.waitForJobs(JOB_BARRIER_TIMEOUT));
   }
 
   @Test


gregorydlogan:
So, this will fix the current issue, but that leaves a test uncovered - dispatching oversized jobs needs to succeed to prevent things from going into deadlocks.

When this exception pops up we're seeing one of two things:
1) A bad-luck roll where we caught the dispatcher in the middle of attempting to dispatch
2) Running the check before the dispatcher can run

Case 1 can be solved by shutting down the scheduler prior to checking if the job has been dispatched, but that might happen prior to the dispacher ever running (see Case 2).

Case 2 is harder to crack - we need to count how many dispatches have happened, but have no way to do that.  The `ScheduledThreadPollExecutor` makes no guarantees about timing, and does not expose a way to see what (if anything) it has done.  Nor can I see an obvious way of subclassing the `JobDispatcher` class - it's an inner class, so the closest I can get is overriding things in the test, and defining some new methods, but they're not visible outside of the `launchDispatcherOnce` method.

I do have a solution which is simple, but leaves some potentially goofy code in the main class: Add a counter to the main `JobDispatcher`.  It would be easy to add, and minimally invasive, and *solve the problem for future cases where ordering might be important*.  What do you think?  I'll do the PR :)

We could also move the `JobDispatcher` out of the inner class, but that's a lot more futzing for (arguably) minimal advantage vs adding a quick counter with rollover protection.

lkiesow:
If you want this test to actually test the dispatching, I suggest to not use the service registry at all but actually use the job dispatcher directly. The test uses the service registry to feed the dispatcher. There is no real reason for using that, is there? And to be fair, I just made that claim without actually having looked at the code again :grin: 

We discussed a few times to move the job dispatcher out of the service registry into its own module since you have to deactivate it on all servers but the admin anyway. So it seems to be sensible to just have it on the admin in the first place. If I remember correctly, @KatrinIhler was recently looking at this. No idea if that's still on the roadmap (Katrin isn't available this week).

gregorydlogan:
Ok, let's park this for now and I can bring it up next week once she's back.  Looking at the code, extracting `JobDispatcher` is going to be a fair amount of churn, and if she's already looking at it/working on it then there's no sense in duplicating effort.

gregorydlogan:
@KatrinIhler ping, are you working on `JobDispatcher`?

KatrinIhler:
@gregorydlogan Dammit, forgot to talk about this with you in the Technical Meeting. I had a note and everything!

I am currently not working on this, other things took precedence. I considered doing this as I was looking into the heartbeat-stuff for SWITCH anyway, but if it's a lot of work as you say, I think other things might be more important for them. So if you wanna take a dig at this, go ahead. ;)

gregorydlogan:
Ok, ty.  I'll take a look at this next month, hopefully.

KatrinIhler:
@gregorydlogan So if I remember correctly, extracting the JobDispatcher turned out to be a non-trivial thing. I think we should still do this sometime in the future, but what does this mean for this PR right now?

gregorydlogan:
It means I still don't like it but this is probably the best solution for now.

I'll merge this shortly.

PR #2042, opened 2020-11-04
More efficient retrieval of active jobs

lkiesow:
This patch improves the service registries mechanism to get active jobs
by not building a static list of active job states and then always
converting that to an array every time the method is called.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
@rute-santos @karendolan I know you've poked at the SR with respect to job loads.  Does this patch affect the job load data on your clusters, or induce processing overhead?  I'm looking at that `jobCache` variable, and I remember adding it to speed up performance, but I'm not 100% sure it's needed.  This patch is certainly cleaner...

lkiesow:
@gregorydlogan `jobCache` just moved a few lines down. It's still there. There shouldn't be any reason for this to slow down anything. It's just initializing something once instead of doing it over and over again with the same data

gregorydlogan:
Derp, odd that GHs formatter marked it up like that. This looks good to me, I'll merge it soon.

pascalseeland (approved):
Looks good

PR #2044, opened 2020-11-04
Add S3 presigned URL support

Gao-Jun:
This PR add S3 presigned URL support for Opencast.

Currently, when use s3 distribution to store files, anyone can download the media files from S3 or CDN if he/she can obtain the URL of these files.

By taking advantage of presigned URL, authorized users will get S3 file with a 'temporal' link with a valid duration. This will protect the media files in case its URLs are leaked.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
+++ b/etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
@@ -44,4 +44,11 @@ org.opencastproject.distribution.aws.s3.distribution.enable=false
 # If you are using another S3 service/provider, please refer to its documentation.
 #org.opencastproject.distribution.aws.s3.path.style=true
 
+# Enable presigned URL
+# Leave this commented out to use presigned URL fow AWS S3
+#org.opencastproject.distribution.aws.s3.presigned.url=true
 
+# Valid duration for presigned URL in milliseconds, default is 21600000 (6 hours)
+# For detailed information about valid duration of presigned URL,
+# see https://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html
+#org.opencastproject.distribution.aws.s3.presigned.url.valid.duration=21600000
--- a/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
+++ b/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
@@ -30,6 +30,7 @@ import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.mediapackage.MediaPackageElementBuilderFactory;
 import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.MediaPackageException;
+import org.opencastproject.mediapackage.MediaPackageSerializer;
 import org.opencastproject.security.api.AccessControlEntry;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AclScope;
@@ -51,12 +52,14 @@ import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.Dictionary;
 import java.util.Map;
@@ -90,6 +93,9 @@ public class XACMLAuthorizationService implements AuthorizationService, ManagedS
   /** The series service */
   protected SeriesService seriesService;
 
+  /** The serializer for media pacakge */
+  private MediaPackageSerializer serializer;
+
   private static final String CONFIG_MERGE_MODE = "merge.mode";
 
   /** Definition of how merging of series and episode ACLs work */
@@ -108,6 +114,11 @@ public class XACMLAuthorizationService implements AuthorizationService, ManagedS
   public void modified(Map<String, Object> config) {
     // this prevents the service from restarting on configuration updated.
     // updated() will handle the configuration update.
+  }
+
+  @Reference(cardinality = ReferenceCardinality.OPTIONAL)
+  public void setMediaPackageSerializer(MediaPackageSerializer serializer) {
+    this.serializer = serializer;
   }
 
   @Override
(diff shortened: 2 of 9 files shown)


lkiesow:
Hi @Gao-Jun, thanks for all your recent contributions. We really appreciate that!
Since it's more than a few lines of code you are adding, can you fill out an [Individual Contributor License Agreement](https://www.apereo.org/licensing/agreements/icla). If you make your contribution on behalf of an institution, you might also want to file a [Corporate Contributor License Agreement](https://www.apereo.org/licensing/agreements/ccla) (giving you as individual contributor a bit more security). Please let me know if you have any questions regarding the CLA.

If you are interested, we also like to invite you to our weekly technical meetings happening every Tuesday at 2pm UTC at https://meet.opencast.video (p: welcome). I'm not sure if that fits your schedule (and time zone) but if you are interested, we are happy to welcome you there any time.

Gao-Jun:
As a former Sakai developer and [2015 Apereo Fellow](https://www.apereo.org/groups/2015-apereo-fellows), I have already given a copy of ICLA and CCLA to Apereo Foundation in 2014. 

Two files from 2014, which take me ten minutes to find them.

[icla.pdf](https://github.com/opencast/opencast/files/5498210/icla.pdf)
[ccla.pdf](https://github.com/opencast/opencast/files/5498211/ccla.pdf)

Besides, thank you for your invitation.

2 PM UTC is 10 PM in China, I'll try to join your metting, but I'm not sure I can make it every time.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -227,6 +241,16 @@ public void activate(ComponentContext cc) {
...
 
+      // AWS presigned URL
+      String presignedUrlConfigValue = OsgiUtil.getComponentContextProperty(cc, AWS_S3_PRESIGNED_URL_CONFIG, "false");
+      presignedUrl = StringUtils.equalsIgnoreCase("true", presignedUrlConfigValue);
+      logger.info("AWS use presigned URL: {}", presignedUrl);
+
+      // AWS presigned URL expiration time in millis
+      String presignedUrlExpTimeMillisConfigValue = OsgiUtil.getComponentContextProperty(cc,
+              AWS_S3_PRESIGNED_URL_VALID_DURATION_CONFIG, null);
+      presignedUrlValidDuration = NumberUtils.toInt(presignedUrlExpTimeMillisConfigValue, DEFAULT_PRESIGNED_URL_EXPIRE_MILLIS);
```
gregorydlogan, line 255: We should be enforcing a maximum here.  Looking at the AWS docs, a simple max of 6 hours ought to do it - then we don't have to figure out what kind of credentials we are working with.

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -110,6 +116,11 @@ public void modified(Map<String, Object> config) {
     // updated() will handle the configuration update.
   }
 
+  @Reference(cardinality = ReferenceCardinality.OPTIONAL)
+  public void setMediaPackageSerializer(MediaPackageSerializer serializer) {
```
gregorydlogan, line 120: I see a total of four `MediaPackageSerializer` implementations in the codebase, two of which would always be present.  At runtime you're going to get *one* of them, but perhaps not the *right* one.  I can see this especially being true if the adopter is distributing via both S3 and Opencast URL signing since these are both implemented with a serializer.

etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
```diff
@@ -44,4 +44,8 @@ org.opencastproject.distribution.aws.s3.distribution.enable=false
 # If you are using another S3 service/provider, please refer to its documentation.
 #org.opencastproject.distribution.aws.s3.path.style=true
 
-
+# Enable presigned URL
+# Leave this commented out to use presigned URL fow AWS S3
+#org.opencastproject.distribution.aws.s3.presigned.url=true
```
gregorydlogan, line 49: There should be a blank line between config and description of the next key.

gregorydlogan (changes requested):
I haven't been able to actually test this (seeing a CNF error about the SAX parser), but it looks good.  Once I've resolved that I'll be able to test properly, but in the meantime there are a couple of minor nits address.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
Gao-Jun: You mean, set the valid duration to 6 hours and remove the AWS_S3_PRESIGNED_URL_VALID_DURATION_CONFIG ?
What about set the default value to 6 hours, and still give users the ability to change it?

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
Gao-Jun: Yes, there are several `MediaPackageSerializer`. Yet, all of these are managed by `ChainingMediaPackageSerializer`, and the `encodeURI`, `decodeURI` method will processed by all serializer one by one.
Serializers will check if the input URI should be processed by them. And `PresignedUrlMediaPackageSerializer` will only handle URI match S3 config.

etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
Gao-Jun: That's easy, I'll update the PR once the former problems solved.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
gregorydlogan: I'm fine setting the default to 6 hours, and leaving the ability to change it, but reading the [presigned url docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html) there is an absolute max duration we can create a valid URL for.  We should be trying to make sure we can't let the Opencast adopter set up requests which are invalid for their account type.

My initial read was incorrect in thinking 6 hours was the value, in actuality 7 days is the max with the right mix of credentials and signing version.  I would set the default to 6 hours, and have it check to make sure the user has not specified a duration longer than 7 days.  Ideally we could check the credential type, but I'm not entirely sure how to do that.  If you know how then adding some logic to automatically select the correct AWS-defined maximum duration would be awesome too.

Gao-Jun:
PR updated.
1. Rebased to current r/9.x branch
2. Set default valid duration to 6 hours, and check whether it exceeds 7 days.
3. Insert blank line between every config keys in config file

gregorydlogan:
@Gao-Jun if the timeout is set small enough that the end user can't actually download the whole file (ie, a 1 minute timeout on a 10GB file), what happens?  Does the download die mid-download?  I'm reworking a bit of the docs, and it occurred to me that we should be warning against too *low* a timeout to avoid this case if it will cause issues.

Gao-Jun:
According to my test (not AWS S3, but a S3 compatiable service). The link only check the start time, that means, if the user doesn't stop the download, the download process will continue even the duration is timeout.

In another case, if a user start to view a long video, and stop it, maybe paella will stop downloading the media. When the user came back after a long time, the link will be expired. But I think in such case, the user will try to refresh the page and get the video work again.

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
KatrinIhler: I think this can currently break in two ways, the first of which we are experiencing right now at SWITCH:

- This will cause workspace ACL URLs to be signed with a Stream Security policy since the signing provider only has a base URL to go on (in this case the admin URL) and that matches here, but the signing provider wasn't previously invoked in this context. This will later cause workspace and WFR URLs not to match, which will cause unnecessary reloading of files.
- As Greg said, there are multiple services providing the MediaPackageSerializer interface available via OSGI, so I don't see how you can guarantee to get the right one (or the one chaining up the rest). But then, we seem to do just that in other classes as well, and I don't understand how that works? Luck??

PR #2049, opened 2020-11-05
Endtime of segments fixed in the editor

### LINKED ISSUE #2021: Studio Input breaks Video Editor
Opening a recording from Studio in the editor does not work:

1. The editor UI is broken. The last segment is missing.
2. The end Time is set to NAN
3. Trying to process the event results in an error in the toolös endpoint since the frontend sends `Na:Na:Na:Na` as end time stamp

Screenshot from 2020-10-22 17-16-34

##### To Reproduce
0. Configure `studio-upload` to push things to the editor
1. Go to studio and record something
2. Make sure to set cut marks in Studio
3. Once it's processed, go to the editor

This happens on `8.x` and newer.

##### Possible workflow fix:

This does not seem to happen when ´schedule-and-upload` is used. It would make the whole workflow a lot simpler if `studio-upload` would just do the preprocessing, and then call `schedule-and-upload`

##### Possible Editor fix

Allow the last segment to have no end which would just translate to whatever the length of the video is or in terms of showing it in the user interface to wherever the right edge is.

marwyg:
fixes #2021 

Added a check to the "segment update" function in the editor. If the `endtime` of a segment is larger than the total duration of the video, it will be set to the end of the video.

### DIFF
--- a/modules/admin-ui/src/main/webapp/scripts/shared/directives/segmentsDirective.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/directives/segmentsDirective.js
@@ -172,6 +172,9 @@ angular.module('adminNg.directives')
              */
         scope.setHumanReadableTimes = function () {
           angular.forEach(scope.video.segments, function(segment, key) {
+            if (segment.end > scope.video.duration) {
+              segment.end = scope.video.duration;
+            }
             segment.startTime = scope.formatMilliseconds(segment.start);
             segment.endTime = scope.formatMilliseconds(segment.end);
           });


gregorydlogan:
@marwyg Is this ready to go?  Not sure what to make of the self assignment :)

marwyg:
@gregorydlogan Sorry about the self assignment. I talked to @lkiesow and he explained the current procedure of reviewing to me. So I unassigned me, because I just made the fix. The commit still has to be reviewed.

modules/admin-ui/src/main/webapp/scripts/shared/directives/segmentsDirective.js
```diff
@@ -181,6 +181,11 @@ angular.module('adminNg.directives')
              * Make sure that times are updates if needed.
              */
         scope.$root.$on('segmentTimesUpdated', function () {
+          angular.forEach(scope.video.segments, function (segment) {
+            if (segment.end > scope.video.duration) {
+              segment.end = scope.video.duration;
+            }
+          });
```
karendolan: Could this be added to the loop before it? Is there any issue that "segment.endTime" will still be old time that is greater than scope.video.duration? In other words, should the "if" check happen before segment.endTime is assigned?

gregorydlogan:
@marwyg Do you have a test workflow for this?

gregorydlogan:
@marwyg ping.

modules/admin-ui/src/main/webapp/scripts/shared/directives/segmentsDirective.js
marwyg: I don't understand exactly what you mean with "the loop before it", but the "if" check has to happen before the assignment of `segment.endTime`. I can put the "if" statement into the loop of `scope.setHumanReadableTimes()` (maybe that's what you mean). Then it wouldn't iterate 2 times over the segments.

marwyg:
I tested it only via debugging. I will try to create a workflow for it.

marwyg:
In the `studio-upload.xml` there is a key named `skipEditorPreview` https://github.com/opencast/opencast/blob/36594e0673c5f74cc5193c9a60163582cacfffe8/etc/workflows/studio-upload.xml#L109 If this is set to `false`, you can record a video with Opencast Studio (top right corner in Opencast) and it will be editable later in the event table in Opencast.

Directly after recording the video in Opencast Studio, there is the possibility to trim the video at the start and the end. When only the start was trimmed, the error occurred later in the editor in Opencast. It created a weird segment with `NaN`.

modules/admin-ui/src/main/webapp/scripts/shared/directives/segmentsDirective.js
karendolan: HI @marwyg , upon revisiting this pull, I don't see which loop I was thinking off. So I'm closing this comment!

karendolan:
Oh, I can't find the loop because you updated the pull! This looks like a totally reasonable fix.

karendolan:
@marwyg Do you know if there is any scenario where the  duration would not be set on the video or default to 0?
Does it make sense to add a sanity check on  video.duration  to be safe? Something like... 
```
            if (scope.video.duration && scope.video.duration > 0 && segment.end > scope.video.duration) {
              segment.end = scope.video.duration;
            }
```

PR #2051, opened 2020-11-05
Limit OpenMP Threads to Prevent Tesseract Blocking the System

lkiesow:
This patch is a mitigation for the new Tesseract 4.x locking up the
system during extract text operations.

We have seen systems under full load for up to half a day when running
Tesseract. This seems to be caused by Tesseract spawning far too many
threads. Limiting the OpenMP threads [1] drops the processing time down
to just a few minutes.

Note that this is not completely free of side-effects since it would
also affect other applications run by Opencast which are using OpenMP.
But none of the default ones do and is basically cannot be worse than
what we see with Tesseract right now.

[1] https://www.openmp.org/spec-html/5.0/openmpse58.html

Default events being processed on develop.opencast.org without limits:

workflow

Default events being processed on develop.opencast.org with limits:

workflow-omp-1

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/assemblies/resources/bin/setenv
+++ b/assemblies/resources/bin/setenv
@@ -39,5 +39,7 @@ export JAVA_MAX_MEM=1G
 export JAVA_PERM_MEM=256M
 export KARAF_NOROOT=true
 
+# Mitigation for new Tesseract 4.x locking up the system
+export OMP_THREAD_LIMIT=1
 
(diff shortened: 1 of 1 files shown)


gregorydlogan:
I see this kind of lockup on the regular and had put it down to my crappy ssd...

Issue #1961, opened 2020-10-14
Style guide needed for the Docs


peculiarPrince:
There should be consistency in Information Given, Structure, and Formatting (at least).

lkiesow:
Some random notes from the meeting:

- [ ] step by step
  - make it actually step by step
  - no important notes at the bottom
- [ ] activemq is a mess
- [ ] make clear recommendations
- [ ] clear distinction between running it on one machine and running it on multiple machines
- [ ] clearly mark optional stuff as optional
  - clearly mark non-optional stuff as non-optional

peculiarPrince:
Issue #1115 falls under "Mark optional or required"

lkiesow:
Pull request #2032 identified a number of issues in the current documentation.
They are marked by `<!-- _`:

```
docs/modules/awss3archive.md:5:<!-- _Add "Note: " here? -->
docs/configuration/basic.md:4:<!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
docs/configuration/message-broker.md:3:<!-- _"Since the release of version 2.x, ..." -->
docs/configuration/message-broker.md:25:<!-- _This is where things start to go up in flames. -->
docs/configuration/message-broker.md:40:<!-- _This step is already mentioned AND explained more clearly in the rpm-el8.md doc at "Install Apache ActiveMQ". -->
docs/configuration/message-broker.md:44:`/usr/share/doc/activemq/README.Debian`. <!-- _This might be better on a new line/list items or as a note at the beginning of the paragraph, "not all OSs configure the same: -Most do this, -Debian does this" -->
docs/configuration/message-broker.md:47:<!-- _Instance of a difference between single and multiple servers/machines? -->
docs/configuration/message-broker.md:52:<!-- _Please be more explicit about when the user will be working in which directory (OC's or ActiveMQ's), e.g. state where to find the directory, or state beforehand that "until stated otherwise, the mentioned config files are found in ActiveMQ's config directory."  -->
docs/configuration/message-broker.md:54:<!-- _Unclear/Check:: Not sure if this is correct information: which `custom.properties` file is meant, OC's or ActiveMQ's? Can the location be given (i.e. where to find the directory as shown in rpm-el8.md "Install Apache ActiveMQ") -->
docs/configuration/message-broker.md:62:<!-- _Unclear/Check:: In which file/directory are we finding this config setting, OC's or ActiveMQ's (I assume it's ActiveMQ)? -->
[...]

PR #2066, opened 2020-11-10
Update Basic Configuration

lkiesow:
This patch addresses some issues with the basic configuration guide being
unclear or giving several options without any advice when to use these.

This is related to #1961

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/configuration/basic.md
+++ b/docs/guides/admin/docs/configuration/basic.md
@@ -1,108 +1,114 @@
 Basic Configuration
 ===================
 
-<!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
-as the Installation Docs do (i.e. Install doc refers to Config doc) -->
-This guide will help you to change the basic configuration settings which are required or at least strongly recommended
-for each Opencast installation. This is basically what you should do, right after installing Opencast on your machine.
+The basic configuration guide will help you to adjust the settings strongly recommended for each Opencast installation.
+This is what you should do right after installing Opencast.
+While there are alternatives for some of these settings, this is the recommended setup.
 
-<!-- _"In most cases, ..." Unclear: the cases being the different Operating Systems? -->
-All settings changes are made to files residing in the Opencast configuration directory. In most cases, that should be either
-`/etc/opencast/` or `/opt/opencast/etc/`. Edit the files using the editor of your choice, e.g.:
-
-    vim /etc/opencast/custom.properties
+All settings changes are made to files residing in the Opencast configuration directory.
+The location of the configuration directory depends on how you installed Opencast.
+If you used the Linux packages, the location is `/etc/opencast`.
 
 
 Step 1: Setting the Server URL
 ------------------------------
 
-By default, only connections from the local machine are accepted by Opencast.  You want to change this if the system
-should be accessible within a network.
+Find the property `org.opencastproject.server.url` in `etc/custom.properties` and set your domain name.
+The value must be set to the URL from which the server can be accessed later.
 
-First, find the property `org.opencastproject.server.url` in your `custom.properties` configuration file and set it to
-your own domain name:
+    org.opencastproject.server.url=https://example.opencast.org
 
-    org.opencastproject.server.url=http://example.com:8080
-
-*Note:* This value will be written to all generated mediapackages and thus cannot be changed easily for already
-processed media, at least not without an extra amount of work involving modifications to the database. That is why you
-should think about this setting carefully.
+*Note:* This value will be written to all generated media packages and thus cannot be changed easily for already
+processed media. Please think about this setting carefully.
 
 
-Step 2: Setting the Login Details
----------------------------------
+Step 2: Setting Authentication Details
+--------------------------------------
 
-There are two authentication methods for Opencast: HTTP Digest authentication and form-based authentication. Both
-methods need a username and a password. Change the password for both! The important keys in the 'custom.properties'
-configuration file are:
+Configure authentication and security details of Opencast, including the login credentials.
+For this, the important keys in the `etc/custom.properties` configuration file are:
 
 * `org.opencastproject.security.admin.user`
     * The user for the administrative account. This is set to `admin` by default.
 * `org.opencastproject.security.admin.pass`
     * The password for the administrative account. This is set to `opencast` by default.
 * `org.opencastproject.security.digest.user`
-    * The user for the communication between Opencast nodes, as well as for capture agents. This is set to
-    `opencast_system_account` by default.
+    * The user for the communication between Opencast nodes. It is sometimes also used by capture agents.
+      This is set to `opencast_system_account` by default.
 * `org.opencastproject.security.digest.pass`
-    * The password for the communication between Opencast nodes and capture agents. This is set to `CHANGE_ME` by default.
+    * The password for the communication between Opencast nodes. It is sometimes also used by capture agents.
+      This is set to `CHANGE_ME` by default.
+* `karaf.shutdown.command`
+    * The security token used for shutting down Opencast. Set this to a random string.
 
-*Note:* The digest credentials are also used for internal communication of Opencast servers. So these keys have to be
-set to the same value on each of your Opencast nodes (Core, Worker, Capture Agent, …)
-
-
-Step 3: Change the default shutdown command
--------------------------------------------
-
-Karaf provides a socket over which you can send a shutdown command. The socket does not provide any kind of
-authentication. Therefore anyone who obtains write access to this socket is able to shutdown karaf and everything
-that runs on it. There is a default `karaf.shutdown.command` defined in `custom.properties`. Change this to something
-secret.
+Make sure that these settings are identical on all nodes of the cluster.
 
 
-Step 4: Setting up Apache ActiveMQ Message Broker
+Step 3: Setting up Apache ActiveMQ Message Broker
 -------------------------------------------------
-<!-- _This is a very nonchalant and subdued description for how important this config probably is -->
-Since version 2.0, Opencast requires a running Apache ActiveMQ instance with a specific configuration.  The message
-broker is mostly run on the admin server of Opencast but can be run separately. It needs to be started before Opencast.
-For more details about the setup, have a look at the [Apache ActiveMQ configuration guide](message-broker.md).
+
+Opencast requires Apache ActiveMQ to relay messages between micro-services.
+For configuration details, please follow the:
+
+- [Apache ActiveMQ configuration guide](message-broker.md)
 
 
-Step 5: Database Configuration
+Step 4: Database Configuration
 ------------------------------
 
-Opencast uses an integrated H2 database by default. While you will find it perfectly functional, it has certain
-drawbacks:
+Opencast uses an integrated H2 database by default, which has certain drawbacks:
 
-* It is rather slow
 * It cannot be used for distributed set-ups
 * Upgrading Opencast with this database is not possible
 
-For testing, it is totally fine to keep the internal database, but you are highly encouraged to switch to a stand-alone
-database for productional use. For more information about database configuration, have a look at the [Database
-Configuration](database.md) section.
+The internal database will suffice for testing, however a stand-alone database is required for production uses.
+Details about the configuration can be found at:
 
-Step 6: Elasticsearch Configuration
+- [Database Configuration](database.md)
+
+
+Step 5: Elasticsearch Configuration
 -----------------------------------
 
-Opencast requires Elasticsearch.
-For more details about the setup, have a look at the [Elasticsearch configuration
-guide](../modules/searchindex/elasticsearch.md).
+Opencast requires Elasticsearch for quick full-text search and data caching.
+For more details about the setup, have a look at the:
 
-Step 7: HTTPS Configuration
+- [Elasticsearch configuration guide](../modules/searchindex/elasticsearch.md).
+
+
+Step 6: HTTPS Configuration
 ---------------------------
-<!-- _Every other instance of linking to another doc so far has been a "do now, but in another doc", 
-whereas here it is very much "not necessary for right now": what to do? -->
-An installation without HTTPS does not make much sense today.
-Thus, make sure to follow [a configuration guide for HTTPS](https/index.md).
 
-Step 8: Setting the Storage Directory (optional)
+This configuration is required in order to:
+
+- Make Opencast available externally
+- Secure connections from/to Opencast
+
+For this, follow one of the
+
+- [configuration guides for HTTPS](https/index.md).
+
+
+Step 7: Setting the Storage Directory (optional)
 ------------------------------------------------
 
-Even though it is not important for all systems – on test setups you can probably omit this – you will often want to set
-the storage directory. This directory is used to store all media, metadata, … Often, an NFS mount is used for this. You
-can set the directory by changing `org.opencastproject.storage.dir` like:
+If you want to use a specific location for storing media, metadata and other data,
+you can set the directory by changing `org.opencastproject.storage.dir`.
 
-    org.opencastproject.storage.dir=/media/mhdatamount
+    org.opencastproject.storage.dir=/path/to/data/folder
 
-<!-- _Instance of "implied Note" -->
-Please keep in mind that the user running Opencast must have read/write permissions to the storage directory.
+Often, an NFS mount is used for data storage.
+Make sure that the user running Opencast has read/write permissions to the storage directory.
+You can check that, for example, by running:
+
+```no-highlight
+sudo -u opencast touch /path/to/data/folder/test
+sudo -u opencast rm /path/to/data/folder/test
+```
+
+
+Finish Installation
+-------------------
+
+If you came here as part of an installation, please head back to the installation guide you used for notes on how to run
+Opencast as a service.


docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
 
 <!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
 as the Installation Docs do (i.e. Install doc refers to Config doc) -->
-This guide will help you to change the basic configuration settings which are required or at least strongly recommended
-for each Opencast installation. This is basically what you should do, right after installing Opencast on your machine.
+This guide will help you to change the basic configuration settings required or at least strongly recommended for each
+Opencast installation. It is basically what you should do, right after installing Opencast on your machine.  There are
```
peculiarPrince: no comma needed after "should do"

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
...
-For more details about the setup, have a look at the [Apache ActiveMQ configuration guide](message-broker.md).
 
-
-Step 5: Database Configuration
+Step 4: Database Configuration
 ------------------------------
 
-Opencast uses an integrated H2 database by default. While you will find it perfectly functional, it has certain
-drawbacks:
+Opencast uses an integrated H2 database by default which has certain drawbacks:
```
peculiarPrince: comma after "by default"

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
...
 Step 1: Setting the Server URL
 ------------------------------
 
-By default, only connections from the local machine are accepted by Opencast.  You want to change this if the system
-should be accessible within a network.
-
-First, find the property `org.opencastproject.server.url` in your `custom.properties` configuration file and set it to
-your own domain name:
+First, find the property `org.opencastproject.server.url` in `etc/custom.properties` and set your domain name.
+The value should be the URL from which the server can be access later.
```
peculiarPrince: suggest "This value will be the URL..." instead

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
...
 
 Step 1: Setting the Server URL
 ------------------------------
 
-By default, only connections from the local machine are accepted by Opencast.  You want to change this if the system
-should be accessible within a network.
-
-First, find the property `org.opencastproject.server.url` in your `custom.properties` configuration file and set it to
-your own domain name:
+First, find the property `org.opencastproject.server.url` in `etc/custom.properties` and set your domain name.
```
peculiarPrince: suggest taking out "First", or adding "Next/Then/After that" to later steps. otherwise this feels out of place.

docs/guides/admin/docs/configuration/basic.md
lkiesow: The setting will not influence the location but must be set to the location. Maybe this is better?

```suggestion
The value must be set to the URL from which the server can be accessed later.
```

peculiarPrince: Yes!

lkiesow:
I've incorporated the suggested changes and updated the pull request.

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
 
 <!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
 as the Installation Docs do (i.e. Install doc refers to Config doc) -->
-This guide will help you to change the basic configuration settings which are required or at least strongly recommended
-for each Opencast installation. This is basically what you should do, right after installing Opencast on your machine.
+This guide will help you to change the basic configuration settings required or at least strongly recommended for each
+Opencast installation. It is basically what you should do right after installing Opencast on your machine.  There are
```
gregorydlogan: `It is basically` -> `This is` ?

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
...
+
+Step 6: HTTPS Configuration
 ---------------------------
-<!-- _Every other instance of linking to another doc so far has been a "do now, but in another doc", 
-whereas here it is very much "not necessary for right now": what to do? -->
-An installation without HTTPS does not make much sense today.
-Thus, make sure to follow [a configuration guide for HTTPS](https/index.md).
 
-Step 8: Setting the Storage Directory (optional)
+To make Opencast available externally and securing connections to it, follow one of the
```
gregorydlogan: `To make Opencast available externally and securing connections to it, ` -> `To secure connections to Opencast`

mliradelc: `To make Opencast available externally ` This part should be even more clear, there was an adopter with this trouble because the guide didn't clear that to have external connections, YOU NEED* to configure HTTPS, otherwise, only localhost can access. 

\* Yes you can configure `org.ops4j.pax.web.cfg` but is not recommended, That should go in a troubleshooting guide or something like that.

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -3,106 +3,102 @@ Basic Configuration
...
-For more details about the setup, have a look at the [Elasticsearch configuration
-guide](../modules/searchindex/elasticsearch.md).
+Opencast requires Elasticsearch for quick full-text search and data caching.
+For more details about the setup, have a look at the:
 
-Step 7: HTTPS Configuration
+- [Elasticsearch configuration guide](../modules/searchindex/elasticsearch.md).
+
+
+Step 6: HTTPS Configuration
```
mliradelc, line 79:
```suggestion
Step 6: HTTPS Configuration (Add external access)
```

docs/guides/admin/docs/configuration/basic.md
gregorydlogan: You don't *need* to configure https, it's sane to provide access via http-only.  It's likely not something that most adopters will want to do, but we shouldn't make it sound like https is required when it's not.

docs/guides/admin/docs/configuration/basic.md
lkiesow: fixed

docs/guides/admin/docs/configuration/basic.md
lkiesow: For the sake of clarity, this guide should provide one way forward, not multiple ways. We should assume that new users do not know what they want (otherwise, they wouldn't read this guide). Providing options like “you can do this but if you don't do it, you need to do something else instead which is something you wouldn't need to do if you did the first thing” just makes the instructions way more confusing and will leave people in a state of “what the hell do I really need to do”.

The first paragraph of this guide states that there are alternatives to what is described in here but this is the recommended set-up. Users should not skip any step except if it is explicitly marked as optional.

I'll try to make this sentence more clear though.

EDIT: And remembering Studio right now, I think parts of Opencast will not even work any longer in a non-local HTTP set-up. I think the web video stuff needs HTTPS iirc.

docs/guides/admin/docs/configuration/basic.md
lkiesow: Let's name it <del>“Reverse Proxy Configuration”. That includes both :D</del>
…isn't such a great idea after all. Let me think about that.

lkiesow:
Updated the docs and (hopefully) made it even more clear that HTTPS is non-optional in this guide.
@peculiarPrince, can you shortly read through this again and check if this is still understandable?
To the others, any other complaints?

gregorydlogan:
I would be tempted to add the note about non-local http not working for all features to the section about https - I did not know that, and thus was about to write another complaint about leading users towards https when they might not need it :)

lkiesow:
Studio over HTTP in Chromium:

---

xy

---

But it feels like we have different target audiences in mind for this guide. I don't see this as a guide catering to people who know it better and seek shortcuts for testing things. It's true that you don't necessarily need everything in this guide: You can run tests without changing the passwords or the shutdown command. If you are testing locally, you don't need to change the server address. But following the steps in here is a reasonable approach to set up a system without easily running into problems or security issues.

For example, even if Studio would not be broken or if you wouldn't run into problems with integrations since basically everything runs in a secure context nowadays and has problems with including non-HTTPS content, what will we really gain from telling people that they don't need HTTPS? A first quick test may be faster. But they will easily miss these steps later causing serious problems.

For developers testing things, that's an entirely different matter. I often set up parts of a system if that is sufficient for my test. If I just want to test the `runtime-info-ui`, I run a worker with no admin. That will work. Should we tell people that they can do that? Likely not, or they will be pretty confused what they should do with just a worker :D

For me, this guide is for new or at least less experienced adopters who want to set up Opencast for testing or production (we nowhere really distinguish between those set-ups – and I don't think we should). I don't think it's a good idea to present a lot of different options to these users. Instead, we should nudge them to what will most likely work regardless of what they finally do with the system. That's why I want to avoid all unnecessary choices. If you say “you can do this or do that”, they don't necessarily know what is best in their situation or what they need in the long run. If we know what they will need in the long run, let's recommend that from the start.

The drawback may seem that the guide is slightly longer and may make people configure something they might not necessarily need in a specific set-up. But that's an illusion since we would have to document both choices otherwise. So you may not have to follow as many steps, but you have to carefully read more steps.

I also think the guide is reasonably short so that people will still actually read it and not skip over things. That's, by the way, also why I've tried to avoid long texts or explanations about items in things guide: If I write a novel about why Opencast uses ActiveMQ and hide crucial information in a long text, people might accidentally skip it. That's why I went with the short, precise “now install this” approach.

_I'd like to say “that are just a few thoughts on this” but this turned out longer than expected_ :smile_cat:

gregorydlogan:
Yep that's fair.  Let's leave this as is.  Approved from my end, @peculiarPrince ?

peculiarPrince:
> But it feels like we have different target audiences in mind for this guide. 

I think this has been a problem from the start. As is, the intended audience changes from doc to doc, and sometimes within the same doc. It would be really good to set the target audience for the user guide and edit the rest of the docs with that audience in mind, and to transfer all the sidenotes about tweaking or possible changes to a "For Those Who Want to Play With OC (and Hopefully Know What They're Doing)" in a seperate guide or set of documentation. 

> I don't see this as a guide catering to people who know it better and seek shortcuts for testing things [...] For me, this guide is for new or at least less experienced adopters who want to set up Opencast for testing or production (we nowhere really distinguish between those set-ups – and I don't think we should). I don't think it's a good idea to present a lot of different options to these users. Instead, we should nudge them to what will most likely work regardless of what they finally do with the system

+1

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -1,108 +1,114 @@
 Basic Configuration
 ===================
 
-<!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
-as the Installation Docs do (i.e. Install doc refers to Config doc) -->
-This guide will help you to change the basic configuration settings which are required or at least strongly recommended
-for each Opencast installation. This is basically what you should do, right after installing Opencast on your machine.
+This guide will help you to adjust the basic configuration settings strongly recommended for each Opencast installation.
+This is what you should do right after installing Opencast.
+There are alternatives for some of these settings but this is the recommended setup.
```
peculiarPrince:
```suggestion
While there are alternatives for some of these settings, this is the recommended setup.
```
Additionally, if there are going to be other guides for those who want to configure alternatives, that could be mentioned/linked here, a la `For those looking for guides on alternative setups, see *link*`

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -1,108 +1,114 @@
 Basic Configuration
 ===================
 
-<!-- _Guide says nothing about how to start running OC, nor does it point to the "Next Step"
-as the Installation Docs do (i.e. Install doc refers to Config doc) -->
-This guide will help you to change the basic configuration settings which are required or at least strongly recommended
-for each Opencast installation. This is basically what you should do, right after installing Opencast on your machine.
+This guide will help you to adjust the basic configuration settings strongly recommended for each Opencast installation.
```
peculiarPrince: I'm unsure if `This guide` refers to this, the Basic Config doc, or to the OC Documentation as a whole, or ... both?

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -1,108 +1,114 @@
...
+Opencast uses an integrated H2 database by default, which has certain drawbacks:
 
-* It is rather slow
 * It cannot be used for distributed set-ups
 * Upgrading Opencast with this database is not possible
 
-For testing, it is totally fine to keep the internal database, but you are highly encouraged to switch to a stand-alone
-database for productional use. For more information about database configuration, have a look at the [Database
-Configuration](database.md) section.
+For testing, it is fine to keep the internal database, but configure a stand-alone database for production.
```
peculiarPrince: `For testing, it is fine to keep the internal database` -> `For testing, the internal database will suffice,`

`but configure a stand-alone database for production.` -> `but for production, a stand-alone database must/should be configured.`

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -1,108 +1,114 @@
...
 
-Step 8: Setting the Storage Directory (optional)
+Opencast requires configuration to:
+
+- Make Opencast available externally
+- Secure connections from/to Opencast
+
+For this, follow one of the
+
+- [configuration guides for HTTPS](https/index.md).
```
peculiarPrince, line 89: Hm... this actually links to a single page "Serve Content Via HTTPS" which then links to the three options of configuration
```suggestion
- [HTTPS configuration guides listed here](https/index.md).
```

docs/guides/admin/docs/configuration/basic.md
```diff
@@ -1,108 +1,114 @@
...
-Step 7: HTTPS Configuration
+Step 6: HTTPS Configuration
 ---------------------------
-<!-- _Every other instance of linking to another doc so far has been a "do now, but in another doc", 
-whereas here it is very much "not necessary for right now": what to do? -->
-An installation without HTTPS does not make much sense today.
-Thus, make sure to follow [a configuration guide for HTTPS](https/index.md).
 
-Step 8: Setting the Storage Directory (optional)
+Opencast requires configuration to:
```
peculiarPrince:
```suggestion
This configuration is required in order to:
```

docs/guides/admin/docs/configuration/basic.md
gregorydlogan: I would read `this guide` as referring to this specific page, rather than the OC docs as a whole.  Perhaps `This page` instead of `This guide` would clarify?

docs/guides/admin/docs/configuration/basic.md
gregorydlogan: Hah, so I thought about commenting on this, but decided against since it's not *technically* wrong.  I would rewrite this as
```The internal database will suffice for testing, however for production uses a stand-alone database is required```

Optionally we could include a list of databases, but I would expect anyone reading this to know what their options and preferences are.

Comment (peculiarPrince) [reply docs/guides/admin/docs/configuration/basic.md]: You're right, it's not technically wrong x) but the OG syntax muddles the message a bit, hence the comment.

How about
`The internal database will suffice for testing, however a stand-alone database is required for production uses.`
?

Comment (gregorydlogan) [reply docs/guides/admin/docs/configuration/basic.md]: That's a cleaner way of saying it, +1

File: docs/guides/admin/docs/configuration/basic.md (hunk 9 above)
Comment (peculiarPrince) [reply docs/guides/admin/docs/configuration/basic.md]: As far as I can tell, the docs use `This guide`, `This page`, and `This document` pretty interchangeably. So I suppose it can stay as is.

File: docs/guides/admin/docs/configuration/basic.md (hunk 8 above)
Comment (lkiesow) [reply docs/guides/admin/docs/configuration/basic.md]: > Additionally, if there are going to be other guides for those who want to configure alternatives

There aren't any complete guides. That information is listed on the specific modules' documentation pages and people will need to look up what they need there.

File: docs/guides/admin/docs/configuration/basic.md (hunk 9 above)
Comment (lkiesow) [reply docs/guides/admin/docs/configuration/basic.md]: Just explicitly named this page:

```diff
diff --git a/docs/guides/admin/docs/configuration/basic.md b/docs/guides/admin/docs/configuration/basic.md
index aa5fcf8dcf..d7821eecaf 100644
--- a/docs/guides/admin/docs/configuration/basic.md
+++ b/docs/guides/admin/docs/configuration/basic.md
@@ -1,7 +1,7 @@
 Basic Configuration
 ===================
 
-This guide will help you to adjust the basic configuration settings strongly recommended for each Opencast installation.
+The basic configuration guide will help you to adjust the settings strongly recommended for each Opencast installation.
 This is what you should do right after installing Opencast.
 There are alternatives for some of these settings but this is the recommended setup.
 

```

docs/guides/admin/docs/configuration/basic.md
lkiesow: Went with

> The internal database will suffice for testing, however a stand-alone database is required for production uses.

lkiesow:
Updated and rebased to the current state of develop.

docs/guides/admin/docs/configuration/basic.md
lkiesow: I like to keep this as it is

peculiarPrince (approved):

lkiesow:
Yay, approval :tada:

PR #2068, opened 2020-11-12
JDK Support proposal (2020-11-11) documentation for 8.x

gregorydlogan:
This PR adds the relevant JDK support documentation for 8.x.  A further PR will be filed after this has been merged forward.

### DIFF
--- a/docs/guides/admin/docs/installation/source-linux.md
+++ b/docs/guides/admin/docs/installation/source-linux.md
@@ -72,12 +72,6 @@ Required for animate service (optional):
     synfig
 
 
-Java version support
---------------------
-
-The only supported Java version for Opencast 8 is JDK 8.  Newer versions will not work with Opencast 8, nor will older versions.
-
-
 ### Dependency Download
 
 Pre-built versions of most dependencies that are not in the repositories can be downloaded from the respective project
--- a/docs/guides/admin/docs/installation/source-linux.md
+++ b/docs/guides/admin/docs/installation/source-linux.md
@@ -71,6 +71,13 @@ Required for animate service (optional):
 
     synfig
 
+
+Java version support
+--------------------
+
+The only supported Java version for Opencast 8 is JDK 8.  Newer versions will not work with Opencast 8, nor will older versions.
+
+
 ### Dependency Download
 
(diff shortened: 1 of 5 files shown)


docs/guides/admin/docs/installation/source-linux.md
```diff
@@ -71,6 +71,13 @@ Required for animate service (optional):
 
     synfig
 
+
+Java version support
```
lkiesow: Isn't this already in here a few lines above this:

> Please make sure to install the following dependencies.
> Required:
>
>    java-1.8.0-openjdk-devel.x86_64 / openjdk-8-jdk

lkiesow (commented):
The installation from source guide already specified the exact version and the packages should usually take care of the required version. Of course, there is the chance that multiple JDKs are installed but that's still easy to miss and can cause problems.  I'm wondering if it would make sense to add something like this to the [startup check script we already have](https://github.com/opencast/opencast/blob/r/8.x/assemblies/resources/bin/check_ports) and not put these comments here, especially since they might also cause people to try installing JDKs separately instead of just installing the package:

```
if ! java -version 2>&1 | grep -q '"1\.8\.'; then
  echo "ERROR: Opencast 8 requires OpenJDK 8
  exit 1
fi
```

That should cause an immediate error with a clear error message.
What do you think?

docs/guides/admin/docs/installation/source-linux.md
gregorydlogan: I deliberately added it again to emphasise that JDK 8, and only 8, is required.  That being said, your solution makes things even more clear, so I'll remove this part.  If you're installing from source and can't follow the directions closely enough then that's not really something we can fix :)

mliradelc:
For me is Ok

mliradelc (approved):
All ok

lkiesow:
Not sure if putting the test directly into `setenv` is the best place since it's more a configuration file but given that this is specifically for Opencast 8 anyway, I don't think we need to argue too much about that. Tested this with Java 8 and Java 11 and it behaves as expected. Merging this into `r/8.x` and reverting this in `r/9.x`.

PR #2075, opened 2020-11-16
Reduce IO Load When Starting a Workflow

lkiesow:
Workflow properties are persisted in the asset manager when a new
workflow is started. This requires that a snapshot of the event is
already present in the asset manager. Since this is not true for
unscheduled events, a snapshot is created on-the-fly.

Unfortunately, this can delay ingests if the storage system is under
high load and can even cause ingests to fail completely.

This patch modifies the behavior by creating a snapshot of an empty
media package instead. This significantly reduces the stress put onto
storage systems.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
@lkiesow checkstyle failure, extra import.

lkiesow:
Fixed

Issue #1892, opened 2020-09-30
iFrameResizer integration into Opencast


orbruno:
Opencast integrated into the LMS OpenOlat via LTI. The fixed height of iFrames always leads to ugly scrollbars.

In other LTI integrations problem has been solved with the library iFrameResizer. With this solution, both the page hosting the iFrame and the page inside the iFrame (in this case Opencast) must include a small JavaScript library. iFrameResizer communicates changed in page height of the iFrame content to the host page and automatically changes the height of the iFrame. So no scrollbars are shown.

iFrameResizer is very widespread and is probably also used in other LMS.

Is there a possibility to integrate iFrameResizer into Opencast. Either as an integral part of Opencast LTI pages or as a manual extension of a specific Opencast instance?

PR #2080, opened 2020-11-18
Add iFrame Resizer to LTI Tools

orbruno:
This patch adds the [iFrame Resizer](https://github.com/davidjbradshaw/iframe-resizer) to the LTI tools. This allows integrations to dynamically adjust the IFrame size to the size of the included content, making it look more like a native tool.


modules/lti/public/index.html
```diff
@@ -30,9 +30,10 @@
 
       You can add webfonts, meta tags, or analytics to this file.
       The build step will place the bundled scripts into the <body> tag.
-
+
```
KatrinIhler: Unintentional whitespace change!

modules/lti/public/iframeResizer.contentWindow.min.js
```diff
@@ -0,0 +1,10 @@
...
+//# sourceMappingURL=iframeResizer.contentWindow.map
```
KatrinIhler, line 10: Whats up with this line?

lkiesow: That's how a minified JavaScript library often looks like… ?

modules/lti/public/index.html
lkiesow:
```suggestion

```

modules/lti/public/iframeResizer.contentWindow.min.js
KatrinIhler: Okay, I was confused why it was commented out, but then nevermind. :D

lkiesow:
This fixes  #1892

LukasKalbertodt (changes requested):
I tried this locally and it works... kind of. 

The series tool is a bit broken because the `<header>` (e.g. "Results 1-2 of 2") has a margin that pushes the `<body>` down without increasing its height. That leads to the bottom of the page being hidden (half of the pagination buttons are hidden). And since this library also sets `overflow: hidden`, users cannot even scroll anymore. That's unfortunate.

</img>


This could be fixed by either fixing our CSS or using [this](https://github.com/davidjbradshaw/iframe-resizer/blob/master/docs/iframed_page/options.md#heightcalculationmethod--widthcalculationmethod) (which curiously also says one should first try fixing the CSS :D). This should certainly be fixed before merging this.

The iframe also doesn't seem to resize when clicking on a video, which is unexpected.

---

Finally, this is an NPM project which means we have a proper dependency management. We shouldn't include minified JS files in our repository and include them directly in the HTML. Instead, `npm install iframe-resizer --save` adds the dependency and adding `import 'iframe-resizer';` in `src/index.ts` includes it into the project. (It's worth noting that the iframe-resizer project description really doesn't explain this part very well...).

I tried this locally and this works, so that's certainly the preferred way of including the library.

LukasKalbertodt:
Superseded by #2562. This PR has some problems still and we don't expect the author to address those, so I just created a new PR.

PR #2083, opened 2020-11-20
Reverting part of #1291 because this code is actually needed

### LINKED ISSUE #1860: [Confirmed][BLOCKER BUG] Endpoint /search/add crashes in presenter Node
**Describe the bug**
From the last 3 weeks, I found a very nasty bug when I was reviewing a PR using a cluster with 3 nodes (Admin, worker, presenter).

The bug appears when the endpoint /search/add is used, this is the error message:

```
> (InjectionUtils:568) - Parameter Class org.opencastproject.mediapackage.MediaPackageImpl has no constructor with single String parameter, static valueOf(String) or fromString(String) methods
```

I tried again using a fresh install of Opencast from the development branch (Using the default security and configuration for nodes) and I had the same problem.

Trying to use the debbuger doesn't works because this error comes before calling the lines of the search/add endpoint  

**To Reproduce**
1. Install opencast 'develop' as a cluster
2. Try to upload a new video

Will fail in the publish-engage step

*Alternatively:*
1. Enter to the presenter mode
2. Go to the search endpoint webinterface
3. Try the `search/add` endpoint with the default example.

Normally should give a 400 error (Bad request), But instead gives a 500 error (Internal server error).

**Environment Information**
 - OS: CentOS 7 (3 Servers in the same network)
 - Opencast develop branch
- Run and built with Java OpenJDK Runtime Environment (build 1.8.0_262-b10)

**Linked OC groups thread:** https://groups.google.com/a/opencast.org/g/dev/c/3dhlEbXtNKo

gregorydlogan:
The MediaPackageImpl class does not demarshall from XML without this method, and it was accidentally cleaned up
during a cleanup purge.  This commit reverts that change, resolving #1860

Fixes #1860

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageImpl.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageImpl.java
@@ -31,6 +31,7 @@ import org.opencastproject.mediapackage.identifier.IdImpl;
 import org.opencastproject.util.DateTimeSupport;
 import org.opencastproject.util.IoSupport;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,6 +39,7 @@ import org.w3c.dom.Node;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


lkiesow:
Yay, obscure usage of possibly different methods for XML stuff :disappointed: 
Anyway, this patch looks reasonable → merging.

mliradelc:
Thanks @gregorydlogan and @lkiesow for taking note of this! :smile: Works perfectly now

karendolan:
Hi @gregorydlogan, how do the other classes that used to use JAXB to marshall items to XML marshall items to XML? For example the workflow set. Do the XML format request APIs still exist, or is it only JSON now?

gregorydlogan:
@karendolan All of our REST APIs use (AFAIK) JAXB to do their serialization, which can spit out XML or JSON depending on settings.  Internally JAXB does everything in XML, converting to JSON as needed. This can lead to fun bugs like JAXB demanding namespace settings in JSON, but that's for another time :)

If you're seeing something which *isn't* using JAXB for its serialization please file an issue (IMO).  If you're seeing other errors similar to this also please file an issue and assign it to me - everything that was working previously should *still* be working with this changeset.

PR #2103, opened 2020-11-26
Only persist users with specific LTI role

### LINKED ISSUE #1690: LTI User Persistence creates too many users
**Describe the bug**
Per Ravishel Naicker on the users list, they have 30K+ students authenticating via LTI, with each one creating a new Opencast user in the database.  The majority of these users are viewer-only users, and should not exist in the database given their permissions.  Instead a regex, or other allow-list based system should be implemented to only persist users who need to be persisted (ie, Instructor roles).

marwyg:
fixes #1690 

With this change, only users with specific LTI roles will be persisted. The roles can be configured in `etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg`.

### DIFF
--- a/etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg
+++ b/etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg
@@ -49,11 +49,25 @@
 #lti.blacklist.user.1=
 
 # Determines whether a JpaUserReference should be created on LTI User Login.
-#
 # This persists the LTI Users in the database, giving them the ability to create long running tasks like ingesting a video.
-#
 # Default: true
-#lti.create_jpa_user_reference=true
+# lti.create_jpa_user_reference = true
+
+# Determines which LTI roles should be persisted in the database on LTI user logins.
+# The "lti.create_jpa_user_reference" config key has to be "true", otherwise this config key will be ignored.
+# The value can be a list of LTI roles identifying users to be persisted or the special value * causing all users to be persisted.
+# The value is not case sensitive.
+#
+# Examples:
+#  - Persist only instructors:
+#    lti.create_jpa_user_reference.roles = instructor
+#  - Persist only instructors and administrators:
+#    lti.create_jpa_user_reference.roles = instructor, administrator
+#  - Persist all users:
+#    lti.create_jpa_user_reference.roles = *
+#    Default value is *, so if nothing is defined, all users will be persisted
+#
+# lti.create_jpa_user_reference.roles = *
 
 # Add Custom Roles to users who has the role with custom_role_name
 # Default: empty no custom roles
--- a/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
+++ b/modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
@@ -51,11 +51,13 @@ import org.springframework.security.oauth.provider.OAuthAuthenticationHandler;
 import org.springframework.security.oauth.provider.token.OAuthAccessProviderToken;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Dictionary;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -123,6 +125,9 @@ public class LtiLaunchAuthenticationHandler implements OAuthAuthenticationHandle
   /** The key to look up whether a JpaUserReferences should be created on login **/
   private static final String CREATE_JPA_USER_REFERENCE_KEY = "lti.create_jpa_user_reference";
 
+  /** The key to look up the LTI roles for which a JpaUserReference should be created **/
+  private static final String LTI_ROLES_TO_CREATE_JPA_USER_REFERENCES_FROM = "lti.create_jpa_user_reference.roles";
+
   /** The security service */
   private SecurityService securityService = null;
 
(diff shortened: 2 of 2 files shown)


marwyg:
Ready for review.

kristofkeppens:
We were having the same performance issues on our installation. We have merged this change into our own branch and can confirm that it works as expected!

lkiesow:
We shortly discussed this on the technical meeting and thought that this could be valuable for `r/9.x` since some people have real problems with too many users. @marwyg, could you rebase this to `r/9.x` but to not break any updates by unexpectedly changing Opencast's behavior then change the default value to persist all users:

```.properties
lti.create_jpa_user_reference.roles = *
```

We can then later change the default back to lecturers only for Opencast 10.

kristofkeppens:
@lkiesow , @marwyg we took this code already in production, since we were facing serious performance issues, but just found out that setting it to "*" does not work. Specifying the types to for instance "administrator, instructor" works fine.

marwyg:
It was rebased to `r/9.x`. 
The default value is set to `*`, which should work now. It means that everything will be persisted. 
Leaving the configuration blank should also work. It will disable the persisting.

Its ready for review again.

modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
```diff
@@ -160,8 +162,8 @@
   /** concurrent attemtps */
   private Map<String, Boolean> activePersistenceTransactions = new ConcurrentHashMap<>(128);
 
-  /** Determines whether a JpaUserReference should be created on lti login */
-  private boolean createJpaUserReference = true;
+  /** Contains the LTI roles where a JpaUserReference shall be created. The default is 'instructor'. */
```
lkiesow:
```suggestion
  /** LTI roles a user must have so a JpaUserReference is created. */
```

etc/org.opencastproject.security.lti.LtiLaunchAuthenticationHandler.cfg
```diff
@@ -48,12 +48,23 @@
...
+#    lti.create_jpa_user_reference = instructor
+#  - Persist only instructors and administrators:
+#    lti.create_jpa_user_reference.roles = instructor, administrator
+#  - Persist all users:
+#    lti.create_jpa_user_reference.roles = *
+#  - No persistence:
+#    lti.create_jpa_user_reference.roles =
+#
+# Default: *
+lti.create_jpa_user_reference.roles = *
```
lkiesow: Setting a default should be unnecessary:

```suggestion
#lti.create_jpa_user_reference.roles = *
```

modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
```diff
@@ -237,9 +239,8 @@ protected void activate(ComponentContext cc) {
       usernameBlacklist.add(username);
     }
 
-    createJpaUserReference = BooleanUtils.toBooleanDefaultIfNull(
-      BooleanUtils.toBooleanObject(StringUtils.trimToNull((String) properties.get(CREATE_JPA_USER_REFERENCE_KEY))),
-      true);
+    ltiRolesForUserCreation = extractLtiRolesForUserCreation(
+            (String) properties.get(LTI_ROLES_TO_CREATE_JPA_USER_REFERENCES_FROM));
```
lkiesow: Actually set the default you claim to set:

```suggestion
    ltiRolesForUserCreation = extractLtiRolesForUserCreation(
            Objects.toString(properties.get(LTI_ROLES_TO_CREATE_JPA_USER_REFERENCES_FROM), "*");
```

lkiesow (changes requested):
Sorry for complaining about this even more :see_no_evil: 

Subsequent 9.x releases should not break previous configurations. E.g. a 9.4 should run with a 9.0 configuration and behave the same whenever possible. With this, you have the three cases:

- `lti.create_jpa_user_reference` was not set
    - The default was `true` and the new default `*` (once it is actually set in code) behaves the same  ✓
- `lti.create_jpa_user_reference` was set to `true`
    - Is the same. The new default would make the behavior stay the same ✓
- `lti.create_jpa_user_reference` was set to `false`
    - Here, the new default would suddenly start persisting users even though you had disabled this. That's a problem.

I guess the best thing to do is to still keep the old key around for now and use something like:
```java
if (createJpaUserReference && ltiRolesForUserCreation != null && ltiRolesForUserCreation.size() > 0) {
```

modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
marwyg: The key "createJpaUserReference" is included again (default true).

modules/security-lti/src/main/java/org/opencastproject/security/lti/LtiLaunchAuthenticationHandler.java
marwyg: The default (*) is set programmatically now.

marwyg:
Now the cases look like this:

`create_jpa_user_reference` = not set (default will be used = true)
`create_jpa_user_reference.roles` = not set (default will be used *)
	**- persist everything**

`create_jpa_user_reference` = false
`create_jpa_user_reference.roles` = anything (doesn't matter)
	**- persist nothing**

`create_jpa_user_reference` = true
`create_jpa_user_reference.roles` = not set (default will be used = *)
	**- persist everything**

`create_jpa_user_reference` = true
`create_jpa_user_reference.roles` = empty string (default will be used = *)
	**- persist everything**

`create_jpa_user_reference` = true
`create_jpa_user_reference.roles` = role_a, role_b
	**- persist only users with role 'role_a' and 'role_b'**

lkiesow (approved):

Issue #1798, opened 2020-08-26
Admin UI does not reflect series role if series is created with only a single role


Arnei:
When creating a new series through the `/series/` API, with an ACL containing a single role, the admin UI does not correctly reflect the role, showing either an empty role, or the given role but without the specified permissions. This seems to be a visual error only, as a correct ACL is still generated and used.

![Bildschirmfoto vom 2020-08-26 11-03-41](https://user-images.githubusercontent.com/14070005/91284169-d5b37600-e78b-11ea-833d-33dd52b8287e.png)

**To Reproduce**
1. Go to https://develop.opencast.org/docs.html?path=/series#updateSeries-5
2. In the example acl, change the `action` tag to e.g. read.
3. Submit the example form
4. Go to https://develop.opencast.org/admin-ng/index.html#!/events/series and check your newly created series acces policy.

**Environment Information**
 - Fedora 31
 - Firefox
 - Opencast 8.5

horstmannmat:
I am having the same issue,  or a little bit more generic, when I try to apply the ROLE_ANONYMOUS, even with more than one role,  to read a series, it does not apply, and I can't see the series at the media-module... 

I don't know if it fits in your issue, but seems pretty similar

lkiesow:
This is an issue related to the old (now removed) admin interface. Closing this.
If this still applies to the new admin interface, please open an issue in the admin interface repository.

PR #2112, opened 2020-11-30
Fix admin interface not displaying the correct role

### LINKED ISSUE #1969: Admin interface access control display broken
If an access control list contains a role not provided in the admin interface's list provider for roles, the interface fails to display the role but displays an empty field instead: 

Screenshot from 2020-10-15 23-19-06

The problem appears on `develop` from 2020-10-15 and likely affects `r/9.x`

mheyen:
If a role is not provided by the admin interface's list provider, it will now be displayed in the role selection.

Previously, there was a bug and the UI displayed [object, Object] in the role selection.
Now it shows the name of the role.

This fixes #1969.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -168,6 +168,9 @@ angular.module('adminNg.controllers')
         },
         changePolicies = function (access, loading) {
           var newPolicies = {};
+          if (!Array.isArray(access)) {
+            access = [access];
+          }
           angular.forEach(access, function (acl) {
             var policy = newPolicies[acl.role];
 
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -431,7 +431,7 @@ angular.module('adminNg.controllers')
           //to resolve before we can add the roles that are present in the series but not in the system
           $scope.access.$promise.then(function () {
             $scope.roles.$promise.then(function() {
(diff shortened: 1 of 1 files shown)


lkiesow (changes requested):
While the issue is a bit harder to reproduce, it also applies to `r/9.x`. It you are ready, it would be great if you could rebase this to `r/9.x`. But there seem to be two issues with this:

First, the interface now seems to be constantly complaining about requiring at least one role with read and write priviledges, even if there is one. Since `ROLE_ADMIN` has always access, we could potentially even drop this restriction:

opencast-acl-warning

Second, ingesting an ACL with just one read role seems to somewhat break the user interface still:

opencast-acl-user-read-only

This was ingested using:

```sh
curl -f -i -u admin:opencast \
  http://localhost:8080/ingest/addMediaPackage/fast \
  -F title=acl-test-user \
  -F creator="J. Doe" \
  -F acl='{"acl": {"ace": [{"allow": true,"role": "ROLE_USER","action": "read"}]}}'
```

lkiesow:
The second error is also reported as #1798

lkiesow:
> This pull request has conflicts frowning_face
> Please resolve those so we can review the pull request.
> Thanks.

Conflicts are against `develop` not against `r/9.x`

mheyen:
The issue with the notification "requiring at least one role with read and write privileges" seems to be fixed as well, at least I cannot replicate it anymore.

lkiesow:
Seems to work. I did see the „requiring at least one role with read and write privileges“ once again but could not reliably reproduce this. In the end this is definitely a huge improvement → merging

mheyen:
I also saw the notification once during processing. But after finishing ingesting, it seems to work correctly. At least it is not _constantly_ complaining anymore.

PR #2128, opened 2020-12-09
Change Workflow Details Button to the blue "more" Icon.

CGreweling:
Replace all "Details" Links with the "more-Icon".
For a more consitent look and feel of the Admin UI.
Anlagen-more-button

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
+++ b/modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
@@ -149,8 +149,8 @@
                   <th translate="EVENTS.EVENTS.DETAILS.ASSETS.TYPE"><!-- Type --></th>
                   <th translate="EVENTS.EVENTS.DETAILS.ASSETS.SIZE"><!-- Size --></th>
                   <th class="medium">
-                    <a translate="EVENTS.EVENTS.NEW.UPLOAD_ASSET.ADD"
-                       class="details-link"
+                    <a
+                       class="more"
                        ng-show="(uploadAssetOptions && !transactions.read_only)"
                        ng-click="openSubTab('newAssetUpload', 'EventAssetAttachmentsResource', 'newassetupload', false, true)"
                        with-role="ROLE_UI_EVENTS_DETAILS_ASSETS_EDIT">
@@ -165,8 +165,8 @@
                   </td>
                   <td>{{ assets.attachments }}</td>
                   <td>
-                    <a translate="EVENTS.EVENTS.DETAILS.ASSETS.DETAILS"
-                       class="details-link"
+                    <a
+                       class="more"
                        ng-show="assets.attachments > 0"
                        ng-click="openSubTab('asset-attachments', 'EventAssetAttachmentsResource', 'attachment')">
                       <!-- Details -->
@@ -179,8 +179,8 @@
                   </td>
                   <td>{{ assets.catalogs }}</td>
                   <td>
-                    <a translate="EVENTS.EVENTS.DETAILS.ASSETS.DETAILS"
-                       class="details-link"
+                    <a
+                       class="more"
                        ng-show="assets.catalogs > 0"
                        ng-click="openSubTab('asset-catalogs', 'EventAssetCatalogsResource', 'catalog')">
                       <!-- Details -->
@@ -191,8 +191,8 @@
                   <td translate="EVENTS.EVENTS.DETAILS.ASSETS.MEDIA.CAPTION"><!-- Media --></td>
                   <td>{{ assets.media }}</td>
                   <td>
-                    <a translate="EVENTS.EVENTS.DETAILS.ASSETS.DETAILS"
-                       class="details-link"
+                    <a
+                       class="more"
                        ng-show="assets.media > 0"
                        ng-click="openSubTab('asset-media', 'EventAssetMediaResource', 'media')">
                       <!-- Details -->
@@ -205,8 +205,8 @@
                   </td>
                   <td>{{ assets.publications }}</td>
                   <td>
-                    <a translate="EVENTS.EVENTS.DETAILS.ASSETS.DETAILS"
-                       class="details-link"
+                    <a
+                       class="more"
                        ng-show="assets.publications > 0"
                        ng-click="openSubTab('asset-publications', 'EventAssetPublicationsResource', 'publication')">
                       <!-- Details -->
@@ -434,8 +434,7 @@
 
                       </td>
                       <td>
-                        <a class="details-link" ng-click="openSubTab('workflow-details', 'EventWorkflowDetailsResource', item.id)"
-                           translate="EVENTS.EVENTS.DETAILS.MEDIA.DETAILS">
+                        <a class="more" ng-click="openSubTab('workflow-details', 'EventWorkflowDetailsResource', item.id)">
                           <!-- Details -->
                         </a>
                       </td>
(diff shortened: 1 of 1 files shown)


lkiesow:
I think it's now less obvious what it does and less consistent since the action icons previously never changed the location of the currently active window/modal

gregorydlogan:
I agree, text is better here than an icon, especially since we're not pressed for space.

ebbertd:
I actually like it. Makes it more consistent.

KatrinIhler:
Some background on motivation: SWITCH has been using it for years, so I wanted to see if this was something for the Community since fewer custom changes are always better. ;)

I like this change because at least for the workflow tab, this means one less column which is always empty:

Before:
image

After:
image

I think the UI looks cleaner this way, and I find it odd that we use icons for details in one spot and text in another. But I can see why people would disagree.

So now what? We could ask for further input on list or something...

gregorydlogan:
I like the look of that change.  If feedback from SWITCH has been positive, hen it probably makes sense to keep this.  Just because I don't like it doesn't mean it shouldn't be in the codebase :)

modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
```diff
@@ -149,8 +149,8 @@ <h2 translate="EVENTS.EVENTS.DETAILS.HEADER" translate-values="{{titleParams}}">
                   <th translate="EVENTS.EVENTS.DETAILS.ASSETS.TYPE"><!-- Type --></th>
                   <th translate="EVENTS.EVENTS.DETAILS.ASSETS.SIZE"><!-- Size --></th>
                   <th class="medium">
-                    <a translate="EVENTS.EVENTS.NEW.UPLOAD_ASSET.ADD"
-                       class="details-link"
```
lkiesow, line 153: This is no details link you are replacing here, is it?

lkiesow (changes requested):
The downside of this patch is that you make things less obvious for new users. „Details“ is self-explainatory. The icon unfortunately is not. So you make the user interface harder to understand for new users with not much gain since you usually have more than enough free space.

> I find it odd that we use icons for details in one spot and text in another

There is a larger difference in the action behind the two: All icons open a new modal dialog while all details links modify the current view.

> So now what? We could ask for further input on list or something...

Since Greg and I can just veto the patch as committers, you have to convince us ;-P
Please think carefully if this really helps new users. I think it will hurt them, but I will let myself be outvoted.


If we decide to go with the icons, make sure that:

- All new “Actions” columns have appropriate headers
- You find suitable icons for the non-details icons
- All icons have proper title tags

modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
CGreweling: > The downside of this patch is that you make things less obvious for new users. „Details“ is self-explainatory. The icon unfortunately is not. So you make the user interface harder to understand for new users with not much gain since you usually have more than enough free space.
> 
> > I find it odd that we use icons for details in one spot and text in another
> 
> There is a larger difference in the action behind the two: All icons open a new modal dialog while all details links modify the current view.
> 
> > So now what? We could ask for further input on list or something...
> 
> Since Greg and I can just veto the patch as committers, you have to convince us ;-P
> Please think carefully if this really helps new users. I think it will hurt them, but I will let myself be outvoted.
> 
> If we decide to go with the icons, make sure that:
> 
>     * All new “Actions” columns have appropriate headers
> 
>     * You find suitable icons for the non-details icons
> 
>     * All icons have proper title tags

Okay lets start a ui discussion then. :)
The main Problem is here, and i think we all agree in that; the "more-icon" is not intuitive and not known from any other software. So the user has to learn what that icon stands for.
To get to the Modals, where "detail-text" is used, he has to use the details icon before. 
So at this point the user must already know what the icon stands for. And the "Details-text" is something new then.
Its always a bad idea to use different symbols for the same meaning in an ui. 

And "Details-text" is in the Opencast Admin ui  the only Link that uses a Link-Text.
That`s my opinion, if you look from a look and feel Interaction-design thing at the oc-admin ui.

As a user i can understand, that one who is used to it, don't like to loose the clear Button; which says what it does.! :)

I personally would prefer the icon-thing. But don't care that much.

As the author of this pr:
I was asked by switch to change the Details-text to that icon only for the "Workflows-Tab". And that makes it more confusing than as it is now, and more confusing than it is in this pr.
So i made the suggestion to change them all.

lkiesow: > To get to the Modals, where "detail-text" is used, he has to use the details icon before.
> So at this point the user must already know what the icon stands for. And the "Details-text" is something new then.
> Its always a bad idea to use different symbols for the same meaning in an ui.

I like to argue that at this point the user has learned that clicking on the icon will open a new modal dialog with additional information about the entity. That's what the icon does. That's not what the details links do.

> And "Details-text" is in the Opencast Admin ui the only Link that uses a Link-Text.

That's not true. A few examples include:

- Publication links
- Add assets (which you also changed to the event details icon)
- All tab and table headers
- Table entries (e.g. presenters, location or series)
- Help menu entries

You also have different icons for event details and series details, but you are now using the event details icon for workflow details, operations, …

Again, I will let myself be outvoted here but if we decide to go with the icons, make sure that:

* All new “Actions” columns have appropriate headers (they are now inconsistent)
* You find suitable icons for the non-details icons (e.g. add assets¹⁾)
* All icons have proper title tags²⁾

---

1) add-asset

---

2) event-details

CGreweling: ``` I like to argue that at this point the user has learned that clicking on the icon will open a new modal dialog with additional information about the entity. That's what the icon does. That's not what the details links do.```
Okay, my understanding for this is a little bit different. When i use the ui i don't have in mind new windows or modals.(maybe this is different if you look on this, with code background ) 
I only think, i klick on something blue on the right(blue text, blue icon or whatever) and i get more information to this line.

My point in case of using this icon for details is, that we are using this icon on most of the main views of the admin ui.
Users, Events, Groups, Access-Policies, Locations. To get "more" informations for this Line of this Objekt.
So in my opinion this means show me the details for this objekt. So why do we need, another symbolic to get details if we are at any other place in the ui. For example more information for a workflow? 
For example "add Asset" this should be something different, because it does something different.

```
That's not true. A few examples include:

    Publication links
This mostly has an additional icon
    Add assets (which you also changed to the event details icon)
    All tab and table headers
    
    Table entries (e.g. presenters, location or series)
    Help menu entries
```
  Maybe we should change that at some of these points.
  
So how do we get a decision here. A Question on list, some kind of vote in the community?

Yes you are right there should be proper titletags.
The other icons should be in an other pr?
But to do this, we should first see, if the pr would be declined or accepted.
And then i should ask Switch if they are willing to pay for these changes.

gregorydlogan: I would prefer whatever yields the least inconsistency.  In my mind that's leaving this PR out.  If SWITCH wants to pay for this plus the rest of the relevant changes we can review them as a whole package then, but until then we shouldn't be doing partial changes like this.

CGreweling: Well sounds good. I will decline this and see, how Switch will proceed in this. 
Thanks @gregorydlogan  @lkiesow for digging into this.

PR #2152, opened 2020-12-15
Fix usage of "organizers" instead of "creators" for series in Admin UI

LukasKalbertodt:
At some point, some of the instances of "creator" were renamed to
"organizer", but apparently a couple instances were forgotten. As a
consequence, it was not possible to sort the series table by
organizers -- a bug which is fixed by this commit.

CC @KatrinIhler as you know stuff about the admin UI and probably touched this code once.

Where I'm still unsure is the two remaining uses of `SeriesIndexSchema.CREATOR`:

https://github.com/opencast/opencast/blob/1a221000150f6474c42beae79494c3ab30c59136/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/series/SeriesQueryBuilder.java#L102-L104

https://github.com/opencast/opencast/blob/1a221000150f6474c42beae79494c3ab30c59136/modules/index-service/src/main/java/org/opencastproject/index/service/impl/index/series/SeriesIndexUtils.java#L109-L111

I *think* both of those can be removed, but I am not sure, so I need another opinion on that. Is there still a "creator" metadata field for series that is used? Because if not: below both of these snippets, there is code handling organizers already, so it seems like the if-branches for the "creators" are never entered.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
@@ -602,7 +602,7 @@ public class SeriesEndpoint implements ManagedService {
             case SeriesIndexSchema.CONTRIBUTORS:
               query.sortByContributors(criterion.getOrder());
               break;
-            case SeriesIndexSchema.CREATOR:
+            case SeriesIndexSchema.ORGANIZERS:
               query.sortByOrganizers(criterion.getOrder());
               break;
             case SeriesIndexSchema.CREATED_DATE_TIME:
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/seriesController.js
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/controllers/seriesController.js
@@ -33,8 +33,8 @@ angular.module('adminNg.controllers')
         label: 'EVENTS.SERIES.TABLE.TITLE',
         sortable: true
       }, {
-        template: 'modules/events/partials/seriesCreatorsCell.html',
-        name:  'creators',
+        template: 'modules/events/partials/seriesOrganizersCell.html',
+        name:  'organizers',
         label: 'EVENTS.SERIES.TABLE.ORGANIZERS',
         sortable: true
       }, {
--- a/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/seriesOrganizersCell.html
+++ b/modules/admin-ui/src/main/webapp/scripts/modules/events/partials/seriesOrganizersCell.html
@@ -1,4 +1,4 @@
-<span ng-repeat="creator in row.creators"
+<span ng-repeat="creator in row.organizers"
       class=metadata-entry>
   {{creator}}
 </span>
--- a/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/delete-series-modal.html
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/partials/modals/delete-series-modal.html
@@ -46,7 +46,7 @@
                   </td>
                   <td>{{ row.title }}</td>
                   <td>
-                    <span class="metadata-entry" ng-repeat="creator in row.creators">{{ creator }}</span>
+                    <span class="metadata-entry" ng-repeat="organizer in row.organizers">{{ organizer }}</span>
                   </td>
                   <td ng-class="row.hasEvents == true ? 'fa fa-check' : ''"></td>
                 </tr>
--- a/modules/admin-ui/src/main/webapp/scripts/shared/resources/seriesResource.js
+++ b/modules/admin-ui/src/main/webapp/scripts/shared/resources/seriesResource.js
@@ -34,7 +34,7 @@ angular.module('adminNg.resources')
           var row = {};
           row.id = r.id;
           row.title = r.title;
-          row.creators = r.organizers;
+          row.organizers = r.organizers;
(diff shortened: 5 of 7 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/SeriesEndpoint.java
```diff
@@ -271,10 +270,10 @@ public Response getSeriesList(@HeaderParam("Accept") String acceptHeader, @Query
             case SeriesIndexSchema.CONTRIBUTORS:
               query.sortByContributors(criterion.getOrder());
               break;
-            case SeriesIndexSchema.CREATOR:
+            case SeriesIndexSchema.ORGANIZERS:
               query.sortByOrganizers(criterion.getOrder());
               break;
-            case EventIndexSchema.CREATED:
+            case SeriesIndexSchema.CREATED_DATE_TIME:
```
KatrinIhler: I'm gonna rain on your parade here and tell you that this probably needs a new API version, because this is not backwards compatible. :( On the other hand, you could also ignore the API for now...

KatrinIhler:
I'd have to put some more thought into this to answer your question, but one thing that comes to mind: What about older series who still have a creator field in their dublin core catalogs?

LukasKalbertodt:
Force pushed, test *should* now be fixed. 

Regarding the external API changes: mhh could point. I guess I could just leave that be and just fix it for the admin UI?

gregorydlogan:
So, I hate to be picky, but I can see someone (*cough* ETH *cough*) wanting *both* fields since the person who created something might not be the organizer...

karendolan:
Does this affect the series Dublincore standard metadata exchange catalog? For example, if I update the Series organizer, will it automatically update the series metadata catalog's DCterm creator value? Is the series metadata catalog still relevant?

LukasKalbertodt:
> So, I hate to be picky, but I can see someone (cough ETH cough) wanting both fields since the person who created something might not be the organizer...

As I said, I'm fine with reverting the changes to the external API. My main focus of this PR was to fix the bug in the admin UI and I just changed the external API "by accident" simply thinking "oh that's probably a bug, let's fix". And the admin-ui API can just change willy nilly, right? It's explicitly non-stable and only internal AFAIK. So my changes to the admin-ui code doesn't really stop anyone from using the "creator" field for series. It just makes it so that the admin-ui sends `sort=organizers` instead of `sort=creator` when clicking on the column labelled "Organizer(s)". The old behavior (`sort=creator`) always resulted in 400 responses from the API.

(And no problem about being picky; that's PR review :D I'm used to that. Compared to others, I feel like the OC community is rather less picky :P)

> Does this affect the series Dublincore standard metadata exchange catalog? For example, if I update the Series organizer, will it automatically update the series metadata catalog's DCterm creator value? Is the series metadata catalog still relevant?

To be honest, I'm not sure as I still don't know enough about these things. But I highly doubt that this PR changes anything about that.

gregorydlogan:
> [...] the admin-ui API can just change willy nilly, right? It's explicitly non-stable and only internal AFAIK. So my changes to the admin-ui code doesn't really stop anyone from using the "creator" field for series. It just makes it so that the admin-ui sends `sort=organizers` instead of `sort=creator` when clicking on the column labelled "Organizer(s)". The old behavior (`sort=creator`) always resulted in 400 responses from the API.

Correct on all counts.  As much as I dislike the current state of the API, we don't want to break it, but the admin API?  Go nuts.

LukasKalbertodt:
Okay, so I just force pushed and now only the admin-ui code is changed.

lkiesow:
Pull request #2177 is a much simpler fix for the sorting issue. It also does not change the metadata to organizer at some places. Since the origin for this field is the DCMI term creator, it's a bit weird to change it to organizer in parts of the code. Of course, it is also a bit weird to translate it as organizer in the first place.

lkiesow:
This should probably go into `develop`

lkiesow:
Greg merged the simple fix and we shortly discussed on the technical meeting that if we want this, this should probably then go into `develop`. The question remains though, do we want this?

gregorydlogan:
I'm concerned this is breaking things for correctness' sake, rather than a defined need.  Is there an adopter where this is a problem?  And is there a migration path for older series/episodes with the previous behaviour in them?

LukasKalbertodt:
> Since the origin for this field is the DCMI term creator, it's a bit weird to change it to organizer in parts of the code. Of course, it is also a bit weird to translate it as organizer in the first place.

I wasn't aware of that origin. In that case, yep it's strange and either we don't translate it as "organizers" or there has to be a discontinuity (a change of labels) in the code in some place anyway. 

> I'm concerned this is breaking things for correctness' sake, rather than a defined need.

To which one might reply "I'm concerned to sacrifice correctness for backwards compatibility's sake". More seriously, I don't yet understand what breakage you are concerned about. This PR only changes on thing about the API: what string to pass as `sort` parameter to get the results sorted by "organizers". This `sort` parameter is only ever used by this one part of the admin-UI in which a fixed string is passed. It's not like we check the series-data first to obtain the metadata field name. So there are no "older series/episodes with previous behavior". And as stated above, the admin-UI API can break. 

*That said*, I'm really not invested in this PR. I created it for a customer, but Lars' #2177 already solves the surface problem. And with what I wrote above in reply to Lars, **I propose to close this** as I don't really see a benefit anymore. Feel free to actually close if you agree.

gregorydlogan:
Author suggests closing thing -> closed. :)

Issue #2146, opened 2020-12-14
Dead Google Transcription jobs can indefinitely throw exception


tyfranck:
Transcription job error due to problems with the Google transcription APIs can result in indefinite exceptions.  
E.g. `Exception when calling the recognition endpoint for media package unknown, job id 6199909346655994050`

karendolan:
Patched in develop branch via commit c87ef69.

PR #2156, opened 2020-12-15
Fix Google transcription service indefinite errors generation #1664 #2146

### LINKED ISSUE #1664: Google speech transcription service keeps generating errors when mediapackage deleted
Currently when a mediapackage with a scheduled transcription job is deleted before the transcription job ends (transcript attached to the mediapackage), the service will indefinitely generate errors as it is not able to find the mediapackage.

Correct behaviour: transcription should stop trying to attach transcript to mediapackage if it doesn't exist and inform the user.

[Issue reported by Carter Wallace. ](https://groups.google.com/a/opencast.org/forum/#!topic/users/hKsO26lLTuE)

tyfranck:
If media package has been deleted or Google speech to text APIs encounters a problem, The Google transcription service will keep generating errors without stopping.
This pull request addresses those issues.

This fixes  #1664  #2146

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
+++ b/modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
@@ -573,12 +573,12 @@ public class GoogleSpeechTranscriptionService extends AbstractJobProducer implem
           JSONParser jsonParser = new JSONParser();
           JSONObject jsonObject = (JSONObject) jsonParser.parse(jsonString);
           Boolean jobDone = (Boolean) jsonObject.get("done");
-          if (jobDone) {
-            resultsArray = getTranscriptionResult(jsonObject);
-          }
           TranscriptionJobControl jc = database.findByJob(jobId);
           if (jc != null) {
             mpId = jc.getMediaPackageId();
+          }
+          if (jobDone) {
+            resultsArray = getTranscriptionResult(jsonObject);
           }
           logger.info("Recognitions job {} has been found, completed status {}", jobId, jobDone.toString());
           EntityUtils.consume(entity);
@@ -604,6 +604,10 @@ public class GoogleSpeechTranscriptionService extends AbstractJobProducer implem
     } catch (TranscriptionServiceException e) {
       throw e;
     } catch (Exception e) {
+      if (hasTranscriptionRequestExpired(jobId)) {
+        // Cancel the job and inform admin
+        cancelTranscription(jobId, "Transcription ERROR", "Transcription job canceled due to errors");
+      }
       String msg = String.format("Exception when calling the recognitions endpoint for media package %s, job id %s",
               mpId, jobId);
       logger.warn(msg, e);
@@ -1007,8 +1011,7 @@ public class GoogleSpeechTranscriptionService extends AbstractJobProducer implem
               try {
                 if (!getAndSaveJobResults(jobId)) {
                   // Job still running, not finished, so check if it should have finished more than N seconds ago
-                  if (j.getDateCreated().getTime() + j.getTrackDuration()
-                          + (completionCheckBuffer + maxProcessingSeconds) * 1000 < System.currentTimeMillis()) {
+                  if (hasTranscriptionRequestExpired(jobId)) {
                     // Processing for too long, mark job as cancelled and don't check anymore
                     database.updateJobControl(jobId, TranscriptionJobControl.Status.Canceled.name());
                     // Delete file stored on Google storage
--- a/modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
+++ b/modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
@@ -114,7 +114,7 @@ public class GoogleSpeechTranscriptionService extends AbstractJobProducer implem
   static final String ACCESS_TOKEN_EXPIRY_NAME = "expires_in";
   private static final int CONNECTION_TIMEOUT = 60000; // ms, 1 minute
   private static final int SOCKET_TIMEOUT = 60000; // ms, 1 minute
-  private static final int ACCESS_TOKEN_MINIMUN_TIME = 60000; // ms , 1 minute
+  private static final int ACCESS_TOKEN_MINIMUM_TIME = 60000; // ms , 1 minute
   // Default workflow to attach transcription results to mediapackage
   public static final String DEFAULT_WF_DEF = "google-speech-attach-transcripts";
   private static final long DEFAULT_COMPLETION_BUFFER = 300; // in seconds, default is 5 minutes
@@ -606,7 +606,7 @@ public class GoogleSpeechTranscriptionService extends AbstractJobProducer implem
     } catch (Exception e) {
       if (hasTranscriptionRequestExpired(jobId)) {
         // Cancel the job and inform admin
-        cancelTranscription(jobId, "Transcription ERROR", "Transcription job canceled due to errors");
+        cancelTranscription(jobId, "Transcription job canceled due to errors");
       }
       String msg = String.format("Exception when calling the recognitions endpoint for media package %s, job id %s",
               mpId, jobId);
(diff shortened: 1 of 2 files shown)


lkiesow:
Would it make sense to bring this into 9.x?
It seems like a bug fix.

tyfranck:
Yes and probably 8.x as well if still supported.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -866,6 +870,37 @@ private String buildResultsFileName(String jobId) {
...
+      // set a time limit based on video duration and maximum processing time
+      if (database.findByJob(jobId).getDateCreated().getTime() + database.findByJob(jobId).getTrackDuration()
+              + (completionCheckBuffer + maxProcessingSeconds) * 1000 < System.currentTimeMillis()) {
+        return true;
+      }
+    } catch (Exception e) {
+      logger.error(String.format("ERROR while calculating transcription request expiration for job: %s", jobId), e);
+    }
+    return false;
+  }
```
karendolan, line 906: Would it make sense to return true when there is an ERROR level error? Or is it useful to allow manual intervention to troubleshot the issue before cancelling the transcript? I'm wonder why it is better to leave the job in perpetual non-expired state during an exception. For example, maybe an issue with the database connection at that point or the transcription job has not been fully at this point.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -1014,7 +1048,7 @@ public void run() {
             // Apply workflow to attach transcripts
```
karendolan, line 1067: "Jobs that get here have state TranscriptionCompleted [... or had an IOException]"

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -866,6 +870,37 @@ private String buildResultsFileName(String jobId) {
...
+        String token = getRefreshAccessToken();
+        deleteStorageFile(mpId, token);
+      } catch (Exception ex) {
+        logger.warn(String.format("could not delete file %s.%s from Google cloud storage", mpId, defaultEncoding), ex);
+      }
+      // Send notification email
+      sendEmail("Transcription ERROR", String.format("%s(media package %s, job id %s).", message, mpId, jobId));
+    } catch (Exception e) {
+      logger.error(String.format("ERROR while deleting transcription job: %s", jobId), e);
+    }
```
karendolan, line 890: It might be helpful to put the sendMail in a finally block to include the general Exception, or, condense into one try-catch block.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -1043,8 +1077,13 @@ private String startWorkflow(String mpId, String wfDefId, Map<String, String> pa
...
-      logger.warn("Media package {} has not been archived yet.", mpId);
+      if (!hasTranscriptionRequestExpired(jobId)) {
+        // Media package not archived but still within completion time? Skip until next time.
+        logger.warn("Media package {} has not been archived yet. Skipped.", mpId);
+      } else {
+        // Close transcription job and email admin
+        cancelTranscription(jobId, "Transcription job canceled, archived media package not found");
+      }
       return null;
     }
```
karendolan, line 1109: It's too bad that the workflow service is not (or cannot be?) queried here to verify that the mpId still exists when the asset manager has no knowledge of the mpId.  For example, when the event is deleted before Google transcript is attached, a 5 hour timeout plus length of video plus buffer has to be waited through before cancelling the transcript job. .... it might be useful to include the countdown in that Skipped log line. For example 
```
logger.warn("Media package {} has not been archived yet or has been deleted. Will keep trying for {} more minutes before cancelling transcript job {}.", mpId, getRemainingTranscriptionExpireTimeInMin(jobId), jobId);
```

karendolan: As a future feature, it might be worth adding a listener for deleted events to the transcript service so it can cancel out jobs when the event is deleted before the transcript is attached.

For example, here the event delete is sent to ActiveMQ but transcription service doesn't listen to that event.
```
2021-01-27T10:18:19,425 | INFO  | (AssetManagerWithMessaging:221) - Send delete message for episode 45b480a8-ace5-47b8-abee-33744cacb5fe to ActiveMQ
2021-01-27T10:31:35,944 | INFO  | (GoogleSpeechTranscriptionService:591) - Recognitions job 4747558756009661662 has been found, completed status true
```

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
```diff
@@ -604,6 +604,10 @@ boolean getAndSaveJobResults(String jobId) throws TranscriptionServiceException,
     } catch (TranscriptionServiceException e) {
       throw e;
     } catch (Exception e) {
+      if (hasTranscriptionRequestExpired(jobId)) {
+        // Cancel the job and inform admin
+        cancelTranscription(jobId, "Transcription job canceled due to errors");
+      }
```
karendolan, line 611: It would help to log an INFO that the job is being cancelled, in addition to the email. Since there are 5 hours of retry logs, this would be the one log bringing the logs to closure for this transcription job. 

Also, IMHO, it would help to add "Google Transcription" to the cancelTranscription email text, since the email message is coming from the Google Transcription service (vs IBM Watson Transcription service)

karendolan:
Hi @tyfranck  the pull works in that it waits for 5 hours + buffer + video duration before cancelling the transcription job. IMHO, 80% of the problem is that the transcription service does not subscribe to delete-event events (or to ask the workflow service if it knows the mpId when asset manager doesn't). But, that's a bigger code change. 

My biggest comment on this pull is to add an INFO log when doing the cancel. Also, is there an Opencast docs section for Google Transcription that can describe the transcript expire delay config? Feel free to comment back a yes or no regarding my comments :).

tyfranck:
Hi @karendolan, thanks for reviewing this. I'll look into your suggestion of transcription service subscribing to delete-event. Do you know anywhere in opencast where that is done so I can re-use? It's obviously bigger change and not part of this PR but I'll investigate for future updates. 
Note that the `5 hours`  and the `buffer` can be changed to any duration in the config.

I'll add an INFO log and update Google transcription docs accordingly for this PR, you're right about that :-).

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: yes, returning `true` will avoid perpetual non-expired state.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: I like the idea of count down, I'll try to add it.
I'll add the second part of your comment in future improvements.

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Agree

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: fixed

modules/transcription-service-google-speech-impl/src/main/java/org/opencastproject/transcription/googlespeech/GoogleSpeechTranscriptionService.java
tyfranck: Count down now added

karendolan (approved):
Looks good. A delete event watcher feature for some future day.

PR #2160, opened 2020-12-16
Fix Ingest by Non-privileged User

lkiesow:
This patch fixes non-scheduled ingests by non-privileged users which can
currently run into the problem that they don't get the necessary access
rights.

What probably happed was:

- Opencast would create a first version in the asset manager not storing the access control list
- Opencast would then try storing workflow parameters in the asset manager but the asset manager would block access since there are no rules allowing that
- Since no metadata were stored, the event would appear to be empty in the admin interface

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/shared-filesystem-utils/src/main/java/org/opencastproject/assetmanager/util/WorkflowPropertiesUtil.java
+++ b/modules/shared-filesystem-utils/src/main/java/org/opencastproject/assetmanager/util/WorkflowPropertiesUtil.java
@@ -21,6 +21,8 @@
 package org.opencastproject.assetmanager.util;
 
 import static org.opencastproject.assetmanager.api.AssetManager.DEFAULT_OWNER;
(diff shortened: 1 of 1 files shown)


gregorydlogan:
To be clear, a non-privileged user is someone with which roles?  ROLE_USER, but not (ROLE_ADMIN or ROLE_COURSE_ADMIN)?

lkiesow:
The issue should appear for anyone who has no `ROLE_ADMIN` or (maybe?) `ROLE_CAPTURE_AGENT`. For my tests, I created a user with all of the `*UI*` roles to be able to upload via admin interface. If you do that on legacy/stable right not, that should result in a 500 error response

lkiesow:
@uzk-rrzk, just confirmed that this resolves the issue in their 8.9 cluster.

gregorydlogan:
This works for me

PR #2161, opened 2020-12-16
Wowza url signing

jchssystems:
This is a feature to sign urls with Wowza.
To configure this, use the file configuration **org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg**

Each property has the form: key._keyId_._propertyName_

**key.wowzatoken.organization** is optional. By default is *, that means any organization.
**key.wowzatoken.url** is mandatory. describes the url prefix that urls must have.
**key.wowzatoken.secret=myTokenPrefix:mySharedSecret** is mandatory. It is a pair separated by **:**. First value, _myTokenPrefix_, means the prefix that all parameters of signed urls will have. Second value, _mySharedSecret_, defines a secret value, to secure urls.

### DIFF
--- a/docs/guides/admin/docs/configuration/stream-security.md
+++ b/docs/guides/admin/docs/configuration/stream-security.md
@@ -220,6 +220,25 @@ Example:
     url.regex.archive=.*archive\/archive\/mediapackage\/.*\/.*\/.*
     url.regex.static=.*static.*
 
+Configuration of Wowza URL Signing
+----------------------------------
+
+To configure Wowza URL Signing, you should use the file configuration `org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg`.
+
+You can find three properties to configure:
+
+    # key.wowzatoken.secret=myTokenPrefix:mySharedSecret
+    # key.wowzatoken.url=http://localhost:8080
+    # key.wowzatoken.organization=mh_default_org
+
+Each property has the form: key.*keyId*.*propertyName*
+
+`key.wowzatoken.organization` is optional. By deafult is `*`, that means any organization.
+
+`key.wowzatoken.url` is mandatory. describes the url prefix that urls must have.
+
+`key.wowzatoken.secret=myTokenPrefix:mySharedSecret` is mandatory. It is a pair separated by `:`. First value, `myTokenPrefix`, means defines the prefix that all parameters of signed urls will have. Second value, `mySharedSecret`, defines a secret value, to secure urls.
+
 Testing
 -------
 
--- a/etc/org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg
+++ b/etc/org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg
@@ -10,7 +10,9 @@
 # A URL signing key has the following attributes:
 
     # Key ID: Key Identifier, e.g. ‘demoKeyOne’
-    # Key secret: Key value, e.g. ‘25DA2BA549CB62EF297977845259A’. The key-length is not predefined, but a key length of
+    # Key secret: Key value pair prefix:secret
+        # prefix is the prefix that all the parameters of the url will have.
+        # secret is a key value, e.g. ‘25DA2BA549CB62EF297977845259A’. The key-length is not predefined, but a key length of
         # at least 128 bit is recommended. Any larger value will not increase security of the underlying algorithm.
     # URL prefix: The URL Signing Provider will only sign URLs that start with this value. This allows to support
         # multiple distributions and different key pairs.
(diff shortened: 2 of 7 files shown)


rfharokh (commented):
Nice of you that you finished my work: 
https://github.com/rfharokh/Opencast/commit/a253359e610a4b5caf4ce9dd2da85cd458125508

But why did you removed my name/refs?

If you need a branch based on the develop branch use this one: https://github.com/rfharokh/Opencast/tree/wowza-url-signing-feature

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/AbstractUrlSigningProvider.java
```diff
@@ -273,7 +295,7 @@ public String sign(Policy policy) throws UrlSigningException {
       return new URI(uri.getScheme(), null, uri.getHost(), uri.getPort(), uri.getPath(),
               URLEncodedUtils.format(queryStringParameters, StandardCharsets.UTF_8), null).toString();
     } catch (Exception e) {
-      getLogger().error("Unable to create signed URL because", e);
+      getLogger().error("Unable to create signed URL because {}", ExceptionUtils.getStackTrace(e));
```
KatrinIhler, line 298: Why this change? The stack trace was logged before.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/provider/impl/WowzaUrlSigningProvider.java
```diff
@@ -46,4 +58,170 @@ public String toString() {
...
+
+    try {
+      URI uri = new URI(policy.getBaseUrl());
+
+      /*
+        For backward compatibility, but i can not see how this could work.
+        According to the documentation "https://www.wowza.com/docs/how-to-protect-streaming-using-securetoken-in-wowza-streaming-engine"
+        if you using token v1 we need a TEA implementation.
+      */
+      if ("rtmp".equals((uri.getScheme()))) {
```
KatrinIhler, line 80: RTMP was removed for Adaptive Streaming anyway a while ago (see #1179). So if this is only used for vod URLs, maybe it can be removed here, too. But I'm not sure.

etc/org.opencastproject.security.urlsigning.provider.impl.WowzaUrlSigningProvider.cfg
```diff
@@ -25,3 +27,7 @@
 
 # key.demoKeyTwo.secret=6EDB5EDDCF994B7432C371D7C274F
```
KatrinIhler, line 28: So does this "old" way no longer work? Then these comments should be removed.

modules/urlsigning-service-impl/src/main/java/org/opencastproject/security/urlsigning/WowzaResourceStrategyImpl.java
```diff
@@ -39,15 +39,29 @@
   private static final String FIND_STREAM_FORMAT = ".{3}[:].*$";
   /** The URI scheme that an RTMP address uses. */
   private static final String RTMP_SCHEME = "rtmp";
+  /** The URI scheme that an http address uses. */
+  private static final String ADAPTIVESTREAMING_HTTP_SCHEME = "http";
```
KatrinIhler, line 43: We removed the "adaptive" in most cases a while ago (see #1179), so just calling it streaming here would be fine.

docs/guides/admin/docs/configuration/stream-security.md
```diff
@@ -220,6 +220,25 @@ Example:
     url.regex.archive=.*archive\/archive\/mediapackage\/.*\/.*\/.*
     url.regex.static=.*static.*
 
+Configuration of Wowza URL Signing
```
KatrinIhler, line 223: So if I understand this correctly, this is not Stream Security, but a different way of signing URLs specifically for Wowza? (Cause I'm pretty sure the first already works.) That's bound to cause some confusion...

KatrinIhler (commented):
Do I assume correctly that this only affects video-on-demand streaming and not livestreaming?

KatrinIhler:
> Nice of you that you finished my work:
> [rfharokh@a253359](https://github.com/rfharokh/Opencast/commit/a253359e610a4b5caf4ce9dd2da85cd458125508)
> 
> But why did you removed my name/refs?

So what's up with that? The code does look pretty similar.

gregorydlogan:
@jchssystems ping, this has been open with questions for quite a while now.

gregorydlogan:
Between the potential copyright issues here, and the lack of progress/response we're going to close this.

PR #2163, opened 2020-12-17
Added upload progressbar to the LTI upload tool

kristofkeppens:
Adding a simple uploadprogress to the upload form when uploading via LTI


modules/lti/package.json
```diff
@@ -9,10 +9,14 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "axios": "^0.21.1",
     "bootstrap": "^4.6.0",
+    "dangerously-set-html-content": "^1.0.8",
+    "engage-ui": "../engage-ui",
+    "http-mock-server": "^0.5.2",
```
lkiesow: This is already listed as devDependency. It's probably not necessary to listr this as runtime dependency as well.

modules/lti/src/components/Upload.tsx
```diff
@@ -108,6 +111,42 @@ class TranslatedUpload extends React.Component<UploadProps, UploadState> {
...
+                    },
+                });
+            }
+        }).catch((_) => {
+            console.log("couldn't retrieve metadata at this time, ignoring...");
+        });
+    }
+
+    setUploadPogress(progress: number) {
+        console.log(progress)
```
lkiesow:
```suggestion
        console.debug(progress)
```

modules/lti/src/components/Upload.tsx
```diff
@@ -108,6 +111,42 @@ class TranslatedUpload extends React.Component<UploadProps, UploadState> {
         }));
     }
 
+    componentWillUnmount() {
```
lkiesow: You removed this block in pull request #2263
This probably shouldn't be in here, should it?

modules/lti/package.json
```diff
@@ -9,10 +9,14 @@
     "@fortawesome/react-fontawesome": "^0.1.14",
     "axios": "^0.21.1",
     "bootstrap": "^4.6.0",
+    "dangerously-set-html-content": "^1.0.8",
+    "engage-ui": "../engage-ui",
```
lkiesow: What do we need to include the `engage-ui` module for?

lkiesow (changes requested):
Initial review without a lot of testing.

modules/lti/package.json
kristofkeppens: These dependencies came from the deeplink PR and are not needed here, removed them.

Arnei:
This might just be me failing to resolve the conflicts properly, but I can't build the lti module with this PR.

```
[INFO] Failed to compile.
[INFO] 
[INFO] ./src/components/Series.tsx
[INFO] Cannot find module: '/home/arnewilken/IdeaProjects/opencast9/modules/lti/node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/esm/objectSpread2'. Make sure this package is installed.
```

kristofkeppens:
@Arnei I fixed the merge conflicts and you should be able to build it now.

Arnei (changes requested):
Works fine in my local OC AllInOne installation and code looks fine too.

Regarding the UI, I would prefer it if the percentage number would not move with the progress bar. It does not add any additional information and can make it stressful to read. Although I'm not a professional UI designer, so if people say it is fine the way it is I'm fine with conceding this request.

It would also be nice if the "Upload successful" message was not in a completely different place from the upload progress bar, but that is arguably not this PRs fault and not a change I am requesting for this PR.

kristofkeppens:
@Arnei we are using the react-bootstrap progressbar ( https://react-bootstrap.github.io/components/progress/ ), without heavy modification of this it's not able to change the way the percentage is shown. We could remove it and only keep it for screen-readers, if this is preferred ?

KatrinIhler:
I talked to Arne and he was fine with the current state if it can't be easily changed. So I'm gonna go ahead and merge this. :)

Issue #2167, opened 2020-12-18
LTI upload mislabels the captions file ( and default workflow as well )


kristofkeppens:
When looking at the paella source code it expects captions as attachment with flavor captions/format+language . However the lti upload now uploads captions with flavor vtt+en/captions . The default partial-ingest workflow also tags */captions instead of captions/* for distribution to engage-download. This way paella never picks up the captions.

PR #2168, opened 2020-12-18
Fixes the lti captions upload together with the default ingest workflow #2167

kristofkeppens:
The LTI upload used the incorrect flavor for the captions, resulting in paella player not able to find the captions. This aims to fix that.

### DIFF
--- a/etc/workflows/partial-ingest.xml
+++ b/etc/workflows/partial-ingest.xml
@@ -98,7 +98,7 @@
       id="tag"
       description="Tagging captions for publication">
       <configurations>
-        <configuration key="source-flavors">*/captions</configuration>
+        <configuration key="source-flavors">captions/*</configuration>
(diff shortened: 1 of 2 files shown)


lkiesow:
@flyapen, does it make sense to bring this into 9.x?

kristofkeppens:
@lkiesow we are running this in 9.x without any issues. Without this the captions upload is unusable so it would make sense to do this!

lkiesow:
Can you then update the pull request to target `r/9.x`?

I could also do that if you prefer, but you would then need to allow that by checking the “Allow edits and access to secrets by maintainers” at the bottom of the side-bar on the right (you unchecked that when creating the pull request).

kristofkeppens:
@lkiesow updated to 9.x

lkiesow (approved):

PR #2211, opened 2021-01-12
Refactor Index Rebuild

KatrinIhler:
This removes classes from the index service module and puts them into two new modules:

- SearchIndex: The AbstractSearchIndex and all directly related classes.
- IndexRebuild: All classes that send or receive messages for the IndexRebuild.

Additionally to classes from the IndexService, these modules picked up some other related classes that were scattered in the rest of the code base for some reason. The actual changes aren't all that much, most of this PR consists of dependency and import adjustments.

Important: This does not improve the IndexRebuild itself in any way, but was a way for me to get started and get a better overview over what does what, and is imo an improvement to before, especially because it separates the AbstractSearchIndex, which is quite reasonable, from the madness that is the IndexService.

I wanted to get this merged ahead of other future changes because conflicts with new changes can arise quickly and they will be a hassle to fix.

Edit: The IndexRebuild module might be refactored further in the future, but as I wasn't quite sure what was necessary yet, I'll leave it as is for now.


karendolan:
Hi @KatrinIhler, is there overarching documentation about the Opencast service indices? For example, how the admin search indices are related to the other service indices?

KatrinIhler:
@karendolan As far as I know there is no overarching documentation. This PR affects the two ElasticSearch Indexes for the Admin UI and the External API. This has nothing to do with the Search Endpoint and its index, or other indexes e.g. for the WorkflowService. 
For more clarity, I'm considering renaming the relevant modules to contain _elasticsearch_ instead of _search_. But I don't know yet when I will get to this.

modules/admin-ui-index/src/main/java/org/opencastproject/adminui/index/AdminUISearchIndex.java
```diff
@@ -22,13 +22,13 @@
...
-import org.opencastproject.index.service.impl.index.series.Series;
-import org.opencastproject.index.service.impl.index.theme.Theme;
-import org.opencastproject.index.service.impl.index.theme.ThemeIndexSchema;
+import org.opencastproject.search.index.AbstractSearchIndex;
+import org.opencastproject.search.index.event.Event;
+import org.opencastproject.search.index.event.EventIndexSchema;
+import org.opencastproject.search.index.group.Group;
+import org.opencastproject.search.index.series.Series;
+import org.opencastproject.search.index.theme.Theme;
+import org.opencastproject.search.index.theme.ThemeIndexSchema;
```
lkiesow: These now use the same namespace as the completely unrelated search service.
That's confusing. Can we use something else?

https://github.com/opencast/opencast/blob/e9f79c2d585e3c5facc24d04681753b948704041/modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchService.java#L23

KatrinIhler: Good point. I'm gonna try to rename this into something else, probably "elasticsearch".

KatrinIhler: I renamed this module and its namespace and also the modules search and search-api, which had the same problem, from `search` to `elasticsearch` (and added an `-impl` in one case). Things shouldn't be quite so confusing now. 

I didn't touch class names for now, otherwise I'd still be here next week. The same goes for modules with `index` in their name, since then I'd have to move the index service, and I don't want to deal with that right now. :D

KatrinIhler:
Checks fell over for some reason, but now everything is fine. :D

lkiesow:
I can take the review but it would be nice if you could talk me through the important parts of this patch in the next couple of days so that I do not have to read through all the 175 modified files.

KatrinIhler:
@lkiesow No problem! I'm gonna ping you so we can talk about a possible time. :D

pascalseeland (approved):
Looks good.

KatrinIhler:
Pascal made me aware that Intellij did some unintentional changes while fixing the import order. I'm gonna revert them, so don't merge just yet. :D

pascalseeland:
I did accept it even with these changes, since they are basically like code cleanup. I'm totally fine with reverting them, but I think should take a look into the entire test suite of the external api why it does need those files all together. As far as I know, it does only use json.

KatrinIhler:
@pascalseeland I reverted them (and similar ones) for now, since these changes have nothing to do with what I'm trying to accomplish here. Maybe we can throw some of them away, that wouldn't surprise me. But that's something for later. :)

pascalseeland:
I also approve the updated version.

KatrinIhler:
Since Pascal reviewed this (thanks for that btw!) and was satisfied, but cannot merge this, I'm gonna go ahead and do that now. :)

PR #2247, opened 2021-01-23
LTI Tools Mock Data and UI Server

lkiesow:
This patch fixes and documents the options for running local copies of
the LTI tools and a server providing mock data for easy testing.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


pascalseeland:
Could you change all files to end either end with .json or remove it from all of them. From my personal preference, I would go with having all files ending in .json, as it makes it easier to know what to expect when you open a file.

pascalseeland (changes requested):
Please change all files to end either end with .json or remove it from all of them. From my personal preference, I would go with having all files ending in .json, as it makes it easier to know what to expect when you open a file. The other option is a direct mapping, but I would prefer rather prefer to know the file type.

lkiesow:
Hi @pascalseeland,
the reason for me changing that is that it's now an exact static representation of the REST endpoints. This allows you for example to easily build a fully functional static test version by just copying the build folder into the mock data. That's something you loose if you just have a bunch of JSON files.

I can change the files if you like me to, I just thought it would be nice and you should be aware that we will loose that.
How do you want me to proceed?

pascalseeland:
Hi @lkiesow,

I get your point. I don't think that anybody would do that, but I'm fine with keeping the functionality. What got me confused what that I was both .json and non-.json. To clear out the structure, I would propose to move the config.json for the server up one level and rename it to mock-server-config.json, so that everything in the mock-server folder is mocking data. And you need to rename the existing-meta-data.json to metadata

And in package.json, please check if the http-mock-server could be moved into a devDependencies.

lkiesow:
All done

pascalseeland (approved):
Can be merged as is.

PR #2253, opened 2021-01-25
Configure POST api/groups access correctly

KatrinIhler:
The rule for POST api/groups in the security configuration doesn't actually match the request. With this fix, ROLE_API_GROUPS_EDIT will actually grant the access it's supposed to.


lkiesow (approved):

mliradelc:
I was working 2 years ago with this endpoint, gladly I will look into :)

mliradelc:
Seems good!!  Merge!!

PR #2260, opened 2021-01-26
Stand-Alone Video Editor

lwillmann:
This is the backend API for a new video editing UI. It is based largely on the existing tools endpoint of the admin-ui and generates compatible smil files. 
It should be possible to access the editing service on the presentation node via LTI.

### Your pull request should…
* [X] have a concise title
* [X] be against the correct branch (features can only go into develop)
* [X] pass automated tests
* [X] have a clean commit history
* [X] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)


modules/editor-service-api/src/main/java/org/opencastproject/editor/api/EditingData.java
```diff
@@ -0,0 +1,171 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.editor.api;
+
+import static com.entwinemedia.fn.data.json.Jsons.arr;
+import static com.entwinemedia.fn.data.json.Jsons.f;
+import static com.entwinemedia.fn.data.json.Jsons.obj;
+import static com.entwinemedia.fn.data.json.Jsons.v;
```
lkiesow: Is using this library in here really necessary?
We wanted to slowly remove this and not extend the usage if possible.

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
```diff
@@ -0,0 +1,197 @@
...
+      return RestUtil.R.serverError();
+    }
+    logger.info("Editor POST-Request received: {}", details);
+
+    JSONParser parser = new JSONParser();
+    EditingData editingInfo;
+    try {
+      editingInfo = EditingData.parse(details);
+    } catch (Exception e) {
+      logger.warn("Unable to parse concat information ({})", details, e);
```
lkiesow: We shouldn't warn admins if users do something wrong (e.g. ask for a non-existing identifier). This will just spam the logs. Let's drop this to debug. You will return an error anyway.
```suggestion
      logger.debug("Unable to parse concat information ({})", details, e);
```

etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -0,0 +1,5 @@
+# Track source Configuration
+
+# Tracks marked with the given tag will be added to the internal publication used by the video editor
+# Default: preview
+#preview.tag=preview
```
lkiesow: Can we implement a fallback for this?
While this is very flexible, this does not necessary work for old content.

Having something like this would be great (trying the next option if the previous one did not return a result):

1. Select tracks tagged with `preview.tag`
2. Select tracks with `fallback.flavor`
3. Select all available (internally published) tracks

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -347,4 +347,10 @@ prop.admin.shortcut.general.main_menu=m
...
 # Default: 0
-#prop.admin.editor.segment.minimum_length=1000
\ No newline at end of file
+#prop.admin.editor.segment.minimum_length=1000
+
+# The url where the video editor ui resides
+#
+# Format: An url
+# Default: #!/events/events/$id/tools/editor
+#prop.admin.editor.url=http://localhost:8080/editor-ui/index.html?mediaPackageId=$id
```
lkiesow:
```suggestion
# The location of the video editor linked in the admin interface
#
# Format: A URL or path
# Common values:
#   The internal editor: #!/events/events/$id/tools/editor
#   The external editting tool: /editor-ui/index.html?mediaPackageId=$id
# Default: #!/events/events/$id/tools/editor
#prop.admin.editor.url=#!/events/events/$id/tools/editor
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+  private void errorExit(final String message, final String mediaPackageId, ErrorStatus status, Exception e)
+          throws EditorServiceException {
+    String errorMessage = MessageFormat.format("{0}. Event ID: {1}", message, mediaPackageId);
+    if (e != null) {
+      logger.warn(errorMessage, e);
+    } else {
+      logger.warn(errorMessage);
+    }
+
+    throw new EditorServiceException(errorMessage, status);
```
lkiesow: - If it is an error, why log this at level `warn`
- Why log this at all if we then throw a new error with this message (which will then be logged)
- Shouldn't we include the previous error in the throw statement to get a peoper stacktrace?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+            .map(p -> tuple(p.getA()[1], p.getA()[2]))
+            .collect(Collectors.toSet());
+
+    final List<TrackData> tracks = Arrays.stream(mp.getElements())
+            .filter(e -> e.getElementType().equals(MediaPackageElement.Type.Track))
+            .map(e -> (Track)e)
+            .filter(e -> elementHasPreviewTag(e) || elementHasPreviewFlavor(e))
+            .map(e -> {
+              final String uri = Arrays.stream(internalPub.getTracks())
+                      .filter(a -> a.getFlavor().getType().equals(e.getFlavor().getType()))
```
lkiesow: Can we use more explanatory names than `e` and `a`?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+            .filter(p -> p.getA()[0].equals("hide"))
+            .filter(p -> p.getB().equals("true"))
+            .map(p -> tuple(p.getA()[1], p.getA()[2]))
+            .collect(Collectors.toSet());
+
+    final List<TrackData> tracks = Arrays.stream(mp.getElements())
+            .filter(e -> e.getElementType().equals(MediaPackageElement.Type.Track))
+            .map(e -> (Track)e)
+            .filter(e -> elementHasPreviewTag(e) || elementHasPreviewFlavor(e))
+            .map(e -> {
```
lkiesow: Very long anonymous functions are easily somewhat confusing.
Maybe make this into a proper function?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+    final Collection<Tuple<String, String>> hiddens = latestWfProperties.entrySet()
+            .stream()
+            .map(p -> tuple(p.getKey().split("_"), p.getValue()))
+            .filter(p -> p.getA().length == 3)
+            .filter(p -> p.getA()[0].equals("hide"))
+            .filter(p -> p.getB().equals("true"))
+            .map(p -> tuple(p.getA()[1], p.getA()[2]))
+            .collect(Collectors.toSet());
+
+    final List<TrackData> tracks = Arrays.stream(mp.getElements())
```
lkiesow: This logic seems weird (or at least I do not understand it):

- You get the media package from the asset manager and scan it for videos with the `preview.tag` (with flavor as fallback)
- Then you get the tracks from the internal publication and try matching the flavors
- You then extract everything else from the internal publication.

My questions here are:
- Why not simply getting the track from the internal publication checking for the correct tags there
- Using this method, you have to archive the preview video in addition to publishing it. Is there any advantage to forcing this?

And there is no documentation about how this works. The current docs suggest that tracks are simply selected by tags.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+
+    URI smilURI;
+    try (InputStream is = IOUtils.toInputStream(smil.toXML(), "UTF-8")) {
+      smilURI = workspace.put(mediaPackage.getIdentifier().toString(), catalogId, EditorService.TARGET_FILE_NAME, is);
+    } catch (SAXException e) {
+      logger.error("Error while serializing the SMIL catalog to XML: {}", e.getMessage());
+      throw new IOException(e);
+    } catch (JAXBException e) {
+      logger.error("Error while serializing the SMIL catalog to XML: {}", e.getMessage());
+      throw new IOException(e);
```
lkiesow: No need to log if you throw the error anyway but maybe include the additional messages in the newly thrown errors?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+    // The properties have the format "hide_flavor_audio" or "hide_flavor_video", where flavor is preconfigured.
+    // We filter all the properties that have this format, and then those which have values "true".
+    final Collection<Tuple<String, String>> hiddens = latestWfProperties.entrySet()
+            .stream()
+            .map(p -> tuple(p.getKey().split("_"), p.getValue()))
+            .filter(p -> p.getA().length == 3)
+            .filter(p -> p.getA()[0].equals("hide"))
+            .filter(p -> p.getB().equals("true"))
+            .map(p -> tuple(p.getA()[1], p.getA()[2]))
+            .collect(Collectors.toSet());
```
lkiesow: Wouldn't using a `Map<String, String>` be easier than using a custom data structure in a set?
I think that's also a bit easier to read.
Something like:
```java
latestWfProperties.entrySet()
            .stream()
            .filter(p -> p.getValue().equals("true"))
            .map(p -> p.getKey().split("_"))
            .filter(p -> p.length == 3)
            .filter(p -> p[0].equals("hide"))
            .collect(Collectors.toMap(p -> p[0], p -> p[2]));
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+
+  @Activate
+  @Modified
+  public void activate(ComponentContext cc) {
+    Dictionary<String, Object> properties = cc.getProperties();
+    expireSeconds =  UrlSigningServiceOsgiUtil.getUpdatedSigningExpiration(properties, this.getClass().getSimpleName());
+    signWithClientIP = UrlSigningServiceOsgiUtil.getUpdatedSignWithClientIP(properties,this.getClass().getSimpleName());
+    // Preview tag
+    previewTag = StringUtils
+            .defaultString((String) properties.get(OPT_PREVIEW_TAG), DEFAULT_PREVIEW_TAG);
```
lkiesow:
```suggestion
    previewTag = Objects.toString(properties.get(OPT_PREVIEW_TAG), DEFAULT_PREVIEW_TAG);
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+  }
+
+  public String getPreviewTag() {
+    return previewTag;
+  }
+
+  @Activate
+  @Modified
+  public void activate(ComponentContext cc) {
+    Dictionary<String, Object> properties = cc.getProperties();
```
lkiesow, line 230: To check: Is properties `null` if the configuration file does not exist?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -0,0 +1,779 @@
...
+    expireSeconds =  UrlSigningServiceOsgiUtil.getUpdatedSigningExpiration(properties, this.getClass().getSimpleName());
+    signWithClientIP = UrlSigningServiceOsgiUtil.getUpdatedSignWithClientIP(properties,this.getClass().getSimpleName());
+    // Preview tag
+    previewTag = StringUtils
+            .defaultString((String) properties.get(OPT_PREVIEW_TAG), DEFAULT_PREVIEW_TAG);
+    logger.debug("Preview subtype configuration set to '{}'", previewTag);
+
+    // Preview subtype
+    previewSubtype = StringUtils.defaultString((String) properties.get(OPT_PREVIEW_SUBTYPE), DEFAULT_PREVIEW_SUBTYPE);
```
lkiesow:
```suggestion
    previewSubtype = Objects.toString(properties.get(OPT_PREVIEW_SUBTYPE), DEFAULT_PREVIEW_SUBTYPE);
```

Same below. `Objects.toString(Object, defaultString)` is actually a bit simpler for these cases.

etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -0,0 +1,5 @@
+# Track source Configuration
```
lkiesow: This is missing most options.

lkiesow (changes requested):

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/EditingData.java
lwillmann: No, it is not necessary ... I switched completly to Gson

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
lwillmann: done

etc/org.opencastproject.editor.EditorServiceImpl.cfg
lwillmann: Implemeted this as suggested

etc/org.opencastproject.organization-mh_default_org.cfg
lwillmann: done

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lwillmann: Ok, I removed the warning log message and included the exception

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lwillmann: The problem by using a `Map<String, String>` is to find a unique key that will not be used twice in the workflow properties. 
The suggested way to use `hide` as a key makes not so much sense in my opinion as the whole map only consists of hidden elements and it leads to an exception if audio and video are configured to be hidden (even if this makes no sense). I can't use `audio` or `video` as key because there might be multiple flavors configured. 
I can't use the flavor as key because it leads to an exception if audio and video are configured to be hidden for the same flavor.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lwillmann: done

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lwillmann: done

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lwillmann: done

etc/org.opencastproject.editor.EditorServiceImpl.cfg
lwillmann: I've described more options

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lkiesow: That makes sense. But if `p[0]` is always `hide` we don't actually need to store it at all, and we could just collect to a set/list containing `p[2]`. And I say all of this, not having had a look at the current code. So ignore it for now, I'll confirm or contradict myself later :grinning:

lkiesow: Okay, my `p[0]` in the collector never made sense. It should have been a `p[1]`. But the current solution seems reasonable if we can have duplicated keys.

lkiesow:
Tested this on a distributed system and seems to work except for a frontend bug. But since it's currently an outdated frontend anyway, that's hopefully fixed by now. Once the frontend gets updated I think this will be ready to go. @pascalseeland, get ready for a review ;-p

lkiesow (approved):

etc/ui-config/mh_default_org/editor/editor-settings.toml
```diff
@@ -0,0 +1,26 @@
+### Ungrouped values, used to make GET-Parameters easier to write
+### Values are assigned here only for debugging purposes
+# Id of the event that the editor should open
+#mediaPackageId = '27cd7156-fda6-4b31-aab5-d56833012caf'
```
gregorydlogan: If this is a debugging key we should probably remove it.

etc/ui-config/mh_default_org/editor/editor-settings.toml
```diff
@@ -0,0 +1,26 @@
+### Ungrouped values, used to make GET-Parameters easier to write
```
gregorydlogan: This is really, *really* nittpicky, but I'd just like to register my annoyance with `toml`, `yaml`, `json` __and__ `css` *all* being used in our UI configuration files.  I get that everyone's written their own piece of the kitchen sink, but at some point I think we're going to have to unify how this all works...

etc/ui-config/mh_default_org/editor/editor-settings.toml
```diff
@@ -0,0 +1,26 @@
...
+# figure out the correct URL by itself.
+#url = 'https://pyca.opencast.org'
+# Opencast username. Used for Http basic authentication.
+#name = "admin"
+# Password belonging to the above username
+# WARNING: When run in Opencast, this file is publicly accessibly
+#password = "opencast"
+
+### Settings of the metadata tab
+[metadata]
```
gregorydlogan, line 14: Were these present previously? If not, why not enable them by default?

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/FlavorData.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.editor.api;
+
+public class FlavorData {
```
gregorydlogan: A similar class to this lives in the common bundle, why not use that?

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/SegmentData.java
```diff
@@ -0,0 +1,70 @@
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    SegmentData that = (SegmentData) o;
+    return Objects.equals(start, that.start) && Objects.equals(end, that.end) && Objects.equals(deleted, that.deleted);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(start, end, deleted);
```
gregorydlogan, line 68: My read of the docs is unclear, but `deleted` here could be null.  It's unclear to me what effect, if any, this would have, but perhaps a simplification of the `deleted` logic by setting it to false unless overriden on line 33 would be better/safer?  This also gets rid of the ternary on line 43.

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/TrackData.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+    return new MediaPackageElementFlavor(flavor.getType(), flavor.getSubtype());
+  }
+
+  public TrackData(final String flavorType, final String flavorSubtype, final TrackSubData audio,
+          final TrackSubData video, String uri, String id) {
+    this.flavor = new FlavorData(flavorType, flavorSubtype);
+    this.audio = audio;
+    this.video = video;
+    this.uri = uri;
```
gregorydlogan, line 53: `uri` is unused?  Or is that just showing up in the json going over the network?

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/WorkflowData.java
```diff
@@ -0,0 +1,39 @@
...
+
+public final class WorkflowData {
+  private final String id;
+  private final String name;
+  private final Integer displayOrder;
+  private final String description;
+
+  public WorkflowData(String id, String name, Integer displayOrder, String description) {
+    this.id = id;
+    this.name = name;
```
gregorydlogan, line 31: Similar to `TrackData`, there are a few unused fields here.

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+  @Path("{mediapackageid}/edit.json")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RestQuery(name = "getEditorData",
+          description = "Returns all the information required to get the editor tool started",
+          returnDescription = "JSON object",
+          pathParameters = { @RestParameter(name = "mediapackageid", description = "The id of the media package",
+                  isRequired = true, type = RestParameter.Type.STRING) }, responses = {
+          @RestResponse(description = "Media package found", responseCode = SC_OK),
+          @RestResponse(description = "Media package not found", responseCode = SC_NOT_FOUND) })
+  public Response getEditorData(@PathParam("mediapackageid") final String mediaPackageId) {
```
gregorydlogan: `mediapackageid` -> `mediapackageId`

You're using camelCase for the rest of the parameters, and `mediapackageId` matches (or should match!) the rest of the system.

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+          returnDescription = "",
+          pathParameters = {
+          @RestParameter(name = "mediapackageid", description = "The id of the media package", isRequired = true,
+                  type = RestParameter.Type.STRING) },
+          responses = {
+          @RestResponse(description = "Editing information saved and processed", responseCode = SC_OK),
+          @RestResponse(description = "Media package not found", responseCode = SC_NOT_FOUND),
+          @RestResponse(description = "The editing information cannot be parsed",
+                  responseCode = HttpServletResponse.SC_BAD_REQUEST) })
+  public Response editVideo(@PathParam("mediapackageid") final String mediaPackageId,
```
gregorydlogan: `mediapackageId`

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+          @Context HttpServletRequest request) {
+    String details;
+    try {
```
gregorydlogan, line 123: I'm seeing three distinct `try-catch` blocks here, why not combine them into a single larger `try-catch`?

modules/editor-service-remote/src/main/java/org/opencastproject/editor/remote/EditorServiceRemoteImpl.java
```diff
@@ -0,0 +1,135 @@
...
+  public String getMetadata(String mediaPackageId) throws EditorServiceException {
+    return doGetForMediaPackage(mediaPackageId, METADATA_SUFFIX);
+  }
+
+  @Override
+  public void setMetadata(String mediaPackageId, String metadata) throws EditorServiceException {
+    doPostForMediaPackage(mediaPackageId, METADATA_SUFFIX, metadata);
+  }
+
+  protected String doGetForMediaPackage(final String mediaPackageId, final String urlSuffix)
```
gregorydlogan, line 74: A lot of this is duplicate of `TrustedHttpClientImpl`.  Is there a reason to avoid using that class?

modules/editor-service-endpoint/src/main/resources/OSGI-INF/editor_service_endpoint.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gregorydlogan, line 1: New code should ideally have inline OSGi annotations rather than external files like this.

gregorydlogan (changes requested):
Mostly minor nitpicks.  I have not had time to look at EditorServiceImpl, that's tomorrow.

etc/ui-config/mh_default_org/editor/editor-settings.toml
Arnei: The metadata and thumbnail tab were present before, but only as placeholders to show how the main menu would look like with more than two menu items. So enabling them by default will just show a placeholder message. Since they are A. supposed to be added in the future and B. supposed to be on/off configurable, I thought it simpler to turn them off through the config instead of commenting out a bunch of code in the frontend

etc/ui-config/mh_default_org/editor/editor-settings.toml
Arnei: Right, makes sense.

modules/editor-service-endpoint/src/main/resources/OSGI-INF/editor_service_endpoint.xml
lwillmann: I absolutely agree, and have tried to do so. Sadly I didn't get it to work with the reference to the AdminUI index.

modules/editor-service-remote/src/main/java/org/opencastproject/editor/remote/EditorServiceRemoteImpl.java
lwillmann: Can you be more specific about the duplicate code? I use the method `getResponse` from `RemoteBase` which use the `TrustetHttpClient`

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/TrackData.java
lwillmann: Yep, it is needed for the json

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/WorkflowData.java
lwillmann: Needed for json

modules/editor-service-endpoint/src/main/java/org/opencastproject/editor/endpoint/EditorRestEndpoint.java
lwillmann: done

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/SegmentData.java
lwillmann: done

modules/editor-service-endpoint/src/main/resources/OSGI-INF/editor_service_endpoint.xml
lkiesow: I remember the admin index being problematic. We would probably need to convert all those OSGi services at once at some point.

Issue #2238, opened 2021-01-19
Reformat code according to checkstyle 


LukasKalbertodt:
Starting with #2216, we have some checkstyle-rules which have exceptions defined for specific files/modules. We want to get rid of those exceptions one by one. So this issue tracks the progress of correctly formatting the code and removing those exceptions in `checkstyle-suppressions.xml`.

I plan on doing a good chunk of this work myself, but obviously, there is a lot to do. Both, fixing the code *and* reviewing all the changes will take quite some time. In case this helps anyone, my workflow for this is roughly like this:

<details>
<summary>My workflow for working on this</summary>

- Temporarily set `<failOnViolation>` and `<failsOnError>` to `false` in `pom.xml`
- Remove one or multiple line from `checkstyle-suppressions.xml` 
  - *Note*: if you remove one line for module `foo`, you should probably remove all lines for that module, including other style rules. If you are working on some files anyway, it's probably best to fix all issues at once.
- Run `mvn checkstyle:check 2>&1 | grep "\[ERROR\]"`
- Use the output of that command to somehow quickly open files in your editor. For example, I use Sublime Text and I would process that output such that I get a file with lines in the format `filename.java:linenumber`. Then, I can do the following:
  - ctrl+x to cut the current line
  - ctrl+p & ctrl+v & enter to navigate to that file and to the specific line number
  - Fix the code
  - ctrl+s & ctrl+w & repeat

</details>

## Modules

If modules get fixed, this list should be updated (I imagine enough people can edit this comment, so that should be fine). The numbers in parenthesis mean count the different violations:

1. Line length
2. Braces
3. Indentation

- [x] `admin-ui-index` (0, 0, 3) #2424
- [x] `admin-ui` (501, 175, 425)
- [x] `adopter-statistics-impl` (0, 5) #2274
- [x] `animate-impl` (0, 0, 7) #2424
- [x] `annotation-impl` (41, 16, 7) #2701
- [x] `asset-manager-api` (46, 0, 0) #2419
- [x] `asset-manager-impl` (113, 16, 70) #2419
- [x] `asset-manager-storage-aws` (14, 8, 10) #2419
- [x] `asset-manager-storage-fs` (0, 3, 10) #2419
- [x] `asset-manager-workflowoperation` (0, 4, 0) #2419
- [x] `authorization-manager` (31, 10) #2270 
- [x] `authorization-xacml` (5, 2, 10) #2666
- [x] `caption-impl` (19, 34, 0)
[...]

PR #2270, opened 2021-02-01
Fix checkstyle violations for `authorization-manager`

LukasKalbertodt:
Part of #2238 

This is my first actual style fixing PR. Please let me know if my choice of style is ok. I am not very familiar with how stuff is usually done in the Java world or in the Opencast world. In particular, you might notice that I tend to put one "item" per line if it increases readability, instead of putting items in the line until it is full. In my experience, this leads to easy to read "list of things" and less noise in git diffs.

But yeah, this is just the beginning of my "style fixing PRs", so please tell me now about things I should format differently. 

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
+++ b/modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
@@ -96,10 +96,19 @@ public abstract class AbstractAclServiceRestEndpoint {
   @GET
   @Path("/acl/{aclId}")
   @Produces(MediaType.APPLICATION_JSON)
-  @RestQuery(name = "getacl", description = "Return the ACL by the given id", returnDescription = "Return the ACL by the given id", pathParameters = { @RestParameter(name = "aclId", isRequired = true, description = "The ACL identifier", type = INTEGER) }, responses = {
+  @RestQuery(
+      name = "getacl",
+      description = "Return the ACL by the given id",
+      returnDescription = "Return the ACL by the given id",
+      pathParameters = {
+          @RestParameter(name = "aclId", isRequired = true, description = "The ACL identifier", type = INTEGER)
+      },
+      responses = {
           @RestResponse(responseCode = SC_OK, description = "The ACL has successfully been returned"),
           @RestResponse(responseCode = SC_NOT_FOUND, description = "The ACL has not been found"),
-          @RestResponse(responseCode = SC_INTERNAL_SERVER_ERROR, description = "Error during returning the ACL") })
+          @RestResponse(responseCode = SC_INTERNAL_SERVER_ERROR, description = "Error during returning the ACL")
+      }
+  )
   public String getAcl(@PathParam("aclId") long aclId) throws NotFoundException {
     final Option<ManagedAcl> managedAcl = aclService().getAcl(aclId);
     if (managedAcl.isNone()) {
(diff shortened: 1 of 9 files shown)


modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -96,10 +96,19 @@
   @GET
   @Path("/acl/{aclId}")
   @Produces(MediaType.APPLICATION_JSON)
-  @RestQuery(name = "getacl", description = "Return the ACL by the given id", returnDescription = "Return the ACL by the given id", pathParameters = { @RestParameter(name = "aclId", isRequired = true, description = "The ACL identifier", type = INTEGER) }, responses = {
+  @RestQuery(
+      name = "getacl",
+      description = "Return the ACL by the given id",
+      returnDescription = "Return the ACL by the given id",
+      pathParameters = {
+          @RestParameter(name = "aclId", isRequired = true, description = "The ACL identifier", type = INTEGER)
```
gregorydlogan, line 104: This formatting doesn't match that in #2273, specifically [here](https://github.com/opencast/opencast/pull/2273/files#diff-f57bf6edb4770a3a203d7f49b0d9b7c9e2f6a9b4ba4e13f1393898721b880d5fR90).  I'm fine with either, with a mild preference for this format.  That being said, my *biggest* preference is that whatever we have is something that an IDE spits out when configured properly.

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -223,21 +306,40 @@ public Response deleteAcl(@PathParam("aclId") long aclId) throws NotFoundExcepti
...
-          @RestParameter(name = "aclId", isRequired = false, description = "The ID of the ACL to apply. If missing the episode ACL will be deleted to fall back to the series ACL", type = INTEGER)}, responses = {
+  @RestQuery(
+      name = "applyAclToEpisode",
+      description = "Immediate application of an ACL to an episode",
+      returnDescription = "Status code",
+      pathParameters = {
+          @RestParameter(name = "episodeId", isRequired = true, description = "The episode ID", type = STRING)
+      },
+      restParameters = {
+          @RestParameter(
```
gregorydlogan, line 317: This formatting doesn't match the changes above :)

modules/authorization-manager/src/test/java/org/opencastproject/authorization/xacml/manager/endpoint/OsgiAclServiceRestEndpointTest.java
```diff
@@ -74,162 +76,183 @@ public void tearDownTest() {
...
+        .expect()
+        .body("ace[0].role", equalTo(SERIES_10_INSTRUCTOR_ROLE))
+        .body("ace[0].action", equalTo(Action.READ.toString()))
+        .body("ace[0].allow", equalTo(true))
+        .statusCode(OK).when().post(host("/acl/reduce"));
   }
 
   @Test
   public void testAclReduceInputEmptyAclExpectsBadRequest() {
     given().formParam("acl", "").formParam("action", "write").formParam("role", NEW_ROLE).expect()
```
gregorydlogan, line 195: This isn't formatted as above.  It would be nice if it were, but I'm guessing there's no obvious way to catch this with the autoformatter.

gregorydlogan (changes requested):

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
LukasKalbertodt: The reason why the `@RestParameter` and `@RestResponse` are in a single line here as opposed to the snippet you linked is simply that here, they fit into one line.

> my biggest preference is that whatever we have is something that an IDE spits out when configured properly.

For that, I still have to experiment quite a bit with the IDE style stuff. While the IDEs shouldn't produce anything that violates our current checkstyle rules, they currently don't produce what I hand formatted here. I will have a look at this again.

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
LukasKalbertodt: Again, here, unlike above, the `@RestParameter` doesn't fit into one line, therefore it's broken into multiple ones.

modules/authorization-manager/src/test/java/org/opencastproject/authorization/xacml/manager/endpoint/OsgiAclServiceRestEndpointTest.java
LukasKalbertodt: Yeah true, difficult to catch this. I basically just look at the reported violations and also fix anything nearby that looks terrible off. At least the stuff that I notice. My PRs don't aim to be perfect, but as long as everyone agrees that they clearly improve the state of the code base, they are probably worth it.

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
LukasKalbertodt: I tried configuring IntelliJ in its settings UI and I couldn't make it produce this formatting. There aren't a lot of settings regarding annotations, unfortunately.

PR #2273, opened 2021-02-01
Fix checkstyle violations for 9 modules

LukasKalbertodt:
Part of #2238 

- `db`
- `dictionary-hunspell`
- `dictionary-regexp`
- `distribution-service-aws-s3-remote`
- `distribution-service-streaming-remote`
- `scheduler-api`
- `smil-api`
- `textanalyzer-impl`
- `textextractor-tesseract`

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/textanalyzer-impl/src/main/java/org/opencastproject/textanalyzer/impl/endpoint/TextAnalysisRestEndpoint.java
+++ b/modules/textanalyzer-impl/src/main/java/org/opencastproject/textanalyzer/impl/endpoint/TextAnalysisRestEndpoint.java
@@ -54,8 +54,8 @@ import javax.ws.rs.core.Response.Status;
  */
 @Path("")
 @RestService(name = "textanalysis", title = "Text Analysis Service",
-  abstractText = "This service enables conversion from one caption format to another.",
-  notes = {
+    abstractText = "This service enables conversion from one caption format to another.",
+    notes = {
         "All paths above are relative to the REST endpoint base (something like http://your.server/files)",
         "If the service is down or not working it will return a status 503, this means the the underlying service is "
         + "not working and is either restarting or has failed",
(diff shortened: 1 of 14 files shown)


modules/textanalyzer-impl/src/main/java/org/opencastproject/textanalyzer/impl/endpoint/TextAnalysisRestEndpoint.java
```diff
@@ -86,17 +86,41 @@ public void activate(ComponentContext cc) {
   @POST
   @Produces(MediaType.TEXT_XML)
   @Path("")
-  @RestQuery(name = "analyze", description = "Submit a track for analysis.", restParameters = { @RestParameter(description = "The image to analyze for text.", isRequired = true, name = "image", type = RestParameter.Type.FILE) }, responses = {
-          @RestResponse(description = "OK, The receipt to use when polling for the resulting mpeg7 catalog.", responseCode = HttpServletResponse.SC_OK),
-          @RestResponse(description = "The argument cannot be parsed into a media package element.", responseCode = HttpServletResponse.SC_BAD_REQUEST),
-          @RestResponse(description = "The service is unavailable at the moment.", responseCode = HttpServletResponse.SC_SERVICE_UNAVAILABLE) }, returnDescription = "The receipt to use when polling for the resulting mpeg7 catalog.")
+  @RestQuery(
```
gregorydlogan, line 89: I'm not a huge fan of this change, mainly because it adds *so much* seemingly boilerplate stuff at the front.  I don't have a better idea though, and this will be much easier to review changes against in the future.

gregorydlogan (approved):

modules/textanalyzer-impl/src/main/java/org/opencastproject/textanalyzer/impl/endpoint/TextAnalysisRestEndpoint.java
LukasKalbertodt: I understand what you mean and I don't mind reformatting it into a more compact form if people prefer it that way. 

That said, yes, it will be lots easier to review changes (clean git diffs). Also: this only adds newlines, so the amount of information doesn't change. It's just that before, it was more difficult to quickly pick out individual bits of that information. So in my opinion (!), this change has only advantages except that it takes up more lines. But I'm not sure that is a real disadvantage. Maybe one just has to get used to it.

But yeah, I will do what the community think is best.

PR #2274, opened 2021-02-01
Fix checkstyle violations for 7 modules

LukasKalbertodt:
Part of #2238 

- `adopter-statistics-impl`
- `conductor`
- `cover-image-impl`
- `cover-image-workflowoperation`
- `crop-api`
- `crop-ffmpeg`
- `crop-workflowoperation`

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
+++ b/modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
@@ -235,9 +235,14 @@ public class CropServiceImpl extends AbstractJobProducer implements CropService,
   }
 
   private File cropFFmpeg(File mediaFile) throws CropException {
-    String[] command = new String[] { binary, "-i", mediaFile.getAbsolutePath(), "-vf", "cropdetect="
-            + greyScaleLimit + ":" + round + ":" + reset,
-            additionalParameters, "-f", "null", "-"};
+    String[] command = new String[] {
+        binary,
+        "-i", mediaFile.getAbsolutePath(),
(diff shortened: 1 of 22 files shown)


gregorydlogan:
Should this be against `r/9.x`?

LukasKalbertodt:
Woopsie daisy. Fixed. Thanks!

modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
```diff
@@ -280,7 +285,7 @@ private File cropFFmpeg(File mediaFile) throws CropException {
 
     if (cropValue > (widthVideo / maxCroppingRatio)) {
       logger.info("Black area in the video is considered too large for cropping. File: " + mediaFile.getAbsolutePath()
```
gregorydlogan, line 287: Is something like this a thing we can detect with checkstyles?  It would be nice to require the use of the object parameters rather than allowing stringbuilding.

gregorydlogan (approved):

modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
LukasKalbertodt: Good question. A quick google search didn't show any built-in checkstyle rules for that, unfortunately. :/

gregorydlogan: I didn't figure, but it was worth a shot :)

PR #2289, opened 2021-02-03
Start Index Rebuild directly

KatrinIhler:
Currently we're sending messages to start a (complete or partial) Index Rebuild and the services send messages back to log the progress and report any errors.

With this PR, the Index Rebuild is now started by directly calling repopulate() on the respective service. All services that implement IndexProducer (= have repopulate()) make themselves available to the IndexRebuildService via OSGI so this can work. 

Since we know how many of those services we need because of the Service enum, the IndexRebuildService implements BundleActivator and uses a ServiceListener so it will only register itself once all IndexProducers have been obtained. This way, there's no possibility of missing services when we start the index rebuild, and we don't have to hardcode the number either.

To keep the log messages for progress & errors of the services somewhat consistent even without the messages, I used the otherwise now redundant AbstractIndexProducer to implement some methods for that. No need to re-invent the wheel 8 times when showing percentages. It's not perfect, but it gets the job done.

Since we no longer need the Index messages, quite a lot of related code (e.g. IndexRecreateObject) could be removed.
I also had to move the IndexProducer for the AssetManager to make this work.

Now we can add a new IndexProducer by simply implementing the interface and adding it to the Service enum (which is now order-sensitive = determines the order the index is re-built).

~~**Attention:** This PR includes #2211 and should not be merged before!~~


lkiesow:
> Since we know how many of those services we need because of the Service enum, the IndexRebuildService implements BundleActivator and uses a ServiceListener so it will only register itself once all IndexProducers have been obtained. This way, there's no possibility of missing services when we start the index rebuild, and we don't have to hardcode the number either.

That won't work reliably since OSGi allows to have the same service multiple times, and you could have more service instances than service types. If you know the types anyway, a safe method would be to have a static binding to each type or to do dynamic bindings and then to manually check for each type.

Btb, I did not read the actual code ;-P

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
```diff
@@ -0,0 +1,228 @@
...
+   *
+   * @param index
+   *           The index to rebuild.
+   * @param service
+   *          The {@link IndexRebuildService.Service} to re-add data from.
+   *
+   * @throws IndexRebuildException
+   *           Thrown if the index rebuild failed.
+   */
+  private void recreateService(AbstractSearchIndex index, IndexRebuildService.Service service)
```
lkiesow: This does not recreate a service, does it?
The names here are confusing.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
```diff
@@ -0,0 +1,228 @@
...
+          throws IndexRebuildException {
+
+    if (!indexProducers.containsKey(service)) {
+      throw new IllegalStateException(format("Service %s is not available", service));
+    }
+
+    IndexProducer indexProducer = indexProducers.get(service);
+    logger.info("Starting to recreate index {} for service '{}'", index.getIndexName(), service);
```
lkiesow: Same here. This will not recreate the index, will it? The index is already there and existing entries would not be modified, would they?

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * The services whose data is indexed by ElasticSearch.
+   * Attention: The order is relevant for the index rebuild and should not be changed!
+   */
+  public enum Service {
+    Groups, Acl, Themes, Series, Scheduler, Workflow, AssetManager, Comments
+  }
+
+  private static final Logger logger = LoggerFactory.getLogger(IndexRebuildService.class);
+  private final Map<IndexRebuildService.Service, IndexProducer> indexProducers = new HashMap<>();
```
lkiesow: Not thread-safe. Use a ConcurrentHashMap

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
```diff
@@ -0,0 +1,228 @@
...
+     *
+     * @param bundleContext
+     *           The bundle context.
+     */
+    private IndexProducerListener(BundleContext bundleContext) {
+      this.bundleContext = bundleContext;
+    }
+
+    @Override
+    public void serviceChanged(ServiceEvent serviceEvent) {
```
lkiesow, line 258: If I interpret this correctly, you are manually searching for OSGi services of a given type. Why not simply use a dynamic binding?

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/AbstractIndexProducer.java
```diff
@@ -0,0 +1,215 @@
...
+   * @param org
+   *           The organization (can be null).
+   */
+  protected void logIndexRebuildProgress(String indexName, int total, int current, int batchSize, Organization org) {
+    final int responseInterval = (total < 100) ? 1 : (total / 100);
+    if (responseInterval == 1 || batchSize > responseInterval || current == total
+            || current % responseInterval < batchSize) {
+
+      if (org == null) {
+        logger.info("Updating index '{}' for service '{}': {}/{} finished, {}% complete.", indexName, getService(),
```
lkiesow, line 145: Would it make sense to pass the logger along as well to not loose context information?
(A question for most of the methods in there)

lkiesow (dismissed):
> Since we no longer need the Index messages,

Before anyone else gets confused by this sentence: This pull request just removes a fraction of the messages sent during an index-rebuilt. Which is still a nice step forward!

Before you had:

```
index(repopulate) --- **message** ---> service(loop over data) --+
  ^                                            ^                 |
  |                                            |                 |
  |                                            |                 |
  +--------------------------------- message --+- prep data -----+
```

Now you have

```
index --- **func_call** ---> service((repopulate | loop over data) --+
  ^                                                ^                 |
  |                                                |                 |
  |                                                |                 |
  +------------------------------------- message --+- prep data -----+
```

---

A few things I found while quickly skimming though the code.

KatrinIhler:
> That won't work reliably since OSGi allows to have the same service multiple times, and you could have more service instances than service types. If you know the types anyway, a safe method would be to have a static binding to each type or to do dynamic bindings and then to manually check for each type.
> 
> Btb, I did not read the actual code ;-P

We actually do not simply count the services, but check if each type is available (since we have one entry per type in the map, we can simply count the map entries.) So I think that would work fine even with duplicate services.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: I know! In my defense, I didn't invent those, they were already like this (just in another place). I was too lazy to rename them, but maybe I should...

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: See above. Well, in this instance, the index might have been cleared before, but we don't really know that. I rewrote so many of those senseless log messages, but I still missed some...

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: Good point, will do.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: Since I couldn't figure out how to only activate the IndexRebuildService if we have one service of each type obtained. (We can't just skip a service during the rebuild, since the order is important...) Maybe I missed something? I know I can set MinimumCardinality, but that's a hardcoded number...

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/AbstractIndexProducer.java
KatrinIhler: I did not think of that. Good idea. :D

KatrinIhler:
> ```
> index --- **func_call** ---> service((repopulate | loop over data) --+
>   ^                                                ^                 |
>   |                                                |                 |
>   |                                                |                 |
>   +------------------------------------- message --+- prep data -----+
> ```

I think this is also confusing, but maybe I just don't understand it correctly? :D In my opinion, there are actually two separate things going on here:

- The IndexRebuildService sends a message to the service to start repopulating. The service sends messages back to inform about the progress and any errors. - This loop is what this PR removes and replaces with a direct call to repopulate() and a couple of logging methods in the parent class of each service.

- In repopulate(), the service sends messages for each element (yikes) to trigger the actual index update in some MessageReceiver. This is internal to each service, the IndexRebuildService knows nothing about this. This still happens with this PR. The plan is to change this for each service, maybe in multiple PRs.

These interactions use entirely different queues and message objects and are completely independent of one another.

Maybe this makes it clearer?

KatrinIhler:
@lkiesow I incorporated your findings :)

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
lkiesow: Couldn't you just have something like this in `activate()`:

```pascal
while services.length() < X; do
  sleep(1);
end
```

Or just refuse to start if you don't have all the services.
It's similar to what you do here just with less confusing custom code.

KatrinIhler: Okay, I could've done that, but I think then I prefer my solution. I'm not happy about the additional code either, but the basic principle isn't that hard to understand, and I think it's a much cleaner solution to use what OSGI has to offer us than busy waiting or failing later.

lkiesow:
> We actually do not simply count the services, but check if each type is available (since we have one entry per type in the map, we can simply count the map entries.) So I think that would work fine even with duplicate services.

That doesn't actually work when one service is removed, but it's more a theoretical problem since we usually don't do that unless we shut down Opencast, and it is unlikely that will cause problems.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
lkiesow: I agree that the basic principle isn't that hard but as you said, you had to read through a lot of OSGi documentation to actually get what all components and methods actually do. This means that understanding this and figuring out any problems will probably not be easy for people while not adding much benefit.

So, my thoughts are basically that I would probably be annoyed if I had to deal with this code in a year or two since this is doing “something special” and not doing what basically everybody else does. Not sure if you would like to deal with this code in two years.

---

Maybe another option would also be to use static bindings:

Right now (with this patch applied), to add a new index you would

- Implement the Interface `IndexProducer`
- Add a new type to `IndexRebuildService.Service` in this class
- Make `getService()` return this type

So, you have to implement the service and modify thgis class anyway.
Instead, we could just use a static binding which would then really simplify the OSGi magic.
That would mean that implementing a new service would be:

- Implement the Interface `IndexProducer`
- Add binding for the specific new implementation to `IndexRebuildService`

---

Or yet another idea, just use a dynamic binding and return an error when starting the rebuild if not all service types are available.

---

In the end, this is no veto against getting this in. It should work like this and is a clever approach. I just wouldn't want to deal with this later. But if others do, I'm fine with it ;-p

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+  }
+
+  /**
+   * Register this service at OSGI.
+   *
+   * @param bundleContext
+   *           The bundle context.
+   */
+  private void registerIndexRebuildService(BundleContext bundleContext) {
```
lkiesow: I just realized that the registration isn't actually thread-safe. I think it would be if you just mark this method as synchronized.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: Lars and I talked about this some more and realized that the alternatives aren't really that much better. I don't like just returning an error if services aren't available, but having a static binding for each service isn't that trivial either since we can't reference the sub classes without introducing cyclic dependencies (since all these sub classes implement IndexProducer from the same module), so we would need to differentiate them using target properties, and those are essentially hard-coded strings... so we came to the agreement that I would add a comment explaining how the listening stuff works to make this a bit more maintainable.

KatrinIhler:
> That doesn't actually work when one service is removed, but it's more a theoretical problem since we usually don't do that unless we shut down Opencast, and it is unlikely that will cause problems.

True, i didn't think of that. You could solve this by using lists in the map, but I think this is too much complexity for a use-case that will probably never happen. If somebody runs into this, we can still change it.

modules/index-rebuild/src/main/java/org/opencastproject/index/rebuild/IndexRebuildService.java
KatrinIhler: Dammit, multi-threading was never a strong area of mine... Fixed.

pascalseeland (approved):
The PR looks good to me. I also think that nobody should start a rebuild and then shutdown the required service. I also think the added documentation in IndexRebuildService is sufficient to understand the code.

KatrinIhler:
Since Pascal approved this, but isn't a committer, I'm gonna merge this for him.

PR #2293, opened 2021-02-04
Added the access_policy field to elasticsearch and made it searchable

### LINKED ISSUE #2104: No longer possible to filter events table by user name
Back in 7.x it was possible to filter the events table in the Admin UI by username, this way you could get every event the user had read and write rights on. Since 8.x this no longer seems to work. Was this an intentional change (if yes - why?) or an unintended side effect?

marwyg:
closes #2104 

The events table in the admin ui should be searchable by the access_policy again, so someone can find events by the read and write rights.

### DIFF
--- a/modules/elasticsearch-impl/src/main/resources/elasticsearch/series-mapping.json
+++ b/modules/elasticsearch-impl/src/main/resources/elasticsearch/series-mapping.json
@@ -23,7 +23,7 @@
 
           "license": { "type" : "keyword" },
 
-          "access_policy": { "type" : "text", "index" : false, "store" : false },
+          "access_policy": { "type" : "text" },
 
           "managed_acl": { "type" : "keyword" },
 
--- a/modules/elasticsearch-impl/src/main/resources/elasticsearch/event-mapping.json
+++ b/modules/elasticsearch-impl/src/main/resources/elasticsearch/event-mapping.json
@@ -55,7 +55,7 @@
 
         "rights": { "type" : "keyword" },
 
-        "access_policy": { "type" : "text", "index" : false, "store" : false },
+        "access_policy": { "type" : "text" },
 
         "managed_acl": { "type" : "keyword" },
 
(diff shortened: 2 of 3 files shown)


modules/elasticsearch-impl/src/main/resources/elasticsearch/series-mapping.json
```diff
@@ -23,7 +23,7 @@
 
           "license": { "type" : "keyword" },
 
-          "access_policy": { "type" : "text", "index" : false, "store" : false },
+        "access_policy": { "type" : "text" },
```
LukasKalbertodt: Accidentally incorrect indentation change.

LukasKalbertodt (changes requested):
I tested this locally and it seems to work. That is: entering the name of a user or a group into the full-text search field shows events that contain that user/group in its ACL. Note however, that this is not equivalent to "all events the user has access to". If an event's ACL contains a group and a user gets access to the event through that group, searching for the user in the full-text search won't show the event. 

Also: Issue #2104 (which this PR fixes) was caused by https://github.com/opencast/opencast/pull/1090.  That PR and the linked issue seems like the only intended change was the `type: "keyword"` to `type: "text"`. So I wonder if the `index:false, store: false` addition was by mistake or whether there was a point in doing so? Maybe @wsmirnow can help us out here?

modules/elasticsearch-impl/src/main/resources/elasticsearch/series-mapping.json
marwyg: Removed the incorrect indentation. 

I talked with @wsmirnow about this. I think there were no particular reason for the change(?) I can't exactly remember what he said, only that he gave me the OK to remove the options, so the field can be searchable again.

marwyg: So if it works and @wsmirnow don't have any complaints, is it ready to be merged?

wsmirnow: I can't remember the exact reason for PR #1090. The [ES documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html) matches the use case here, imo. Using type `text` and `index=true` (default preset) make sense for me.

LukasKalbertodt: Ok I guess then it all makes sense and this can be merged.

LukasKalbertodt (approved):
After the discussion above we concluded this change makes sense!

KatrinIhler:
Merged since Lukas approves. :D

PR #2310, opened 2021-02-08
Gracefully handle missing Shibboleth User References

lkiesow:
Shibboleth will usually check on login if a user is present in the
system, importing the user data as user reference if not or continuing
otherwise.

This behavior fails if a user exists but no user reference, causing the
login process to fail.

This might, for example, be caused if another userprovider is active to
supply the roles which a user has while Shibboleth is used for login.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/security-aai/pom.xml
+++ b/modules/security-aai/pom.xml
@@ -56,6 +56,10 @@
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.compendium</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


gregorydlogan (approved):
I can't actually test this, but it looks right to me.  Do we have a Shib adopter who can test who is *not* ELAN? :)

lkiesow:
I did that for @ypatios. They do have this on their system already.

PR #2311, opened 2021-02-10
Update Comments in ElasticSearch Indices directly

KatrinIhler:
With this PR, the EventCommentsService no longer sends messages to update the ElasticSearch indices, but does this directly via method call. This affects the usual creating, updating and deleting of comments as well as the index rebuild.

Other servives will hopefully follow soon in a similar manner.

~~**Attention**: This PR is based on #2289 and should not be merged before that one.~~


pascalseeland (changes requested):
Generally looking at this PR I think it is fine. The only thing I would propose is splitting up the parts exposing the AbstractSearchIndex and the actual comment updating in two PRs. This way, the followup-PRs also only change the small parts containing their service.

KatrinIhler:
@pascalseeland Thank you for taking a look at this PR! I can see your point. However, I just stumbled over a problem with my approach, so I might have to restructure my PRs anyway. I will ping you once the rest is ready for review again. :)

KatrinIhler:
Rebased on develop.

TODO: Add locking mechanism.

KatrinIhler:
Rebased onto #2595. Ready for review again!

pascalseeland (approved):

KatrinIhler:

PR #2318, opened 2021-02-12
Serverless HLS leaves files open

### LINKED ISSUE #2317: Serverless HLS leaves files open
**Describe the bug**
We had a “too many open files” error on engage and we tracked it down to the serverless HLS feature, which leaves the .m3u8 files open when using the AWS S3 distribution.
The culprit method is also used by the composer service so it could also happen on workers.

**To Reproduce**
1. Encode lots of recordings with serverless HLS
2. Run lots of aws s3 distribution jobs for days
3. At some point, Opencast engage will not be able to function anymore due to "too many open files"
4. Run "sudo lsof" and see all temporary files created by the aws s3 distribution  left open in the system

**Environment Information**
 - OS: affect all
 - Opencast version with minor release: 9.x, develop

rute-santos:
Ensure all files are properly closed when distributing serverless hls to aws s3.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/AdaptivePlaylist.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/AdaptivePlaylist.java
@@ -21,6 +21,7 @@
 
 package org.opencastproject.mediapackage;
 
+
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.data.Function2;
 
@@ -200,7 +201,8 @@ public interface AdaptivePlaylist extends Track {
    */
   static Set<String> getReferencedFiles(File file, boolean segmentsOnly) throws IOException {
     Set<String> allFiles = new HashSet<String>(); // don't include playlist variants
-    Set<String> segments = getVariants(file).stream().filter(isPlaylistPred.negate()).collect(Collectors.toSet());
+    Set<String> segments = getVariants(file).stream().filter(isPlaylistPred.negate())
+            .collect(Collectors.toSet());
     Set<String> variants = getVariants(file).stream().filter(isPlaylistPred).collect(Collectors.toSet());
 
     if (!segmentsOnly)
@@ -289,6 +291,7 @@ public interface AdaptivePlaylist extends Track {
     }
     return new ArrayList<File>(map.values());
   }
+
 
   /**
    * Fix all the HLS file references in a manifest when a referenced file is renamed
@@ -734,6 +737,7 @@ public interface AdaptivePlaylist extends Track {
       List<Track> newTracks = new ArrayList<Track>();
       Rep rep = segments.get(0); // use segment dir as temp space
 
+
       // Lambda to rewrite a track using the passed in functions, using closure
       Function<Rep, Boolean> rewriteTrack = (trackRep) -> {
         File srcFile = trackRep.origMpfile;
@@ -766,6 +770,7 @@ public interface AdaptivePlaylist extends Track {
         newTrack.setLogicalName(fileMap.get(trackRep.name)); // set logical name for publication
         return true;
       };
+
 
       try {
         // Rewrite the variants and masters tracks in order and throw exception if there are any failures
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/AdaptivePlaylist.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/AdaptivePlaylist.java
@@ -21,7 +21,6 @@
 
 package org.opencastproject.mediapackage;
 
-
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.data.Function2;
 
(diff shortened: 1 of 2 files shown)


modules/common/src/main/java/org/opencastproject/mediapackage/AdaptivePlaylist.java
```diff
@@ -159,10 +162,12 @@ static boolean checkForMaster(File file) throws IOException {
...
-    Optional<Matcher> m = Files.lines(file.toPath()).map(variantPatt::matcher).filter(Matcher::find).findFirst();
-    if (m.isPresent()) {
-      m = null;
-      return true;
+    try (Stream<String> lines = Files.lines(file.toPath())) {
+      Optional<Matcher> m = lines.map(variantPatt::matcher).filter(Matcher::find).findFirst();
+      if (m.isPresent()) {
+        m = null;
+        return true;
+      }
```
lkiesow: Why set `m = null`? It's a local variable anyway and would be discarded due to its scope, wouldn't it?
This also would allow for the much simpler:

```suggestion
    try (Stream<String> lines = Files.lines(file.toPath())) {
      return lines
          .map(masterPatt::matcher)
          .filter(Matcher::find)
          .findFirst()
          .isPresent();
    }
```

I also changed the formatting since I think that's easier to read like this, but that might just be my preference.

lkiesow: Thinking about this for a moment, isn't this simply:

```java
return lines
    .map(masterPatt::matcher)
    .anyMatch(Matcher::find);
```

rute-santos: I suppose so. This code was there before and I did not change it, just fixed the bug. I can make the requested change.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -921,19 +921,12 @@ protected String process(Job job) throws Exception {
...
-         * } else {
-         * restoredElement = restoreElement(channelId, mediaPackage, elementIds, null);
-         * }
-         * return (restoredElement != null) ? MediaPackageElementParser.getAsXml(restoredElement) : null;
+         * TODO Commented out due to changes in the way the element IDs are passed (ie, a list rather than individual
+         * ones per job). This code is still useful long term, but I don't have time to write the necessary wrapper code
+         * around it right now. case Restore: String fileName = arguments.get(3); MediaPackageElement restoredElement =
+         * null; if (StringUtils.isNotBlank(fileName)) { restoredElement = restoreElement(channelId, mediaPackage,
+         * elementIds, fileName); } else { restoredElement = restoreElement(channelId, mediaPackage, elementIds, null);
+         * } return (restoredElement != null) ? MediaPackageElementParser.getAsXml(restoredElement) : null;
```
KatrinIhler: I'm not too happy about commented code in general (though that's not your fault), but I feel like this makes this snippet very hard to read. I don't think the space won by this change is worth it.

KatrinIhler (commented):
I haven't tested this, but the code looks reasonable to me (except for one tiny thing).

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
rute-santos: This change was unintentional (as well as others that changed the original formatting). I will revert all of them so that you only see what really changed.

KatrinIhler: Actually, most of the formatting changes - except for this one - seemed quite reasonable. Maybe put them in a separate commit? :)

rute-santos: Hi @KatrinIhler , I don't think this is related at all with this PR?

KatrinIhler: it's not, but since you already did the work - you could also file a separate PR? I'd approve it. :D

rute-santos: Ok, separate PR then. No problem!

KatrinIhler (approved):

PR #2329, opened 2021-02-15
Fixes Solr search failing when titles containing lots of upper case characters

### LINKED ISSUE #2308: episode search endpoint fails for sname values containing lots of uppercase characters.
**Describe the bug**
Solr search fails for specific serie titles. For instance when you have a series with as title a name that contains a lot of uppercase letters ( TestSeriesWithUpperCaseLetters ) and you use the /search/episode.json with this title as series name you get 0 results back. Looking at the query in SOLR I noticed that this gets cast to lowercase when doing the search. In most cases this works fine, but when there are a lot of uppercase characters SOLR seems to think that this is too much for a match. 

Is there a reason why everything gets cast to lowercase before doing the search? I removed this for the title field and this seems to work fine for us, but it might have unforeseen consequences. 

This is using opencast  9.1.

kristofkeppens:
When a series title contains a lot of upper case characters the solr search fails for this title because there occurs a cast to lower case before the search action. This fixes this for the title and id. Other fields still get cast to lower cause, although I think this should be safe to remove as well.

This fixes #2308

### DIFF
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -711,7 +711,7 @@ public class SeriesRestService {
       q.setText(text.toLowerCase());
     }
     if (StringUtils.isNotEmpty(seriesId)) {
-      q.setSeriesId(seriesId.toLowerCase());
+      q.setSeriesId(seriesId);
     }
     if (StringUtils.isNotEmpty(seriesTitle)) {
       q.setSeriesTitle(seriesTitle);
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -714,7 +714,7 @@ public class SeriesRestService {
       q.setSeriesId(seriesId.toLowerCase());
     }
     if (StringUtils.isNotEmpty(seriesTitle)) {
-      q.setSeriesTitle(seriesTitle.toLowerCase());
+      q.setSeriesTitle(seriesTitle);
     }
     if (StringUtils.isNotEmpty(creator)) {
       q.setCreator(creator.toLowerCase());
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -708,7 +708,7 @@ public class SeriesRestService {
       q.setEdit(edit);
     }
     if (StringUtils.isNotEmpty(text)) {
-      q.setText(text.toLowerCase());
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
I'm having a bit of trouble reviewing this PR. On the one hand, the change is tiny and it seems very reasonable that we *don't* want to convert the search query to lower case. The current code seems fishy. But:

- I was not able to reproduce #2308 with Opencast 9.1. Of course, this being some kind of search index, I can very well imagine the problem only occurring in larger data sets.
- I was unable to `git blame` my way to an answer to the question "why was `toLowerCase()` added in the first place?". This is pre-git stuff and thus lost somewhere in the squashed SVN history.
- If we agree that the `toLowerCase` should be removed here, there is no reason to not also remove the other `toLowerCase()` on the other fields, right? You said so in the issue as well.

So yeah, that's keeping me from approving this PR. But maybe these considerations are overkill and someone with a more "practical oriented mindset" want to review this PR instead :D

kristofkeppens:
@LukasKalbertodt I'll gladly remove the other toLowerCase()'es as well, I don't really see a negative impact to that, but I can imagine some unwanted side effects as well.

lkiesow:
> If we agree that the toLowerCase should be removed here, there is no reason to not also remove the other toLowerCase() on the other fields, right? You said so in the issue as well.

Actually, I think there is. Wouldn't it be pretty weird for one field to behave different from another field? And why would we call this a bug at one field but think it's correct at another field?

I think we can only guess why this was added, but more likely than not there is no real reason. The only thing I can come up with is the idea of matching regardless of case which this is obviously not achieving. So, whatever we do, we should do consistently.

LukasKalbertodt:
> Actually, I think there is. Wouldn't it be pretty weird for one field to behave different from another field? And why would we call this a bug at one field but think it's correct at another field? [...] So, whatever we do, we should do consistently.

I fully agree with you here. I think you misunderstood my prior comment as I wanted to say exactly that (I shouldn't use double negation :D).

But yeah I guess then the other fields should certainly be changed as well.

kristofkeppens:
@LukasKalbertodt I removed the toLowerCase from the other fields as well.

LukasKalbertodt (approved):
Diff looks good, the project compiles, all tests pass and the search API seems to work still. :+1: 

So this is kind of an approval, but with the two asterisks [mentioned above](https://github.com/opencast/opencast/pull/2329#issuecomment-786784081): I couldn't reproduce the bug this PR fixes and we are not 100% sure this won't break anything. But as long as everyone agrees this is likely an improvement and probably won't break anything, we can merge it and see.

KatrinIhler (approved):
I can't reproduce the bug either on a modified 8.10 (productive system, so lots of data in there). Maybe other conditions have to be met for this problem to appear.
Still, I think this change is reasonable, and if it breaks something, at least it will make us aware of a deeper problem going on? I also tested this a bit and searching still seems to work as expected. 
So I'm in favor of merging this. IMO, we shouldn't be afraid to remove code just because we don't know why it was added.
@LukasKalbertodt Do you want to merge this or should I? :)

PR #2332, opened 2021-02-17
Fix resolution scaling by removing `force_original_aspect_ratio`

### LINKED ISSUE #2297: Standard workflow fails when resizing ultra-wide input video because resolution is not divisible by 2
**Describe the bug**
Uploading an ultra wide (roughly 2.4:1), high resolution (2048x858) video results in a workflow processing error. The workflow runs this ffmpeg command:

```
ffmpeg -nostdin -nostats -i Spring.webm -c:v libx264 -crf 23 -maxrate 1200k -bufsize 2400k -profile:v high -level 4.0 -pix_fmt yuv420p -x264opts keyint=25:min-keyint=25:no-scenecut -movflags +faststart -filter:v scale=1280:-2:force_original_aspect_ratio=decrease,fps=25 -c:a aac -b:a 64k -ac 1 -f mp4 out.mp4
```

In particular, see `scale=1280:-2:force_original_aspect_ratio=decrease`. The `scale=1280:-2` means: width of 1280 pixels, height such that the aspect ratio stays and make sure it is divisible by 2. So in this case: 2048/1280 = 1.6; 858/1.6 = 536.25, which would be rounded to 536. That would be a good output resolution, BUT the `force_original_aspect_ratio=decrease` now leads to the width being decreased to 1279 because 1279/536 is closer to the original aspect ration than 1280/536 is. But now the width is not divisible by 2 which means the codec will throw an error. 

From my understanding, the `force_original_aspect_ratio` does not make sense here because in this setup, if it gets active, it would most of the time lead to a odd width or height. That's because if it gets active, it would usually decrease the width by only 1. @lkiesow made me aware of [the `force_divisible_by` parameter](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20190708154340.26514-1-lkiesow@uos.de/). Which would make sense here I guess?

**To Reproduce**
1. Download this awesome Blender movie: https://upload.wikimedia.org/wikipedia/commons/a/a5/Spring_-_Blender_Open_Movie.webm
2. Go to stable.opencast.org
3. Upload video with "Process upon upload and schedule" workflow and with output 720p resolution
4. Notice how the WF fails

**Environment Information**
 - Opencast version: 9.1

LukasKalbertodt:
Fixes #2297

Ideally, we would use `force_divisible_by=2`, but that option requires a fairly new ffmpeg version and we don't want to bump the ffmpeg version requirement in a minor Opencast update. For example, Ubuntu 20.04 default repositories contain a version that's too old. 

All the resize operations will still maintain the aspect ratio due to the `-2` as the height term. See the linked issue for why the `force_original_aspect_ratio` is not required.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/encoding/adaptive-streaming-movies.properties
+++ b/etc/encoding/adaptive-streaming-movies.properties
@@ -29,7 +29,7 @@ profile.adaptive-2160p.http.suffix = -preview.mp4
 profile.adaptive-2160p.http.ffmpeg.command = -i #{in.video.path} \
   -c:v libx264 -crf 23 -maxrate 10000k -bufsize 10000k -profile:v high -level 4.0 -pix_fmt yuv420p \
   -x264opts keyint=25:min-keyint=25:no-scenecut -movflags +faststart \
-  -filter:v scale=3840:-2:force_original_aspect_ratio=decrease,fps=25 \
+  -filter:v scale=3840:-2,fps=25 \
   -c:a aac -b:a 96k -ac 1 \
   -f mp4 #{out.dir}/#{out.name}#{out.suffix}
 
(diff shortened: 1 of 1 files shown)


lkiesow (approved):

KatrinIhler:
Since one month has passed and this looks reasonable, I'm gonna go ahead and merge this. :D

PR #2335, opened 2021-02-17
Distinguish Between Documentation and Configuration Checks

lkiesow:
This patch stops calling the configuration tests documentation tests and
puts them in a separate workflow with separate rules. This is hopefully
less confusing.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/checkstyle/check-docs.sh
+++ b/docs/checkstyle/check-docs.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -ue
+
+ret=0
+cd docs/guides
+
+for docs in admin developer; do
+  cd $docs
+  echo "Markdown doc $docs build log:"
+  mkdocs build &> mkdocs.log
+  cat mkdocs.log
+  if grep \
+       -e 'WARNING.*Documentation file' \
+       -e 'pages exist in the docs directory' \
+       -e 'is not found in the documentation files' \
+        mkdocs.log;
+    then
+      echo "$docs did not build correctly!"
+      ret=1
+  fi
+  cd ..
+done
+
+if ! npm test; then
+  echo "npm test failed!"
+  ret=1
+fi
+
+exit $ret
--- a/docs/checkstyle/check-docs.sh
+++ b/docs/checkstyle/check-docs.sh
@@ -7,17 +7,21 @@ cd docs/guides
 
 for docs in admin developer; do
   cd $docs
-  echo "Markdown doc $docs build log:"
-  mkdocs build &> mkdocs.log
-  cat mkdocs.log
-  if grep \
-       -e 'WARNING.*Documentation file' \
-       -e 'pages exist in the docs directory' \
-       -e 'is not found in the documentation files' \
-        mkdocs.log;
+  echo "Building $docs documentation…"
+  if mkdocs build &> mkdocs.log; then
+    if grep \
+        -e 'WARNING.*Documentation file' \
+        -e 'pages exist in the docs directory' \
+        -e 'is not found in the documentation files' \
+         mkdocs.log;
     then
       echo "$docs did not build correctly!"
+      cat mkdocs.log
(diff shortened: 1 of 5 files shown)


docs/checkstyle/check-docs.sh
```diff
@@ -0,0 +1,30 @@
...
+for docs in admin developer; do
+  cd $docs
+  echo "Markdown doc $docs build log:"
+  mkdocs build &> mkdocs.log
+  cat mkdocs.log
+  if grep \
+       -e 'WARNING.*Documentation file' \
+       -e 'pages exist in the docs directory' \
+       -e 'is not found in the documentation files' \
+        mkdocs.log;
```
LukasKalbertodt: Why these specific errors only? Why not also fail on "ERROR"? Or just check the exit code of `mkdocs`?

docs/checkstyle/check-docs.sh
```diff
@@ -0,0 +1,30 @@
...
+set -ue
+
+ret=0
+cd docs/guides
+
+for docs in admin developer; do
+  cd $docs
+  echo "Markdown doc $docs build log:"
+  mkdocs build &> mkdocs.log
+  cat mkdocs.log
```
LukasKalbertodt: Would it make sense to move this line into the `if` so that the log isn't always printed?

docs/checkstyle/check-docs.sh
```diff
@@ -0,0 +1,30 @@
...
+
+set -ue
+
+ret=0
+cd docs/guides
+
+for docs in admin developer; do
+  cd $docs
+  echo "Markdown doc $docs build log:"
+  mkdocs build &> mkdocs.log
```
LukasKalbertodt: This just silently fails if `mkdocs` or any of the extensions is not installed. Not relevant for CI, but maybe confusing if devs try to execute it locally?

LukasKalbertodt (commented):
Should `test-documentation-build.yml` be renamed to just `test-documentation.yml`?

I also left a couple inline remarks, but these have nothing to do with your change, but also apply already to the code before this PR. So feel free to ignore them. This PR is certainly an improvement.

docs/checkstyle/check-docs.sh
lkiesow: With the modification to this test, it does exactly that contrary to what happened with the `| tee ...` before.
This means, the test now fails if `mkdocs` fails.
These are just additional things to check for.

LukasKalbertodt: Oh yeah, of course, that makes sense!

lkiesow:
> Should test-documentation-build.yml be renamed to just test-documentation.yml?

Yes, but then we should do that for all the workflows.
So I keep it out of this pull request.

I hopefully addressed all your other comments

docs/checkstyle/check-docs.sh
```diff
@@ -0,0 +1,34 @@
...
+         mkdocs.log;
+    then
+      echo "$docs did not build correctly!"
+      cat mkdocs.log
+      ret=1
+    fi
+  else
+    echo mkdocs exited abnormally.
+    cat mkdocs.log
+  fi
```
LukasKalbertodt:
```suggestion
  else
    echo mkdocs exited abnormally.
    cat mkdocs.log
    ret=1
  fi
```

Maybe still let the script fail here?

LukasKalbertodt (commented):
Only one thing left!

LukasKalbertodt:
By the way: I cannot mark my own comments as "resolved". Weird. Feel free to resolve them yourself.

docs/checkstyle/check-docs.sh
lkiesow: Definitely. Good catch.

LukasKalbertodt (approved):

PR #2344, opened 2021-02-21
Auto-generate OAI-PMH database

lkiesow:
This addresses my incorrect assumption that OAI-PMH is no longer in any
workflows by default and setting up the database is thus optional. It is
still in the retract workflow which will currently fail if the database
was not manually prepared in advance.

This patch changes the code so that an SQL code is generated which works
with all databases and ensures modifications update the modification
date property on OAI-PMH entries. This will already generate a valid
OAI-PMH repository on its own but is still compatible to the previous
trigger approach.

If a user sets up database trigger, that will ensure that the
modification date is not set in the code but the moment it is actually
being persistent in the database. This works in favor of some less
resilient harvesters. For details see background below.

This does not modify any exising database structure. It will only effect
new database table creation and does not require a migration.

Background: What are the triggers for?
--------------------------------------

OAI-PMH is a harvesting protocol supporting selective harvesting,
meaning that you can e.g. ask only for records being modified after a
given date or time. This allows harvesters to pick up the harvesting
where they last stopped and not harvesting everything every time.

The problem with setting the modification date in code is that an entry
may be written to the database during an active harvest and the
modification date of this new entry may be before the response date of
the harvest. This would mean that if a harvester assumes all records are
contained in the last response, some may get lost.

The Guidelines for Harvester Implementers addresses this problem and
harvesters should deal with this:

- http://www.openarchives.org/OAI/2.0/guidelines-harvester.htm

> Items in a repository may change or be added during a harvest, or
> after a harvest within the same datestamp (i.e. the same day if the
> datestamp is YYYY-MM-DD). This means that to incrementally harvest
> from a repository, a harvester should overlap successive incremental
> harvests by one datestamp increment (i.e. one day if the granularity
> is YYYY-MM-DD). Furthermore, since it is repository implementation
> dependent whether changes that occur during the harvest will be
> reflected in the response, the from argument of the next incremental
> harvest should be based on the the responseDate returned in the first
> partial-list response of a sequence. When harvesting from repositories
> which use a datestamp granularity of one second, it is advisable to
> overlap by a small additional amount to account for any discrepancy
> between the reported responseDate and the time at the repository when
> any search necessary to answer the request was performed.

But at least VideoLaunge did not and the usage of SQL triggers did
protect that harvester implementation from tripping.

This means, the situation after applying this patch is:

- OAI-PMH works out of the box with no modifications
- If you have a very delicate harvester implementation, you can add
  triggers to improve the situation.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/oaipmh-persistence/src/main/java/org/opencastproject/oaipmh/persistence/OaiPmhEntity.java
+++ b/modules/oaipmh-persistence/src/main/java/org/opencastproject/oaipmh/persistence/OaiPmhEntity.java
@@ -81,11 +81,7 @@ public class OaiPmhEntity {
 
   /** The last modification date */
   @Column(
-      name = "modification_date",
-      insertable = false,
-      updatable = false,
-      // this is H2 syntax - Opencast uses a dedicated database dependent schema in production
-      columnDefinition = "TIMESTAMP AS CURRENT_TIMESTAMP()")
+      name = "modification_date")
   @Temporal(TemporalType.TIMESTAMP)
   private Date modificationDate;
 
@@ -107,6 +103,7 @@ public class OaiPmhEntity {
    * Default constructor without any import.
    */
   public OaiPmhEntity() {
+    this.modificationDate = new Date();
   }
 
   /**
@@ -125,6 +122,7 @@ public class OaiPmhEntity {
    */
   public void setMediaPackageId(String mediaPackageId) {
     this.mediaPackageId = mediaPackageId;
+    this.modificationDate = new Date();
   }
 
   /**
@@ -140,6 +138,7 @@ public class OaiPmhEntity {
    */
   public void setOrganization(String organization) {
     this.organization = organization;
+    this.modificationDate = new Date();
   }
 
   /**
@@ -157,6 +156,7 @@ public class OaiPmhEntity {
    */
   public void setDeleted(boolean deleted) {
     this.deleted = deleted;
+    this.modificationDate = new Date();
   }
 
   /**
@@ -174,6 +174,7 @@ public class OaiPmhEntity {
    */
   public void setSeries(String series) {
     this.series = series;
+    this.modificationDate = new Date();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


gregorydlogan:
This looks right to me, but I don't use OAI-PMH for anything.  @JamesUoM @ts23 any chance you can take a quick look at this and make sure it's sane?

JamesUoM:
If a OAIPMH object is created in the java with a modification time t_m  just before the start of a harvest t_h0 , but is not committed to the the database until after t_h0 then it will not get harvested during harvest h0 or the subsequent h1 unless h1 overlaps h0 completely. 

In theory you only need to set the modification time right before you commit the object to the database, but the time between setting the modification and the database transaction completing causes the edge case. Therefore I'm not sure if the additional modification date assignments do anything useful.

lkiesow:
Hi @JamesUoM, that's exactly the problem addressed in OAI's harvester implementation guide I cited above. Harvesters should expect that persisting data takes time and should not harvest from the last point in time they harvested but overlap the last harvest by a bit to avoid this issue.

That said, this does not remove the current behavior: If you continue to use the database trigger (which is still in our docs), you get the exact same behavior you had before. The Java code would set a date, but the trigger would just overwrite that.

But if you have a harvester which does not need that, or you don't use OAI-PMH at all, you can skip setting up the trigger which makes your life a bit simpler.

Right now, even if you do not use OAI-PMH, the retract workflow will fail with an SQL exception if Opencast runs with MySQL, MariaDB or PostgreSQL due to the current internal SQL syntax.

Or to put it in other words: Nothing changes for you, but things will be simpler for everyone else.

lkiesow:
@ts23, @JamesUoM ping. Next release is coming up and retraction is now broken for quite some time in Opencast 9. Would be nice to get that fixed. I can also ask someone else for a review if you don't have time?

gregorydlogan:
This appears to work for me.

PR #2351, opened 2021-02-22
Update Groups in Elasticsearch Indices directly

KatrinIhler:
With this PR, the JpaGroupRoleProvider no longer sends messages to update the ElasticSearch indices, but does this directly via method call. This affects the usual creating, updating and deleting of groups as well as the index rebuild.

See also #2311 for the same change to the CommentsService.

**Attention:** This PR is based on #2311 and should not be merged before that one.


KatrinIhler:
Replaced by #2572

PR #2354, opened 2021-02-22
Update Themes in Elasticsearch Indices directly

KatrinIhler:
With this PR, the ThemesDatabaseService no longer sends messages to update the Admin UI index, but does this directly via method call. This affects the usual creating, updating and deleting of themes as well as the index rebuild.

Be aware that in contrast to all other services, the themes service puts its data _only_ into the Admin UI index. I didn't notice this at first either, as you can see in the commit history... For the future, this can probably be removed completely (see discussion below).

See also #2311 and #2351 for similar changes to other services.

*Note:* Since the Create Theme UI seems to be broken, to test this you can use the rest endpoints to create, update and delete themes.

**Attention:** This PR is based on #2351 and should not be merged before that one.


lkiesow:
- If we shouldn't merge this, does it make sense to turn this into a draft pull request?
- Does it actually make sense to store something like themes in Elasticsearch? Wouldn't it suffice  to get the data directly from the services? That seems easier, shouldn't be a problem due to the limited amount of data and could spare us from maintaining this part of the index at all. – This is more a general question than a complaint about this pull request.

KatrinIhler:
@lkiesow 

1. I'm not sure. My thoughts were that someone might want to review multiple pull requests at once - e.g. the ones changing the services since the changes are very similar - and then merge them one after the other. However, I'm not sure whether that would cause conflicts that would need rebasing? The commits that are shared between PRs are completely identical, hash and all, so maybe Github is gonna be smart about it, but I don't know.

2. Good point... probably not. I bet this is just because all other tables in the Admin UI pull their data from the index. This is especially relevant since the themes are actually only in the Admin UI index (so they aren't identical after all!). I'm gonna look into this when I try to throw one index away, since that question is gonna come up anyway.

KatrinIhler:
Just realized that this would re-index Themes for both indices, even though they're only updated in the Admin UI Index. Applied a quick fix since the goal is to rip this out later anyway.

KatrinIhler:
@pascalseeland I've updated this PR so it's ready to review again! :D

What I did:
- Rebased it onto the new base PR #2531
- Cleaned up some more code
- Biggest change: Re-implemented the locking mechanism that existed in the BaseMessageReceiver (and that I had previously overlooked) so there will be no problems when updating the same theme at the same time.

As usual, if you have any questions, just ping me. :)

KatrinIhler:
@pascalseeland Had to rethink the locking for a moment, but I think I want to do it like this. Any chance you could get this merged today? Otherwise I'm gonna have to do some more rebasing for a PR that does locking for the rest of the index objects, and I don't wanna >_<

KatrinIhler:
Rebased onto #2595.

KatrinIhler:
Okay, and now it also works. :D

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/theme/IndexTheme.java
```diff
@@ -51,7 +58,7 @@
         "licenseSlideDescription", "licenseSlideBackground", "watermarkActive", "watermarkFile", "watermarkPosition" })
 @XmlRootElement(name = "theme", namespace = IndexObject.INDEX_XML_NAMESPACE)
 @XmlAccessorType(XmlAccessType.NONE)
-public class Theme implements IndexObject {
+public class IndexTheme implements IndexObject {
```
pascalseeland, line 61: Somehow I don't like the IndexTheme renaming. I fully understand why you rename it and I think its better than before but I'm not sure if IndexedTheme would not be a better name.

pascalseeland (approved):
Besides the naming issue it is fine. I'm also fine with merging it like this and in a second step also look at the series and event class which will probably have the same naming issue. But then please create an issue before merging so it does not get lost.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/theme/IndexTheme.java
KatrinIhler: Ha, I couldn't decide between those two! Maybe I'll change that later...

KatrinIhler: Ha, I couldn't decide between those two!

By the way: Series and Events do not have this issue! That's because they are representing data from different services, while the IndexTheme is a 1:1 representation of a Theme object, thus the naming problem.

The longterm plan is to stop indexing Themes anyway, since it doesn't make much sense, but this was out of scope for now. So this problem will hopefully resolve itself.

KatrinIhler:

PR #2368, opened 2021-03-01
Improve Email Workflow Operation

lkiesow:
This patch fixes several minor issues with the email workflow operation.
The patch is backwards compatible to not break old integrations.

- Until now, an `TO` header like `J. Doe <jdoe@example.com>` would cause
  the operation to fail due to it trying to use e.g. `J.` as a separate
  address. While the default stays the same, a new operation option
  `address-separator` has been introduced to work around it. Using `,`
  usually makes the most sense.

- Inclusion of one invalid mail address can now be skipped by using
  `skip-invalid-address`. This makes sense, especially is a field like
  _contributor_ is used which can easily contain an invalid field. This
  means, mails are semd only to what can be identified as valid
  addresses.

- Internal representation of arrays as strings with separators has been
  minimized, catching a few edge cases.

- Instead of causing a NullPointerException, the operation will be
  skipped if no recipients are defined.

---

For testing, use a workflow like this:

```xml
<definition xmlns="http://workflow.opencastproject.org">
  <id>send-email</id>
  <title>Send Email</title>
  <tags><tag>upload</tag></tags>
  <operations>
    <operation id="send-email">
      <configurations>
        <configuration key="to">${(catalogs['episode']['contributor'])}</configuration>
        <configuration key="subject">test</configuration>
        <configuration key="address-separator">,</configuration>
        <configuration key="skip-invalid-address">true</configuration>
      </configurations>
    </operation>
  </operations>
</definition>
```

Ingest an event like this:

```
curl -f -i -u admin:opencast http://localhost:8080/ingest/addMediaPackage/send-email \
  -F 'flavor=presentation/source' -F 'BODY=@some-file.txt' \
  -F title=test -F creator=test \
  -F contributor=username1 \
  -F contributor=username2 \
  -F contributor='J. Doe <jdoe@example.com>' \
  -F contributor=non-existing-user
```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/notification/EmailWorkflowOperationHandler.java
+++ b/modules/notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/notification/EmailWorkflowOperationHandler.java
@@ -40,8 +40,10 @@ import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Objects;
+
 import javax.mail.MessagingException;
-import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
 
 /**
@@ -67,6 +69,9 @@ public class EmailWorkflowOperationHandler extends AbstractWorkflowOperationHand
   static final String SUBJECT_PROPERTY = "subject";
   static final String BODY_PROPERTY = "body";
   static final String BODY_TEMPLATE_FILE_PROPERTY = "body-template-file";
+  static final String ADDRESS_SEPARATOR_PROPERTY = "address-separator";
+  static final String ADDRESS_SEPARATOR_DEFAULT = ", \t";
+  static final String SKIP_INVALID_ADDRESS_PROPERTY = "skip-invalid-address";
   static final String IS_HTML = "use-html";
 
   /*
@@ -78,6 +83,7 @@ public class EmailWorkflowOperationHandler extends AbstractWorkflowOperationHand
   @Override
   protected void activate(ComponentContext cc) {
     super.activate(cc);
+    logger.debug("Activating email workflow operation handler");
   }
 
   /**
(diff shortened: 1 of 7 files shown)


rute-santos:
This looks good to me. I suggest that you also add tests for new configuration options (different separator and skip-invalid-address true/false).

lkiesow:
@rute-santos, thanks for taking a look at this. I've slightly modified the original draft so that the delimiter is not fixed any longer, but you can freely choose whatever you want. Due to the current defaults, it should still be backwards compatible. I've also added tests for the two new options. This now is ready to be reviewed.

modules/notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/notification/EmailWorkflowOperationHandler.java
```diff
@@ -94,28 +100,33 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+    String[] to = processDestination(workflowInstance, operation, TO_PROPERTY);
+    String[] cc = processDestination(workflowInstance, operation, CC_PROPERTY);
+    String[] bcc = processDestination(workflowInstance, operation, BCC_PROPERTY);
+
+    if (to.length + cc.length + bcc.length == 0) {
+      logger.info("No recipients. Skipping operation.");
+      return createResult(srcPackage, Action.SKIP);
+    }
+
+    String subject = applyTemplateIfNecessary(workflowInstance, operation, SUBJECT_PROPERTY, ", ");
```
rute-santos: Why ", " and not let the service use its default DEFAULT_DELIMITER_FOR_MULTIPLE? Just curious. Also, I think this should be a constant as it's used in a few places.

lkiesow: My idea was that if you have two subjects, the resulting text would just look more natural like this:

- this is a subject a, this is subject b

rather than

- this is a subject a,this is subject b

But this should seldom happen and honestly, I don't really care either way. What would you prefer?

- Using `, ` (as a local const)
- Using the current default `,`
- Or even making `, ` the new default

Pick your favorite ;-)

rute-santos: I don't care which one either, but it should be a constant (local or not). So how about option 3 to simplify?

lkiesow: @rute-santos, done. A new default is set and the inline strings have been removed. Can you take another shot at the review?

rute-santos (approved):
Looks good to me!

PR #2387, opened 2021-03-08
Update Managed ACLs in Elasticsearch Indices directly

KatrinIhler:
With this PR, the AclService no longer sends messages to update the Elasticsearch indices, but does this directly via method call. This affects the updating and deleting of ManagedACLs.

An Index Rebuild for this Service is actually not necessary since the fields are already updated by the SeriesService and AssetManager respectively. So the AclService is no longer called during an Index Rebuild.

See #2311, #2351 and #2354 for similar changes to other services.

*Note:* Showing the correct managed ACL currently seems to be broken in the Admin UI for both series and events. I tested this by taking a look at the Event object with a debugger... :(

**Attention:** This PR is based on #2354 and should not be merged before that one.


KatrinIhler:
Rebased on develop. 

TODO: Add locking mechanism.

KatrinIhler:
Rebased onto #2595. Ready for review again!

pascalseeland (approved):

KatrinIhler:

PR #2391, opened 2021-03-10
Perform `check-availibility` WF check with system user

LukasKalbertodt:
This is to prevent problems when the user starting the workflow is not
authorized to access the distributed assets. This should be rare in
theory, but in practice, the authorization for static files might use
a cache for permissions. This can easily lead to workflows failing
because the the availability check encounters a 403 response due to
permission caching.

Using the system user here doesn't seem like a problem: the availability check was
never supposed to test any authorization-related things, but simply
whether the published file exists at all.

I tested this patch on a multi-node system and it seems to work. And in particular, it solves the problem I encountered where workflows would fail because of that.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
+1 for the patch but the tests fail with NullPointerExceptions

PR #2418, opened 2021-03-16
Update Test Server Builds

lkiesow:
This patch includes the preparation for updating the test servers into
the main repository builds by building the distribution tarballs on
`push` events and uploading them to S3.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -122,6 +122,7 @@ jobs:
             hunspell \
             hunspell-de-de \
             procps \
+            s3cmd \
             sox \
             tar \
             tesseract-ocr \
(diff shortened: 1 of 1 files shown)


lkiesow:
Not sure if anyone wants to try testing this or if we just want to deploy this and see if it works (I tested it on my account)

gregorydlogan:
Imma do this. I can see a sec issue already I suspect :)

lkiesow:
> I can see a sec issue already I suspect

If you suspect a security issue, I'm not going to deploy this :D
Can you give me more details?

If it's that this workflow is also run on pull request, be aware that from such a run you cannot access secrets even if you modify the Actions workflow. So you can't push arbitrary code via pull request. It only works if it is already in the repository.

gregorydlogan:
Yep, that's what I missed.  I thought this was firing on PRs, which presents the obvious issue of echoing the creds to the log, or sending them somewhere via some other method.

My only other complaint here is that this won't fire on develop since it's not prefixed with `r/`.

lkiesow:
> My only other complaint here is that this won't fire on develop since it's not prefixed with r/.

Good catch. Hopefully fixed that.

gregorydlogan:
Lgtm

PR #2428, opened 2021-03-21
Fix Workflow Instance Endpoint Documentation

lkiesow:
The date filter does not actually apply to workflow start dates but to
the event metadata.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
```diff
@@ -444,8 +444,8 @@ public Response getWorkflowsAsXml(@QueryParam("state") List<String> states, @Que
           @RestParameter(name = "seriesTitle", isRequired = false, description = "Filter results by series title", type = STRING),
           @RestParameter(name = "creator", isRequired = false, description = "Filter results by the mediapackage's creator", type = STRING),
           @RestParameter(name = "contributor", isRequired = false, description = "Filter results by the mediapackage's contributor", type = STRING),
-          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by workflow start date.", type = STRING),
-          @RestParameter(name = "todate", isRequired = false, description = "Filter results by workflow start date.", type = STRING),
+          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by event date.", type = STRING),
```
KatrinIhler: It's actually the created date (so not e.g. the bibliographic start date, which I would expect), maybe add that here?

lkiesow:
```suggestion
          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by `created` Dublin Core metadata .", type = STRING),
```
?

KatrinIhler:
```suggestion
          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by event created date.", type = STRING),
```
?

KatrinIhler: Actually, I take that back, I got confused by the use of "created" in the key, but I took a closer look at it just now and it's actually the bibliographic start date, not the created date. We have too many dates...

lkiesow: „The bibliographic start date“ is the Dublin Core „created“ field, isn't it?
So… which suggestion do you want me to apply here?

KatrinIhler: No, I think it's "temporal", actually.
(So "Start Date" in the event metadata is "temporal" in the catalog, and "Created" is "created"... which would actually make sense?)

lkiesow: > No, I think it's "temporal", actually.

No, it's not. The DC field `temporal` contains start and end date information used by capture agents for recordings. Thats the technical start (and end) date. That's not what is being used by this endpoint. You can test this easily on http://develop.opencast.org since the uploaded events have no `temporal` metadata, only `created` and you can still use this field for filtering the workflows.


Are you happy with:
```suggestion
          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by the date contained in the `created` Dublin Core metadata. Use UTC dates in the format <pre>2020-12-24T08:30:00Z</pre>.", type = STRING),
```

---

Example DC metadata to show that's not temporal (from develop.opencast.org):


```xml
<dublincore>
  <dcterms:created xsi:type="dcterms:W3CDTF">2021-04-07T00:03Z</dcterms:created>
  <dcterms:creator>Rosatrieu</dcterms:creator>
  <dcterms:description>The video shows how espresso is typically made.</dcterms:description>
  <dcterms:extent xsi:type="dcterms:ISO8601">PT54.335S</dcterms:extent>
  <dcterms:identifier>ID-espresso-video</dcterms:identifier>
  <dcterms:isPartOf>ID-wiki-commons</dcterms:isPartOf>
  <dcterms:license>CC-BY-SA</dcterms:license>
  <dcterms:title>Espresso_video</dcterms:title>
</dublincore>
```

KatrinIhler: Okay, so I spent probably way too much time on this?

1. "temporal" contains the bibliographic start date, not the technical one. Reasoning:
 
The config mapping dublin core entries to event metadata (`org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-common.cfg`) says so:

```
property.startDate.inputID=temporal
property.startDate.label=EVENTS.EVENTS.DETAILS.METADATA.START_DATE

property.created.inputID=created
property.created.label=EVENTS.EVENTS.DETAILS.METADATA.CREATED
```

And when I schedule an event via the UI and later change the bibliographic (!) start date, the temporal field contains that value, not the technical start date. Actually, the technical start date then isn't in the dublin core catalog at all - but you only see that when the two dates are not the same.

The events on develop are uploaded via a script, so that's probably why they don't have a temporal field - when I upload one via the UI, it does.

2. The workflow code is using the bibliographic start date, not the created date. The code says so:

WorkflowServiceSolrIndex:

    if (mp.getDate() != null) {
      doc.addField(CREATED_KEY, mp.getDate());
    }

(I know it says "created", but the Mediapackage Start Date is actually the bibliographic one:)

CommonEventCatalogUIAdapter:

     // Mediapackage start date is set by event metadata start date. The "created" metadata field is not used.
    MetadataField startDate = abstractMetadata.getOutputFields().get("startDate");
    mediaPackage.setDate(sdf.parse((String) startDate.getValue()));

Though it doesn't make much of a difference, since these dates are very often identical, and created is probably used as a fallback somewhere.

----------------------------------------------------
So I guess I'm still arguing for "temporal" here. :D

gregorydlogan:
Did we ever come to a conclusion on this?  I know @lkiesow said @KatrinIhler is right, but I don't want to merge this until I'm sure this is good to go...

KatrinIhler:
@gregorydlogan I think the actual change still has to be done, but that should be quick.

KatrinIhler:
@lkiesow Ping :D

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
lkiesow: Then, based on what you found out, let's go with:

```suggestion
          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by the creation date metadata field of the media package which is usually populated via the `temporal` Dublin Core element. Use UTC dates in the format <pre>2020-12-24T08:30:00Z</pre>.", type = STRING),
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
```diff
@@ -455,8 +455,8 @@ public Response getWorkflowsAsXml(@QueryParam("state") List<String> states, @Que
...
           @RestParameter(name = "creator", isRequired = false, description = "Filter results by the mediapackage's creator", type = STRING),
           @RestParameter(name = "contributor", isRequired = false, description = "Filter results by the mediapackage's contributor", type = STRING),
-          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by workflow start date.", type = STRING),
-          @RestParameter(name = "todate", isRequired = false, description = "Filter results by workflow start date.", type = STRING),
+          @RestParameter(name = "fromdate", isRequired = false, description = "Filter results by the creation date metadata field of the media package which is usually populated via the `temporal` Dublin Core element. Use UTC dates in the format <pre>2020-12-24T08:30:00Z</pre>.", type = STRING),
+          @RestParameter(name = "todate", isRequired = false, description = "Filter results by the creation date metadata field of the media package which is usually populated via the `temporal` Dublin Core element. Use UTC dates in the format <pre>2020-12-24T08:30:00Z</pre>.", type = STRING),
```
KatrinIhler, line 459: Please don't hate me. \*ducks\*

```suggestion
          @RestParameter(name = "todate", isRequired = false, description = "Filter results by the bibliographic start date of the event, which maps to the `temporal` field in the Dublin Core catalog. Use UTC dates in the format <pre>2020-12-24T08:30:00Z</pre>.", type = STRING),
```

KatrinIhler (changes requested):

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/endpoint/WorkflowRestService.java
lkiesow: The code actually looks at the date attribute of the media package and not at any metadata directly. That field is populated by values from the Dublin Core catalog (either `temporal` or (if `temporal` does not exist) `created`). That's why I don't agree with your phrasing. But since we can't find a common ground here, let's leave it as it was.

PR #2435, opened 2021-03-22
Added Download Dropdown to Series LTI-Tools

Arnei:
This PR allows users to pass a new URL parameter "download". It enables a download button in the Series view on each episode.
The button will open a dropdown that allows the user to download individual tracks from that episode.

The PR is sponsored by the University of Ulm

Bildschirmfoto vom 2021-03-22 14-39-47

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/lti/src/utils.ts
+++ b/modules/lti/src/utils.ts
@@ -1,5 +1,5 @@
 import { parse as parseQuery } from "query-string";
-import { Track } from "./OpencastRest";
+
 
 export function parsedQueryString() {
     return parseQuery(window.location.search);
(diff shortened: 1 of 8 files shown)


wsmirnow:
I've got an error `Fehler: Cannot read property 'resolution' of undefined`.  The episodes contains adaptive streaming publications. These tracks looks like

```
{
  "id": "track-1",
  "type": "presentation/delivery",
  "ref": "track:reference-id",
  "transport": "HLS",
  "mimetype": "application/x-mpegURL",
  "tags": {
    "tag": ["360p-quality", "engage-download", "engage-streaming"]
  },
  "url": "https://streaming.server.edu/opencast/smil:engage-player_track-1_presentation.smil/playlist.m3u8",
  "duration": 900052,
  "live": false
}
```
The video tag is missing.

Arnei:
> I've got an error `Fehler: Cannot read property 'resolution' of undefined`. The episodes contains adaptive streaming publications. 

Thanks for the quick feedback. I have added error handling for tracks that are missing the video tag.

As per suggestion I have also tried to exclude tracks from adaptive streaming publications in the list of downloadable tracks.

Arnei:
Removed the check for the LTI role "Instructor", so that every LTI authenticated person can see the button. Reasoning is that the users of this button are most likely not gonna be the instructors, but the students, who typically rank lowest on the ranking list of access rights.

modules/lti/src/utils.ts
```diff
@@ -1,5 +1,75 @@
...
+  }
+  return s.charAt(0).toUpperCase() + s.slice(1);
+}
+
+/**
+ * Sorts an array of tracks alphabetically by their type
+ * If tracks have the same type, they are then sorted by resolution
+ * @param tracks the tracks to be sorted
+ */
+export const sortByType = (tracks: Track[]) => {
```
pascalseeland: I would prefer to move this function into its own ts file. Something like trackUtils.ts. As such, utils will only contain general usable functions and track specific stuff can go into trackUtils

modules/lti/src/utils.ts
```diff
@@ -1,5 +1,75 @@
...
+  })
+  return tracks;
+}
+
+/**
+ * Helper function for sortByType. Sorts a resolution string by the number it contains
+ * @param descA resolution of the first track
+ * @param descB resolution of the second track
+ */
+const sortResolutions = (descA: string, descB: string) => {
```
pascalseeland: Either also move this function to trackUtils or make it a general function for comparing two resolutions. Also I would prefer to rename it compareResolutions, as it does not sort them, like the sortTracks above

pascalseeland (changes requested):
Please take a look at utils.ts. Otherwise this is fine

modules/lti/src/utils.ts
wsmirnow: I am pretty sure @KatrinIhler is not a fan of utility classes, that never been used as intended. I share her opinion. If we have a specific case, where we can reuse this function, then we can move it out.

wsmirnow: Ok, It is currently in a utils file, sorry for noise. But do you really want more utils files? really? :D

wsmirnow (changes requested):
Can you document the new `download` option in `/docs/guides/admin/docs/modules/ltimodule.md` please.

modules/lti/src/utils.ts
wsmirnow: Ok, It is currently in a utils file, sorry for noise. But do you really want more utils files? really? :D

KatrinIhler: I have been summoned? :D (It's especially funny since I was just talking to Malte about his utils file. :rofl:)

I can't be bothered to look at this closely right now, so just a general request: no more utility files than absolutely necessary, and if you can name them better than something-utils, please do. :D

wsmirnow (approved):

pascalseeland (approved):

modules/lti/src/utils.ts
pascalseeland: The main reason for my request was because I did not want to have a utils file which contains a lot of miscellaneous function.

KatrinIhler:
Approved twice, so I'm gonna merge this.

PR #2462, opened 2021-03-30
Automatically closing resources on the getAsset AM rest endpoint

gregorydlogan:
Using autoclose to close input stream for fetched resources.  This was causing filehandle leaks in Osnabruck testing on S3.

### Your pull request should…

* [x] have a concise title
* [ ] ~close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists~
* [x] pass automated tests

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
@@ -71,6 +71,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


KatrinIhler:
@gregorydlogan This doesn't work for me. It's no longer possible to download assets via the Admin UI, I get 500 Server Error: Stream closed. :/

gregorydlogan:
Assigning this to me so that I get around to fixing it.

KatrinIhler:
@gregorydlogan Since I was the one that started this whole thing - our S3 problem was solved by setting the amount of max connections higher, so it's possible this is not actually a problem. I also set max connections to 1 on our test system and then downloaded assets multiple time, so it seems like the connections are closed at least if nothing goes wrong...

gregorydlogan:
Still worth checking out. The docs imply the stream isn't automatically closed when it falls out of scope, so I'd like to run this past a profiler. I'll get to that eventually :)

KatrinIhler:
@gregorydlogan Would it make sense to make this a draft for now?

gregorydlogan:
This is not a pressing concern anymore, and this isn't the fix anyway.

PR #2476, opened 2021-04-09
Remove broken admin-frontend test

lkiesow:
This is identical to #2471 which should have gone into `r/9.x`. Sorry.

---

Currently, the Opencast admin interface frontend does not build in
Germany any longer since we switched to daylight saving time (UTC+2).
The admin interface frontend module fails with:

    Expected spy put to have been called with:
      [ 'filter', 'furniture', undefined, '2014-12-31T23:00:00.001Z/2015-01-03T22:59:59.999Z' ]
    but actual calls were:
      [ 'filter', 'furniture', undefined, '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z' ].

Looking at the test, it does not make a lot of sense since it would
compensate for local time zones by just setting a fixed time after the
conversion, rendering the test mostly useless anyway.

Unfortunately, in some times zones it may also happen that only one of
the dates changes which will still let the test fail.

Since the test does not make a lot of sense anyway, this patch removes
that particular test to fix the Opencast build process.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/test/test/unit/shared/directives/tableFilterDirectiveSpec.js
+++ b/modules/admin-ui-frontend/test/test/unit/shared/directives/tableFilterDirectiveSpec.js
@@ -191,20 +191,6 @@ describe('adminNg.directives.adminNgTableFilter', function () {
                 expect(filter.prefilled.to).toBe(false);
                 expect(filter.prefilled.from).toBe(false);
             })
-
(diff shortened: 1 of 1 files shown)


pascalseeland (approved):

lkiesow:
Closed in favor of #2478

PR #2477, opened 2021-04-09
Fixed test for daylight saving time

rubends:
Fixed the broken admin-frontend test for daylight saving time.


lkiesow:
Will be merged from `r/9.x`

PR #2493, opened 2021-04-14
Fixed "No response from service" for videogrid

### LINKED ISSUE #2137: Videogrid Remote fails with BadMessageException
**Describe the bug**
When trying to process a video conference recording with ~50 webcam videos, the video grid remote throws an exception, causing the videogrid WOH to fail:

```
2020-12-10T17:43:33,445 | WARN  | (HttpChannel:579) - /videogrid/videogrid org.eclipse.jetty.http.BadMessageException: 400: Unable to parse form content
```
My suspicion is that the bug is caused by the way the Videogrid WOH sends ffmpeg commands to the Videogrid Service. Currently, the WOH generates all necessary commands and sends them as to the service as a single argument.

**To Reproduce**
1. Have an Opencast setup with at least two servers (e.g. admin-presentation and worker).
2. Ingest a new mediapackage. Use /ingest/addPartialTrack to add a ton of videos that all start at different times and whose duration overlaps.
3. Start a workflow on the mediapackage that runs the Videogrid WOH on the ingested videos.

**Environment Information**
 - OS: Fedora 31
 - Opencast 9.x

Arnei:
The videogrid operation would sometimes fail for many videos (more than ~50).
This was due to the WOH sending all the ffmpeg commands to the service as a single job argument, which broke the service if it became to long/large.
This commit changes the WOH and the service in a way that instead of sending all commands in a single job, each command is now send as its own job.

Resolves #2137.

How to test:
* Have a distributed Opencast installation (seperate admin and worker nodes)
* Create a workflow with the Videogrid Operation (An example: https://github.com/elan-ev/opencast-bigbluebutton-integration/blob/master/post-archive/bbb-upload-9-2.xml)
* Use the Ingest Service to create a mediapackage, use "AddPartialTrack" to add a couple of presenter tracks and then start your workflow
* If you've got (processing) time, try to test with a lot (more than ~50) presenter tracks that all have different start times when added with "AddPartialTrack"

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
This reviews *so* much better with the `?w=1` query parameter

PR #2494, opened 2021-04-14
Fix Processing of Unicode Titles

lkiesow:
This patch fixes an issue with serializing Dublin Core catalogs with
multi-byte Unicode characters which would result in vroken XML data,
which in turn would cause workflow failures an potential data loss.

The reason for for this is an old, unpatched bug in Xalan:

- https://stackoverflow.com/q/11952289/2352895
- https://issues.apache.org/jira/browse/XALANJ-2560

To test this, ingest a video using:

    curl -f -i -u admin:opencast \
      http://localhost:8080/ingest/addMediaPackage/fast \
      -F mediaUri=https://www.pexels.com/video/854982/download/ \
      -F title="I 🖤 Opencast"

Note that while this fixes the processing and broken data, this does not
yet catch all occurences of broken XML. Notably, there is still a
problem in media package internals causing playback problems.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
@lkiesow I'm pretty sure GitHub Actions doesn't support UTF8 in the branch name, since this appears to have broken the tests in such a way that *there aren't any records of the tests* :D

Edit Nevermind, page refresh appears to have resolved the issue.

PR #2500, opened 2021-04-15
Make media package handle multi-byte Unicode characters

lkiesow:
Parsing media packages had a strange internal behavior of parsing the
XML, possibly modifying the XML, serializing the XML again to then parse
the XML again. The serializer used in this process could not handle
multi-byte Unicode characters properly which broke the entire process.

This was used as part of the search service which caused the Media
module and player to break on such events.

This patch fixes the problem by directly parsing the XML in most cases
and using a different serializer in the other cases.

Note that this will not fix already broken XML!
It just prevents breaking new XML.

---

Due to the very strange parsing behavior previously present in the code,
I would like to get two separate reviews on this.
If you happen to have further questions, please do not hesitate to ask.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
@@ -27,7 +27,6 @@ import org.opencastproject.mediapackage.identifier.Id;
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


Arnei (approved):
I've tested this on my local r/9.x AllInOne by trying to through multi-byte unicode characters in every text field and file name I could think of, and then test out Opencast functions. So far it seems to work.

Might be good to note that this won't fix events with XMLs that are already broken, those stay broken.

Can't really comment on the code, seems to be reasonable at least.

PR #2501, opened 2021-04-15
Handle multiple creators in Media Module

lkiesow:
This patch fixes the problem that the media module would just display a
single random creator, even if the event has multiple creators.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
+++ b/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
@@ -657,8 +657,12 @@ function($, bootbox, _, alertify, jsyaml) {
         }
         tile = tile + '<div class="infos">';
 
-        if (data.dcCreator) {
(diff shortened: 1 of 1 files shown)


Arnei (approved):
Works in my local AllInOne Opencast 9 installation.
Would be nice if I could see all creators somehow, but that problem is arguably not part of this PR.
Bildschirmfoto vom 2021-04-20 12-34-37

Also makes me wonder why the search endpoint returns a field with a single 'dcCreator' in the first place :thinking:

lkiesow:
> Also makes me wonder why the search endpoint returns a field with a single 'dcCreator' in the first place thinking

That's a good question.

gregorydlogan:
Well, this is better than picking a single one.  I'd like to see a search for "alice" bring up things by "alice, bob, charlie", but considering the new media module is coming Soon(tm) I think we can deal with this for now.

PR #2512, opened 2021-04-16
Add Build Date in User Interface

### LINKED ISSUE #1215: Show build date
Opencast lists its version and commit hash (if build from source). Having the build date a well would be really helpful, especially on the test servers:

Screenshot from 2019-11-06 16-38-50

lkiesow:
This patch adds the build date to the build information visible to the
system administrator in the admin interface.

This fixes #1215

Screenshot from 2021-04-16 20-34-18

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/index.html
+++ b/modules/admin-ui-frontend/app/index.html
@@ -114,9 +114,20 @@
 
     <footer id="main-footer" ng-controller="NavCtrl">
       <div class="default-footer">
-        <div ng-if="version.version" class="meta">
+        <div ng-if="version.version"
+             class="meta"
+             title="{{ 'BUILD.VERSION' | translate }}">
           Opencast {{ version.version }}
-          <span with-role="ROLE_ADMIN"> - {{ version.buildNumber || 'undefined' }}</span>
+          <span ng-if="version.buildNumber"
+                with-role="ROLE_ADMIN"
+                title="{{ 'BUILD.COMMIT' | translate }}">
+            - {{ version.buildNumber || 'undefined' }}
+          </span>
+          <span ng-if="version['last-modified']"
+                with-role="ROLE_ADMIN"
+                title="{{ 'BUILD.DATE_DESC' | translate }}">
+            - {{ 'BUILD.BUILT_ON' | translate }} {{ version['last-modified'] }}
+          </span>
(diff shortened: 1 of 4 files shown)


Arnei (commented):
Works in my local installation. I got some questions about usability though.

Why is this hidden in a title tag? It's not like there's no space in the footer left , and the additional visual clutter seems minimal. Kind of makes this feature something only for people that are "in the know".

Maybe I am not enough of a sys-admin, but it would not have been immedeately clear to me that this date is supposed to be the latest build date. It could also have been the date of when the latest commit was commited or something. Maybe changing the string to "Last build on {Date}" would make the intended meaning more clear to some. (Or have the date in the footer and the explanation for what that date is in the title)

gregorydlogan:
> Maybe I am not enough of a sys-admin, but it would not have been immedeately clear to me that this date is supposed to be the latest build date. It could also have been the date of when the latest commit was commited or something. Maybe changing the string to "Last build on {Date}" would make the intended meaning more clear to some. (Or have the date in the footer and the explanation for what that date is in the title)

I too would prefer it to be visible rather than hidden in alt text.  I suspect this is the *commit* date since it's being written on the hash.  Adding something like "Built on" makes sense since this is (probably) for nontechincal users who wouldn't know what to do with the hash.  That being said, adding that extra text would require new translation keys, which would need to go into `develop` - perhaps a follow-on PR?

lkiesow:
Updated this and set some additional explainations.

Arnei (approved):
Yup, that seems fairly unambiguous.

But, as Greg already mentioned, might need to go into develop now instead of r/9.x?

lkiesow:
> That being said, adding that extra text would require new translation keys, which would need to go into develop - perhaps a follow-on PR?

No. The rules have changes for this and this should be fine for `r/9.x`:

> Accepting patches for minor releases
> 
> - Patches for minor releases must not
>    - …
>    - Modify __existing__ translations

I did not modify any. I just added new ones.

mliradelc (approved):
Code Wise and Testing for me is all ok. 
If there are no objections about if it should be here or in the develop branch I can merge it.

I will wait after the technical meeting of tomorrow.

PR #2520, opened 2021-04-18
Do not build against Java 8

lkiesow:
Opencast 10 does not support Java 8 any longer, so we don't need to run
our tests against it. This makes Actions run the tests against Java 13
instead.

The Opencast build fails with newer versions of Java. Mid-term, it would
make sense to build against all LTS versions we support, plus the latest
version of Java.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -8,6 +8,7 @@ on:
       - 'docs/checkstyle/**'
       - 'docs/log4j/**'
       - 'assemblies/**'
+      - '.github/**'
   push:
     paths:
       - 'pom.xml'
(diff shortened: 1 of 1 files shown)


gregorydlogan:
While I'm fine testing against 13, the relevant section of the proposal from Nov 6 reads

```
Opencast 10: Formal support for JDK 11, with a provisional "it should work" for newer versions


Going forward, I would like to see a given version of Opencast support the latest two LTS JDKs as a general rule.

```

Since jdk 17 isn't generally available until September I see 13 (which seems like the most widely used) as a good stepping stone.

lkiesow:
That's why all LTS + latest: The LTS is what we support and thus should definitely test, the latest is what we need eventually since it will evolve into the next LTS at some point. Unfortunately, 13 seems the newest version Opencast builds against right now and setting compatibility to JDK 11 already breaks the build :(

PR #2529, opened 2021-04-19
Expose some of the S3 client configuration for assets

KatrinIhler:
This exposes some of the S3 client configuration for assets so we can change it.
This includes:
- Max Connections
- Connection Timeout
- Max Retries

Additionally I improved the related documentation a bit.

### DIFF
--- a/docs/guides/admin/docs/modules/awss3distribution.md
+++ b/docs/guides/admin/docs/modules/awss3distribution.md
@@ -82,18 +82,24 @@ Opencast Service Configuration
 The Opencast AWS S3 Distribution service has five configuration keys, which can be found in the
 `org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg` configuration file.
 
-|Key name|Value|Example|
-|--------|-----|-------|
-|org.opencastproject.distribution.aws.s3.distribution.enable|True to enable S3 distribution, false otherwise|true|
-|org.opencastproject.distribution.aws.s3.region|The AWS region to set|us-west-2|
-|org.opencastproject.distribution.aws.s3.endpoint|The endpoint of AWS S3 service. Only used with S3 compatible service|https://s3.service.com|
-|org.opencastproject.distribution.aws.s3.path.style|True to enable path style access URL for bucket, false otherwise|false|
-|org.opencastproject.distribution.aws.s3.bucket|The S3 bucket name|example-org-dist|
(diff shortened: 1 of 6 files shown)


gregorydlogan:
Can we add these config bits to the distribution code as well?  I can see someone expecting in both places.

KatrinIhler:
@gregorydlogan I was afraid you'd ask that...

KatrinIhler:
@gregorydlogan Added the same options to distribution and couldn't resist cleaning up the code and the documentation a bit...

docs/guides/admin/docs/modules/awss3distribution.md
```diff
@@ -82,18 +82,24 @@ Opencast Service Configuration
...
+| org.opencastproject.distribution.aws.s3.max.connections              | Number of max connections                                                                | 50                           |                                                                                       |
+| org.opencastproject.distribution.aws.s3.connection.timeout           | Connection timeout in ms                                                                 | 10000                        |                                                                                       |
+| org.opencastproject.distribution.aws.s3.max.retries                  | Number of max retries                                                                    | 100                          |                                                                                       |
+| job.load.aws.s3.distribute                                           | Distribute job load                                                                      | 0.1                          |                                                                                       |
```
gregorydlogan, line 100: So, I hate to be a complainer here, but these aren't exposed on the *other* S3 integration...

Is there a reason you're exposing these?  I don't personally care, but I feel like the chance of someone needing to mess with these values is *so* slim that if they actually *do* need to, then they probably need some kind of larger refactor rather than to mess with the loads.

That being said, if this is something that the sponsor wants I'm cool with it - but it should be in both places :D

gregorydlogan (approved):
Approved pending comment!

docs/guides/admin/docs/modules/awss3distribution.md
KatrinIhler: I didn't expose the job loads, the job loads are already part of the configuration. I just added them to the documentation! Please don't shoot the messenger! :D (I totally agree with you btw, but then, most of the s3 code should probably be shared anyway...)

etc/org.opencastproject.distribution.aws.s3.AwsS3DistributionServiceImpl.cfg
```diff
@@ -47,9 +47,20 @@ org.opencastproject.distribution.aws.s3.distribution.enable=false
 
 # Enable presigned URL
 # Leave this commented out to use presigned URL fow AWS S3
```
mtneug: I think this is not true. Presigned URLs are not used by default.

```suggestion
```

KatrinIhler: True, I fixed the value, but missed the comment. Thx!

PR #2531, opened 2021-04-20
Hand over Elasticsearch Index to Services for Index Rebuild

KatrinIhler:
This PR contains some basic work to improve the index rebuild that was previously part of the pull requests for each service. This way those PRs can hopefully be independent from one another.

This does the following:
- Hand over the AbstractSearchIndex to the service via repopulate() instead of just giving the index name - this way the service can later do the index update directly.
- Expose the AbstractSearchIndex interface via OSGI so we can later depend on it.
- Use OSGI Annotations for the indices so annotated services (e.g. GroupRoleProvider) can also use them.

The first two commits were part of #2311, the last was in #2351. @pascalseeland  You already reviewed the first one, so this should be quick! :)


pascalseeland (approved):

KatrinIhler:
Gonna merge since since Pascal approved. :)

PR #2546, opened 2021-04-28
Default Visibility in Admin Interface

lkiesow:
This patch updates the default visibility in the admin interface so
users will only see events and series they have write access to. This
prevents them trying to modify events they cannot, resulting in errors
for those users.

While I was originally against this, the recent development of having
end users on the interface has shown that this is something everyone
configures anyway and which prevents lots of user errors. Hence, it
probably makes sense to set this as default after all.

This obviously has no effect on admin users since they have write access
to everything anyway.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


wsmirnow (approved):
+1

Issue #1111, opened 2019-10-02
AWS AutoScalingTerminationStateServiceTest regularly fails


lkiesow:
The `AutoScalingTerminationStateServiceTest.testLifeCyclePolling` fails on an irregular basis making both buildbot and [Travis](https://api.travis-ci.org/v3/job/592679363/log.txt) fail from time to time:

```
[INFO] --- maven-surefire-plugin:3.0.0-M3:test (default-test) @ opencast-termination-state-aws ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.opencastproject.terminationstate.aws.AutoScalingTerminationStateServiceTest
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.583 s <<< FAILURE! - in org.opencastproject.terminationstate.aws.AutoScalingTerminationStateServiceTest
[ERROR] testLifeCyclePolling(org.opencastproject.terminationstate.aws.AutoScalingTerminationStateServiceTest)  Time elapsed: 3.44 s  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.opencastproject.terminationstate.aws.AutoScalingTerminationStateServiceTest.testLifeCyclePolling(AutoScalingTerminationStateServiceTest.java:132)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   AutoScalingTerminationStateServiceTest.testLifeCyclePolling:132 expected:<1> but was:<0>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

lkiesow:
Looking at the high number of today's pull requests you could see this error popping up multiple times, breaking a number of test builds. @gregorydlogan, assuming that you wrote those tests, can you take a look?

<details>
<summary>Mudule build log of error copied from buildbot</summary>

```
[INFO] Building Opencast :: termination-state-aws 9-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:3.0.0:clean (default-clean) @ opencast-termination-state-aws ---
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0-M1:enforce (enforce-maven) @ opencast-termination-state-aws ---
[INFO] 
[...]

PR #2553, opened 2021-04-30
Automatic cleaning of working file repository

lwillmann:
If one configures it workflows to produce a 'failed-zip' in case of errors (See: https://docs.opencast.org/develop/admin/#workflowoperationhandlers/zip-woh/) the zip file with the failed workflow is stored in the working file repository and may stay there until forever if it is not removed manually. At our customer Switch the 'failed.zip's have taken a large amount of diskspace from the Working File Repository. We decided to introduce an automatic cleaning process similar to the `org.opencastproject.workspace.impl.WorkspaceCleaner` to cleanup the Working File Repository continuously. 

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -140,6 +140,23 @@ org.opencastproject.workspace.cleanup.period=86400
 org.opencastproject.workspace.cleanup.max.age=2592000
 
 
+######### Working File Repository Cleanup #########
+
+# The scheduled period in seconds, at which a working file repository cleanup operation is performed.
+# 86400 seconds equals 24 hours.
+# Default value: -1 (Disable cleanup scheduler)
+org.opencastproject.working.file.repository.cleanup.period=86400
+
+# The maximum age a file must reach in days before a deletion of the file in the workspace cleanup operation is
+# performed.
+# Default value: -1 (max age will never be reached)
+org.opencastproject.working.file.repository.cleanup.max.age=100
+
+# A comma separated lists of collections in the working file repository that shouldbe cleaned up.
+# Default value: Not provided (no collection will be cleaned up)
+org.opencastproject.working.file.repository.cleanup.collections=failed.zips
+
+
 ######### ACTIVE MQ BROKER #########
 
 # This configures the connection to the Active MQ broker so that we can send and receive messages.
--- a/modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
+++ b/modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
@@ -160,9 +160,8 @@ public class WorkingFileRepositoryImpl implements WorkingFileRepository, PathMap
       try {
         garbageCollectionPeriodInSeconds = Integer.parseInt(period);
       } catch (NumberFormatException e) {
-        logger.warn("Invalid configuration for working file garbage collection period ({}={})",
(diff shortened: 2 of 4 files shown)


modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
```diff
@@ -141,6 +152,47 @@ public void activate(ComponentContext cc) throws IOException {
 
     registeredMXBean = JmxUtil.registerMXBean(workingFileRepositoryBean, JMX_WORKING_FILE_REPOSITORY_TYPE);
 
+    // Determine garbage collection period
+    int garbageCollectionPeriodInSeconds = -1;
+    String period = StringUtils.trimToNull(
+            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_PERIOD_KEY));
+    if (period != null) {
+      try {
+        garbageCollectionPeriodInSeconds = Integer.parseInt(period);
```
lkiesow, line 161: I'm a fan of failing fast if there is a known problem instead of unexpectedly going into a somewhat unknown state. In other words, if the configuration is completely broken, let the service fail and let the user fix the configuration.

etc/custom.properties
```diff
@@ -140,6 +140,23 @@ org.opencastproject.workspace.cleanup.period=86400
 org.opencastproject.workspace.cleanup.max.age=2592000
 
 
+######### Working File Repository Cleanup #########
```
lkiesow, line 143: Why make this a system-wide configuration and not a service configuration?

lkiesow (changes requested):

lkiesow:
I remember not cleaning up the acls since this caused problems in a few cases. Did that get resolved by now?

etc/custom.properties
lwillmann: Well I thought cleaning up the working file repository is essencialy the same as cleaning up the workspace. The Workspace Cleanup is configured in the custom.properties file, thus I placed the configuration for the WFR cleanup just below in the same file. Should I change the location?

modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
lwillmann: @lkiesow  Tried to fix this, now the build fails in another module for org.opencastproject.terminationstate.aws.AutoScalingTerminationStateServiceTest 
can you check this?

lkiesow: That seems to be https://github.com/opencast/opencast/issues/1111

lwillmann:
> I remember not cleaning up the acls since this caused problems in a few cases. Did that get resolved by now?
Sorry, I don't know about the problem. Can you be more specific on how this is related to this PR.

lkiesow:
>> I remember not cleaning up the acls since this caused problems in a few cases. Did that get resolved by now?
>
> Sorry, I don't know about the problem. Can you be more specific on how this is related to this PR.

Judging by https://github.com/opencast/opencast/pull/1380, this should probably be fine:

> If we cleanup the ACLs at the end of the workflow, the last workflow update that happens after will set an empty acl into the index. So set an exception as we do in other workflows.

PR #2554, opened 2021-04-30
Retract publications before deleting events

lwillmann:
Currently the external API deletes events without previously retracting any possible publications. To enable deletion of events in one step, this PR uses the same behavior as in the AdminUI where publications are retracted before the event is deleted (see: https://github.com/opencast/opencast/pull/862). As this is a change of the external API a new version is introduced.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -107,6 +107,7 @@ import org.opencastproject.util.doc.rest.RestQuery;
 import org.opencastproject.util.doc.rest.RestResponse;
 import org.opencastproject.util.doc.rest.RestService;
 import org.opencastproject.util.requests.SortCriterion;
+import org.opencastproject.workflow.api.WorkflowDatabaseException;
 import org.opencastproject.workflow.api.WorkflowInstance;
 
 import com.entwinemedia.fn.Fn;
@@ -187,6 +188,12 @@ public class EventsEndpoint implements ManagedService {
   /** Subtype of previews required by the video editor */
   private static final String DEFAULT_PREVIEW_SUBTYPE = "preview";
 
+  /** ID of the workflow used to retract published events */
+  private static final String RETRACT_WORKFLOW = "retract.workflow.id";
+
+  /** Default ID of the workflow used to retract published events */
+  private static final String DEFAULT_RETRACT_WORKFLOW = "delete";
+
   /** The logging facility */
   private static final Logger logger = LoggerFactory.getLogger(EventsEndpoint.class);
 
@@ -198,6 +205,8 @@ public class EventsEndpoint implements ManagedService {
   private String previewSubtype = DEFAULT_PREVIEW_SUBTYPE;
 
   private Map<String, MetadataField> configuredMetadataFields = new TreeMap<>();
+
+  private String retractWorkflowId = DEFAULT_RETRACT_WORKFLOW;
 
   /** The resolutions */
   private enum CommentResolution {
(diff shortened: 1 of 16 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -170,7 +171,7 @@
 
 @Path("/")
 @Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0,
-            ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0 })
+            ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0, ApiMediaType.VERSION_1_6_0 })
```
KatrinIhler: I think you missed some @Produces for methods, for example for getEventScheduling(). Some methods can be available for less versions than the rest, mostly when they were added later. Sorry. :D (I can show you my commit later.)

lwillmann: Thanks for the review @KatrinIhler. I added now hopefully all missing @produces. It seems there was also one missing for 1.5.0 in the statistics endpoint.

KatrinIhler: Yeah, I found that one as well. :D

lwillmann:
Update to external API Version 1.7.0 because of #2572

KatrinIhler:
@lwillmann @pascalseeland  If, improbable as it may be, this gets into OC 10, you don't actually need a new API version since I already introduced 1.6 for that very release. :) If it's OC 11, you do.

lwillmann:
@KatrinIhler ok, I must admit the API versioning stuff is a little bit confusing to me. Anyhow I did split the PR now into two commits one is just the feature and the other one for the versioning stuff. So @pascalseeland it is up to you ... if it gets into OC10 just take the first commit.

pascalseeland:
Sorry for the delay. As we discussed in the technical meeting we agree that adding a return value is fine the first commit is fine to be merged.

KatrinIhler:
@gregorydlogan Maybe you can merge the first commit before cutting OC 10?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -421,14 +433,45 @@ public Response getEventMedia(@HeaderParam("Accept") String acceptHeader, @PathP
...
-
-    return Response.noContent().build();
+          throws SearchIndexException, UnauthorizedException {
+    final Opt<Event> event = indexService.getEvent(id, externalIndex);
+    if (event.isNone()) {
+      return RestUtil.R.notFound(id);
+    }
+    final Runnable doOnNotFound = () -> {
+      try {
+        externalIndex.delete(Event.DOCUMENT_TYPE,id.concat(getSecurityService().getOrganization().getId()));
```
gregorydlogan: This line no longer works with the index changes -> incorrect number of arguments.

gregorydlogan (dismissed):
Might make sense to rebase onto `r/10.x`, and drop the commit which removes the API version change.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
KatrinIhler: Sorryyyy

lwillmann: Ok, rebased to `r/10.x` and changed the `delete` method arguments ... maybe just one time this gets merged before Katrin does the next change ;-)

KatrinIhler: Sounds like a challenge ;)

pascalseeland (approved):

KatrinIhler:
@lwillmann Do I see this correctly in that the API version isn't checked on deletion, so the behavior is the same whether I use API version 1.6.0 or 1.7.0? In that case introducing a new version doesn't really have an effect.

lwillmann:
@KatrinIhler You are right, even if there is no version 1.7.0 in Opencast 10 ;-). I thought there is no need to check the version because it doesn't do something critical if there is no publication and it is kind of a bugfix for previous versions if there is a publication. As discussed, I've added more documentation to clear things up.

KatrinIhler:
I'm okay with this now and since Pascal also approved this, I'm gonna merge it.

PR #2558, opened 2021-05-03
Add organization ID to the S3 distribution object path

mtneug:
When distributing files with S3 the organization ID is curretly not part
of the S3 object path. Adding it brings it on par with the download
distribution.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
Does this not need any migration of old paths? E.g. will retraction of old paths still work?

mtneug:
@lkiesow I think a migration is not necessary. When retracting the [(old) URL is read from the mediapacakge](https://github.com/opencast/opencast/blob/72240d3079bed34a1e87a1355f45cca3e6b1b9d1/modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java#L810-L825). I have not tested this, however.

KatrinIhler (approved):
I think Matthias is right. My test case left something to be desired, but it tried to retract the old URLs without the organization. Rest looks reasonable as well.

PR #2559, opened 2021-05-03
Allow proper mapping of tenant hostnames to URLs

mtneug:
Opencast has some assumptions about how hosts are accessed in different
organizations. Especially the schema and port number are assumed to be
equal or in certain cases just the default protocol port number.

This changes the host mapping configuration to explicitly specify a URL.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -52,26 +52,17 @@ name=Opencast Project
 #
 # Example Configuration:
 #
-# prop.org.opencastproject.host.admin-presentation.opencast.com=tenant1.admin-presentation.opencast.com
-# prop.org.opencastproject.host.ingest.opencast.com=tenant1.ingest.opencast.com
-# prop.org.opencastproject.host.worker.opencast.com=tenant1.worker.opencast.com
+# prop.org.opencastproject.host.admin-presentation.opencast.com=http://tenant1.admin-presentation.opencast.com
+# prop.org.opencastproject.host.ingest.opencast.com=https://tenant1.ingest.opencast.com
+# prop.org.opencastproject.host.worker.opencast.com=https://tenant1.worker.opencast.com:8433
 #
 # Important notices:
 #  - Values read from here are stored in the database (table oc_organization_node) and will not be removed if a node is
 #    removed from this list.
 #  - Do not include schema or port in the URLs.
 #
-# Default: localhost
-# prop.org.opencastproject.host.localhost = localhost
-
-# Port for this tenant.
-#
-# When run behind a proxy server, set this to -1 since most proxies won't forward the
-# original port and set the corresponding field in the request to this value.
-#
-# Value: <integer port number>
-#
-port=8080
+# Default: http://localhost:8080
+# prop.org.opencastproject.host.localhost = http://localhost:8080
 
 # Identifier of the Administrative role.
 #
--- a/modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
+++ b/modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
@@ -164,7 +164,11 @@ public final class SecurityUtil {
 
   /** Extract hostname and port number from a URL. */
   public static Tuple<String, Integer> hostAndPort(URL url) {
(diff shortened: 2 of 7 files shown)


etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -52,26 +52,17 @@ name=Opencast Project
 #
 # Example Configuration:
 #
-# prop.org.opencastproject.host.admin-presentation.opencast.com=tenant1.admin-presentation.opencast.com
-# prop.org.opencastproject.host.ingest.opencast.com=tenant1.ingest.opencast.com
-# prop.org.opencastproject.host.worker.opencast.com=tenant1.worker.opencast.com
+# prop.org.opencastproject.host.admin-presentation.opencast.com=http://tenant1.admin-presentation.opencast.com
+# prop.org.opencastproject.host.ingest.opencast.com=https://tenant1.ingest.opencast.com
```
lkiesow, line 56: Does this mean that you cannot have a tenant that reacts to both http and https any longer?

modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
```diff
@@ -164,7 +164,11 @@ public static String getSystemUserName(ComponentContext cc) {
 
   /** Extract hostname and port number from a URL. */
   public static Tuple<String, Integer> hostAndPort(URL url) {
-    return tuple(StringUtils.strip(url.getHost(), "/"), url.getPort());
+    int port = url.getPort();
+    if (port < 0) {
```
lkiesow, line 168: Previously, `port=-1` meant any port. Does this now mean something else?

lkiesow (commented):
Haven't really looked at this closely. Just some questions…

modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
mtneug: When a URL is parsed, `-1` means that no port is explicitly specified in the URL. We can also check for `-1`, but every port number below 0 is invalid.

etc/org.opencastproject.organization-mh_default_org.cfg
mtneug: No, if I understand the code correctly.

The schema only matters for the `ServiceRegistryEndpoint` `/available.{xml|json}`. pyCA, for instance, should then only use the schema and port specified here. Note that the other service registry endpoints will still report the URL of the service registration (i.e. `org.opencastproject.server.url`). I would expect them to be tenant-specific as well, but I don't know if this is intentionally limited to this one endpoint.

Opencast currently determines the tenant using only the hostname. You should be able to route traffic through different ports or a different protocol schema. As mentioned in the updated docs, this also means that it is currently not possible to have `tennen1` use `http://localhost:8080` and `tennen2` `http://localhost:8081`.

So including a schema here is necessary for the `ServiceRegistryEndpoint` and to not explicitly specifying a port number.

mtneug: mmh I was wrong. The port is used to differentiate between tenants. Will update the docs.

modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
mtneug: Can this be resolved?

mliradelc: @lkiesow is this solved for you? 🙂

mtneug: ?

gregorydlogan (approved):
LGTM, but I also don't have a multitenant system to test with...

modules/common/src/main/java/org/opencastproject/security/util/SecurityUtil.java
lkiesow: Don't wait for me. If this is fine, it's fine. It was just something I noticed. I didn't do a proper review of this.

lkiesow:
@mtneug, can you add a release notes snippet for this?
See https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs/releasenotes

PR #2560, opened 2021-05-03
Add additional s3 operations

mtneug:
This PR adds additional S3 distribution operations:

* Rename existing `publish-aws` und `retract-aws` to `publish-engage-aws` and `retract-engage-aws` to make it clear where they publish to
* Add `retract-partial-aws` (This uses the `engage` channel as well, but there is currently no partial retract for other channels)
* Add `publish-configure-aws`
* Add `retract-configure-aws`

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


CGreweling:
@mtneug Just one question. Is it possible to use the OSGI ``@Reference`` Binding in the Java Class instead of the OSGI-XML here?

gregorydlogan (changes requested):
Concur with @CGreweling, this should use the newer OSGi bits.  Otherwise looks good to me.

mtneug:
@CGreweling, @gregorydlogan: can you tell me how this would be done in this case? I have two OSGi services per class with varying references some even missing.

gregorydlogan:
@mtneug this style: https://github.com/opencast/opencast/blob/develop/modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentAdminRoleProviderImpl.java

Edit: The above is not what the issue is, the issue is that multiple instances of the same class need to use different dependencies and we're not sure how to do that with the annotations.

gregorydlogan:
@mtneug aside from the conflicts I'd say this is ready to go in.

gregorydlogan (approved):

mtneug:
@gregorydlogan: conflicts are fixed and since the new workflows don't use S3 they don't need to be updated.

PR #2562, opened 2021-05-03
Add "iFrame Resizer" library to LTI tools

### LINKED ISSUE #1892: iFrameResizer integration into Opencast

LukasKalbertodt:
Fixes #1892  

See commit messages for motivation and context.

This replaces #2080. That other PR had some problems that are addressed here.

If you want to test this, I used this dummy HTML parent page:

<details>

```
<html>
<head>
  <title>Test Iframe</title>
  <script src="iframeResizer.min.js"></script>
</head>
<body>
  <h1>Yes hello here is doggo</h1>
  <p>Some very important text</p>
  <iframe id="myframe" src="http://localhost:8080/ltitools/index.html?subtool=series&series=e1f32683-1095-4110-aaf5-6c3674f02574"></iframe>
  <p>This is a footer, mister</p>
  <script>
    iFrameResize({ log: true }, '#myframe')
  </script>
</body>
```

</details>

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/lti/package-lock.json
+++ b/modules/lti/package-lock.json
@@ -6832,6 +6832,11 @@
       "resolved": "https://registry.npmjs.org/iferr/-/iferr-0.1.5.tgz",
       "integrity": "sha1-xg7taebY/bazEEofy8ocGS3FtQE="
     },
+    "iframe-resizer": {
+      "version": "4.3.2",
+      "resolved": "https://registry.npmjs.org/iframe-resizer/-/iframe-resizer-4.3.2.tgz",
+      "integrity": "sha512-gOWo2hmdPjMQsQ+zTKbses08mDfDEMh4NneGQNP4qwePYujY1lguqP6gnbeJkf154gojWlBhIltlgnMfYjGHWA=="
+    },
     "ignore": {
       "version": "4.0.6",
       "resolved": "https://registry.npmjs.org/ignore/-/ignore-4.0.6.tgz",
--- a/modules/lti/package.json
+++ b/modules/lti/package.json
@@ -11,6 +11,7 @@
     "bootstrap": "^4.3.1",
     "i18next": "^17.0.16",
     "i18next-browser-languagedetector": "^3.0.3",
+    "iframe-resizer": "^4.3.2",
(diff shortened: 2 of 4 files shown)


Arnei:
Works in Firefox, the iframe automatically sizes its height. Does not size its width to accommodate for really long words or for the video player (when clicking on an event), but I've been told that is to be expected. Also the iframeResizer spams the console.

Does not work in Iridium (Version 2020.11 (Fedora Build) (64-Bit)), not quite sure yet if that is due to my testing setup or not. Possibly interesting error messages:
`Blocked autofocusing on a <input> element in a cross-origin subframe.`
`iframeResizer.js:848 Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('http://localhost:8080') does not match the recipient window's origin ('null').`

LukasKalbertodt:
Thanks for trying this out!

> Also the iframeResizer spams the console.

That's due to the `{ log: true }` in my test HTML. Shouldn't happen otherwise.

> Does not size its width 

My understanding is that the height is most important here. The use case is most likely an iframe inside an LMS where the iframe width would be `100%` (i.e. same width as the parent container) and only the height needs to be adjusted according to the content.

> Does not work in Iridium [...]

I think I saw a similar error message once when trying locally. I think the browsers actually just get really confused with `file://` and `localhost` origins. 
And in any case: this PR doesn't really contain any code on its own, but only includes a library.

Arnei (approved):

PR #2566, opened 2021-05-04
Add support for basic authentication with Elasticsearch

mtneug:
Elasticsearch allows for basic authentication. This adds two optional
settings for specifying a username and password. By default these are
null which will deactivate the use of basic authentication (as is the
current behavior).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/search/pom.xml
+++ b/modules/search/pom.xml
@@ -88,6 +88,15 @@
       <artifactId>httpcore-osgi</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>${httpcomponents-httpclient.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient-osgi</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.lucene</groupId>
       <artifactId>lucene-core</artifactId>
       <version>${lucene.version}</version>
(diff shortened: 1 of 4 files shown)


modules/search/pom.xml
```diff
@@ -87,6 +87,15 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpcore-osgi</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
```
lkiesow: Isn't all of `httpclient` contained in `httpclient-osgi`?

mtneug: I have no clue. The dependency analyzer has given me some trouble saying I didn't declare that dependency. The CI build does the same now for the `admin-ui` and I have no idea why 😞 .

lkiesow:
Updated the PR. Can you check this still works for you?

gregorydlogan:
Linking to the relevant docs to enable this on the ES side would be nice, as would adding it to the docker-compose files in the devel-dependency directory if possible.

mtneug:
> Updated the PR. Can you check this still works for you?

Forgot to answer. This is now running and working.

mtneug:
Conflict is fixed

mtneug:
> Linking to the relevant docs to enable this on the ES side would be nice, as would adding it to the docker-compose files in the devel-dependency directory if possible.

I added the reference to the documentation. I wouldn't add it to the developer compose files as these should be kept simple lowering the risk of errors due to misconfiguration.

gregorydlogan:
@mliradelc can we move this forward?

lkiesow:
Since the release is fast approaching, I dare to steal this from you @mliradelc.
We have had two devs from two institutions working on this and reviewing each other.
→ Let's get this in.

PR #2573, opened 2021-05-07
Refactoring workflows and encoding profiles

Arnei:
Based on and replaces #1504, as the original author cannot find the time to finish. For more details, see #1504.

To test this, Lars proposed this checklist:

<details>
  <summary>Plan for testing:</summary>
  
* Testing new workflows
 
   * [ ]  Upload presentation and publish directly
   * [ ]  Upload presenter and publish directly
   * [ ]  Upload two streams and publish directly
   * [ ]  Upload presentation, send to editor and publish from there
   * [ ]  Upload presenter, send to editor and publish from there
   * [ ]  Upload two streams, send to editor and publish from there
   * [ ]  Upload presentation, send to editor, cut out element and publish
   * [ ]  Upload presenter, send to editor, cut out element and publish
   * [ ]  Upload two streams, send to editor, cut out element and publish
   * [ ]  Retract event and publish it again
   * [ ]  Republish metadata
   * [ ]  Upload via studio (if the new workflow is used directly, needs more tests otherwise)

* Testing the upgrade path
 
   * [ ]  Upload presentation using old workflow, send to editor and publish from there using the new workflow
   * [ ]  Upload presenter using old workflow, send to editor and publish from there using the new workflow
   * [ ]  Upload two streams using old workflow, send to editor and publish from there using the new workflow
   * [ ]  Upload presentation using old workflow, send to editor, cut out element and publish using the new workflow
   * [ ]  Upload presenter using old workflow, send to editor, cut out element and publish using the new workflow
   * [ ]  Upload two streams using old workflow, send to editor, cut out element and publish using the new workflow
   * [ ]  Retract event published using old workflow and publish it again using the new workflow
   * [ ]  Republish metadata of event published using the old workflow

</details>

Regarding unanswered comments on the previous PR:
> I ran all the listed tests and only Retract event published using old workflow and publish it again using the new workflow does not work. Retracting from engage and publishing to engage again works. But the event is not retracted from the external api and oai-pmh.

The new workflows only publishes to or retracts from engage per default. Users are expected to add other operations to the workflows themselves, based on the publication channels they wish to support. So if you previously have published to the external api and oai-pmh, you should add the related retract operations to your retract workflow.

> Opencast has some encoding profiles hardcoded.

This PR adds any hardcoded profile that I found back to `opencast.properties`.


etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+  #{if-height-geq-1260} \
+  #{if-height-geq-1800}
+
+# Preview image for the player, shown before the movie is started
+profile.player-preview.http.name = cover image for engage
+profile.player-preview.http.input = visual
+profile.player-preview.http.output = image
+profile.player-preview.http.suffix = -player.jpg
+profile.player-preview.http.ffmpeg.command = -i #{in.video.path} -frames:v 1 \
+  -filter:v select=eq(n\\,1)+lt(t\\,#{time}),fps=1,reverse,scale=-2:720 \
```
lkiesow: This should suffer from the same problem fixed by https://github.com/lkiesow/opencast/commit/3a8e413c0efbeee737264aa6b021afead9c3bfce

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+  #{out.dir}/#{out.name}#{out.suffix}
+
+# Slide preview images as shown in the player
+profile.player-slides.http.name = slide preview image for engage
+profile.player-slides.http.input = visual
+profile.player-slides.http.output = image
+profile.player-slides.http.suffix = .#{time}.jpg
+profile.player-slides.http.ffmpeg.command = -ss #{time} -i #{in.video.path} \
+  -r 1 -frames:v 1 -filter:v yadif,scale=160:-1 \
```
lkiesow: We recently discussed in the technical meeting if we do not want to drop `yadif`. I think most were in favor of dropping it. Here is our chance.

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+    -c:v libx264 -profile:v high -level 4.0  -crf 23 \
+    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 800k -bufsize 1600k \
+    -c:a aac -b:a 64k -ac 2 \
+    -movflags faststart #{out.dir}/#{out.name}#{out.suffix.480p-quality} \
+  -filter:v scale=w=1280:h='trunc(1280/dar/2)*2',setsar=1,fps=25 -pix_fmt yuv420p \
+    -c:v libx264 -profile:v high -level 4.0  -crf 23 \
+    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 1200k -bufsize 2400k \
+    -c:a aac -b:a 96k -ac 2 \
+    -movflags faststart #{out.dir}/#{out.name}#{out.suffix.720p-quality} \
+  #{if-height-geq-900} \
```
lkiesow: By now, [Resolution Based Encoding](https://docs.opencast.org/r/9.x/admin/#workflowoperationhandlers/encode-woh/#resolution-based-encoding) also supports `if-width-or-height-geq-<width>-<height>` which may be a better choice?

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+profile.adaptive-parallel.http.input = visual
+profile.adaptive-parallel.http.output = visual
+profile.adaptive-parallel.http.jobload = 4
+profile.adaptive-parallel.http.suffix.480p-quality = -480p.mp4
+profile.adaptive-parallel.http.suffix.720p-quality = -720p.mp4
+profile.adaptive-parallel.http.suffix.1080p-quality = -1080p.mp4
+profile.adaptive-parallel.http.suffix.1440p-quality = -1440p.mp4
+profile.adaptive-parallel.http.suffix.2160p-quality = -2160p.mp4
+profile.adaptive-parallel.http.ffmpeg.command.if-height-geq-900 = \
+  -filter:v scale=w=1920:h='trunc(1920/dar/2)*2',setsar=1,fps=25 -pix_fmt yuv420p \
```
lkiesow: We previously used `scale=1280:-2` instead of `scale=1280:1280:force_original_aspect_ratio=decrease:force_divisible_by=2` since we were late to the game and `force_divisible_by`  requires a recent FFmpeg version we did not want to require in a minor version bump.

But since this is for the next major version, I would suggest using a scale filter like this:

```
ffmpeg
  -nostdin -nostats
  -i Spring.webm
  -c:v libx264 -crf 23 -maxrate 1200k -bufsize 2400k -profile:v high -level 4.0 -pix_fmt yuv420p
  -x264opts keyint=25:min-keyint=25:no-scenecut
  -movflags +faststart
  -filter:v scale=1280:1280:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25
  -c:a aac -b:a 64k -ac 1
  -f mp4 out.mp4
```

Documentation can be found at http://ffmpeg.org/ffmpeg-all.html#Options-85

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+  -r 1 -frames:v 1 -filter:v yadif,scale=-1:54 \
+  #{out.dir}/#{out.name}#{out.suffix}
+
+# Image to video
+profile.image-movie.work.name = image to video
+profile.image-movie.work.input = image
+profile.image-movie.work.output = visual
+profile.image-movie.work.suffix = -image-video.mp4
+profile.image-movie.work.ffmpeg.command = -loop 1 -i #{in.video.path} \
+  -c:v libx264 -r 30 -t #{time} -pix_fmt yuv420p \
```
lkiesow, line 151: This may result in:

```
[libx264 @ 0x509bac0] width not divisible by 2 (225x225)                                                                                                                                                                                      
Error initializing output stream 0:0 -- Error while opening encoder for output stream #0:0 - maybe incorrect parameters such as bit_rate, rate, width or height
```

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+  -movflags faststart #{out.dir}/#{out.name}#{out.suffix}
+
+profile.mp4-preview.dual.http.name = preview video (picture-by-picture)
+profile.mp4-preview.dual.http.input = visual
+profile.mp4-preview.dual.http.output = visual
+profile.mp4-preview.dual.http.suffix = -preview-composite.mp4
+profile.mp4-preview.dual.http.ffmpeg.command = -i #{in.video.path} \
+  #{compositeCommand} \
+  -c:v libx264 -preset veryfast -crf 23 -profile:v baseline -pix_fmt yuv420p -tune film \
+  -c:a aac -strict -2 -ar 22050 -ab 64k \
```
lkiesow: Why `-strict -2`? That shouldn't be necessary any longer (was years ago when the `aac` encoder was marked experimental).

etc/encoding/opencast.properties
```diff
@@ -0,0 +1,243 @@
...
+  #{out.dir}/#{out.name}#{out.suffix}
+
+profile.mp4-preview.http.name = preview video
+profile.mp4-preview.http.input = visual
+profile.mp4-preview.http.output = visual
+profile.mp4-preview.http.suffix = -preview.mp4
+profile.mp4-preview.http.ffmpeg.command = -i #{in.video.path} \
+  -filter:v yadif,scale=-2:360 \
+  -c:v libx264 -preset veryfast -crf 23 -profile:v baseline -pix_fmt yuv420p -tune film \
+  -c:a aac -strict -2 -ar 22050 -ab 64k \
```
lkiesow:
```suggestion
  -c:a aac -ar 22050 -ab 64k \
```

etc/workflows/publish.xml
```diff
@@ -181,21 +63,16 @@
...
-        description="Remove temporary processing artifacts">
+      id="cleanup"
+      fail-on-error="false"
+      description="Remove temporary processing artifacts">
       <configurations>
+        <!-- On systems with shared workspace or working file repository -->
+        <!-- you want to set this option to false. -->
         <configuration key="delete-external">true</configuration>
-        <!-- FixMe Don't clean up ACLs until workflow service no longer looks for them in the WFR. -->
         <configuration key="preserve-flavors">security/*</configuration>
```
lkiesow: Do we still need this? Removing the comment suggest that the problem is fixed? In that case, we should  do a full cleanup.

lkiesow (changes requested):

etc/encoding/opencast.properties
Arnei: So in this particular example, instead of 

`#{if-height-geq-900}`

it would be more desirable to use

`#{if-width-or-height-geq-1600-900}`

to avoid upscaling videos with sufficient height but absurdly small width?

etc/encoding/opencast.properties
Arnei: This encoding profile is also used in the partial import WOH for generating gap fillers before concatenation, I assume it would not be wise to change the resolution of those. Maybe it would be better to create a seperate encoding profile for the partial import WOH.

Edit: Nevermind, just realized this doesn't matter.

etc/workflows/publish.xml
Arnei: Seems like the ACLs are still required for parsing ActiveMQ messages, so we have to keep them.

etc/encoding/opencast.properties
Arnei: While I'm otherwise absolutely in favor of using it, I'm wondering if `force_divisble_by` is still too new an option. If I install ffmpeg on my Fedora 32 via packet manager, the version I'm getting (4.2) does not yet include it.

lkiesow: It's over a year old now, we control most FFmpeg builds on servers and can just require a specific version for a new major version.

EDIT: It's actually even from 2019

gregorydlogan:
FTR, this does *not* work with the current Debian packaged ffmpeg.

lkiesow:
Fixed merge conflicts by merging with `develop`.

gregorydlogan:
This may or may not cause issues with #3084, depending on what's actually in these changes (I haven't looked).

lkiesow (approved):

PR #2575, opened 2021-05-08
Fixing unchecked directory list() call

gregorydlogan:
This change resolves an issue report on the users list where retracting would sometimes cause NPEs when the list().length check was done.  Reading the docs closely, .list() returns null in the case where the File is *not* a directory, or if an IO error occurs.  Since an IO error will likely show up as a *ton* of other errors, we're going to check if the .list() call returns null, and assume that a null here is because the File isn't pointing at a directory.  This probably shouldn't be happening, but we'll try and delete whatever it's pointing at anyway.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
You based this on the 9.4 release instead of a release branch.
Can you rebase this?

gregorydlogan:
Whoops!  Done, thanks @lkiesow .

PR #2584, opened 2021-05-10
Add a Maven project for the Checkstyle rules to be consumed by external projects

JulianKniephoff:
[The Opencast Annotation Tool](/opencast/annotation-tool) would like to format its Java sources in a style consistent with Opencast. We currently have a redundant copy of the file in the repo which already triggers me slightly, but now that there is actually some movement regarding checkstyle thanks to @LukasKalbertodt it's probably outdated as well.

Now, the Maven checkstyle plugin can look up its configuration in the classpath, so you can load configuration files from Maven dependency artifacts. This PR creates such an artifact for our checkstyle configuration! :tada: 

Putting the POM directly in the `checkstyle` folder feels a bit icky since that folder already contains stuff not directly related to checkstyle (albeit related to style checking ...). However, didn't want to completely rename/restructure that directory only to have people not like the result, so I went down the slightly ugly but very simple route for now, looking for feedback on how the community would like to structure this going forward.

With that in mind I'm interested in any and all reviews, but I'm especially interested in what @gregorydlogan and @LukasKalbertodt have to say; @gregorydlogan because the entire goal here is to get this artifact deployed to Central eventually, and I'm not sure I did everything required for that, and @LukasKalbertodt because I don't want to disrupt the whole "restyling" effort.

### Your pull request should…

* [x] have a concise title
* [x] be against the correct branch
* [x] pass automated tests

### DIFF
--- a/docs/checkstyle/check-config.sh
+++ b/docs/checkstyle/check-config.sh
@@ -5,11 +5,11 @@ set -ue
 ret=0
 
 echo Checking pom.xml files for tabs or trailing spaces…
-if grep -rn $'\t' modules assemblies pom.xml --include=pom.xml; then
+if grep -rn $'\t' modules assemblies docs/checkstyle pom.xml --include=pom.xml; then
   echo "Tabs found!"
   ret=1
 fi
-if grep -rn ' $' modules assemblies pom.xml --include=pom.xml; then
+if grep -rn ' $' modules assemblies docs/checkstyle pom.xml --include=pom.xml; then
   echo "Trailing spaces found!"
   ret=1
 fi
--- a/docs/checkstyle/pom.xml
+++ b/docs/checkstyle/pom.xml
@@ -27,27 +27,34 @@
         </includes>
       </resource>
     </resources>
-
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-jar-plugin</artifactId>
-        <version>3.2.0</version>
-        <executions>
-          <!-- Generate an empty JavaDoc-jar to comply with Maven Central rules -->
-          <execution>
-            <id>empty-javadoc-jar</id>
-            <phase>package</phase>
-            <goals>
-              <goal>jar</goal>
-            </goals>
-            <configuration>
-              <classifier>javadoc</classifier>
-              <classesDirectory>${basedir}/javadoc</classesDirectory>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
   </build>
+
+  <!-- Generate an empty JavaDoc-jar to comply with Maven Central rules when we are releasing -->
+  <profiles>
+    <profile>
+      <id>release</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-jar-plugin</artifactId>
+            <version>3.2.0</version>
+            <executions>
+              <execution>
+                <id>empty-javadoc-jar</id>
+                <phase>package</phase>
+                <goals>
+                  <goal>jar</goal>
+                </goals>
+                <configuration>
+                  <classifier>javadoc</classifier>
+                  <classesDirectory>${basedir}/javadoc</classesDirectory>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
 </project>
--- a/docs/checkstyle/pom.xml
+++ b/docs/checkstyle/pom.xml
@@ -32,6 +32,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
+        <version>3.2.0</version>
         <executions>
           <!-- Generate an empty JavaDoc-jar to comply with Maven Central rules -->
           <execution>
--- a/docs/checkstyle/pom.xml
+++ b/docs/checkstyle/pom.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.opencastproject</groupId>
+  <artifactId>checkstyle</artifactId>
+  <version>9-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <name>Opencast Checkstyle Rules</name>
+  <parent>
+    <groupId>org.opencastproject</groupId>
+    <artifactId>base</artifactId>
+    <version>9-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+
+  <properties>
+    <checkstyle.skip>true</checkstyle.skip>
+  </properties>
+
+  <build>
+    <resources>
+      <resource>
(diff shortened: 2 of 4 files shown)


docs/checkstyle/pom.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.opencastproject</groupId>
+  <artifactId>checkstyle</artifactId>
+  <version>9-SNAPSHOT</version>
```
lkiesow, line 6: This looks like it could easily be forgotten when doing a version bump.
Why do we need to have code for external projects in here?
Shouldn't they live in their codebase?

JulianKniephoff: The Maven version plugin updates it.

The point is I want to use some of the "code" (rather config) from your repo in mine without duplicating it. For that it needs to be exposed somehow.

docs/checkstyle/check-config.sh
```diff
@@ -5,11 +5,11 @@ set -ue
 ret=0
 
 echo Checking pom.xml files for tabs or trailing spaces…
-if grep -rn $'\t' modules assemblies pom.xml --include=pom.xml; then
+if grep -rn $'\t' modules assemblies docs/checkstyle pom.xml--include=pom.xml; then
```
lkiesow:
```suggestion
if grep -rn $'\t' modules assemblies docs/checkstyle pom.xml --include=pom.xml; then
```

docs/checkstyle/pom.xml
lkiesow: Why not just copy the configuration? If it's an external project it can have its own rules and this would indirectly promise a stable API for checkstyle rules which get published to Maven Central while it would serve no purpose within this project.

It's more a -0 than a -1 from me but in general, I think that things which are not used by Opencast should not live within Opencast since it confuses people (wtf is this for). But let's see what others think about this.

gregorydlogan: I could see this being a major help if the longer term goal is to merge the OAT into the mainline codebase, rather than keeping it external.  It's also useful in the sense that if our existing code contributors need to work on it the code is already formatted the same.

That being said, I'm not sure this is the right way to do it.  @JulianKniephoff we're already publishing a base pom[1], could we shoehorn this checkstyle file into that?  I would expect that our 'base' object defines things which are basic to the entire ecosystem, which would include things like checkstyle rules.  I feel like having a separate artifact for this is kinda hacky.

1: https://repo1.maven.org/maven2/org/opencastproject/base/, also, damnit I forgot to publish 9.4.  Again.

JulianKniephoff: @lkiesow you bring up a few good points I hadn't considered, yet. So I'm actually more of a +0 as opposed to a +1 myself now. :sweat_smile: However, I still think this could be useful. External projects could (for whatever reason) use the Opencast style without having to duplicate the config. Granted, since these rules might just change out from under them, they probably want to pin the version they depend on rather than just always using the latest, and at that point you might say they could also just copy the file, sure. :shrug: 

I also don't think it hurts, though. I personally don't expect any stability from this, beyond the fact that from experience and git history I can tell that the corresponding files don't change that much. :wink: Which is actually more than I can say about the **other** Opencast packages/modules that external projects like the Annotation Tool depend on, which are also published to Central, and which we also don't give any stability guarantees for, as was stated rather explicitly in some technical meeting where I brought up the Annotation Tool breaking because of changes in `common`.

Now, whether or not that's a good thing is a completely different discussion, which also kind of ties in with @gregorydlogan's point: Does Opencast want to support external projects like the Annotation Tool? I personally think it should, because I don't think it scales to pull everything that might have dependencies to Opencast into the main source tree. For one thing, not every project wants to/can evolve at the same speed as Opencast or support as many concurrent versions. That's also why I personally would like to keep the Annotation Tool separate. :thinking: Although I'm also not dead set on that opinion; I also see some advantages of the AT being integrated more tightly, like piggybacking on the existing test/demo infra, or not having to deal with incompatible dependency versions while supporting multiple versions of Opencast.

Keeping the style of the project consistent with Opencast is useful in any case, though, I think, exactly for the reasons Greg mentioned. Like, the AT doesn't need to live in `opencast/opencast` for committers to contribute; everyone is welcome in our little corner here. :wink: 

As for how to package this: I'm personally a fan of small, targeted artifacts. Who knows what I will pull in if I depend on `org.opencastproject.base` only to get the checkstyle configuration. :thinking: Depends on what else is in it, too, of course. But do we have other things already? It would also seem kind of weird to me if there is **just** the checkstyle config in this generic package. :thinking: Although in the specific case of the Annotation Tool, I might actually want to make `org.opencastproject.base` my parent POM, now that I think about it. (Although I don't actually know whether that works at the moment.) And automatically getting the checkstyle config I want with that actually sounds good. x) (**EDIT**: Nevermind. While an interesting idea this would require some more thought and/or changes to make it work.)

Anyway, for this to work the `packaging` of `base` would need to change. (It's a `pom` now, but needs to be a `jar` to contain resources.) I don't know what implications this might have. :thinking:

docs/checkstyle/check-config.sh
JulianKniephoff: Whoops. :sweat_smile:

LukasKalbertodt:
> @LukasKalbertodt because I don't want to disrupt the whole "restyling" effort.

I don't think this PR would interfere with that.

---

Other than that, I can't say a lot about the code changes. From the discussion above, I'm also not sure if this is still worth it. I understood this PR as an effort to avoid duplicate configurations, but due to the "stability" problems mentioned, Julian already said:

> at that point you might say they could also just copy the file

So what are the advantages of this PR again, now that we concluded others would need to pin to a specific version anyway?

But yeah, I'm very neutral on this.

gregorydlogan:
The more I think about this the less sense it makes.

Our check styles rules are in a little bit of flux right now, but in general don't change much. If you're concerned with using the same rules as upstream maybe @LukasKalbertodt can gin up a goal checkstyle config (or are we already there?) for you to use right now, and if we end up changing a buncha rules in the future we can address exporting our rules then?

JulianKniephoff:
> So what are the advantages of this PR again, now that we concluded others would need to pin to a specific version anyway?

I mean, it's still easier to change a dependency version than to manually keep a file in sync. :shrug: But yeah, sure, none of the arguments on either side are very strong IMO. This is a pretty benign thing we are talking about here, and as far as I'm concerned we can close and/or postpone this particular discussion.

**However**, I do think there is a larger issue at play here as I alluded to above already which would render this whole discussion kind of moot if we agreed on it; either we would just do it as a matter of course (although we might still discuss timing and implementation details, of course), or this PR wouldn't exist in the first place.

And that issue is whether or not Opencast should do/have/expose certain things in support of external projects which aim to extend it. This does not only include the style guide, but more importantly things like dependency versions like @mtneug mentioned in the technical recently. Or "internal API" stability. (Or alternatively a dedicated new API for this purpose.)

Counting the casual mentions of "a plugin system" in the technical meetings makes me think that at least some of us agree with me that Opencast **should** do that. Maybe it's not on the level of Maven and OSGi bundles/Karaf features, but however it may look in the end Opencast will end up with some artifacts (in the general sense, not necessarily Maven artifacts) in its source tree that won't make sense if you look at Opencast in isolation.

Anyway, this got a bit out of hand, sorry. :sweat_smile: Maybe I'm blowing this out of proportion, but I would be interested in people's thoughts on this general matter in any case. :shrug: But maybe at this point we should move this to the dev list instead of further abusing this straw man PR? :thinking:

gregorydlogan:
Agree, let's move further discussion to the dev list.

PR #2590, opened 2021-05-10
New woh: conditional-config

rute-santos:
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
@@ -8,8 +8,12 @@ Description
 
 The conditional-config operation sets a workflow configuration variable based on conditions that are
 tested in sequence.
+
 If _condition-1_ is true, _value-1_ is assigned to the variable, if _condition-2_ is true, _value-2_ is
 assigned, and so on. If no conditions are true, the value specified in _no-match_ is assigned.
+
+Conditions are tested in alphabetical order and the first condition that is true sets the variable so,
+if two conditions are true, the value is set by the first one that gets evaluated.
 
 This operation is useful to reduce workflow complexity when the same operation is repeated many times and
 executed based on distinct if-conditions.
--- a/docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
@@ -1,0 +1,83 @@
+Conditional Config Workflow Operation
+=====================================
+
+ID: `conditional-config`
+
+Description
+-----------
+
+The conditional-config operation sets a workflow configuration variable based on conditions that are
+tested in sequence.
+If _condition-1_ is true, _value-1_ is assigned to the variable, if _condition-2_ is true, _value-2_ is
+assigned, and so on. If no conditions are true, the value specified in _no-match_ is assigned.
+
+This operation is useful to reduce workflow complexity when the same operation is repeated many times and
+executed based on distinct if-conditions.
+
+
+Parameter Table
+---------------
+
+|configuration key|description|example|
+|-----------------|-----------|-------|
+|configuration-name|Name of workflow configuration variable to be set|encoding-profile|
+|condition-**X**|Condition to be tested, same syntax as workflow conditions|(${presenter_work_resolution_x} &gt; 1600)|
+|value-**X**|Value to be assigned to configuration variable if condition-<X> is true|encoding-profile-x|
+|no-match|Value to be assigned to configuration variable if none of the conditions are true|encoding-profile-none|
+
+All _condition-**X**_ are sorted as strings and then evaluated in sequence e.g. condition-1, condition-2, etc.
+
+
+Example
+-------
+
+Set presenter encoding profile based on presenter/work media attributes:
+
+```xml
+<operation id="conditional-config" 
+           description="Evaluate media properties and set presenter-encoding-profiles configuration">
+	<configurations>
+		<configuration key="configuration-name">presenter-encoding-profiles</configuration>
+		<configuration key="condition-1">
+			(${presenter_work_framerate} == 25) AND (${presenter_work_resolution_x} &gt; 1600)
+			AND (${presenter_work_bitrate} &gt; 1999999)
+		</configuration>
+		<configuration key="value-1">
+			z-full-3m-presenter,z-threequarters-1500k-presenter,z-half-700k-presenter,
+			z-quarter-300k-presenter,z-lowbr-160k-presenter,multiencode-hls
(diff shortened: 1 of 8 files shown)


gregorydlogan:
Without looking at the code:
- Needs a note in the release notes
- _Could_ be moved to 9.x, although I don't actually care if we do.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConditionalConfigWorkflowOperationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+      String condition = operation.getConfiguration(key);
+      if (StringUtils.isEmpty(condition)) {
+        continue;
+      }
+      boolean result = false;
+      try {
+        result = WorkflowConditionInterpreter.interpret(condition.trim());
+        logger.debug("Evaluate condition: {}, result: {}", condition, result);
+      } catch (IllegalArgumentException e) {
+        logger.warn("Invalid condition to evaluate: {}.", condition);
```
mtneug: This should be an error.

```suggestion
        logger.error("Invalid condition to evaluate: {}.", condition);
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConditionalConfigWorkflowOperationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+        throw new WorkflowOperationException(String.format("Invalid condition: %s", condition.trim()), e);
+      }
+      // If condition true, set variable and return
+      if (result) {
+        String valuePropName = VALUE_PREFIX + key.substring(CONDITION_PREFIX.length());
+        value = operation.getConfiguration(valuePropName);
+        if (StringUtils.isEmpty(value)) {
+          logger.warn("Condition {}  evaluated to true, but no value informed to set {} variable in {}",
```
mtneug: This should be an error.
```suggestion
          logger.error("Condition {}  evaluated to true, but no value informed to set {} variable in {}",
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/ConditionalConfigWorkflowOperationHandler.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    // If we got here, no condition was evaluated true so we use the no-match configuration if there
+    String noMatch = operation.getConfiguration(NO_MATCH);
+    if (StringUtils.isNotEmpty(NO_MATCH)) {
+      properties.put(wfConfigName, noMatch);
+      logger.debug("Configuration key '{}' of workflow {} is set to value '{}'", wfConfigName, id, noMatch);
+      return createResult(workflowInstance.getMediaPackage(), properties, Action.CONTINUE, 0);
+    }
+    return createResult(Action.CONTINUE);
```
mtneug: I guess one could argue both ways: the operation evaluated a bunch of expressions, so it did something, but really nothing has changed after this operation. So maybe `SKIP` would be more appropriate? Regardless, I'm fine with either one.

modules/workflow-workflowoperation/src/test/java/org/opencastproject/workflow/handler/workflow/ConditionalConfigWorkflowOperationHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+    Assert.assertEquals("value no-match", result.getProperties().get(CONFIGURATION_VAR));
+  }
+
+  private WorkflowInstance replaceVars(WorkflowInstance wf, String key, String value) throws WorkflowParsingException {
+    HashMap<String, String> wfConfig = new HashMap<String, String>();
+    wfConfig.put(key, value);
+
+    // Replace operation configuration with variables form wf config
+    final Function<String, String> systemVariableGetter = k -> null;
+    String xml = WorkflowConditionInterpreter.replaceVariables(WorkflowParser.toXml(workflowInstance),
```
mtneug: `workflowInstance` is used directly.
```suggestion
    String xml = WorkflowConditionInterpreter.replaceVariables(WorkflowParser.toXml(wf),
```

docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
```diff
@@ -0,0 +1,83 @@
...
+---------------
+
+|configuration key|description|example|
+|-----------------|-----------|-------|
+|configuration-name|Name of workflow configuration variable to be set|encoding-profile|
+|condition-**X**|Condition to be tested, same syntax as workflow conditions|(${presenter_work_resolution_x} &gt; 1600)|
+|value-**X**|Value to be assigned to configuration variable if condition-<X> is true|encoding-profile-x|
+|no-match|Value to be assigned to configuration variable if none of the conditions are true|encoding-profile-none|
+
+All _condition-**X**_ are sorted as strings and then evaluated in sequence e.g. condition-1, condition-2, etc.
```
mtneug, line 32: Could you add a description of what happens if multiple conditions are `true`? Conditions are evaluated in alphabetical order, which might be surprising (`condition-10` is evaluated before `condition-2`).

mtneug (changes requested):
I still need to run a test workflow, but it looks good from what I see so far. Only some minor complaints.

docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
rute-santos: Will do. Yes, they are sorted alphabetically. The idea was that the part after the '-' could be anything, not only numbers, but I can change it to numbers if you think it is better. Please let me know.

modules/workflow-workflowoperation/src/test/java/org/opencastproject/workflow/handler/workflow/ConditionalConfigWorkflowOperationHandlerTest.java
rute-santos: Good catch!

docs/guides/admin/docs/workflowoperationhandlers/conditional-config-woh.md
mtneug: No, I would also prefer any string. Just wanted to document the behaviour ;).

mtneug (approved):

mtneug:
Works for me.

PR #2593, opened 2021-05-11
Use UTF-8 for config files

mtneug:
By default, config files are assumed to be encoded using ISO-8859-1 (latin1). This changes that to UTF-8.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/assemblies/resources/system.properties.append
+++ b/assemblies/resources/system.properties.append
@@ -7,3 +7,6 @@ org.terracotta.quartz.skipUpdateCheck=true
 javax.xml.parsers.DocumentBuilderFactory=org.apache.xerces.jaxp.DocumentBuilderFactoryImpl
 javax.xml.parsers.SAXParserFactory=org.apache.xerces.jaxp.SAXParserFactoryImpl
 javax.xml.transform.TransformerFactory=org.apache.xalan.processor.TransformerFactoryImpl
+
+# Decode files with UTF-8
+felix.fileinstall.configEncoding=UTF-8
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncodingProfileScanner.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncodingProfileScanner.java
@@ -41,6 +41,8 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Dictionary;
 import java.util.HashMap;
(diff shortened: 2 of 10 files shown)


lkiesow (commented):
- Do we actually need to include this in the fileinstall configurations?
- If we include it in the `custom.properties`, properties in there are still encoded `latin1` aren't they?
- Would it make sense to set this in the `config.properties`? I vaguely remember that one to be loaded before the other ones. Maybe that would cause them to be loaded as UTF-8 Already?

mtneug:
1. This setting is a fileinstall setting. So why not? I'm not sure if files other than property files are then loaded with UTF-8, but it shouldn't hurt.
2. Probably.
3. Probably. `config.properties` includes `custom.properties` and has fileinstall settings for the general configuration files. As I understand it, `config.properties` is generated automatically by Maven. I would have to take a look at how I would change that.

lkiesow:
Take a look at https://github.com/opencast/opencast/blob/develop/assemblies/resources/system.properties.append
Maybe we could even include it there.

> This setting is a fileinstall setting. So why not? I'm not sure if files other than property files are then loaded with UTF-8, but it shouldn't hurt.

More configuration isn't always better. Especially if we want everything to be UTF-8 anyway. If we can just set a global default instead of setting it  over and over again, I think that would be preferable.

mliradelc:
@mtneug Should I continue with the review?  Because I can merge it before the FF.

lkiesow:
@mliradelc, if you want to take it from here, feel free to unassign me and assign yourself. I don't mind doing one less review.

mliradelc:
> Take a look at https://github.com/opencast/opencast/blob/develop/assemblies/resources/system.properties.append
> Maybe we could even include it there.
> 
> > This setting is a fileinstall setting. So why not? I'm not sure if files other than property files are then loaded with UTF-8, but it shouldn't hurt.
> 
> More configuration isn't always better. Especially if we want everything to be UTF-8 anyway. If we can just set a global default instead of setting it over and over again, I think that would be preferable.

@mtneug What do you think about to make it global?

gregorydlogan:
@mtneug @mliradelc is this still a WIP, or are we still waiting on something?

mtneug:
Current status:

* all config files directly in `etc` except for `custom.properties` should now use UTF-8 as configured in `system.properties`
* files in `etc/acl`, `etc/branding`, `etc/security`, `etc/workflows`, `etc/statistics` should already be loaded in the correct encoding (probably UTF-8). These are mainly XML files that specify an encoding. `etc/statistics` contains JSON files that I assume are loaded in UTF-8.
* files in `etc/email`, `etc/encoding`, `etc/feeds`, `etc/listproviders` are now loaded in UTF-8. This required a code change since we load the property files ourselves.
* Property files read in by the `execute-once` and `execute-many` operations now also use UTF-8.

There are some other locations we load in property files, but I'm unsure about those. I leave them out for now.

lkiesow:
You mentioned in the technical meeting, that some code already converts the properties internally. How does this effect code like:

```
modules/dictionary-hunspell/src/main/java/org/opencastproject/dictionary/hunspell/DictionaryServiceImpl.java:98:      binary = new String(binary.getBytes("ISO-8859-1"), "UTF-8");
```

Does this still work?

mtneug:
> Does this still work?

No, and thanks for reminding me to test this again. I thought I tested it, but apparently not. The conversion is no longer needed and interpreting it as non-UTF-8 will lead to wrong characters.

PR #2596, opened 2021-05-11
New  woh: select-version

rute-santos:
The select-version woh replaces the current media package in the running workflow with another version from the asset manager.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/asset-manager-workflowoperation/pom.xml
```diff
@@ -39,10 +39,24 @@
       <artifactId>opencast-workflow-service-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.entwinemedia.common</groupId>
+      <artifactId>functional</artifactId>
+    </dependency>
```
rute-santos: This was missing before, not related to this change. I was getting an error when building.

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/SelectVersionWorkflowOperationHandler.java
```diff
@@ -0,0 +1,201 @@
...
+import org.opencastproject.mediapackage.MediaPackageElementFlavor;
+import org.opencastproject.mediapackage.selector.SimpleElementSelector;
+import org.opencastproject.workflow.api.AbstractWorkflowOperationHandler;
+import org.opencastproject.workflow.api.WorkflowInstance;
+import org.opencastproject.workflow.api.WorkflowOperationException;
+import org.opencastproject.workflow.api.WorkflowOperationInstance;
+import org.opencastproject.workflow.api.WorkflowOperationResult;
+import org.opencastproject.workflow.api.WorkflowOperationResult.Action;
+
+import com.entwinemedia.fn.data.Opt;
```
Arnei, line 41: If I recall correctly, we want to avoid depending on entwine libraries as much as possible, so we can maybe eventually get rid of them. So it would be nice if you could rewrite the WOH to require them.

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/SelectVersionWorkflowOperationHandler.java
```diff
@@ -0,0 +1,201 @@
...
+
+  /** The configuration options for this handler */
+  private static final SortedMap<String, String> CONFIG_OPTIONS;
+
+  private AssetManager assetManager;
+
+  static {
+    CONFIG_OPTIONS = new TreeMap<String, String>();
+    CONFIG_OPTIONS.put(OPT_VERSION,
+            "Select the specific version. If informed, takes precedence over the other options.");
```
Arnei: From the docs, I would've expected the WOH to fail if both of mutually exclusive way of selecting a version are present. To avoid unexpected behaviour that might be hard to debug, I would either change this to actually fail if mutually exclusive ways of selecting a version are specified, or instead mention the precedence order in the docs.

Arnei (changes requested):
Having an easy way to run workflows on previous snapshots seems like a great feature to me.

Did some basics tests in my local AllInOne installation, worked fine there so far. Code looks  clean and concise. I got some minor remarks down below .

A nice-to-have to go with this PR would be some documentation on how to figure out the snapshot version that one wants to get back to (i.e. without having to ssh into the storage and ls the contents of all snapshots of the mediapackage in question). For example, if I want to find out what the last working snapshot version of mediapackage was before I broke it.

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/SelectVersionWorkflowOperationHandler.java
rute-santos: I could not agree more :) The problem is that Opt, Stream, etc are returned from the asset manager api so, until that's changed, I don't see a way to not use them...

modules/asset-manager-workflowoperation/src/main/java/org/opencastproject/workflow/handler/assetmanager/SelectVersionWorkflowOperationHandler.java
rute-santos: Good point!

rute-santos:
Hi @Arnei ,
I think I have addressed all your comments. Please let me know if there are any additional issues.
Thanks so much for reviewing!
Rute

docs/guides/admin/docs/workflowoperationhandlers/select-version-woh.md
```diff
@@ -17,6 +17,29 @@ There are currently two ways of specifying which version to select, which are **
 - A version number
 - A combination of *source-flavors* and *no-tag* (see below).
 
+### How to find the desired version of the media package?
+
+Every time a workflow runs a _snapshot_ operation, a new version of the media package is saved to the Asset Manager, starting with version 0.
+
+There isn't currently an easy way to list all the versions of a media package so here are some tips:
+
+1. Open the event details, go to the _Workflows_ tab, and count all _snapshot_ operations that were run until you reach the one you want. To get to the operation list, for each workflow, choose _Details,_ and then _Operations_ > _Details_.
```
Arnei: Is this actually reliable? For example, if I upload an event via the Admin UI, "version 0" seems to be created before the first workflow operation has a chance to run, and only contains a couple XML files, not the media files. Maybe there are more cases where Opencast creates new mp versions that the user needs to know about in order to reliable use this tip?

rute-santos: Probably. I will remove this because it isn't 100% reliable. I will leave only the database part. Or please let me know if you have other suggestions.

Arnei: Okay, makes sense. Sadly I don't have any other suggestions.

Arnei (approved):

KatrinIhler (changes requested):
All in all this looks very good. Code is clean and well-documented and you provided a good test with it.
I just have one complaint: Instead of defining the OSGI dependencies in an XML file, you should use OSGI annotations. We decided a while ago that we would rather not have new code introduce OSGI XML files. Could you change that, please? :D

rute-santos:
Done. Hopefully I got that right :)

KatrinIhler (dismissed):

KatrinIhler:
@rute-santos Sorry for dropping the ball on this one - OSGI annotations look good now. However, this PR now has conflicts and one check is failing, though I'm not entirely sure why. Maybe the Maven Dependency plugin doesn't like that the old XML files are defined as Service Components are defined and the new one is missing? (Because there's no longer an XML file...)  :thinking: Maybe you can take another look at this before the Feature Freeze?

mliradelc:
Hi, looking the test that failed, is only a few  trailing spaces found by the test.
```
Checking pom.xml files for tabs or trailing spaces…
+ echo Checking pom.xml files for tabs or trailing spaces…
+ grep -rn '	' modules assemblies pom.xml --include=pom.xml
+ grep -rn ' $' modules assemblies pom.xml --include=pom.xml
modules/asset-manager-workflowoperation/pom.xml:25:    </dependency>    
+ echo 'Trailing spaces found!'
+ ret=1
+ echo Checking configuration files for tabs or trailing spaces…
+ grep -rn '	' etc
Trailing spaces found!
```

Besides that, also is the conflict that @KatrinIhler has found, with that two fixed I think is ready for merge 😄

rute-santos:
I think everything is good now. Thanks for the review and comments!

mliradelc:
All seems good now, @rute-santos thanks for your awesome work 😄 , Merged!

Issue #2561, opened 2021-05-03
Workflow variable defaults are broken


wsmirnow:
Opencast provide a way to set an default for specific variables in the workflow definition as described in [Docs](https://docs.opencast.org/r/9.x/admin/#configuration/workflow/#accept-user-input). Providing defaults should work with/for:

- default value should be applied for not existing variables
- the value of existing variables should be used instead of provided default value

For testing you can use the simple workflow, attached to the ticket. The cleanup operation is a good choice for testing this scenario imo.

```
<?xml version="1.0" encoding="UTF-8"?>
<definition xmlns="http://workflow.opencastproject.org">

  <id>test-workflow-variable-default</id>
  <title>Test Workflow variable default</title>
  <tags>
    <tag>archive</tag>
  </tags>
  <displayOrder>100</displayOrder>
  <description>
    This is a simple workflow to test workflow variable default value is applied.
  </description>
  <configuration_panel>
  </configuration_panel>
  <operations>

    <operation
      id="defaults"
      description="Applying default configuration values">
      <configurations>
        <configuration key="someVar">foo</configuration>
      </configurations>
    </operation>

    <!-- Test workflow defaults with an cleanup operation -->

    <operation
      id="cleanup"
      if="${someVar}"
      description="Test #1: should be executed">
      <configurations>
        <configuration key="delete-external">false</configuration>
        <configuration key="preserve-flavors">security/*</configuration>
      </configurations>
    </operation>

    <operation
      id="cleanup"
      if="${someVar:foo} == 'foo'"
      description="Test #2: should be executed">
      <configurations>
        <configuration key="delete-external">false</configuration>
        <configuration key="preserve-flavors">security/*</configuration>
      </configurations>
    </operation>

    <operation
      id="cleanup"
      if="${someVar:foo} == 'bar'"
      description="Test #3: should not be executed">
      <configurations>
        <configuration key="delete-external">false</configuration>
        <configuration key="preserve-flavors">security/*</configuration>
      </configurations>
    </operation>

    <operation
      id="cleanup"
      if="${someOtherVar}"
      description="Test #4: should not be executed">
      <configurations>
        <configuration key="delete-external">false</configuration>
        <configuration key="preserve-flavors">security/*</configuration>
      </configurations>
    </operation>

    <operation
      id="cleanup"
      if="${someOtherVar:true}"
      description="Test #5: should be executed">
      <configurations>
        <configuration key="delete-external">false</configuration>
[...]

PR #2614, opened 2021-05-18
Fix variable always resolving to the default value even when set

mtneug:
Workflow variables such as `${my-var:my-default}` would always be
resolved to `my-default` even when `my-var` is set. This is because
the variable lookup incorrectly uses `my-var:my-default` instead
of simply `my-var`.

Related #2561

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
Works. Merging.

PR #2620, opened 2021-05-19
Replacing remaining Freenode references with Matrix

gregorydlogan:
We either missed some, or never actually checked, but I've replaced the remaining links to Freenode with Matrix links.

### DIFF
--- a/docs/guides/index.html
+++ b/docs/guides/index.html
@@ -108,8 +108,8 @@
         </div>
         <div class="col-md-6">
           <h4>IRC channel (Chat)</h4>
-          <p>For occasional quick questions we have an IRC channel on <a
-            href="https://webchat.freenode.net/?channels=opencast">Freenode</a>: <em>#opencast</em>.</p>
+          <p>For occasional quick questions we have a Matrix <a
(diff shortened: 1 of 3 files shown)


docs/guides/index.html
```diff
@@ -108,8 +108,8 @@ <h4>Mailing lists</h4>
         </div>
         <div class="col-md-6">
           <h4>IRC channel (Chat)</h4>
-          <p>For occasional quick questions we have an IRC channel on <a
-            href="https://webchat.freenode.net/?channels=opencast">Freenode</a>: <em>#opencast</em>.</p>
+          <p>For occasional quick questions we have a Matrix <a
+            href="https://app.element.io/#/room/#opencast-community:matrix.org">channel</a>: <em>#opencast-community</em>.</p>
```
lkiesow, line 112: This is irrelevant for docs.opencast.org and will cause merge conflicts. The landing page is now generated from 
https://github.com/opencast/opencast/blob/develop/docs/guides/.infrastructure/index.html.j2

gregorydlogan: So, delete this whole page then?

lkiesow: I wouldn't touch it in 9.x but then it's also not really important and the tests have passed so… I'll fix the merge conflict.
Merging…

PR #2621, opened 2021-05-19
Adding Dependable ignores for solr and spring-security

gregorydlogan:
If you merge this, remember to close #2597 and #2598!

### Your pull request should…

* [ ] have a concise title
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/#development-process/#acceptance-criteria-for-patches-in-different-versions)
* [ ] pass automated tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -52,3 +52,11 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


lkiesow:
I know, I caused this (sorry) but I think this is not what we would want since this configuration would globally activate the regular dependency updates for all Java dependencies while most cannot be updated separately right now. That's why I would tend to drop this unless you have a good idea to avoid those problems.

gregorydlogan:
I figured that global dep updates were already being done with Dependabot, and that this would disable (a couple of) the problematic ones.  I don't have a better idea, so I'm fine dropping this.

PR #2648, opened 2021-05-26
Spelling Fixes

gregorydlogan:
This pull request fixes a few spelling issues.

- 817c658 should be non-controversial, since it's all internal changes
- 13844ac might not be able to go into 10.x?  I've fixed the spelling in the visible text, but this does not alter the translation keys. 
 It's unclear what the rules mean IMO.  Can we fix spelling without altering the keys in a release branch?
- c248b19 fixes spelling in both the visible text *and* translation keys.  I wouldn't be surprised if this isn't allowed in 10.x, but I think it makes sense in context.

### Your pull request should…

* [x] have a concise title
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/#development-process/#acceptance-criteria-for-patches-in-different-versions)
* [x] pass automated tests


lkiesow:
> 817c658 should be non-controversial, since it's all internal changes

Not entirely sure but I think this might actually modify the filters of the external API. Not that I think this would actually hurt anyone and I'm fine with merging this. We also didn't so far make a lot of noise about translations and I would leave it to the release managers to decide if we want to invalidate some translations :grin:

gregorydlogan:
I feel like this is worth doing in 10.x, since the translations are easy to fix later.  But I'm obviously not the reviewer :)

pascalseeland:
I don't think it will change the external API, as to my understanding, the external API maps the CANCELLED state to STOPPED (https://github.com/opencast/opencast/blob/2f783712e96264b9d23fc926b2a6f7a4e0a2a980/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/event/Event.java#L116). 
I would vote for fixing the translation and the keys. I would vote that wrong text is the same as wrong code.

pascalseeland (approved):

lkiesow:
Two devs approve → I approve :D

PR #2649, opened 2021-05-27
Update series index directly

KatrinIhler:
Ooof. Okay. So, what this does:

- Update admin ui & external api index directly by the series service without sending messages (see my other recently merged PRs for similar changes).
- But the series service still sends those messages for other updates - ripping all that out wasn't in scope, sorry.
- Only update the index once per series instead of the three times we did that before :D
- I annotated some dependencies since the SeriesService uses OSGI annotations and doesn't play well with non-annotated classes.
- In that context I ripped out the ServicePublisher so I could annotate the JpaAclDB directly. That wrapper code was only used for this one case, so I felt like it was unnecessary overhead anyway (sorry Greg!)
- I removed the applyAcl methods from the AclService since they were now causing a dependency loop between the AclService and the SeriesService. The applyAclToSeries() method only called the SeriesService anyway... I moved that behavior into the endpoints (admin UI/External API) where it was used. That means those two endpoints belong to the AclService are now also gone.
- I fixed some hardcoded dependency versions in the poms (because I made that mistake as well).
- The creator is now only set for new series and not overwritten every time somebody updates a series. What was up with that anyway?

Stay tuned for the last two services left to be done, the AssetManager and the SchedulerService! :D


modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -432,8 +471,12 @@ public void updateSeriesProperty(String seriesID, String propertyName, String pr
           throws SeriesException, NotFoundException, UnauthorizedException {
     try {
       persistence.updateSeriesProperty(seriesID, propertyName, propertyValue);
-      messageSender.sendObjectMessage(SeriesItem.SERIES_QUEUE, MessageSender.DestinationType.Queue,
-              SeriesItem.updateProperty(seriesID, propertyName, propertyValue));
+
+      // update ES indices
```
pascalseeland, line 475: Since it was not there in the original code, this looks fine. In case you know the reasons, I would prefer to add them here as a comment.

modules/transcription-service-workflowoperation/pom.xml
```diff
@@ -75,12 +80,6 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.opencastproject</groupId>
-      <artifactId>opencast-transcription-service-amberscript</artifactId>
-      <version>10-SNAPSHOT</version>
-      <scope>compile</scope>
-    </dependency>
```
pascalseeland: Since this has nothing to do with actual PR. I do see the value of fixing it, but I would prefer this file not to be part of this PR.

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -304,100 +288,6 @@ public Response deleteAcl(@PathParam("aclId") long aclId) throws NotFoundExcepti
...
-        }
-      } catch (AclServiceException e) {
-        logger.error("Error applying acl to series {}", seriesId);
-        return serverError();
-      }
-    }
-    // acl not found
-    return notFound();
-  }
-
```
pascalseeland: This should be discussed in the technical meeting, as this is changing the API. It is an internal API and 10.x is still not released, so I'm not sure this qualifies for a minor release patch.

modules/elasticsearch-api/pom.xml
```diff
@@ -3,14 +3,6 @@
...
   <artifactId>opencast-elasticsearch-api</artifactId>
   <name>Opencast :: elasticsearch-api</name>
-  <dependencies>
-    <dependency>
-      <groupId>org.opencastproject</groupId>
-      <artifactId>opencast-common</artifactId>
-      <version>10-SNAPSHOT</version>
-      <scope>compile</scope>
-    </dependency>
-  </dependencies>
```
pascalseeland: Please move to a different PR together with translation-workflow-service/pom.xml

pascalseeland (changes requested):
Overall I like the changes. The issues with the API need to be discussed in the dev meeting. Afterwards we can decide on the target branche.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
KatrinIhler: I'm not quite sure what you're referring to here. Maybe this comment belongs to another line?

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
KatrinIhler: This was discussed in the Technical Meeting. The Series ACL endpoint has to be removed for this PR and has a good replacement among the series service endpoints, so this was approved. The Episode ACL endpoint is used by the stud.iP integration however. Since I mostly removed it for consistency, I have decided to re-add it to make their lives easier, though we should think about a replacement in another spot for the future.

KatrinIhler:
@pascalseeland @CGreweling I have re-added the Episode ACL endpoint that the studip integration uses, though we should think about a replacement for the future, since having it in that spot doesn't make a whole lot of sense imo. Had to throw out another unused dependency to make this work.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
KatrinIhler: I guess this is related to the theme property check in the line below? That was done before, but when the messages were received, so I just moved it. That part of the code is a bit funky anyway...

KatrinIhler:
@pascalseeland Moved unrelated dependency declaration improvements to #2730 :)

KatrinIhler:
So I actually had to touch this again, because the re-added endpoint causes a loop with my new (not yet published) AssetManager changes between the AssetManager and the AclServiceFactory, which both need each other. I could've made the dependencies of the AclService dynamic, but instead I decided to move the functionality from the ACLService into the endpoint. This makes the ACLService a bit cleaner, and the endpoint actually already required the necessary dependencies, they just weren't used. 

All in all, this isn't perfect, but I think it's okay and I hope I'm done with it now....

gregorydlogan:
> All in all, this isn't perfect, but I think it's okay and I hope I'm done with it now....

You've been tagged as 'knows how the index works', until its gone you're going to keep having to work with it now :p

KatrinIhler:
> > All in all, this isn't perfect, but I think it's okay and I hope I'm done with it now....
> 
> You've been tagged as 'knows how the index works', until its gone you're going to keep having to work with it now :p

I just meant this PR, I'm not done with the index myself. ;) (And yeah, I know how this works...)

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -843,12 +844,37 @@ public Response applyAclToEvent(@PathParam("eventId") String eventId, @FormParam
 
       Source eventSource = getIndexService().getEventSource(optEvent.get());
       if (eventSource == Source.ARCHIVE) {
-        if (getAclService().applyAclToEpisode(eventId, accessControlList)) {
+        Opt<MediaPackage> mediaPackage = getAssetManager().getMediaPackage(eventId);
+        Option<AccessControlList> aclOpt = Option.option(accessControlList);
+        // the episode service is the source of authority for the retrieval of media packages
+        if (mediaPackage.isSome()) {
+          MediaPackage episodeSvcMp = mediaPackage.get();
```
CGreweling, line 851: what does eipisdodeSvcMP stands for? its better than epM or so, but a little longer would make it more readable.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -843,12 +844,37 @@ public Response applyAclToEvent(@PathParam("eventId") String eventId, @FormParam
...
+                MediaPackage mp = getAuthorizationService().setAcl(episodeSvcMp, AclScope.Episode, acl).getA();
+                getAssetManager().takeSnapshot(mp);
+              } catch (MediaPackageException e) {
+                logger.error("Error getting ACL from media package", e);
+              }
+            }
+
+            // if none EpisodeACLTransition#isDelete returns true so delete the episode ACL
+            @Override
+            public void enone() {
```
CGreweling, line 867: enone means what?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
```diff
@@ -941,7 +941,8 @@ public Response deleteSeriesTheme(@PathParam("seriesId") String seriesID) throws
           @RestResponse(responseCode = SC_OK, description = "The ACL has been successfully applied"),
           @RestResponse(responseCode = SC_BAD_REQUEST, description = "Unable to parse the given ACL"),
           @RestResponse(responseCode = SC_NOT_FOUND, description = "The series has not been found"),
-          @RestResponse(responseCode = SC_INTERNAL_SERVER_ERROR, description = "Internal error") })
+          @RestResponse(responseCode = SC_INTERNAL_SERVER_ERROR, description = "Internal error"),
+          @RestResponse(responseCode = SC_UNAUTHORIZED, description = "If the current user is not authorized to perform this action") })
```
CGreweling, line 945: Maybe you mean The current user is not authorized. Not a question if he is authorized?

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -335,69 +338,43 @@ public Response applyAclToEpisode(@PathParam("episodeId") String episodeId, @For
       return notFound();
     }
     try {
-      if (aclService.applyAclToEpisode(episodeId, Options.join(macl))) {
+      Option<AccessControlList> aclOpt = Options.join(macl).map(toAcl);
+      Opt<MediaPackage> mediaPackage = getAssetManager().getMediaPackage(episodeId);
+      // the episode service is the source of authority for the retrieval of media packages
+      if (mediaPackage.isSome()) {
+        MediaPackage episodeSvcMp = mediaPackage.get();
```
CGreweling, line 345: episodeSvcMp = ?

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -335,69 +338,43 @@ public Response applyAclToEpisode(@PathParam("episodeId") String episodeId, @For
...
+        aclOpt.fold(new Option.EMatch<AccessControlList>() {
+          // set the new episode ACL
+          @Override
+          public void esome(final AccessControlList acl) {
```
CGreweling, line 349: esome = episodeIsSome?

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
```diff
@@ -335,69 +338,43 @@ public Response applyAclToEpisode(@PathParam("episodeId") String episodeId, @For
...
+              MediaPackage mp = getAuthorizationService().setAcl(episodeSvcMp, AclScope.Episode, acl).getA();
+              getAssetManager().takeSnapshot(mp);
+            } catch (MediaPackageException e) {
+              logger.error("Error getting ACL from media package", e);
+            }
+          }
+
+          // if none EpisodeACLTransition#isDelete returns true so delete the episode ACL
+          @Override
+          public void enone() {
```
CGreweling, line 361: enone = episodeIsNone

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/util/AccessInformationUtil.java
```diff
@@ -107,8 +108,9 @@ public static JSONObject serializeManagedAcl(ManagedAcl managedAcl) {
    *           if the <code>acl</code> parameter is null
    */
   public static JSONObject serializePrivilegesByRole(AccessControlList acl) {
-    if (acl == null)
+    if (acl == null) {
       throw new IllegalArgumentException("The parameter trans must not be null");
```
CGreweling, line 112: where ist the parameter trans here? Is the comment correct here?

CGreweling (commented):
Just some small style comments from me. 
I will try to test this till next tuesday.

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/util/AccessInformationUtil.java
KatrinIhler: I just fixed a check style error here since this code was moved and thus wasn't excluded from check style anymore. I'm not responsible for the rest. :D

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
KatrinIhler: As with almost all of the stylistic comments, I didn't actually write this code, I just moved it into a different spot. I agree that it's kinda ugly, but fixing it is not in scope for me right now.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
KatrinIhler: See above. :)

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
KatrinIhler: See above. :)

modules/authorization-manager/src/main/java/org/opencastproject/authorization/xacml/manager/endpoint/AbstractAclServiceRestEndpoint.java
KatrinIhler: See above. :)

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
KatrinIhler: I think I just copied this from somewhere else, but yeah, the description could be better. Gonna fix this.

CGreweling:
I testes this. The good news it works if i recreate the complete index.
But, if i recreate only the series service i get: 

200 from the REST-Endpoint and this in the logs:

```
2021-06-25 11:58:56,651 | ERROR | (IndexEndpoint:136) - Repopulating the index failed
java.lang.IllegalArgumentException: No enum constant org.opencastproject.index.rebuild.IndexRebuildService.Service.series
	at java.lang.Enum.valueOf(Enum.java:240) ~[?:?]
	at org.opencastproject.index.rebuild.IndexRebuildService$Service.valueOf(IndexRebuildService.java:66) ~[!/:?]
	at org.opencastproject.index.rebuild.IndexRebuildService.rebuildIndex(IndexRebuildService.java:147) ~[!/:?]
	at org.opencastproject.adminui.endpoint.IndexEndpoint.lambda$recreateIndexFromService$1(IndexEndpoint.java:134) [!/:?]
	at org.opencastproject.security.util.SecurityContext.lambda$runInContext$0(SecurityContext.java:68) [!/:?]
	at org.opencastproject.security.util.SecurityContext.runInContext(SecurityContext.java:58) [!/:?]
	at org.opencastproject.security.util.SecurityContext.runInContext(SecurityContext.java:67) [!/:?]
	at org.opencastproject.adminui.endpoint.IndexEndpoint.lambda$recreateIndexFromService$2(IndexEndpoint.java:131) [!/:?]

```

The seperate recreate index does not work, mybe this works with your other pr`s according to the index rebuild are merged in the code @KatrinIhler ?

KatrinIhler:
@CGreweling Okay, this is odd. How did you test this? Did you just check out the PR or did you pull it into an existing code base?

CGreweling:
> @CGreweling Okay, this is odd. How did you test this? Did you just check out the PR or did you pull it into an existing code base?

@KatrinIhler  i checked out develop and pulled this branch in. As it is described here in the "command line instructions" in github.

KatrinIhler:
@CGreweling Did you try "Series"? I think the capital S might matter.

mliradelc:
@CGreweling Has you tried more over this?

KatrinIhler:

PR #2650, opened 2021-05-27
sort the options alphabeticaly, this makes the languages field etc more readable

kristofkeppens:
We had some complaints of our users that the language and license options field with the lti upload were difficult to use because there was no kind of sorting applied and the list seemed random. This applies a simple alphabetical sort to the options for the language field and the licenses field.

### DIFF
--- a/modules/lti/src/components/EditForm.tsx
+++ b/modules/lti/src/components/EditForm.tsx
@@ -66,7 +66,7 @@ function collectionToOptions(collection: EventMetadataCollection, translatable: 
         .map(([k, v]) => [parseMetadataCollectionKey(k).label, v])
         .map(([k, v]) => [translatable ? t(k) : k, v])
         .map(([k, v]) => ({ value: v, label: k}))
-        .sort((a,b) => a.label.localeCompare(b.label));
+        .sort((a,b) => a.label.localeCompare(b.label, i18next.language));
 }
 
 function MetadataFieldReadOnly(props: MetadataFieldProps) {
--- a/modules/lti/src/components/EditForm.tsx
+++ b/modules/lti/src/components/EditForm.tsx
@@ -49,10 +49,24 @@ function parseMetadataCollectionKey(s: string): MetadataCollectionKey {
 }
 
 function collectionToOptions(collection: EventMetadataCollection, translatable: boolean, t: i18next.TFunction): OptionType[] {
-    return collectionToPairs(collection)
+    const pairs = collectionToPairs(collection);
+    //Check if we get a JSON object back, and that this contains the order field
+    //Use the order field in this case to determine to order of the items in the dropdown
+    if(pairs[0][0].startsWith("{") === true) {
+        if(Object.prototype.hasOwnProperty.call(JSON.parse(pairs[0][0]), "order") === true) {
+            return pairs
+                .sort((a,b)  => JSON.parse(a[0]).order - JSON.parse(b[0]).order)
+                .map(([k, v]) => [parseMetadataCollectionKey(k).label, v])
+                .map(([k, v]) => [translatable ? t(k) : k, v])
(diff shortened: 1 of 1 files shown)


Arnei (approved):
Makes sense and works in my local installation.

wsmirnow:
This will make the order field configuration from the list provider useless. To change the order, please customize the `etc/listproviders/*.properties` e.g. [licenses.properties](https://github.com/opencast/opencast/blob/r/9.x/etc/listproviders/licenses.properties) or [languages.properties](https://github.com/opencast/opencast/blob/r/9.x/etc/listproviders/languages.properties)

gregorydlogan:
@kristofkeppens rather than declining this PR, do you want to update it to a doc change adding that sorting those list providers will resolve the issue?

kristofkeppens:
@gregorydlogan , @wsmirnow I have to admit that I didn't know about those options but I'm guessing that something else is going wrong then. The LTI tool is doing a request to .../lti-service-gui/new/metadata to retrieve the options and it's values. The list it retrieves there, for instance for the licenses, contains the order field but this list is not ordered. But apart from this, the order in the drowdown is different from the order in the json response as well. I can't seem to figure out how it's ordered in the form, since it doesn't seem logical, but is consistent.

If I understand correctly the solution would be to not order alphabetically, but order based on the order field ? Would this be acceptable?

wsmirnow:
> @gregorydlogan , @wsmirnow I have to admit that I didn't know about those options but I'm guessing that something else is going wrong then. The LTI tool is doing a request to .../lti-service-gui/new/metadata to retrieve the options and it's values. The list it retrieves there, for instance for the licenses, contains the order field but this list is not ordered. But apart from this, the order in the drowdown is different from the order in the json response as well. I can't seem to figure out how it's ordered in the form, since it doesn't seem logical, but is consistent.
> 
> If I understand correctly the solution would be to not order alphabetically, but order based on the order field ? Would this be acceptable?

Yes, if the order field exists we should make use of it. In case of not existing order field, we can sort it alphabetically imo.

modules/lti/src/components/EditForm.tsx
```diff
@@ -49,10 +49,24 @@ function parseMetadataCollectionKey(s: string): MetadataCollectionKey {
...
+                .map(([k, v]) => ({ value: v, label: k}));
+        }
+    }
+    //Otherwise just return an alphabetically sorted list
+    return pairs
         .map(([k, v]) => [parseMetadataCollectionKey(k).label, v])
         .map(([k, v]) => [translatable ? t(k) : k, v])
-        .map(([k, v]) => ({ value: v, label: k }));
+        .map(([k, v]) => ({ value: v, label: k}))
+        .sort((a,b) => a.label.localeCompare(b.label));
```
wsmirnow: Is there a way to pass the current locale as second argument to `localeCompare()`? With german locale on Chrome at least the sort order is not correct, see screenshot below (Dänisch is before Deutsch but should be other way).

image

wsmirnow (changes requested):
The `order` field for sorting works es expected. Alphabetical sorting order is not locale specific, but do the job. Please take a look at the comment.

modules/lti/src/components/EditForm.tsx
Arnei: I'm by no means a grammtics expert, but don't you treat Umlaute the same way as normal letters when sorting in german? So "Dänisch" is effectively "Danisch" and therefore should be before "Deutsch"?

kristofkeppens: It seems that the sorting is correct in german, but by coincidence, according to the documentation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare Dänisch should come before Deutch in german locale, but not in the swedish locale for instance. I'll update the PR to include the second parameter.

PR #2701, opened 2021-06-10
Fix checkstyle violations in 6 modules

LukasKalbertodt:
We are getting there! #2238 

- `annotation-impl`
- `common-jpa-impl`
- `event-comment`
- `fileupload`
- `message-broker-api`
- `message-broker-impl`

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/common-jpa-impl/src/test/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImplTest.java
+++ b/modules/common-jpa-impl/src/test/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImplTest.java
@@ -90,8 +90,8 @@ public class ServiceRegistrationJpaImplTest {
 
   @Test
   public void testQueryStatistics() throws Exception {
-    HostRegistrationJpaImpl host = new HostRegistrationJpaImpl("http://localhost:8081", "http://localhost:8081", "Admin", 1024L,
-            1, 1, true, false);
+    HostRegistrationJpaImpl host = new HostRegistrationJpaImpl(
+        "http://localhost:8081", "http://localhost:8081", "Admin", 1024L, 1, 1, true, false);
     ServiceRegistrationJpaImpl serviceReg = new ServiceRegistrationJpaImpl(host, "NOP", "/nop", false);
 
     Date now = new Date();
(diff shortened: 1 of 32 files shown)


modules/common-jpa-impl/src/test/java/org/opencastproject/serviceregistry/impl/jpa/ServiceRegistrationJpaImplTest.java
```diff
@@ -125,13 +125,13 @@ public void testQueryStatistics() throws Exception {
 
   }
 
- @Test
+  @Test
  public void testToString() throws Exception {
```
lkiesow:
```suggestion
  public void testToString() throws Exception {
```

modules/message-broker-impl/src/main/java/org/opencastproject/message/broker/endpoint/MessageBrokerServiceEndpoint.java
```diff
@@ -39,19 +39,19 @@
...
-  property = {
+    property = {
     "service.description=Message Broker REST Endpoint",
     "opencast.service.type=org.opencastproject.message.broker.endpoint",
     "opencast.service.path=/broker"
-  },
-  immediate = true,
+    },
+    immediate = true,
   // expose interface for MH REST publisher!
```
lkiesow:
```suggestion
```

modules/message-broker-impl/src/main/java/org/opencastproject/message/broker/impl/MessageReceiverImpl.java
```diff
@@ -46,12 +46,12 @@
  * A class to receive messages from a ActiveMQ Message Broker.
  */
 @Component(
-  property = {
+    property = {
     "service.description=Message Broker Receiver",
     "service.pid=org.opencastproject.message.broker.impl.MessageReceiverImpl"
```
lkiesow: Shouldn't this be indented one more level?
There are a few identical cases.

lkiesow (changes requested):

modules/message-broker-impl/src/main/java/org/opencastproject/message/broker/impl/MessageReceiverImpl.java
LukasKalbertodt: Absolutely. So I guess I really need to pay closer attention to that and not just let my script indent everything according to how checkstyle wants it.

Unfortunately, checkstyle is still buggy when it comes to indenting these annotations. And it often suggests absolutely wrong values. So these things are not caught automatically. But yeah, gonna pay more manual attention to that from now on. And the cases in this PR should be fixed.

LukasKalbertodt:
@lkiesow Thanks for the review. I think I fixed all your remarks.

PR #2723, opened 2021-06-13
Limit Ingest Filename Length

### LINKED ISSUE #2699: Filenames of videos, thumbnails, preview can be too long thus workflow fails
**Describe the bug**
The filenames of any process file can be too long, therefore tools like ffmpeg etc. fail with error and the whole workflow might fail.

`2021-06-09T15:53:12,055 | INFO  | (EncoderEngine:461) - /srv/opencast/workspace/mediapackage/b39f3e94-81c0-48ac-9a5f-abb07a208c2c/bf15e33c-0eec-4855-bd92-ca30514723e5/2021_06_09_13_36___WS13_1_Rare_species_may_act_as_problem_solvers_in_critical_situations_of_the_bacterial_co_occurence_network_in_cystic_fibrosis_airways___Marie_M*****_P***__presentation__OC_Studio__8c349465-3440-4563-b2ea-4d8952b6d4f9-preview-composite.mp4: File name too long`

Here so many informations will be concatted together, which the system itself would not need: date, title of the recording, name of the presenter type of the video (presenter/presentation), roomname plus uuids etc.
Seems to me, that these are for human readability only.
My opinion of naming and handling files, internally work with uuids only and use human-readable naming only for downloads and respect filesystem specific conventions (pull human-readable data from db & truncate some info e.g. title).

**To Reproduce**
1. Go to 'Add Event'
2. Enter Title of at least 150 characters approx.
3. Enter Presenter(s) of at least 20 characters approx.
4. Start Studio Upload Workflow
5. Check opencast.log as the webfrontend yields unspecific error details

**Environment Information**
 - OS: Debian Stretch
 - Browser any
 - Opencast version with minor release <=8.11

lkiesow:
This patch limits the filename length of ingested tracks to prevent
problems during processing or download due to file system or process
limitations.

This fixes #2699

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -219,6 +219,9 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
   /** The default is not to automatically skip attachments and catalogs from capture agent */
   public static final boolean DEFAULT_SKIP = false;
 
+  /** The maximum length of filenames ingested by Opencast */
+  public static final int FILENAME_LENGTH_MAX = 75;
+
   /** Managed Property key to allow Opencast series modification during ingest
    * Deprecated, the param potentially causes an update chain reaction for all
    * events associated to that series, for each ingest */
@@ -715,13 +718,18 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
       job = serviceRegistry.updateJob(job);
       String elementId = UUID.randomUUID().toString();
       logger.info("Start adding track {} from input stream on mediapackage {}", elementId, mediaPackage);
+      if (fileName.length() > FILENAME_LENGTH_MAX) {
+        final String extension = "." + FilenameUtils.getExtension(fileName);
+        final int length = Math.max(0, FILENAME_LENGTH_MAX - extension.length());
+        fileName = fileName.substring(0, length) + extension;
+      }
       URI newUrl = addContentToRepo(mediaPackage, elementId, fileName, in);
       MediaPackage mp = addContentToMediaPackage(mediaPackage, elementId, newUrl, MediaPackageElement.Type.Track,
               flavor);
       if (tags != null && tags.length > 0) {
         MediaPackageElement trackElement = mp.getTrack(elementId);
         for (String tag : tags) {
-          logger.info("Adding Tag: " + tag + " to Element: " + elementId);
+          logger.debug("Adding tag `{}` to element {}", tag, elementId);
           trackElement.addTag(tag);
         }
       }


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -715,13 +715,16 @@ public MediaPackage addTrack(InputStream in, String fileName, MediaPackageElemen
       job = serviceRegistry.updateJob(job);
       String elementId = UUID.randomUUID().toString();
       logger.info("Start adding track {} from input stream on mediapackage {}", elementId, mediaPackage);
+      if (fileName.length() > 75) {
+        fileName = fileName.substring(0, 75) + '.' + FilenameUtils.getExtension(fileName);
```
Rillke: 1. Consider marking altered file names as such in case the new name gets offensive.
2. Consider declaring 75 as a constant.
3. If the old file name must not be longer than 75 characters (checked in if condition), the resulting file name probably should also not be longer (currently it's 75 + 1 + file extension length)

```suggestion
        String fileExtension = FilenameUtils.getExtension(fileName);
        String separator = '_lop.';
        int baseNameLength = 75 - fileExtension.length() - separator.length();
        fileName = fileName.substring(0, baseNameLength) + separator + fileExtension;
```

lkiesow: We don't have to hit exactly 75 characters, do we? I'm in favor of keeping things less complicated. I also don't understand your `_lop.` and I'm not sure others will. What about just using the default setting of [`StringUtils.abbreviate(…`)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#abbreviate-java.lang.String-int-) instead?

Rillke: The current implementation will create file names like `Some_File.ex.ext` from `Some_File.ext`, assuming that `Some_File.ext` is 76 characters long.

`StringUtils.abbreviate(…)` looks like a good default but will produce many dots just before the file name extension (and some ill-configured snakeoil virus protection may have an issue with that). I have no problem causing false snakeoil alerts, though.

lkiesow: Catching half of the extension is actually a good point. That could look weird. I'll update the code.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -715,13 +715,18 @@ public MediaPackage addTrack(InputStream in, String fileName, MediaPackageElemen
       job = serviceRegistry.updateJob(job);
       String elementId = UUID.randomUUID().toString();
       logger.info("Start adding track {} from input stream on mediapackage {}", elementId, mediaPackage);
+      if (fileName.length() > 75) {
```
CGreweling: Maybe set this as a variable at the beginning of the class? final int MAX_FILENAME_LENGTH=75 ?

CGreweling (commented):
Nice! Will test this, and if possible put this in our production system.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
lkiesow: Done

CGreweling:
It works! But i do not see the approve button here any more. @lkiesow i think you have to do something at my review so i can proceed. Just for me to understand what to do next.
From my perspective, this can be merged.

Rillke:
@CGreweling 

Files changed -> Review Changes (green button top right) -> Approve

https://user-images.githubusercontent.com/2311611/122437805-59985200-cf9a-11eb-9174-e33dcff7f81b.mp4

PR #2732, opened 2021-06-14
One Asset Manager to rule them all

KatrinIhler:
This merges the different layers of the Asset Manager until there is only one interface and one implementation (complete with OSGI annotations). I did this change step by step to ensure that behavior remains the same, and in the end did some cleanup where I changed the order of methods to keep similar ones together, dissolved small private methods that were only used by one other method (because boy, does the AssetManager like small private methods) and removed methods that weren't used.

I also merged and refactored the related test classes to reduce code duplication. I kicked out the test suites since I think no one uses them and there were already so many files with such long names... but this might be controversial? Idk, I have no hard feelings there.

An improvement for the future would be to rip out the Entwine Functional library that is used excessively here, but unfortunately that would have taken too much time. Some methods might also better live elsewhere to reduce the size of the AssetManagerImpl class. The class structure to create the queries could also probably be simplified a lot.

Still, I think this is an improvement because this massively reduces the complexity of the AssetManager and opens the door to further optimizations, because I no longer have to look at 5 files to figure out what takeSnapshot() _actually_ does. It doesn't even make testing much harder, so that's also nice. In summary, I was surprised how easy this was, even though my brain was running on 100% the whole the time. The tests were actually harder than the implementation itself...

(If you don't like my refactoring of the Asset Manager, we can probably just skip the last three commits. Everything until then is mostly copy & paste and then straightening out the places where it went boom.)


modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestTasksEndpoint.java
```diff
@@ -126,56 +133,61 @@ public TestTasksEndpoint() throws Exception {
...
-          }
-        };
-      }
-
-      @Override
-      public Database getDb() {
-        return db;
-      }
-
+  org.opencastproject.assetmanager.api.AssetManager mkAssetManager(final Workspace workspace) throws Exception {
```
lkiesow: Why not import this and use
```suggestion
  AssetManager mkAssetManager(final Workspace workspace) throws Exception {
```

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerJobProducer.java
```diff
@@ -47,10 +49,10 @@
 import java.util.List;
 import java.util.function.Consumer;
 
-public class TieredStorageAssetManagerJobProducer extends AbstractJobProducer {
+public class AssetManagerJobProducer extends AbstractJobProducer {
```
rute-santos, line 52: Currently, the asset-manager-workflowoperation pom.xml has _opencast-asset-manager-impl_ as a dependency.
It would be great to define an AssetManagerJobProducer interface in asset-manager-api with:
public interface AssetManagerJobProducer {
  public Job moveByIdAndDate(final String mpId, final Date start, final Date end, final String targetStorage);
  public Job moveByIdAndVersion(final Version version, final String mpId, final String targetStorage);
  public Job moveById(final String mpId, final String targetStorage);
  public Job moveByDate(final Date start, final Date end, final String targetStorage);
}
and then implement an AssetManagerJobProducerImpl in asset-manager-impl so that the asset-manager-workflowoperation pom.xml would depend on the api and not the impl.

gregorydlogan: This does, kinda, make sense, since the WOHs depend on the specific implementation of `impl`.  That being said, I don't have any issue with this being done.

This could also be a chance to simplify the design a bit.  The original reason why there even *is* an `AMJPI` is because there was no way to extend two classes in a single subclass since Java doesn't do multiple inheritance.  I have not looked through the code here, but it's quite possible that the `AMJP` can be collapsed into the `AM` itself.

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestTasksEndpoint.java
KatrinIhler: I think this was simply moved over from another spot, but yeah, importing will look a lot nicer. Will fix this.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerJobProducer.java
KatrinIhler: I think what Greg is suggesting sounds like a good idea, but I don't have much more time to spend on this. But I bet there are a few low-hanging fruits to simplify left, so knock yourselves out! :D

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestTasksEndpoint.java
KatrinIhler: Done :)

mliradelc:
Tested with #2743 All good!

KatrinIhler:

PR #2734, opened 2021-06-14
MH-13617 Prepare AV fix for fast workflow: add textual warning to docs

### LINKED ISSUE #2733: Clarify that recordings archived with "fast" workflow cannot be used with other Opencast workflows
**Describe the bug**
I ingest a recording with the "fast" workflow. Once it archives, I attempt to run another Opencast workflow on it and it fails because there are no prepared tagged tracks.

**To Reproduce**
1. Ingest a recording using the "fast" testing workflow
2. Once it archives, run an edit workflow on the archived event
3. See that the edit workflow fails because of no source "prepare" tagged tracks

**Environment Information**
 - Opencast version's after commit https://github.com/opencast/opencast/pull/956

Reference comments in https://github.com/opencast/opencast/pull/1861 regarding making fast compatible with production workflows. Creating replacement pull containing textual warnings that fast archived recordings are not compatible with production workflows.

karendolan:
This pull clarifies to the user that recordings archived with the "fast" testing workflow are not compatible with other Opencast workflows. 

Closes #2733
Replacement pull for https://github.com/opencast/opencast/pull/1861

### DIFF
--- a/docs/guides/developer/docs/api/workflow-api.md
+++ b/docs/guides/developer/docs/api/workflow-api.md
@@ -5,6 +5,8 @@ The Workflow API is available since API version 1.1.0.
 
 # Workflow instances
 
+**Warning:** these examples reference the `fast` testing workflow. Recordings archived from the `fast` testing workflow can not be used with other Opencast workflows.
+
 ### GET /api/workflows
 
 Returns a list of workflow instances.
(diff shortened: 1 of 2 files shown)


lkiesow:
My argument wasn't that we shouldn't make it compatible, but that we shouldn't make it complicated to make it compatible ;-)
But this definitely works for now, so let's get this emrged.
Maybe I can propose a PR based on my suggested changes later.

PR #2736, opened 2021-06-14
Fix processing of `fast` events

lkiesow:
This patch ensures that users can process events ingested using the
`fast` workflow using either “Start Task” → “Publish” or selecting
“Publish” in the video editor.

This is a continuation of the idea of pull request #1861 with a
different approach being used:

- This adds no unnecessary additional steps and no additional processing
  to the fast testing workflow
- This stores no duplicates of any media
- This only copies the tracks if necessary
- This works with already processed data (fixes the problem for all
  events stored in the asset manager)

It also fixes a problem of #1861 which would fail in the editor due to
no encoding quality being selected.

---

@karendolan, @gregorydlogan, this essentially implements the idea I outlines before in my comment. What do you think about this approach? Would this work for you as well?

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/workflows/publish-after-cutting.xml
+++ b/etc/workflows/publish-after-cutting.xml
@@ -25,6 +25,7 @@
         <configuration key="publishToYouTube">false</configuration>
         <configuration key="thumbnailType">0</configuration>
         <configuration key="thumbnailPosition">1</configuration>
+        <configuration key="flagQuality720p">true</configuration>
       </configurations>
     </operation>
 
--- a/docs/guides/developer/docs/api/workflow-api.md
+++ b/docs/guides/developer/docs/api/workflow-api.md
@@ -5,8 +5,6 @@ The Workflow API is available since API version 1.1.0.
 
 # Workflow instances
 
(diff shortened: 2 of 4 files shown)


etc/workflows/publish-after-cutting.xml
```diff
@@ -25,6 +25,7 @@
         <configuration key="publishToYouTube">false</configuration>
         <configuration key="thumbnailType">0</configuration>
         <configuration key="thumbnailPosition">1</configuration>
+        <configuration key="flagQuality720p">true</configuration>
```
karendolan, line 28: What is the flag quality for?

lkiesow: The problem with any event processed by a workflow like fast is that the configuration is different from “Schedule and Upload”. Using “Start Task”, that's no problem since you have a new configuration panel and will set the options again. But that's not true for the editor which relies on previously set options. This means that when starting “Publish” from the editor, no quality-flags are set and no encoding will be done. That ultimately results in no processed media and a failure during publication. That would also be a problem with your previous approach.

This flag sets the same default set in “Schedule and Upload”, meaning that at least one quality will be encoded if nothing else is set. If something else had been set before (e.g. because “Schedule and Upload” had been used), this will have no effect.

PR #2742, opened 2021-06-17
AssetManager updates Elasticsearch indices directly

KatrinIhler:
With this PR, the AssetManager now directly updates the Elasticsearch indices for the Admin UI and the External API instead of sending messages for that. It still sends these messages for other purposes, however.

This change was relatively straightforward, the only hurdle was the ACL updating in the database which had to be moved and which only worked prior because the messages are not immediately processed (timing issues, yeah!)

This is based on and includes #2732, which I only did because it makes this change easier. Additionally, I included #2649 because it ensures those changes play well together and because this PR relies on refactoring done for the series changes. So if you want to review this, take a look at those two first or focus on the last five commits of this PR.


KatrinIhler:

PR #2763, opened 2021-06-21
Hello World Configuration Example

lkiesow:
This patch adds an example about how to properly (re-)load configuration
in a service to the hello world module to give new developers a starting
point.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/hello-world-impl/src/main/java/org/opencastproject/helloworld/impl/HelloWorldServiceImpl.java
+++ b/modules/hello-world-impl/src/main/java/org/opencastproject/helloworld/impl/HelloWorldServiceImpl.java
@@ -23,9 +23,14 @@ package org.opencastproject.helloworld.impl;
 
 import org.opencastproject.helloworld.api.HelloWorldService;
 
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Modified;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Per discussion in the technical meeting, it would be nice to have example OSGi DI bits so that when people copy and paste this they have something to work from.  I'm fine if we want to merge this without that, but it would be a nice to have :)

lkiesow:
I'll do that eventually, but not add it to this pull request since it's not really related.

modules/hello-world-impl/src/main/java/org/opencastproject/helloworld/impl/HelloWorldServiceImpl.java
```diff
@@ -42,9 +47,28 @@
   /** The module specific logger */
   private static final Logger logger = LoggerFactory.getLogger(HelloWorldServiceImpl.class);
 
+  /** Message to print */
+  private String message = "Hello World";
```
KatrinIhler, line 51: That default here is unnecessary, right? Since we already use one in activate(). Or does the compiler complain if we leave this out?

```suggestion
  private String message;
```

gregorydlogan: The tests fail without this, if nothing else.

PR #2785, opened 2021-06-30
Added Arne Wilken as a developer to the main pom file

Arnei:
As tasked by Greg, this PR adds myself as developer to the main pom file

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1693,6 +1693,21 @@
         <github.nick>mliradelc</github.nick>
       </properties>
     </developer>
+    <developer>
+      <id>awilken</id>
+      <name>Arne Wilken</name>
+      <email>wilken@elan-ev.de</email>
+      <organization>ELAN e.V.</organization>
+      <organizationUrl>https://elan-ev.de/</organizationUrl>
+      <roles>
+        <role>developer</role>
+      </roles>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1693,6 +1693,21 @@
...
+      <name>Arne Wilken</name>
+      <email>wilken@elan-ev.de</email>
+      <organization>ELAN e.V.</organization>
+      <organizationUrl>https://elan-ev.de/</organizationUrl>
+      <roles>
+        <role>developer</role>
+      </roles>
+      <timezone>Europe/Berlin</timezone>
+      <properties>
+        <irc.nick>Arne Wilken</irc.nick>
```
lkiesow: I don't think this is your IRC nick, is it?
If I remember correctly, Freenode (and maybe even IRC) does not allow spaces.

Arnei: ...Come to think of it, should I rather be putting my Matrix id there (e.g. `@arnewilken:chat.virtuos.uni-osnabrueck.de`)? Seeing as I don't really use IRC, actually.

lkiesow: Doesn't make sense as `<irc.nick>` specifically  mentions IRC.
I don't care if you want to add something like `<matrix.user>` as this isn't used anywhere anyway.

PR #2793, opened 2021-07-03
More HTML validation for AdminUI

### LINKED ISSUE #2792: Series access policy rules not disabled for viewers
**Describe the bug**
When a user with view-only permissions on a series or view-only permissions on series ACLs views a series access policy, they can still tick check boxes and get a message "The access rules have been saved".

**To Reproduce**
1. Create a user with VIEW ROLEs on all admin UI actions, notably `ROLE_ADMIN_UI ROLE_UI_SERIES_DETAILS_ACL_VIEW ROLE_UI_SERIES_DETAILS_METADATA_VIEW ROLE_UI_SERIES_DETAILS_VIEW ROLE_UI_SERIES_VIEW ROLE_UI_EVENTS_VIEW`
2. Create a new series, grant user created in step 1 view permissions in the access policy step
3. Log in as the new user, switch to "series", open the details of the created series and switch to "Access policy"
4. Tick the "write" checkbox for the role of your user and get the message "The access rules have been saved" in spite of a 400 error returned by the server

Screenshot from 2021-07-03 11-55-26

**Environment Information**
 - OS: n/a
 - Browser: Firefox
 - Opencast version with minor release 10.x (develop.opencast.org), 8.x (legacy.opencast.org)

N.b. trying to reproduce this on stable.opencast (9.x) caused the system to only report 500 errors after step 4 for events and series in admin UI. Sorry for breaking this system :-(

Rillke:
- Turn on several more checks on HTML validator.
  - More checks help catch more errors when doing changes
    to the Admin UI in future.
- Facilitates organizations to create accessibility reports
  by showing which accessibility rules are enabled and which
  are yet to be done.
  Accessibility and its declaration is a requirement in the EU
  for most public entities (and upcoming for e-commerce):
  https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A32016L2102
- Fix invalid HTML.
  - Duplicate attributes removed or merged.
  - Elements which can't have nested content.
  - Remove wrong closing tags, add correct ones.
- Remove deprecated cellpadding/-spacing attributes
- May cause the statistics screen look a little different than before
  (stakeholders actively using the statistics are endorsed to re-target
  CSS rules accordingly)

Resolves #2792

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
Hi @Rillke,
we don't have to fuss about this too much this time and can probably just get it in like it is, for the future, it would be really nice to split something like this though:

- This resolves #2792, but for a reviewer it is really hard to detect which part of the changes actually fix the problem and you have to dig around quite a bit to identify this.
- You target the legacy branch of Opencast. Doing so with large patches is usually a bad idea, especially if they don't come with actual benefits to counter the disadvantages:
  - It's much more likely to have merge conflicts between old and new code if someone also has changes something in the newer versions. This causes pain for developers and adds another potential source for errors.
  - Legacy usually gets less testing, making it easier to miss an error in something like this, which then has a greater impact on user's systems

This means, it would have been nice to have a pull request against `r/9.x`, fixing #2792 and one or possibly two pull requests against `develop` not actually doing anything but introducing some linting to avoid future problems.

Rillke:
Hi @lkiesow your comment and objection is very helpful and much appreciated as well as your review.

For the record, this is (which addresses both issues, invalid HTML and impaired functionality):
https://github.com/opencast/opencast/pull/2793/files#diff-80e16f6cbbc5df7f2dd3acc6606d49f934bee07d5ff75d617252f3b57a02525eR221

One comment though:

> Legacy usually gets less testing, making it easier to miss an error in something like this, which then has a greater impact on user's systems

In general, also legacy branches benefit from increased test coverage, however I see the point with merge conflicts and it's not worth committer's time.

Issue #2557, opened 2021-05-03
LTI-Upload-Tool form validation 


Arnei:
If the user makes a mistake in filling out the required information for the upload in the LTI-Upload-Tool (e.g. they forgot to add a title), they will be notified with a generic error message only after the upload is completed. It would be nice if the users could be notified about their mistake much earlier, at least before the lengthy upload. A more specific error message would also be nice.

Environment Information
 - OS:  CentOS 7
 - Browser: Chrome
 - Opencast version: Develop

PR #2804, opened 2021-07-06
Add dfxp as subtitle format to LTI upload

kristofkeppens:
This adds the possibility to upload dfxp subtitles via the LTI upload module.
When using webvtt you will be prompted to set the language of the subtitles, when using dfxp the language will be parsed from the subtitle file itself, so there is no need to manually set this.

### DIFF
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -277,8 +277,6 @@ public class LtiServiceImpl implements LtiService, ManagedService {
         captionsMediaPackage.setURI(captionsUri);
       }
 
-      final MetadataList metadataList = new MetadataList();
-      final MediaPackageElementFlavor flavor = new MediaPackageElementFlavor("dublincore", "episode");
       final EventCatalogUIAdapter adapter = getEventCatalogUIAdapter();
 
       final DublinCoreMetadataCollection collection = adapter.getRawFields();
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -314,11 +314,11 @@ public class LtiServiceImpl implements LtiService, ManagedService {
       eventHttpRequest.setAcl(accessControlList);
 
       //Set workflow json object
-      JsonObject workflowConfig = new JSONObject();
+      JSONObject workflowConfig = new JSONObject();
 
-      workflowConfig.put('workflow', workflow);
-      workflowConfig.put('configuration', workflowConfiguration);
-
+      workflowConfig.put("workflow", workflow);
+      workflowConfig.put("configuration", (JSONObject) new JSONParser()
+          .parse(workflowConfiguration));
       eventHttpRequest.setProcessing(workflowConfig);
       eventHttpRequest.setMetadataList(metadataList);
       metadataList.add(adapter, collection);
@@ -326,7 +326,7 @@ public class LtiServiceImpl implements LtiService, ManagedService {
       JSONObject source = new JSONObject();
       source.put("type", "UPLOAD");
       eventHttpRequest.setSource(source);
-      indexService.createEvent(r);
+      indexService.createEvent(eventHttpRequest);
     } catch (SchedulerException e) {
       if (e.getCause() instanceof NotFoundException || e.getCause() instanceof IllegalArgumentException) {
         throw new RuntimeException("unable to create event", e.getCause());
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -235,6 +235,8 @@ public class LtiServiceImpl implements LtiService, ManagedService {
   public void upsertEvent(
           final LtiFileUpload file,
           final String captions,
+          final String captionsFormat,
+          final String captionsLanguage,
           final String eventId,
           final String seriesId,
           final String metadataJson) throws UnauthorizedException, NotFoundException {
@@ -251,19 +253,23 @@ public class LtiServiceImpl implements LtiService, ManagedService {
         throw new RuntimeException("Unable to create media package for event");
       }
       if (captions != null) {
-        final MediaPackageElementFlavor captionsFlavor = new MediaPackageElementFlavor("captions", "vtt+en");
+        final MediaPackageElementFlavor captionsFlavor = new MediaPackageElementFlavor(captionsFormat + "+" + captionsLanguage, "captions");
         final MediaPackageElementBuilder elementBuilder
             = MediaPackageElementBuilderFactory.newInstance().newElementBuilder();
         final MediaPackageElement captionsMpe = elementBuilder
                 .newElement(MediaPackageElement.Type.Attachment, captionsFlavor);
-        captionsMpe.setMimeType(mimeType("text", "vtt"));
-        captionsMpe.addTag("lang:en");
+        if ("dfxp".equals(captionsFormat)) {
+          captionsMpe.setMimeType(mimeType("application", "xml"));
+        } else {
+          captionsMpe.setMimeType(mimeType("text", captionsFormat));
+        }
+        captionsMpe.addTag("lang:" + captionsLanguage);
         mp.add(captionsMpe);
         final URI captionsUri = workspace
                 .put(
                         mp.getIdentifier().toString(),
                         captionsMpe.getIdentifier(),
-                        "captions.vtt",
+                        "captions." + captionsFormat,
                         new ByteArrayInputStream(captions.getBytes(StandardCharsets.UTF_8)));
         captionsMpe.setURI(captionsUri);
       }
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -250,9 +250,9 @@ public class LtiServiceImpl implements LtiService, ManagedService {
       throw new RuntimeException("No workflow configured, cannot upload");
     }
     try {
-      final EventHttpServletRequest r = new EventHttpServletRequest();
-      final MediaPackage mp = ingestService.createMediaPackage();
-      if (mp == null) {
+      final EventHttpServletRequest eventHttpRequest = new EventHttpServletRequest();
+      final MediaPackage mediaPackage = ingestService.createMediaPackage();
+      if (mediaPackage == null) {
         throw new RuntimeException("Unable to create media package for event");
       }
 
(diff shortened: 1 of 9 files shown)


modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
+      r.setMediaPackage(ingestService.addTrack(file.getStream(), file.getSourceName(),
+          MediaPackageElements.PRESENTER_SOURCE, mp));
+
+      JSONArray metadataJsonArray = (JSONArray) new JSONParser().parse(metadataJson);
+
       JSONArray collectionJsonArray = MetadataJson.extractSingleCollectionfromListJson(metadataJsonArray);
       MetadataJson.fillCollectionFromJson(collection, collectionJsonArray);
 
       replaceField(collection, "isPartOf", seriesId);
+      replaceField(collection, "duration", "6000");
```
Arnei: What is setting the metadata duration supposed to achieve? It does not seem to have any effect in the final mediapackage.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
       AccessControlList accessControlList = null;
 
-      // If series is set and it's ACL is not empty, use series' ACL as default
-      if (StringUtils.isNotBlank(seriesId)) {
-        accessControlList = seriesService.getSeriesAccessControl(seriesId);
-      }
-
-      if (accessControlList == null || accessControlList.getEntries().isEmpty()) {
-        accessControlList = new AccessControlList(
+      accessControlList = new AccessControlList(
```
Arnei: Why are we hardcoding the ACLs here? Is this sure to not break any user workflows?

modules/lti/src/components/EditForm.tsx
```diff
@@ -161,11 +168,24 @@ class TranslatedEditForm extends React.Component<EditFormProps> {
...
+                <>
+                    <div className="form-group">
+                        <label htmlFor="caption">{this.props.t("LTI.CAPTION")}</label>
+                        <input type="file" className="form-control-file" onChange={this.onChangeCaptionFile.bind(this)} />
+                        <small className="form-text text-muted">{this.props.t("LTI.CAPTION_DESCRIPTION")}</small>
+                    </div>
+                    { this.props.captionFormat === "vtt" && languageOptions.length > 0 &&
+                        <div className="form-group">
+                            <label htmlFor="caption_language">{this.props.t("LTI.CAPTION")} {this.props.t("LTI.LANGUAGE")}</label>
+                            <Select
```
Arnei, line 180: This allows me to select no option when uploading, resulting in the language "null". Does not seem to break anything, but it would be nice to force them to set a language to avoid confusion later on (which would arguably be part of adding form validation  #2557).

modules/lti/src/components/EditForm.tsx
```diff
@@ -161,11 +168,24 @@ class TranslatedEditForm extends React.Component<EditFormProps> {
...
             {this.props.withUpload === true &&
-                <div className="form-group">
-                    <label htmlFor="caption">{this.props.t("LTI.CAPTION")}</label>
-                    <input type="file" className="form-control-file" onChange={this.onChangeCaptionFile.bind(this)} />
-                    <small className="form-text text-muted">{this.props.t("LTI.CAPTION_DESCRIPTION")}</small>
-                </div>
+                <>
+                    <div className="form-group">
+                        <label htmlFor="caption">{this.props.t("LTI.CAPTION")}</label>
+                        <input type="file" className="form-control-file" onChange={this.onChangeCaptionFile.bind(this)} />
```
Arnei, line 174: Does not check for file type, allowing to upload basically anything. Having a simple check here to spot missclicks early would be nice (which would arguably be part of adding form validation #2557)

Arnei (commented):
There are a couple of changes here that don't seem directly related to the topic of the PR, it would be nicer to have those as a seperate PR instead. I'm mainly talking about the introduction and subsequent use of EventHttpServletRequest here, please correct me if I'm wrong.

More comments/questions down below.

Also at some point the we probably want to restyle the Upload UI, the file upload section looks really cramped.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
gregorydlogan: Agree, this is only appropriate if there is no series, per the previous code.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
+      r.setMetadataList(metadataList);
+      metadataList.add(adapter, collection);
 
-      final Map<String, String> configuration = gson.fromJson(workflowConfiguration, Map.class);
-      configuration.put("workflowDefinitionId", workflow);
-      ingestService.ingest(mp, workflow, configuration);
+      JSONObject source = new JSONObject();
+      source.put("type", "UPLOAD");
+      r.setSource(source);
+      indexService.createEvent(r);
```
gregorydlogan: This creates the event in the index service, but I would be surprised if it correctly ingested the underlying dxfp.  If it *does* ingest it then (IMO) this is a bug since I would __never__ expect the index service to do that.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
-
-      final EventCatalogUIAdapter adapter = getEventCatalogUIAdapter();
+      r.setMediaPackage(mp);
+      final MetadataList metadataList = new MetadataList();
+      final MediaPackageElementFlavor flavor = new MediaPackageElementFlavor("dublincore", "episode");
+      final EventCatalogUIAdapter adapter = catalogUIAdapters.stream()
+          .filter(e -> e.getFlavor().equals(flavor)).findAny().orElse(null);
+      if (adapter == null) {
+        throw new RuntimeException("no adapter found");
+      }
```
lkiesow: Not entirely sure about the syntax but there should be something like:

```suggestion
          .filter(e -> e.getFlavor().equals(flavor))
          .findAny()
          .orElseThrow(() -> new RuntimeException("no adapter found"));
```

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
       throw new RuntimeException("No workflow configured, cannot upload");
     }
     try {
-      MediaPackage mp = ingestService.createMediaPackage();
+      final EventHttpServletRequest r = new EventHttpServletRequest();
```
lkiesow: Please help those poor people who have to read a lot of code:
```suggestion
      final EventHttpServletRequest eventHttpRequest = new EventHttpServletRequest();
```

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
lkiesow: The index service can use the ingest service in the background. This means, this might actually work. But it also ties together several parts of Opencast to achieve its goal in a roundabout way. If you talk to the ingest service anyway: Let's use it directly. Less complicated and less likely to break because someone changes something in a completely unrelated part of Opencast.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
lkiesow: To be honest, this whole block looks like a security issue. The logic is: If no series is applied, make the episode readable __and__ writable for all users authenticated via OAuth? And, yes, the new logic makes it worse by removing that series check.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
-      final Map<String, String> configuration = gson.fromJson(workflowConfiguration, Map.class);
-      configuration.put("workflowDefinitionId", workflow);
-      ingestService.ingest(mp, workflow, configuration);
+      JSONObject source = new JSONObject();
+      source.put("type", "UPLOAD");
+      r.setSource(source);
+      indexService.createEvent(r);
+    } catch (SchedulerException e) {
+      if (e.getCause() != null && e.getCause() instanceof NotFoundException
+              || e.getCause() instanceof IllegalArgumentException) {
```
lkiesow: `null instanceof ClassX` is always `false` anyway
```suggestion
      if (e.getCause() instanceof NotFoundException || e.getCause() instanceof IllegalArgumentException) {
```

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,60 +250,84 @@ public void upsertEvent(
...
           new AccessControlEntry("ROLE_OAUTH_USER", "write", true),
           new AccessControlEntry("ROLE_OAUTH_USER", "read", true));
-      }
 
-      this.authorizationService.setAcl(mp, AclScope.Episode, accessControlList);
-      mp = ingestService.addTrack(file.getStream(), file.getSourceName(), MediaPackageElements.PRESENTER_SOURCE, mp);
+      r.setAcl(accessControlList);
+      r.setProcessing(
+              (JSONObject) new JSONParser().parse(
+                      String.format("{\"workflow\":\"%s\",\"configuration\":%s}", workflow, workflowConfiguration)));
```
lkiesow: Hand-crafting a JSON object can easily go very wrong. Can we use GSON or JSONSimple to build this?

modules/lti/src/components/EditForm.tsx
kristofkeppens: Indeed, I was planning to do the form validation for all form items in a separate PR.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
kristofkeppens: I can confirm that this does work, but you are right that we should use the most direct way to achieve this. We'll look into it.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
kristofkeppens: Not sure why this check on series was removed, if this is giving issues we need to find a different way to resolve this cause this is indeed a security issue! What would be the default ACL in opencast when no series is applied, read/write for admin and read for user I suppose ?

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
kristofkeppens: removed, no idea why this was there tbh

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
kristofkeppens: Reverted this change and using the ingestService now.

modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
```diff
@@ -246,38 +248,48 @@ public void upsertEvent(
...
+                        "captions." + captionFormat,
                         new ByteArrayInputStream(captions.getBytes(StandardCharsets.UTF_8)));
-        captionsMpe.setURI(captionsUri);
+        captionsMediaPackage.setURI(captionsUri);
       }
 
-      JSONArray metadataJsonArray = (JSONArray) new JSONParser().parse(metadataJson);
-
+      final MetadataList metadataList = new MetadataList();
+      final MediaPackageElementFlavor flavor = new MediaPackageElementFlavor("dublincore", "episode");
```
Arnei: These are never used

Arnei (changes requested):
While I do appreciate the effort of making the Upload UI look less cramped, doing so only for the captions is inconsistent and makes the UI arguably more confusing. Therefore I'd propose undoing those changes and keep overhauling the UI for a seperate PR.
Bildschirmfoto vom 2021-07-21 11-29-31

gregorydlogan (approved):
Aside from the changes already requested this looks ready to go.

Arnei (approved):

PR #2809, opened 2021-07-08
Ingest with tags

lkiesow:
This patch allows users to specify tags for media much like flavors,
when ingesting via `addMediaPackage`.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
@@ -91,6 +91,7 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
@@ -640,10 +641,11 @@ public class IngestRestService extends AbstractJobProducerEndpoint {
           @RestParameter(description = "URL of series DublinCore Catalog", isRequired = false, name = "seriesDCCatalogUri", type = RestParameter.Type.STRING),
           @RestParameter(description = "Series DublinCore Catalog", isRequired = false, name = "seriesDCCatalog", type = RestParameter.Type.STRING),
           @RestParameter(description = "Access control list in XACML or JSON form", isRequired = false, name = "acl", type = RestParameter.Type.STRING),
+          @RestParameter(description = "Tag of the next media file", isRequired = false, name = "tag", type = RestParameter.Type.STRING),
           @RestParameter(description = "URL of a media track file", isRequired = false, name = "mediaUri", type = RestParameter.Type.STRING) },
       bodyParameter = @RestParameter(description = "The media track file", isRequired = true, name = "BODY", type = RestParameter.Type.FILE),
       responses = {
-          @RestResponse(description = "Ingest successfull. Returns workflow instance as xml", responseCode = HttpServletResponse.SC_OK),
+          @RestResponse(description = "Ingest successful. Returns workflow instance as xml", responseCode = HttpServletResponse.SC_OK),
           @RestResponse(description = "Ingest failed due to invalid requests.", responseCode = HttpServletResponse.SC_BAD_REQUEST),
           @RestResponse(description = "Ingest failed. Something went wrong internally. Please have a look at the log files",
               responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR) },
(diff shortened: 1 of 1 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -640,10 +641,12 @@ protected Response addMediaPackageElement(HttpServletRequest request, MediaPacka
           @RestParameter(description = "URL of series DublinCore Catalog", isRequired = false, name = "seriesDCCatalogUri", type = RestParameter.Type.STRING),
           @RestParameter(description = "Series DublinCore Catalog", isRequired = false, name = "seriesDCCatalog", type = RestParameter.Type.STRING),
           @RestParameter(description = "Access control list in XACML or JSON form", isRequired = false, name = "acl", type = RestParameter.Type.STRING),
+          @RestParameter(description = "Flavor of the next media file", isRequired = false, name = "flavor", type = RestParameter.Type.STRING),
```
Arnei: Why are we duplicating the "flavor" parameter here? It's already specified as required earlier. Looks good otherwise.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -838,7 +847,8 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
               } catch (java.net.URISyntaxException e) {
                 return badRequest(String.format("Invalid URI %s for media", value), e);
               }
-              ingestService.addTrack(mediaUrl, flavor, mp);
+              ingestService.addTrack(mediaUrl, flavor, tags.toArray(new String[0]), mp);
```
gregorydlogan, line 853: Ok, bare with me here.  We want to be able to tag, and flavor some combination of media files, and dc catalogs.  To do this we are forced to rely on the *ordering* of the parameters in the request, right?

In that case shouldn't we be clearing `flavor` as well, since otherwise it's going to be applied to anything *else* we might ingest as well?  c.f: a request with parameter ordering `flavor`, `tag`, `t1`, `t2` is going to apply the flavor and tag to `t1`, but only the flavor to `t2`.

lkiesow: Hood point. But removing functionality from the ingest API is dangerous, so I rather not do that.
I could try and make this consistent by only clearing this if new tags are set after a media file:

```
flavor=a
tag=b
tag=c
track_a (with flavor a and tags b and c)
track_b (with flavor a and tags b and c)
```

and

```
flavor=a
tag=b
tag=c
track_a (with flavor a and tags b and c)
flavor=x
tag=y
track_b (with flavor x and tag y)
```
Or do you think it is save to disallow a second track without dedicated flavor?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
lkiesow: You are right → fixed.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
gregorydlogan: I think it's likely safe to disallow tracks without flavours. My read of the rest docs says that you need to ingest (flavor, track) pairs already.

I'm also struggling to imagine a legit usecase for setting a single flavour and ingesting multiple files. Concatenation after hard clips/crashes on the CA?

lkiesow: As discussed at the technical meeting, a warning has now been added in 588a4df8a8f0ce63bed3cab0ae6916b41b9cf7ca

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -893,6 +910,20 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
     }
   }
 
+  private void warnIfFlavorAlreadyUsed(final boolean used) {
+    if (used) {
+      logger.warn("\n"
+          + "********************************************\n"
+          + "* Warning: Re-use of flavors during ingest *\n"
+          + "*          is deprecated and will be       *\n"
+          + "*          removed s*on! Declare a flavor  *\n"
```
gregorydlogan: Is the `*` in `s*on` meant to be there?

lkiesow: fixed

PR #2826, opened 2021-07-12
Update Deprecated Code

lkiesow:
This is a batch update of old and mostly deprecated code I've updated
since I was working on and with this as part of #2815 anyway.

It shouldn't change how the code works. It mostly fixes logger
statements and replaces custom code with functions from Java which have
been introduced since and which make things a lot easier. This is also
the reason why I code remove a bunch of old utility functions in the
end.

- I rather push this towards `develop` instead of stable since it changes a lot of code and doesn't really do anything except for cleaning up
- This pull request contains #2815

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/StaticMetadataServiceDublinCoreImpl.java
```diff
@@ -124,16 +123,11 @@ public void activate(@SuppressWarnings("rawtypes") Map properties) {
...
-            .map(new Function<DublinCoreCatalog, StaticMetadata>() {
-              @Override
-              public StaticMetadata apply(DublinCoreCatalog episode) {
-                return newStaticMetadataFromEpisode(episode);
-              }
-            })
-            .getOrElse((StaticMetadata) null);
+    Catalog[] catalogs = mp.getCatalogs(MediaPackageElements.EPISODE);
+    if (catalogs.length > 0) {
+      return newStaticMetadataFromEpisode(DublinCoreUtil.loadDublinCore(workspace, catalogs[0]));
```
gregorydlogan, line 128: This seems like it's making the assumption that the *first* catalog is *always* the episode catalog.  Is that a safe assumption?

gregorydlogan (approved):

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -622,37 +612,31 @@ static void addEpisodeMetadata(final SolrInputDocument doc, final StaticMetadata
...
 
   static List<DField<String>> fromMValue(List<MetadataValue<String>> as) {
-    return map(as, new ArrayList<DField<String>>(), new Function<MetadataValue<String>, DField<String>>() {
-      @Override
-      public DField<String> apply(MetadataValue<String> v) {
-        return new DField<String>(v.getValue(), v.getLanguage());
-      }
-    });
+    return as.stream()
+        .map(v -> new DField<>(v.getValue(), ""))
```
KatrinIhler, line 632: Why did you change the second parameter from v.getLanguage() to "" here?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -622,37 +612,31 @@ static void addEpisodeMetadata(final SolrInputDocument doc, final StaticMetadata
...
 
   static List<DField<String>> fromDCValue(List<DublinCoreValue> as) {
-    return map(as, new ArrayList<DField<String>>(), new Function<DublinCoreValue, DField<String>>() {
-      @Override
-      public DField<String> apply(DublinCoreValue v) {
-        return new DField<String>(v.getValue(), v.getLanguage());
-      }
-    });
+    return as.stream()
+        .map(v -> new DField<>(v.getValue(), ""))
```
KatrinIhler, line 638: Same here.

KatrinIhler (commented):
I basically only found one thing I don't understand, rest looks very nice. :D (Delete aaall the custom functional code.)

KatrinIhler:
Okay, just ignore my comments, I took a look at the other pull request. ;) So I guess we're waiting for the other one to be merged ahead to develop because then this will surely conflict?

KatrinIhler (approved):

lkiesow:
> So I guess we're waiting for the other one to be merged ahead to develop because then this will surely conflict?

Why would it? It's the same commit. And it's in `develop` now.

KatrinIhler:
Including PRs in other PRs has caused conflicts for me a lot, even though the commits are the same, hash and all. :thinking: Maybe I'm doing something wrong.

Either way, let's merge this. :D

PR #2834, opened 2021-07-15
Remove Security-related Workaround in AssetManager

KatrinIhler:
This workaround was introduced by #150 because the old onion-model AssetManager saved the ACL properties _after_ sending the TakeSnapshot message. So if you used the media package belonging to the newly-created snapshot, the asset URLs would point at the archive, which means that when AuthorizationService.getActiveACL(mp) got called to attach the ACL to the message, 
the AssetManager would check the ACL properties in the database to see if the user was allowed to request the security-related assets... but those properties weren't set yet, so this would fail for anyone who wasn't an Admin.

But because of the layers of the AssetManager, the security layer had no way to do this at the right time: _after_ the snapshot was created, but _before_ the message is sent. To do this correctly, you would have to switch the layers so that security stuff could happen before the messaging stuff, but that would have obviously been a major change. (Not so flexible after all, huh?) So we worked around that by using the old media package before the snapshot happened, because those URLs still pointed at the WFR, which doesn't do that problematic security check. But this required like ten lines of explanation, which is no bueno.

With the new all-in-one AssetManager introduced by #2732, this suddenly becomes an easy thing to fix by just changing the order in takeSnapshot(), and I actually did already fix it in that PR because the direct index updates would otherwise fail. I just overlooked the workaround for the messaging stuff.

So with this PR: Workaround begone, so it can't confuse any more people.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -407,20 +407,10 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
                 mkPropertyName(ace.getRole(), ace.getAction())), Value.mk(ace.isAllow())));
       }
 
-      // We pass the original media package here, instead of using
-      // snapshot.getMediaPackage(), for security reasons. The original media
-      // package has elements with URLs of type http://.../files/... in it. These
-      // URLs will be pulled from the Workspace cache without a HTTP call.
-      //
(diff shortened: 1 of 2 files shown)


KatrinIhler:
To test this, use a Non-Admin user to create a new event. If I did something wrong, that should blow up. ;)

lkiesow:
Looking at the tests, this seems to break the schaduler.

KatrinIhler:
@lkiesow Already on my to-do list (of course only running the asset manager tests was too optimistic. >_>)

KatrinIhler:
I fixed the scheduler tests, and I'm not happy with the way I did it. But I think any better way would mean a rewrite of the Scheduler tests, which is not in itself a bad idea but is very much out of scope for what I considered a minor fix. Let's see what people say...

PR #2836, opened 2021-07-15
Target Java 11

lkiesow:
This patch sets the target Java version for most modules to 11 which is the
effective minimum requirement for Opencast 10 anyway. This allows the usage of
newer Java syntax and features (Example incuded).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/security-aai/pom.xml
+++ b/modules/security-aai/pom.xml
@@ -13,6 +13,7 @@
   <properties>
     <opencast.basedir>${project.basedir}/../..</opencast.basedir>
     <checkstyle.skip>false</checkstyle.skip>
+    <java.release>8</java.release>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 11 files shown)


LukasKalbertodt (approved):
Very nice! I was looking forward to that. 

It compiles fine for me, tests are green and it seems to work. The patch also looks good to me. I don't quite understand the reason for those `Import-Package` changes, but oh well. It works :P

lkiesow:
> I don't quite understand the reason for those Import-Package changes, but oh well. It works :P

The new bundler, probably in combination with the Java 10 (11?) changes to the internal Java modules seems to cause the imports to work slightly different so that more requirements end up in the manifests. There aren't available at runtime which causes OSGi to complain (wiring errors). These additional libraries shouldn't be necessary (it did run just fine without them before after all), so I adjusted the imports. That, and I fixed a few actual errors which seemed to have worked by luck before.

modules/security-aai/pom.xml
```diff
@@ -13,6 +13,7 @@
   <properties>
     <opencast.basedir>${project.basedir}/../..</opencast.basedir>
     <checkstyle.skip>false</checkstyle.skip>
+    <java.release>8</java.release>
```
gregorydlogan, line 16: What is the AAI code doing that restricts it to JDK 8?

lkiesow: Test failures based on the Spring test library it is using.

PR #2846, opened 2021-07-19
Some bug fixes to IBM Watson transcription service

rute-santos:
- Do not pass user_token if not registered to get callback notifications (to avoid warning).
- Use preemptive http basic auth (Watson returns a 503 in some cases when too much load instead of prompting for auth credentials).
- Check for jobs in endless waiting state.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
+++ b/modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
@@ -68,17 +68,23 @@ import org.opencastproject.workspace.api.Workspace;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
+import org.apache.http.HttpHost;
 import org.apache.http.HttpStatus;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.AuthCache;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.protocol.HttpClientContext;
 import org.apache.http.entity.FileEntity;
+import org.apache.http.impl.auth.BasicScheme;
+import org.apache.http.impl.client.BasicAuthCache;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.DefaultHttpRequestRetryHandler;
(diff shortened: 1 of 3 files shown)


karendolan:
@rute-santos  can you provide guidance on how someone without IBM Watson  account can test this.

rute-santos:
We can provide someone temporary credentials to test. I have to say that the issues fixed by this PR are not easily reproducible because they only happen when there's a problem on IBM's side. FWIW, we have had these changes working fine in our production environment for a few months.

modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
```diff
@@ -740,12 +774,10 @@ public MediaPackageElement getGeneratedTranscription(String mpId, String jobId)
...
 
   protected CloseableHttpClient makeHttpClient() {
-    CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
-    credentialsProvider.setCredentials(AuthScope.ANY, new UsernamePasswordCredentials(user, psw));
     RequestConfig reqConfig = RequestConfig.custom().setConnectTimeout(CONNECTION_TIMEOUT)
             .setSocketTimeout(SOCKET_TIMEOUT).setConnectionRequestTimeout(CONNECTION_TIMEOUT).build();
-    return HttpClients.custom().setDefaultCredentialsProvider(credentialsProvider).setDefaultRequestConfig(reqConfig)
-            .build();
+    return HttpClients.custom().setDefaultRequestConfig(reqConfig)
+            .setRetryHandler(new DefaultHttpRequestRetryHandler(3, true)).build();
```
gregorydlogan, line 780: Is 3 defined in their spec?  It seems like a reasonable number of attempts, but I'm not a fan of hardcoded values.

gregorydlogan (commented):
Looks good too me.  If you have this running in prod then we're likely fine to merge.

modules/transcription-service-ibm-watson-impl/src/main/java/org/opencastproject/transcription/ibmwatson/IBMWatsonTranscriptionService.java
rute-santos: No, it isn't. I didn't realize this fix wasn't already there... I looked at our ticket history and this was a fix we made more than a year ago for dealing with stale connections in the http client library (we would get a "failed to respond" error, but a re-try of the operation a few minutes later would just work). 3 was an arbitrary number and we didn't see those errors anymore. Again, not reproducible :(
I can make it a constant in the Java class or it could be a configuration option in the service, which I honestly think is overkill. It's always good to retry when using an external service and 3 also seems a good number to me:)

gregorydlogan: Ok. I'm not a huge fan but the choice seems reasonable here. Chances are pretty good that if 3 attempts fail then N will fail too - the service is probably having some downtime. Adding user configurability to this isn't gaining anything but code complexity :) 

Let's just leave this as is.

PR #2849, opened 2021-07-20
Add exception for extron smp351 technical catalog

geichelberger:
Like the BBB integration the Extron SMP351 ingests a JSON structure as an xml catalog. Without that fix the SMP fails to ingest to Opencast caused by the bad request.

Tested with Extron firmware 2.14

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -936,8 +936,9 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
       final boolean isJSON;
       try (InputStream inputStream = workingFileRepository.get(mediaPackageId, elementId)) {
         try (BufferedReader reader  = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
-          // Exception for current BBB integration which is ingesting a JSON array as catalog
-          isJSON = reader.read() == '[';
+          // Exception for current BBB integration and Extron SMP351 which is ingesting a JSON array/object as catalog
+          int firstChar = reader.read();
+          isJSON = firstChar == '[' || firstChar == '{';
         }
       }
 


lkiesow:
What are they ingesting as a catalog?

gregorydlogan:
Doesn't this also effectively undo the recent change WRT to catalog format requirements?

geichelberger:
They ingest a catalog of the type technical/extron-smp-351.
```
{
 "package": { 
   "version":"3.001.0002", 
   "name": "SMP-NAME_DATE_TIME",
   "location": "SMP_LOCATION",
   "db_id": "113887",
   "event_id": "EVENT_ID",
   "owner":"",
   "source_device":"DEVICE",
   "device_id":"",
   "record_mode":"0",
   "package_size":   7325075, 
   "streams":  [
      {
      "number":1,
      "timedtext": [
                   ],
      "recordings":  [
         {
            "number":1,
            "name":"RECORDING.m4v",
            "created":"2021-07-20T14:46:30+02:00",
            "start":"00:00:00",
            "start_ms":0,
            "duration":"00:00:17",
            "duration_ms":17666,
            "file_size":7323642
         }
      ],
      "chapters":  [
      ],
      "thumbnails": [
		{ "timestamp":15033,"name":"00015033.jpg","title":""}
      ]
   }
   ],
   "metadata": {

		"dc:contributor":"",
		"dc:coverage":"",
		"dc:creator":"",
		"dc:date":"2021-07-20T12:46:30Z",
		"dc:description":"",
		"dc:format":"",
		"dc:identifier":"EVENT_ID",
		"dc:language":"",
		"dc:publisher":"",
		"dc:relation":"",
		"dc:rights":"",
		"dc:source":"",
		"dc:subject":"",
		"dc:title":"Ingesst Test 1",
		"dc:type":"",
		"dc:course":""
   },
   "local_time":"2021-07-20T14:46:30+02:00" ,
   "utc_time":"2021-07-20T12:46:30Z" ,
   "timezone":{"id":"CET","description":"(UTC+01:00) Central European Time"}
 }
}
```

gregorydlogan:
IMO the solution here is to get Extron to patch their firmware to attach this as an Attachment, rather than a Catalog.  This doesn't contain anything that's not already in an Opencast metadata file somewhere.

geichelberger:
I have no idea why they add that stuff at all, and we will definitely report the problem to Extron, but it is easier to have a temporary patch in the meantime. Extron is not the fastest at fixing bugs, and we are not the only ones using Extron as capture agents.

mtneug:
This would open it up to any non-trivial JSON documents again 😄 . Maybe this can be automatically ingested as an attachment instead? This would be very unexpected, though.

lkiesow:
@gregorydlogan, if you do not disagree, I would like to merge this. It will help a number of adopters since the SMP351 is a popular hardware device used with Opencast and being stubborn about this will hurt lots of people.

That said, it would be nice to get rid of this eventually. If you could reach out to Extron @geichelberger, that would be great. If you want, add me in CC. They likely just need to replace the call to [`/ingest/addCatalog`](https://develop.opencast.org/docs.html?path=/ingest#addCatalogInputStream-3) with [`/ingest/addAttachment`](https://develop.opencast.org/docs.html?path=/ingest#addAttachmentInputStream-1). All parameters should be identical.

> This would open it up to any non-trivial JSON documents again smile . Maybe this can be automatically ingested as an attachment instead? This would be very unexpected, though.

Doing that is just confusing. Even allowing JSON again, XML would not still be checked, so it's still an improvement.

gregorydlogan:
Yeah go ahead. I might jot like it but at the end of the day we likely won't see a firmware fix broadly applied for a year at least. In the meantime this fixes things, but let's rememeber to roll this back our in OC 11 or 12.

reiferschris:
@geichelberger If you have access to the new firmware version 3.x by extron, you can implement Lars suggestion simply by modifying the manifest URL in the publish settings of the smp. Just change the last part `/ingest/addCatalog` to `/ingest/addAttachment.`
Unfortunately we don't have CAs with the old firmware and I don't remember if this setting was available before.

ziegenberg:
The new firmware version 3.x requires a so-called "Flex OS App" for Opencast. To our knowledge, there's currently only one for Panopto and Kaltura but none for Opencast publicly available.

reiferschris:
We got the required “Flex OS App” for Opencast from extron and it works. I'm sure they will give it to you if you ask for it, now that you know it exists. I'm pretty sure that Konstanz also has access to this app, so its existence is hardly a secret.

lkiesow:
So, let's get this merged for now.
For anyone wondering:
This will be part of the 10.2 release scheduled to be released on August 18th.

PR #2850, opened 2021-07-20
Fix lti rest endpoint annotations and ogsi properties

### LINKED ISSUE #2180: LTI Rest Endpoint on Admin throws Exception
The REST docs of the LTI service on the admin node throws an exception:

lti-rest-endpoint error

This happens with the new tools on `r/9.x`

### LINKED ISSUE #2442: The lti-service-remote bundle is missing its endpoint definitions
**Describe the bug**
As reported by a user on Matrix: 
`Hey, does anybody know a reason why deleting an event from the LTI series view on the presentation node redirects to /login as if the user is not authenticated, which then returns a '405 not allowed', while doing exactly the same from an admin node works perfectly?`

In investigating why the 405 was being returned, I noticed that *[the entire endpoint class is missing the endpoint annotations](https://github.com/opencast/opencast/blob/develop/modules/lti-service-remote/src/main/java/org/opencastproject/lti/endpoint/LtiServiceRemoteImpl.java#L166)*, hence the 405 since the endpoint doesn't exist without them.  Looks like they were never present, and this was missed during code review.

geichelberger:
Declaring the seriesId as a path parameter caused the IllegalStateException #2180. 
The problem described in #2442 should be caused by two OSGi services having the same `opencast.service.type` property value.

Fixes #2180  and should fix #2442

### DIFF
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/endpoint/LtiServiceRestEndpoint.java
@@ -220,8 +220,9 @@ public class LtiServiceRestEndpoint {
               name = "eventId",
               description = "The event (id) to copy",
               isRequired = true,
-              type = STRING
-          ),
+              type = STRING)
+      },
+      restParameters = {
           @RestParameter(
               name = "seriesId",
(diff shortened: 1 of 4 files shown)


Arnei (approved):

lkiesow:
The endpoints are still having lots of issues, but this improves the situation → merging

PR #2851, opened 2021-07-20
Server statistics can overwelm database

lkiesow:
The server statistics page in the admin interface can create very complex and slow queries due to the including of live job statistics. If multiple people had this page open at the same time in a larger cluster, the responses could even be slower than the request, causing the requests to pile up and essentially taking down the database, causing the whole cluster to break down.

This patch fixes the problem by caching the response so that only one request at a time with a maximum frequency of 30 sec will be sent to Opencast.

The data and data retrieval methods are still the same and partially still make no sense, but this should already fix the immediate danger to the cluster.

As a side effect, this also removes the Entwine functional library and fixes some response errors.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):
LGTM

PR #2855, opened 2021-07-21
Added speech to text (Vosk)

marwyg:
This commit adds a new workflow operation that can generate subtitles with [Vosk](https://alphacephei.com/vosk/). Here is an example, how the workflow operation can look like:

```
    <operation
        id="speechtotext"
        fail-on-error="true"
        description="Generates subtitles">
      <configurations>
        <configuration key="source-flavor">*/source</configuration>
        <configuration key="target-flavor">captions/vtt+en</configuration>
        <configuration key="target-element">attachment</configuration>
        <configuration key="target-tags">archive,subtitle,engage-download</configuration>
        <configuration key="language-code">eng</configuration>
      </configurations>
    </operation>
```

The "target-element" (optional) field describes, where the ".vtt" file should be appended in the media package (possibilities are: as a **track** or as an **attachment**). The default is "attachment".

The "language-code" (optional) defines the language model that vosk shall use. It should match the directory name of the language model. The default is "eng".

**Dependencies:**

For this to work, you have to install vosk-cli: https://github.com/elan-ev/vosk-cli
Further requirements:
- ffmpeg
- vosk (with pip - package installer for python)
- webvtt-py (pip)

If the subtitles shall show up in the paella player, the `target-flavor` field has to end with `.../vtt+LANG`, like in the example above. In the workflow operation example above `LANG` was set to `en`. It will just determine what selectbox option will appear in the paella player.


marwyg:
I cannot upload files with a ".py" file extension. So please rename the script to "test_webvtt.py".
[test_webvtt.txt](https://github.com/opencast/opencast/files/6856848/test_webvtt.txt)

Here is also the requirements.txt including the necessary python packages:
[requirements.txt](https://github.com/opencast/opencast/files/6856876/requirements.txt)

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,322 @@
...
+
+
+  @Activate
+  public void activate(ComponentContext cc) {
+    super.activate(cc);
+    logger.debug("Activated speech to text service");
+  }
+
+  @Override
+  public void updated(Dictionary properties) throws ConfigurationException {
```
lkiesow: Don't implement an unnecessary managed service

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,322 @@
...
+              String.format("This service can't handle operations of type '%s'", job.getOperation()));
+    }
+
+    List<String> arguments = job.getArguments();
+    String language = arguments.get(1);
+    URI mediaFile = new URI(arguments.get(0));
+
+    // prepare the output file
+    File preparedOutputFile = new File(workspace.rootDirectory(),
+            String.format("subtitles/%d/%s.%s", job.getId(), FilenameUtils.getBaseName(mediaFile.getPath()), "vtt"));
```
lkiesow: collection?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,322 @@
...
+
+    logger.debug("Finished updating speech-to-text service");
+  }
+
+  /**
+   * If configuration changes, this will change the currently used speech-to-text engine.
+   *
+   * @param properties Contains the configuration properties.
+   */
+  private void updateUsedEngine(Dictionary properties) {
```
lkiesow: We only have Vosk.

- best case: This does nothing
- all other cases: this breaks something

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,322 @@
...
+    String language = arguments.get(1);
+    URI mediaFile = new URI(arguments.get(0));
+
+    // prepare the output file
+    File preparedOutputFile = new File(workspace.rootDirectory(),
+            String.format("subtitles/%d/%s.%s", job.getId(), FilenameUtils.getBaseName(mediaFile.getPath()), "vtt"));
+    FileUtils.forceMkdirParent(preparedOutputFile);
+
+    // copy the mediafile to the workspace directory to get an absolute path
+    File mediaFileWorkspaceCopy = copyToWorkspace(job, mediaFile);
```
lkiesow: workspace.get?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,322 @@
...
+
+    URI subtitlesURI;
+    try {
+      File subtitlesFile = selectedEngine.generateSubtitlesFile(
+              mediaFileWorkspaceCopy.toURI(), preparedOutputFile, language);
+      // copy the output file and delete the temp folder in the next line
+      subtitlesURI = workspace.putInCollection("subtitles-" + job.getId(),
+              subtitlesFile.getName(), new FileInputStream(subtitlesFile));
```
lkiesow: try-with-res?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/endpoint/SpeechToTextServiceRestEndpoint.java
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * A service endpoint to expose the {@link SpeechToTextService} via REST.
+ */
+@Path("/")
+@RestService(
+    name = "speechtotext",
+    title = "Speech to Text Service",
+    abstractText = "Create subtitles for video or audio clips.",
+    notes = { "The file format is webVTT (*.vtt)" }
```
lkiesow:
```suggestion
    notes = { "The file format is WebVTT (*.vtt)" }
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,140 @@
...
+  private static final String engineName = "Vosk";
+  private static final String VOSK_ROOT_PATH_CONFIG_KEY = "vosk.root.path";
+  private File voskRootDirectory;
+
+  @Override
+  public String getEngineName() {
+    return engineName;
+  }
+
+  @Activate
```
lkiesow, line 74: add `@Modified`

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,140 @@
...
+  public void activate(ComponentContext cc) {
+    logger.debug("Activating Vosk as viable speech-to-text engine...");
+  }
+
+  @Override
+  public void updated(Dictionary properties) throws ConfigurationException {
```
lkiesow: Use `activate()`

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,140 @@
...
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.speechtotext.api.SpeechToTextEngine#generateSubtitlesFile(URI, File, String)
+   */
+  @Override
+  public File generateSubtitlesFile(URI mediaFile, File preparedOutputFile, String language)
+          throws SpeechToTextEngineException {
+
+    final List<String> command = new ArrayList<>();
+    command.add(voskRootDirectory.getAbsolutePath() + "/venv/bin/python3"); // TODO: change this
```
lkiesow: separate project…

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -0,0 +1,317 @@
...
+      MediaPackageElementFlavor targetFlavor = track.getFlavor().applyTo(tagsAndFlavors.getSingleTargetFlavor());
+      subtitleMediaPackage.setFlavor(targetFlavor);
+
+      List<String> targetTags = tagsAndFlavors.getTargetTags();
+
+      Job inspection = mediaInspectionService.enrich(subtitleMediaPackage, true);
+      if (!waitForStatus(inspection).isSuccess()) {
+        throw new SpeechToTextServiceException(String.format(
+                "Transcription for '%s' failed at enriching process", trackURI));
+      }
```
lkiesow: Do we need this?

lkiesow (commented):
As discussed: Take these as comments

gregorydlogan:
I haven't read the code, does this add an external dependency? If so, please update the docs to reflect that this should be installed.

Rillke:
> For this to work, you have to install vosk-cli: https://github.com/elan-ev/vosk-cli

Can't see this, 404. Is this a private repo?

lkiesow:
Ups, it was. But no longer

marwyg:
I changed the test class, so it will skip if the vosk-cli package isn't installed.

ziegenberg:
Just a short question: What needs to be done for this PR to go forward? TU Wien is interested in this going into the next release. Is there a code review or some testing missing?

ziegenberg:
Any updates on this PR? Is there something we (TU Wien) could help with? Like testing?

CGreweling:
@ziegenberg we are pushing this in our production system at the UOS right now. @marwyg maybe you could push our findings so far to this pr soon?

lkiesow:
@marwyg, can you give us a short update about the current state?

- I've marked this as draft at some point, since I then knew that parts were still missing. Is it ready to be reviewed now?
- There is an empty `etc/org.opencastproject.speechtotext.impl.engine.VoskEngine.cfg`?
- Documentation about dependencies (Vosk needs to be installed) is still missing

marwyg:
There was a configuration key at the beginning, but I guess I just delete the config file then. I will add the dependencies and remove the "draft" status.

marwyg:
It's ready for review. I attached 2 Workflows to this comment for testing. One is the "fast" workflow just with the speech to text workflow operation in it. The other one is for the "actions" menu above the event table. With this workflow you can chose multiple events in the table for subtitle generation, but you have to specify a source flavor and select whether the chosen events were already published or not.

[fast_subtitles.txt](https://github.com/opencast/opencast/files/7417469/fast_subtitles.txt)
[vosk-subtitles.txt](https://github.com/opencast/opencast/files/7417493/vosk-subtitles.txt)

(Please notice, that they are .txt files)

ziegenberg:
It works on our dev setup, but fails with an NPE in production. We will try to look into this tomorrow and provide more details.

modules/speech-to-text-remote/src/main/java/org/opencastproject/speechtotext/remote/SpeechToTextServiceRemoteImpl.java
```diff
@@ -0,0 +1,92 @@
...
+      }
+      Job receipt = JobParser.parseJob(response.getEntity().getContent());
+      logger.info("Completed transcription for {}", mediaFile);
+      return receipt;
+    } catch (IOException e) {
+      throw new SpeechToTextServiceException("Failed building service request", e);
+    } finally {
+      closeConnection(response);
+    }
+  }
```
geichelberger, line 94: We get a NPE because the RemoteServiceManager is missing. My understanding is that annotations of the superclasses are ignored by default. There are two possible solutions:
Add following maven-bundle-plugin instruction `<_dsannotations-options>inherit</_dsannotations-options>` or 
define the References
```
 @Reference
  public void setTrustedHttpClient(TrustedHttpClient client) {
    this.client = client;
  }

  @Reference
  public void setRemoteServiceManager(ServiceRegistry remoteServiceManager) {
    this.remoteServiceManager = remoteServiceManager;
  }
```

lkiesow: I vaguely remember that the `<_dsannotations-options>norequirements</_dsannotations-options>` had some side effects and you should set the annotations in each service class

geichelberger: sorry wrong dsannotation-options i meant `inherit`

marwyg:
@geichelberger I added the references now. As I debugged the code, the `client` and `remoteServiceManager` variables didn't seem to be null anymore.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -157,6 +157,7 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-admin-ui-frontend/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-admin-ui-index/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-animate-workflowoperation/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-speech-to-text-workflowoperation/${project.version}</bundle>
```
lkiesow: The allinone distribution should contain workflowoperation, api and impl

docs/guides/admin/docs/workflowoperationhandlers/speech-to-text-woh.md
```diff
@@ -0,0 +1,57 @@
...
+Requirements
+------------
+
+In order for it to work, you have to install the vosk-cli package. An installation instruction can be found here: 
+https://github.com/elan-ev/vosk-cli. 
+
+Further requirements:
+- ffmpeg
+- vosk (with pip - package installer for python)
+- webvtt-py (pip)
```
lkiesow: These are just transitive dependencies from the above-mentioned tool, and users do not need to care about these if they installed that tool successfully.

```suggestion
```

docs/guides/admin/docs/workflowoperationhandlers/speech-to-text-woh.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+Operation Examples
+------------------
+
+If you want to display the subtitles in the Paella player, the WebVTT files have to be published and
+the target SubFlavor has to be configured like this: ```vtt+LANG```, where ```LANG``` is
+the language option, that will be displayed in the Paella player. Please note the description above
+for the ```language-code``` field.
```
lkiesow: Inline code just uses a single tick in Marksown:

```suggestion
the target SubFlavor has to be configured like this: `vtt+LANG`, where `LANG` is
the language option, that will be displayed in the Paella player. Please note the description above
for the `language-code` field.
```

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
```diff
@@ -0,0 +1 @@
+speechtotext.engine=Vosk
```
lkiesow: This lacks documentation. What is this configuration for?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+  public void modified(Map<String, Object> config) throws ConfigurationException {
+    logger.debug("Modified speech-to-text service");
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.osgi.service.cm.ManagedService#updated(java.util.Dictionary)
+   */
+  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
```
lkiesow: Why manually implement `ManagedService`? There seems to be no good reason. As you created a Declarative Service, you already have a managed service and get all configuration in the `@activate` and `@modified` marked methods. These can even be the same method (you can just use both annotations at the same method).

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+    // we currently just use vosk as engine, so we hardcode it here to prevent config errors and confusion
+    // if an another engine will be added, use the "updateUsedEngine" method
+    if (availableEngines == null || availableEngines.size() <= 0) {
+      throw new ConfigurationException("No speech to text engines are available");
+    }
+    for (SpeechToTextEngine engine : availableEngines) {
+      if (engine.getEngineName().equals(DEFAULT_ENGINE_NAME)) { // hardcoded "vosk" as default
+        selectedEngine = engine;
+      }
+    }
```
lkiesow: OSGi guarantees that this will never happen. Without one, the service would not start up. However, if someone would implement two, your complete service start could fail since the second engine could be selected but is not guaranteed to exist. Either you need to evaluate the used engine at runtime, or you just use a static binding to the only existing engine right now and remove the configuration.

The second would be my preference since the configuration is just confusing for users right now since they cannot actually configure anything.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+
+    jobLoad = LoadUtil.getConfiguredLoadValue(properties, JOB_LOAD_CONFIG, JOB_LOAD_DEFAULT, serviceRegistry);
```
lkiesow, line 122: This seems to be missing from the configuration file.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+      if (engine.getEngineName().equals(configuredEngineName)) {
+        selectedEngine = engine;
+      }
+    }
+    if (selectedEngine == null) {
+      logger.error("Couldn't find speak-to-text engine with name '{}'.", configuredEngineName);
+    } else {
+      logger.debug("Set speech-to-text engine to {}", selectedEngine.getEngineName());
+    }
+  }
```
lkiesow: This is not used at all
```suggestion
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -0,0 +1,325 @@
...
+    String language = arguments.get(1);
+    URI mediaFile = new URI(arguments.get(0));
+
+    File tmpSubtitlesFile = null;
+    URI subtitleFilesURI;
+    String vttFileName = String.format("%s.%s", FilenameUtils.getBaseName(mediaFile.getPath()), "vtt");
+    try {
+      // prepare the output vtt file
+      tmpSubtitlesFile = new File(String.format("%s/%s/%s/%s",
+              workspace.rootDirectory(), "collection", getCollectionId(job), TMP_PREFIX + vttFileName));
```
lkiesow: I think you want to use `workspace.getCollectionURI()`

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,152 @@
...
+        ManagedService.class
+    },
+    property = {
+        "service.description=Vosk implementation of the SpeechToTextEngine interface",
+        "service.pid=org.opencastproject.speechtotext.impl.engine.VoskEngine"
+    }
+)
+public class VoskEngine implements SpeechToTextEngine, ManagedService {
+
+  private static final Logger log = LoggerFactory.getLogger(VoskEngine.class);
```
lkiesow: Not really important, but can we name this `logger` for consistency? That's what basically everywhere else is used.

```suggestion
  private static final Logger logger = LoggerFactory.getLogger(VoskEngine.class);
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
lkiesow: This still manually implements a `ManagedService` while you already get all configuration via the `@activate` and `@modified` marked method(s). Any particular reason for that?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  public void updated(Dictionary properties) throws ConfigurationException {
+    if (properties == null) {
+      return;
+    }
+    log.debug("Start updating Vosk configuration.");
+    voskExecutable = StringUtils.defaultIfBlank(
+            (String) properties.get(VOSK_EXECUTABLE_PATH_CONFIG_KEY), VOSK_EXECUTABLE_DEFAULT_PATH);
```
lkiesow: This configuration is missing in the missing configuration file ;-)

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -0,0 +1,152 @@
...
+          throws SpeechToTextEngineException {
+
+    final List<String> command = new ArrayList<>();
+    command.add(voskExecutable);
+    command.add("-i");
+    command.add(mediaFile.toString());
+    command.add("-o");
+    command.add(preparedOutputFile.getAbsolutePath());
+    command.add("-l");
+    command.add(language);
```
lkiesow: Seems simpler, though this does not really matter and I don't have any strong feelings about wither way:

```suggestion
    final List<String> command = Arrays.asList(
      voskExecutable,
      "-i", mediaFile.toString(),
      "-o", preparedOutputFile.getAbsolutePath(),
      "-l", language);
```

modules/speech-to-text-impl/src/test/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImplTest.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  @Test
+  public void testVoskSpeechToText() throws Exception {
+    URI videoUri = getClass().getResource("/speech_to_text_test.mp4").toURI();
+    Job job = speechToTextService.transcribe(videoUri, "eng");
+    File output = new File(new URI(job.getPayload()));
+    StringBuilder textFromSpeech = new StringBuilder();
+    try (Stream<String> stream = Files.lines(Paths.get(output.getPath()), StandardCharsets.UTF_8)) {
+      stream.forEach(s -> textFromSpeech.append(s).append("\n"));
```
lkiesow: Isn't this just a more complex way of
```java
Files.lines(Paths.get(output.getPath()), StandardCharsets.UTF_8).collect(Collectors.joining("\n"));
```

I'm not actually sure what this is supposed to test except for the possible `IOException` which, when thrown, would already let the test fail by itself, wouldn't it?

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -0,0 +1,318 @@
...
+    if (!waitForStatus(job).isSuccess()) {
+      throw new WorkflowOperationException(
+              String.format("Speech-to-Text job for media package '%s' failed", parentMediaPackage));
+    }
+
+    // subtitles file is generated now, put it into the media package
+    try {
+      URI output = new URI(job.getPayload());
+      String id = UUID.randomUUID().toString();
+      InputStream in = workspace.read(output);
```
lkiesow: Close the `InputStream`
```suggestion
      try (InputStream in = workspace.read(output)) { ...
```

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -0,0 +1,318 @@
...
+      URI uri = workspace.put(parentMediaPackage.getIdentifier().toString(), id,
+              FilenameUtils.getName(output.getPath()), in);
+
+      MediaPackageElement subtitleMediaPackage;
```
lkiesow: This name really confused me while reading the code. The code ends up adding media packages to media packages. Let's name this element ;-)

```suggestion
      MediaPackageElement subtitleMediaPackageElement;
```

lkiesow (changes requested):
Reviewing the code, I found a few more errors. If you have questions about any comments, please let me know.

Additionally, here are two test media files which are a lot smaller and should still work:

- The same audio stream, just without the video stream (which you shouldn't need): https://data.lkiesow.io/opencast/speech_to_text_test.mp4
- The audio as slightly higher compressed mono stream: https://data.lkiesow.io/opencast/speech_to_text_test_small.mp4

I'll try doing some runtime tests tomorrow to finish the review.

assemblies/karaf-features/src/main/feature/feature.xml
marwyg: They get included with "opencast-services-processing-heavy-load", but I just realized, that the modules are sorted alphabetically. I rearranged the speech-to-text modules a bit.

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
marwyg: It's used for the method `updateUsedEngine`. The idea was, that you can change the used engine with this key. But I guess it won't happen in the near future, so I just delete it.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
marwyg: Hm I don't quite remember, I think something didn't work as expected when I did it like that. But maybe I did something wrong, I will try it again.

docs/guides/admin/docs/workflowoperationhandlers/speech-to-text-woh.md
```diff
@@ -0,0 +1,57 @@
...
+|target-element    |no      |Define where to append the subtitles file. Possibilities are: as a 'track' or as an 'attachment'. The default is "attachment".|
+|language-code     |no      |The language of the video or audio source (default is "eng). It has to match the name of the language model directory. See 'vosk-cli'.|
+|target-tags       |no      |Tags for the subtitle file                                                       |
+
+
+Requirements
+------------
+
+In order for it to work, you have to install the vosk-cli package. An installation instruction can be found here: 
+https://github.com/elan-ev/vosk-cli.
```
lkiesow: Let's turn this into a link

```suggestion
In order for it to work, you have to install the vosk-cli package. Installation instructions can be found at
[github.com/elan-ev/vosk-cli](https://github.com/elan-ev/vosk-cli).
```

docs/guides/admin/docs/workflowoperationhandlers/speech-to-text-woh.md
```diff
@@ -0,0 +1,57 @@
...
+for the ```language-code``` field.
+
+```XML
+<operation
+    id="speechtotext"
+    description="Generates subtitles for video and audio files">
+  <configurations>
+    <configuration key="source-flavor">*/source</configuration>
+    <configuration key="target-flavor">captions/vtt+en</configuration>
+    <configuration key="target-element"> attachment | track </configuration>
```
lkiesow: Since this is supposed to be an example, it should contain one of the possible options:

```suggestion
    <configuration key="target-element">attachment</configuration>
```

assemblies/karaf-features/src/main/feature/feature.xml
lkiesow: Ups, missed that. Ignore my comment :)

lkiesow:
> - The audio as slightly higher compressed mono stream: https://data.lkiesow.io/opencast/speech_to_text_test_small.mp4

Trying out both, it's interesting to see that the one with the higher compression and worse audio quality actually generates the more accurate result :grin:

lkiesow:
Runtime tests look good. Apart from the two additional comments about documentations, I don't think I have anything to add.

marwyg:
> > * The audio as slightly higher compressed mono stream: https://data.lkiesow.io/opencast/speech_to_text_test_small.mp4
> 
> Trying out both, it's interesting to see that the one with the higher compression and worse audio quality actually generates the more accurate result grin

Hm 🤔 maybe it's worth to do some more testing. Nonetheless, I replaced the original media file with the smaller one.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
marwyg: Seems to work.

lkiesow (approved):

PR #2863, opened 2021-07-25
Fix Maven Plugin Multithreading Problems

lkiesow:
This patch fixes a few issues appearing when building with multiple
threads:

- The download-maven-plugin file locking [was broken](https://github.com/maven-download-plugin/maven-download-plugin/issues/95), causing several modules (e.g. studio, editor) to fail. Updating the plugin fixes the problem.

- The phantomjs-maven-plugin is not thread-safe and causes all Theodul modules to fail. Unfortunately, [it is no longer maintained](https://github.com/klieber/phantomjs-maven-plugin#phantomjs-maven-plugin) and we should consider removing it. Especially, since it is only needed for running the Jasmine tests which are disabled by default and also partly broken and known to be unreliable (that's why they were deactivated back then).

  For now, this patch makes Maven skip the phantomjs-maven-plugin as well if the Jasmine tests are disabled.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -617,17 +617,17 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-assembly-plugin</artifactId>
-          <version>3.1.0</version>
+          <version>3.3.0</version>
         </plugin>
         <plugin>
           <groupId>com.googlecode.maven-download-plugin</groupId>
           <artifactId>download-maven-plugin</artifactId>
-          <version>1.5.0</version>
+          <version>1.6.6</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-resources-plugin</artifactId>
-          <version>3.1.0</version>
+          <version>3.2.0</version>
         </plugin>
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
@@ -701,6 +701,7 @@
             </execution>
           </executions>
           <configuration>
+            <skip>${skipJasmineTests}</skip>
             <version>1.9.8</version>
           </configuration>
         </plugin>


mliradelc:
I was looking how to run/try this, is
```
$ mvn clean install -T $(sysctl hw.logicalcpu | grep -o '[0-9]') -Pdev
```

The correct way to run this?

(`...sysctl hw.phys...` is for macOS only)

mtneug:
`nproc` gives you the number of threads on your system. But `1.0C` should give you that aswell.

mliradelc:
> `nproc` gives you the number of threads on your system. But `1.0C` should give you that aswell.

Yes, but you need to install `coreutils` before to use it, anyway, is more easy to use `nproc` 😄 

Is it 1.0C single thread?

mtneug:
I'm using the following:

```
# sequential
$ rm -rf ~/.m2
$ mvn clean install

# parallel
$ rm -rf ~/.m2
$ mvn clean install --threads 1.0C -Pnone
```

The parallel build fails for `editor-service-endpoint-remote`:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.2:unpack (unpack) on project opencast-editor-service-endpoint-remote: Unable to find/resolve artifact.: Could not find artifact org.opencastproject:opencast-editor-service-endpoint:jar:10-SNAPSHOT in central (https://repo.maven.apache.org/maven2/) -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :opencast-editor-service-endpoint-remote
```

mtneug (changes requested):
Fix build with clean local repository.

mtneug:
If we want to support parallel builds, we should also use that in CI. What do you think?

lkiesow:
> If we want to support parallel builds, we should also use that in CI. What do you think?

I think that's a good idea, but I thought, I let people play with it for a bit and figure out more problems first :grin:

lkiesow:
> Is it 1.0C single thread?

No. That's a core modifier. It depends on your machine. For me, that will result in 16 threads.
On an octa-core it should result in 8 threads, …

lkiesow:
> `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.1.2:unpack (unpack) on project opencast-editor-service-endpoint-remote: Unable to find/resolve artifact.: Could not find artifact org.opencastproject:opencast-editor-service-endpoint:jar:10-SNAPSHOT in central (https://repo.maven.apache.org/maven2/) -> [Help 1]`

Kind of sounds like a potential problem?

mtneug:
To be clear: `editor-service-endpoint` has not been built yet. Apparently, `editor-service-endpoint-remote` depends on `editor-service-endpoint`. Since it's not locally available, Maven tries to download it. Looking at the POM file, it seems some class files are copied over?

lkiesow:
That's a perfect example of something working by accident. There is definitely a dependency definition missing, since it clearly depends on the other package. Will you create a pull request fixing that, or shall I?

mliradelc:
> > Is it 1.0C single thread?
> 
> No. That's a core modifier. It depends on your machine. For me, that will result in 16 threads.
> On an octa-core it should result in 8 threads, …

Thanks for the answer, the docs I found on the internet, didn't tell anything about that is for choosing all the cores.

lkiesow:
> To be clear: `editor-service-endpoint` has not been built yet. Apparently, `editor-service-endpoint-remote` depends on `editor-service-endpoint`. Since it's not locally available, Maven tries to download it. Looking at the POM file, it seems some class files are copied over?

That's now fixed in #2879. Anything else holding this one back?

lkiesow:
Pull request #2879 is merged. Anything holding this back?

mtneug (approved):
No LGTM

mtneug:
Found another problem on systems with a clean npm install: the `opencast-engage-paella-player` module runs `npm cache verify` which could interfere with parallel `npm install` operations. I don't see why Opencast builds should execute `npm cache verify` so let's remove it: #2917

PR #2876, opened 2021-07-29
I am a test, pls ignore me

KatrinIhler:


lkiesow:
For testing, you could also just push to your own repository. The workflow will run there as well.

KatrinIhler:
I guess I was a bit confused about the details - for PRs, all changes are taken into account, but if I push to a branch, it's only the last commit, right? This also provides a nice overview, but it clutters up the repo, especially because you can't delete PRs, only close them...  Maybe next time, I'll aim a PR at my own repo, just figured out that's possible. :D

(And even with all the testing, I still got it wrong... :cry:)

lkiesow:
If you need help, I originally wrote that test ;-)

PR #2857, opened 2021-07-22
One Elasticsearch index to rule them all

KatrinIhler:
(Is the joke getting old yet?)

With this PR, there will be only one Elasticsearch index supporting both the Admin UI and the External API. The old index implementations were removed, the new one replaced the AbstractSearchIndex class that already contained most of the relevant functionality. The index-rebuild module was merged with the elasticsearch-index module for simplification.

The endpoints to clear and rebuild the index were moved to /index. Both the Admin UI and External API endpoints related to that were removed. This might slightly violate the External API versioning, but there was no good way to keep this backwards-compatible, and you shouldn't trigger an index rebuild from an external application anyway.

The new index has a new identifier that Opencast uses to find it. When updating to OC 11, there will be three different ways to get this to work (can you tell that I couldn't decide which one was better?)

1. Just start an index rebuild. A simple way to go for smaller systems.
2. If an index rebuild will not be required for other reasons, you can use an included migration script to rename your admin UI index. There might be some things to consider about the cloning process used if you have a complex Elasticsearch setup, but for most people this should be straightforward.
3. Both too much work? If you don't care about consistency, you can use the new config option to just keep using your old admin UI index without renaming it.

Additionally, there's a second script to remove the old indices if you want to. I hope I made it sufficiently clear that that should be used with caution. 

I hope I added/updated all the relevant docs. I cleared the release notes for OC11, I think that was the right thing to do?

While I was at it, I did some very minor cleanup where unnecessary complex method parameters and unused classes were concerned. I also killed the EventIndex interface.

I added some configuration and split up the index name into name and identifier (one used for logging, one to find the actual index), because I felt they had different use cases. In general, I struggled with some indecisiveness where naming was concerned, so I hope the commits aren't too confusing. If you struggle understanding something, just ask.

PS: Forgot to add the best thing - the index rebuild should now be twice as fast ;)

PR #2877, opened 2021-07-29
Only run Github database test for sql scripts

KatrinIhler:
This way, the test script won't fail if we put any other scripts into our upgrade folders, as currently happening for #2857.

I tested this by creating a small test PR. When you put an sql script into an upgrade folder, that causes this test run, but a simple text file now won't trigger the test script, while it did before.

### DIFF
--- a/.github/workflows/test-database-upgrade.yml
+++ b/.github/workflows/test-database-upgrade.yml
@@ -3,10 +3,10 @@ name: Test database upgrade
 on:
   pull_request:
     paths:
-      - 'docs/upgrade/**'
+      - 'docs/upgrade/*/mysql5.sql'
   push:
     paths:
-      - 'docs/upgrade/**'
+      - 'docs/upgrade/*/mysql5.sql'
 
 jobs:
   # Test database upgrade path.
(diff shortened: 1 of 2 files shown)


KatrinIhler:
@gregorydlogan Do you want to take a look at this?

lkiesow (changes requested):
> This way, the test script won't fail if we put any other scripts into our upgrade folders, as currently happening for #2857.

Not true. This way, the test will not __run__ if you put other things in there. But the test will run and fail as soon as the next person actually touches the scripts.

lkiesow:
Modify https://github.com/opencast/opencast/blob/develop/docs/upgrade/.test.sh and test if `$dir/mysql5.sql` exists. Skip a folder if it does not.

lkiesow:
Or use
```
for sqlfile in *_to_*/mysql5.sql; do
   ...
done
```

KatrinIhler:
@lkiesow Okay, I think now I got it. The first commit ensures that the test only runs if an SQL script was changed. The second commit will check if a version folder actually contains an SQL script before trying to run it. :)

lkiesow (approved):

PR #2880, opened 2021-07-29
Remove Unwanted Logs

lkiesow:
The error message is part of the Exception anyway and will still be
logged (and looks scary) if the exception is caught and handled
gracefully.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
@@ -1192,7 +1192,6 @@ public class IndexServiceImpl implements IndexService {
       metadataList = getMetadataListWithAllEventCatalogUIAdapters();
       MetadataJson.fillListFromJson(metadataList, (JSONArray) new JSONParser().parse(metadataJSON));
     } catch (final org.json.simple.parser.ParseException e) {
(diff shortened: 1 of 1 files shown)


gregorydlogan (approved):

karendolan:
@lkiesow is this a common log? The error appears to be commonly sent out the REST endpoint to whatever system was trying to update Opencast metadata. There is no way to debug the issue from the Opencast side without that log.

lkiesow:
I encountered it when using the editor service REST interface. Opencast was working completely as expected. And in such a situation, there should be no stack trace being logged. In that part of the code, you would get sensible error codes sent back to clients and you could enable additional logging if you need to. The problem is basically logging like this can end up in two cases:

- The exception in the line below is caught and gracefully handled later, in which case there should be no error log. It's misleading in that it implies there is something wrong with your system.
- The exception in the line below is actually causing an unexpected error, in which case this is automatically logged at most places (e.g. jax-rs (never shure if it's rs or ws :D) will log that automatically if you do not catch that in your endpoint) or, if caught, should be logged at the place where it is actually causing the unexpected error (e.g. right before returning an Internal Server Error). Keeping the log line in this case would cause the same problem to be logged multiple times, which is pretty confusing. I have people send me stack-traces with half the information often enough :(

This obviously doesn't guarantee that all endpoint do log errors properly. But keeping this just in case while doing harm by producing misleading stack traces is also not great.

PR #2892, opened 2021-08-02
Fix checkstyle violations in 6 `userdirectory*` modules

LukasKalbertodt:
Work on #2238 

- `userdirectory-brightspace`
- `userdirectory-canvas`
- `userdirectory-ldap`
- `userdirectory-moodle`
- `userdirectory-sakai`
- `userdirectory`

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
+++ b/modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
@@ -118,7 +118,8 @@ public class BrightspaceUserProviderFactory implements ManagedServiceFactory {
   @Override
   public void updated(String pid, Dictionary properties) throws ConfigurationException {
     logger.debug("updated BrightspaceUserProviderFactory");
-    String adminUserName = StringUtils.trimToNull(bundleContext.getProperty(SecurityConstants.GLOBAL_ADMIN_USER_PROPERTY));
+    String adminUserName = StringUtils.trimToNull(
+        bundleContext.getProperty(SecurityConstants.GLOBAL_ADMIN_USER_PROPERTY));
     String organization = (String) properties.get(ORGANIZATION_KEY);
     String urlStr = (String) properties.get(BRIGHTSPACE_URL);
     String systemUserId = (String) properties.get(BRIGHTSPACE_USER_ID);
@@ -152,9 +153,14 @@ public class BrightspaceUserProviderFactory implements ManagedServiceFactory {
     logger.debug("creating new brightspace user provider for pid={}", pid);
 
 
-    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(pid,
-            new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey), org, cacheSize,
-            cacheExpiration, adminUserName);
+    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(
+        pid,
+        new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey),
+        org,
+        cacheSize,
+        cacheExpiration,
+        adminUserName
+    );
     this.providerRegistrations
             .put(pid, this.bundleContext.registerService(UserProvider.class.getName(), provider, null));
     this.providerRegistrations
--- a/modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
+++ b/modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
@@ -152,19 +152,14 @@ public class BrightspaceUserProviderFactory implements ManagedServiceFactory {
 
     logger.debug("creating new brightspace user provider for pid={}", pid);
 
-
-    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(
-        pid,
-        new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey),
-        org,
-        cacheSize,
-        cacheExpiration,
-        adminUserName
-    );
+    BrightspaceClientImpl clientImpl
+        = new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey);
+    BrightspaceUserProviderInstance provider
+        = new BrightspaceUserProviderInstance(pid, clientImpl, org, cacheSize, cacheExpiration, adminUserName);
     this.providerRegistrations
-            .put(pid, this.bundleContext.registerService(UserProvider.class.getName(), provider, null));
+        .put(pid, this.bundleContext.registerService(UserProvider.class.getName(), provider, null));
     this.providerRegistrations
-            .put(pid, this.bundleContext.registerService(RoleProvider.class.getName(), provider, null));
+        .put(pid, this.bundleContext.registerService(RoleProvider.class.getName(), provider, null));
   }
 
   /**
--- a/modules/userdirectory-canvas/src/main/java/org/opencastproject/userdirectory/canvas/CanvasUserRoleProvider.java
+++ b/modules/userdirectory-canvas/src/main/java/org/opencastproject/userdirectory/canvas/CanvasUserRoleProvider.java
@@ -65,393 +65,377 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 @Component(
-  property = {
-    "service.description=Provides for Canvas users and roles"
-  },
-  immediate = true,
-  service = {UserProvider.class, RoleProvider.class}
+    property = {
+        "service.description=Provides for Canvas users and roles"
+    },
+    immediate = true,
+    service = {UserProvider.class, RoleProvider.class}
 )
 public class CanvasUserRoleProvider implements UserProvider, RoleProvider {
 
-    private static final Logger logger = LoggerFactory.getLogger(CanvasUserRoleProvider.class);
+  private static final Logger logger = LoggerFactory.getLogger(CanvasUserRoleProvider.class);
 
-    private static final String LTI_LEARNER_ROLE = "Learner";
-    private static final String LTI_INSTRUCTOR_ROLE = "Instructor";
-    private static final String PROVIDER_NAME = "canvas";
+  private static final String LTI_LEARNER_ROLE = "Learner";
+  private static final String LTI_INSTRUCTOR_ROLE = "Instructor";
+  private static final String PROVIDER_NAME = "canvas";
 
-    /** The key to look up the organization identifier in the service configuration properties */
-    private static final String ORGANIZATION_KEY = "org.opencastproject.userdirectory.canvas.org";
-    private static final String DEFAULT_ORGANIZATION_VALUE = "mh_default_org";
+  /** The key to look up the organization identifier in the service configuration properties */
+  private static final String ORGANIZATION_KEY = "org.opencastproject.userdirectory.canvas.org";
+  private static final String DEFAULT_ORGANIZATION_VALUE = "mh_default_org";
 
-    /** The key to look up the URL of Canvas instance */
-    private static final String CANVAS_URL_KEY = "org.opencastproject.userdirectory.canvas.url";
-    /** The key to look up the token of the user to invoke RESTful service of Canvas */
-    private static final String CANVAS_USER_TOKEN_KEY = "org.opencastproject.userdirectory.canvas.token";
+  /** The key to look up the URL of Canvas instance */
+  private static final String CANVAS_URL_KEY = "org.opencastproject.userdirectory.canvas.url";
+  /** The key to look up the token of the user to invoke RESTful service of Canvas */
+  private static final String CANVAS_USER_TOKEN_KEY = "org.opencastproject.userdirectory.canvas.token";
 
-    private static final String CACHE_SIZE_KEY = "org.opencastproject.userdirectory.canvas.cache.size";
-    private static final Integer DEFAULT_CACHE_SIZE_VALUE = 1000;
-    private static final String CACHE_EXPIRATION_KEY = "org.opencastproject.userdirectory.canvas.cache.expiration";
-    private static final Integer DEFAULT_CACHE_EXPIRATION_VALUE = 60;
+  private static final String CACHE_SIZE_KEY = "org.opencastproject.userdirectory.canvas.cache.size";
+  private static final Integer DEFAULT_CACHE_SIZE_VALUE = 1000;
+  private static final String CACHE_EXPIRATION_KEY = "org.opencastproject.userdirectory.canvas.cache.expiration";
+  private static final Integer DEFAULT_CACHE_EXPIRATION_VALUE = 60;
 
-    /** The keys to look up which roles in Canvas should be considered as instructor roles */
-    private static final String CANVAS_INSTRUCTOR_ROLES_KEY = "org.opencastproject.userdirectory.canvas.instructor.roles";
-    private static final String DEFAULT_CANVAS_INSTRUCTOR_ROLES = "teacher,ta";
-    /** The keys to look up which users should be ignored */
-    private static final String IGNORED_USERNAMES_KEY = "org.opencastproject.userdirectory.canvas.ignored.usernames";
-    private static final String DEFAULT_INGROED_USERNAMES = "admin,anonymous";
+  /** The keys to look up which roles in Canvas should be considered as instructor roles */
+  private static final String CANVAS_INSTRUCTOR_ROLES_KEY = "org.opencastproject.userdirectory.canvas.instructor.roles";
+  private static final String DEFAULT_CANVAS_INSTRUCTOR_ROLES = "teacher,ta";
+  /** The keys to look up which users should be ignored */
+  private static final String IGNORED_USERNAMES_KEY = "org.opencastproject.userdirectory.canvas.ignored.usernames";
+  private static final String DEFAULT_INGROED_USERNAMES = "admin,anonymous";
 
-    private Organization org;
-    private String url;
-    private String token;
-    private int cacheSize;
-    private int cacheExpiration;
-    private Set<String> instructorRoles;
-    private Set<String> ignoredUsernames;
-    private LoadingCache<String, Object> cache = null;
-    private final Object nullToken = new Object();
+  private Organization org;
+  private String url;
+  private String token;
+  private int cacheSize;
+  private int cacheExpiration;
+  private Set<String> instructorRoles;
+  private Set<String> ignoredUsernames;
+  private LoadingCache<String, Object> cache = null;
+  private final Object nullToken = new Object();
 
-    @Activate
-    public void activate(ComponentContext cc) throws ConfigurationException {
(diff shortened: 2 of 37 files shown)


LukasKalbertodt:
CI failure is not my fault: 

> IO Failed to bind to 0.0.0.0/0.0.0.0:57928

If someone could restart that job, that'd be great!

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
```diff
@@ -152,9 +153,14 @@ public void updated(String pid, Dictionary properties) throws ConfigurationExcep
...
-    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(pid,
-            new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey), org, cacheSize,
-            cacheExpiration, adminUserName);
+    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(
+        pid,
+        new BrightspaceClientImpl(urlStr, applicationId, applicationKey, systemUserId, systemUserKey),
+        org,
+        cacheSize,
+        cacheExpiration,
+        adminUserName
```
KatrinIhler, line 162: Okay, I hate to be that person, but this is an absolute pet peeve of mine. :D I think this takes up way too much space now for not that much benefit.

modules/userdirectory-canvas/src/main/java/org/opencastproject/userdirectory/canvas/CanvasUserRoleProvider.java
```diff
@@ -65,393 +65,402 @@
...
+  private String token;
+  private int cacheSize;
+  private int cacheExpiration;
+  private Set<String> instructorRoles;
+  private Set<String> ignoredUsernames;
+  private LoadingCache<String, Object> cache = null;
+  private final Object nullToken = new Object();
+
+  @Activate
     public void activate(ComponentContext cc) throws ConfigurationException {
```
KatrinIhler: Indentation of method header looks odd here. There are more cases down below.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
KatrinIhler: Okay, I hate to be that person, but this is an absolute pet peeve of mine. :D I think this takes up way too much space now for not that much benefit. I'm also against this for the sake of consistency, since most parts of Opencast don't do it this way.

modules/userdirectory-canvas/src/main/java/org/opencastproject/userdirectory/canvas/CanvasUserRoleProvider.java
```diff
@@ -65,393 +65,402 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(CanvasUserRoleProvider.class);
 
-    private static final String LTI_LEARNER_ROLE = "Learner";
-    private static final String LTI_INSTRUCTOR_ROLE = "Instructor";
-    private static final String PROVIDER_NAME = "canvas";
+  private static final String LTI_LEARNER_ROLE = "Learner";
+  private static final String LTI_INSTRUCTOR_ROLE = "Instructor";
+  private static final String PROVIDER_NAME = "canvas";
 
     /** The key to look up the organization identifier in the service configuration properties */
```
KatrinIhler: Indentation of doc strings is off now.

modules/userdirectory-ldap/src/main/java/org/opencastproject/userdirectory/ldap/OpencastLdapAuthoritiesPopulator.java
```diff
@@ -232,7 +247,11 @@ public OpencastLdapAuthoritiesPopulator(String attributeNames, String prefix, St
           logger.debug("Could not find any attribute named '{}' in user '{}'", attributeName, userData.getDn());
         }
       } catch (ClassCastException e) {
-        logger.error("Specified attribute containing user roles ('{}') was not of expected type String", attributeName, e);
+        logger.error(
```
KatrinIhler, line 250: If the e gets a line to itself, that would make our logging statements a lot more verbose than they are now... I do not like it. :D

modules/userdirectory-sakai/src/main/java/org/opencastproject/userdirectory/sakai/SakaiUserProviderInstance.java
```diff
@@ -338,37 +355,37 @@ protected User loadUserFromSakai(String userName) {
...
+      logger.warn("Invalid regular expression for user pattern {} - disabling checks", userPattern);
+      userPattern = null;
+    }
 
-      int code;
+    int code;
 
-      try {
+    try {
           // This webservice does not require authentication
```
KatrinIhler: Comments are again not indented with the code.

modules/userdirectory-sakai/src/main/java/org/opencastproject/userdirectory/sakai/SakaiUserProviderInstance.java
```diff
@@ -598,92 +622,102 @@ public long countUsers() {
 
   // RoleProvider methods
 
-   @Override
+  @Override
    public List<Role> getRolesForUser(String userName) {
```
KatrinIhler: Indentation of method headers not correct.

modules/userdirectory-sakai/src/main/java/org/opencastproject/userdirectory/sakai/SakaiUserProviderInstance.java
```diff
@@ -598,92 +622,102 @@ public long countUsers() {
...
-       roles.add(new JaxbRole(query, jaxbOrganization, "Sakai Site Role", Role.Type.EXTERNAL));
-     } else {
+      roles.add(new JaxbRole(query, jaxbOrganization, "Sakai Site Role", Role.Type.EXTERNAL));
+    } else {
        // Site ID - return both roles
-       roles.add(new JaxbRole(sakaiSite + "_" + LTI_INSTRUCTOR_ROLE, jaxbOrganization, "Sakai Site Instructor Role", Role.Type.EXTERNAL));
-       roles.add(new JaxbRole(sakaiSite + "_" + LTI_LEARNER_ROLE, jaxbOrganization, "Sakai Site Learner Role", Role.Type.EXTERNAL));
-     }
+      roles.add(new JaxbRole(
+          sakaiSite + "_" + LTI_INSTRUCTOR_ROLE,
```
KatrinIhler, line 707: Same complaint as usual - more space used for little benefit, yadda yadda.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserDirectoryPersistenceUtil.java
```diff
@@ -713,14 +736,15 @@ public static JpaGroup findGroupByRole(String role, String orgId, EntityManagerF
...
+      if (em != null) {
         em.close();
+      }
     }
   }
 
 
   public static void removeGroup(String groupId, String orgId, EntityManagerFactory emf) throws NotFoundException,
-  Exception {
+          Exception {
```
KatrinIhler: Isn't that a bit much indentation? Alternatively maybe we can put the whole throws clause into a new line? Same below.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/api/GroupRoleProvider.java
```diff
@@ -40,21 +40,21 @@
...
      * {@inheritDoc}
      *
      * @see org.opencastproject.security.api.RoleProvider#getRolesForUser(String)
      */
-    List<Role> getRolesForGroup(String groupName);
+  List<Role> getRolesForGroup(String groupName);
 
     /**
      * {@inheritDoc}
      *
```
KatrinIhler: Comment indentation is off again.

KatrinIhler (dismissed):
I think now I will see white space changes highlighted whenever I close my eyes...

Anyway. :D Mostly this looks fine. There are some things that were missed when changing the indentation (mostly doc strings and sometimes method headers).

The rest mostly comes down to preference. I really do not like spreading out parameters of method headers or method calls out over so many lines. I think the benefits, while there, are not really worth all that "wasted" space. It's also not at all consistent with the rest of Opencast (and I know Opencast itself is not consistent here - or anywhere, really - but I'd still argue that this is more unusual.)

I know other people think differently (just argued with Julian about this :D), so if we can't agree, maybe we should ask the rest of the devs what they think.

I guess when breaking long lines I would also break them later than you do (if possible), because I feel like this makes it easier to see that this is supposed to be one line vs two independent lines? But that's not really a hill I'm willing to die on.

modules/userdirectory-ldap/src/main/java/org/opencastproject/userdirectory/ldap/OpencastLdapAuthoritiesPopulator.java
LukasKalbertodt: I joined a couple lines here, so that should be fixed.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserDirectoryPersistenceUtil.java
LukasKalbertodt: Well. When I started this whole adventure, I check how much throw statements were usually indented in Opencast, and the most common indentation was actually 8. So I went with that (and people approved that in the PR that introduced the rules). 

But I agree that this seems weird here. I changed it so that the whole `throws` part with both exception is in a new line. I think it looks better now.

LukasKalbertodt:
I pushed. All inline comments that I reacted with :+1: to should be fixed (and lots more).

---

> There are some things that were missed when changing the indentation (mostly doc strings and sometimes method headers).

Thanks a ton for catching those! I had too much faith in checkstyle, which is apparently not warranted. By now I'm pretty disappointed by the quality of checkstyle. It has tons of false positives and false negatives and is just uhg. The indentation I always fixed with a simple script, but since checkstyle is not to be trusted, I have to do it manually from now on.

> [...] so many lines. I think the benefits, while there, are not really worth all that "wasted" space. 

Could you explain the actual disadvantages of "wasted space"? I guess you are not talking about the few additional `\n` bytes required to store the bytes. So you must be talking about some other disadvantage that is caused by having more lines. "more lines" is not a disadvantage in itself as far as I can tell. So I'm curious what you see as the actual problem with that?

Some examples you left line comments on I might even agree with you. That is, for call expressions. But this is also a consequence of methods in this codebase often having *tons* of parameters. Which, as far as I am concerned, is a bad thing in itself.

That said, as I stated from the very beginning, I'm just here to do the dummy work and will stick to what the community wants. 

> It's also not at all consistent with the rest of Opencast (and I know Opencast itself is not consistent here - or anywhere, really - but I'd still argue that this is more unusual.)

I'm more than half done with the checkstyle reformat, so by now there is lots of code which, at least for method signatures and annotations, uses the "one line per item if it doesn't fit into one line" rule. 

> [...] maybe we should ask the rest of the devs what they think.

Well, I fairly explicitly requested feedback on the first few PRs I've created of this kind. The idea was that there, we can have all that discussion in the beginning and all subsequent PRs don't really spawn discussions. I mentioned that in the technical meeting, people noticed the PRs and commented on them.

I'm not saying all remaining PRs should just be merged or that no one should comment on them, but I'm certainly not going back and changing all the other stuff again.

modules/userdirectory-brightspace/src/main/java/org/opencastproject/userdirectory/brightspace/BrightspaceUserProviderFactory.java
LukasKalbertodt: In this specific example I would tend to agree with you and I have no problem with joining a few lines here. How would you format this whole statement?

gregorydlogan: I would create `BrightspaceClientImpl` prior to `provider`, then wrap similar to how it was previously wrapped.

In general I agree with @KatrinIhler, although I think the userdirectories is particular are egregiously bad in their use of excessively long and similar method headers.  Some of the cases I think the spacing [improves](https://github.com/opencast/opencast/pull/2892/files#diff-d2dce0faedd65be61b36882b6fcaa3c825abc17fe7ec0cd35d4746e44d97211bR329) things, some it definitely [doesn't](https://github.com/opencast/opencast/pull/2892/files#diff-683c0474e824c7a4fe47ea7cc59ad3006b0581a2cd897f2c4465ea0e16683018R84) - sadly 'neither of the above' isn't a helpful comment.  If I had to pick I would probably break it out line by line as it currently stands in this PR.  

My rule of thumb is if the variables names (yes, really) are short and different enough that it parses well on fewer lines I tend to keep them on those lines.  In the case of something like this I would do
```
    BrightspaceUserProviderInstance provider = new BrightspaceUserProviderInstance(pid, BSCI, org, cacheSize,
        cacheExpiration, adminUserName);
```
which has the double benefit of breaking `cacheSize`, and `cacheExpiration` apart.  Or, rename those two variables and remove the `cache` prefix, at which point it probably fits all on one line :D

modules/userdirectory-ldap/src/main/java/org/opencastproject/userdirectory/ldap/OpencastLdapAuthoritiesPopulator.java
```diff
@@ -354,8 +371,11 @@ private void addAuthorities(Set<GrantedAuthority> authorities, final String[] va
     if (values != null) {
       Organization org = securityService.getOrganization();
       if (!organization.equals(org)) {
-        throw new SecurityException(String.format("Current request belongs to the organization \"%s\". Expected \"%s\"",
-                org.getId(), organization.getId()));
+        throw new SecurityException(String.format(
+            "Current request belongs to the organization \"%s\". Expected \"%s\"",
+            org.getId(),
```
gregorydlogan: I would join this line and the next.

I would also like to see `org` and `organization` be renamed so they're clearer.  Ugh.  That being said, that particular change is out of scope.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserAndRoleDirectoryServiceImpl.java
```diff
@@ -336,17 +352,30 @@ private User loadUser(Tuple<String, String> orgUser) {
     for (Role role : roles) {
       if (Role.Type.EXTERNAL_GROUP.equals(role.getType())) {
         // Load roles granted to this group
-        logger.debug("Resolving transitive roles for user {} from external group {}", user.getUsername(), role.getName());
+        logger.debug(
+            "Resolving transitive roles for user {} from external group {}",
+            user.getUsername(),
```
gregorydlogan: I would join these.

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/UserAndRoleDirectoryServiceImpl.java
```diff
@@ -336,17 +352,30 @@ private User loadUser(Tuple<String, String> orgUser) {
...
           if (roleProvider instanceof GroupProvider) {
             List<Role> groupRoles = ((GroupProvider) roleProvider).getRolesForGroup(role.getName());
             if (groupRoles != null) {
               for (Role groupRole : groupRoles) {
                 derivedRoles.add(JaxbRole.fromRole(groupRole));
               }
-              logger.debug("Adding {} derived role(s) for user {} from internal group {}", derivedRoles.size(), user.getUsername(), role.getName());
+              logger.debug(
+                  "Adding {} derived role(s) for user {} from internal group {}",
+                  derivedRoles.size(),
```
gregorydlogan: I would join these as well.  In general, the form
```
debug(
"my logging statement with random includes {} {} {}",
var1, var2, var3);
```
parses best to my eyes, although I get that autoformatting that is probably hard :D

gregorydlogan (approved):
In general I approve this changes, comments below.

Regarding the "wasted space" issue, IMO it's a personal taste thing.  In theory your IDE should abstract that away, but a lot of time I'm actually working in (unassisted) vim, which makes removing things like that harder.  I don't mind the method headers - I don't *like* it, but I also don't have a better solution.  What really hurts is in mid-method, where having the whole thing in your head is important.  In those cases it's a fine line between wasting space and drawing attention to a (probably vastly overcomplicated) method call.  That ends up being author's discretion, and I really can't see setting a reasonable, measurable threshold for when we space vs when we don't.

What you're doing is IMO correct, keep up with the good work!

LukasKalbertodt:
> What really hurts is in mid-method, where having the whole thing in your head is important. In those cases it's a fine line between wasting space and drawing attention to a (probably vastly overcomplicated) method call.

That's a good point. But yeah, hard to find good precise rules when to do what.

---

I pushed a commit fixing all your remarks, I think. For one, I did it slightly differently than suggested. Feel free to add a comment again.

gregorydlogan:
Much better, thank you @LukasKalbertodt.  @KatrinIhler you're still marked as wanting changes, do you still want them after this last commit?

KatrinIhler:
Greg explained well what I meant (and of course it wasn't the waste of bytes :D):

There should be some balance between spreading out complicated code so it's easier to understand, and drawing too much attention to verbose, but simple code that it doesn't deserve, which means I have to scroll more when trying to understand the higher structure of a method or class. But obviously, where that balance lies is often a matter of preferences, so you can't make rules for that.

I also obviously don't want you to go back and change everything again, I'm not that cruel. :D And you're right that I should've complained sooner, but going through these PRs takes so much time that I procrastinated it too long...

So yeah, I saw that you made some changes, and I don't want to drag this out longer than I have to (especially considering that I'm going on a three-week vacation starting tomorrow...), so I think I can live with the rest. :)

LukasKalbertodt:
> drawing too much attention to verbose, but simple code that it doesn't deserve

Yeah I guess that's a good point and explains the problem with "too many lines". 

> I'm not that cruel. :D

I know :wink: 

> going through these PRs takes so much time

Absolutely! That's why I also didn't expect people to review these, except the first couple. And honestly, most of these PRs were merged without review, I think. They were just tested to still compile and work. 

And at the risk of repeating myself: my PRs' main purpose is to get rid of the checkstyle violations → such that we can remove the exception for the corresponding module → such that in the future, those violations cannot be reintroduced. My PRs do not claim to have the best code style nor are they the "last word" on code style. I very much welcome other code style improvement PRs that change the code I touched. But as long as my PRs prevent future bad style violations and as long as they are at least an improvement over the previous code, then they serve their purpose as far as I am concerned.

lkiesow:
> my PRs' main purpose is to get rid of the checkstyle violations → such that we can remove the exception for the corresponding module → such that in the future, those violations cannot be reintroduced. My PRs do not claim to have the best code style nor are they the "last word" on code style. 

I think that absolutely makes sense. I've reviewed a few of your pull requests, and I could have complained about more things. But in most cases these were problems with the existing code, and the style changes just highlighted that once more. I don't think it's fair to say you need to fix everything just because you touch something ;-)

Let's get to a common style first and let's make an effort to improving code when we are actually working on it, rather than just auto-formatting it.

PR #2897, opened 2021-08-03
LTI upload / edit form styling, adding some sensible padding and margins

kristofkeppens:
The upload / edit form is cramped together and not easy to read. This adds some sensible padding and margins to the different form elements.


lkiesow:
Already looks way better.
For anyone curious how this changes the interface:

Screenshot from 2021-08-10 23-24-00

PR #2899, opened 2021-08-04
Hello World Workflow Operation

lkiesow:
This patch provides a very simple example of how a workflow operation
works and can be a starting point for new developments.

Like the other hello-world modules, this is intentionally not included
in the Opencast distributions and thus not listed in the documentation
since people cannot actually use it.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/hello-world-workflowoperation/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+            <Build-Number>${buildNumber}</Build-Number>
+            <Export-Package>
+              org.opencastproject.workflow.handler.logging;version=${project.version}
```
KatrinIhler: I don't think this is supposed to be here.

KatrinIhler (changes requested):
Looks good as a whole. :) Just found one think that's most likely a copy-paste error.

modules/hello-world-workflowoperation/pom.xml
lkiesow: fixed

KatrinIhler (approved):

PR #2906, opened 2021-08-06
Retry Documentation Deployment

lkiesow:
While concurrent build within a branch cancel each other to not run into
problems, deployments can still fail when merges happen in quick
succession on different branches. This patch will mitigate the issue by
rebasing to the updated code in case of conflicts when pushing the
updated docs.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -99,4 +99,9 @@ jobs:
     - name: push updates
       working-directory: docs.opencast.org/
       run: |
+        set -e
+        if ! git push origin gh-pages; then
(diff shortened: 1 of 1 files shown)


mtneug:
If this happens, doesn't that mean that the docs are potentially inconsistent? If two commits happen in quick succession and both also change the docs, the slowest one would "win", but there can be some files remaining from the faster commit, can't they?

lkiesow:
Builds from the same branch cancel each other so that there always the
newest will win. This can only happen with builds from different
branches which will always change files in different paths.

PR #2913, opened 2021-08-09
Fix series endpoint update series method by skipping the ACL update if the parameter is empty

geichelberger:
This pull request again allows the ACL parameter to be optional.
The commit https://github.com/opencast/opencast/commit/d260846aa1269c422d464eb1b14d5c6ce1aad04c changed that behavior and required the ACL parameter, which leads to a broken SeriesServiceRemote update series function.

Endpoint POST /
https://develop.opencast.org/docs.html?path=/series#updateSeries-5


lkiesow (approved):
Woks as expected. Will merge after the technical meeting if no one else objects.

…though in general, I think it's possible to set Series with no ACL and I find it curious that it's not possible to remove an ACL. But that's a different issue and nothing to discuss here.

PR #2914, opened 2021-08-09
Metadata to ACL Operation

lkiesow:
This patch adds a metadata-to-acl workflow operation which can be used
to automatically add access rights for users mentioned in metadata
fields to the access control list.

The operation works like this:

1. Get values for configured field from the episode Dublin Core catalog
2. Look up if a user with this username exists
3. Add read and write permissions for this user

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


Arnei (approved):
Works and looks good.

Might want to emphasize that this looks up a users `username` in the docs, cause if you try to add users to fields such as presenter(s) or contributor(s) in the Admin UI, they are added by the users `name`.

lkiesow:
Added another sentence mentioning this. Now it's in both the general description and in the description how this works.

The admin interface's behavior is actually pretty strange, since it presents the username to the user but seems to replace it in the backend so that the metadata value has changed when you next come back to the interface. But, I guess, that's a different issue.

Arnei (approved):

PR #2920, opened 2021-08-10
Add search form input field epFrom on sort if url parameter is set

### LINKED ISSUE #2919: When sorting the episodes of a series, the filtering by the selected series is lost
**Describe the bug**
In the media module, click on the "Series" tab and select one of the series available. Then, the episodes of this series are shown.
When selecting one of the sorting options from the "Sort" dropdown on the top right, all available episodes are shown, not only the ones that belong to the currently selected series.

**To Reproduce**
_Steps to reproduce the behavior:_
1. Go to the media module and click on the "Series" tab
2. Select one of the series available to see its episodes
4. Use the "Sort" dropdown on the top right and select any option for sorting

_Expected behaviour:_
The episodes of the **series selected** are shown in a sorted order.

_Current behaviour:_
All available episodes of **all series** on media module are shown in a sorted order.

**Environment Information**
- OS: Windows 10 20H2
- Browser: Chrome 92.0.4515.131 (Official Build) (64-bit) || Firefox 90.0.2 (64-Bit)

geichelberger:
With this pull request, the media module remembers whether episodes of a certain series have been filtered and takes this into account when sorting.

Before this fix, applying the sort criterion removed the epFrom filter parameter.

Fixes #2919

### DIFF
--- a/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
+++ b/modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
@@ -541,6 +541,12 @@ function($, bootbox, _, alertify, jsyaml) {
     });
 
     $($oc_sort_dropdown).on('change', function() {
+      let epFrom = GetURLParameter('epFrom');
+      if (epFrom) {
+        $('#oc-search-form .form-group').append(
+          '<input type=\'hidden\' name=\'epFrom\' value=\'' + _.escape(epFrom) + '\' />'
+        );
+      }
       $($oc_search_form).submit();
     });
 


ziegenberg:
This would be nice to have in the next point release.

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
```diff
@@ -541,6 +541,12 @@ function($, bootbox, _, alertify, jsyaml) {
     });
 
     $($oc_sort_dropdown).on('change', function() {
+      let epFrom = GetURLParameter('epFrom');
+      if (epFrom) {
+        $('#oc-search-form .form-group').append(
+          '<input type=\'hidden\' name=\'epFrom\' value=\'' + epFrom + '\' />'
```
lkiesow: This allows adding unescaped user input. This means that I could generate malicious links and send them to users, which then have my code executed if they click on it. Iirc, there should already be an escape method somewhere.

lkiesow (changes requested):

modules/engage-ui/src/main/resources/ui/js/app/engage-ui.js
ziegenberg: `_.escape` should do the trick, I guess...

lkiesow: I think that's the method I was remembering

geichelberger: 🤦 OMG! Sorry, that should not happen. Lesson learned. After a hard day of meetings, you shouldn't fix a bug in a hurry.
Is there a way to restart the CI pipeline? The build failed on a module that has no modifications.

lkiesow: Ignore that issue for now. Merging `r/9.x` into `r/10.x` just broke  something. Will be fixed as soon as the tests on #2922  succeed and that one gets merged.

lkiesow (approved):

PR #2527, opened 2021-04-19
Removes Safari from Karma admin-ui-frontend tests.

mliradelc:
This commit removes safari to be use in the admin-ui Karma tests.
 Solves the issue when Opencast is build in macOS fails the admin-ui-frontend tests

This change adds the requirement to have Firefox or Chrome installed when Opencast is
build from source in macOS

### Your pull request should…

* [x] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] be against the correct branch (features can only go into develop)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #2720, opened 2021-06-12
Fixed admin interface tests in certain timezones

lkiesow:
This patch fixes an admin interface unit test which fails in certain timezones
with active daylight saving time (e.g. CEST).

The error is:

```
       Chrome Headless 93.0.4535.3 (Linux x86_64) adminNg.directives.adminNgTableFilter #selectFilterPeriodValue with both from- and to dates sets the time period filter value (accounting for localized day) FAILED
[INFO]  Expected spy put to have been called with:
[INFO]    [ 'filter', 'furniture', undefined, '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z' ]
[INFO]  but actual calls were:
[INFO]    [ 'filter', 'furniture', undefined, '2014-12-31T23:00:00.001Z/2015-01-02T22:59:59.999Z' ].
[INFO]
[INFO]  Call 0:
[INFO]    Expected $[3] = '2014-12-31T23:00:00.001Z/2015-01-02T22:59:59.999Z' to equal '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z'.
[INFO]  Error: Expected spy put to have been called with:
[INFO]    [ 'filter', 'furniture', undefined, '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z' ]
[INFO]  but actual calls were:
[INFO]    [ 'filter', 'furniture', undefined, '2014-12-31T23:00:00.001Z/2015-01-02T22:59:59.999Z' ].
[INFO]  Call 0:
[INFO]    Expected $[3] = '2014-12-31T23:00:00.001Z/2015-01-02T22:59:59.999Z' to equal '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z'.
[INFO]      at <Jasmine>
[INFO]      at UserContext.<anonymous> (test/unit/shared/directives/tableFilterDirectiveSpec.js:206:37)
[INFO]      at <Jasmine>
       Firefox 89.0 (Fedora 0.0.0) adminNg.directives.adminNgTableFilter #selectFilterPeriodValue with both from- and to dates sets the time period filter value (accounting for localized day) FAILED
[INFO]  Expected spy put to have been called with:
[INFO]    [ 'filter', 'furniture', undefined, '2014-12-30T23:00:00.001Z/2015-01-02T22:59:59.999Z' ]
[INFO]  but actual calls were:
[...]

PR #2925, opened 2021-08-10
Test only with Firefox or Chrome

lkiesow:
As discussed at today's technical meeting this patch closes #2527 and
replaces it with a patch limiting the tests to Firefox or Chrome so we
know that we actually run in a somewhat known and safe environment.

---

Note that `r/9.x` seems to have the same issue I fixed in #2720 for `r/10.x`.
Make sure to build with something like this if you test this locally:

```
TZ="America/Regina" mvn clean install -Pnone -DtrimStackTrace=false
```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/test/karma.conf.js
+++ b/modules/admin-ui-frontend/test/karma.conf.js
@@ -59,26 +59,18 @@ module.exports = function (config) {
             enabled: true,
             usePhantomJS: false,
             preferHeadless: true,
-            // post processing of browsers list
-            // here you can edit the list of browsers used by karma
-            postDetection: function(availableBrowsers) {
-                var result = availableBrowsers;
-                /* Leaving this commented out as an example.
-                   If we ever want to disable an installed browser (c.f: IE) we can exclude it like this
-                //Remove PhantomJS if another browser has been detected
-                if (availableBrowsers.length > 1 && availableBrowsers.indexOf('PhantomJS')>-1) {
-                    var i = result.indexOf('PhantomJS');
-                    if (i !== -1) {
-                        result.splice(i, 1);
-                    }
-                }
-		*/
-		if (availableBrowsers.length < 1) {
+            // limit the list of browsers used by karma
+            postDetection: (browsers) => {
+                const allowed = new Set(['ChromeHeadless', 'FirefoxHeadless'])
+                console.info('Detected browsers: ' + browsers);
+                browsers = browsers.filter(browser => allowed.has(browser));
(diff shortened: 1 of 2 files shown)


mtneug:
`FirefoxHeadless` on macOS fails for me, but those seem to be local related (unexpected date format etc.). Configuring Firefox to prefer English didn't solve the issues. `ChromeHeadless` works though. Safari is no longer used.

gregorydlogan:
It's worth testing both `TZ="America/Regina` and `TZ="America/Boston"` since the former does not observe DST, whereas the latter does.  Because we *all* love TZ test regressions!

This works for me, but I'm also on Debian.

modules/admin-ui-frontend/test/karma.conf.js
```diff
@@ -59,26 +59,18 @@ module.exports = function (config) {
             enabled: true,
             usePhantomJS: false,
             preferHeadless: true,
-            // post processing of browsers list
-            // here you can edit the list of browsers used by karma
-            postDetection: function(availableBrowsers) {
```
mliradelc, line 64: I prefer the function `availableBrowsers` and exclude the ones that they have known problems. Set by stone the browsers can be troublesome in the future.

mliradelc (commented):
Tried with my Mac,  same problems that @mtneug has. In my case, macOS is set on English

modules/admin-ui-frontend/test/karma.conf.js
lkiesow: Fine with me. Closing this pull request then.

lkiesow:
> It's worth testing both `TZ="America/Regina` and `TZ="America/Boston"` since the former does not observe DST, whereas the latter does.  Because we *all* love TZ test regressions!

Actually, no. There is still a timezone related build problem in `r/9.x` as I mentioned. It's fixed in `r/10.x`. Tests will thus fail. But that has nothing to do with what this pull request tried to do.

mtneug:
Why close this?

lkiesow:
Because @mliradelc disagrees with these changes, which means we can also use his pull request once it's fixed. For me, this was just an attempt at getting this done. I don't have this problem and won't spend a lot of effort on this. If @mliradelc opposes the solution and wants to solve this differently, that's fine with me.

mliradelc:
Reopened after discussion in the Chat, see #2527

mliradelc:
One question, should be this against 9.x instead of develop?

PR #2928, opened 2021-08-13
Added config option for disabling the embed-code button 

jduehring:
This PR adds a configuration option for disabling the embed-code button in the admin ui. 

The configuration can be changed by setting a flag on the `prop.adminui.embed_button_display` option in `etc/org.opencastproject.organization-mh_default_org.cfg` . If it's set to any other value than `false` or not set at all, it defaults to `true`.
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventsController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventsController.js
@@ -52,10 +52,7 @@ angular.module('adminNg.controllers')
         $scope.editorUrl = '#!/events/events/$id/tools/editor';
       }
       if (angular.isDefined(user.org.properties['adminui.embed_button_display'])) {
-        var embedButtonConfig = user.org.properties['adminui.embed_button_display'];
-        if (embedButtonConfig === 'false') {
-          $scope.showEmbedButtons = embedButtonConfig;
-        }
+        $scope.showEmbedButtons = user.org.properties['adminui.embed_button_display'] !== 'false';
       }
     }).catch(angular.noop);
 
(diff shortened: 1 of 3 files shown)


modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventsController.js
```diff
@@ -49,6 +51,12 @@ angular.module('adminNg.controllers')
       if (!$scope.editorUrl) {
         $scope.editorUrl = '#!/events/events/$id/tools/editor';
       }
+      if (angular.isDefined(user.org.properties['adminui.embed_button_display'])) {
+        var embedButtonConfig = user.org.properties['adminui.embed_button_display'];
+        if (embedButtonConfig === 'false') {
+          $scope.showEmbedButtons = embedButtonConfig;
+        }
```
lkiesow: You set `showEmbedButtons` to the boolean value `true` or the string `"false"` which may work (not actually sure) but is confusing and may break eventually. You could use something like this to always set this to a boolean value:

```suggestion
        $scope.showEmbedButtons = user.org.properties['adminui.embed_button_display'] !== 'false';
```

jduehring: Makes sense, updated the code with your suggested change.

mtneug:
You can already hide this icon by not assigning the `ROLE_UI_EVENTS_EMBEDDING_CODE_VIEW` role to a user.

lkiesow:
You are right:

https://github.com/opencast/opencast/blob/52d7d8ffc4aa9de59066a30592e20a7525f48c76/modules/admin-ui-frontend/app/scripts/modules/events/partials/eventActionsCell.html#L69

Is there still a reason to merge this? Or can we then just close this?

mtneug:
Closing in favour of using `ROLE_UI_EVENTS_EMBEDDING_CODE_VIEW`. Feel free to open it again if you disagree.

Issue #2135, opened 2020-12-11
Use proxy_cookie_flags instead of hacked proxy_cookie_path


mtneug:
**Describe the bug**

The documentation recommends the use of `proxy_cookie_path` for setting cookie flags. Since nginx 1.19.3 the new configuration key `proxy_cookie_flags` exists that is meant or setting cookie flags. I don't know if we should migrate now since 1.19.3 is a relatively new version and people probably still use 1.18 / 1.17.

gregorydlogan:
We should probably document both, since the new behaviour will be the right way going forward but we'll have adopters on the old version for a very long time still.

lkiesow:
Pull request #2942 has essentially fixed this, but left a comment for older versions we should remove eventually. My suggestion is to remove it in the Opencast 12 documentation, leaving adopters about another year to upgrade. Let's leave this open to remind us.

PR #2942, opened 2021-08-18
Update Nginx example regarding proxy_cookie_path

jojoob:
Recommend proxy_cookie_flags for Nginx >=1.19.3
but mention the proxy_cookie_path workaround for Nginx <1.19.3

This request addresses #2135

### DIFF
--- a/docs/guides/admin/docs/configuration/https/nginx.md
+++ b/docs/guides/admin/docs/configuration/https/nginx.md
@@ -113,7 +113,10 @@ http {
             proxy_redirect          http://$host https://$host;
 
             # Make sure to serve cookies only via secure connections.
-            proxy_cookie_path / "/; HTTPOnly; Secure";
+            proxy_cookie_flags ~ secure httponly;
+            # When using Nginx <1.19.3 replace the above 'proxy_cookie_flags' line
+            # with the (uncommented) 'proxy_cookie_path' line below.
+            #proxy_cookie_path / "/; HTTPOnly; Secure";
 
             # Depending on your integration, you may also want to allow cookies
             # to be used on other sites. In that case, use this instead:


docs/guides/admin/docs/configuration/https/nginx.md
```diff
@@ -113,7 +113,10 @@ http {
             proxy_redirect          http://$host https://$host;
 
             # Make sure to serve cookies only via secure connections.
-            proxy_cookie_path / "/; HTTPOnly; Secure";
+            proxy_cookie_flags ~ secure httponly;
```
lkiesow, line 116: Not opposing this but out of interest: What is the advantage of this over `proxy_cookie_path`?

docs/guides/admin/docs/configuration/https/nginx.md
```diff
@@ -113,7 +113,10 @@ http {
...
+            # When using Nginx <1.19.3 replace the above 'proxy_cookie_flags' line
```
lkiesow, line 117: When using old Nginx versions… update? Do we need a fallback for older versions? We don't have them at any of the other flags.

jojoob: I totally agree. I added this because of the [concerns](https://github.com/opencast/opencast/issues/2135#issuecomment-743192297) raised by @gregorydlogan

docs/guides/admin/docs/configuration/https/nginx.md
jojoob: It is simply a cleaner way of setting cookie flags instead of exploiting the proxy_cookie_path directive. It helps to avoid misconfiguration. Appending flags to the path string is undocumented and could eg. lead to same flags repeated if previously present (`path=/; Secure` -> `path=/; Secure; Secure`.

docs/guides/admin/docs/configuration/https/nginx.md
lkiesow: Fine with me. Then, let's leave this for now and not close the issue, but add a task to remove the comment with Opencast 12. That gives people almost another year.

lkiesow (approved):
While Apereo has not yet filed you ICLA, they have received it. Therefore, let's get this merged.

PR #2946, opened 2021-08-19
Rebuild ExternalIndex while triggering an index recreation

Gao-Jun:
While invoke index rebuild API with  `/admin-ng/index/createIndex`, only `adminUISearchIndex` is cleaned and rebuilt.

This PR make `externalIndex` cleaned and rebuilt as the same way as `adminUISearchIndex`.

If `externalInbox` is not processed in index creation process, some search results will be conflict with each other from different endpoint. This will affect some scenes like batch update data or clean up data.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -182,6 +182,11 @@
     </dependency>
     <dependency>
       <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-external-api-index</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
       <artifactId>opencast-list-providers-service</artifactId>
       <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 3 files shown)


gregorydlogan:
This will, I imagine, conflict with #2857 which will likely go in first.  It might be worth your time to look at that PR and see if it solves this issue already.  If not it might make sense to base you work from that branch so that you don't have to come back to as many merge conflicts :D

lkiesow:
Isn't that what [`/api/recreateIndex`](https://develop.opencast.org/docs.html?path=/api#recreateIndex-2) is for?
If I understood @KatrinIhler correctly, this would become a single index with Opencast 11 anyway, but for now, we probably do not want to mix them, do we?

Gao-Jun:
@lkiesow My mistake, this patch affect 3 endpoints `clearIndex`, `recreateIndex/{service}`, `recreateIndex`.
Besides, I looked into #2857, it's a huge PR with a very good idea.
Yet, this PR still avoid some misleading result in migration and test environment.
So, I think this PR could be merged, and after #2857 is merged, only a few lines should be modified to use one single index.

Gao-Jun:
Sorry, my fault.
Today, I found that, the `recreateIndex` for `adminUISearchIndex` is located at `/admin-ng/index/recreateIndex`.
And, the `recrateIndex` for `externalIndex` is located at `/api/recreateIndex`.
So, just as the API name metioned, there are two different endpoint to handle different index.
I think this PR should not be merged.

PR #2968, opened 2021-08-27
Fix isActive check for workflows

lkiesow:
This patch fixes the `isActive` check for workflows from the workflow
utils which should be the inverse of `isTerminated` but was actually
missing the `failing` state.

This also marks this method as deprecated since with the existence of
`status.isTerminated()`, there is no real benefit in having a separate
utils method for this check.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowUtil.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowUtil.java
@@ -34,10 +34,9 @@ public final class WorkflowUtil {
    *          The workflow state to check.
    * @return True if the workflow is currently active, not stopped or failed.
    */
+  @Deprecated
   public static boolean isActive(WorkflowState workflowState) {
-    return WorkflowState.INSTANTIATED.equals(workflowState)
-            || WorkflowState.RUNNING.equals(workflowState)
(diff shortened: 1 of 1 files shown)


modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowUtil.java
```diff
@@ -34,10 +34,9 @@ private WorkflowUtil() {
...
    */
+  @Deprecated
   public static boolean isActive(WorkflowState workflowState) {
-    return WorkflowState.INSTANTIATED.equals(workflowState)
-            || WorkflowState.RUNNING.equals(workflowState)
-            || WorkflowState.PAUSED.equals(workflowState);
+    return !workflowState.isTerminated();
   }
 
   /**
```
mtneug, line 42: The `#isActive(String)` method needs to be updated as well.

mtneug (changes requested):

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowUtil.java
lkiesow: done.

PR #2994, opened 2021-09-09
Provide a minimal "No Op" workflow for debugging and maintenance

JulianKniephoff:
This can be useful sometimes, for example when parts of our indices
concerning workflows get out of sync with the database, and you see
inconsistent information about the state of a workflow in the events
table vs. the event details. In a case like that, you can just run
some workflow using the `/workflow/start` API to force Opencast
to go through all the proper channels to update the database
and the indices in a (hopefully) consistent manner.

Having a workflow for that purpose handy is thus often useful,
so admins don't have to have it floating around somewhere external,
or come up with the definition from scratch every time. You can just
specify `noop` as a parameter to the `/workflow/start` endpoint.
Also, this workflow intentionally has no `tags` so that it doesn't
show up anywhere in the UI to confuse people.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/workflows/noop.xml
+++ b/etc/workflows/noop.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<definition xmlns="http://workflow.opencastproject.org">
+  <id>noop</id>
+  <title>Do nothing (mostly for debugging and maintenance purposes)</title>
+  <operations>
+    <!-- We need at least one operation for this to be considered a valid workflow definition -->
+    <operation id="defaults" />
+  </operations>
+</definition>


gregorydlogan (approved):
I like it.  Thanks for making it hidden by default :D

Arnei:
Let me play devil's advocate for a bit here. I'm against cluttering the workflow directory with very specific workflows for the sake admin convenience. This workflow in particular is so simple that the people who decide it is needed on their installation can create it themselves no problem. Furthermore, if events are broken so often that it should be included in every installation, doesn't this mean there is an underlying issue that should be fixed, rather than fighting symptoms with this workflow?

It might be more useful to have this mentioned as a troubleshooting method in the docs somewhere, rather than simply adding a workflow for people "in the know".

mtneug:
[We have such a workflow](https://zivgitlab.uni-muenster.de/educast-nrw/opencast/workflows/-/blob/5cf1daf26729e7f6fc26537afcef687a7d604340/src/workflows/educast-main-fix-no-op.xml) for some years now and don't really use it (we also [have one for failing a recording](https://zivgitlab.uni-muenster.de/educast-nrw/opencast/workflows/-/blob/5cf1daf26729e7f6fc26537afcef687a7d604340/src/workflows/educast-main-fix-fail.xml)). I'm not really against adding it upstream, but we may also choose to put it in <https://github.com/opencast/community-workflows>.

JulianKniephoff:
> This workflow in particular is so simple that the people who decide it is needed on their installation can create it themselves no problem.

Sure, but it's annoying. And thinking about how often I heard "just run a dummy workflow on the event" here at ELAN in the last couple of months that also becomes a matter of scale. And of course these are "first world problems", but so is having a clean workflow folder, IMO. :thinking: 

> Furthermore, if events are broken so often that it should be included in every installation, doesn't this mean there is an underlying issue that should be fixed, rather than fighting symptoms with this workflow?

Yes, but that's disproportionately harder, and probably won't happen anytime soon. This workflow is a practical solution to a real problem we have and that probably won't go away that easily., even though we all agree that it should, and that this is not the best solution.

Like, finding these inconsistency issues is hard. Yes, they are frequent enough to be annoying, but they are often not reproducible enough to be amenable to the usual debugging techniques. And apart from the big crowd fundings no client is going to pay us to wildly try to recreate them and/or read and understand the entirety of the index service to get at them logically, when we can just run a workflow to fix their broken event.

> It might be more useful to have this mentioned as a troubleshooting method in the docs somewhere, rather than simply adding a workflow for people "in the know".

A note in the docs (be it in addition to providing the workflow or instead of) is probably a good idea. Let's see what people say about this PR on Wednesday, and if we don't decide to close it then already, I might write one. :wink: Although I'm already thinking about where to put it. I don't think we have a generic "Trouble Shooting" section or something like that, right? :thinking: 

> but we may also choose to put it in https://github.com/opencast/community-workflows.

So yeah, besides a note in the docs, this could be an alternative. I also thought about adding a script to the [helper scripts](https://github.com/opencast/helper-scripts). That could even talk to the `/workflow`-endpoints directly. Thoughts on that? :thinking:

gregorydlogan:
I don't remember the conclusion we came to here, does anyone else?

JulianKniephoff:
I don't think there really was one. :thinking:

gregorydlogan:
I'm going to merge this then.  We can remove it later if it's really that big a deal.

PR #2814, opened 2021-07-09
Add track fields `is_master_playlist` and `is_live` to external API

mtneug:
This PR adds the two fields `is_master_playlist` and `is_live` to the external API when returning data about tracks. This includes the previously undocumented endpoint `/api/events/{event_id}/media`. This endpoint is now documented as well and additionally returns the `has_audio` and `has_video` fields analogously to the publication endpoints.

* [ ] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3002, opened 2021-09-13
Sign publication URL of events in External API

KatrinIhler:
The External API already offers the `sign`-parameter to sign distribution URLs when requesting one or more events with their publications.
With this change, from version 1.7.0 onward the publication URLs themselves will also be signed if that flag is set and a signing provider is configured accordingly. The reason for this change is that SWITCH uses Stream Security to protect an external player.

This affects the following endpoints of the External API:
- getEvent
- getEvents
- getEventPublication
- getEventPublications

Version 1.7.0 was already introduced by #2814 for OC11, so I'm piggybacking off of that. ;) I also amended the docs accordingly.

### DIFF
--- a/docs/guides/developer/docs/api/events-api.md
+++ b/docs/guides/developer/docs/api/events-api.md
@@ -62,15 +62,15 @@ the response:
 
 Query String Parameter     |Type                         | Description
 :--------------------------|:----------------------------|:-----------
-`sign`                     | [`boolean`](types.md#basic) | Whether public distribution urls should be signed
-`withacl`                  | [`boolean`](types.md#basic) | Whether the acl metadata should be included in the response
-`withmetadata`             | [`boolean`](types.md#basic) | Whether the metadata catalogs should be included in the response
-`withpublications`         | [`boolean`](types.md#basic) | Whether the publication ids and urls should be included in the response
+`sign`                     | [`boolean`](types.md#basic) | Whether publication urls (version 1.7.0 or higher) and distribution urls should be pre-signed.
+`withacl`                  | [`boolean`](types.md#basic) | Whether the acl metadata should be included in the response.
+`withmetadata`             | [`boolean`](types.md#basic) | Whether the metadata catalogs should be included in the response.
+`withpublications`         | [`boolean`](types.md#basic) | Whether the publication ids and urls should be included in the response.
 `withscheduling`           | [`boolean`](types.md#basic) | Whether the scheduling information should be included in the response (version 1.1.0 and higher).
 
 By setting the optional `sign` parameter to `true`, the method will pre-sign distribution urls if URL signing is turned
-on in Opencast. Remember to consider the [maximum validity of signed URLs](security-api.md#Introduction) when caching
-this response.
+on in Opencast. Starting with version 1.7.0, publication urls will also be signed if configured. Remember to consider
+the [maximum validity of signed URLs](security-api.md#Introduction) when caching this response.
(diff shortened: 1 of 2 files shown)


docs/guides/developer/docs/api/events-api.md
```diff
@@ -62,15 +62,15 @@ the response:
 
 Query String Parameter     |Type                         | Description
 :--------------------------|:----------------------------|:-----------
-`sign`                     | [`boolean`](types.md#basic) | Whether public distribution urls should be signed
-`withacl`                  | [`boolean`](types.md#basic) | Whether the acl metadata should be included in the response
-`withmetadata`             | [`boolean`](types.md#basic) | Whether the metadata catalogs should be included in the response
-`withpublications`         | [`boolean`](types.md#basic) | Whether the publication ids and urls should be included in the response
+`sign`                     | [`boolean`](types.md#basic) | Whether publication urls (version 1.7.0 or higher) and distribution urls should be pre-signed.
```
mliradelc, line 65: I don't understand about the meaning of `pre-signed`, Is it another signing before the Stream security signing?

mliradelc (changes requested):

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1603,39 +1602,33 @@ public Response getEventPublications(@HeaderParam("Accept") String acceptHeader,
...
+    if (url == null || !sign) {
+      return url;
     }
 
-    String location = uri.toString();
-    if (urlSigningService.accepts(location)) {
+    if (urlSigningService.accepts(url.toString())) {
       try {
-        location = urlSigningService.sign(location, expireSeconds, null, null);
+        return URI.create(urlSigningService.sign(url.toString(), expireSeconds, null, null));
```
mliradelc, line 1626: Would be a nice idea to add a `logger.debug` to know if the url was succesfully signed

mliradelc: The code works, I've checked with the debugger 😄

docs/guides/developer/docs/api/events-api.md
KatrinIhler: No, I just meant that the URL is already signed when requesting the event, in contrast to signing the URL explicitly afterwards with requests to /signing. This wasn't supposed to be confusing...

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
KatrinIhler: I'd rather not add a debug log for something like that, since it's pretty apparent if the URL is signed or not, and this code could potentially be called a lot. :D

mliradelc: I don't have strong opinions on this matter, so it's ok for me :)

docs/guides/developer/docs/api/events-api.md
mliradelc: Ok, so pre-sign means We don't make a request to the signing services after the query. from what I understand with versions lower than 1.6 all stays like always.

mliradelc (approved):
After testing and the doubts are now explained, its ok

PR #3003, opened 2021-09-14
Set jobLoad to 0 when jobCache is empty for very small jobLoads

Arnei:
There was an issue with a extremely low jobLoad values at ETH, which Opencast cannot handle very well (I think it was permanently blocking workers).

Logs:
```
mh03-sbox | 2021-09-11T10:54:04,822 | DEBUG | (AbstractJobProducer:175) - 308 Current load on this host: 5.9604645E-7, job's load: 4.0, job's status: DISPATCHING, max load: 4.0
mh03-sbox | 2021-09-11T10:54:04,822 | DEBUG | (AbstractJobProducer:192) - 308 Declining job 216264 of type org.opencastproject.composer with load 4 because load of 4 would exceed this node's limit of 4.
```

I'm not sure how that load came to be, or if this PR actually fixes the issue for ETH, but it seems reasonable to cleanly set the jobLoad to 0 if the jobCache is empty.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
@@ -950,7 +950,7 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
     }
     logger.debug("Current host load: {}, job load cache size: {}", format("%.1f", localSystemLoad), jobCache.size());
 
-    if (jobCache.isEmpty() && Math.abs(localSystemLoad) > 0.01f) {
+    if (jobCache.isEmpty() && localSystemLoad != 0) {
       logger.warn("No jobs in the job load cache, but load is {}: setting job load to 0",
               format("%.2f", localSystemLoad));
       localSystemLoad = 0;


mtneug:
Why was there a `Math.abs`? Can this be negative 😮 ?

gregorydlogan:
It absolutely can be, yes :)

So, the reason this check was doing a comparison rather than exact 0 is that pretty rarely do the default jobloads exactly end up with a zero value.  We have too many jobs with loads of 0.1, which when added and subtracted repeatedly end up with floating point bits left over.  You probably want [Math.signum](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#signum-double-) for this kind of check.

That being said, I'm not sure your change here is entirely correct.  Perhaps we should look at the load, and if it's less than say 0.1 we round it to 0, then continue with the rest of the logic - I suspect no one is running with loads < 0.1.  That eliminates the floating point error cruft once it hits a certain point as well.  I've thought about this a few times over the years wondering when the FP error would come back and bite us.  Apparently you're the first :)

Arnei:
I'm not deeply knowledgable about jobLoad and jobCache, so I'll defer to any suggestions. 

If you want set load to 0 if it is less than 0.1, does that mean you would not set it to 0 if it was greater than 0.1? Because that was how it worked before, and I don't understand why the jobLoad should be greater than 0 if the jobCache is empty. I'm also not sure what to do with negative jobLoad values though.

gregorydlogan:
> If you want set load to 0 if it is less than 0.1, does that mean you would not set it to 0 if it was greater than 0.1? Because that was how it worked before

Ah, good point.  This does change the behaviour somewhat, but I feel like almost no one changes the job loads outside of encoding, and in those cases they only get *larger*, or at least aren't likely to be less than 1.  I chose 0.1 when I built this as a default value to keep from dispatching *all* of jobs to a single node, there's no actual meaning to the 0.1 value.  It's intended to mean that a job imposes a load, but not enough of a load to be worth figuring out just how much load :)

> I don't understand why the jobLoad should be greater than 0 if the jobCache is empty.

It's a [known issue](https://en.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems) that's outside of the bounds that we can fix I suspect ;)

> I'm also not sure what to do with negative jobLoad values though.

Ignore them IMO.  If you want to build a check to ensure that they *can't* be used I guess I'm fine with that, but I'd say that's outside the bounds of this PR.

Arnei:
>> I don't understand why the jobLoad should be greater than 0 if the jobCache is empty.

> It's a known issue that's outside of the bounds that we can fix I suspect ;)

Ah, that's not what I meant. From your previous answer I gathered you would change the check to

`if (jobCache.isEmpty() && Math.abs(localSystemLoad) < 0.1f)`.

so I was wondering why we should not set the localSystemLoad to 0 if it is greater than 0.1f,  especially since the jobCache is empty at that point. My naive assumption here is "no jobs --> no load".

gregorydlogan:
Y'know, you're approach is probably better.  I think the fix as it stands works for me.

PR #3040, opened 2021-10-01
Deactivate Parallel Builds in CI

lkiesow:
This patch deactivates the recently activated parallel builds in the CI
system (GitHub Actions). Speed is not so much a concern there since we
need to wait at least to the next Technical Meeting for approval anyway.
Thus, we do not gain much by activating this.

But parallel Maven builds make identifying reasons for errors
incredibly hard. Especially if they happen in slow building modules
(e.g. in some JavaScript builds). Before you would just look at the end
of the build log. Now you need to scroll through thousands of lines of
logs to identify the root cause for a build failure. That is extremely
annoying as a reviewer.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -67,7 +67,7 @@ jobs:
       - name: build opencast
         run: |
           mvn --batch-mode -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
-            clean install -Pnone -T 1C
+            clean install -Pnone
 
       - name: build assemblies
         if: matrix.java == 11


gregorydlogan:
What about running both a parallel and non-parallel build in (drum roll please) parallel? The point being to check that parallel builds continue to work, but also have the non-threaded version to catch a sane build log.

mtneug:
> What about running both a parallel and non-parallel build in (drum roll please) parallel? The point being to check that parallel builds continue to work, but also have the non-threaded version to catch a sane build log.

I like that idea.

gregorydlogan:
Per discussion in the technical meeting, let's turn parallel builds off.  ci.opencast.org will (once deployed properly) continue with parallel builds.

PR #3044, opened 2021-10-04
Notes fetch comments in backend

Arnei:
The optional "notes" column causes fetching comments for each event from the backend on each refresh, even when not displayed.   This leads to way too many requests which causes flickering and made the notes column unusable on large systems.

This PR attempts a quick fix, by removing the repeated requests from the frontend and instead modifying the events endpoint in the backend to return comments with each event.

A more optimal long-term solution would probably be to add comments to the index, but that seems to be a larger undertaking.


lkiesow:
This will still cause a lot of additional work being put on the admin, with most people probably not using this.
Can we just deactivate this by default, allowing admins to opt-in if they need the feature?

PR #3049, opened 2021-10-08
Removed ManagedService from Admin UI backend

### LINKED ISSUE #1217: ToolsEndpoint restarts on configuration update
…since it is a declarative OSGI service while also implementing the ManagedService interface. This basically means that on a configuration change, the configuration is updated, then the service is shut down and reloaded and then the configuration is loaded yet again.

The endpoint should not implement the ManagedService but should the OSGi declarative service method `modified()` to handle configuration updates without reloading the whole service.

Arnei:
The ManagedService interface would cause service restarts on configuration updates. This PR removes said interface and replaces it with osgi annotations.

Resolves #1217.

### Your pull request should…

* [x] have a concise title
* [x]  [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] pass automated tests


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/OsgiEventEndpoint.java
```diff
@@ -208,12 +240,14 @@ public UrlSigningService getUrlSigningService() {
...
+  @Reference
   public void setUrlSigningService(UrlSigningService urlSigningService) {
     this.urlSigningService = urlSigningService;
   }
 
-  @Override
-  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
+  @Modified
+  public void updated(ComponentContext cc) {
+    Dictionary<String, Object> properties = cc.getProperties();
```
lkiesow: I wouldn't call it `updated()` but switch to something like `modified()` instead. It's not technically required, but users would likely expect a `ManagedService` if they encounter `updated(…)`.

Another minor thing is that instead of getting the `Dictionary` from the `ComponentContext`,  you could just use:

```java
@Modified
public void modified(Map<String, Object> properties) {
```

There are more changes like this. I'm not commenting on the other ones.

lkiesow (commented):
Generally looks good. Just a very minor complaint.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/OsgiEventEndpoint.java
Arnei: Switched `updated()` with `modified()`.

Also switched `ComponentContext` for `Map` where easily possible. There are some `modified()` methods that use `UrlSigningServiceOsgiUtil` which expects `Dictionary` instead of `Map`, and I dare not modify `UrlSigningServiceOsgiUtil` to use `Map` as UrlSigningServiceOsgiUtil` is used in other classes outside of this module.

lkiesow (approved):

KatrinIhler:
@Arnei I'm currently looking into how @Modified works for another PR and I think this one might be introducing a bug - if my understanding is correct, @Modified is only called when the configuration changes, but not on startup. If we want to read the config on startup as well, we either need to add @Activate, or call modified() from an existing activate() method.

Arnei:
> @Arnei I'm currently looking into how @Modified works for another PR and I think this one might be introducing a bug - if my understanding is correct, @Modified is only called when the configuration changes, but not on startup. If we want to read the config on startup as well, we either need to add @activate, or call modified() from an existing activate() method.

That sounds correct. What places in the code did you find where this causes an issue?

KatrinIhler:
@Arnei I think this might be an issue for AdminUIConfiguration, SeriesEndpoint & OsgiEventEndpoint.

Arnei:
Just tested this for the SeriesEndpoint and it seems you are quite correct, the `.cfg` is not read on startup as one would expect.

lkiesow:
Sorry, it seems I've missed that in my review.
You can add both annotations to one method though:

```java
@Activate
@Modified
public void modified(ComponentContext cc) {
```

PR #3050, opened 2021-10-08
Fixed limit value returned by search service endpoint

### LINKED ISSUE #1173: Search service returns incorrect limit in some cases
Via MH-9801:

https://stable.opencast.org/search/series.xml?limit=20 will return
```
<ns2:search-results xmlns:mp="http://mediapackage.opencastproject.org" xmlns:ns2="http://search.opencastproject.org" offset="0" limit="0" total="0" searchTime="2">
<ns2:query>
...
```

Obviously, the limit was set to 20 and not to 0. This always happens if less then limit objects are in the result set. But clearly less results should not effect the limit of possible results which could have been returned if enough data were available.

Arnei:
Previously, the limit value in the response xml/json was guessed at. This changes the value to actual limit that was being used.

Resolves #1173.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
@@ -155,10 +155,7 @@ public class SolrRequester {
    * @throws SolrServerException
    *           if the solr server is not working as expected
    */
-  private SearchResult createSearchResult(final SolrQuery query, final SearchQuery sQuery) throws SolrServerException {
-
-    final boolean signed = sQuery.willSignURLs();
-    final long limit = sQuery.getLimit();
+  private SearchResult createSearchResult(final SolrQuery query, final boolean signed) throws SolrServerException {
 
     // Execute the query and try to get hold of a query response
     QueryResponse solrResponse = null;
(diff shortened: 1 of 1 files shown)


lkiesow (approved):

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
```diff
@@ -169,7 +172,7 @@ private SearchResult createSearchResult(final SolrQuery query, final boolean sig
     final SearchResultImpl result = new SearchResultImpl(query.getQuery());
     result.setSearchTime(solrResponse.getQTime());
     result.setOffset(solrResponse.getResults().getStart());
-    result.setLimit(solrResponse.getResults().size());
+    result.setLimit(limit);
```
lkiesow: Tested this, but unfortunately, it does not work like this.
Setting no limit or setting `limit=0` causes the limit to be set to the default value:

```
    if ((q.getLimit() > 0) && (q.getLimit() < QUERY_MAX_ROWS)) {
      query.setRows(q.getLimit());
    } else {
      query.setRows(QUERY_MAX_ROWS);
    }
```

…but this would claim 0. The same goes for values exceeding the `QUERY_MAX_ROWS`.

lkiesow (changes requested):

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
Arnei: Thanks for checking. It seems the `row` parameter of the SolrQuery would then make for a more appropriate `limit`.

lkiesow (approved):

Issue #3017, opened 2021-09-24
Search Service not returning signed URLs


tkurze:
### Short description
The search service is not returning signed URLs for events even when instructed to do so and stream security has been configured correctly and is working otherwise (/api/events, /sign, e.g.).

#### Example
For example: When calling /search/episode.json?sign=true, a list is returned, but the url(s) are unchanged: no policy, no signature, etc.:
{
  "search-results": {
    "offset": 0,
    "limit": 3,
    "total": 3,
>[...]<
                },
              "url": "http://localhost:8080/static/mh_default_org/engage-player/ecf93214-2d13-4a6e-bf2d-e8d4edb477d2/71b809c1-25d0-4b1e-a48f-95fd7dc57fb3/output.mp4",
              "size": 16145008,
         >[...]<

#### External API (seems) unaffected
A call to /signing/accepts providing the above URL return true, and /signing/sign returns: 
http://localhost:8080/static/mh_default_org/engage-player/ecf93214-2d13-4a6e-bf2d-e8d4edb477d2/71b809c1-25d0-4b1e-a48f-95fd7dc57fb3/output.mp4?policy=eyJTdGF0ZW1lbnQiOnsiQ29uZGl0aW9uIjp7IkRhdGVMZXNzVGhhbiI6MH0sIlJlc291cmNlIjoiaHR0cDpcL1wvbG9jYWxob3N0OjgwODBcL3N0YXRpY1wvbWhfZGVmYXVsdF9vcmdcL2VuZ2FnZS1wbGF5ZXJcL2VjZjkzMjE0LTJkMTMtNGE2ZS1iZjJkLWU4ZDRlZGI0NzdkMlwvNzFiODA5YzEtMjVkMC00YjFlLWE0OGYtOTVmZDdkYzU3ZmIzXC9vdXRwdXQubXA0In19&keyId=demoKeyOne&signature=339463ca139cda2d721f235ed4786c727c48bf87ff4744209da2a9b5fa372c07

### Impact
This behavior impacts the Media Module when using external web servers to deliver content (Apache, e.g.) as no valid signature (etc.) is provided, Apache does (correctly) deny access to protected resources, rendering the media module, engage, etc. useless.
This bug does *not* concern access through the external API, /api/events, e.g.

### Concerned Opencast Versions
[...]

PR #2915, opened 2021-08-10
Unify loadFromXml behavior

Gao-Jun:
All loadFromXml function should handled with MediaPackageSerializer.

There are three loadFromXml functions implemented in `MediaPackageBuilderImpl.java`. Two of them (`loadFromXml(String xml)` and `loadFromXml(Node xml)`) will handle media package with registered `MediaPackageSerializer`. Yet, the third one `loadFromXml(InputStream is)` handle media package without registered `MediaPackageSerializer`.

This PR unify the behavior of `loadFromXml` function to make them all handle media packages with registered `MediaPackageSerializer`.

This change make a test case in `SearchServiceImplTest.java` fail. Because it use `loadFromXml(InputStream is)` to get the resource file. When it use `DefaultMediaPackageSerialzerImpl` to handle media packages, the url of the resource is translated to an absolute path which can not be accessed by `getClass().getResourceAsStream()`. So, I modify the test case, use `FileInputStream` to handle the absolute path.

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3058, opened 2021-10-11
Stream Security Fix

mliradelc:
This PR fixes the non-signing issue that was reported on the issue in #3017 

I need your comments about this line `TransformerFactory.newInstance().newTransformer().transform(xmlSource, outputTarget);` The checkstyle algorithm makes some alarms to use another class instead, but from what I've read, it's a well-known class and method to use for transform back to `inputStream`

How to try it
1. Enable stream security signing.
2. upload an event
3. use the search endpoint and look if the stream links are signed.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
@@ -27,14 +27,23 @@ import org.opencastproject.mediapackage.identifier.Id;
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.Result;
+import javax.xml.transform.Source;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
 import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathExpressionException;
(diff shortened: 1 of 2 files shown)


lkiesow:
Is this a duplicate of #2915 ?

mliradelc:
> Is this a duplicate of #2915 ?

Thanks for the hint, is kind of but not the same, #2915 streamline all the Methods for `loadFromXml`, ~~but doesn't rewrite the URLs if there are serializers.~~ The method `loadFromXml (Node xml)` has the method to rewrite the URLs, but goes `MediapackageImpl.valueof(Node xml)` instead `MediapackageImpl.valueof(inputStream is)`  

I tried the same approach but going directly to `valueof(Node xml)` and the search service for some reason didn't like it. 

Maybe a modification from that PR can make it work.

modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
```diff
@@ -100,6 +108,25 @@ public MediaPackage createNew(Id identifier) {
...
+        //Convert InputStream to XML document to rewrite the URLs
+        Document xml = XmlSafeParser.parse(is);
+        rewriteUrls(xml, serializer);
+
+        // Reconverting back to inputStream
+        ByteArrayOutputStream os = new ByteArrayOutputStream();
+        Source xmlSource = new DOMSource(xml);
+        Result outputTarget = new StreamResult(os);
+        XmlSafeParser.newTransformerFactory().newTransformer().transform(xmlSource, outputTarget);
+        is = new ByteArrayInputStream(os.toByteArray());
```
lkiesow: Why not just

```suggestion
        return loadFromXml(XmlSafeParser.parse(is));
```

mliradelc: Thanks Lars for the review, it's my first time that I make a PR to a critical part of opencast. 

I tried that, but for some reason the Return from`loadFromXml(node xml)` is `.valueof(node xlm)` makes a different result than `valueof(is)`, In my tests made that the search index won't rewrite the Url, Perhaps I made wrong the tests, if you can test it it would be great 😃 

The other reason that I wanted to avoid that it was my inexperience in the common package, making it the more transparent possible to avoid any bugs.

gregorydlogan: @mliradelc it's because your solution is calling `rewriteUrls`, and @lkiesow's does not.

Try this, starting from the same point as the above suggestion:
```
        //Convert InputStream to XML document to rewrite the URLs
        Document xml = XmlSafeParser.parse(is);
        rewriteUrls(xml, serializer);
        return loadFromXml(xml);
```

gregorydlogan:
@mliradelc can you rebase this please?

mliradelc:
> @mliradelc can you rebase this please?

I don't know what happened but now there are like 400 commits, I'm looking right now how to fix it.

mliradelc:
> > @mliradelc can you rebase this please?
> 
> I don't know what happened but now there are like 400 commits, I'm looking right now how to fix it.

@gregorydlogan  done!

PR #3064, opened 2021-10-12
Disabling Static Files Authorization for external software

nemuecahit:
We discovered that after an Update, the ILIAS authentication doesn't work if the static files authorization is enabled. To help other users with the same problem, we want to give an explanation on the cause and how to solve it.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/configuration/serving-static-files.md
+++ b/docs/guides/admin/docs/configuration/serving-static-files.md
@@ -32,6 +32,9 @@ The configuration options for requiring authentication can be found in the servi
 An alternative method to this is Opencast's [token based authorization system](stream-security.md) which allows (and
 requires) to defer all security checks to external systems.
 
+With Opencast 10, a new default configuration for static files authorization was introduced.
+For some external software, like Ilias, it is required that all of your media is accessable to every user. If that is not the case, the static files authorization must be disabled by setting the `authentication.required` option in `etc/org.opencastproject.fsresources.StaticResourceServlet.cfg` to `false`.  
+
 
 Available Static File Authorization Plug-ins
 --------------------------------------------
--- a/docs/guides/admin/docs/configuration/serving-static-files.md
+++ b/docs/guides/admin/docs/configuration/serving-static-files.md
@@ -29,11 +29,10 @@ access control lists published to Engage before allowing access.
 The configuration options for requiring authentication can be found in the service's configuration file at
 `etc/org.opencastproject.fsresources.StaticResourceServlet.cfg`.
 
+For some external software, like the ILIAS plugin, it is required to set `authentication.required` to `false` as static files authorization is active by default.
+
 An alternative method to this is Opencast's [token based authorization system](stream-security.md) which allows (and
 requires) to defer all security checks to external systems.
-
-With Opencast 10, a new default configuration for static files authorization was introduced.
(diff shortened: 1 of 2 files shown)


docs/guides/admin/docs/upgrade.md
```diff
@@ -68,3 +68,5 @@ in:
 ```
 etc/org.opencastproject.fsresources.StaticResourceServlet.cfg
 ```
+
+If you are using external software, like Ilias for example, it is required that all of your media is accessable to every user. If that is not the case, the static files authorization must be disabled by setting the `authentication.required` option in the file above to `false`.
```
lkiesow: This is not true. For example, Stud.IP would be external software, but you can (and probably should) still use this feature because users are authenticated against Opencast. You should just deactivate this if you do not want Opencast to check for authorization. And that's basically what the sentence above the configuration in here says.

If you want to tell people about the problem in ILIAS, maybe be specific about that?

```suggestion
The [ILIAS plugin](https://github.com/fluxapps/OpenCast), for example, does not authenticate users against Opencast. This is why you probably want to disable this feature and make your media publically accessable if you use that plugin.
```

docs/guides/admin/docs/configuration/serving-static-files.md
```diff
@@ -32,6 +32,9 @@ The configuration options for requiring authentication can be found in the servi
 An alternative method to this is Opencast's [token based authorization system](stream-security.md) which allows (and
 requires) to defer all security checks to external systems.
 
+With Opencast 10, a new default configuration for static files authorization was introduced.
```
lkiesow: Avoid mentioning specific Opencast versions. The docs are specific to the versions anyway:

```suggestion
Static files authorization is active by default.
```

docs/guides/admin/docs/configuration/serving-static-files.md
```diff
@@ -32,6 +32,9 @@ The configuration options for requiring authentication can be found in the servi
...
+For some external software, like Ilias, it is required that all of your media is accessable to every user. If that is not the case, the static files authorization must be disabled by setting the `authentication.required` option in `etc/org.opencastproject.fsresources.StaticResourceServlet.cfg` to `false`.
```
lkiesow:
```suggestion
For some external software, like the ILIAS plugin, it is required to set `authentication.required` to `false`.  
```

The paragraph in line 29 talks about where to find the “configuration options for requiring authentication”. Maybe move this paragraph directly after that one? Then, we also do not need to mention the configuration file again.

lkiesow (changes requested):
This targets `develop` should this go into `r/10.x` instead?

lkiesow (approved):

docs/guides/admin/docs/upgrade.md
```diff
@@ -67,4 +67,5 @@ in:
 
 ```
 etc/org.opencastproject.fsresources.StaticResourceServlet.cfg
-```
+
+The [ILIAS plugin](https://github.com/fluxapps/OpenCast), for example, does not authenticate users against Opencast. If you are using that plugin, this is why you probably want to disable this feature and make your media publically accessable.
```
lkiesow: Fixing the line length test failure…

```suggestion
The [ILIAS plugin](https://github.com/fluxapps/OpenCast), for example, does not authenticate users against Opencast.     
If you are using that plugin, this is why you probably want to disable this feature and make your media publically       
accessable.
```

PR #3074, opened 2021-10-14
Improve development dependency containers

JulianKniephoff:
I'm using these more and more these days, even multiple instances of them at once. Unfortunately that's not quite possible as is, so I'm always carrying this untracked change with me. Maybe someone else will find it useful.

This patch clarifies some things in the `readme.md` for the `docker-compose`-files (see especially the changes in the first commit and that commit's commit message for more info), but most importantly it makes the container names configurable, so you can have multiple instances of these compose files in parallel.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/scripts/devel-dependency-containers/.env
+++ b/docs/scripts/devel-dependency-containers/.env
@@ -1,0 +1,1 @@
+COMPOSE_PROJECT_NAME=opencast
--- a/docs/scripts/devel-dependency-containers/docker-compose-all-sql.yml
+++ b/docs/scripts/devel-dependency-containers/docker-compose-all-sql.yml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   elasticsearch:
-    container_name: opencast-elasticsearch
+    container_name: $COMPOSE_PROJECT_NAME-elasticsearch
     image: docker.io/library/elasticsearch:7.9.2
     ports:
       - 127.0.0.1:9200:9200
@@ -16,7 +16,7 @@ services:
     volumes:
       - ../activemq/activemq.xml:/activemq-opencast.tpl.xml:ro
   postgresql:
-    container_name: opencast-postgresql
+    container_name: $COMPOSE_PROJECT_NAME-postgresql
     image: docker.io/library/postgres:13.2
     ports:
       - 127.0.0.1:5432:5432
@@ -25,7 +25,7 @@ services:
       POSTGRES_PASSWORD: dbpassword
       POSTGRES_DB: opencast
   mariadb:
-    container_name: opencast-mariadb
+    container_name: $COMPOSE_PROJECT_NAME-mariadb
     image: docker.io/library/mariadb:10.5.9
     ports:
       - 127.0.0.1:3306:3306
--- a/docs/scripts/devel-dependency-containers/docker-compose-mariadb.yml
+++ b/docs/scripts/devel-dependency-containers/docker-compose-mariadb.yml
@@ -1,7 +1,7 @@
 version: '2'
 services:
   elasticsearch:
-    container_name: opencast-elasticsearch
+    container_name: $COMPOSE_PROJECT_NAME-elasticsearch
     image: docker.io/library/elasticsearch:7.9.2
     ports:
       - 127.0.0.1:9200:9200
@@ -9,14 +9,14 @@ services:
     environment:
       discovery.type: single-node
   activemq:
-    container_name: opencast-activemq
+    container_name: $COMPOSE_PROJECT_NAME-activemq
     image: quay.io/lkiesow/activemq
     ports:
       - 127.0.0.1:61616:61616
     volumes:
       - ../activemq/activemq.xml:/activemq-opencast.tpl.xml:ro
   mariadb:
-    container_name: opencast-mariadb
+    container_name: $COMPOSE_PROJECT_NAME-mariadb
     image: docker.io/library/mariadb:10.5.9
     ports:
       - 127.0.0.1:3306:3306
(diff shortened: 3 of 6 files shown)


lkiesow:
-1 to this change.
It completely breaks the compose files for podman:

```
% podman-compose up                                                                                                                          (git)-[JulianKniephoff-improve-dev-dep-containers] [125]
using podman version: podman version 3.4.0
podman pod create --name=devel-dependency-containers --share net -p 127.0.0.1:61616:61616
a36fd27f8bbd6373fc44eab057727a65961c19a161931c5ad2401185a910b6c1
0
podman create --name= --pod=devel-dependency-containers --label io.podman.compose.config-hash=123 --label io.podman.compose.project=devel-dependency-containers --label io.podman.compose.version=0.0.1 --label com.docker.compose.container-number=1 --label com.docker.compose.service=activemq -v /home/lars/dev/opencast/docs/scripts/devel-dependency-containers/../activemq/activemq.xml:/activemq-opencast.tpl.xml:ro --add-host elasticsearch:127.0.0.1 --add-host :127.0.0.1 --add-host activemq:127.0.0.1 --add-host :127.0.0.1 quay.io/lkiesow/activemq
Error: bad format for add-host: ":127.0.0.1"
125
podman start -a 
Error: name or ID cannot be empty
125
```

It describes non-existing `podman-compose` flags and attributes incorrect behavior:

```
% podman-compose down --volumes                                                                                                                (git)-[JulianKniephoff-improve-dev-dep-containers] [0]
usage: podman-compose [-h] [-f file] [-p PROJECT_NAME] [--podman-path PODMAN_PATH] [--no-ansi] [--no-cleanup] [--dry-run] [-t {1pod,1podfw,hostnet,cntnet,publishall,identity}]
                      {help,version,pull,push,build,up,down,ps,run,start,stop,restart,logs} ...
podman-compose: error: unrecognized arguments: --volumes
```

I think the concept would need to be expanded since pod names would still clash.
I didn't test that, though.

I'm not a huge fan of making this more complicated for the purpose of serving very specific needs.

Finally, I don't think that the recommended default behavior should be to keep all data (`stop` vs `down`).

gregorydlogan:
I'm very against keeping data as well, that seems like it would be very dependent on what you're doing at a given time. Certainly for my use (ie doing most of the merges) it would be massively bad for me to keep state between tests :)

I like the idea of being able to have multiple branches running in parallel, and don't have a problem with the general concept but it does need to work with both docker- and podman-compose out of the box. I already have some local tooling that already does part of this via sed and shell scripts, and it's quite handy so +1 for the general idea.

JulianKniephoff:
Addressing @lkiesow's comments first:

> -1 to this change. It completely breaks the compose files for podman:

Apparently you are right. Admittedly I didn't test that before. I was naively assuming that `podman{,-compose}` are compatible with their `docker` counterparts like everyone keeps repeating. :sweat_smile: 

It **might** work if you alias `docker` to `podman` and just use `docker-compose`, but I didn't test that either. I'm just assuming because the variable substitution and pulling values out of `.env` seem to be `docker-compose`-features, and don't have anything to do with `docker` vs. `podman`.

Anyway, I don't see a way to easily make it work with `podman-compose`, so since we want that (which I can respect) I'ma go ahead and close this. However, I still can't keep myself from addressing the other points.

> I'm not a huge fan of making this more complicated for the purpose of serving very specific needs.

I disagree that this is a "very specific need," and take Greg's comment and his custom tooling as supporting evidence. (I hope that's okay, @gregorydlogan. :wink:) Like, this idea is something that I imagine anyone involved in Opencast development (and projects around it) for any extended period of time could use. Maybe other people handle multiple long-lived branches differently, and I would be curious about how, but uploading a video anew every time you want to test something if you switched tasks in between, or everyone maintaining their own bash-and-sed-abominations shouldn't be the solution, IMO. :thinking:

And that's the other point where I would disagree: We should accept minor inconveniences to make the developer experience better, I think. And as it is, this is rather minor in my opinion. Heck, apart from the fact that we would probably have to extend it to work with `podman-compose`, I actually think this is making things simpler and less specific, allowing for more flexibility basically for free, because it was already in the tool; we just indirectly deactivated it before. :shrug: 

> Finally, I don't think that the recommended default behavior should be to keep all data (`stop` vs `down`).

Not sure I would call my wording a recommendation, but sure, one option is mentioned first, I guess. :shrug: 

However, I found the previous phrasing pretty misleading. It's not true that the containers don't preserve state. It's just that `docker-compose down` removes that state. If you forget to do that (which can totally happen if you start everything with `-d` because everything runs in the background you might just forget, but is also what happened to me who runs without `-d` exactly to **avoid** forgetting, but who didn't know that `Ctrl+C`-ing out of that **also** just stops the containers as opposed to bringing them `down`) you might be confused by leftover state. I think it's better to just mention both options.

But anyway, I know/understand everything now and am no longer confused. For all I care, we can leave it as is. Alternatively, feel free to tell me to extract the doc changes into their own PR or something like that. :shrug: 

And finally to address @gregorydlogan's concern:

> I'm very against keeping data as well, that seems like it would be very dependent on what you're doing at a given time.

I didn't actually change whether or not data is kept. Both are already possible. I merely pointed that out.

mtneug:
If you don't set `contianer_name`, `docker-compose` will set one based on the project and service name. I assume that `podman-compose` will do the same. The project name is by default the folder name of the `docker-compose.yaml` but can be changed with the `--project-name` flag ([example](https://github.com/opencast/opencast-docker/#quick-start) in the Opencast Docker repo).

JulianKniephoff:
Yeah ok, that would be an alternative, but I assumed we don't want the ugly default name. :thinking:

PR #3075, opened 2021-10-14
Expose ACLs in search service (via `SearchResultItem`)

LukasKalbertodt:
Like other PRs before, these changes are required by the upcoming Tobira module. The module will enable the new video portal Tobira to synchronize with Opencast somewhat efficiently. It's still in development in https://github.com/elan-ev/opencast-tobira/. We want to push all tobira-unrelated improvements to the community as early as possible though. Hence this PR.

---

The ACL information is available in the search index in non-XML form. Code using the search service might be interested in those ACLs without needing to load an attachment and parse XML. So it makes sense to provide this information as part of `SearchResultItem`.

Unfortunately, to make the `SearchService` remote impl work (which uses the search API), the access control list has to be serialized and exposed in that API. Offering more information in the de-facto public and often-used search REST API was not the goal of this PR. But I don't really see a good alternative.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
@@ -725,8 +725,9 @@ public class SolrIndexManager {
         actionPermissions = new ArrayList<String>();
         permissions.put(entry.getAction(), actionPermissions);
       }
-      actionPermissions.add(entry.getRole());
-
+      if (adminRole != null && !entry.getRole().equals(adminRole)) {
+        actionPermissions.add(entry.getRole());
+      }
     }
 
     // Write the permissions to the solr document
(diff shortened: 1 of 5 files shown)


modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrIndexManager.java
```diff
@@ -721,12 +721,9 @@ static void setAuthorization(SolrInputDocument doc, SecurityService securityServ
         logger.debug("Search service doesn't know how to handle action: {}", entry.getAction());
         continue;
       }
-      if (acl == null) {
```
lkiesow, line 724: Memo for myself: In the old code, `acl` could never have been `null`. This always evaluated to false.

lkiesow (approved):

PR #3076, opened 2021-10-14
Store modification & deletion dates for series and add range-lookup method to `SeriesService`

LukasKalbertodt:
Like other PRs before, these changes are required by the upcoming Tobira module. The module will enable the new video portal Tobira to synchronize with Opencast somewhat efficiently. It's still in development in https://github.com/elan-ev/opencast-tobira/. We want to push all tobira-unrelated improvements to the community as early as possible though. Hence this PR.

---

To synchronized with Opencast efficiently, Opencast needs to know when a "thing" last changed. That was already true for events (at least in the search service), but Opencast did not know this information for series. This PR changes that by adding a modified date field to series. 

To notify other applications, like Tobira, about a deleted series, we now store a deletion date instead of actually deleting the series.

To use this new information, a new method is added to the `SeriesService` that allows modification date range lookup and returns series sorted by modification date. Unfortunately, to implement the remote impl, we needed to add a new JSON API as well. This is described as internal and is not expected to be used outside of Opencast itself. 

Also see commit message for more details. This PR can probably be reviewed commit-by-commit fairly well.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/upgrade/10_to_11/postgresql9.sql
+++ b/docs/upgrade/10_to_11/postgresql9.sql
@@ -1,5 +1,5 @@
 -- Increase mime_type field size
-ALTER TABLE oc_assets_asset MODIFY COLUMN mime_type VARCHAR (255);
+ALTER TABLE oc_assets_asset ALTER COLUMN mime_type TYPE VARCHAR (255);
 
 -- Add modified and deletion date fields to series.
 --
@@ -20,4 +20,4 @@ UPDATE oc_series
 UPDATE oc_series
     SET modified_date = TIMESTAMP '1970-01-01 00:00:01'
     WHERE modified_date IS NULL;
-ALTER TABLE oc_series MODIFY modified_date TIMESTAMP NOT NULL;
+ALTER TABLE oc_series ALTER COLUMN modified_date SET NOT NULL;
--- a/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreGsonAdapter.java
+++ b/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreGsonAdapter.java
@@ -1,0 +1,57 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.metadata.dublincore;
+
+
+import com.google.gson.JsonParser;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+import org.json.simple.parser.ParseException;
+
+import java.io.IOException;
+
+
+/** Gson adapter to provide proper (de)serialization for {@code DublinCoreCatalog} */
+public class DublinCoreGsonAdapter extends TypeAdapter<DublinCoreCatalog> {
+  @Override
+  public void write(JsonWriter out, DublinCoreCatalog catalog) throws IOException {
+    out.jsonValue(catalog.toJson());
+  }
+
+  @Override
+  public DublinCoreCatalog read(JsonReader in) throws IOException {
+    try {
+      // TODO: this is a bit stupid. To avoid duplicating logic of
+      // `DublinCoreJsonFormat.read(JSONObject json)`, we convert given reader
+      // into a string and then parse it again with `DublinCoreJsonFormat.read`.
+      // At some point we should either parse into a `JSONObject` (from json.simple)
+      // directly or avoid parsing twice some other way.
+      String raw = JsonParser.parseReader(in).toString();
+      return DublinCoreJsonFormat.read(raw);
+    } catch (ParseException e) {
+      // Gson throws parse errors as `IllegalStateException`. For example, see `beginObject` of
+      // `JsonReader`.
+      throw new IllegalStateException("could not parse JSON when deserializing DublinCoreCatalog: ", e);
+    }
+  }
+}
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
@@ -49,6 +49,7 @@ import org.opencastproject.rest.RestConstants;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AccessControlParser;
 import org.opencastproject.security.api.UnauthorizedException;
+import org.opencastproject.series.api.Series;
 import org.opencastproject.series.api.SeriesException;
 import org.opencastproject.series.api.SeriesQuery;
 import org.opencastproject.series.api.SeriesService;
@@ -68,6 +69,7 @@ import com.entwinemedia.fn.data.Opt;
 import com.entwinemedia.fn.data.json.JValue;
 import com.entwinemedia.fn.data.json.Jsons;
 import com.entwinemedia.fn.data.json.SimpleSerializer;
+import com.google.gson.Gson;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -84,7 +86,10 @@ import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.text.ParseException;
+import java.util.Date;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.DELETE;
@@ -124,6 +129,8 @@ import javax.ws.rs.core.Response;
 public class SeriesRestService {
 
   private static final String SERIES_ELEMENT_CONTENT_TYPE_PREFIX = "series/";
+
+  private static final Gson gson = new Gson();
 
   /** Logging utility */
   private static final Logger logger = LoggerFactory.getLogger(SeriesRestService.class);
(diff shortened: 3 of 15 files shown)


modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -383,6 +403,60 @@ public DublinCoreCatalog getSeries(String seriesId) throws NotFoundException, Se
...
+    // Validate parameters
+    if (limit <= 0) {
+      throw new IllegalArgumentException("limit has to be > 0");
+    }
+
+    // Make sure the user is actually an administrator of sorts
+    User user = securityService.getUser();
+    if (!user.hasRole(GLOBAL_ADMIN_ROLE) && !user.hasRole(user.getOrganization().getAdminRole())) {
+      throw new UnauthorizedException(user, getClass().getName() + ".getModifiedInRangeForAdministrativeRead");
+    }
```
mtneug, line 419: Does this mean the Tobira API user needs to be (org) admin? I would not give Tobira global admin rights. Org admin maybe, but I'm still unsure if this is a good design. 

Do you expect this method to be used internally or can we simply rely on Spring security for the REST endpoint?

LukasKalbertodt: Currently, we need the Tobira user to be org admin, yes. I was also not happy with this, but ultimately, Tobira needs read access to all events that should potentially appear on the Tobira page. Further, it needs write access to all events that could potentially modified through Tobira. For most institutions that's often simply "Tobira needs read and write access to all videos". After all, users of Tobira perform actions in Opencast *through* Tobira. So it kind of makes sense. This is the same for some other integrations, e.g. the Ilias plugin which also requires a very powerful Opencast user.

Of course, you could always add a special Tobira role to all ACLs to avoid handing out org admin rights. 

Your last quesiton I don't quite understand. I expect the REST endpoint to be used *only* by the series service remote impl.

mtneug: This is not true. ILIAS and Moodle don't require org admin rights and you can thus have videos that are not seen by those systems.

LukasKalbertodt: I just said "a very powerful Opencast user", not org admin. Though yes, that's still not technically correct because the user used by Ilias/Moodle does not *need* to have lots of privileges. But it typically does, i.e. is allowed to read/write the majority (if not all) events.

Again, I also don't like the Tobira module requiring org admin right now. And I would like to change that. I think this can be easily changed in the future. I'm just saying the OC user used by Tobira has very similar requirements to OC users used by Ilias/Moodle.

LukasKalbertodt:
This is still missing an appropriate SQL migration. And I noticed that the code likely has to be adjusted to deal with old series that don't have a valid modification date. So I will work on this and mark as "read for review" as soon as it's done.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
mtneug: Fair enough and I think we see this similarly :)

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesEntity.java
```diff
@@ -49,17 +53,39 @@
...
+    indexes = {
+        @Index(name = "IX_oc_series_modified_date", columnList = ("modified_date")),
+    }
+)
 @NamedQueries({
-    @NamedQuery(name = "Series.findAll", query = "select s from SeriesEntity s"),
-    @NamedQuery(name = "Series.getCount", query = "select COUNT(s) from SeriesEntity s"),
+    @NamedQuery(
+        name = "Series.findAll",
+        query = "select s from SeriesEntity s where s.deletionDate is null"
```
Arnei, line 67: Could you explain to me when do we have a series that is deleted still in the db? As far as I know, deleted series are removed from the db, so storing a deletionDate makes little sense?

LukasKalbertodt: This PR also changes that. When deleting a series, it is not actually deleted from the DB, but we only set the deleted date to a non-null value.

```diff
- em.remove(entity);
+ entity.setModifiedDate(new Date());
+ entity.setDeletionDate(new Date());
+ em.merge(entity);
```

https://github.com/opencast/opencast/pull/3076/commits/afb82d27d61370bb2d5903e4a25881c9aad11401#diff-8ebbecff33097385ffb1bdcc22e6a8acd012f150472562fa17ab98838c7afd43L172-L173

Arnei: Right, not sure how I missed that. Thanks for the link

Arnei: Does that mean there is no way of permanently deleting a series from the db anymore short of directly removing them from the db by hand?

LukasKalbertodt: Yes, that is correct. I mean, for Opencast, it is deleted as all SQL queries retrieving series now explicitly check for `deletion_date is null`. But yes, the data is still in the database. 

Maybe instead of just setting the modification and deletion dates on removal, all other columns should be set to empty or dummy values? But that could always be changed later I guess?

Arnei (approved):

Arnei:
Looks good to me, but I think I'd like a second opinion here. Particularly on the "series aren't actually deleted anymore"-part.  So far, I've not actually managed to come up with a good argument on why this would be a problem, but I still feel like it could be a problem. Maybe there are some database limitations or weird data privacy laws I'm unaware of.

docs/upgrade/10_to_11/postgresql9.sql
```diff
@@ -0,0 +1,23 @@
...
+UPDATE oc_series
+    SET modified_date = (
+        SELECT MAX(oc_search.modification_date)
+            FROM oc_search
+            WHERE oc_search.series_id = oc_series.id
+    );
+UPDATE oc_series
+    SET modified_date = TIMESTAMP '1970-01-01 00:00:01'
+    WHERE modified_date IS NULL;
+ALTER TABLE oc_series ALTER COLUMN modified_date SET NOT NULL;
```
lkiesow, line 23: Does it make sense to run at the end:

```sql
VACUUM (VERBOSE, ANALYZE) oc_series;
```

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreGsonAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.opencastproject.metadata.dublincore;
+
+
+import com.google.gson.JsonParser;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+import org.json.simple.parser.ParseException;
```
lkiesow, line 29: If you use Gson, you probably also want to use Gson's exception type:
[`com.google.gson.JsonParseException](https://www.javadoc.io/doc/com.google.code.gson/gson/2.6.2/com/google/gson/JsonParseException.html)

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -345,6 +350,91 @@ public Response getSeriesAccessControlListJson(@PathParam("seriesID") String ser
...
+      return badRequest("Parameter 'limit' <= 0, but it has to be > 0");
+    }
+
+    try {
+      final List<Series> series = seriesService.getAllForAdministrativeRead(
+          new Date(from),
+          Optional.ofNullable(to).map(millis -> new Date(millis)),
+          limit);
+
+      Gson gson = new Gson();
```
lkiesow: Gson is thread-safe, and you can make this a class field. Constructing this every time has quite some overhead and caused problems for Opencast at several places.

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -345,6 +350,91 @@ public Response getSeriesAccessControlListJson(@PathParam("seriesID") String ser
...
+          .build();
+    }
+  }
+
+  /**
+   * Returns a {@code Response} object representing a BAD_REQUEST with the given
+   * message as body. Also logs the message.
+   */
+  private static Response badRequest(String msg) {
+    logger.warn("Bad request to /series/allInRangeAdministrative: {}", msg);
```
lkiesow: This lets users spam the logs with warnings.
For a client error, the error is returned.
Make this a debug log:
```suggestion
    logger.debug("Bad request to /series/allInRangeAdministrative: {}", msg);
```

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -169,12 +177,15 @@ public void deleteSeries(String seriesId) throws SeriesServiceDatabaseException,
           throw new UnauthorizedException(currentUser + " is not authorized to update series " + seriesId);
         }
       }
-      em.remove(entity);
+
+      entity.setModifiedDate(new Date());
+      entity.setDeletionDate(new Date());
```
lkiesow: Maybe make sure they are actually identical?
```suggestion
      var date = new Date();
      entity.setModifiedDate(date);
      entity.setDeletionDate(date);
```

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -383,6 +403,60 @@ public DublinCoreCatalog getSeries(String seriesId) throws NotFoundException, Se
...
+        if (from.after(to.get())) {
+          throw new IllegalArgumentException("`from` is after `to`");
+        }
+
+        q = em.createNamedQuery("Series.getAllModifiedInRange", SeriesEntity.class)
+            .setParameter("from", from)
+            .setParameter("to", to.get())
+            .setMaxResults(limit);
+      } else {
+        q = em.createNamedQuery("Series.getAllModifiedSince", SeriesEntity.class)
```
lkiesow, line 436: Above, you check if the user is an admin or an organizational admin. This query returns all events from all organizations. This means, you allow admins from one organization to access events from another organization. I don't think that's intended, is it?

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -383,6 +403,60 @@ public DublinCoreCatalog getSeries(String seriesId) throws NotFoundException, Se
...
+            .setParameter("since", from)
+            .setMaxResults(limit);
+      }
+
+      final List<Series> out = new ArrayList<>();
+      for (SeriesEntity entity : q.getResultList()) {
```
lkiesow, line 443: `SeriesEntity` holds basically the same data the new `Series` does. Would it make sense to have `SeriesEntity` implement/extend `Series`?

modules/series-service-remote/src/main/java/org/opencastproject/series/remote/SeriesServiceRemoteImpl.java
```diff
@@ -322,6 +332,47 @@ public DublinCoreCatalog getSeries(String seriesID) throws SeriesException, NotF
...
+      }
+
+      if (response.getStatusLine().getStatusCode() == SC_BAD_REQUEST) {
+        throw new SeriesException("internal server error when fetching /allInRangeAdministrative.json");
+      } else if (response.getStatusLine().getStatusCode() == SC_UNAUTHORIZED) {
+        throw new UnauthorizedException("got UNAUTHORIZED when fetching /allInRangeAdministrative.json");
+      } else {
+        // Retrieve and deserialize data
+        Gson gson = new Gson();
+        Type type = new TypeToken<ArrayList<Series>>() { }.getType();
```
lkiesow: Make these class fields.

lkiesow (changes requested):

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -345,6 +350,91 @@ public Response getSeriesAccessControlListJson(@PathParam("seriesID") String ser
...
+              name = "from",
+              isRequired = true,
+              description = "Start of date range (inclusive) in milliseconds "
+                  + "since 1970-01-01T00:00:00Z. Has to be >=0.",
+              type = Type.INTEGER
+          ),
+          @RestParameter(
+              name = "to",
+              isRequired = false,
+              // TODO: this shows the default value as 0 despite us not setting this value!
```
lkiesow, line 375:
```suggestion
              defaultValue = "",
```

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
LukasKalbertodt: I don't really think so. `SeriesEntity` for me is just the connection to the DB and I would not want to mix it with `Series`. For example, `Series` contains `DublinCoreCatalog dublinCore`. If `SeriesEntity` would derive from `Series` that would have to be `null` all the time? And from experience, in these situations, sooner or later there will be more and more differences between the "database adapter" and the thing you want to use in your actual code.

docs/upgrade/10_to_11/postgresql9.sql
LukasKalbertodt: Not sure to be honest. At least in postgres (as far as I know), the first 8 nullable columns don't have any overhead as there is always an 8bit mask per row for exactly that. But :shrug:

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
LukasKalbertodt: Nop, that still shows `0` in the test form box :/

LukasKalbertodt:
@lkiesow I think I fixed all of your remarks (except the four which I comment-replied to). Most changes have been amended to the respective commits. Only the security fix is its own commit.

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreGsonAdapter.java
LukasKalbertodt: Well, I am not throwing that exception. I am catching it. Because I'm using an existing method which uses JSON simple. It's a bit stupid but was the easiest solution. Everything else would have required writing lots more code and likely duplicating logic. Kinda annoying to have multiple JSON libraries.

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
lkiesow: Interesting. Well… let's ignore that for now.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
lkiesow: This is now resolved in the query `getAllModifiedSince`

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
lkiesow: I would have probably solved that by making `Series` an interface and just requiring the methods, but fair enough, it's not really important.

lkiesow (approved):

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -303,13 +316,20 @@ public DublinCoreCatalog storeSeries(DublinCoreCatalog dc)
...
     try {
       tx.begin();
-      SeriesEntity entity = getSeriesEntity(seriesId, em);
-      if (entity == null) {
+      SeriesEntity entity = getPotentiallyDeletedSeriesEntity(seriesId, em);
+      if (entity == null || entity.isDeleted()) {
+        // If the series existed but is marked deleted, we completely delete it
+        // here to make sure no remains of the old series linger.
+        if (entity != null) {
+          this.deleteSeries(seriesId);
```
KatrinIhler, line 324: Just ran into this while testing my migration scripts for Bern - this doesn't actually seems to work, since deleteSeries() would only mark the entity as deleted again, not remove it (found no method to do this), but mostly because this throws a NotFound for entities which are already marked as deleted. To test this, just create a series, delete it, and then try to recreate a series with the same id. Since you know how this feature is supposed to work, can you take a look at this?

PR #3079, opened 2021-10-18
Put max load config of server into right place

KatrinIhler:
Currently the max load config for a server doesn't do anything because it's in the wrong place - it was moved into the service registry config (most likely by #272), but the code looks for it in the custom.properties. So I put it back there. ;)

For the future it might be nice for this to be hot-loaded, but that change would rather be something for develop.


gregorydlogan (approved):
Hotloading is on my todo list, especially since I'm also kinda-sorta working on ripping the job dispatcher out of the SR.  LGTM.

PR #3082, opened 2021-10-19
Fix Job Load Warning

lkiesow:
This patch prevents the job dispatcher from printing our warnings on
rounding errors.

Previously, you would see errors like this:

    WARN  | (ServiceRegistryJpaImpl:954) -
            No jobs in the job load cache,
            but load is -0.00: setting job load to 0

Which were probably caused by the float value being not exactly zero.
This patch prevents the warning if it is more or less zero and we can
assume that this was a rounding error.

This will also always set `localSystemLoad` to zero if the cache is
empty, resulting in the same behavior for that as before. Assuming that
was correct, this should still be correct, just without any annoying
warnings for rounding problems.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):
LGTM

Arnei (approved):

PR #3100, opened 2021-10-25
Remove ActiveMQ dependencies

gregorydlogan:
This PR removes all trace (hopefully) of ActiveMQ from our codebase.  It's no longer required for build, or runtime environments.

Notes:
- The overall messaging structure has not changed: The issue was passing things to ActiveMQ, not the message bus arch in general.
- Each major service is it's own commit, I would review each separately
- Release notes have not been added yet, but I figured I (probably?) also missed something anyway.

Work sponsored by our generous Crowdfunding initiatives!

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
Tagging @mliradelc for livestreaming testings, and @KatrinIhler for load testing on something larger than I can set up here.

lkiesow:
We should test if OAI-PMH auto updates.

gregorydlogan:
@wsmirnow can we test this against an ETH test box for that?

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -221,7 +221,6 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-lti-service-impl/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-mattermost-notification-workflowoperation/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-message-broker-api/${project.version}</bundle>
```
lkiesow, line 224: If we don't need ActiveMQ any longer, we should also not need the `message-broker-api` any longer.

gregorydlogan: Take a look at what's in there now. It's still in use - I added three interfaces for the various flavours of messaging. Happy to rename the bundle, just not sure what to call it. Something like `opencast-messaging-api`?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -78,8 +78,8 @@
 import org.opencastproject.mediapackage.MediaPackageElements;
 import org.opencastproject.mediapackage.MediaPackageParser;
 import org.opencastproject.mediapackage.MediaPackageSupport;
-import org.opencastproject.message.broker.api.MessageSender;
 import org.opencastproject.message.broker.api.assetmanager.AssetManagerItem;
+import org.opencastproject.message.broker.api.update.IAssetManagerUpdateHandler;
```
geichelberger: Why change the naming convention and use an I as a prefix?

mliradelc:
> Tagging @mliradelc for livestreaming testings, and @KatrinIhler for load testing on something larger than I can set up here.

No problem, I will do the livestreaming tests 😄

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
gregorydlogan: Bah, I meant to change those back.  They colided with some pre-existing classes, so I had to pick a different name.  I'll update the names to remove the prefix.

modules/asset-manager-impl/src/test/java/org/opencastproject/assetmanager/impl/AssetManagerMessagingTest.java
```diff
@@ -40,17 +39,17 @@
 import java.io.Serializable;
 
 /**
- * Test message sending to ActiveMQ.
+ * Test message sending via the AM.
```
mliradelc:
```suggestion
 * Test message sending via the Asset Manager.
```

mliradelc: It will be useful for the javadocs

mliradelc (commented):
Only a little change for the javadoc processing

wsmirnow:
> @wsmirnow can we test this against an ETH test box for that?

Hi @gregorydlogan, for sure. But it is easier to enable OAI-PMH autorepublish (see [config](https://github.com/opencast/opencast/blob/9557d33487ac92eed2ff7a432cbfe13a5e7ee2c0/etc/org.opencastproject.event.handler.OaiPmhUpdatedEventHandler.cfg#L2)) and test it locally by serving http://localhost:8080/oaipmh/default?verb=GetRecord&metadataPrefix=matterhorn&identifier=<MP_ID>

gregorydlogan:
@wsmirnow thanks for the tip, OAIPMH seems to work as expected :D

docs/guides/admin/docs/releasenotes/activemq.txt
```diff
@@ -0,0 +1 @@
+ActiveMQ is no longer required by Opencast.  Assuming you have not integrated ActiveMQ into anything else it is safe to uninstall.
```
lkiesow: People should know if they have and this may make users unsure if they aren't using it.
```suggestion
ActiveMQ is no longer required by Opencast.  It is safe to uninstall.
``

File: modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java [hunk 5]
```diff
@@ -197,6 +196,8 @@
 
   private final HashMap<String, RemoteAssetStore> remoteStores = new LinkedHashMap<>();
 
+  private final List<AssetManagerUpdateHandler> handlers = new ArrayList<>();
```
lkiesow: `ArrayList` is not thread-safe, is it?

modules/conductor/src/main/java/org/opencastproject/event/handler/ConductingEpisodeUpdatedEventHandler.java
```diff
@@ -21,100 +21,43 @@
...
-          logger.error("Problem while getting episode update message events", t);
-        } finally {
-          securityService.setOrganization(null);
-          securityService.setUser(null);
-        }
+    AssetManagerItem.TakeSnapshot snapshotItem = (AssetManagerItem.TakeSnapshot) messageItem;
+    if (AssetManagerItem.Type.Update.equals(snapshotItem.getType())) {
+      // the OAI-PMH handler is a dynamic dependency
+      if (oaiPmhUpdatedEventHandler != null) {
+        oaiPmhUpdatedEventHandler.handleEvent(snapshotItem);
```
lkiesow, line 73: Why does this call the `OaiPmhUpdatedEventHandler`? Why this class? Why doesn't `OaiPmhUpdatedEventHandler` implement `AssetManagerUpdateHandler` directly?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1436,11 +1438,16 @@ public static Snapshot rewriteUris(Snapshot snapshot, Fn<MediaPackageElement, UR
             mpCopy);
   }
 
+  public void fireEventHandlers(AssetManagerItem item) {
```
lkiesow, line 1466: Not sure where to put this comment, so, let me put it here since this is where the event handlers are triggered. We have had the `MessageItems` (I assume) so that we could have one ActiveMQ queue. Now that we can talk to the event handlers directly, wouldn't it be much easier for the `AssetManagerUpdateHandler` to require the two methods `update()` and `delete()` instead of stuffing all data into the same `MessageItem`, decide what type it is and then do some special threatment based on the type? Basically, the `MessageItem` now just adds complexity, no more functionality. Or did I miss something?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -244,6 +245,15 @@ public void setAssetStore(AssetStore assetStore) {
     this.assetStore = assetStore;
   }
 
+  @Reference(name = "eventHandlers", cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
```
lkiesow: Just wondering why you assign a name here? From what I can tell, `name = "eventHandlers"` isn't actually used anywhere. Is there a reason for doing so?

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/update/AssetManagerUpdateHandler.java
```diff
@@ -19,29 +19,10 @@
...
-    return true;
-  }
-
-  @Override
-  protected Session getSession() {
-    return session;
-  }
+import org.opencastproject.message.broker.api.assetmanager.AssetManagerItem;
 
+public interface AssetManagerUpdateHandler {
```
lkiesow, line 26: Easily confused with `AssetManagerUpdatedEventHandler`. Though I think this name is fine and that one should be renamed.

modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/impl/SchedulerServiceImpl.java
```diff
@@ -229,19 +225,24 @@
 
   /** The list of registered event catalog UI adapters */
   private List<EventCatalogUIAdapter> eventCatalogUIAdapters = new ArrayList<>();
+  private List<SchedulerUpdateHandler> schedulerUpdateHandlers = new ArrayList<>();
```
lkiesow: `ArrayList` is not thread safe: https://stackoverflow.com/a/2444195/2352895

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -133,6 +132,8 @@
 
   private AclServiceFactory aclServiceFactory;
 
+  private ArrayList<SeriesUpdateHandler> updateHandlers = new ArrayList<>();
```
lkiesow, line 124: Not thread safe: https://stackoverflow.com/a/2444195/2352895

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -653,6 +654,10 @@ public void repopulate(final AbstractSearchIndex index) throws IndexRebuildExcep
     }
   }
 
+  private void triggerEventHandlers(SeriesItem item) {
+
+  }
+
```
lkiesow: There is some code missing here.
Without this, no updates are triggered.

lkiesow (changes requested):
The patch generally looks hood and should (apart from a few minor things) work as expected. The structure itself is still very much like before, and from an architectural point of view, this partly leads to a somewhat weird structure. I'll go through details below. But despite that, this should already be a big improvement.


## Conductor & Message-Broker-API

It's a bit odd to see what code goes into which bundle. For example, the `AssetManager` now requires `UpdateHandler` defined in the `message-broker-api` which are then implemented in the `conductor` and execute code in other services.

Screenshot from 2021-11-08 20-02-35

- `message-broker-api` and `conductor` are now basically belong together as `api` and `impl` bundles, and the name should reflect that _if_ we want to keep both around.
- If we want to drop the “messaging” packages
    - Instead of putting the handler all in the `conductur` (or whatever we end up naming the `impl` bundle), we could also put the handler into the bundles they interact with. For example, we could put the OAI-PMH handler into the OAI-PMH publication service. Downside: We may need two implementations when we have remotes. But the services itself could just implement the handler interface, which would also make things simpler.
    - The handler interfaces could also go into the services they belong to. E.g. the `AssetManager` handler interface could go into the asset manager. The 
    - If we want that, this needs some more thoughts.

The placement of things in the `message-broker-api` is also a bit confusing. If there is a package `assetmanager`, I would have expected all asset manager related things in there:

Screenshot from 2021-11-08 21-43-16


## NessageItems

The `MessageItems` may have made sense with ActiveMQ, but now they are just an unnecessary layer of complexity. For example, the `AssetManager` would construvt a `MessageItem` by setting a type and putting specific data in that container, the Handler then needs to react to the type and extract the whole thing again:

a

Since we now have a direct connection and different interfaces for all “messages” anyway, why not just have different methods you could call for each handler? For example, the `AssetManagerUpdateHandler` interface should define `update)…=? and `delete(…)` which would directly accept the required data. That would be so much easier and less confusing:

b

## Indirect Messaging/Handler Chaining

We unnecessarily chain handlers. For example, the `AssetManager` will call a handler in the `conductor` which does nothing but to call the `OaiPmhHandler` in turn. That handler could just have implemented the `AssetManagerEventHandler` interface in the first place. The new code can now deal with several separate handlers for each event and the chaining is just confusing.

We can basically remove the `Composer` instances in the following diagram, which shows the new event handler structure:

Screenshot from 2021-11-08 22-38-44


## Dynamic vs Static Bindings

Using dynamic bindings like this is a nice way of making the whole thing flexible and easy to extend with additional receivers _but_ this comes at a cost:

1. With the message queue, before we had the guarantee that when the asset manager  would send a message, this message would be received by OAI-PMH eventually, regardless of the current state of that service. This promise does not hold any longer. Updates while booting up, for example, could now be easily lost. This wouldn't happen with a static binding which would give us the exact same guarantee.
    - The downside of static bindings would be that we need to explicitely declare all connections. But since we do not have many connections, it could be worth it if we avoid data loss.
2. Right now, we want specific connections, but loosing connections would be hard to notice. For example, I could move an `AssetManagerUpdateHandler` service to a worker node which has no asset manager and where the handler would just do nothing. It wouldn't complain, it just wouldn't work. A static binding would complain instantly, making it very obvious that there is a problem.


---

Sorry for the very long review. We don't have to change everything I described. I gave multiple contradicting ideas anyway ;-)

lkiesow:
Addition for the technical meeting on Nov 09, 2021. Discussing how things should look like in the end.
This is definitely out of scope for this pull request:

index

KatrinIhler:
+1 for static bindings. Otherwise, what Lars said. ;)

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
gregorydlogan: Habit more than anything.  IIRC it makes some of the reference display bits from Karaf easier to use, but definitely not required.

gregorydlogan:
Assuming I didn't mess anything up in the rebase this should be ready for review @lkiesow .

modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/update/ISeriesUpdateHandler.java
```diff
@@ -19,22 +19,10 @@
...
-
-  /**
-   * Try reconnecting if there was no reconnection attempt in the immediate past.
-   *
-   * @return If the re-connection attempt was successful.
-   */
-  boolean reconnect();
+import org.opencastproject.message.broker.api.series.SeriesItem;
 
+public interface ISeriesUpdateHandler {
```
geichelberger: I prefix

geichelberger:
Some interfaces have an I prefix.

gregorydlogan:
Bah, forgot to do that while I was squashing the changes again.  Fixed!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -171,13 +171,14 @@
 
   private static final String MANIFEST_DEFAULT_NAME = "manifest";
 
+  private ArrayList<AssetManagerUpdateHandler> handlers = new ArrayList<>();
```
geichelberger: I think that has to be threadsafe,  like a CopyOnWriteArrayList.

gregorydlogan:
I don't think we *need* to do this since I'm betting that other parts of the AM would fall over if truly running multithreaded, but it's also not going to hurt to be threadsafe anyway.

gregorydlogan:
@lkiesow I think I've satisfied everyone, this good to go?

gregorydlogan:
Rebased!

lkiesow:
I went through this again and re-thinking the current structure, I think that we could potentially lose data when (re-)starting Opencast since we have no guarantee that really any event handler is already registered when we try updating data. It also does not fix itself eventually, but we basically just lost the update. For more details, I've trued explaining the whole thing in more details here:

- [data.lkiesow.io/opencast/opencast-update-messages.mp4](https://data.lkiesow.io/opencast/opencast-update-messages.mp4)

Do you agree that this is a problem?

Now on to a suggestion for a simple way on how to fix this. Since we want to [eventually refactor the whole thing anyway](https://github.com/opencast/opencast/pull/3100/#issuecomment-964257638) and we know that static bindings are a problem right now due to some circular dependencies, I suggest keeping the fix as simple as possible for now.

That fix could be just waiting to send the message until all expected event handlers are available. Maybe a little bit nicer, but basically something like this:

```java
public void fireEventHandlers(AssetManagerItem item) {
  while (handlers.size() != 2) {
    logger.info("Not all event handlers are registered yet. Waiting for 10s.");
    Thread.sleep(10000);
  }
  for (AssetManagerUpdateHandler handler : handlers) {
    handler.execute(item);
  }
}
```

I'm also bringing this to list, hoping to get more opinions on this. Do you agree? Do you see a better way? Am I overreacting and this isn't actually a problem?

gregorydlogan:
This is a very good point.  It's an edge case in that you would have to have workflows in mid-process at startup, but it's quite true that this would either cause errors, or drop messages.

Your proposed solution would, in theory work.  It's a horrible hack, but hey this is just temporary.  My only concern would be if the system starts up in a way that it will never exit that loop - in that case you've got a dropped message.  At some point we have to draw a line and do *something* though, so I'm ok with the proposed solution.

lkiesow:
Added a minimal patch for some typos, … I found. Remove it if you don't agree with it [3c75bd4](https://github.com/opencast/opencast/pull/3100/commits/3c75bd47edbc58c4448f51a6cd9a446f013b640c)

gregorydlogan (commented):
LGTM

gregorydlogan:
This should be ready to go again.

lkiesow:
> I went through this again and re-thinking the current structure, I think that we could potentially lose data when (re-)starting Opencast since we have no guarantee that really any event handler is already registered when we try updating data. It also does not fix itself eventually, but we basically just lost the update. For more details, I've trued explaining the whole thing in more details here:
> 
>     * [data.lkiesow.io/opencast/opencast-update-messages.mp4](https://data.lkiesow.io/opencast/opencast-update-messages.mp4)
> 
> 
> Do you agree that this is a problem?
> 
> Now on to a suggestion for a simple way on how to fix this. Since we want to [eventually refactor the whole thing anyway](https://github.com/opencast/opencast/pull/3100/#issuecomment-964257638) and we know that static bindings are a problem right now due to some circular dependencies, I suggest keeping the fix as simple as possible for now.
> 
> That fix could be just waiting to send the message until all expected event handlers are available. Maybe a little bit nicer, but basically something like this:
> 
> ```java
> public void fireEventHandlers(AssetManagerItem item) {
>   while (handlers.size() != 2) {
>     logger.info("Not all event handlers are registered yet. Waiting for 10s.");
>     Thread.sleep(10000);
>   }
>   for (AssetManagerUpdateHandler handler : handlers) {
>     handler.execute(item);
>   }
> }
> ```
> 
> I'm also bringing this to list, hoping to get more opinions on this. Do you agree? Do you see a better way? Am I overreacting and this isn't actually a problem?

You never fixed this, did you? There are only two additional commits fixing merge conflicts. Or did I miss something?

gregorydlogan:
Ah, good point.  I've applied the hack :)

gregorydlogan:
... and fixed the tests to not be a runaway.  7+ hour build on my local infra.  Whoops! :D

lkiesow:
Looks broken, I'll investigate…

Screenshot from 2022-05-13 12-26-00

lkiesow:
Added logging for the event handlers.
Good thing we introduced this check, otherwise, we would already have lost messages.
It seems like only one event handler registers with the asset manager:

```
karaf@root()> log:display | grep AssetManagerImpl
2022-05-13 13:55:19,266 | INFO  | (AssetManagerImpl:206) - Activating AssetManager.
2022-05-13 13:55:21,526 | ERROR | (AssetManagerImpl:254) - Added class org.opencastproject.liveschedule.message.AssetManagerEventUpdateHandler event handler to asset manager
karaf@root()> 
```

lkiesow:
Seems like the conductor never activates its services:

```
karaf@root()> bundle:list |grep conductor
 80 x Active   x  82 x 12.0.0.SNAPSHOT            x Opencast :: conductor
karaf@root()> bundle:services 80
karaf@root()>
```

…which makes sense since the files have neither OSGi annotations nor are there any OSGi XML files to declare the services.

gregorydlogan:
Bah, I must have missed those in the rebase because it was working at that point. I'll try and get them done before the branch cut.

lkiesow:
Not really your fault. Seems like #3255 [broke this](https://github.com/opencast/opencast/pull/3255/files#diff-6903fd0392932ec6f36f3e69d22808c75924bff91b26321b1a0b0d247370a2d7).

lkiesow:
#3766 should fix the problem

lkiesow:
Merged `develop` which now includes #3766

lkiesow (approved):

PR #2669, opened 2021-06-02
Reworked Partial Import WOH to use concat demuxer

Arnei:
The concatenation in the Partial Import WOH can require a lot of working memory. Using the concat demuxer instead of concat filter should lower the load considerably.

To that end, all source videos need to be pre-encoded. This commit enforces the use of a preencode encoding profile. It also adds an encoding profile that ensures the necessary prerequisites for concat demuxing.

Resolves #2569.

To test this, create a workflow with the Partial Import WOH as per documentation. Start the workflow by creating a mediaPackage through the `/ingest` API, particularly `/addPartialTrack` to specify the start time of each video. 

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/#development-process/#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3106, opened 2021-10-28
Added back encoding profiles removed by previous commit

Arnei:
PR #2573 removed among others an encoding profile that was added in PR #2669. This PR aims at readding those encoding profiles.


lkiesow:
Are they used by default? Iirc, the idea was to ship only what is used by default.

Arnei:
They are not used by default in the sense that they are hardcoded into a WOH somewhere, which is why I added them to the docs of the WOH where they are relevant.

lkiesow:
Makes sense

PR #3107, opened 2021-10-29
Force same sample rate for audio files in PartialImport WOH encoding profiles

Arnei:
The PartialImport WOH does not only concatenate input files, but also self-generated ones, one of which are silent audio files. During concatenation of input audio files and generated silent audio files, "Non-monotonous DTS" warnings could appear and the output file would be distorted. For example, the output file could have a different length from the sum of its parts or the quality would be significantly lowered. Forcing the same sample rate alleviates the problem (thanks Lars!).

To test this PR, you want to test the PartialImport WOH with at least one video file, and one audio file which is shorter than the video files, so the operation is forced to generate some silent audio tracks.

Should this PR be approved, #3106 should be updated accordingly.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/encoding/opencast-movies.properties
+++ b/etc/encoding/opencast-movies.properties
@@ -167,7 +167,7 @@ profile.import.silent.name = Generate silent audio tracks for filling gaps
 profile.import.silent.input = nothing
 profile.import.silent.output = audio
 profile.import.silent.suffix = -silent-audio.mp4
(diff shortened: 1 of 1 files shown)


lkiesow (approved):

Rillke:
Ah, this probably caused some of the recordings to be asynchronous.

Rillke:
Why did you choose 44,1 kHz not 48 kHz btw?

Arnei:
To be perfectly honest, I didn't have a sophisticated reason. I simply looked at what sampling rate ffmpeg was generating the silent audio files with and chose that. If there's consensus that 48 kHz is preferred, I absolutely wouldn't mind changing to that.

PR #3126, opened 2021-11-02
Add config for service error states to service registry

KatrinIhler:
This modifies the use of one config for the service registry and adds another. With this, you can either disable error states for services completely or only for specific services. The warning state behavior remains unchanged. Default behavior is unchanged. Release Notes were added. Tests were added. :)

### DIFF
--- a/etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
+++ b/etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
@@ -4,9 +4,9 @@
 # Default: 10
 #max.attempts=10
 
-# Comma-separated list of services that should never go into error state.
+# Comma-separated list of service types that should never go into error state, e.g. org.opencastproject.assetmanager
 # Default: None
-#no.error.state.services=
+#no.error.state.service.types=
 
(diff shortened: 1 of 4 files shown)


etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
```diff
@@ -1,9 +1,13 @@
 # Configuration for the service registry
 
-# Number of failed jobs on a service before to set it in error state.
+# Number of failed jobs on a service before to set it in error state. Set to -1 to disable error states completely.
 # Default: 10
 #max.attempts=10
 
+# Comma-separated list of services that should never go into error state.
```
ts23: can you put an example what to specify as the service

ts23 (commented):
visually its good

etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
KatrinIhler: Added example and clarified config key. :)

gregorydlogan (approved):
LGTM

ts23 (approved):

PR #3149, opened 2021-11-09
Use original file extension in the asset manger

mtneug:
Up until now the asset manager might change the file extension based on
the mime type (or `.unknown` if the mime type is not known). This PR uses
the original file extention used during the upload and only relies on
known mime types if there is no file extension.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
@@ -244,7 +244,7 @@ public abstract class AbstractAssetManagerRestEndpoint extends AbstractJobProduc
   }
 
   @GET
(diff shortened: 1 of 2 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -310,11 +311,12 @@ public Response getAsset(@PathParam("mediaPackageID") final String mediaPackageI
             }
           }
 
-          final String fileName = mediaPackageElementID
-                  .concat(".")
-                  .concat(asset.getMimeType().bind(suffix).getOr("unknown"));
+          if (StringUtils.isEmpty(fileName)) {
```
lkiesow: Haven't actually run the code, but looks good.
Minor suggestion, also prevent blank file names like "`  `".

```suggestion
          if (StringUtils.isBlank(fileName)) {
```

PR #3156, opened 2021-11-09
Add S3 version of duplicate-event

mtneug:
`duplicate-event` only uses the download distribution. The new operation
`duplicate-event-aws` uses the `aws` distribution instead.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/workflow-workflowoperation/pom.xml
+++ b/modules/workflow-workflowoperation/pom.xml
@@ -169,6 +169,7 @@
               OSGI-INF/operations/copy.xml,
               OSGI-INF/operations/defaults.xml,
               OSGI-INF/operations/duplicate-event.xml,
+              OSGI-INF/operations/duplicate-event-aws.xml,
               OSGI-INF/operations/error-resolution.xml,
               OSGI-INF/operations/export-wf-properties.xml,
               OSGI-INF/operations/failing.xml,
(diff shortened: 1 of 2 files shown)


ts23:
I was under the impression that duplicate event works on the asset manager level rather than the distribution side of things, so I don't quite understand why we duplicate-event should distribute in the first place. I think we should handle the distribution in a start-workflow wfh started by the duplicate-event-wf?

mtneug:
@ts23 the `duplicate-event` operation will only [publish the new event to the `internal` channel](https://github.com/opencast/opencast/blob/d3e361cdd4ad0fdddc8b8bce89b7b7bf0a8cab8e/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java#L453-L505). Yes, we could remove that and then handle everything in the `start-workflow` operation. The latter did not exist when `duplicate-event` was first implemented.

One problem might be that typically the elements for the `internal` publication are not archived making it harder (impossible?) to publish them again. But maybe this is not a good enough argument and we should archive those elements?

ts23:
If I understand it correctly we duplicate from archive, so we get the stuff out of the archive and put the copies back into the archive to be processed later so we can do whatever we like with the duplicates, ie. running an arbitrary WF on them.
 
I think I turned the functionality of duplicate-wf around:
 the original:  ...  make copy(ies) and modify the original  ... to ... make copy(ies) and keep orig and run a WF on all copies

ts23:
we could actually consider sticking the startWF into the duplication

mtneug:
Summary from the OC dev meeting: `duplicate-event` should no longer deal with publications and we thus don't need two versions of `duplicate-event`.

PR #3159, opened 2021-11-09
Remove check for tracks for OAI-PMH

mtneug:
OAI-PMH can be used to only publish metadata so there shouldn't be a
track check as in the other publication types. The check is only done
during the republish anyway so the initial publish already allows for
track-less publications.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


mliradelc (commented):
Approved, I agree with @mtneug over this. Also the exception is too generic to have any use of.

I will wait a little if there are other opinions about this.

mliradelc (approved):
Approved, I agree with @mtneug over this. Also the exception is too generic to have any use of.

I will wait a little if there are other opinions about this.

KatrinIhler (approved):
I think this is pretty non-controversial, let's merge this :)

PR #3161, opened 2021-11-10
Removed Solr from Series Service

Arnei:
Removes all Solr functionality from the series service. Functions that relied on the solr index were either removed or now query the database directly.

- Moved the series filter query endpoint to the external api, while trying to preserve as much functionality as possible (for example, there is no pagination, so instead an offset variable is offered)

This PR can wait until Opencast 12, it is mainly open for reviewing purposes.

Work sponsored by our generous Crowdfunding initiatives!

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests


mtneug:
Do you expect/observe performance degradations? Should this maybe instead query Elastic-/OpenSearch (though I do expect the DB to scale quite good for normal queries)?

Arnei:
I did not observe performance degradations, but I also only tested this on my local machine, where database and server run on the same hardware. However, the DB queries are quite simple, so I also expect this to scale well.

Lars and I had discussed that we did not want to query the Elastic-/OpenSearch, in order to get away from the currently somewhat cyclic dependency where the External API and the Series Service are trying to get information from each other.

mliradelc (changes requested):
Thanks for the PRs 😄 

Because of the changes in the pagination feature in the api, A comment in the documentation and in the changelog is required

mliradelc:
I will Merge it, but this needs the notes to be added before the release in one month ;)

Arnei:
Sorry for the late response @mliradelc , we discussed in the technical meeting that it would be fine for this PR to wait for Opencast 12, but I forgot you weren't there. I was gonna add the notes after the branch cut, especially since Matthias also mentioned that this changes the External API and would therefore need a new API version (*sigh*). I think it would be better to revert the merge if can still reasonably do so?

mliradelc:
Hi @Arnei  thanks for your comment, of course we can revert the commit 🙂 , ~~I will do that after the tech meeting of today and mark it as a draft. Feel free to undraftit  when you feel that is ready for review.~~

After the meeting was decided to hold the revert because more info is needed.

Update: After conversation with @Arnei this PR will be reverted

PR #3172, opened 2021-11-15
Improve Admin UI Performance

KatrinIhler:
This PR contains multiple changes to improve the performance of the admin UI for bigger systems (in this case SWITCH).

The commits in detail:
1. When loading the metadata for new or existing events, fill the series collection only once, since this might take a while if you have lots of series. To do this, we hand over a custom list provider query if we only want to show series with write access.
2. Don't load event metadata just to get the series list provider for a query, since this fills all of the metadata collections, which is expensive. Instead, just get depend on the series list provider explicitly.
3. Some aesthetic improvements suggested by Intellij, including removal of unused parameters.
4. Remove some filters for the Admin UI tables. _This change might be controversial._  This removes either filters I deemed unnecessary (mostly name filters - why list all of the names in a drop down when they are right there in the table, and can be filtered with the search field?)  or that are very expensive (all filters that list all users or roles in the system). All of these fields should still be searchable with a full text query in the search field. This will result in some tables (e.g. ACL templates) to not have filters at all.
5. The last is a bug fix for multi tenant systems that was necessary to get this to work, since I earlier changed the order of loading common and extended metadata. With this change, we will get and set the common UI catalog adapters in the index service by organization, instead of overwriting the one adapter again and again so the last organization to be loaded "wins".

Since these changes build up on each other, I would rather not split them apart. They are contained in separate commits, so I think they are still easy enough to review. However, I could file No. 4 separately if it's not deemed appropriate for a minor release.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -86,8 +86,10 @@ import org.opencastproject.index.service.exception.UnsupportedAssetException;
 import org.opencastproject.index.service.impl.util.EventUtils;
 import org.opencastproject.index.service.resources.list.provider.EventsListProvider.Comments;
 import org.opencastproject.index.service.resources.list.query.EventListQuery;
+import org.opencastproject.index.service.resources.list.query.SeriesListQuery;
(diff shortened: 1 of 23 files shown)


gregorydlogan:
IMO commit 4 is fine.  The filters you removed are either things which would not occur to me to filter anyway.  Worst case I'm expecting to do a 10.7 at some point so we can bring 'em back then if it's an issue.

PR #3174, opened 2021-11-15
Fix NullPointerException in `CachedJWT` if `exp` claim does not exist

LukasKalbertodt:
The `exp` claim is optional: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4
If it is not present, the token is treated as not having an expiration
time.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/CachedJWT.java
+++ b/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/CachedJWT.java
@@ -57,7 +57,7 @@ public class CachedJWT {
    * @return Boolean indicating whether the cached JWT has expired.
    */
   public boolean hasExpired() {
-    return !this.expiresAt.after(new Date());
+    return this.expiresAt != null && !this.expiresAt.after(new Date());
   }
 
   /**
--- a/modules/security-jwt/src/test/java/org/opencastproject/security/jwt/JWTGenerator.java
+++ b/modules/security-jwt/src/test/java/org/opencastproject/security/jwt/JWTGenerator.java
@@ -132,6 +132,21 @@ public final class JWTGenerator {
     );
   }
 
+  public String generateValidNonExpiringSymmetricJWT() {
+    return generateValidNonExpiringJWT(getSymmetricAlgorithm());
+  }
+
+  private String generateValidNonExpiringJWT(Algorithm algorithm) {
+    return JWT.create()
+        .withIssuer(issuer)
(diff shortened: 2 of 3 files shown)


joneug:
Besides the `CachedJWT` class, validation also happens in the `JWTVerifier` class in [line 130](https://github.com/opencast/opencast/blob/develop/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java#L130) (using the method `verify` implemented in the `java-jwt` library). I assume that the library is handling a missing expiration according to the specification. But I did not test it / look it up in the library's source code. Maybe we should add a test case for that in `JWTVerifierTest`? And also in `JWTLoginTest` for the case that the JWT was cached?

LukasKalbertodt:
The `verify` method (used when there is no cached JWT) is working. I noticed this bug when I tried one JWT -> it worked, then tried the exact same request again -> NullPointerException.

joneug:
Ok, great! I still suggest adding a test case for that. The following code tests exactly the flow you described (first request without cache, second request with cache).

In `JWTGenerator`:

```java
  public String generateValidNonExpiringSymmetricJWT() {
    return generateValidNonExpiringJWT(getSymmetricAlgorithm());
  }

  private String generateValidNonExpiringJWT(Algorithm algorithm) {
    return JWT.create()
        .withIssuer(issuer)
        .withAudience(clientId)
        .withClaim(usernameKey, username)
        .withClaim(nameKey, name)
        .withClaim(emailKey, email)
        .withClaim(rolesKey, roles)
        .sign(algorithm);
  }
```

In `JWTLoginTest`:

```java
  @Test
  public void testNonExpiringLoginWithCache() {
    Object username;
    String jwt = generator.generateValidNonExpiringSymmetricJWT();

    username = authFilter.getPreAuthenticatedPrincipal(
        mockRequest(jwt)
    );
    assertEquals(username, generator.getUsername());

    // Make sure that the cache is used and no calls to the mocked user reference provider are made
    reset(userReferenceProvider);

    username = authFilter.getPreAuthenticatedPrincipal(
        mockRequest(jwt)
    );
    assertEquals(username, generator.getUsername());
  }
```

LukasKalbertodt:
Oh yeah totally! Sorry, didn't want to sound like tests were not useful. And thanks a ton for providing those tests ^_^ I pushed another commit with exactly your tests.

joneug (approved):

PR #1285, opened 2019-12-11
Implement save buttons for metadata

pmiddend:
This PR changes the way metadata is updated in the Admin UI. Instead of the metadata fields being updated *immediately*, there’s a “Save metadata” button at the bottom to bulk-save the metadata. This is consistent with the already-existing bulk update metadata modal (and this modal was also adapted).

Editing a metadata field now doesn’t produce a little checkmark at the right, but a red “is dirty” symbol that I picked at random.

Also included in this PR is the ability to trigger a workflow on all affected events. Meaning: all events in the series in case of a series metadata change, a single event in case of the normal “Edit event” modal or all selected events in case of the “Bulk edit events” modal.

Editing metadata and then trying to exit the modal results in a prompt to confirm unsaved changes.

Also fixed is the “is a workflow running on the event” detection, which behaved weirdly.

Sneakily, this also introduces a small style change to the Javascript DI mechanism, adding more newlines. This hopefully makes merge conflicts easier to resolve and is a bit easier to read.

This work is sponsored by the University of Vienna as part of the 2019/2020 crowdfunding campaign.

PR #3176, opened 2021-11-16
Save Buttons for Metadata

KatrinIhler:
This PR introduces save buttons for the metadata of existing series and events in the Admin UI, similarly to what #1285 attempted. This change affects both the common and extended metadata tab. Now, when changing the value of a metadata field, instead of directly sending a request to the back end, the field is only marked as dirty. The request is only sent when the user clicks the save button. If the user attempts to close the window with unsaved changes, a window with a warning will pop up.
I changed the tests accordingly and added a few new ones, hopefully covering most cases where things could go wrong. But I didn't manage to actually test this today, so this is a draft PR for now. (@gregorydlogan This way you can already take a look at it without having to wait for tomorrow. :))
However, an older version of the series changes has already been running on a production system for over a month and the code is very similar to the one for the event metadata, so hopefully there won't be too many nasty surprises. I'm gonna take another look at this tomorrow and then also add some screenshots.

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -733,7 +733,7 @@ angular.module('adminNg.controllers')
 
     $scope.close = function() {
       if (($scope.unsavedChanges([$scope.commonMetadataCatalog]) === false
-           && $scope.unsavedChanges($scope.extendedMetadataCatalogs))
+           && $scope.unsavedChanges($scope.extendedMetadataCatalogs)  === false)
           || confirmUnsaved()) {
         Modal.$scope.close();
       }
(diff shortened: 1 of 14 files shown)


KatrinIhler:
Alright, here are the promised screen shots. (The events metadata tab was changed accordingly.)

At the beginning, the save button is disabled.
save-button-disabled

 When changing the value of a metadata field, that field is marked dirty, and the save button becomes enabled. (Changing the value back will unmark the field again.)
field-marked-dirty

When clicking the save button, the field is unmarked and a notification is displayed on top of the tab.
saved-notification

Trying to close the window with unsaved changes will result in a warning.
warning-unsaved-changes

KatrinIhler:
Found two minor bugs while testing, but I think now it's ready. I also tested extended metadata. (Two extended metadata catalogs for series doesn't seem to work well, but since it's already falling apart when creating the series, I don't think that's my fault.)

gregorydlogan:
This appears to work as is says on the tin.  @mtneug can you verify that extended metadata catalogs still work as expected with your (complicated) setup?  I'm not hugely familiar with the extended metadata side of things but this *looks* right to me.

gregorydlogan:
@KatrinIhler does this mean we can (finally!) close #1285?

KatrinIhler:
@gregorydlogan Well, technically that PR does a little bit more - e.g. it also modifies the bulk event metadata editing (which I forgot), but more importantly it adds the ability to trigger a workflow on all affected events.

But I don't really care - we can always salvage parts from the PR after we close it, and this was the most important feature after all.

gregorydlogan:
Can discuss next Tuesday 👍

PR #3186, opened 2021-11-19
Use comma as separator for WFR cleanup config

lwillmann:
It is possible to specify multiple collections for the Working File Repository cleanup. The comment on how to configure the cleanup in the custom.properties file says: 
_A comma separated lists of collections in the working file repository that should be cleaned up._ 
but actually the split operation in the code expects spaces between each collection. Thus this PR changes the implementation to match the documentation.  

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
+++ b/modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
@@ -180,7 +180,7 @@ public class WorkingFileRepositoryImpl implements WorkingFileRepository, PathMap
     // Determine which collections should be garbage collected
     List<String> collectionsToCleanUp = null;
     String[] cleanupCollections = StringUtils.split(StringUtils.trimToNull(
-            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)));
+            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)),',');
     if (cleanupCollections != null) {
       collectionsToCleanUp = Arrays.asList(cleanupCollections);
     }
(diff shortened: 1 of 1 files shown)


modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
```diff
@@ -180,7 +180,7 @@ public void activate(ComponentContext cc) throws IOException {
     // Determine which collections should be garbage collected
     List<String> collectionsToCleanUp = null;
     String[] cleanupCollections = StringUtils.split(StringUtils.trimToNull(
-            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)));
+            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)),',');
```
mtneug: Could you add a space:

```suggestion
            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)), ',');
```

mtneug (changes requested):

modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
```diff
@@ -180,7 +180,7 @@ public void activate(ComponentContext cc) throws IOException {
     // Determine which collections should be garbage collected
     List<String> collectionsToCleanUp = null;
     String[] cleanupCollections = StringUtils.split(StringUtils.trimToNull(
-            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)));
+            cc.getBundleContext().getProperty(WORKING_FILE_REPOSITORY_CLEANUP_COLLECTIONS_KEY)), ',');
```
KatrinIhler: I did this recently in #3126 like this:

```
String noErrorStateServiceTypesStr = StringUtils.trimToNull((String) properties.get(
            NO_ERROR_STATE_SERVICE_TYPES_CONFIG_KEY));
    if (noErrorStateServiceTypesStr != null) {
      noErrorStateServiceTypes = Arrays.asList(noErrorStateServiceTypesStr.split("\\s*,\\s*"));
```

The advantage of this is that you can also use white spaces after the commas without affecting the results. It also uses the built-in String split functionality - though in this case you have to check for null before splitting, and handle that case accordingly.

lwillmann: @KatrinIhler Yes, this is a better solution.

PR #3197, opened 2021-11-22
Elasticsearch Retry Config

KatrinIhler:
This adds configuration options to Elasticsearch for retrying failed requests. This is helpful if the index is temporarily unavailable, e.g. if Elasticsearch doesn't have enough heap to fulfill the request. You can set how many times a retry will be attempted, as well as how long to wait between attempts. These can be configured separately for simple get requests and for requests that also update or delete, since the latter are usually more important. Note that a get request that is necessary for an update will be considered part of the update, so those config settings apply.

Additionally I cleaned up some code (mostly recommendations from Intellij), and also made the index name (not identifier) changeable during runtime. Updates to the documentation and text for the release notes are included.

@wsmirnow asked me to aim this at OC 10, but as this depends on #2857, that's unfortunately not possible. Not sure if OC 11 is the right branch, since it isn't out yet, or if this would be better aimed at develop because this adds functionality to existing code. Any opinions?

### DIFF
--- a/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
+++ b/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
@@ -7,11 +7,11 @@
 # Default: opencast
 #index.identifier=opencast
 
+# Config that can be updated during runtime:
+
 # The index name. Will be used for logging.
 # Default: Elasticsearch
 #index.name=Elasticsearch
-
-# Config that can be updated during runtime:
 
 # How many times to retry get requests to the Elasticsearch in case of ElasticsearchStatusException.
 # Default: None
--- a/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
+++ b/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
@@ -14,12 +14,12 @@
 #index.name=Elasticsearch
 
 # How many times to retry get requests to the Elasticsearch in case of ElasticsearchStatusException.
-# Default: None
+# Default: 0
 #max.retry.attempts.get=0
 
 # How many times to retry update requests to the Elasticsearch in case of ElasticsearchStatusException (includes reads
 # necessary for updates).
-# Default: None
+# Default: 0
 #max.retry.attempts.update=0
 
 # How long to wait between retry attempts of get requests (in milliseconds).
--- a/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
+++ b/etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
@@ -1,5 +1,7 @@
 # Elasticsearch index configuration
 
+# Config that cannot be updated during runtime:
+
 # The base identifier of the Elasticsearch index. Subindices will be named after the following schema:
 # identifier_subindex, e.g. opencast_event. If an index with this name doesn't exist, Opencast will create it.
 # Default: opencast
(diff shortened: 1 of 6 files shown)


mliradelc:
Hi Kathrin, @joneug and me as release Managers of version 11 we decided to include your PR into opencast 11, can you target to `r/11.x`?

KatrinIhler:
> Hi Kathrin, @joneug and me as release Managers of version 11 we decided to include your PR into opencast 11, can you target to `r/11.x`?

Already am. :)

etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg
```diff
@@ -1,10 +1,31 @@
...
 #index.identifier=opencast
 
+# Config that can be updated during runtime:
+
 # The index name. Will be used for logging.
 # Default: Elasticsearch
 #index.name=Elasticsearch
+
+# How many times to retry get requests to the Elasticsearch in case of ElasticsearchStatusException.
+# Default: None
```
gregorydlogan: I'll fix these after merge, but IMO defaults should match verbatim with the setting below it.  You and I can parse None == 0, but someone who *isn't* as fluent might not be able to.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -130,15 +139,39 @@ public void activate(ComponentContext ctx) throws ComponentException {
...
+   * OSGi callback for configuration changes.
+   *
+   * @param properties
+   *          The configuration
+   */
+  @Modified
+  public void modified(Map<String, Object> properties) {
+    super.modified(properties);
+
+    maxRetryAttemptsGet = NumberUtils.toInt((String) properties.get(MAX_RETRY_ATTEMPTS_GET_PROPERTY),
```
gregorydlogan, line 160: Out of curiosity, why are these here rather than in the superclass?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -130,15 +139,39 @@ public void activate(ComponentContext ctx) throws ComponentException {
...
+            DEFAULT_MAX_RETRY_ATTEMPTS_GET);
+    retryWaitingPeriodGet = NumberUtils.toInt((String) properties.get(RETRY_WAITING_PERIOD_GET_PROPERTY),
```
gregorydlogan, line 162: I have to be that guy, but `max.waiting.*=-1` does bad things since Thread.sleep throws an `IllegalArgumentException` for a negative waiting time.

Further objections: While *technically* correct in that a maximum retry count of -1 never retries, I'm pretty sure we should throw an exception, or at least warn that this is a dumb idea.  Also, setting non-integer values yeilds 0s, but I'm not sure there's much you can do with that since you're using NumberUtils.

gregorydlogan (changes requested):

mliradelc:
Last or two weeks ago, we talked in the dev meeting about this PR to be delayed for the next minor release. I will leave as a draft

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
KatrinIhler: The reasoning was that properties that are used by the logic of the super class will get read in by that one, while the others will be left to the sub class. The sub class decides which configured retry attempts/waiting period are passed over to the super class as method parameters depending on if it's a read or update, which the super class doesn't know about, so I figured these would be mainly the sub class's responsibility. 

But let's be honest, the whole separation between those two classes is a bit messy anyway. I tried to straighten it out a bit in the last few PRs, but... yeah.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
KatrinIhler: The code actually checks for negative waiting periods:
```
          if (retryWaitingPeriod > 0) {
            Thread.sleep(retryWaitingPeriod);
          }
```

But I'm gonna add a warning for negative configured values.

The waiting period is in milliseconds, that should be enough granularity for people... :laughing:

KatrinIhler:
@gregorydlogan Finally addressed all your findings, ready for review again :)

PR #3198, opened 2021-11-23
Fixed Admin UI Endpoint Configuration not loading on Opencast startup

Arnei:
As mentioned in PR #3049 by @KatrinIhler, said PR caused some of the admin ui endpoint configuration files to not be read on Opencast start-up anymore, but only when the configuration files were modified at runtime. This should fix that.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
@@ -124,15 +124,11 @@ import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Date;
-import java.util.Dictionary;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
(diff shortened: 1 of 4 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
```diff
@@ -239,6 +243,12 @@ protected void activate(ComponentContext cc) {
       logger.debug("Configured server url is {}", ccServerUrl);
       if (ccServerUrl != null)
         this.serverUrl = ccServerUrl;
+
+      Dictionary dict = cc.getProperties();
+      List<String> keys = Collections.list(dict.keys());
+      Map<String, Object> dictCopy = keys.stream()
+              .collect(Collectors.toMap(Function.identity(), dict::get));
+      modified(dictCopy);
```
KatrinIhler: You can also add the properties to the activate method as a parameter, then you don't need to transform the dictionary to a map. :) Both activate() and modified() can take multiple parameters, e.g. component context, properties or bundle context.

KatrinIhler (changes requested):
Just one thing, rest looks good.

KatrinIhler (approved):
Looks good now, merging! :D

PR #3203, opened 2021-11-25
start-workflow-wfh allows to pass mediapackage id from previous WFH

ts23:
Update StartWF WFH to allow passing MP ids in from previous WF operations. this can be useful for Duplicate MP workflows that can then start other operation on the duplicates

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -16,8 +16,8 @@ For each duplicated event the new media package ID is stored as a workflow prope
 
 |Name                                    |Example                                                             |Description                                    |
 |----------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|
-|duplicate\_media\_package\_*number*\_id |`duplicate\_media\_package\_1\_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
-|duplicate\_media\_package\_ids          |`duplicate\_media\_package\_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
+|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
+|duplicate\_media\_package\_ids          |`duplicate_media_package_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
 
 \* will be deprecated
 
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -16,7 +16,10 @@ For each duplicated event the new media package ID is stored as a workflow prope
 
 |Name                                    |Example                                                             |Description                                    |
 |----------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|
-|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       |
+|duplicate\_media\_package\_*number*\_id |`duplicate\_media\_package\_1\_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
+|duplicate\_media\_package\_ids          |`duplicate\_media\_package\_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
+
+\* will be deprecated
 
 Parameter Table
 ---------------
--- a/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
@@ -20,7 +20,7 @@ The StartWorkflowWorkflowOperationHandler can be used to start a new workflow fo
 <operation id="start-workflow">
   <configurations>
     <configuration key="workflow-definition">fast</configuration>
-    <configuration key="media-packages">${duplicate\_media\_package\_ids}</configuration>
+    <configuration key="media-packages">${duplicate_media_package_ids}</configuration>
     <configuration key="key">value</configuration>
     <configuration key="publish">true</configuration>
   </configurations>
--- a/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
@@ -8,7 +8,7 @@ The StartWorkflowWorkflowOperationHandler can be used to start a new workflow fo
 
 |Configuration Key         |Example                              |Description                                     |
 |--------------------------|-------------------------------------|------------------------------------------------|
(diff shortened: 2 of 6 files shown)


mtneug:
@ts23 why not simply use `<configuration key="media-package">${duplicate_media_package_1_id}</configuration>`?

ts23:
> @ts23 why not simply use `<configuration key="media-package">${duplicate_media_package_1_id}</configuration>`?

doesn't solve the issue of multiple mps being created

mtneug:
Alternative approach to keep it consistent with other operations:

- extend the `duplicate-event` operation with an additional variable `duplicate_media_package_ids` that holds a comma-separated list of the new media package ids
- change the `media-package` property of the `start-workflow` operation such that multiple comma-separated media package ids can be supplied similar to how the common `source-flavors` property works
- it might be a good idea to change/deprecate the property name from `media-package` to `media-packages` (again `source-flavor` -> `source-flavors` serves as a precedent)

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
```diff
@@ -104,22 +97,33 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
       throw new WorkflowOperationException(format("Workflow Definition '%s' not found", configuredWorkflowDefinition));
     } catch (Exception e) {
       throw new WorkflowOperationException(e);
     }
 
+    for (String mpId : asList(configuredMediaPackageIDs)) {
+      // Get media package
+      Opt<MediaPackage> mpOpt = assetManager.getMediaPackage(mpId);
+      if (mpOpt.isNone()) {
+        throw new WorkflowOperationException(format("Media package %s not found", mpId));
```
mtneug: Maybe it would be better to record the error but continue with the other IDs and only then fail? Or do you think a fast fail would be better?

etc/workflows/duplicate-event.xml
```diff
@@ -59,6 +59,14 @@
...
     </operation>
 
+    <operation
+      id="start-workflow">
+      <configurations>
+        <configuration key="media-packages">${duplicate_media_package_ids}</configuration>
+        <configuration key="workflow_definition">fast</configuration>
+      </configurations>
+    </operation>
+
```
mtneug, line 69: I assume this is only for the review? Could you remove it for the merge?

mtneug (changes requested):
The feature works and only have some small points.

etc/workflows/duplicate-event.xml
ts23: I was going to leave the example in but it be default. Or do we not provide usage examples?

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
ts23: we should make that optional on fail_on_error (fast fail if fail-on-error = true, record and continue if fail-on-error=false)

gregorydlogan:
Does this related to #3155?

mtneug:
No this is for the `start-workflow` operation. #3155 allows access to organisation properties in workflows.

etc/workflows/duplicate-event.xml
mtneug: I would remove it as this is most likely used in production and not everybody will use `fast` for that. In any case, `workflow_definition` should be `workflow-definition`.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
```diff
@@ -51,7 +53,7 @@
   private static final Logger logger = LoggerFactory.getLogger(StartWorkflowWorkflowOperationHandler.class);
 
   /** Name of the configuration option that provides the media package ID */
-  public static final String MEDIA_PACKAGE_ID = "media-package";
+  public static final String MEDIA_PACKAGE_ID = "media-packages";
```
mtneug: This is a breaking change that is not allowed in OC 11. I would suggest only deprecating the `media-package` option but still allowing it. It should be possible to simply merge the `media-packages` and `media-package` strings possible with a seperator.

mtneug (changes requested):
This works, but there are still two remaining issues I found. I'm sorry for the late review.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/StartWorkflowWorkflowOperationHandler.java
ts23: should be fixed

docs/guides/admin/docs/workflowoperationhandlers/start-workflow-woh.md
```diff
@@ -20,7 +20,7 @@ The StartWorkflowWorkflowOperationHandler can be used to start a new workflow fo
 <operation id="start-workflow">
   <configurations>
     <configuration key="workflow-definition">fast</configuration>
-    <configuration key="media-package">e72f2265-472a-49ae-bc04-8301d94b4b1a</configuration>
+    <configuration key="media-packages">${duplicate\_media\_package\_ids}</configuration>
```
mtneug: `_` in backticks does not need to be escaped:

```suggestion
    <configuration key="media-packages">${duplicate_media_package_ids}</configuration>
```

docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
```diff
@@ -16,7 +16,10 @@ For each duplicated event the new media package ID is stored as a workflow prope
 
 |Name                                    |Example                                                             |Description                                    |
 |----------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|
-|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       |
+|duplicate\_media\_package\_*number*\_id |`duplicate\_media\_package\_1\_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
+|duplicate\_media\_package\_ids          |`duplicate\_media\_package\_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
```
mtneug: `_` in backticks does not need to be escaped:

```suggestion
|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
|duplicate\_media\_package\_ids          |`duplicate_media_package_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
```

mtneug (changes requested):
Only some nitpicking in the documentation. Can go in otherwise.

Thanks for the patience.

mtneug (approved):

etc/workflows/duplicate-event.xml
```diff
@@ -59,6 +59,14 @@
       </configurations>
     </operation>
 
+    <operation
+      id="start-workflow">
+      <configurations>
+        <configuration key="media-packages">${duplicate_media_package_ids}</configuration>
+        <configuration key="workflow-definition">fast</configuration>
```
KatrinIhler, line 66: The WOH itself is cool, but I don't agree with this change to the workflow definition. We shouldn't include fast, since it's a testing workflow and people shouldn't run this by accident on their production cluster (as just happened to us), and also - why include an import workflow here? Since the initial event that was duplicated was already processed, the import has already happened, the only thing we might want to do is publish - but I wouldn't do that by default either.

PR #3205, opened 2021-11-30
Fix Date Time Picker

### LINKED ISSUE #1976: Admin UI: Add event -> Source -> Recording metadata -> Start date's picker bulky and text labels covered by controls
**Describe the bug**
I see 5 issues here, all about the picker so they go into a single report.

When opening the start date's picker in Admin UI Add event -> Source -> Recording metadata, the **now** and **done** buttons are bulky (1) so they do not fit into one line. The time sliders cover label texts (2). Additionally, **Microsecond** and **Millisecond** are probably only required in very rare cases (even **Second** is something I never required) (3). If someone needs such a granularity, which is not reflected in the text field (4) and therefore can't be precisely chosen by the user, they should be given the opportunity to enter this in the text field through keyboard, however, on my screen the text field is **covered** by the bulky picker (5).

**To Reproduce**
1. Go to https://develop.opencast.org/admin-ng/index.html#!/events/events?modal=new-event-modal&tab=source
2. Click on **Add event**
3. Add a title and proceed to **Source** and set focus to the **Start date** field
Screenshot 2020-10-18 at 09 59 14

**Environment Information**
 - OS: OSX Catalina
 - Firefox
 - Opencast version with minor release `Opencast 10.0.0.SNAPSHOT - ad548b9`

lkiesow:
This patch fixes the interface of the date time picker which was missing
all its style attributes since they were not properly included in bower
and was additionally partially overwritten by random other attributes.

This also caused hidden fields like microseconds to show up which was
not helpful.

To make things even simpler, this also switches from the slider
interface for times to the selector interface, which is much more
common.

screenshot-from-2021-11-29_22-39-16

_Other than depicted, this is now asking for hour and minute only as discussed in the technical meeting._

This fixes #1976

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/directives/datetimepickerDirective.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/directives/datetimepickerDirective.js
@@ -61,9 +61,14 @@ angular.module('adminNg.directives')
         };
 
 
-        var optionsObj = {};
-        optionsObj.timeInput = true;
-        optionsObj.showButtonPanel = true;
+        var optionsObj = {
+          controlType: 'select',
+          showMillisec: false,
+          showMicrosec: false,
+          showTimezone: false,
+          oneLine: true,
+          timeFormat: 'HH:mm'
+        };
         optionsObj.onClose = function () {
           var newDate = element.datetimepicker('getDate');
           setTimeout(function(){
--- a/modules/admin-ui-frontend/app/index.html
+++ b/modules/admin-ui-frontend/app/index.html
@@ -6,7 +6,7 @@
     <meta name="description" content="">
     <meta name="viewport" content="width=device-width">
     <!-- Place favicon.ico and apple-touch-icon.png in the root directory -->
-    <!-- build:css styles/vendor.css -->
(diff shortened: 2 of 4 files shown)


modules/admin-ui-frontend/app/scripts/shared/directives/datetimepickerDirective.js
```diff
@@ -61,9 +61,14 @@ angular.module('adminNg.directives')
...
-        optionsObj.timeInput = true;
-        optionsObj.showButtonPanel = true;
+        var optionsObj = {
+          controlType: 'select',
+          showMillisec: false,
+          showMicrosec: false,
+          showTimezone: false,
+          oneLine: true,
+          timeFormat: 'HH:mm'
+        }
```
geichelberger: The pipeline fails because of the missing semicolon.

lkiesow: Yep, seems like the old eslint version in 10.x has slightly different rules about this.
But it's already fixed.

PR #3218, opened 2021-12-01
Opencast Plugin Manager

lkiesow:
This patch add a new plugin manager, allowing users to easily enable or
disable certain parts of Opencast. This is similar to (and in fact uses)
the Karaf feature configuration with the huge benefit that users to not
have to configure exact versions and keep that updated. Instead they
just need to specify the plugin they want to load.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


geichelberger:
The transcription-service-workflowoperation bundle references the transcription-service-amberscript bundle, which results in a missing requirement if you enable only the google-speech plugin.

lkiesow:
Great, sorry. That's definitely a weird dependency. I'll look into it and fix it in some way.

modules/plugin-manager/src/main/java/org/opencastproject/plugin/impl/PluginManagerImpl.java
```diff
@@ -0,0 +1,152 @@
...
+            .filter(e -> BooleanUtils.toBoolean(Objects.toString(e.getValue(), "")))
+            .map(Map.Entry::getKey)
+            .collect(Collectors.toSet());
+    logger.debug("Active plugin configuration: {}", activePlugins);
+
+    // Verbose Karaf logs active?
+    verbose = BooleanUtils.toBoolean(Objects.toString(properties.get(VERBOSE)));
+
+    Thread thread = new Thread(new PluginStarter());
+    thread.start();
```
geichelberger, line 85: What happens if the thread is still running and the configuration changes?

geichelberger:
This PR #3718 should solve the dependency issue. It would be cleaner to split the workflow operation bundle but having three bundles with just one or two classes seems odd.

lkiesow:
Rebased on top of `develop` and fixed all merge conflicts.
Should hopefully now work as expected.

lkiesow:
> This PR #3718 should solve the dependency issue. It would be cleaner to split the workflow operation bundle but having three bundles with just one or two classes seems odd.

Seems like that one did not help. Even after the rebase, you get the error. Maybe I just turn this into one transcription service plugin for now…

```
karaf@root()> bundle:list  |grep transc
401 x Active   x  82 x 12.0.0.SNAPSHOT            x Opencast :: transcription-service-api
402 x Active   x  82 x 12.0.0.SNAPSHOT            x Opencast :: transcription-service-google-speech-impl
403 x Active   x  82 x 12.0.0.SNAPSHOT            x Opencast :: transcription-service-persistence
404 x Waiting  x  82 x 12.0.0.SNAPSHOT            x Opencast :: transcription-service-workflowoperation
karaf@root()> bundle:diag 404
Opencast :: transcription-service-workflowoperation (404)
---------------------------------------------------------
Status: Waiting
Declarative Services
org.opencastproject.transcription.workflowoperation.AmberscriptAttachTranscriptionOperationHandler (354)
  missing references: TranscriptionService
org.opencastproject.transcription.workflowoperation.StartTranscriptionOperationHandler (359)
  missing references: TranscriptionService
org.opencastproject.transcription.workflowoperation.AttachTranscriptionOperationHandler (356)
  missing references: TranscriptionService
org.opencastproject.transcription.workflowoperation.AmberscriptStartTranscriptionOperationHandler (355)
  missing references: TranscriptionService

```

geichelberger:
Does opencast work? Because the first time, that wiring error caused karaf to fail on startup (If I remember correctly.). The Waiting bundle state is okay because these services have a reference target filter. The Google OSGi Services seem to be active.

geichelberger:
A dirty solution would be to make the TranscriptionService references optional.

lkiesow:
For now, I just simplified that by moving the transcription services into one module. We can always split them up more later.

Unfortunately, I found another problem though. Activating and then deactivating the Theodul modules will deactivate all of Opencast's REST endpoints. It is likely a bug of the Theodul modules and not of the plugin manager, but this could still be pretty annoying for users.

My guess would be the cusom OSGi service registration handling in the Theodul plugin manager, but it's too late today to investigate any further.,

geichelberger:
I think it has nothing to do with theodul. Its caused by removing the bundle which last created an endpoint. You can reproduce the behavior with just the transcription service. 

It could be caused by the way we createEndpoints and destroy them. If we create an Endpoint we destroy the jax-rs server and recreate it with the new server factory.
```
    if (server != null) {
      logger.debug("Destroying JAX-RS server");
      server.stop();
      server.destroy();
    }

    server = sf.create();
```

After destroying the last servlet, which caused the recreation of the jax-rs server, all other servlets are offline.

https://github.com/opencast/opencast/blob/e57ff7216be8d52e8546ec8cd1b0c09eae835757/modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java#L269-L368

geichelberger:
This is a proof of my theory: https://github.com/geichelberger/opencast/commit/595e5021c60fa57702f4a6043ccbdde202f0944d
The proof recreates the JAX-RS server on destroyEndpoint if there are still servlets registered.

With recreating the server, the rest endpoints are still available after deactivating a plugin.

docs/guides/admin/docs/modules/plugin-management.md
```diff
@@ -0,0 +1,32 @@
...
+    - Legacy annotation functionality. We do not recommend turning this on. It also requires additional configuration in
+      the player. It's kept for a few legacy use-cases.
+- opencast-plugin-player-theodul
+    - Alternative player for Opencast.
+- opencast-plugin-transcription-amberscript
+    - Support for the Amberscript transcription service
+- opencast-plugin-transcription-google-speech
+    - Support for the Google cloud transcription service
+- opencast-plugin-transcription-ibm-watson
+    - Support for the IBM cloud transcription service
```
geichelberger: After the latest change, this should only mention the opencast-plugin-transcription-service.

docs/guides/admin/docs/modules/plugin-management.md
```diff
@@ -0,0 +1,32 @@
+Plugin Management
+=================
+
+Opencast comes with a number of plugins which allow you to turn additional functionality off and on.
+Plugins can be enabled at runtime in `etc/org.opencastproject.plugin.impl.PluginManagerImpl.cfg` by setting the specific
+configuration to `on` or `off`.
+
```
geichelberger, line 7: Maybe mention the current limitation that deactivating features at runtime can cause troubles.

docs/guides/admin/docs/modules/plugin-management.md
```diff
@@ -0,0 +1,32 @@
+Plugin Management
+=================
+
+Opencast comes with a number of plugins which allow you to turn additional functionality off and on.
+Plugins can be enabled at runtime in `etc/org.opencastproject.plugin.impl.PluginManagerImpl.cfg` by setting the specific
+configuration to `on` or `off`.
```
geichelberger, line 6: I would remove the off, since disabling features does currently causes troubles, and you cannot enable a feature by setting the configuration to off.

assemblies/pom.xml
```diff
@@ -121,10 +121,16 @@
...
               <feature>opencast-security-jwt</feature>
-              <feature>opencast-moodle</feature>
-              <feature>opencast-sakai</feature>
-              <feature>opencast-canvas</feature>
-              <feature>opencast-brightspace</feature>
+              <feature>opencast-plugin-legacy-annotation</feature>
+              <feature>opencast-plugin-player-theodul</feature>
+              <feature>opencast-plugin-transcription-amberscript</feature>
+              <feature>opencast-plugin-transcription-google-speech</feature>
+              <feature>opencast-plugin-transcription-ibm-watson</feature>
```
geichelberger, line 128: opencast-plugin-transcription-services

geichelberger (changes requested):
LGTM, only some minor changes are required. I would add it to OC 12 even with the problems if you want to deactivate a feature at runtime. The plugin config would not be the only config that can blow up opencast at runtime.

gregorydlogan:
I'm going to mark this as on hold, pending you saying it's ready for review again

docs/guides/admin/docs/modules/plugin-management.md
lkiesow: done in the config

docs/guides/admin/docs/modules/plugin-management.md
lkiesow: Works for most plugins. The only one it does seem to cause trouble is Theodul. Specifically marked this in the config file.

lkiesow:
This is ready for a (hopefully) final review.

modules/plugin-manager/src/main/java/org/opencastproject/plugin/impl/PluginManagerImpl.java
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * A simple tutorial class to learn about Opencast Services
+ */
+@Component(
+    property = {
+        "service.description=Plugin Manager Service"
+    },
+    immediate = true,
+    service = PluginManagerImpl.class
+)
```
geichelberger, line 50: What do you think of using a short and on-point configuration name by setting the configurationPid to
org.opencastproject.plugins

The benefit would be if we change the plugin manager implementation, we could still use the same configuration file.

geichelberger (changes requested):
LGTM, currently, I only tested the development distribution. I will try the others asap.

geichelberger:
I tested the other distributions it is working as expected.

modules/plugin-manager/src/main/java/org/opencastproject/plugin/impl/PluginManagerImpl.java
lkiesow: I don't really care either way. This is consistent with what we use basically everywhere else in Opencast. I also don't expect that we will dramatically change the implementation, but keep the exact same configuration file. Chances are that we need to adjust it, if we change the implementation.

That said, I don't mind changing this, if you want me to. It's a quick and easy change. I'll do whatever it takes to get this merged :grin:  Do you want me to change it?

gregorydlogan:
@geichelberger says this works, IMO the config file should match the rest of them.  If we're going to start changing naming styles like that we should consider doing *all* of them at the same time, in another PR.  I'm going to merge this so it can finally be in the codebase.

PR #3221, opened 2021-12-03
Link Configuration Files in GitHub

lkiesow:
This patch automatically links configuration files referenced in the
documentation to the actual configuration files in GitHub. This is done
in addition to the tooltip added to `etc/` explaining where to find
these files.

Screenshot from 2021-12-03 14-31-32

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/javascript/extra.js
+++ b/docs/guides/admin/docs/javascript/extra.js
@@ -50,12 +50,30 @@ function addTitleToCodeTag() {
     SPAN.classList.add("etc-span");
     SPAN.title = '"etc" represents the configuration directory of Opencast.\nThis directory is often located at "/etc/opencast".';
 
+    let branch = 'develop';
+    if (window.location.host == 'docs.opencast.org') {
+        branch = window.location.pathname.substring(1).replace(/\/admin.*/, '');
+    }
+    const repobase = `https://github.com/opencast/opencast/blob/${branch}/`;
+
     var codeElementList = document.getElementsByTagName("CODE");
     for (var i = 0; i < codeElementList.length; i++) {
         var CODE = codeElementList[i];
+
         if (typeof CODE.innerText !== 'undefined') {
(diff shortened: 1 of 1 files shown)


docs/guides/admin/docs/javascript/extra.js
```diff
@@ -50,12 +50,30 @@ function addTitleToCodeTag() {
...
     for (var i = 0; i < codeElementList.length; i++) {
         var CODE = codeElementList[i];
+
         if (typeof CODE.innerText !== 'undefined') {
             if (CODE.innerText.startsWith('etc/')) {
                 CODE.innerHTML = CODE.innerHTML.replace(/^etc/, SPAN.outerHTML);
+
+                // Link repository
+                let a = document.createElement('a');
+                a.innerText = '⇲';
```
ziegenberg: Instead of the rotated icon ⇲ (aka South East Arrow to Corner `&#8690;`), one could use 📝 (`&#128221;`) or 📄 (`&#128196;`).

gregorydlogan: I feel like those aren't actually any better though.  At least to my eyes, the 128221 reads as 'edit this' and the other is just a gray blob.

mliradelc: In that case, a gearbox should be better ⚙️, but I like the arrow shown in the example.  The magnify glass also should work 🔍.

lkiesow:
Before we continue down the rabbit hole of what arbitrary character could we misuse for this, if you feel strongly about not using the arrow, we could just use the actual GitHub logo instead:

Screenshot from 2021-12-15 01-56-51

What do you think?

gregorydlogan:
This seems like a good solution to me. Makes it clear it's offsite, and on gh.

lkiesow:
Since I liked the result, I already included this  in this pull request yesterday. Happy to get a review for this.

PR #3148, opened 2021-11-09
Allow empty track duration

mtneug:
The track duration might be empty and in with case the external API request will fail. An empty track duration happens e.g. for subtitle tracks.

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3231, opened 2021-12-07
Fix possible type change in external API

mtneug:
PR #3148 introduced a possible type change in the external API from int
to string for the duration field of tracks in case the field is `null`.
As discussed, this PR changes the return value to JSON `null` instead.

Note that the behavior is still inconsistend when it comes to durations:

* `{eventId}/media` will omit the field
* Event duration has a type change to `""`
* this changes Track duration to `null`

IMO I would leave the behavior of `{eventId}/media` as this is done with
all the fields. I'm unsure about Event duration though as this could be
considered an API change (the last one is none as a default value was
only introduced with 11.x). What do you think?

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


mtneug:
As discussed in the technical meeting: the Event duration now also returns JSON `null` by default.

mliradelc (approved):
Like we talked last week, this change makes much more sense, because a blank space can not be seen for the developer/user.

PR #3250, opened 2021-12-10
Improve deleting events from index

KatrinIhler:
This contains a bugfix and some improvements when deleting events from the index. The whole thing is kinda messy, so I'll try to summarize what I did:
- Move partial deletion of scheduling information, workflow information and asset version of indexed events into their respective services since the index itself shouldn't be aware of the details
- Make these partial delete methods use the regular update functionality of the index - this fixes a bug where the event wasn't locked when making these changes, which could lead to unintended side effects like events not getting deleted properly
- Completely remove the heuristic that if an indexed event has neither a asset version, workflows and scheduling information, we want to delete the event. This is bad and whoever wrote that should feel bad.
- Instead, explicitly delete the whole event at the end of the removeEvent() method of the index service - since, you know, we know that's what we wanna do here.
- This makes it necessary for the index service to know the Elasticsearch index, but since we only got one now, that's fine. Maybe we should change the rest of the code accordingly (not handing the index over from the endpoints), but that's not really important.
- Delete the event if the three services AssetManager, SchedulerService and WorkflowService either succesfully deleted their stuff, or couldn't find it. I believe that catches a few edge cases that could lead to trouble before, like a service not finding the relevant data in their database and then not removing their stuff from the index. Doesn't matter now.
- This also makes the doOnNotFound handler in the admin ui and external api endpoints unnecessary. Yay, less code.
- The partial deletion stuff is still relevant in other cases (e.g. deleting a workflow from an event), so I couldn't get rid of it entirely.
- For the SchedulerService and WorkflowService, I tried to ensure that the partial deletion from the index also happens in case of Not Found exceptions. The AssetManager is unfortunately pretty decoupled from that information, so I couldn't manage it there.
- Fixed the scheduler service logic regarding Not Found since hat seemed to be broken by a prior change.
- Made the removal of a workflow instance from the Elasticsearch index independent of whether we can find the instance in the solr index so we can also do that on Not Found.
- I wrote wrapper methods of the index delete() method for the respective index objects since I felt we were exposing too many details there.
- Additionally I did some minor code style changes and fixed stuff Intellij warned me about for code I was already working on.

Edit: I changed the values of the removedScheduling, removedWorkflow, ... variables in the index service to only be true if there was actually something removed, not in case of Not Found, and introduced a bug in the return value with that, so now that's considered there.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -393,16 +393,9 @@ public abstract class AbstractEventEndpoint {
     if (event.isNone()) {
       return RestUtil.R.notFound(id);
     }
-    final Runnable doOnNotFound = () -> {
-      try {
-        getIndex().delete(Event.DOCUMENT_TYPE,id,getSecurityService().getOrganization().getId());
-      } catch (SearchIndexException e) {
-        logger.error("error removing event {}: {}", id, e);
-      }
-    };
     final IndexService.EventRemovalResult result;
     try {
-      result = getIndexService().removeEvent(event.get(), doOnNotFound, getAdminUIConfiguration().getRetractWorkflowId());
+      result = getIndexService().removeEvent(event.get(), getAdminUIConfiguration().getRetractWorkflowId());
     } catch (WorkflowDatabaseException e) {
       logger.error("Workflow database is not reachable. This may be a temporary problem.");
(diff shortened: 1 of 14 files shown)


Arnei (approved):
The changes seem reasonable to me and the code looks good (Yay less code!). Works as expected in my local test installation, albeit I only tested with test data, not "real world" data.

mliradelc:
If is a bug fix, Should this backported to 11.x or 10.x? or is only a bug for develop?

KatrinIhler:
Since this touches quite a lot, I wasn't comfortable with porting it back. Too afraid to break something in a minor release :D

PR #3248, opened 2021-12-09
Switch to Java Optional

lkiesow:
This patch switches from Opencast's old, internal Option class to Java's
Optional class to simplify the Inbox code.

This should not change any functionality.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3251, opened 2021-12-10
Inbox may fail at startup

### LINKED ISSUE #3243: Inbox may fail at startup
The inbox service may start up prior to the organization configuration being loaded. Since the inbox sets an internal security context to run as a specific user in a specific organization, this can lead to the inbox not finding that user or organization and thus failing to configure an inbox. Instead, a warning is printed to the logs:

```
2021-12-09 01:50:36,916 | WARN  | (InboxScannerService:218) - Cannot create security context for user admin, organization mh_default_org. Either the organization or the user does not exist                                                  
…
2021-12-09 01:50:37,222 | INFO  | (OrganizationDirectoryServiceImpl:239) - Creating organization 'mh_default_org'
```

This is a timing issue and will only pop up from time to time.

A workaround is to restart Opencast or to adjust the inbox configuration in any way at runtime once organization and user are loaded.

lkiesow:
The inbox service may start up prior to the organization configuration
being loaded. Since the inbox sets an internal security context to run
as a specific user in a specific organization, this can lead to the
inbox not finding that user or organization and thus failing to
configure an inbox.

This patch makes the inbox wait for the organization and user to show up
instead of failing. This means that startup order does not matter any
longer.

This fixes #3243
<strike>This is based on #3248</strike>

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
@@ -179,22 +179,28 @@ public class InboxScannerService implements ArtifactInstaller, ManagedService {
 
     var securityContext = getUserAndOrganization(securityService, orgDir, orgId, userDir, userId)
             .map(a -> new SecurityContext(securityService, a.getB(), a.getA()));
-    // Only setup new inbox if security context could be acquired
-    if (securityContext.isPresent()) {
-      // remove old file install configuration
-      fileInstallCfg.foreach(removeFileInstallCfg);
-      // set up new file install config
-      fileInstallCfg = some(configureFileInstall(cc.getBundleContext(), inbox, interval));
-      // create new scanner
-      Ingestor ingestor = new Ingestor(ingestService, securityContext.get(), workflowDefinition,
-              workflowConfig, mediaFlavor, inbox, maxThreads, seriesService, maxTries, secondsBetweenTries);
-      this.ingestor = some(ingestor);
-      new Thread(ingestor).start();
-      logger.info("Now watching inbox {}", inbox.getAbsolutePath());
-    } else {
-      logger.warn("Cannot create security context for user {}, organization {}. "
-              + "Either the organization or the user does not exist", userId, orgId);
+    while (securityContext.isEmpty()) {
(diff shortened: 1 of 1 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
```diff
@@ -179,29 +176,31 @@ public synchronized void updated(Dictionary properties) throws ConfigurationExce
...
+    while (securityContext.isEmpty()) {
+      logger.debug("Could not create security context for user {}, organization {}. "
+              + "Either the organization or the user does not exist (yet).", userId, orgId);
+        try {
+          Thread.sleep(5000);
+        } catch (InterruptedException e) {
+          return;
+        }
+      securityContext = getUserAndOrganization(securityService, orgDir, orgId, userDir, userId)
+              .map(a -> new SecurityContext(securityService, a.getB(), a.getA()));
```
mtneug, line 191: Instead of busy waiting (potentially forever if the org does not exists / is removed), you can register a listener to the `OrganizationDirectoryService#addOrganizationDirectoryListener(OrganizationDirectoryListener)`. Those will be called once an organization is added, removed, updated.

Not sure if it is best to register the `InboxScannerService` instance or an anonymous in place implementation.

mtneug (changes requested):

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
lkiesow: Using `addOrganizationDirectoryListener()` is potentially possible but makes this horribly complex since you would have to deal with lots of concurrency problems that can happen. After all, the organization can already exist, the organization may only exist in the future, or it may happen exactly parallel to registering any listeners.

In fact, there is actually a good chance of this happening in parallel since organization creation and this code running roughly happens at the same time, which is the reason for this issue only popping up from time to time.

That is why you would need to register the listener, then check if the organization was already there, unregister it again if it existed already and would need a synchronized method to start the Ingestor to prevent starting it twice by accident. Given that this should usually only wait for a very short time, if at all, I don't think that complexity is worth it, do you?

PR #3253, opened 2021-12-10
CVE-2021-44228 (Apache Log4j Remote Code Execution) mitigation

gregorydlogan:
This can be/should be removed once we've updated to pax-logging-log4j2:1.11.10 or newer.

This fixes [GHSA-mf4f-j588-5xm8](https://github.com/opencast/opencast/security/advisories/GHSA-mf4f-j588-5xm8).

### DIFF
--- a/assemblies/resources/bin/setenv
+++ b/assemblies/resources/bin/setenv
@@ -34,7 +34,7 @@
 # export KARAF_DEBUG       # Enable debug mode
 # export KARAF_REDIRECT    # Enable/set the std/err redirection when using bin/start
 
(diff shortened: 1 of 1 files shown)


gregorydlogan:
@JulianKniephoff this should go into the last 9.x release, and then we can merge it forward from there.

gregorydlogan:
Aaand I just found out this only works for log4j 2.10 to 2.15. Not entirely sure what we have, but I don't think that applies.

pascalseeland:
ops4j-logging used by oc9 includes 2.13.2 (https://github.com/ops4j/org.ops4j.pax.logging/blob/de10f3b46facf72e41ac7b7750eb4b703732ec0e/pom.xml#L231)
so this fix does work

lkiesow:
Merging this without discussion in the technical meeting since it is urgent and should make the release tomorrow.

PR #3273, opened 2021-12-17
Changed embed code selection design

### LINKED ISSUE #3214: Embed code selection design
We have had several complaints about the design (mainly the choice of colors) for the embed code selection where in particular the yellow highlighting clashes with the overall blueish page:

a

We would like to implement a somewhat more neutral style for this dialog.
Here is a quick first mockup of how this could look like:

d

narickmann:
This fixes #3214

Looks now like this:
code-design

### DIFF
--- a/modules/admin-ui-frontend/app/styles/views/modals/_embedded-code.scss
+++ b/modules/admin-ui-frontend/app/styles/views/modals/_embedded-code.scss
@@ -21,19 +21,20 @@
 
 .embedSizeButton {
     display: inline-block;
-    background-color: #466DC3;
-    border: 1px solid silver;
+    background-color: #E2E1E1;
+    border: 1px solid #CCC;
     border-radius: 10px;
     cursor: pointer;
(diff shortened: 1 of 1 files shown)


modules/admin-ui-frontend/app/styles/views/modals/_embedded-code.scss
```diff
@@ -21,19 +21,20 @@
...
     vertical-align: middle;
     font-size: 14px;
     text-align: center;
-    color: white;
+    color: black;
 }
 
 .embedSizeButtonSelected {
-    border: 5px solid #FFE04F;
+    //border: 5px solid #FFE04F;
```
lkiesow: This line should probably not have been part of the patch:
```suggestion
```

lkiesow (changes requested):

lkiesow (approved):
- Looks definitely better than before
- ICLA is filed with Apereo

PR #3274, opened 2021-12-17
Index Extended Metadata

KatrinIhler:
With this PR, extended metadata of events and series are put into the Elasticsearch index. This can be used to filter by extended metadata fields via the full text search. The series extended metadata is indexed by the Series Service and the event extended metadata by the Asset Manager.

Some more detailed information on what this PR does/doesn't and why/why not:

1. The first commit renames "Api index" to "Elasticsearch index". This was most likely an oversight during a rebase of one of my previous PRs and I took this chance to fix this.
2. The metadata of a single catalog will be rewritten completely on each update, so fields that are no longer configured will be removed. The same is however, not entirely true for catalogs that are no longer configured - the series service lacks awareness of this, so old series catalogs may linger, though there is an endpoint to remove them in the External API.
3. I added an OSGI property to the catalog UI adapters on registration so the Asset Manager can get only those for extended metadata. This is necessary because the Asset Manager can't tell whether a catalog contains extended event metadata without them.
4. I changed the implementation of DublinCoreByteFormat so we don't swallow exceptions there.
5. I added an XML adapter to transform the map with the extended metadata, since the XML is what's used to read a series from the index, and without this change, incremental updates will lose catalogs.
5. I added a new endpoint/method to the series service so we can hand over the whole dublin core catalog instead of transforming to a byte stream and back. Since we only ever use the series element endpoints for extended metadata internally, and the endpoints aren't explicitly stable, I intend to change or remove them accordingly. (Actually, I have already begun the work, but I lack time to finish it before my vacation, so expect a separate PR in January. This might be better anyway since this change might be controversial.)

(I know this change isn't exactly pretty, but please be merciful, I really tried. :laughing:)

This work was sponsored by the University of Vienna.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -473,12 +473,12 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
   }
 
   /**
-   * Update the event in the API index.
+   * Update the event in the Elasticsearch index.
    *
    * @param snapshot
    *         The newest snapshot of the event to update
    * @param index
-   *         The API index to update
+   *         The Elasticsearch index to update
    */
   private void updateEventInIndex(Snapshot snapshot) {
     final MediaPackage mp = snapshot.getMediaPackage();
@@ -531,12 +531,12 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
   }
 
   /**
-   * Remove the event from the API index
+   * Remove the event from the Elasticsearch index
    *
    * @param eventId
    *         The id of the event to remove
    * @param index
-   *         The API index to update
+   *         The Elasticsearch index to update
    */
   private void removeArchivedVersionFromIndex(String eventId) {
     final String orgId = securityService.getOrganization().getId();
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -133,6 +133,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -530,7 +531,8 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
 
       // extended metadata
       event.resetExtendedMetadata();  // getting rid of old data
-      for (EventCatalogUIAdapter extendedCatalogUIAdapter : extendedEventCatalogUIAdapters.get(organization)) {
+      for (EventCatalogUIAdapter extendedCatalogUIAdapter : extendedEventCatalogUIAdapters.getOrDefault(organization,
+              Collections.emptyList())) {
         for (Catalog catalog: mp.getCatalogs(extendedCatalogUIAdapter.getFlavor())) {
           try (InputStream in = workspace.read(catalog.getURI())) {
             EventIndexUtils.updateEventExtendedMetadata(event, DublinCores.read(in),
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -81,6 +81,7 @@ import org.opencastproject.mediapackage.MediaPackageSupport;
 import org.opencastproject.message.broker.api.assetmanager.AssetManagerItem;
 import org.opencastproject.message.broker.api.update.AssetManagerUpdateHandler;
 import org.opencastproject.metadata.dublincore.DublinCores;
+import org.opencastproject.metadata.dublincore.EventCatalogUIAdapter;
 import org.opencastproject.security.api.AccessControlEntry;
 import org.opencastproject.security.api.AccessControlList;
 import org.opencastproject.security.api.AccessControlParser;
@@ -184,6 +185,7 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
   private Database db;
   private AclServiceFactory aclServiceFactory;
   private ElasticsearchIndex index;
+  private Map<String, List<EventCatalogUIAdapter>> extendedEventCatalogUIAdapters = new HashMap<>();
 
   // Settings for role filter
   private boolean includeAPIRoles;
@@ -284,6 +286,20 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
   @Reference
   public void setIndex(ElasticsearchIndex index) {
     this.index = index;
+  }
+
+  @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC,
+          target = "(common-metadata=false)")
+  public synchronized void addCatalogUIAdapter(EventCatalogUIAdapter catalogUIAdapter) {
+    List<EventCatalogUIAdapter> list = extendedEventCatalogUIAdapters.computeIfAbsent(
+            catalogUIAdapter.getOrganization(), k -> new ArrayList());
+    list.add(catalogUIAdapter);
+  }
+
+  public synchronized void removeCatalogUIAdapter(EventCatalogUIAdapter catalogUIAdapter) {
+    if (extendedEventCatalogUIAdapters.containsKey(catalogUIAdapter.getOrganization())) {
+      extendedEventCatalogUIAdapters.get(catalogUIAdapter.getOrganization()).remove(catalogUIAdapter);
+    }
   }
 
   /**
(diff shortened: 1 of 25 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -493,15 +509,30 @@ private void updateEventInIndex(Snapshot snapshot, ElasticsearchIndex index) {
...
         } catch (IOException | NotFoundException e) {
-          throw new IllegalStateException(String.format("Unable to load dublin core catalog for event '%s'",
+          throw new IllegalStateException(String.format("Unable to load common dublin core catalog for event '%s'",
                   mp.getIdentifier()), e);
         }
       }
 
+      // extended metadata
+      event.resetExtendedMetadata();  // getting rid of old data
+      for (EventCatalogUIAdapter extendedCatalogUIAdapter : extendedEventCatalogUIAdapters.get(organization)) {
```
Arnei: There is a chance for an uncaught NullPointerException here (which I realized due to an index rebuild in my local not-extended-metadata-having installation)

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -1737,6 +1737,55 @@ public Response getSeriesElement(
...
+  @RestQuery(
+          name = "updateExtendedMetadata",
+          description = "Updates extended metadata of a series",
+          returnDescription = "An empty response",
+          pathParameters = {
+                  @RestParameter(name = "seriesId", description = "The series identifier", type = STRING,
+                          isRequired = true),
+                  @RestParameter(name = "type", description = "The catalog type", type = STRING,
+                          isRequired = true),
+                  @RestParameter(name = "dc", description = "The catalog with extended metadata.", type = TEXT,
```
Arnei, line 1252: Since it's not a path `dc` should probably be listed under `restParameters` instead of `pathParameters`?

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -1737,6 +1737,55 @@ public Response getSeriesElement(
     }
   }
 
+  @PUT
+  @Path("{seriesId}/extended-metadata/{type}")
```
Arnei: The hyphen seems to break the rest docs. Or I am going crazy. Could be both, really.

Arnei: Seems a bit weird to have a way to set extended metadata without a way to get it. Might be reasonable to add a @GET endpoint?

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
```diff
@@ -1737,6 +1737,55 @@ public Response getSeriesElement(
...
+  @RestQuery(
+          name = "updateExtendedMetadata",
+          description = "Updates extended metadata of a series",
+          returnDescription = "An empty response",
+          pathParameters = {
+                  @RestParameter(name = "seriesId", description = "The series identifier", type = STRING,
+                          isRequired = true),
+                  @RestParameter(name = "type", description = "The catalog type", type = STRING,
```
Arnei: Might be good to give an example for `type`. Also, is type a unique identifier?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/objects/series/SeriesIndexUtils.java
```diff
@@ -122,17 +122,40 @@ public static SearchMetadataCollection toSearchMetadata(Series series) {
...
+   * @param extendedMetadata
+   *          the extended metadata map
+   */
+  private static void addExtendedMetadata(SearchMetadataCollection doc, Map<String, Map<String,
+          List<String>>> extendedMetadata) {
+    for (String type: extendedMetadata.keySet()) {
+      Map<String, List<String>> extendedMetadataByType = extendedMetadata.get(type);
+      for (String name: extendedMetadataByType.keySet()) {
+        List<String> values = extendedMetadataByType.get(name);
+        String fieldName = SeriesIndexSchema.EXTENDED_METADATA_PREFIX.concat(type + "_" + name);
```
Arnei, line 153: This naming schema causes trouble if there are two extended metadata catalogs with the same type AND a property with the same inputID. Unlike with events, the `type` only seems to be the maintype of the flavor, not maintype + subtype. Basically this causes the two catalogs to have a kind of "shared" field in the index. No idea how realistic that scenario is though.

Arnei (changes requested):
Couple of minor things, looks good and works in my local test installation otherwise.

Given the recent discussions about high availability and index rebuilds, would you expect extended metadata to be significant to the speed of index rebuilds under "normal" circumstances? Do you think it would be a good idea to make adding extended metadata to the index configurable?

modules/index-service/src/main/java/org/opencastproject/index/service/catalog/adapter/CatalogUIAdapterFactory.java
```diff
@@ -120,6 +121,8 @@ public void updated(String pid, Dictionary<String, ?> properties) throws Configu
     final String type = getCfg(properties, CONF_TYPE_KEY);
     Option<String> optCommonMetadata = getOptCfg(properties, CONF_COMMON_METADATA_KEY);
     final boolean isCommonMetadata = optCommonMetadata.isSome() ? Boolean.parseBoolean(optCommonMetadata.get()) : false;
+    Dictionary serviceProperties = new Properties();
+    serviceProperties.put("common-metadata", isCommonMetadata);
```
geichelberger: The class has a `common-metadata` constant CONF_COMMON_METADATA_KEY.

KatrinIhler: Good find, thx!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
KatrinIhler: To be certain: You meant the `extendedEventCatalogUIAdapters.get()` can return null, right?

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
KatrinIhler: Yep!

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
KatrinIhler: You can actually already get extended metadata via GET /{seriesId}/elements/{elementType}. And the new endpoint is actually just a wrapper for PUT /{seriesId}/elements/{elementType} as well. I intended to get rid of the old endpoints, but then that fell of my radar, not sure if/when that's going to happen now. But for now, this doesn't break the old endpoints, even if they're not explicitly stable.

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
KatrinIhler: This refers to the type (so the part before the /) of the catalog flavor, and is used in a bunch of series rest endpoints already. The series service expects this part to be unique. I tried to be a little clearer in the description.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/objects/series/SeriesIndexUtils.java
KatrinIhler: There should never be two extended metadata catalogs with the same type for series, because this will already break in the database. The series service expects this to be unique. This is actually documented somewhere if I remember correctly. And I totally agree that this is bonkers, but I'm not touching that particular hornets nest with a stick right now.

KatrinIhler:
@Arnei Thanks for your review, sorry that I took so long to actually address it. :sweat_smile: 

> Given the recent discussions about high availability and index rebuilds, would you expect extended metadata to be significant to the speed of index rebuilds under "normal" circumstances? Do you think it would be a good idea to make adding extended metadata to the index configurable?

Since we need to read the extended metadata from the file system for events, this could impact the performance of the index rebuild. However, this would only affect those organizations that use extended metadata, of which there are few.

I guess you could make this configurable by having the new field in the index scheme, but just not indexing the data if this is turned off. Not sure if this is a good idea. @gregorydlogan Maybe we want to talk about this PR in the next technical meeting again?

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
KatrinIhler: Nope, you're right. Fixed!

gregorydlogan:
Seems reasonable to me, will add to the notes for next week.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/objects/series/SeriesIndexUtils.java
Arnei: I'll take your word for it :)

modules/series-service-impl/src/main/java/org/opencastproject/series/endpoint/SeriesRestService.java
Arnei: Thanks for that. Endpoints with type fields that don't specify what `type` refers to is a little pet peeve of mine (there's a couple more of those in Opencast).

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
Arnei: I absolutely do not remember anymore, but your assertion seems likely.

KatrinIhler: Well, I fixed that one. :D

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -286,6 +289,20 @@ public void setIndex(ElasticsearchIndex index) {
     this.index = index;
   }
 
+  @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC,
```
gregorydlogan, line 292: Doesn't this need an `unbind` to tell it to call `removeCatalogUIAdapter`?  Or is OSGi smart enough to figure that out on its own?

gregorydlogan (commented):
Looks good, aside from this one question

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
geichelberger: OSGi should be smart enough, I think it uses keywords like bind/unbind add/remov.

KatrinIhler: Yeah, I think OSGi infers this.

gregorydlogan:
Per discussion in the technical meeting, there are concerns that this may have perf issues with respect to reindexing, but rather than watching this PR to ensure there are no conflicts let's focus on getting anything that pops up fixed.  I'm going to merge this.

PR #3293, opened 2021-12-20
Fix pax-logging in startup.properties

lkiesow:
This patch fixes some pax-logging replacements done when assembling
Opencast's distributions.

The reason why Opencast works as expected even without this patch is,
that we had to make sure to include the new bundles in the distribution
and the easiest way of doing that was for Opencast's Karaf features to
require the bundles.

That is why `pax-logging-log4j2` will automatically be started when
starting any of Opencast's Karaf features, automatically mitigating the
replacement error.

The only side-effect was that the logger's start level was slightly
higher than before. In practice, that meant that logging would start
about a second later.

This patch now ensures the start level again drops to `8` and no one
gets confused about why we have `pax-logging-api` twice in the
`startup.properties`.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


geichelberger:
LGTM

geichelberger (commented):
LGTM

geichelberger (approved):

lkiesow:
Thanks for the review → merging

PR #3255, opened 2021-12-12
Migrate to declarative services

geichelberger:
Current status: ~~It does not blow up completely.~~ It seems to work.

* [x] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3300, opened 2021-12-22
Fixup osgi services and xmls

geichelberger:
This PR is a preparation for PR #3255.

Some OSGI XML files have shown inconsistencies. Methods were specified that did not exist and dynamic references were missing unbind methods. When switching to Declarative Services this would lead to compilation errors.

Side effect it resolves the shutdown errors of Opencast. Missing deactivate Method etc.

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
+++ b/modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
@@ -52,5 +52,5 @@
              interface="org.opencastproject.security.urlsigning.service.UrlSigningService"
              bind="setUrlSigningService"/>
   <reference name="asset-manager" interface="org.opencastproject.assetmanager.api.AssetManager"
-             cardinality="1..1" policy="dynamic" bind="setAssetManager"/>
+             cardinality="1..1" policy="dynamic" bind="setAssetManager" unbind="unsetAssetManager"/>
 </scr:component>
--- a/modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
+++ b/modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
@@ -52,5 +52,5 @@
              interface="org.opencastproject.security.urlsigning.service.UrlSigningService"
              bind="setUrlSigningService"/>
   <reference name="asset-manager" interface="org.opencastproject.assetmanager.api.AssetManager"
-             cardinality="1..1" policy="dynamic" bind="setAssetManager" unbind="unsetAssetManager"/>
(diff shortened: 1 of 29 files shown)


modules/admin-ui/src/main/resources/OSGI-INF/event_endpoint.xml
```diff
@@ -52,5 +52,5 @@
              interface="org.opencastproject.security.urlsigning.service.UrlSigningService"
              bind="setUrlSigningService"/>
   <reference name="asset-manager" interface="org.opencastproject.assetmanager.api.AssetManager"
-             cardinality="1..1" policy="dynamic" bind="setAssetManager"/>
+             cardinality="1..1" policy="dynamic" bind="setAssetManager" unbind="unsetAssetManager"/>
```
lkiesow: Not against your patch… but why on earth do we have a dynamic binding for this? I'm pretty sure the service will not work without the asset manager being present. And that's probably a question for all the other dynamic asset manager bindings as well. It seems like someone started and everyone else copied that XML…

gregorydlogan: Honestly, I think that's exactly what happened.  I know I've told many a newbie to 'just copy the OSGi descriptors from some other service'.  In hindsight, a mistake.

IMO, there's no reason for an unbind here.  Or in many other places.

geichelberger: Is there a difference in behavior between policy dynamic or static, if the cardinality is 1..1? If the asset manager unbinds, the event endpoint service will be deactivated.

geichelberger: ` If the cardinality constraints cannot be maintained after a component configuration is activated, that is the reference becomes unsatisfied, the component configuration must be deactivated.`

http://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-reference.cardinality

lkiesow: You are right, I missed the cardinality declaration. Still, there is a slight difference between `static, 1..1` and `dynamic, 1..1`. The latter lets you replace bound services at runtime, while the first would require a restart of the service. Dynamic may thus break if you do something with the specific service during activation. I don't think that's relevant for Opencast in most cases. In general, I would still prefer static, since that is simpler (less configuration, no unbind, …) and gives you some guarantees due to the guaranteed component restart. Of course, unless you implement something like a plugin system where you are actually often switching references and restarting the components every time would become costly.

PR #3340, opened 2022-01-10
Match Inbox Events Against Schedule

### LINKED ISSUE #3240: Match inbox events against schedule
This is based on an implementation for #3239.

Capture agents often record files to file names containing metadata which allows us to identify an event by looking at a specific time in the schedule of a certain location (capture agent).

It would be nice if the inbox could use this to identify an event and extend the metadata of ingested media.
This needs to be an optional setting and should be turned off by default.

lkiesow:
Capture agents often record to filenames containing metadata which
allows us to identify an event by looking at a specific time in the
schedule of a certain location (capture agent).

If `created` and `spatial` metadata could be identified, this patch
allows us to look up the schedule, using all available event data fram
there if we had a unique match.

Data taken from the schedule, overwriting inbox data are:

- All metadata
- Workflow definition
- Workflow configuration

Notes:

- This patch fixes #3240.
- This is based on #3327.
- You want #3339 for this.

__Addition__: Use ffprobe to get additional metadata

In addition to the metadata extraction from the file name, this patch
allows users to have ffprobe run on the media files to get some
additional metadata from the media container.

The extracted metadata are creation time and duration, which may help to
match scheduled events.

With the media file now having a duration, there is now also an
additional option allowing to filter out some events which are barely
conflicting (imaging two back to back recordings and the second one
starts a minute early).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
@@ -27,6 +27,7 @@ import static org.opencastproject.util.data.Collections.dict;
 import static org.opencastproject.util.data.Tuple.tuple;
 
 import org.opencastproject.ingest.api.IngestService;
+import org.opencastproject.scheduler.api.SchedulerService;
 import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.UserDirectoryService;
@@ -34,6 +35,7 @@ import org.opencastproject.security.util.SecurityContext;
 import org.opencastproject.series.api.SeriesService;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.felix.fileinstall.ArtifactInstaller;
@@ -119,12 +121,14 @@ public class InboxScannerService implements ArtifactInstaller, ManagedService {
 
   public static final String INBOX_METADATA_REGEX = "inbox.metadata.regex";
   public static final String INBOX_DATETIME_FORMAT = "inbox.datetime.format";
+  public static final String INBOX_SCHEDULE_MATCH = "inbox.schedule.match";
 
   private IngestService ingestService;
   private SecurityService securityService;
   private UserDirectoryService userDir;
   private OrganizationDirectoryService orgDir;
   private SeriesService seriesService;
+  private SchedulerService schedulerService;
 
   private ComponentContext cc;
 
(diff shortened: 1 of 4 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
```diff
@@ -291,4 +480,40 @@ public void cleanup(final File artifact) {
...
+
+  class JsonTags {
+    @SerializedName(value = "creation_time")
+    private String creationTime;
+
+    Date getCreationTime() throws ParseException {
+      if (creationTime == null) {
+        return  null;
+      }
+      DateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSz");
```
gregorydlogan, line 510: You're stripping Z below, and using z here.  Is that going to break anything?

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
```diff
@@ -172,6 +187,20 @@ public synchronized void updated(Dictionary properties) throws ConfigurationExce
...
 
+    // Metadata parsing configuration
+    var metadataPattern = Optional.ofNullable(properties.get(INBOX_METADATA_REGEX))
+            .map(Objects::toString)
+            .map(Pattern::compile);
+    var dateFormatter = Optional.ofNullable(properties.get(INBOX_DATETIME_FORMAT))
+            .map(Objects::toString)
+            .map(DateTimeFormatter::ofPattern)
+            .orElse(DateTimeFormatter.ISO_DATE_TIME);
+    var ffprobe = BooleanUtils.toBoolean((String) properties.get(INBOX_METADATA_FFPROBE))
```
gregorydlogan, line 198: You're using boolean here, but below treat it as a string.  Why not treat the whole thing as a boolean and default to false?

Edit: to be clear, I'm referring to https://github.com/opencast/opencast/pull/3340/files#diff-91b7b32da6b6f75846087746803d8660b4282f48f39e365d5e00250b412be6c5R210 as 'below'

gregorydlogan (approved):
In general this looks ok, a couple of questions remain.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/InboxScannerService.java
lkiesow: Not sure what you mean. The result is always a `String` and always used as such. Only the first argument of the ternary operator is a boolean. This is just saying, if using ffprobe is turned on, set this to the configured ffprobe binary, otherwise set this to `null`.

I guess you could treat both configuration options differently here, but then you would have a similar check at the end.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
lkiesow: That's… magic :D
The `z` in the format string means: “parse a timezone here”.
The `Z` in the date string will not work. Java's parser want's the `+0000` format. That's what the replace is for.
In short, `z` parses `+0000`.

PR #3343, opened 2022-01-11
Password strength indicator

### LINKED ISSUE #3215: Password strength indicator
Users tend to select bad passwords, and it is a good idea to remind them of that. Therefore, it would be nice to have a basic password strength indicator for the web interfaces which allows users to set passwords:

Screenshot from 2021-12-01 13-11-10 (copy)

As an indicator, a bar would be nice which will be red at < 33%, yellow at < 66% and green above that.

Password strength is usually not an exact science, but a simple approach would be something like this:
- https://stackoverflow.com/a/32440338/2352895

narickmann:
This fixes #3215.
Simple strength bar, checks if the password contains at least one uppercase letter, one lowercase letter, one number, one special character and is at least 8 characters long. (Colors of strength-bar: red, orange, yellow, green, darkgreen)

strength-bar-1

strength-bar-4

strength-bar-6

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
@@ -41,45 +41,92 @@ angular.module('adminNg.directives')
         link: function (scope, elem, attrs, ctrl) {
           scope.$watch('password', function (pw) {
 
-            scope.strength =
-                isValid(pw && /[A-Z]/.test(pw)) +
-                isValid(pw && /[a-z]/.test(pw)) +
-                isValid(pw && /\d/.test(pw)) +
-                isValid(pw && /(?=.*\W)/.test(pw)) +
-                isValid(pw && /^.{8,}$/.test(pw));
+            scope.rule =
+                requirements(pw && /[A-Z]/.test(pw)) +
+                requirements(pw && /[a-z]/.test(pw)) +
+                requirements(pw && /\d/.test(pw)) +
+                requirements(pw && /(?=.*\W)/.test(pw)) +
+                requirements(pw && /^.{8,}$/.test(pw));
 
-            var bar = 'background:white; width:0%;';
-            var pw_strength = '';
-            setProgBar(scope.strength);
+            // bad passwords from https://en.wikipedia.org/wiki/List_of_the_most_common_passwords
+            // plus Opencast's default password
+            const bad_passwords = ['0', '111111', '1111111', '123', '123123', '123321',
+              '1234', '12345', '123456', '1234567', '12345678', '123456789', '1234567890',
+              '12345679', '123qwe', '18atcskd2w', '1q2w3e', '1q2w3e4r', '1q2w3e4r5t',
+              '3rjs1la7qe', '555555', '654321', '666666', '7777777', '888888',
+              '987654321', 'aa12345678', 'abc123', 'admin', 'dragon', 'Dragon', 'google',
+              'iloveyou', 'Iloveyou', 'lovely', 'Monkey', 'mynoob', 'password',
+              'password1', 'password12', 'password123', 'princess', 'qwerty', 'qwerty123', 'qwertyuiop',
+              'Qwertyuiop', 'welcome', 'zxcvbnm', 'opencast' ];
 
-            function isValid(rule) {
+            var bar_color = 'background:white;';
+            var bar_width = 'width:0%;';
+            var bar_text = '';
+            var strength = calcStrength(scope.rule, pw);
+
+            setProgBar(strength);
+
+            function requirements(rule) {
               return rule ? 1 : 0;
             }
 
-            function setProgBar(validRules) {
-              if (validRules == 1) {
-                bar = 'background:red; width:20%;';
-                pw_strength = 'Very Weak';
+            function calcStrength(rule, pw){
+
+              if (bad_passwords.indexOf(pw) > -1) {
+                strength = 0;
+                return strength;
               }
-              else if (validRules == 2) {
-                bar = 'background:darkorange; width:40%;';
-                pw_strength = 'Weak';
-              }
-              else if (validRules == 3) {
-                bar = 'background:gold; width:60%;';
-                pw_strength = 'Good';
-              }
-              else if (validRules == 4) {
-                bar = 'background:green; width:80%;';
-                pw_strength = 'Strong';
-              }
-              else if (validRules == 5) {
-                bar = 'background:#388ed6; width:100%;';
-                pw_strength = 'Very Strong';
-              }
-              document.getElementById('bar').style = bar;
-              document.getElementById('pw').innerHTML = pw_strength;
+              // for every unused rule/requirement = -5 points
+              var usedRules = (rule - 5) * 5;
+
+              var uniqueChars = (pw.split('').sort().join('').replace(/(.)\1+/g, '').length) * 2;
+              var pw_length = pw.length * 4;
+              var lowerCase = (pw.length - pw.replace(/[a-z]/g, '').length) * 2;
+              var upperCase = (pw.length - pw.replace(/[A-Z]/g, '').length) * 2;
+              var number = (pw.length - pw.replace(/[0-9]/g, '').length) * 4;
+              var symbol = (pw.length - pw.replace(/\W/g, '').length) * 6;
+
+              var strength = usedRules + uniqueChars + pw_length + lowerCase + upperCase + number + symbol;
+              return strength;
             }
+
+            function setProgBar(strength) {
+
+              if (strength >= 90) {
+                bar_color = 'background:#388ed6;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Very strong';
+              }
+              else if (strength >= 70) {
+                bar_color = 'background:green;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Strong';
+              }
+              else if (strength >= 50) {
+                bar_color = 'background:gold;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Good';
+              }
+              else if (strength >= 30) {
+                bar_color = 'background:darkorange;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Weak';
+              }
+              else if (strength >= 1) {
+                bar_color = 'background:red;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Very weak';
+              }
+              else if (strength == 0){
+                bar_color = 'background:white;';
+                bar_width = 'width:' + 0 + '%;';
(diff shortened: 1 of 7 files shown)


gregorydlogan:
Quick comments:
- The colour contrast for the yellow line is hard to make out
- The green colour in the last image should probably match the rest of the UI branding.

lkiesow:
An additional textual indicator (weak, good, strong) would also be helpful

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,93 @@
...
+              else if (validRules == 4) {
+                bar = 'background:green; width:80%;';
+                pw_strength = 'Strong';
+              }
+              else if (validRules == 5) {
+                bar = 'background:#388ed6; width:100%;';
+                pw_strength = 'Very Strong';
+              }
+              document.getElementById('bar').style = bar;
+              document.getElementById('pw').innerHTML = pw_strength;
```
lkiesow: Safer and should work the same:

```suggestion
              document.getElementById('pw').innerText = pw_strength;
```

narickmann:
Yellow is now darker (gold) and darkgreen is now blue. Also added a textual indicator. Pls let me know, if the colors are okay or if you want me to change them or the textual indicator.

strength-bar-1
strength-bar-2
strength-bar-3
strength-bar-4
strength-bar-5
strength-bar-6

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,93 @@
...
+
+        link: function (scope, elem, attrs, ctrl) {
+          scope.$watch('password', function (pw) {
+
+            scope.strength =
+                isValid(pw && /[A-Z]/.test(pw)) +
+                isValid(pw && /[a-z]/.test(pw)) +
+                isValid(pw && /\d/.test(pw)) +
+                isValid(pw && /(?=.*\W)/.test(pw)) +
+                isValid(pw && /^.{8,}$/.test(pw));
```
lkiesow: This just awards a single point to all passwords longer than 8 characters (identical to `pw.length >= 8`).
How about awarding a point for each 8 unique characters instead? That way, a very long password with lots of different characters gets extra points, while just repeating a single character 8 times gets no points.
Something like:

```suggestion
                evalLength(pw);
            function evalLength(password) {
                let unique = '';
                for (const c of password) {
                    if (u.indexOf(c) < 0) {
                        unique += c;
                    }
                }
                return Math.floor(unique.length / 8);
            }
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,93 @@
...
+            var bar = 'background:white; width:0%;';
+            var pw_strength = '';
+            setProgBar(scope.strength);
+
+            function isValid(rule) {
+              return rule ? 1 : 0;
+            }
+
+            function setProgBar(validRules) {
+              if (validRules == 1) {
```
lkiesow: Maybe additionally check against the most commonly used bad passwords:

```suggestion
// bad passwords from https://en.wikipedia.org/wiki/List_of_the_most_common_passwords
// plus Opencast's default password
const bad_passwords = ["0", "111111", "1111111", "123", "123123", "123321",
	"1234", "12345", "123456", "1234567", "12345678", "123456789", "1234567890",
	"12345679", "123qwe", "18atcskd2w", "1q2w3e", "1q2w3e4r", "1q2w3e4r5t",
	"3rjs1la7qe", "555555", "654321", "666666", "7777777", "888888",
	"987654321", "aa12345678", "abc123", "admin", "dragon", "Dragon", "google",
	"iloveyou", "Iloveyou", "lovely", "Monkey", "mynoob", "password",
	"password1", "password123", "princess", "qwerty", "qwerty123", "qwertyuiop",
	"Qwertyuiop", "welcome", "zxcvbnm", "opencast" ]

              if (bad_passwords.indexOf(pw) >= 0) {
                  // This is one of the most commonly used (bad) passwords
                  bar = 'background:black; width:10%;';
                  pw_strength = 'Bad password';
              }
              else if (validRules == 1) {
```

lkiesow (changes requested):
I don't think I would necessarily agree with the algorithm:

```
yzvhcklsvojttgtpbljurxmgcrdkdvvopncywiunsdszxsafltczxbrohmafdgnvmzghhmjcalynbmxlemitrrnyqnkaoliesadchhrpckwslaxcksvlgjabaetleqxo
-> Weak
```

while

```
aA1
-> Strong
```

Now, we don't need to (and never will be) perfect, but maybe just factor in the length.
I'll add a few suggestions for that in the code below.

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,93 @@
...
+              }
+              else if (validRules == 3) {
+                bar = 'background:gold; width:60%;';
+                pw_strength = 'Good';
+              }
+              else if (validRules == 4) {
+                bar = 'background:green; width:80%;';
+                pw_strength = 'Strong';
+              }
+              else if (validRules == 5) {
```
lkiesow: Using my suggestion, this check also would need a slight adjustment:

```suggestion
              else if (validRules >= 5) {
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,140 @@
...
+            function calcStrength(rule, pw){
+
+              if (bad_passwords.indexOf(pw) > -1) {
+                strength = 0;
+                return strength;
+              }
+              // for every unused rule/requirement = -5 points
+              var usedRules = (rule - 5) * 5;
+
+              var uniqueChars = (pw.split('').sort().join('').replace(/(.)\1+/g, '').length) * 2;
```
lkiesow: This will actually count unique characters only. Meaning that e.g. `aaa` would count as `0` and not as `1`. This is probably not what you intended. What you want is probably:

```suggestion
              var uniqueChars = new Set(pw).size * 2;
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,140 @@
...
+                bar_color = 'background:red;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Very weak';
+              }
+              else if (strength == 0){
+                bar_color = 'background:white;';
+                bar_width = 'width:' + 0 + '%;';
+                bar_text = 'Bad password';
+              }
+
```
lkiesow, line 120: Instead of repeating this in every condition above, you could set this once here:

```suggestion
              bar_width = 'width:' + strength + '%;';
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,140 @@
...
+        },
+
+        link: function (scope, elem, attrs, ctrl) {
+          scope.$watch('password', function (pw) {
+
+            scope.rule =
+                requirements(pw && /[A-Z]/.test(pw)) +
+                requirements(pw && /[a-z]/.test(pw)) +
+                requirements(pw && /\d/.test(pw)) +
+                requirements(pw && /(?=.*\W)/.test(pw)) +
```
lkiesow: I'm actually not entirely sure but isn't this just identical to

```suggestion
                requirements(pw && /\W/.test(pw)) +
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,140 @@
...
+
+              var uniqueChars = (pw.split('').sort().join('').replace(/(.)\1+/g, '').length) * 2;
+              var pw_length = pw.length * 4;
+              var lowerCase = (pw.length - pw.replace(/[a-z]/g, '').length) * 2;
+              var upperCase = (pw.length - pw.replace(/[A-Z]/g, '').length) * 2;
+              var number = (pw.length - pw.replace(/[0-9]/g, '').length) * 4;
+              var symbol = (pw.length - pw.replace(/\W/g, '').length) * 6;
+
+              var strength = usedRules + uniqueChars + pw_length + lowerCase + upperCase + number + symbol;
+              return strength;
```
lkiesow: Let's assume we chose the password `a`.
The password `a`…
- is not in the list of bad passwords. Thus, we continue to the ranking.
- does not match 4 out of the 5 rules, giving us a starting score of `-20`.
- gets `2` points for having one unique character.
- gets `4` points for length.
- gets `2` points for lowercase character length.

We end up with `-20 + 2 + 4 + 2 = -12`.
But `-12` does not match any of your conditions below.
You probably want all of these to be very weak:

```suggestion
              return Math.max(1, usedRules + uniqueChars + pw_length + lowerCase + upperCase + number + symbol);
```

modules/admin-ui-frontend/app/scripts/shared/directives/passwordStrengthDirective.js
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            function setProgBar(strength) {
+
+              if (strength >= 90) {
+                bar_color = 'background:#388ed6;';
+                bar_width = 'width:' + strength + '%;';
+                bar_text = 'Very strong';
```
lkiesow: These strings should probably be translated:

https://github.com/opencast/opencast/blob/cdfd04c33725fc71c3d682859fc863fe6937b668/modules/admin-ui-frontend/app/scripts/shared/services/restServiceMonitor.js#L72

lkiesow (changes requested):

lkiesow (changes requested):
Generally looks good.

The only thing missing is the source translation. [lang-en_US.json](https://github.com/opencast/opencast/blob/52b63cef0beb180a64c51e67f5db68ddc21517b9/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json) is used as source of the translation. All other translation files are automatically overwritten by whatever was translated on Crowding. You modified `en_GB` and `de_DE`. Both will be overwritten. You can probably just copy over the changes from `en_GB` to `en_US`.

gregorydlogan:
@naweidner are you moving forward with this?

narickmann:
Hopefully I've adressed through all the remaining open points.

lkiesow (approved):

PR #3372, opened 2022-01-22
Fix port checking

lkiesow:
Opencast's start script includes a check for if the TCP port Opencast
wants to use is already in use. Unfortunately, this would also check if
you are connected to this host on another machine. This prevented me
from starting Opencast while I was accessing Opencast on another
machine.

This patch improves the situation by looking at local TCP ports only.

Reference: [`/proc/net/tcp`](https://www.kernel.org/doc/Documentation/networking/proc_net_tcp.txt)

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/assemblies/resources/bin/check_ports
+++ b/assemblies/resources/bin/check_ports
@@ -41,7 +41,7 @@ if [ "$PROGNAME" = "start-opencast" ] && [ "$nargs" -eq "0" ]; then
       return 0;
     fi
     if [ -d /proc/net ] ; then
-      if cat /proc/net/tcp* | awk '{print $2, $3, $4}' | grep " 0[1A]$" | grep -q "$hexPort"; then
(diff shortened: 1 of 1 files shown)


gregorydlogan (approved):
Looks reasonable (as much as reading the contents of /proc/net/tcp can be reasonable) and seems to work.  This is probably also going to resolve the issue I've been seeing where connecting to a remote Opencast instance on 8080, and then trying to start a completely unrelated local one would fail with port in use.

lkiesow:
That's exactly the reason I wrote this change.
The port check still isn't great though and there are more things which can fail
This just makes it a bit better.

The documentation I linked above also stated that these files are deprecated.
That makes me wonder if we want to keep this script around for long anyway.
Still, for now this works, and works better than it did before ;-)

gregorydlogan:
Long term yeah we should probably migrate to something using `tcp_diag`, but I suppose the question is whether this is present in a Mac env, and which version(s) of their OS do/do not have it.

PR #3384, opened 2022-01-26
Add DublinCore Catalog to Scheduler JSON

lkiesow:
This patch adds the episode Dublin Core catalog to the JSON version of
the capture agent calendars. The new JSON response will look like this:

```json
[
  {
    "agent": {…},
    "episode-dublincore": {…}
  }
]
```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
+++ b/modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
@@ -578,10 +578,13 @@ public class SchedulerRestService {
         }
       }
 
-      var result = new ArrayList<TechnicalMetadata>();
+      var result = new ArrayList<Map<String, Object>>();
       for (var event: service.search(agent, Opt.none(), Opt.none(), Opt.some(new Date()), endDate)) {
         var id = event.getIdentifier().toString();
-        result.add(service.getTechnicalMetadata(id));
+        result.add(Map.of(
+                "data", service.getTechnicalMetadata(id),
+                "episode-dublincore", service.getDublinCore(id).toXmlString()
+                ));
       }
 
       final ResponseBuilder response = Response.ok(gson.toJson(result));


modules/scheduler-impl/src/main/java/org/opencastproject/scheduler/endpoint/SchedulerRestService.java
```diff
@@ -578,10 +578,13 @@ public Response getCalendarJson(
         }
       }
 
-      var result = new ArrayList<TechnicalMetadata>();
+      var result = new ArrayList<Map<String, Object>>();
       for (var event: service.search(agent, Opt.none(), Opt.none(), Opt.some(new Date()), endDate)) {
         var id = event.getIdentifier().toString();
-        result.add(service.getTechnicalMetadata(id));
+        result.add(Map.of(
+                "agent", service.getTechnicalMetadata(id),
```
mtneug: Not sure about the `agent` naming. What do you think about `meta`?

lkiesow: I don't really care that much, though I don't quite understand what `meta` would mean in this context, since you then would have meta(data?) and other metadata?

mtneug: mmh ok fair enough. Still not convinced about `agent` as this does not describe the agent 😄. Maybe `episode`, `event`, `data`?

lkiesow: `episode` and `event` clash with the `episode-dublincore` field.
`data` is pretty much neutral. So, I change it like this?

```json
[
  {
    "data": {…},
    "episode-dublincore": {…}
  }
]
```

…an alternative would still be to have it inside the object:


```json
[
  {
    "id": …,
    "…": …,
    "episode-dublincore": {…}
  }
]
```

What would you prefer?

mtneug: Personally, I would prefer the first one. So just a rename.

lkiesow: Then `data` it is. Updated.

mtneug (approved):

PR #3409, opened 2022-02-01
duplicate WFH changes title and StartDateTime

ts23:
For rolling forward Lectures from one year to the next we need to be able to change the Title and Start Date and Time of the duplicated mediapackage.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -33,6 +33,10 @@ Parameter Table
 |copy-number-prefix   |`copy`                                     |The prefix used for the number of the copy which is appended to the title of the new event.     |
 |number-of-events     |`2`                                        |How many events to create.|
 |max-number-of-events |`1000`                                     |How many events are allowed to be created at maximum.|
+|no-suffix            |`true`                                     |Boolean to decide if the number suffix (`copy #`) is not attached (true ... no attachment, false ...`copy #` is attached) |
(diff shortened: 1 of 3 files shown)


joneug:
I tested this and the given title is set correctly. However, the given date is only reflected in the "Created" metadata filed not in the "Start date" field. Is this what you wanted?

Also, I cannot start the duplicate-event workflow without specifying a start date. If I try that the following exception is thrown: `java.lang.RuntimeException: Bug. At least one of the value columns must be non null.`

gregorydlogan:
@ts23 ping

ts23:
@gregorydlogan Pong ... still looking at it ... making progress ... trying to sort startDate

gregorydlogan:
Converting to a draft since I almost accidentally merged this today :D

tyfranck:
I'll pick this one up. I don't have access to  @ts23 repository so will probably have to create a new pull request based on this.

tyfranck:
I've created a new PR #3805 which should address the issue highlighted.  Please take a look. 
I'm going to close this PR.

tyfranck:
> I tested this and the given title is set correctly. However, the given date is only reflected in the "Created" metadata filed not in the "Start date" field. Is this what you wanted?

Should now be fixed please see PR #3805

> Also, I cannot start the duplicate-event workflow without specifying a start date. If I try that the following exception is thrown: `java.lang.RuntimeException: Bug. At least one of the value columns must be non null.`

This is an UI issue with input set as date.  Should now be possible to run duplicate-event workflow without start date.

tyfranck:
See PR #3805

PR #3444, opened 2022-02-11
Fix frontend-no-prebuilt profile

lkiesow:
This patch fixes the frontend-no-prebuilt maven profile which can be
used to have maven run the local npm version instead of downloading one.
That has the big advantage of not having to remove any `none*` folders
when you switch versions.

To try this, run:

```
❯ mvn clean install -Pdev,frontend-no-prebuilt
```

The profile did exist for a long time but was broken due to some tasks
being executed in the wrong order like trying to copy the frontend to
the target folder before actually building it.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/pom.xml
+++ b/modules/admin-ui-frontend/pom.xml
@@ -44,8 +44,9 @@
                 </goals>
                 <phase>initialize</phase>
                 <configuration>
-                  <executable>node_modules/.bin/bower</executable>
+                  <executable>npx</executable>
                   <arguments>
+                    <argument>bower</argument>
                     <argument>--allow-root</argument>
                     <argument>install</argument>
                   </arguments>
@@ -56,17 +57,18 @@
                 <goals>
                   <goal>exec</goal>
                 </goals>
-                <phase>compile</phase>
+                <phase>initialize</phase>
                 <configuration>
-                  <executable>node_modules/grunt-cli/bin/grunt</executable>
+                  <executable>npx</executable>
                   <arguments>
+                    <argument>grunt</argument>
                     <argument>build</argument>
                     <argument>--skipTests=${skipTests}</argument>
                   </arguments>
                 </configuration>
               </execution>
               <execution>
-                <id>esline</id>
+                <id>eslint</id>
                 <goals>
                   <goal>exec</goal>
                 </goals>
--- a/modules/engage-paella-player/pom.xml
+++ b/modules/engage-paella-player/pom.xml
@@ -37,11 +37,11 @@
                 </configuration>
               </execution>
               <execution>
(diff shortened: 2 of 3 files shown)


mliradelc:
Tried and I'm having this problem with the lti module when is being compiled with the parameters you give in the description.

```
node:internal/modules/cjs/loader:488
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './2020/CreateDataProperty' is not defined by "exports" in /Users/maxlira/repos/oc-official/modules/lti/node_modules/es-abstract/package.json
    at new NodeError (node:internal/errors:371:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:429:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:683:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/Users/maxlira/repos/oc-official/modules/lti/node_modules/object.getownpropertydescriptors/implementation.js:3:26) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}

Node.js v17.3.0
[ERROR] Command execution failed.
```

I tried again but with the normal `mvn clean install -Pdev` without any issues

lkiesow:
You are using Node 17.x while Opencast supports the latest LTS which is 16.x.
Not sure if that causes the issue and we will eventually need to fix it anyway, but can you try using 16 for testing this?
This patch is only about the build process and not about the code after all.

lkiesow:
Yep, it's likely you using Node 17 since [Node 16 already gives a deprecation warning](https://github.com/ljharb/es-abstract/issues/133) for this.

mliradelc:
I though that maven downloads the nodejs version required for the project. I have to check that.

mliradelc (approved):
Works as intended! Merged!

mliradelc:
Note: 
If any encounter problem compiling, use `Node version Manager, NVM` to change to the latest LTS version (In this case the 16)

lkiesow:
The idea behind the `frontend-no-prebuilt` profile is basically to tell Maven to not download Node but to use the system version.
That is why you can do things like jumping between Opencast versions without having to remove it.
You are using the same version of node after all.
The downside of this is that you are now responsible for your environment yourself.

Issue #1374, opened 2020-01-21
When there is a series without a title, the metadata tabs break


JulianKniephoff:
After creating series with an empty title, which is surprisingly easy to do, considering that it is marked as a required field in the admin UI, you can't enter new or edit existing metadata anymore. The metadata tab is just empty:

![image](https://user-images.githubusercontent.com/123272/72820379-28b1d180-3c6f-11ea-9049-85d5eba0398f.png)

Apparently the `metadata` and `metadata.json` endpoints have a problem serializing this (which they have to do since they return a list of possible series to display). Interestingly the requests still succeed with a 200 code; they just don't have any content.

The serialization error can be seen in the log, and is ultimately due to some `NullPointerException`. Unfortunately I could not investigate further, yet.

karendolan:
@JulianKniephoff I am surprised the Series service did not reject the creation of a series without a title. This seems  like an easy fix to enforce a default series title (i.e. `S-<seriesId>`) upon series creation.

[EDIT] I recommend the default series title because many places in the code expect a minimum ID and Title for both series and episode. If one or the other does not exist, it impacts more than the External API api and the Admin api.

ebbertd:
@JulianKniephoff Is this fixed now?

JulianKniephoff:
Seems to be fixed! :D

PR #3460, opened 2022-02-16
fix admin-ui endpoint crash, when a series without title exists

darolfes:
The metadata tab would break, when there is a series without title (see #1374 ).
This pullrequest fixes the endpoint, that causes this behavior, by simple removing series with empty titles. Meaning the metadata tab loads again, but series with empty titles are not selectable (other series work fine).

Not sure if this should close the issue, as this only fixes a major symptom, but not the underlying problem. Empty series titles still cause small problems in other areas (e.g. attributes of series with empty titles sometimes can't be modified) and can still be created over the rest api.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1189,7 +1189,7 @@ public abstract class AbstractEventEndpoint {
 
     // Remove null keys
     Map<String, String> seriesCollection = seriesField.getCollection();
-    while (seriesCollection.remove(null) != null) { }
+    seriesCollection.remove(null);
     seriesField.setCollection(seriesCollection);
 
     return;
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -1172,6 +1172,29 @@ public abstract class AbstractEventEndpoint {
     }
   }
 
+  /**
+   * Removes emtpy series titles from the collection of the isPartOf Field
+   * @param ml the list to modify
+   */
+  private void removeSeriesWithNullTitlesFromFieldCollection(MetadataList ml) {
+    // get Series MetadataField from MetadataList
+    MetadataField seriesField = Optional.ofNullable(ml.getMetadataList().get("dublincore/episode"))
+            .flatMap(titledMetadataCollection -> Optional.ofNullable(titledMetadataCollection.getCollection()))
+            .flatMap(dcMetadataCollection -> Optional.ofNullable(dcMetadataCollection.getOutputFields()))
+            .flatMap(metadataFields -> Optional.ofNullable(metadataFields.get("isPartOf")))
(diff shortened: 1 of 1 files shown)


modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1172,6 +1172,29 @@ public Response createEventCommentReply(@PathParam("eventId") String eventId, @P
...
+            .flatMap(dcMetadataCollection -> Optional.ofNullable(dcMetadataCollection.getOutputFields()))
+            .flatMap(metadataFields -> Optional.ofNullable(metadataFields.get("isPartOf")))
+            .orElse(null);
+    if (seriesField == null || seriesField.getCollection() == null) {
+      return;
+    }
+
+    // Remove null keys
+    Map<String, String> seriesCollection = seriesField.getCollection();
+    while (seriesCollection.remove(null) != null) { }
```
lkiesow: This is just a confusing way of just doing:
```suggestion
    seriesCollection.remove(null);
```
Map keys have to be unique, or don't they? That means there can only be one element with key `null` in the `Map`. That means you only ever call map twice in this loop and the last time it will never do anything.

darolfes: Yes, of course.

gregorydlogan:
@darolfes while I'm fine with this solution, are you able to address the underlying API issue?  It should not be possible to create invalid series via our own rest api :D

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -1172,6 +1172,29 @@ public Response createEventCommentReply(@PathParam("eventId") String eventId, @P
...
+   */
+  private void removeSeriesWithNullTitlesFromFieldCollection(MetadataList ml) {
+    // get Series MetadataField from MetadataList
+    MetadataField seriesField = Optional.ofNullable(ml.getMetadataList().get("dublincore/episode"))
+            .flatMap(titledMetadataCollection -> Optional.ofNullable(titledMetadataCollection.getCollection()))
+            .flatMap(dcMetadataCollection -> Optional.ofNullable(dcMetadataCollection.getOutputFields()))
+            .flatMap(metadataFields -> Optional.ofNullable(metadataFields.get("isPartOf")))
+            .orElse(null);
+    if (seriesField == null || seriesField.getCollection() == null) {
+      return;
```
karendolan, line 1187: It would be helpful to log with INFO or WARN, here, the ID of each series that is being omitted from the admin UI

KatrinIhler:
@gregorydlogan Fixing the API, while a good idea, is unfortunately not in scope for this. But this way we can at least stop it from bringing down the Admin UI upload every couple of months...

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
KatrinIhler: Be aware that the series still show up in the Admin UI table (though series with empty titles will always be at the bottom of the last page, no matter how you sort...), it just won't be in the series dropdown in the metadata tab.

PR #3471, opened 2022-02-21
Matrix room description update fix

Alaschinger:
The Matrix room description of the development team doesn't automatically update to the latest version in case of a new release. This should be fixed by using the 'tags' instead of the 'releases' list.


lkiesow (approved):
Apereo got an ICLA from Alex.

PR #3540, opened 2022-03-17
Fix NPE when signing publication URLs

mtneug:
Publication URLs might be null. Signing them leads to a NPE.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -2697,6 +2697,9 @@ public abstract class AbstractEventEndpoint {
   }
 
   private URI signUrl(URI url) {
+    if (url == null) {
+      return null;
+    }
     if (getUrlSigningService().accepts(url.toString())) {
       try {
         String clientIP = null;


reiferschris (approved):
If I made no mistakes while testing this, I can approve this change. Our dev-system does not show errors in the assets/publications tab anymore after applying this fix. Thank you!

KatrinIhler:
Is this going to be a problem for the External API as well? It can sign publication URLs since #3002.

reiferschris:
I tried testing if the external api has problems but could not reproduce it with querying one of the /api/events//{eventId}/publications/{publicationId}?sign={sign} Endpoints.. But I am not even sure if that would cause it to fail in the first place.
I found this line in your code @KatrinIhler https://github.com/opencast/opencast/pull/3002/files#diff-606d2a3ea1a0603656eb303a059677b604f3bb095b7f81ec97519869f5a751d0R1620 which I would assume prevents this from being a problem, looks kind of similiar to the fix Matthias provided.

mliradelc (approved):
Can confirm, this fix works.

From @reiferschris confirmation, we are sure that the External API is not affected by this bug

mtneug:
@mliradelc: I wondered as well and looked at the event endpoints. [The `getSignedUrl` method](https://github.com/opencast/opencast/blob/87d371ce918dd2b4009d17206d1b26ee6b8cb616/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java#L1620-L1633) actually has a null check so it should not be affected.

Issue #3510, opened 2022-03-03
Write permission gets lost during cleanup operation


jojoob:
**Describe the bug**
When creating an event via admin UI, uploading a video for the presentation track and select the fast workflow the video processes just fine.
On file system level however the permissions looking odd:

```bash
root@ocadmin:~# ls -l /opencast1/archive/mh_default_org/d9766e49-6b42-43d1-a253-641c6ea7908d/*
/opencast1/archive/mh_default_org/d9766e49-6b42-43d1-a253-641c6ea7908d/0:
total 12
-rw-r--r-- 2 opencast opencast  521 Mar  3 15:14 9e6e7e6e-832c-4d7d-ac98-a8417e71e8d4.xml
-rw-r--r-- 1 opencast opencast  830 Mar  3 15:14 manifest.xml
-rw-r--r-- 4 opencast opencast 2281 Mar  3 15:14 security-policy-episode.xml

/opencast1/archive/mh_default_org/d9766e49-6b42-43d1-a253-641c6ea7908d/1:
total 1928
-rw-r--r-- 2 opencast opencast     521 Mar  3 15:14 9e6e7e6e-832c-4d7d-ac98-a8417e71e8d4.xml
-r--r--r-- 1 opencast opencast 1956169 Mar  3 15:14 dc0dee77-121f-493b-94f4-627e65061e58.webm
-rw-r--r-- 1 opencast opencast    5160 Mar  3 15:16 manifest.xml
-rw-r--r-- 4 opencast opencast    2281 Mar  3 15:14 security-policy-episode.xml

root@ocadmin:~# find /opencast1/ ! -perm -u=w -path "*d9766e49-6b42-43d1-a253-641c6ea7908d*" -printf "%M %p\n"
-r--r--r-- /opencast1/archive/mh_default_org/d9766e49-6b42-43d1-a253-641c6ea7908d/1/dc0dee77-121f-493b-94f4-627e65061e58.webm
```

When starting a "Publish" workflow on this just uploaded and successful processed event the workflow fails.

<details>

```
job-operation
Distribute
job-type
org.opencastproject.distribution.download
stack-trace
org.opencastproject.serviceregistry.api.ServiceRegistryException: Error handling operation 'Distribute'
	at org.opencastproject.distribution.download.DownloadDistributionServiceImpl.process(DownloadDistributionServiceImpl.java:695)
	at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:313)
[...]

PR #3541, opened 2022-03-17
Downgrade commons-io

jojoob:
This pull request partly reverts #3328 because of an issue under some environments. See #3510


mliradelc (approved):
Tested, I tried unpublish, then publish and I'm not getting any issues.

Approved!

mliradelc:
After talking about in the technical meeting of today, It was agreed to merge it :)

jojoob:
It should be save (as far as #3510 is concerned) to upgrade to 2.12 once it is released.

PR #3544, opened 2022-03-21
Fix bug: Editor sometimes doesn't skip deactivated sections on Safari Browser

darolfes:
Under some circumstances the Safari Browser doesn't skip deactivated Sections in the Editor. For Example when using the
"Play Transition"-Button. This PR fixes this problem.

Reason for the Bug: Safari doesn't trigger a "Playing"-Event on finishing the seek (in contrast to firefox/chrome). This Event would
restore the Playback-State in the Editor. Instead the Editor remains in the Seeking-State, resulting in faulty behavior.

This fix additionally listens on the "Seeked"-Event, which is triggered by Safari (and some other Browsers), when
the seek is finished and restores the State.


mliradelc (approved):
Tested on my Mac. works as expected

PR #3553, opened 2022-03-23
Fully Automated Release

gregorydlogan:
In theory this should work, however...
- I have not tested that the calculatings for the major and minor versions actually work the way I think they do
- The text of the release is not properly set.  We probably want logic to do different release text depending on if it's a major release, or a minor.
  - In the case of a major we (IMO) have something very vague and ensure the RMs come back and update that text
  - In the case of a minor we (IMO) have something static like "This is a maintenance release for X.y", and the RMs can come back to update it if they feel like they need to.
- The release is still set to be a draft.  Before this goes in we should remove that.

Any other issues or objections people can see?

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -41,6 +41,14 @@ jobs:
         id: get_version
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
 
+      - name: Get the major version
+        id: get_major_version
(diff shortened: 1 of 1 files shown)


lkiesow:
Looks good. Once tested, we should also switch from draft mode to release mode.

gregorydlogan:
#4091 is cleaner :P

PR #3555, opened 2022-03-24
Rework the userdirectory-brightspace module

kristofkeppens:
This fixes the userdirectory module to work with the latest version of the brightspace api, reworks the way it works ( identical to the canvas userprovider ) so you retrieve the user roles in an identical fashion to the roles you get via LTI.


gregorydlogan:
I don't really have a way to test this, but considering you're using this in production I'm going to assume it works :)

PR #3567, opened 2022-03-24
Fix workflow definition installation

mtneug:
When installing workflow definitions there is a chance that a workflow
definition which fails to load will result in stalling the boot of
further services. This happens when the faulty workflow definition is
loaded last and the `ReadinessIndicator` service is not registered.
This rearranges the code to make sure the service is always registered
when the last workflow definition was handled.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


mliradelc:
I think this PR is very important because it's a big bug, maybe with @joneug we should make a new release of opencast to address this asap.

I'm trying to trigger the bug, but I can't. I have seen this bug before, @mtneug Have you any tips on how to trigger more easily the bug?

mtneug:
Most should hopefully not be affected so I would not create a new release just for that. I see this when testing locally with multiple tenants and workflow definitions that only belong to a specific org. In this case, it could be the case that the org is not yet created so the installation of the workflow definition will fail.

To trigger this, you could introduce a syntax error or org that does not exist in all/most workflow definitions.

PR #3586, opened 2022-03-29
Auto-generated series for Capture Agents

Alaschinger:
Integrated code from switchcast in the ingest-service module to auto-generate a series, if an event without series is ingested. Events without series may be inaccessible for users without admin rights. So this patch provides, that people with rights to the CA have access to the respective series and the contained events.


etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg
```diff
@@ -54,3 +54,9 @@
 #org.opencastproject.download.source = http://localhost/.*
 #org.opencastproject.download.user = opencast_system_account
 #org.opencastproject.download.password = CHANGE_ME
+
+# Control if events without series are added to an auto generated CA series. This ensures that an event without series is accessible for users with rights to the CA.
+#
+# Default: false
```
lkiesow: I think this does something else. It's a prefix, not a boolean. Though, having this unset will disable the functionality. Also, the default is not `false` but EMPTY.

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
```diff
@@ -1949,4 +1976,83 @@ public Track xapply(Job job) throws MediaPackageException {
...
+    }
+    //Get capture agent name
+    String captureAgentId = null;
+    Catalog[] catalog = mp.getCatalogs(MediaPackageElementFlavor.flavor("dublincore", "episode"));
+    if (catalog.length == 1) {
+      try (InputStream catalogInputStream = workingFileRepository.get(mp.getIdentifier().toString(), catalog[0].getIdentifier())) {
+        DublinCoreCatalog dc = dublinCoreService.load(catalogInputStream);
+        captureAgentId = getCaptureAgent(dc);
+      } catch (Exception e) {
+        logger.warn("Unable to determine capture agent name", e);
```
lkiesow: Will this cause a warning being logged every time someone ingests something not from a capture agent?

lkiesow (changes requested):

lkiesow (approved):

PR #3603, opened 2022-03-31
Termination service persists maintenance

### LINKED ISSUE #3602: Termination State Services leaves nodes in maintenance mode
**Describe the bug**
The termination state service puts a node into maintenance mode while if completes jobs and terminates. If a node is started with the same IP with will start in the maintenance mode and not accept jobs.

**To Reproduce**
Run AWS autoscaling nodes in a small enough subnet that when terminated the same internal IP address will be reused. The new instance will start is the maintenance mode of the previous.

**Environment Information**
AWS autoscaling with termination state service enabled

JamesUoM:
Fixes #3602 Termination State Services leaves nodes in maintenance mode

Check on start up that host is not in maintenance, due to state from a previous instance


JamesUoM:
I'm out of practice, against wrong branch

PR #3604, opened 2022-03-31
New editor - Save subtitles in Opencast

Arnei:
A new feature is currently in development for the new editor -  a subtitle editor. It will allow editing or creating subtitle tracks in the new editor UI and then saving them in Opencast. This modifies the editor service endpoint to allow the saving of subtitles. 

The POST "{mediaPackageId}/edit.json" endpoint is extended to accept a list of subtitles and flavors which are to be saved in the mediapackage. Subtitles in existing flavors are overwritten, while subtitles in new flavors are added as new tracks.

Does not perform any checks (e.g. validity) on the submitted subtitles.

Depends on #3601

* [ ] have a concise title
* [ ] pass automated tests


modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+   *
+   * @param mediaPackage
+   *          the media package to at the SMIL catalog
+   * @param subtitles
+   *          the subtitles to be added
+   * @throws IOException
+   */
+  private void addSubtitleTrack(MediaPackage mediaPackage, List<PostEditingData.Subtitle> subtitles)
+          throws IOException {
+    if (subtitles != null) {
```
metzenseifner: Consider replacing with `Objects.requireNonNull(subtitles, "Argument subtitles may not be null");`, thereby throwing a meaningful NullPointerException. Reasoning: subtitles being null is clearly an invalid state. As it is now, the caller has no way of knowing that something went awry.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -786,6 +867,12 @@ public void setEditData(String mediaPackageId, EditingData editingData) throws E
       errorExit("Unable to add SMIL cutting catalog to archive", mediaPackageId, ErrorStatus.UNKNOWN, e);
     }
 
+    try {
+      addSubtitleTrack(mediaPackage, editingData.getSubtitles());
+    } catch (IOException e) {
```
metzenseifner, line 1055: } catch (IOException | NullPointerException e) {

(Should you decided to heed my suggestion about throwing NullPointerException)

metzenseifner (changes requested):
I think my suggestions would help give the user better feedback in the problem case.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
Arnei: Thanks, will do.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
Arnei: Caught the NullPointerException in a different clause in [7a78ba0](https://github.com/opencast/opencast/pull/3604/commits/7a78ba0efafe259f6663aaf6efaf818e59d61f00), because it is perfectly valid for the frontend to not send any subtitles.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+      // Memorize uri of the previous track file for deletion
+      URI oldTrackURI = null;
+      if (track != null) {
+        oldTrackURI = track.getURI();
+      }
+
+      // Put updated filename in working file repository and update the track.
+      InputStream is = IOUtils.toInputStream(subtitle.getSubtitle(), "UTF-8");
+      URI subtitleUri = workspace.put(mediaPackage.getIdentifier().toString(), subtitleId, "subtitle.vtt",
+              //                  EditorService.TARGET_FILE_NAME,
```
lkiesow:
```suggestion
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+          throws IOException, NullPointerException {
+    Objects.requireNonNull(subtitles, "Argument subtitles may not be null");
+
+    for (PostEditingData.Subtitle subtitle : subtitles) {
+      // Generate ID for new tracks
+      String subtitleId = UUID.randomUUID().toString();
+      String trackId = null;
+
+      // Check if subtitle already exists
+      for (Track t : mediaPackage.getTracks()) {
```
lkiesow, line 512: This allows users to easily replace any track with any content. It may be a good idea to limit this to caption flavors only.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+      Track track = mediaPackage.getTrack(trackId);
+
+      // Memorize uri of the previous track file for deletion
+      URI oldTrackURI = null;
+      if (track != null) {
+        oldTrackURI = track.getURI();
+      }
+
+      // Put updated filename in working file repository and update the track.
+      InputStream is = IOUtils.toInputStream(subtitle.getSubtitle(), "UTF-8");
```
lkiesow: It's a good idea to always close input streams

```suggestion
      try (InputStream is = IOUtils.toInputStream(subtitle.getSubtitle(), "UTF-8")) {
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+      }
+
+      track.setURI(subtitleUri);
+      track.setIdentifier(subtitleId);
+      for (String tag : getSmilCatalogTags()) {
+        track.addTag(tag);
+      }
+      track.setChecksum(null);
+
+      if (oldTrackURI != null) {
```
lkiesow: The old and new URIs could be identical, in which case you would not want to delete this here

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+          logger.info("Could not remove track from workspace. Could be it was never there.");
+        }
+      }
+
+      try {
+        assetManager.takeSnapshot(mediaPackage);
+      } catch (AssetManagerException e) {
+        logger.error("Error while adding the updated media package ({}) to the archive", mediaPackage.getIdentifier(),
+                e);
+        throw new IOException(e);
```
lkiesow: Don't log an error if you return an error. Add the message to the thrown exception:

```suggestion
        throw new IOException("Error while adding the updated media package " + mediaPackage + " to the archive", e);
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+      }
+    }
```
lkiesow, line 627: You leave files in the workspace. Once added to the asset manager, you should call `workspace.delete(…)` to remove the subtitle from there.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -420,6 +422,84 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+        track = (Track) mpeBuilder.elementFromURI(subtitleUri, MediaPackageElement.Type.Track, subtitle.getFlavor());
+        mediaPackage.add(track);
+        logger.info("Creating new track for flavor: " + track.getFlavor());
+      }
+
+      track.setURI(subtitleUri);
+      track.setIdentifier(subtitleId);
+      for (String tag : getSmilCatalogTags()) {
+        track.addTag(tag);
+      }
```
lkiesow: This does not fit the documentation for these tags which is:

```properties
# Sets the tag that should be applied to the generated smil catalog                                                      
# Default: archive                                                                                                       
#smil.catalog.tags=archive 
```

A subtitle is no SMIL catalog. I would just drop these lines:

```suggestion
```

lkiesow (changes requested):
I found a few minor problems, but nothing major. Before we merge this, we may want to sit down and discuss how subtitles should actually be named/tagged/flavored though. This code (and related pull requests) feel like we have too many implicit assumptions about how things could work.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
Arnei: Yeah, I wasn't sure what the bigger issue was, being able to replace any track, or being limited to the (implicit standard)`captions/*` flavor.

lkiesow:
Maybe you missed the CI failure, but this does not build any longer

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/EditingData.java
```diff
@@ -61,10 +61,10 @@ public EditingData(List<SegmentData> segments, List<TrackData> tracks, List<Work
     this.waveformURIs = waveformURIs;
   }
 
-  public static EditingData parse(String json) {
+  public static PostEditingData parse(String json) {
```
lkiesow: Why would a static method of the class `EditingData` return `PostEditingData`?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lkiesow: This is now caught by
```java
if (!subtitle.getFlavor().matches(captionsFlavor)) {
```

lkiesow (changes requested):
I'll write it down here as well as on the frontend pull request since it touches both parts:

I did only briefly skim through the code, but if I interpret the code correctly, subtitles are loaded from the internal publication, but then saved to the asset manager. This means that you don't necessary get what you save the next time you open the editor. It also means that you may overwrite completely different subtitle files and you might also miss subtitle files which are only in the asset manager.

My suggestion would be to always get the subtitles from the asset manager?

Arnei:
Rebased this on r/12.x

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/PostEditingData.java
```diff
@@ -0,0 +1,63 @@
...
+package org.opencastproject.editor.api;
+
+import org.opencastproject.mediapackage.MediaPackageElementFlavor;
+
+import java.util.List;
+
+/**
+ * Wrapper for additional data that may be received by a post request from the editor
+ */
+public final class PostEditingData extends EditingData {
```
lkiesow: Does it make sense to use this for both ways from and to the editor? The only difference is the subtitle and we could provide it using this method as well.

_(see also comment above)_

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -437,6 +447,82 @@ void addSmilToArchive(MediaPackage mediaPackage, final Smil smil) throws IOExcep
...
+   * @param subtitles
+   *          the subtitles to be added
+   * @throws IOException
+   */
+  private void addSubtitleTrack(MediaPackage mediaPackage, List<PostEditingData.Subtitle> subtitles)
+          throws IOException {
+    for (PostEditingData.Subtitle subtitle : subtitles) {
+      if (!subtitle.getFlavor().matches(captionsFlavor)) {
+        break;
+      }
```
lkiesow: Maybe raise an exception which ends in a 400 Bad Request for the user, so it is clear that the request wasn't processed successfully and, in particular, wasn't saved as sent by the user. Also, maybe not save anything if half of the request is bad? Otherwise, you don't know the state you end up with.

lkiesow (changes requested):
Generally looks good and you would now also get the saved subtitles if access control lets you, which might be a problem in a cluster with the current solution (see below). 

We also found another minor issue which should never be a problem if everything goes well, but still might be worth fixing.

For getting the subtitle to the frontend, right now, you just replace the internal publication with a track containing a direct URI to the asset manager. I'm not sure if you can always access the asset manager directly like this, e.g., if you are on a presentation node.

You could use `EditingData` for sending the subtitle instead of replacing parts of the internal publications. This also has the advantage that the codes becomes simpler since you don't need `PostEdittingData` and `edit.json` has always the same structure.

Arnei:
> Generally looks good and you would now also get the saved subtitles if access control lets you, which might be a problem in a cluster with the current solution (see below).
> 
> We also found another minor issue which should never be a problem if everything goes well, but still might be worth fixing.
> 
> For getting the subtitle to the frontend, right now, you just replace the internal publication with a track containing a direct URI to the asset manager. I'm not sure if you can always access the asset manager directly like this, e.g., if you are on a presentation node.
> 
> You could use `EditingData` for sending the subtitle instead of replacing parts of the internal publications. This also has the advantage that the codes becomes simpler since you don't need `PostEdittingData` and `edit.json` has always the same structure.

To avoid any problems with accessing the asset manager, I've changed it just send the complete subtitle file contents instead. As you already mentioned, we already do this from Frontend-to-Backend anyway, so now it's at least consistent for both ways. Which also allows for removal of `PostEditingData`

lkiesow (approved):
Looks good and seems to work as expected. It should also be fully backwards compatible. Merging.

PR #3606, opened 2022-03-31
Editor - Cut subtitle tracks

Arnei:
A new feature is currently in development for the new editor -  a subtitle editor. It will allow editing or creating subtitle tracks in the new editor UI and then saving them in Opencast. This modifies the videoeditor service to be able to cut subtitle tracks based on the webvtt format.

As ffmpeg is severly limited in how it allows cutting subtitles (does not work with filters), this parses the webvtt subtitles into a list container to allow for fine grained editing by the videoeditor. ~~The WebVTT parser is still rather rudimentary and likely needs some improvements before merging.~~ EDIT: The WebVTT parser is still rather rudimentary, but should not fail as long as the given WebVTT file is valid.

The videoeditor now discriminates between the usual audio/video elements and reference elements (aka everything that is not explicitly specified to be an audio or video element). Within the reference elements, it looks for `.vtt` files and cuts them according to the SMIL. 

A workflow operation might look like this the following. Unfortunately, subtitle file language is encoded in the flavor subtype, making this rather awkward.
```xml
    <operation
        id="editor"
        description="Cutting the recording according to the edit decision list">
      <configurations>
        <configuration key="source-flavors">captions/vtt+en</configuration>
        <configuration key="smil-flavors">smil/tmp</configuration>
        <configuration key="target-smil-flavor">smil/tmp</configuration>
        <configuration key="target-flavor-subtype">vtt+en+trimmed</configuration>
        <configuration key="interactive">false</configuration>
      </configurations>
    </operation>
```

Furthermore, this PR modifies the engage-paella-player module to allow the Paella Player to display subtitles from tracks. It also removes the "vtt+language" requirement from the flavor subtype, so you can now have "source+en" or "prepared+en".

Depends on #3601

### Your pull request should…

* [ ] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/engage-paella-player-7/src/js/EpisodeConversor.js
+++ b/modules/engage-paella-player-7/src/js/EpisodeConversor.js
@@ -318,7 +318,7 @@ function readCaptions(potentialNewCaptions, captions) {
         captions.push({
           id: potentialCaption.id,
           lang: captions_lang,
-          text: captions_lang || 'unknown language';,
+          text: captions_lang || 'unknown language',
           url: potentialCaption.url,
           format: captions_format
         });
--- a/modules/engage-paella-player-7/src/js/EpisodeConversor.js
+++ b/modules/engage-paella-player-7/src/js/EpisodeConversor.js
@@ -302,32 +302,23 @@ function readCaptions(potentialNewCaptions, captions) {
       if (captions_match) {
         let captions_lang = captions_match[3];
 
-        // TODO: read the lang from the dfxp file
-        //if (captions_format == "dfxp") {}
-
         if (!captions_lang && potentialCaption.tags && potentialCaption.tags.tag) {
           if (!(potentialCaption.tags.tag instanceof Array)) {
             potentialCaption.tags.tag = [potentialCaption.tags.tag];
           }
           potentialCaption.tags.tag.forEach((tag)=>{
             if (tag.startsWith('lang:')){
-              let split = tag.split(':');
-              captions_lang = split[1];
+              captions_lang = tag.substring('lang:'.length);
             }
           });
         }
 
-        let captions_label = captions_lang || 'unknown language';
-        //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
-
-        let captions_format = potentialCaption.url.
-          substring(potentialCaption.url.lastIndexOf('.') + 1, potentialCaption.url.length)
-          || potentialCaption.url;
+        let captions_format = potentialCaption.url.split('.').pop();
 
         captions.push({
           id: potentialCaption.id,
           lang: captions_lang,
-          text: captions_label,
+          text: captions_lang || 'unknown language';,
           url: potentialCaption.url,
           format: captions_format
         });
@@ -366,8 +357,7 @@ function getCaptions(episode) {
           }
           currentCatalog.tags.tag.forEach((tag)=>{
             if (tag.startsWith('lang:')){
-              let split = tag.split(':');
-              captions_lang = split[1];
+              captions_lang = tag.substring('lang:'.length);
             }
           });
         }
--- a/modules/engage-paella-player-7/src/js/EpisodeConversor.js
+++ b/modules/engage-paella-player-7/src/js/EpisodeConversor.js
@@ -293,41 +293,73 @@ function processAttachments(episode, manifest, config) {
   }
 }
 
-function getCaptions(episode) {
-  const result = [];
-  let attachments = episode.mediapackage?.attachments?.attachment;
-  if (!(attachments instanceof Array)) {
-    attachments = attachments ? [attachments] : [];
-  }
+function readCaptions(potentialNewCaptions, captions) {
+  potentialNewCaptions.forEach((potentialCaption) => {
+    try {
+      let captions_regex = /^captions\/([^+]+)(\+(.+))?/g;
+      let captions_match = captions_regex.exec(potentialCaption.type);
 
+      if (captions_match) {
+        let captions_lang = captions_match[3];
 
-  attachments.forEach(att => {
-    const exp = /captions\/([a-z\d]+)(?:\+([a-z]+))?/.exec(att.type);
-    if (exp) {
-      const format = exp[1];
-      const lang = exp[2] || '';
-      result.push({
-        id: att.id,
-        lang: lang,
-        text: lang || 'Unknown language',
-        format: format,
-        url:att.url
-      });
+        // TODO: read the lang from the dfxp file
+        //if (captions_format == "dfxp") {}
+
+        if (!captions_lang && potentialCaption.tags && potentialCaption.tags.tag) {
+          if (!(potentialCaption.tags.tag instanceof Array)) {
+            potentialCaption.tags.tag = [potentialCaption.tags.tag];
+          }
+          potentialCaption.tags.tag.forEach((tag)=>{
+            if (tag.startsWith('lang:')){
+              let split = tag.split(':');
+              captions_lang = split[1];
+            }
+          });
+        }
+
+        let captions_label = captions_lang || 'unknown language';
+        //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
+
+        let captions_format = potentialCaption.url.
+          substring(potentialCaption.url.lastIndexOf('.') + 1, potentialCaption.url.length)
+          || potentialCaption.url;
+
+        captions.push({
+          id: potentialCaption.id,
+          lang: captions_lang,
+          text: captions_label,
+          url: potentialCaption.url,
+          format: captions_format
+        });
+      }
     }
+    catch (err) {/**/}
   });
+}
 
-  let catalogs = episode.mediapackage?.metadata?.catalog;
-  if (!(catalogs instanceof Array)) {
-    catalogs = catalogs ? [catalogs] : [];
-  }
+function getCaptions(episode) {
+  var captions = [];
 
+  var attachments = episode.mediapackage.attachments.attachment;
+  var catalogs = episode.mediapackage.metadata.catalog;
+  var tracks = episode.mediapackage.media.track;
+  if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
+  if (!(catalogs instanceof Array)) { catalogs = catalogs ? [catalogs] : []; }
+  if (!(tracks instanceof Array)) { tracks = tracks ? [tracks] : []; }
+
+  // Read the attachments
+  readCaptions(attachments, captions);
+
+  // Read the tracks
+  readCaptions(tracks, captions);
+
+  // Read the catalogs
   catalogs.forEach((currentCatalog) => {
     try {
       // backwards compatibility:
       // Catalogs flavored as 'captions/timedtext' are assumed to be dfxp
       if (currentCatalog.type == 'captions/timedtext') {
         let captions_lang;
-
         if (currentCatalog.tags && currentCatalog.tags.tag) {
           if (!(currentCatalog.tags.tag instanceof Array)) {
             currentCatalog.tags.tag = [currentCatalog.tags.tag];
@@ -341,7 +373,7 @@ function getCaptions(episode) {
         }
 
         let captions_label = captions_lang || 'unknown language';
-        result.push({
+        captions.push({
           id: currentCatalog.id,
           lang: captions_lang,
           text: captions_label,
@@ -353,7 +385,7 @@ function getCaptions(episode) {
     catch (err) {/**/}
   });
 
-  return result;
+  return captions;
 }
 
 export function episodeToManifest(ocResponse, config) {
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -274,6 +274,10 @@ class OpencastToPaellaConverter {
           if (videoCanvas) {
             currentStream.canvas = [videoCanvas];
           }
+        } else {
+          // Explicitly set the type, else Paella will assume this track to be 'video' type (Causing an exception
+          // if it is actually not)
+          currentStream.type = 'noType';
         }
       }
     });
@@ -379,20 +383,11 @@ class OpencastToPaellaConverter {
     return paellaStreams;
   }
 
-  getCaptions(episode) {
-    var captions = [];
-
-    var attachments = episode.mediapackage.attachments.attachment;
-    var catalogs = episode.mediapackage.metadata.catalog;
-    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
-    if (!(catalogs instanceof Array)) { catalogs = catalogs ? [catalogs] : []; }
-
-
-    // Read the attachments
-    attachments.forEach((currentAttachment) => {
+  readCaptions(potentialNewCaptions, captions) {
+    potentialNewCaptions.forEach((potentialCaption) => {
       try {
         let captions_regex = /^captions\/([^+]+)(\+(.+))?/g;
-        let captions_match = captions_regex.exec(currentAttachment.type);
+        let captions_match = captions_regex.exec(potentialCaption.type);
 
         if (captions_match) {
           let captions_format = captions_match[1];
@@ -401,11 +396,11 @@ class OpencastToPaellaConverter {
           // TODO: read the lang from the dfxp file
           //if (captions_format == "dfxp") {}
 
-          if (!captions_lang && currentAttachment.tags && currentAttachment.tags.tag) {
-            if (!(currentAttachment.tags.tag instanceof Array)) {
-              currentAttachment.tags.tag = [currentAttachment.tags.tag];
+          if (!captions_lang && potentialCaption.tags && potentialCaption.tags.tag) {
+            if (!(potentialCaption.tags.tag instanceof Array)) {
+              potentialCaption.tags.tag = [potentialCaption.tags.tag];
             }
-            currentAttachment.tags.tag.forEach((tag)=>{
+            potentialCaption.tags.tag.forEach((tag)=>{
               if (tag.startsWith('lang:')){
                 let split = tag.split(':');
                 captions_lang = split[1];
@@ -417,16 +412,33 @@ class OpencastToPaellaConverter {
           //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
 
           captions.push({
-            id: currentAttachment.id,
+            id: potentialCaption.id,
             lang: captions_lang,
             text: captions_label,
-            url: currentAttachment.url,
+            url: potentialCaption.url,
             format: captions_format
           });
         }
       }
       catch (err) {/**/}
     });
+  }
+
+  getCaptions(episode) {
+    var captions = [];
+
+    var attachments = episode.mediapackage.attachments.attachment;
+    var catalogs = episode.mediapackage.metadata.catalog;
+    var tracks = episode.mediapackage.media.track;
+    if (!(attachments instanceof Array)) { attachments = attachments ? [attachments] : []; }
+    if (!(catalogs instanceof Array)) { catalogs = catalogs ? [catalogs] : []; }
+    if (!(tracks instanceof Array)) { tracks = tracks ? [tracks] : []; }
+
+    // Read the attachments
+    this.readCaptions(attachments, captions);
+
+    // Read the tracks
+    this.readCaptions(tracks, captions);
 
     // Read the catalogs
     catalogs.forEach((currentCatalog) => {
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -390,7 +390,6 @@ class OpencastToPaellaConverter {
         let captions_match = captions_regex.exec(potentialCaption.type);
 
         if (captions_match) {
-          let captions_format = captions_match[1];
           let captions_lang = captions_match[3];
 
           // TODO: read the lang from the dfxp file
@@ -410,6 +409,10 @@ class OpencastToPaellaConverter {
 
           let captions_label = captions_lang || 'unknown language';
           //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
+
+          let captions_format = potentialCaption.url.
+            substring(potentialCaption.url.lastIndexOf('.') + 1, potentialCaption.url.length)
+            || potentialCaption.url;
 
           captions.push({
             id: potentialCaption.id,
--- a/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
+++ b/modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
@@ -411,7 +411,7 @@ class OpencastToPaellaConverter {
           captions.push({
             id: potentialCaption.id,
             lang: captions_lang,
-            text: captions_lang || 'unknown language';,
+            text: captions_lang || 'unknown language',
             url: potentialCaption.url,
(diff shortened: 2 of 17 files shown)


mliradelc:
Hi Arnei, have you tried with Paella 7? Because you modified Paella 6 to make it work

Arnei:
> Hi Arnei, have you tried with Paella 7? Because you modified Paella 6 to make it work

I have not, thanks for the notice.

Arnei:
Rebased this on r/12.x

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -293,41 +293,73 @@ function processAttachments(episode, manifest, config) {
...
-      const lang = exp[2] || '';
-      result.push({
-        id: att.id,
-        lang: lang,
-        text: lang || 'Unknown language',
-        format: format,
-        url:att.url
-      });
+        // TODO: read the lang from the dfxp file
+        //if (captions_format == "dfxp") {}
```
lkiesow: Do we still support `dfxp`? Is this a leftover from coding this?

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -293,41 +293,73 @@ function processAttachments(episode, manifest, config) {
...
+
+        if (!captions_lang && potentialCaption.tags && potentialCaption.tags.tag) {
+          if (!(potentialCaption.tags.tag instanceof Array)) {
+            potentialCaption.tags.tag = [potentialCaption.tags.tag];
+          }
+          potentialCaption.tags.tag.forEach((tag)=>{
+            if (tag.startsWith('lang:')){
+              let split = tag.split(':');
+              captions_lang = split[1];
```
lkiesow: Arguably simpler and only removes the prefix instead of splitting on every `:` character:
```suggestion
              captions_lang = tag.substring('lang:'.length);
```

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -293,41 +293,73 @@ function processAttachments(episode, manifest, config) {
...
+            }
+          });
+        }
+
+        let captions_label = captions_lang || 'unknown language';
+        //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
```
lkiesow: Do we still need the translation? If not, this can go?

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -293,41 +293,73 @@ function processAttachments(episode, manifest, config) {
...
+
+        let captions_format = potentialCaption.url.
+          substring(potentialCaption.url.lastIndexOf('.') + 1, potentialCaption.url.length)
+          || potentialCaption.url;
```
lkiesow: this or alternative below
```suggestion
        let captions_format = potentialCaption.url.
          substring(potentialCaption.url.lastIndexOf('.') + 1)
          || potentialCaption.url;
```

lkiesow: or even just
```suggestion
        let captions_format = potentialCaption.url.split('.').pop();
```

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -416,17 +410,38 @@ class OpencastToPaellaConverter {
           let captions_label = captions_lang || 'unknown language';
           //paella.utils.dictionary.translate("CAPTIONS_" + captions_lang);
 
+          let captions_format = potentialCaption.url.
+            substring(potentialCaption.url.lastIndexOf('.') + 1, potentialCaption.url.length)
+            || potentialCaption.url;
```
lkiesow: Is this simpler?
```suggestion
          let captions_format = potentialCaption.url.split('.').pop();
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -265,7 +259,12 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
                     index = srcIndex; // default src
                   }
 
-                  videoclips.add(new VideoClip(index, begin / 1000.0, end / 1000.0));
+                  // Sort out ref elements
+                  if (media.getMediaType() == SmilMediaElement.MediaType.REF) {
+                    refElements.add(new VideoClip(index, begin, end));
```
lkiesow, line 264: The variable `begin` and `end` are in ms:
```java
long begin = media.getClipBeginMS();
```
Is it correct to pass milliseconds to `VideoClip` here while two lines down you pass seconds to the constructor?

lkiesow: This is actually correct, since the constructor for `VideoClip(…)` is mad and treats `double` as milliseconds and `long as seconds. But it's probably still easier to understand here if we do at least the same here, using either only `long` or only `double`.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -279,19 +278,88 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
...
-
-      if (error != null) {
-        FileUtils.deleteQuietly(tempDirectory);
-        throw new ProcessFailedException("Editing pipeline exited abnormaly! Error: " + error);
+
+      // Don't mix video/audio and subtitles
+      if (videoclips.size() > 0 && refElements.size() > 0) {
+        throw new ProcessFailedException("Can not process media elements together with ref elements. "
+                + "There likely is an error in the SMIL file");
+      }
```
lkiesow, line 287: This is very likely just me being confused by SMIL: Are `refElements` in SMIL subtitles?

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -279,19 +278,88 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
...
+
+      // get output file extension
+      String outputFileExtension = null;
+      if (videoclips.size() > 0) {
+        outputFileExtension = properties.getProperty(VideoEditorProperties.DEFAULT_EXTENSION, ".mp4");
+      }
+      if (refElements.size() > 0) {
+        String extension = FilenameUtils.getExtension(sourceTrackUri);
+        if (VideoEditorProperties.WEBVTT_EXTENSION.equals(extension)) {
+          outputFileExtension = properties.getProperty(VideoEditorProperties.WEBVTT_EXTENSION, ".vtt");
```
lkiesow, line 297: If we load this from the configuration file, this should be added to the configuration file as an option.
Though, I'm not sure if we actually need this to be configurable.
Don't we always want `.vtt`?

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -279,19 +278,88 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
...
+          // Edit
+          List<WebVTTSubtitleCue> cutCues = new ArrayList<>();
+          for (VideoClip time: cleanclips) {
+            for (WebVTTSubtitleCue cue : subtitle.getCues()) {
+              if ((time.getStart() - SUBTITLE_GRACE_PERIOD) <= cue.getStartTime()
+                      && (time.getEnd() + SUBTITLE_GRACE_PERIOD) >= cue.getEndTime()) {
+                cue.setStartTime((long) (cue.getStartTime() - time.getStart()));
+                cue.setEndTime((long) (cue.getEndTime() - time.getStart()));
+                cutCues.add(cue);
+              }
```
lkiesow: If I understand this code block correctly, this will only let subtitle lines through which are completely within a video clip, basically fulfilling (with some grace period):

```
clip.start < subtitle.start < subtitle.end < clip.end
```

That way, we filter out all subtitle lines which are not within a clip. But is it intentional, that we then also filter out subtitles which are mostly in a clip? An alternative would be to shorten them?
Basically, this subtitle would be removed, wouldn't it?

```
   [----clip-----]
[--subtitle--]
```

Also, I don't think, the new start/end is set correctly if you have multiple clips.
Example:

```
   [----clip1-----]     [----clip2--------]
      [--sub1--]           [-sub2]  [-sub2]
1  2  3  4  5  6  7  8  9  10 11 12 13 14 15 ← time
```
if we calculate start based on this, we would get:
```
sub1.start = sub1.start - clip1.start = 3 - 2 = 1 (correct)
sub2.start = sub2.start - clip2.start = 10 - 9 = 1 (not correct)
```
Or am I missing something?

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/base/SubtitleCue.java
```diff
@@ -0,0 +1,100 @@
...
+    this.startTime = cue.getStartTime();
+    this.endTime = cue.getEndTime();
+    this.lines = cue.getLines();
+  }
+
+  protected SubtitleCue() {
+    this.lines = new ArrayList<>();
+  }
+
+  protected SubtitleCue(long startTime, long endTime) {
```
lkiesow, line 48: JavaDoc to know that times are in ms would be great

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/base/Subtitle.java
```diff
@@ -0,0 +1,54 @@
...
+    this.cues = new ArrayList<T>();
+  }
+
+  public void addHeaderLine(String headerLine) {
+    this.headerLines.add(headerLine);
+  }
+  public List<String> getHeaderLines() {
+    return headerLines;
+  }
+  public void setHeaderLines(List<String> headerLines) {
```
lkiesow:
```suggestion
  }
  
  public List<String> getHeaderLines() {
    return headerLines;
  }
  
  public void setHeaderLines(List<String> headerLines) {
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTParser.java
```diff
@@ -0,0 +1,189 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Parses WebVTT from a file into a datastructure to allow for easy modification.
+ * Throws exceptions if the read WebVTT is invalid.
+ *
+ * TODO: Comments are currently ignored and discarded. Find a good way to keep comments
+ *  without compromising easy editing.
```
lkiesow, line 39: I think that's okay. The editor also does not support comments and you can still store them in the original version you keep in the asset manager if you want to.

Looking at [comments in WebVTT](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API#webvtt_comments), you could probably have a list of subtitle elements which can either be of type `cue` or of type `comment`.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -279,19 +278,88 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
...
+      // Or give up
+      // TODO: It might be better if subtitle tracks were assigned the mediatype "texttrack" in the first place
+      if (refElements.size() > 0) {
+        List<VideoClip> cleanclips = sortSegments(refElements);    // remove very short cuts that will look bad
+        String extension = FilenameUtils.getExtension(sourceTrackUri);
+        if (VideoEditorProperties.WEBVTT_EXTENSION.equals(extension)) {
+          // Parse
+          WebVTTParser parser = new WebVTTParser();
+          FileInputStream fin = new FileInputStream(sourceFile);
+          WebVTTSubtitle subtitle = parser.parse(fin);
```
lkiesow: Close the input stream:
```suggestion
          WebVTTSubtitle subtitle;
          try (FileInputStream fin = new FileInputStream(sourceFile)) {
            subtitle = parser.parse(fin);
          }
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTParser.java
```diff
@@ -0,0 +1,189 @@
...
+
+  public WebVTTParser(Charset charset) {
+    this.charset = charset;
+  }
+
+  public WebVTTSubtitle parse(InputStream is) throws IOException, SubtitleParsingException {
+    return parse(is, true);
+  }
+
+  public WebVTTSubtitle parse(InputStream is, boolean strict) throws IOException, SubtitleParsingException {
```
lkiesow: The parameter `strict` isn't actually used anywhere, is it?

```suggestion
  public WebVTTSubtitle parse(InputStream is) throws IOException, SubtitleParsingException {
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTSubtitle.java
```diff
@@ -0,0 +1,62 @@
...
+    return this.cues;
+  }
+  public void setCues(List<WebVTTSubtitleCue> cues) {
+    this.cues = cues;
+  }
+
+  public void addRegion(WebVTTSubtitleRegion region) {
+    this.regions.add(region);
+  }
+  public List<WebVTTSubtitleRegion> getRegions() {
```
lkiesow:
```suggestion
  }
  
  public List<WebVTTSubtitleRegion> getRegions() {
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTSubtitle.java
```diff
@@ -0,0 +1,62 @@
...
+    this.style = new ArrayList<>();
+  }
+
+  public void addCue(WebVTTSubtitleCue cue) {
+    this.cues.add(cue);
+  }
+  public List<WebVTTSubtitleCue> getCues() {
+    return this.cues;
+  }
+  public void setCues(List<WebVTTSubtitleCue> cues) {
```
lkiesow:
```suggestion
  }
  
  public List<WebVTTSubtitleCue> getCues() {
    return this.cues;
  }
  
  public void setCues(List<WebVTTSubtitleCue> cues) {
```

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -279,19 +278,88 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
...
+                cue.setEndTime((long) (cue.getEndTime() - time.getStart()));
+                cutCues.add(cue);
+              }
+            }
+          }
+          subtitle.setCues(cutCues);
+
+          // Write
+          WebVTTWriter writer = new WebVTTWriter();
+          writer.write(subtitle, new FileOutputStream(outputPath));
```
lkiesow: Use `try-with-resources` to make sure to close the output stream?

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTWriter.java
```diff
@@ -0,0 +1,116 @@
...
+    } catch (UnsupportedEncodingException e) {
+      throw new IOException("Encoding error in input subtitle");
+    }
+  }
+
+  private String formatTimeCode(long ms) {
+    long milliseconds = (ms % 1000);
+    long seconds = (long)Math.floor((ms / 1000) % 60);
+    long minutes = (long)Math.floor((ms / (1000 * 60)) % 60);
+    long hours = (long)Math.floor((ms / (1000 * 60 * 60)));
```
lkiesow: `Math.floor(…)` does an implicit conversion to a `double` which is why you need to then cast it as a long. But since you only work with `long`, you are already doing an integer division, and you don't need that at all, do you?

```suggestion
    long milliseconds = ms % 1000L;
    long seconds = ms / 1000L % 60L;
    long minutes = ms / (1000L * 60L) % 60L;
    long hours = ms / (1000L * 60L * 60L;
```

lkiesow (changes requested):
I found a few issues in here, although most of my comments are minor things are don't necessarily need to be changed.

But I think cuts aren't necessarily correct if you have multiple clips. So far, that's just my assumption from reading the code. I haven't tested this yet.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: Yup, leftover from porting the changes from Paella6 to Paella7. Will remove.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: Inclined to agree, will change.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: I have no idea. You're probably right.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: That's a lot easier to read, thanks!

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: Agreed, will rewrite

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: Been a bit, but if I remember correctly: According to official SMIL definition no, `refs` are not subtitles, but just one of the synonyms for media elements (https://www.w3.org/TR/REC-smil/smil-extended-media-object.html#edef-ref). BUT in Opencast, refs are basically anything that are not video or audio, and the videoeditor fails if it finds one. That's also why we later check if the ref element links to a `.vtt` file, since that is the only kind of ref element we are equippped to process.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: We're not loading from a configuration file. VideoEditorProperties is an internal interface. Or are you referring to something different?

And yes, we always wan `.vtt`.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: > But is it intentional, that we then also filter out subtitles which are mostly in a clip? An alternative would be to shorten them?

Yes, this is intentional. Way back when we had a discussion about this and while I don't quite recall all the finer points, I think the takeaway were
- Subtitles should not give information to the user that the editor did not want the user to see/hear.
- It's far, far beyond the scope of this project to create an algorithm that can shorten a subtitle to fit the set cuts and get reasonable results that don't comprise the above point.

Therefore, it is safer to just drop the line entirely.

As for the calculations, I was pretty sure they made sense to me back then, but now I can't tell why you would be wrong. I'll look into it again.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTParser.java
Arnei: If you think it's okay I'll leave it as is.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/subtitle/webvtt/WebVTTWriter.java
Arnei: That does look substantially cleaner, thanks.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: Okay, so i tried to change this to make more similar by having `new VideoClip(index, begin / 1000.0, end / 1000.0));` both times, but that actually screws the calculations over?

I am also not quite sure yet how you arrived at the conclusion that the constructor treats `double` as `ms` and `long` as `s` respectively. The constructor merely casts long to double, doesn't it?

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: Looked into it again, you were right. I updated the calculations.

modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
Arnei: Update: Lars was looking at the constructors of a different `VideoClip` class. Also, the code is correct anyway, since the videoclips are later used for ffmpeg commands which expects seconds, while the subtite classes work with milliseconds. Added a comment to reflect that.

lkiesow (approved):
Looks hood and we can definitely add it like this although overall captions handling in Opencast is still pretty messy due to various different flavors being used. Bringing this in productions, the previous habit of always using `captions/vtt+{lng}` doesn't really work any longer since we need a way to distinguish between cut and uncut subtitles. Maybe we can use `captions/source` and `captions/delivery`. But that's a problem for another day.

mtneug:
> Looks hood and we can definitely add it like this although overall captions handling in Opencast is still pretty messy due to various different flavors being used. Bringing this in productions, the previous habit of always using `captions/vtt+{lng}` doesn't really work any longer since we need a way to distinguish between cut and uncut subtitles. Maybe we can use `captions/source` and `captions/delivery`. But that's a problem for another day.

As mentioned elsewhere, we intend to use `captions/source+{lng}` and `captions/delivery+{lng}`. The issue, of course, is that replacing only `source` with `delivery` and leaving the `+{lng}` untouched, is tricky with Opencast. Currently, we simply enumerate supported languages.

PR #3634, opened 2022-04-06
Update adopter registration code

gregorydlogan:
This pull request updates the adopter registration bits to:
 - Work properly
 - Log more things
 - Update the Terms of Service

Open questions:
 - Do we care about available disk space?  We would have to pull the data dir out of the config, then find the space from there since otherwise we get the node's root disk.  This data is likely not useful IMO.
 - Services are logged as a ",\n" delimited string.  I'm not sure how better to do this - creating a database entry for each service on each node seems excessive, but the string is a particularly dumb (and potentially problematic, due to field lengths) way of doing it.
 - This needs unit tests, although realistically integration tests would be useful too.

This goes hand-in-hand with https://github.com/opencast/adopter-registration-server/pull/17

---

To test this, set something like `org.opencastproject.adopter.registration.server.url=http://localhost:5000/` in `custom.properties`.

---

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
+++ b/modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
@@ -4,18 +4,11 @@
 </h1>
 
 <p style="padding-bottom: 15px">
-  Thank you for using Opencast. We really appreciate your registration.
-  You can change the data you provided at any time. To do so, go to the “Opencast Registration”
-  in the menu “Help” of the administrative user interface. If you deleted your Opencast
-  installation, want to remove your account but cannot anymore, please write an email to
-  support@opencast.org.
-</p>
-
-<p style="padding-bottom: 15px">
-  Although Opencast is free, open-source software it is important for us,
-  to know how Opencast is used. For instance, hardware vendors want to know how
-  large the market is before they make their devices Opencast compatible. Even public
-  funding for further development relies on such data.
+  Opencast is free, open-source software - so why do we bother you with registration and feedback?
+  Well, data on the usage of Opencast is important for us.  When it comes to discussions with hardware vendors about
+  cooperation and compatibility they want to know how large the market is before they make any decision. Likewise,
+  public funding for further development relies on such data.  Finally, we also need the feedback to further develop
+  Opencast in line with its usage.
 </p>
 
 <p style="padding-bottom: 15px">
@@ -67,7 +60,7 @@
 </p>
 
 <p style="padding-bottom: 15px">
-  If provided, we will publish the name of your institution on opencast.org and set
+  If any data is provided, we will publish the name of your institution on opencast.org and set
   a geographical marker for your institution on the Opencast adopter’s map.
 </p>
 
@@ -76,7 +69,7 @@
 </p>
 
 <p style="padding-bottom: 15px">
-  This data is anonymized and will be stored without any relation to your organizational
+  This data itself is anonymized and will be stored without any relation to your organizational
   or personal data unless you allow us to link this information. The data will be sent in
    a regular interval. The data sent may include the following information:
 </p>
@@ -110,12 +103,12 @@
 </p>
 
 <p style="padding-bottom: 15px">
-  Error reports can be automatically sent so that we can monitor if there
+  Additionally, error reports can be automatically sent so that we can monitor if there
   are systematic errors within Opencast that were not detected in our QA.
 </p>
 
 <p style="padding-bottom: 15px">
-  Note that this data will be linked to your organization so that we are
+  Note that this specific data will be linked to your organization so that we are
   able to warn you in case we found a critical behavior. The error reports
   may also contain information about your content. This data will not be public.
   It will only be accessible to Opencast committers.
@@ -131,6 +124,13 @@
   <li>A timestamp when the error occurred</li>
   <li>The Opencast version</li>
 </ul>
+
+<p style="padding-bottom: 10px">
+  Finally, you can change the data you provided at any time. To do so, go to the “Opencast
+  Registration” in the menu “Help” of the administrative user interface. If you deleted your
+  Opencast installation, want to remove your account but cannot anymore, please write an
+  email to support@opencast.org
+</p>
 
 
 
--- a/modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
+++ b/modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
@@ -23,7 +23,40 @@
   agree to sending data from all categories.
 </p>
 
+<h1 style="font-size: x-large;margin: 15px 0">
+  Data Stewardship
+</h1>
 
+<p style="padding-bottom: 10px; font-weight: bold">
+  Location
+</p>
+<p style="padding-bottom: 15px">
+  This data will be stored within the European Union, and will be GDPR compliant.
+  The current location can be found <a href="https://docs.opencast.org/develop/developer/#infrastructure/">here</a>
+</p>
+
+<p style="padding-bottom: 10px; font-weight: bold">
+  Storage Duration
+</p>
+<p style="padding-bottom: 5px;">
+  Organizational Profile
+</p>
+<p style="padding-bottom: 10px">
+  Your organizational profile will be purged 6 months after its last update.
+</p>
+<p style="padding-bottom: 5px">
+  Usage statistics and technical data
+<p style="padding-bottom: 15px">
+  Your technical statistics will be purged 24 months after their last update
+</p>
+
+<p style="padding-bottom: 10px; font-weight: bold">
+  Access to Information
+</p>
+<p style="padding-bottom: 15px">
+  Acess to the raw data is restricted to the Opencast board, who can authorize release to commiters on a case-by-case basis.
+  Aggregated summaries of the data will be made available from time to time at locations of the board's choosing.
+</p>
 
 <h1 style="font-size: x-large;margin: 15px 0">
   What Happens With Your Data
@@ -44,26 +77,29 @@
 
 <p style="padding-bottom: 15px">
   This data is anonymized and will be stored without any relation to your organizational
-  or personal data. The data will be sent in a regular interval. The data sent may include
-  the following information:
+  or personal data unless you allow us to link this information. The data will be sent in
+   a regular interval. The data sent may include the following information:
 </p>
 
 <ul style="padding-left: 30px; list-style: disc; padding-bottom: 15px; color: #333">
   <li>A timestamp at which this data has been created</li>
-  <li>An identifier marking the data source but is impossible
-    to link to your personal or organizational data
-  </li>
+  <li>An optional identifier marking the data source to link your organizational data to your install</li>
   <li>The number and status of events in your Opencast system</li>
   <li>The total duration of all media in your Opencast system</li>
   <li>The number of series in your Opencast system</li>
+  <li>The number of Opencast users (does not collect info from external user providers</li>
   <li>The number of capture agents connected to your Opencast system</li>
+  <li>The number of jobs in your Opencast system</li>
+  <li>The number of tenants in your Opencast system</li>
   <li>The version of your Opencast system</li>
   <li>Technical data regarding servers in your Opencast system
     <ul style="padding-left: 80px; list-style: circle">
       <li>A unique anonymous identifier for each server</li>
+      <li>The hostname for each server</li>
       <li>The number of CPU cores of each server</li>
       <li>The system memory of each server</li>
       <li>The disk space of each server</li>
+      <li>The maximum load value for each server </li>
       <li>The Opencast services running on each server</li>
     </ul>
   </li>
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
@@ -132,7 +132,7 @@ public class Form implements IForm {
 
   // If this is true, then the next stats send pass should delete the registration, then delete this object.
   @Column(name = "delete_me")
-  private transient boolean deleteMe = false;
+  private boolean deleteMe = false;
 
 
   //================================================================================
@@ -190,6 +190,7 @@ public class Form implements IForm {
       this.registered = f.registered;
     }
     this.dateModified = new Date();
+    this.deleteMe = f.deleteMe;
   }
 
   @Override
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
@@ -367,12 +367,4 @@ public class Form implements IForm {
     return Objects.requireNonNullElse(termsVersionAgreed, TERMSOFUSEVERSION.PRE_2022);
   }
 
-  public void setTermsVersionAgreed(TERMSOFUSEVERSION version) {
-    if (version == null) {
-      this.termsVersionAgreed = TERMSOFUSEVERSION.PRE_2022;
-    } else {
-      this.termsVersionAgreed = version;
-    }
-  }
-
 }
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
@@ -49,6 +49,14 @@ import javax.persistence.TemporalType;
 })
 public class Form implements IForm {
 
+  public enum TERMSOFUSEVERSION {
+    PRE_2022, APRIL_2022
+  };
+
+  public static final TERMSOFUSEVERSION getLatestTermsOfUse() {
+    return TERMSOFUSEVERSION.APRIL_2022;
+  }
+
 
   //================================================================================
   // Properties
@@ -114,6 +122,13 @@ public class Form implements IForm {
   @Column(name = "registered")
   private boolean registered;
 
+  // The default here is the original terms of use.
+  // Note that this object doesn't get instanciated unless you've agreed to the terms *at some point*
+  // so assuming a default of PRE_2022 is reasonable.  Either no agreement -> no db object at all, or
+  // agreement -> PRE_2022 default
+  @Column(name = "terms_version_agreed")
+  private TERMSOFUSEVERSION termsVersionAgreed = TERMSOFUSEVERSION.PRE_2022;
+
 
   //================================================================================
   // Constructor and Methods
@@ -141,6 +156,9 @@ public class Form implements IForm {
     this.allowsStatistics = allowsStatistics;
     this.allowsErrorReports = allowsErrorReports;
     this.agreedToPolicy = agreedToPolicy;
+    if (this.agreedToPolicy) {
+      this.termsVersionAgreed = getLatestTermsOfUse();
+    }
     this.registered = registered;
   }
 
@@ -160,6 +178,7 @@ public class Form implements IForm {
     this.allowsStatistics = f.allowsStatistics;
     this.allowsErrorReports = f.allowsErrorReports;
     this.agreedToPolicy = f.agreedToPolicy;
+    this.termsVersionAgreed = f.termsVersionAgreed;
     if (!this.registered) {
       // overwrite this field only when an adopter isn't registered yet
       // once an adopter is registered, he stays registered
@@ -341,6 +360,18 @@ public class Form implements IForm {
 
   public void setRegistered(boolean registered) {
     this.registered = registered;
+  }
+
+  public TERMSOFUSEVERSION getTermsVersionAgreed() {
+    return null == termsVersionAgreed ? termsVersionAgreed : TERMSOFUSEVERSION.PRE_2022;
+  }
+
+  public void setTermsVersionAgreed(TERMSOFUSEVERSION version) {
+    if (version == null) {
+      this.termsVersionAgreed = TERMSOFUSEVERSION.PRE_2022;
+    } else {
+      this.termsVersionAgreed = version;
+    }
   }
 
 }
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
@@ -130,6 +130,10 @@ public class Form implements IForm {
   @Column(name = "terms_version_agreed")
   private TERMSOFUSEVERSION termsVersionAgreed = TERMSOFUSEVERSION.PRE_2022;
 
+  // If this is true, then the next stats send pass should delete the registration, then delete this object.
+  @Column(name = "delete_me")
+  private transient boolean deleteMe = false;
+
 
   //================================================================================
   // Constructor and Methods
@@ -367,4 +371,11 @@ public class Form implements IForm {
     return Objects.requireNonNullElse(termsVersionAgreed, TERMSOFUSEVERSION.PRE_2022);
   }
 
+  public void delete() {
+    this.deleteMe = true;
+  }
+
+  public boolean shouldDelete() {
+    return this.deleteMe;
+  }
 }
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
@@ -25,6 +25,7 @@ import org.opencastproject.security.api.Organization;
 import org.opencastproject.util.EqualsUtil;
 
 import java.util.Date;
+import java.util.Objects;
 
 import javax.persistence.Access;
 import javax.persistence.AccessType;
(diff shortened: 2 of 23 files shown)


ebbertd:
While I can not review the code in this PR, I did have a look at the terms of use. After reading these I do have open questions which I think should be addressed in the terms of use.

- Where will the data be stored?
- For how long will the data be stored?
- Who has access to this data? Basically, who is "we" as stated in the terms of use?

modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
```diff
@@ -43,16 +43,14 @@ <h1 style="font-size: x-large;margin: 15px 0">
 </p>
 
 <p style="padding-bottom: 15px">
-  This data is anonymized and will be stored without any relation to your organizational
+  This data is anonymized and can be stored without any relation to your organizational
```
ebbertd: *can* is quite vague. Could you specify in which case the usage statistics and technical data will be linked with the organizational data?

gregorydlogan: Ah.  I meant that the data is anonymized, and can be stored w/o... *unless* you say we can link it internally.

What about something like `This data is anonymized by default, but will be linked to your organizational profile if you allow it.`?

gregorydlogan:
Good points.
 
> * Where will the data be stored?

No idea currently, but somewhere GDPR compliant I would bet.  I would guess ELAN/Osnabruck.  We should probably figure that out before addressing...

> * For how long will the data be stored?

Forever, unless removal is requested?  We could also say something like 2 years from time of last ping?

> * Who has access to this data? Basically, who is "we" as stated in the terms of use?

The Opencast committer body + board, or at least a subset of it.  I can't think of a reason why we'd be releasing it to anyone else, but we could add a clause about releasing anonymized summaries of technical information to third parties if that seems useful.

modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
ebbertd: That sounds reasonable. Does that mean there is a checkbox (or something like it) in which admins can choose to link their organizational profile with the usage statistics and technical data?

ebbertd:
> > * Where will the data be stored?
> 
> No idea currently, but somewhere GDPR compliant I would bet. I would guess ELAN/Osnabruck. We should probably figure that out before addressing...

Given that most adopters are in Europe I also think that storage of this data in Europe, and therefore GDPR compliant, makes sense. We should state the storage location and organization storing the data.

> > * For how long will the data be stored?
> 
> Forever, unless removal is requested? We could also say something like 2 years from time of last ping?

Do we actually need to store the raw data forever? We might need to differentiate between the usage statistics, technical data, organizational profile and personal profile containing contact information. As I understand it this data serves different purposes, therefore we have different storage duration requirements.

*Usage statistics and technical data*

From what I understand this data needs to be aggregated across institutions to be useful to us. Please correct me if I'm wrong. If that is the case, do we still need to store the raw data once this is done? If that is not the case we should delete it after a fixed time.

*Organizational profile*

This is useful for us to show how large the market is and for us to know where our adopters are located. Therefore, we need to store this data as long as this organization is transmitting it to us. Most likely we will also want to aggregate this information in some way, like summing up how many organizations transmitted this data in a given year. For this case I like your proposal of storing this information for 2 years after last ping.

*Personal profile containing contact information*

This data is mainly used to contact the admins about updates and security issues. So if the contact information changes I think we should immediately delete the old contact information.

As for the case that they stop transmitting data to us, it kinda seems long to me to save contact information for 2 years. What about 6 months?

Although, we might want to contact them for other purposes apart from updates and security issues. Again the case that they stopped transmitting data to use. We might want to contact them in this case to ask why they stopped using Opencast or why they only stopped the data transmission. Both cases would be a chance to get valuable feedback.

> 
> > * Who has access to this data? Basically, who is "we" as stated in the terms of use?
> 
> The Opencast committer body + board, or at least a subset of it. I can't think of a reason why we'd be releasing it to anyone else, but we could add a clause about releasing anonymized summaries of technical information to third parties if that seems useful.

That the committer body and board could have access to this data seems reasonable to me. Although, in practice it might only be a subset. Nevertheless, we should state this and link the list of current committers and board members as well as the processes defined for inclusion and exclusion in those groups.

gregorydlogan:
> > > * Where will the data be stored?
> > 
> > 
> > No idea currently, but somewhere GDPR compliant I would bet. I would guess ELAN/Osnabruck. We should probably figure that out before addressing...
> 
> Given that most adopters are in Europe I also think that storage of this data in Europe, and therefore GDPR compliant, makes sense. We should state the storage location and organization storing the data.

I would be against saying anything beyond that it would be stored in a GDPR compliant location.  We might have to move between institutions, if nothing else :)  We could include a link to our infra docs, which could contain the current storing organization.

> Do we actually need to store the raw data forever? We might need to differentiate between the usage statistics, technical data, organizational profile and personal profile containing contact information. As I understand it this data serves different purposes, therefore we have different storage duration requirements.
> _Usage statistics and technical data_
> 
> From what I understand this data needs to be aggregated across institutions to be useful to us. Please correct me if I'm wrong. If that is the case, do we still need to store the raw data once this is done? If that is not the case we should delete it after a fixed time.

I'm hesitant, especially considering we've never had this kind of fine-grained detail, to commit to aggregating it already.  We don't know what we need, and it's *not* going to be a burden to store it raw.  I'm fine with removing the relevant technical data when the org goes offline though.

> _Personal profile containing contact information_
> 
> This data is mainly used to contact the admins about updates and security issues. So if the contact information changes I think we should immediately delete the old contact information.

Absolutely.  We don't actually have a provision for storing the old data since the adopter registration key is what serves as the primary key in the backend.

> As for the case that they stop transmitting data to us, it kinda seems long to me to save contact information for 2 years. What about 6 months?
> 
> Although, we might want to contact them for other purposes apart from updates and security issues. Again the case that they stopped transmitting data to use. We might want to contact them in this case to ask why they stopped using Opencast or why they only stopped the data transmission. Both cases would be a chance to get valuable feedback.

All of this data updates (IIRC) daily, so going offline for 6 full months is pretty indicative of being truly gone.  I think removing contacts after 6 months, and technical after 24 months seems reasonable.

> > > * Who has access to this data? Basically, who is "we" as stated in the terms of use?
> > 
> > 
> > The Opencast committer body + board, or at least a subset of it. I can't think of a reason why we'd be releasing it to anyone else, but we could add a clause about releasing anonymized summaries of technical information to third parties if that seems useful.
> 
> That the committer body and board could have access to this data seems reasonable to me. Although, in practice it might only be a subset. Nevertheless, we should state this and link the list of current committers and board members as well as the processes defined for inclusion and exclusion in those groups.

In practice it's going to be you, me, Lars, and maybe Rudiger, but that's also not a reasonable thing to include :D

+1 for defining process for inclusion and exclusion, that seems reasonable.  Strong -1 to listing the current members.  If that changes, then we have to update these terms again.  IMO a link to somewhere in our documentation would be better - the terms say that you're agreeing that anyone on this list has access, and here's how that gets changed, and then we can change the list(s) without requiring a re-agreement to the terms.

ebbertd:
> > > > * Where will the data be stored?
> > > 
> > > 
> > > No idea currently, but somewhere GDPR compliant I would bet. I would guess ELAN/Osnabruck. We should probably figure that out before addressing...
> > 
> > 
> > Given that most adopters are in Europe I also think that storage of this data in Europe, and therefore GDPR compliant, makes sense. We should state the storage location and organization storing the data.
> 
> I would be against saying anything beyond that it would be stored in a GDPR compliant location. We might have to move between institutions, if nothing else :) We could include a link to our infra docs, which could contain the current storing organization.

Good idea, lets do that.

> 
> > Do we actually need to store the raw data forever? We might need to differentiate between the usage statistics, technical data, organizational profile and personal profile containing contact information. As I understand it this data serves different purposes, therefore we have different storage duration requirements.
> > _Usage statistics and technical data_
> > From what I understand this data needs to be aggregated across institutions to be useful to us. Please correct me if I'm wrong. If that is the case, do we still need to store the raw data once this is done? If that is not the case we should delete it after a fixed time.
> 
> I'm hesitant, especially considering we've never had this kind of fine-grained detail, to commit to aggregating it already. We don't know what we need, and it's _not_ going to be a burden to store it raw. I'm fine with removing the relevant technical data when the org goes offline though.

Okay, fine by me.

> 
> > _Personal profile containing contact information_
> > This data is mainly used to contact the admins about updates and security issues. So if the contact information changes I think we should immediately delete the old contact information.
> 
> Absolutely. We don't actually have a provision for storing the old data since the adopter registration key is what serves as the primary key in the backend.
> 
> > As for the case that they stop transmitting data to us, it kinda seems long to me to save contact information for 2 years. What about 6 months?
> > Although, we might want to contact them for other purposes apart from updates and security issues. Again the case that they stopped transmitting data to use. We might want to contact them in this case to ask why they stopped using Opencast or why they only stopped the data transmission. Both cases would be a chance to get valuable feedback.
> 
> All of this data updates (IIRC) daily, so going offline for 6 full months is pretty indicative of being truly gone. I think removing contacts after 6 months, and technical after 24 months seems reasonable.
> 
> > > > * Who has access to this data? Basically, who is "we" as stated in the terms of use?
> > > 
> > > 
> > > The Opencast committer body + board, or at least a subset of it. I can't think of a reason why we'd be releasing it to anyone else, but we could add a clause about releasing anonymized summaries of technical information to third parties if that seems useful.
> > 
> > 
> > That the committer body and board could have access to this data seems reasonable to me. Although, in practice it might only be a subset. Nevertheless, we should state this and link the list of current committers and board members as well as the processes defined for inclusion and exclusion in those groups.
> 
> In practice it's going to be you, me, Lars, and maybe Rudiger, but that's also not a reasonable thing to include :D
> 
> +1 for defining process for inclusion and exclusion, that seems reasonable. Strong -1 to listing the current members. If that changes, then we have to update these terms again. IMO a link to somewhere in our documentation would be better - the terms say that you're agreeing that anyone on this list has access, and here's how that gets changed, and then we can change the list(s) without requiring a re-agreement to the terms.

Sorry for the misunderstanding. But you got what I intended to say. :) Liking to the docs on committers and the board is a good idea. This way we can just have our normal processes while maintaining transparency.

gregorydlogan:
Per discussion with the board:
- Access to the raw data is restricted to the board, with release to committers on a case-by-base basis
- Aggregated summaries of the data will be made available from time to time at locations of the board's choosing

Greg will also be backporting this to OC 10/11 in a separate PR.

modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
gregorydlogan: Yes.  `Which data do you want to share? > Usage statistics`, which is false by default.  So you can register your org, without sharing your hardware data.

gregorydlogan:
@ebbertd @lkiesow I'm going to try and get the back port done for the 10.12 release. I'm going to need a relatively quick review of the changed ToU on Tuesday. Can I ping you for that, or are you not going to be available?

gregorydlogan:
This should be ready for review, at least with respect to the ToU changes.  I'll do one more pass through the data logic tomorrow before I backport this to 10.x.

modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
```diff
@@ -23,7 +23,40 @@
...
+  Usage statistics and technical data
+<p style="padding-bottom: 15px">
+  Your technical statistics will be purged 24 months after their last update
+</p>
 
+<p style="padding-bottom: 10px; font-weight: bold">
+  Access to Information
+</p>
+<p style="padding-bottom: 15px">
+  Acess to the raw data is restricted to the Opencast board, who can authorize release to commiters on a case-by-case basis.
```
ebbertd, line 50: We could include a link to the Governance document explaining how the board is formed and who is currently on the board. See: https://docs.opencast.org/develop/developer/#governance/#current-board

ebbertd:
> @ebbertd @lkiesow I'm going to try and get the back port done for the 10.12 release. I'm going to need a relatively quick review of the changed ToU on Tuesday. Can I ping you for that, or are you not going to be available?

Feel free to ping me.  The ToU are fine by me.

gregorydlogan:
Converting this to a draft because I'm going to force-push this into r/10.x now that it hasn't made 10.12.  Will un-draft once we're ready.

modules/admin-ui-frontend/app/scripts/modules/registration/partials/terms-of-use.html
gregorydlogan: IMO this is out of scope for this PR. If someone wants to know about how the sausage is made, they'll find it.

gregorydlogan:
Completely redid the commit structure, but now it should work, and with 10.x.  Once this goes in I'll file a follow on PR with the OSGI bits swapped to the new style.

lkiesow:
I found a major problem with the new version. It seems like the explanation view is missing:

https://user-images.githubusercontent.com/1008395/167475475-61be8c7e-1091-48ee-9fab-3292ed3aa2ec.mp4

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
```diff
@@ -343,4 +362,16 @@ public void setRegistered(boolean registered) {
     this.registered = registered;
   }
 
+  public TERMSOFUSEVERSION getTermsVersionAgreed() {
+    return null == termsVersionAgreed ? termsVersionAgreed : TERMSOFUSEVERSION.PRE_2022;
```
lkiesow:
```suggestion
    eturn Objects.requireNonNullElse(termsVersionAgreed, TERMSOFUSEVERSION.PRE_2022);
```

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
```diff
@@ -343,4 +362,16 @@ public void setRegistered(boolean registered) {
...
+  }
+
+  public void setTermsVersionAgreed(TERMSOFUSEVERSION version) {
+    if (version == null) {
+      this.termsVersionAgreed = TERMSOFUSEVERSION.PRE_2022;
+    } else {
+      this.termsVersionAgreed = version;
+    }
+  }
```
lkiesow: This doesn't seem to be used at all:
```suggestion
```

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -100,6 +121,9 @@ public class ScheduledDataCollector extends TimerTask {
   /** The Opencast version this is running in */
   private String version;
 
+  /** The timer for shutdown uses */
+  private Timer t;
```
lkiesow: Maybe more than a single character?
```suggestion
  private Timer timer;
```

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -146,7 +175,13 @@ public void run() {
...
     }
 
-    if (adopter.isRegistered() && adopter.agreedToPolicy()) {
+    // Don't send data unless they've agreed to the latest (at time of writing) terms.
+    // Pre April 2022 doesn't allow collection of a bunch of things, and doens't allow linking stat data to org
+    // so rather than burning time turning various things off (after figuring out what needs to be turned off)
+    // we just don't send anything.  By the time we need to update the ToU again this whole thing would need reworking
+    // anyway, so we'll run with this for now.
+    if (adopter.isRegistered() && adopter.agreedToPolicy()
+        && adopter.getTermsVersionAgreed() == Form.TERMSOFUSEVERSION.APRIL_2022) {
```
lkiesow: Isn't `adopter.agreedToPolicy()` implied in `adopter.getTermsVersionAgreed() == Form.TERMSOFUSEVERSION.APRIL_2022`?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -186,9 +221,22 @@ private String collectGeneralData(Form adopterRegistrationForm) {
...
-    serviceRegistry.getHostRegistrations().forEach(host -> statisticData.addHost(new Host(host)));
+    statisticData.setAdopterKey(adopterKey);
+    serviceRegistry.getHostRegistrations().forEach(host -> {
+      Host h = new Host(host);
+      try {
+        String services = serviceRegistry.getServiceRegistrationsByHost(host.getBaseUrl())
+            .stream()
+            .map(sr -> sr.getServiceType())
+            .collect(Collectors.joining(",\n"));
+        h.setServices(services);
```
lkiesow, line 254: Why add a comma separated list of the services and not the host?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -199,6 +247,42 @@ private String collectStatisticData(String statisticKey) throws Exception {
...
+        long current = statisticData.getCACount();
+        int orgCAs = caStateService.getKnownAgents(systemAdminUser, org).size();
+        statisticData.setCACount(current + orgCAs);
+
+        //Calculate the total number of minutes for this org, add it to the total
+        current = statisticData.getTotalMinutes();
+        long orgDuration = 0L;
+        try {
+          SearchResult sr = searchService.getForAdministrativeRead(sq);
+          orgDuration = Arrays.stream(sr.getItems())
```
lkiesow: First, this doesn't work iirc since the search service automatically applies a limit to queries by default. Though maybe the `getForAdministrativeRead` bypasses that.

If it does get all, this will load hundreds of thousands of Java objects into memory. I don't think that's a good idea. This may cause harm to the node.

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentStateServiceImpl.java
```diff
@@ -373,10 +373,21 @@ public void removeAgent(String agentName) throws NotFoundException {
...
   public Map<String, Agent> getKnownAgents() {
+    return getKnownAgents(null, null);
+  }
+
+  public Map<String, Agent> getKnownAgents(User u, Organization o) {
     agentCache.cleanUp();
     EntityManager em = null;
-    User user = securityService.getUser();
-    Organization org = securityService.getOrganization();
+    User user = u;
```
lkiesow: Maybe use `Objects.requireNonNullElse(…`?

modules/capture-admin-service-api/src/main/java/org/opencastproject/capture/admin/api/CaptureAgentStateService.java
```diff
@@ -93,6 +95,13 @@ public interface CaptureAgentStateService {
    */
   Map<String, Agent> getKnownAgents();
 
+  /**
+   * Returns the list of known agents that a given user and organization is authorized to schedule
+   *
+   * @return A {@link java.util.Map} of name-agent pairs.
+   */
+  Map<String, Agent> getKnownAgents(User u, Organization o);
```
lkiesow: Why this new method if you run this in `runAs` anyway:

```java
SecurityUtil.runAs(securityService, org, systemAdminUser, () -> {
  …
}
```

lkiesow (changes requested):

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
gregorydlogan: A setter would be required by JPA, right?  The extra logic ensures that we get a non-null in the case of an old database being loaded by new code.  This *may* not be required if the default terms version in `termsVersionAgreed` gets set when the new column is created, but I had already written this part when I figured out how to get JPA to add columns.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: No.  Agreed to policy is a boolean saying that they have agreed to *some version* of the policy, but not necessarily the latest.

Edit: But you're right, in that the *logic* here does imply that.  We're not sending stuff unless the latest policy is agreed to...  I'll drop the extra clause.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: This is sent on a per-host basis.  If you've got three hosts, you get three separate lists of services, one for each node.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: Ah, good point.  Luckily, the first page of results appears to contain a correct `total` value.

Edit: Wrong kind of total...

modules/capture-admin-service-api/src/main/java/org/opencastproject/capture/admin/api/CaptureAgentStateService.java
gregorydlogan: Good point, unnecessary changes.  Will revert.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: Per discussion in chat, this is disabled completely for now and will make a reappearance later.

gregorydlogan:
I forgot to fix the modal's state. Tomorrow. 

Edit: This is now done.

lkiesow:
Now contacts the server, but here is how to break it:

Update registration, checking:

- [x] Usage statistics
- [x] Error reports
- [x] I have read and agree to the privacy policy and terms of use

Hit save. Reload the page.
Open up the registration again.
Uncheck both data checkboxes:

- [ ] Usage statistics
- [ ] Error reports
- [x] I have read and agree to the privacy policy and terms of use

Hit save. Reload the page.
Open up the registration again.
You are stuck in the „Thank you“ modal with no buttons

lkiesow:
Here is a video to follow along when I break this:

https://user-images.githubusercontent.com/1008395/167723316-f30be62a-a24b-4a85-b579-e6b90a75ee22.mp4

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/Form.java
gregorydlogan: This isn't needed according to further testing, so I've removed it.

gregorydlogan:
I can't reproduce the behaviour you're seeing anymore, although none of these fixes should have resolved it.  As much as I hate saying it, have you cleared your cache?  I did see some very wonky behaviour while I was fixing [d91b7fe](https://github.com/opencast/opencast/pull/3634/commits/d91b7fece984e27e00695d05c96104aa47a665d5), and switching to private browsing (and thus a clean cache) resolved the issue.

lkiesow:
I did but the previous errors could easily be caused by the settings being set in a weird way. Let me test the new state…

lkiesow:
Looks far better and almost works now…
What does not work is deleting your registration:

- Start Opencast, register
- Restart Opencast, so the data is sent immediately
- Delete your registration via admin interface
  - The data is not deleted on the registration server
- Restart Opencast to trigger scheduled data sending
  - No change

I don't think we are allowed to keep user data if a user specifically chose to delete it.
I did not read through the TOS (maybe something about this is in there?) but even with that I find it unexpected and think we shouldn't keet the data.

gregorydlogan:
Huh, forgot to update that.  I don't think we *can* keep the data if they choose to delete it, can we?  I would interpret the user clicking "Delete" as a deletion request, and (to maintain GDPR compliance) wouldn't we be obligated to remove our copy?

Even if we aren't, keeping it would be rude.  Will fix shortly.

gregorydlogan:
I didn't quite have time to test that the deletion actually works, but it *looks* like it should @lkiesow.

If you get around to testing this and can get it into 11.7 great, if not it can wait until 10.14/11.8/12.0.

gregorydlogan:
@lkiesow this is now ready for review (again).

lkiesow:
This now looks good. I still want to suggest two minor improvements to the delete dialog:

- Can we turn “Are you sure you want to delete your registration data?” into “Are you sure you want to delete your registration data? Confirming this will also remove your registration data from the Opencast project server with the next update interval.”
- Can we turn the green “Confirm” into a red “Irrevocably Delete”? In particular, switching from green to red for this action would be good.

I don't think another review of this is required. Feel free to merge this with or without the minor change.

PR #3635, opened 2022-04-06
Fix bug with adding a series when a deleted one still lingers

LukasKalbertodt:
`this.deleteSeries` only marks the series as deleted again, not actually
removing it. This will clash if a series with the same ID is added again
as it will cause a unique constraint violation. Instead, we completely
remove it here. Unfortunately, without the `flush` it didn't work: the
unique violation was still triggered. Whether `flush` is the best
solution: 🤷, I hardly know stuff about JPA.

This was noticed by Katrin: https://github.com/opencast/opencast/pull/3076#discussion_r839548962

CC @KatrinIhler 

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
@@ -321,7 +321,8 @@ public class SeriesServiceDatabaseImpl implements SeriesServiceDatabase {
         // If the series existed but is marked deleted, we completely delete it
         // here to make sure no remains of the old series linger.
         if (entity != null) {
-          this.deleteSeries(seriesId);
+          em.remove(entity);
+          em.flush();
         }
 
         // no series stored, create new entity


modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
```diff
@@ -321,7 +321,8 @@ public DublinCoreCatalog storeSeries(DublinCoreCatalog dc)
         // If the series existed but is marked deleted, we completely delete it
         // here to make sure no remains of the old series linger.
         if (entity != null) {
-          this.deleteSeries(seriesId);
+          em.remove(entity);
+          em.flush();
```
mtneug, line 325: Do we need a flush? This is all happening in a transaction anyway.

mtneug (commented):
I still need to test this but this looks reasonable.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/persistence/SeriesServiceDatabaseImpl.java
LukasKalbertodt: As the commit message says: "Unfortunately, without the `flush` it didn't work: the unique violation was still triggered."
Which really confuses me. Changes by one statement in a transaction should be seen by statements that are executed afterwards. But somehow, this is not the case here? But then again, this is all JPA and I don't know what extra logic it adds. Just calling `flush` was the easiest solution I found. If you have a better one or can explain why it behaves like it does, I'd be very grateful ^_^

tyfranck: I have tested this PR and it works fine. My attempt at explaining the need for `flush`  .... 
I suspect the reason why `flush` is necessary is partly due to the fact that JPA for performance reason does not  always execute DB instructions such as `INSERT` or `DELETE` until the transaction is committed.
I think in this case, we have `em.remove()` but before that is sent to the DB (e.g transaction committed), the same id (primary key) is used to define another Entity. So JPA probably decides that it doesn't need to send a `DELETE` instruction anymore maybe for performance reason or who knows? 
`Flush` obviously will bypass all of that and run DB DELETE. Alternative option to `flush` could be to create a different transaction for the delete bit but I don't think it is necessary here.

lkiesow:
Thanks for the review, @tyfranck → merging

PR #3661, opened 2022-04-11
Check if user can be loaded before starting the workflow

darolfes:
Port to the opencast community edition.
This PR simply adds a small check before starting a workflow that ensures the workflow user can be loaded.
Seems like that the corresponding bug has already been fixed, but since this is just an additional check, there seems to be no real downside to adding this.

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -585,11 +585,7 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
 
       // Get the current user
       User currentUser = securityService.getUser();
-      if (currentUser == null)
-        throw new SecurityException("Current user is unknown");
-
-      if (userDirectoryService.loadUser(currentUser.getUsername()) == null)
-        throw new SecurityException(String.format("Current user '%s' can not be loaded", currentUser.getUsername()));
+      validUserOrThrow(currentUser);
 
       // Get the current organization
       Organization organization = securityService.getOrganization();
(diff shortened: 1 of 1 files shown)


metzenseifner (approved):
I'd approve this. The lines before it concerns me a bit. Perhaps it would be a good idea to abstract the two checks into a validation function. I'll let pass on the baton to one of the committers.

SecurityService.getUser() may fail or return a User object that is possibly unknown to the UserDirectoryService.

```
      // Get the current user
      User currentUser = securityService.getUser();
      if (currentUser == null)
        throw new SecurityException("Current user is unknown");

...
// this commit code follows
```


I imagine something along the lines of

```
User currentUser = securityService.getUser();
validOrThrow(currentUser); //throws security exception with appropriate string inside
```

gregorydlogan:
I'm not opposed to merging this, but I'm struggling to understand how the failure case here could ever happen.  Do you happen to know what the original cause of this check being added was?  How can a user, who was just fetched from the security service, not be loadable by the user directory?

metzenseifner:
I know that it is possible to delete a User account (removing from the User Service,I suppose) and restart an old workflow that was created by a now non-existant user account (it was likely an old workflow stuck in a failed state that had been restarted). I had this case in Innsbruck. I am not sure how it happens at a technical level though.

metzenseifner:
Also I'd like to know what the ticket owner means with

> Seems like that the corresponding bug has already been fixed

KatrinIhler:
> Also I'd like to know what the ticket owner means with
> 
> > Seems like that the corresponding bug has already been fixed

We ran into this bug at SWITCH because anonymous users couldn't run workflows, since they can't be loaded at workflow start (because they aren't persisted). Anonymous users can now run workflows, but it's possible that this bug applies to other scenarios, as you described above. In that case, I'd rather have the workflow not be started rather than have a zombie workflow in the system that can't ever run. :)

KatrinIhler:
@gregorydlogan See my comment above - anonymous users for example are available from the security service, but they aren't persisted in the database and thus can't be loaded from the user service. The same is true for a user that's no longer available.

darolfes:
Updated: Moved the checks into a validation function.

PR #3670, opened 2022-04-13
Added track upload endpoint the External Events Api

Arnei:
Adds a new endpoint to the External API Events endpoints. It allows uploading a track to an event by sending the updated media package to the archive. It also allows removing all other tracks of the specified flavor. It does not start a workflow.

The motivation for this feature is to offer an external endpoint that allows uploading subtitles as tracks to an event after event creation and without access to the Admin UI. However, the endpoint is currently not restricted to subtitle files or the "captions/*" flavor, mainly because it was easier to code while also giving the endpoint additional functionality beyond its originally intended use case.

The endpoint does currently not place a restriction on which file types can be uploaded. Maybe it should?

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


modules/external-api/pom.xml
```diff
@@ -194,6 +203,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-asset-manager-api</artifactId>
+      <version>12-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
geichelberger: The dependency is already declared at line 175.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1963,4 +1977,93 @@ private Optional<Response> updateSchedulingInformation(
...
+        if (item.isFormField()) {
+          if ("flavor".equals(fieldName)) {
+            String flavorString = Streams.asString(item.openStream());
+            try {
+              newFlavor = MediaPackageElementFlavor.parseFlavor(flavorString);
+            } catch (IllegalArgumentException e) {
+              return RestUtil.R.badRequest(String.format("Could not parse flavor %s; %s", flavorString, e.getMessage()));
+            }
+          } else if ("overwriteExisting".equals(fieldName)) {
+            String hhh = Streams.asString(item.openStream());
```
geichelberger: Does hhh have a meaning?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1963,4 +1977,92 @@ private Optional<Response> updateSchedulingInformation(
...
+        Track[] existing = mp.getTracks(newFlavor);
+        for (int i = 0; i < existing.length; i++) {
+          mp.remove(existing[i]);
+          logger.debug("Overwriting existing asset {} {}", tmpFlavor, newFlavor);
+        }
+      }
+      // correct the flavor of the new attachment
+      Track[] elArray = mp.getTracks(tmpFlavor);
+      elArray[0].setFlavor(newFlavor);
+      logger.debug("Updated asset {} {}", tmpFlavor, newFlavor);
```
geichelberger: It's possible to have multiple track parameters that would only correct the flavor for the first entry.

image

lkiesow:
@Arnei, is this ready for another review?

Arnei:
Yes, I hope I addressed all change requests by geichelberger sufficiently.

geichelberger:
This endpoint creates a new snapshot of the media packages, which causes troubles if there is currently a workflow running. The other event endpoints altering the media package are changing the workflow version of the media package if a workflow is running.

 e.g.
https://github.com/opencast/opencast/blob/08eab7331bcab7fca4db1f5a812071b852332ffd/modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java#L1319-L1365

lkiesow:
Adding a track to a running workflow can have very weird side effects. You also don't even know if you end up with the new track in the resulting snapshot, if the workflow even creates one. I would much prefer the API to answer with something like [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) (This response is sent when a request conflicts with the current state of the server).

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
```diff
@@ -93,7 +93,7 @@
 @Path("/")
 @Produces({ ApiMediaType.JSON, ApiMediaType.VERSION_1_0_0, ApiMediaType.VERSION_1_1_0, ApiMediaType.VERSION_1_2_0,
             ApiMediaType.VERSION_1_3_0, ApiMediaType.VERSION_1_4_0, ApiMediaType.VERSION_1_5_0,
-            ApiMediaType.VERSION_1_6_0, ApiMediaType.VERSION_1_7_0 })
+            ApiMediaType.VERSION_1_6_0, ApiMediaType.VERSION_1_7_0, ApiMediaType.VERSION_1_8_0 })
```
geichelberger: I think the API version needs to be changed.

geichelberger (changes requested):

modules/external-api/src/main/java/org/opencastproject/external/endpoint/GroupsEndpoint.java
Arnei: Just to be sure, the issue here is that the API version should already be 1_8_0 and needs to be bumped to 1_9_0 everywhere, right?

geichelberger: Yes, I think so. Else it would add a feature to a released API version. @lkiesow, correct me if I am wrong.

gregorydlogan:
Is there still a plan to push forward with this?

Arnei:
It kind of fell of my radar. I'll put it back on there ^^'

modules/external-api/src/test/resources/event-track-update-mediapackage.xml
```diff
@@ -0,0 +1,3 @@
+<ns2:mediapackage xmlns:ns2="http://mediapackage.opencastproject.org" id="trackupdateevent"
+                  duration="-1">
+</ns2:mediapackage>
```
geichelberger: No newline at end of file.

modules/external-api/src/test/resources/event-track-update-mediapackage-updated.xml
```diff
@@ -0,0 +1,11 @@
...
+                  duration="-1">
+  <ns2:media>
+    <ns2:track type="addTrack/temporary" id="6b574c7b-bf85-4023-bc41-c3145fa189d6">
+      <ns2:tags />
+      <ns2:url>/av.mov
+      </ns2:url>
+      <ns2:duration>-1</ns2:duration>
+    </ns2:track>
+  </ns2:media>
+</ns2:mediapackage>
```
geichelberger: No newline at end of file.

geichelberger (changes requested):
Two files are missing a newline at the end of the file.

geichelberger (approved):
LGTM

Issue #3509, opened 2022-03-02
User gets removed from groups after Shibboleth login (since OC11)


jduehring:
I am working with a customer who uses Shibboleth to Login into the Opencast Admin-UI and who has described the following behaviour: 

1. Enable the bootstrap user and use it to add some user xy to some group, for example the `Default System Administrator` group.
2. Disable bootstrap user and check whether user xy has entered the group and has gotten the new admin roles.
Now...
2.1 If the user xy was already logged in and one just refreshed the admin-ui, one could see that everything seemed to have worked.
2.2 But if the user needed to login, he got removed from the group and the respective roles.
3. Same happens after user xy logs out and in again after 2.1

-> So basically every time a user is logging in, he gets removed from all assigned groups. 

Since this behaviour came up after upgrading to OC 11, I checked and found [this](https://github.com/opencast/opencast/pull/3157) merge/commit, which we confirmed led to the new behaviour. 

Then I started looking into the code and got a bit confused.
This method:
https://github.com/opencast/opencast/blob/9035e7bbfe1ca294dbad783388d41eeb174fd0b0/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaUserReferenceProvider.java#L376-L394
is called each time a user is logging in. The aim of this method seems to be to check the assigned roles of a user and with the next call compare them with the groups the user is part of. If a role does not match with the primary group role, the user get removed from the respective group. 

But after some debugging it seems like the call `user.getRoles()` only gets the actual roles of the user, not the primary role from a group (the one typically starting with `ROLE_GROUP`). My assumtion is that this leads to an empty `internalGroupRoles` / `rolelist` and therefore the user gets removed from the group in each case. 

Before the mentioned commit, this check didn't even take place, since it was restricted to groups with a specific prefix.

It would be great, if someone can reproduce this behaviour to confirm it has nothing to do with any false configurations or just give some more insight. It is hard to test any changes for this module in my local enviroment. 

**Environment Information**
 - OS: [Rocky Linux 8.5]
 - Opencast version [11.3]

lkiesow:
Initial thoughts after reading this:

```java
(Role.Type.GROUP.equals(role.getType()) || (Role.Type.INTERNAL.equals(role.getType())
```

Do the roles exist at all on the users? If so, could they have a different type?

> it seems like the call user.getRoles() only gets the actual roles of the user, not the primary role from a group (the one typically starting with `ROLE_GROUP`).

[...]

PR #3672, opened 2022-04-13
Disable group removement upon missing group roles

jduehring:
This PR addresses #3509 . When logging in via Shibboleth there is currently a check which matches all the primary group roles of a user with its corresponding group memberships. If a primary group role is not present, but the user is part of the corresponding group, the user gets removed from the group. 

The problem is that when adding a new user to a group, the user will get saved in the database, but won't get the primary group role and therefore will be removed from the group after the next login. 

I do not really see a reason for making it a hard constraint that the primary group role must be present in the users role list in order to be part of that group. It should be sufficient to rely on the `oc_group_member` database table to track all group memberships. 
For that reason I removed the part in the code which was responsible for the removement of group memberships when checking the primary group roles.

I would need someone with a Shibboleth system to test, if this fixes the problem, since I do not have a suitable local setup to test this. 

I would also appreciate feedback if the removal of group memberships served any special purpose which I might have missed.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
+++ b/modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
@@ -306,6 +306,25 @@ public class JpaGroupRoleProvider implements AAIRoleProvider, GroupProvider, Gro
   }
 
   /**
+   * Removes a user from all groups
+   *
+   * @param userName
+   *          the username
+   * @param orgId
+   *          the user's organization
+   *
+   */
+  public void removeMemberFromAllGroups(String userName, String orgId) throws UnauthorizedException {
+    // List of the user's groups
+    List<JpaGroup> membership = UserDirectoryPersistenceUtil.findGroupsByUser(userName, orgId, emf);
+    for (JpaGroup group : membership) {
+      logger.debug("Removing user {} from group {}", userName, group.getRole());
+      group.getMembers().remove(userName);
+      addGroup(group);
+    }
+  }
+
+  /**
    * Loads a group from persistence
    *
    * @param groupId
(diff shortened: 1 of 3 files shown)


majosch:
I've just deployed this PR (on Opencast 11.6) to our test installation (Rocky 8.5, shibboleth 3.3.0) and can confirm that users are not removed from groups after login/logout anymore.

Martin Schamberger
(University of Vienna)

modules/userdirectory/src/main/java/org/opencastproject/userdirectory/JpaGroupRoleProvider.java
```diff
@@ -280,18 +279,9 @@ public void updateGroupMembershipFromRoles(String userName, String orgId, List<S
...
-          membershipRoles.add(group.getRole());
-        } else {
-          // remove user from this group
-          logger.debug("Removing user {} from group {}", userName, group.getRole());
-          group.getMembers().remove(userName);
-          addGroup(group);
-        }
-      } catch (UnauthorizedException e) {
-        logger.warn("Unauthorized to add or remove user {} from group {}", userName, group.getRole(), e);
+      if (roleList.contains(group.getRole())) {
```
gregorydlogan, line 282: This *seems* reasonable, but breaks things in at least one subtle way:

- `JpaUserAndRoleProvider`:496 uses `updateGroupMembershipFromRoles` to set the group membership to an empty set before deleting the user.  Removing the removal here will leave dangling references.

gregorydlogan (changes requested):
IMO this behaviour is important enough that this change should be accompanied by a unit test as well.  The fact that you can make this change *without changing existing testing* is worrying enough...

gregorydlogan:
@majosch @jduehring can you confirm that there aren't dangling references left behind?  This should make the next 11.x release.

jduehring:
@gregorydlogan I put the removal in a separate method, including some unit testing and checked for other usages of `updateGroupMembershipFromRoles`. There should be no dangling references left.

gregorydlogan (approved):
LGTM

Issue #3387, opened 2022-01-27
URLs returned by /search signed twice with token authentication


jojoob:
**Describe the bug**
I configured "Stream Security"/token based authentication on an Opencast 11 distributed setup on Docker using the [minimal configuration example](https://docs.opencast.org/r/11.x/admin/#configuration/stream-security/#minimal-configuration-example) from the docs. Everything works when doing the manual [testing](https://docs.opencast.org/r/11.x/admin/#configuration/stream-security/#testing) steps. But engage UI does not work: thumbs and videos are not loading because the URLs delivered by the "/search/" API are signed twice. URLs delivered by the external API (`curl 'https://opencastadmin.htw-berlin.de/api/events?sign=true&withpublications=true'`) are signed correct.

![ksnip_20220127-112108](https://user-images.githubusercontent.com/4180675/151339730-a3bc507f-e946-4315-9854-40a652778867.png)

`curl https://opencast.htw-berlin.de/search/episode.json`:

```json
[...]
                            {
                            "id": "4b3132c7-ba44-4700-bf3c-aebbbe7517bb",
                            "type": "presentation/player+preview",
                            "ref": "track:track-5",
                            "mimetype": "image/jpeg",
                            "tags":
                            {
                                "tag": "engage-download"
                            },
[...]

PR #3674, opened 2022-04-15
Avoid rewriteUrls being invoked twice while loading mediapackage.

Gao-Jun:
This PR is related to #3058 , which add `rewriteUrls` to `loadFromXml(InputStream is)`, and with a `loadFromXml(Node xml)` as return value.

Yet, another `rewriteUrls` is also exist in `loadFromXml(Node xml)`. So, while invoking  `loadFromXml(InputStream is)`, `rewriteUrls` will be called twice. That means, URLs in mediapackage will be rewrited twice, and cause an unexpected result.

For example, I use `PresignedUrlMediaPackageSerializer` to generate s3 presigned URL, and the current logic will return an unexpected presigned presigned URL.

Solution: remove the invoke of `rewriteUrls` in `loadFromXml(InputStream is)`.
All 3 `loadFromXml` method will eventually invoke `rewriteUrls` in  `loadFromXml(Node xml)` only once.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
+++ b/modules/common/src/main/java/org/opencastproject/mediapackage/MediaPackageBuilderImpl.java
@@ -104,10 +104,6 @@ public class MediaPackageBuilderImpl implements MediaPackageBuilder {
   public MediaPackage loadFromXml(InputStream is) throws MediaPackageException {
     try {
       Document xml = XmlSafeParser.parse(is);
-      if (serializer != null) {
-        //Convert InputStream to XML document to rewrite the URLs
-        rewriteUrls(xml, serializer);
-      }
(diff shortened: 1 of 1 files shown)


mliradelc:
In our cluster, the signing only makes one string. But I'm starting to suspect that there are other components using this code (The S3 module for example) and the bug #3387 appears on those instances.

I will try to run this PR in our cluster to see if I overlooked something

jojoob:
I cherry-picked the patch onto the code of version 11.6 I can confirm it fixes #3387. But since I just tested for this specific issue I can not make a statement if it introduces any other problems.

mliradelc (approved):
Tested in our cluster, still Sign the queries from the search service.

PR #3679, opened 2022-04-19
Don't store invalid ACLs submitted to api-endpoint

darolfes:
Port of an old bugfix to the community edition.
From the original commit:
```
This is a workaround for the following problem: When the user navigates
through series in the series details modal, the admin ui attempts to
store an invalid intermediate state of the series ACLs.
```

The mentioned bug in the admin ui has seemingly been fixed, but it is still possible to submit ACLs to the api that have an empty role or action.
The commit fixes this, but I am unsure if this is still wanted/needed as the original problem no longer exists.
Feel free to close this PR if you feel that this change is unnecessary.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
@@ -968,7 +968,7 @@ public class SeriesEndpoint {
     }
 
     if (!accessControlList.isValid()) {
-      logger.error("POST api/series/{}/access: Invalid series ACL detected", seriesId);
+      logger.debug("POST api/series/{}/access: Invalid series ACL detected", seriesId);
       return badRequest();
     }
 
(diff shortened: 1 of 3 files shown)


modules/common/src/main/java/org/opencastproject/security/api/AccessControlList.java
```diff
@@ -68,6 +68,13 @@ public List<AccessControlEntry> getEntries() {
     return entries;
   }
 
+  public boolean isValid() {
+    for (AccessControlEntry entry : entries) {
+      if (!entry.isValid()) return false;
+    }
+    return true;
```
lkiesow: Maybe a bit simpler?
```suggestion
    return entries.stream().allMatch(AccessControlEntry::isValid);
```

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
```diff
@@ -967,6 +967,11 @@ public Response applyAclToSeries(@PathParam("seriesId") String seriesId, @FormPa
       return badRequest();
     }
 
+    if (!accessControlList.isValid()) {
+      logger.error("POST api/series/{}/access: Invalid series ACL detected", seriesId);
```
lkiesow: Let's not log client errors as errors in the backend:
```suggestion
      logger.debug("POST api/series/{}/access: Invalid series ACL detected", seriesId);
```

lkiesow (changes requested):

lkiesow (approved):

PR #3681, opened 2022-04-19
Make deletion of live publication in case of capture errors configurable

darolfes:
Port of to the community edition.
This PR makes the deletion of a live publication in case of a capture error configurable.

### DIFF
--- a/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/message/SchedulerUpdateHandler.java
+++ b/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/message/SchedulerUpdateHandler.java
@@ -31,21 +31,39 @@ import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.util.NotFoundException;
 
 import org.apache.commons.lang3.BooleanUtils;
+import org.osgi.service.component.ComponentContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Dictionary;
 import java.util.Map;
+import java.util.Objects;
 
 public class SchedulerUpdateHandler extends UpdateHandler {
 
   private static final Logger logger = LoggerFactory.getLogger(SchedulerUpdateHandler.class);
 
   private static final String DESTINATION_SCHEDULER = "SCHEDULER.Liveschedule";
+  private static final String DELETE_ON_CAPTURE_ERROR = "live.deleteOnCapureError";
 
   protected SchedulerService schedulerService;
 
+  private boolean deleteOnCaptureError = true;
+
   public SchedulerUpdateHandler() {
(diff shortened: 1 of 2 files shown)


modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/message/SchedulerUpdateHandler.java
```diff
@@ -31,23 +31,41 @@
...
+import java.util.Dictionary;
 import java.util.Map;
+import java.util.Objects;
 
 public class SchedulerUpdateHandler extends UpdateHandler {
 
   private static final Logger logger = LoggerFactory.getLogger(SchedulerUpdateHandler.class);
 
   private static final String DESTINATION_SCHEDULER = "SCHEDULER.Liveschedule";
+  private static final String DELETE_ON_CAPTURE_ERROR = "live.deleteOnCapureError";
```
lkiesow, line 47: Did you miss adding the configuration file/entry for this?

lkiesow (dismissed):

lkiesow:
Also, a short explanation about the reasoning for this patch would be great. The
The live stream should not work any longer, should it?
Why do we want to keep it around?

KatrinIhler:
@lkiesow Apparently this error can also be caused by temporary issues from which the live stream can recover, but by then the publication has been retracted, so users can no longer watch it. SWITCH had an issue with that with the Extrons and thus patched it out. I thought this might be interesting for the Community as well.

lkiesow:
Looks good.
Unfortunately, this patch does not work for Opencast 12 or `develop` due to

- the removal of ActiveMQ and the restructuring caused by it
- the switch to OSGi annotation, which the `activate()` method would need

Do you need this for `r/11.x`? If you need it, my suggestion is to create another patch for `r/12.x`, merge both in the respective branches and revert the `11.x` one before merging that branch all the way to `develoop`.

If you don't need it for `r/11.x` I suggest targeting `r/12.x` with this patch directly.

All in all, the code got moved around, but it's still pretty much the same.

KatrinIhler:
@darolfes I would suggest an alternative implementation for 12.x, so that we can pull this one into 11.x and then revert it for later versions, as Lars described. But please finish your other work first, that's more important. :)

lkiesow:
Converted to draft, so we know this isn't done. Feel Please convert it back once it is.

darolfes:
r/12.x version is here: #4085

mliradelc:
From @lkiesow comment, is it ready for review and merge? Because I remember that it was converted into draft.

We will release a new version 11 the next week, @darolfes Is #4085 the same PR only targeted for 12.x, or it has some differences?

gregorydlogan:
This is the 11.x version, 4085 is 12.x.  I'm going to merge this, forward 11.x, then do 12.x later today

PR #3682, opened 2022-04-19
Make creation of default External API group configurable

darolfes:
This PR makes the default group creation of the external api configurable.


lkiesow:
What is this default group for the external API? Why is this created at all, and what is it used for?

KatrinIhler:
@lkiesow Quick explanation: The External API group loader will create an EXTERNAL APPLICATIONS group for each organization if it doesn't exist. If you have a different set of groups already, this might be annoying, so SWITCH patched it out. I thought it might be nice to be able to turn this off.

mliradelc:
Because it's a configuration that can anyone activate/deactivate, I will merge it

PR #3687, opened 2022-04-20
Changed inbox behaviour for additional files for scheduled events

### LINKED ISSUE #3649: Inbox Scanner assignment of scheduled events: issue when ingesting >1 recordings for same scheduled event
**Describe the bug**
If more that one recordings are sequentially copied in the inbox folder for the same scheduled event, there are issues with ingesting (usually eg in case of 2 files, the second one hangs in the inbox).
This is a frequent case as lecturers often take a break or they run a short test recording at the beginning of a lecture.

Probably this issue is related to this one:
https://github.com/opencast/opencast/issues/2276

**To Reproduce**
1. Schedule an event that lasts 2 hours
2. copy a file that corresponds to the first part of this event to the inbox folder
3. copy a second file that corresponds to the second half of the same scheduled event to the inbox folder

**Suggested solution**
Automatic ingesting of scheduled events should allow for more than one parts of the same event to be ingested; this could be done by:
– before ingesting, check status of respective media package (maybe wf running?)
– if status "scheduled" then ingest as usual
– if status eg "running" then copy the event to a new media package and ingest the new media to this new media package (named as "... – part 2", "...–part n", for example

Arnei:
Resolves #3649

Previously, additional files for a scheduled event to the inbox would overwrite the previous event or cause an exception in case a workflow was currently running on the scheduled event. Now, new events are created for additional files instead.

Unfortunately I did not find a (reasonably easy) way to number the new events as requested in the linked issue. For now, new events are named after the pattern "originalTitle - newEventId". 

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
@@ -25,19 +25,27 @@ import static java.lang.String.format;
 import static org.opencastproject.scheduler.api.RecordingState.UPLOAD_FINISHED;
 
 import org.opencastproject.ingest.api.IngestService;
+import org.opencastproject.mediapackage.Catalog;
 import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.MediaPackageElements;
+import org.opencastproject.mediapackage.MediaPackageException;
+import org.opencastproject.mediapackage.identifier.Id;
+import org.opencastproject.mediapackage.identifier.IdImpl;
 import org.opencastproject.metadata.dublincore.DublinCore;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalog;
+import org.opencastproject.metadata.dublincore.DublinCoreUtil;
 import org.opencastproject.metadata.dublincore.DublinCores;
 import org.opencastproject.metadata.dublincore.EncodingSchemeUtils;
 import org.opencastproject.metadata.dublincore.Precision;
+import org.opencastproject.scheduler.api.Recording;
 import org.opencastproject.scheduler.api.SchedulerService;
 import org.opencastproject.security.util.SecurityContext;
 import org.opencastproject.series.api.SeriesService;
 import org.opencastproject.util.IoSupport;
+import org.opencastproject.util.NotFoundException;
 import org.opencastproject.workflow.api.WorkflowInstance;
+import org.opencastproject.workspace.api.Workspace;
 
 import com.google.common.util.concurrent.RateLimiter;
 import com.google.gson.Gson;
@@ -57,6 +65,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.sql.Timestamp;
 import java.text.DateFormat;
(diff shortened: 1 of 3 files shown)


modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
```diff
@@ -239,6 +249,31 @@ public RetriableIngestJob call() {
...
+                  try {
+                    Recording recordingState = schedulerService.getRecordingState(id);
+                    if (recordingState.getState().equals(UPLOAD_FINISHED)) {
+                      mediaPackage = (MediaPackage) mediaPackage.clone();
+                      Id newId = IdImpl.fromUUID();
+                      mediaPackage.setIdentifier(newId);
+
+                      // Update dublincore title
+                      try {
+                        String newTitle = mediaPackage.getTitle() + " - " + newId;
```
lkiesow: Some insider knowledge: Never try getting correct metadata from the media package. Always extract it from the Dublin Core catalog. My first test ended up like this:

Screenshot from 2022-04-20 16-40-36

lkiesow: Maybe something like:

```diff
@@ -255,15 +255,17 @@ public class Ingestor implements Runnable {
   try {
     Recording recordingState = schedulerService.getRecordingState(id);
     if (recordingState.getState().equals(UPLOAD_FINISHED)) {
+      var referenceId = mediaPackage.getIdentifier().toString();
       mediaPackage = (MediaPackage) mediaPackage.clone();
-      Id newId = IdImpl.fromUUID();
-      mediaPackage.setIdentifier(newId);
+      mediaPackage.setIdentifier(IdImpl.fromUUID());

       // Update dublincore title
       try {
-        String newTitle = mediaPackage.getTitle() + " - " + newId;
         DublinCoreCatalog dc = DublinCoreUtil.loadEpisodeDublinCore(workspace, mediaPackage).get();
+        var newTitle = dc.get(DublinCore.PROPERTY_TITLE).get(0).getValue()
+            + " (" + Instant.now().getEpochSecond() + ")";
         dc.set(DublinCore.PROPERTY_TITLE, newTitle);
+        dc.set(DublinCore.PROPERTY_REFERENCES, referenceId);
         mediaPackage = updateDublincCoreCatalog(mediaPackage, dc);
         mediaPackage.setTitle(newTitle);
       } catch (Exception e) {
```

This…
- uses the DC title
- adds the original event as reference
- switches from the very long UUID to the ingest time (as unix timestamp) which will at least give you an order for the ingested events

lkiesow (changes requested):

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
Arnei: Very good suggestions, should be implemented now.

lkiesow (approved):
Now this looks better.
Screenshot from 2022-04-22 13-03-01

ypatios:
It seems that the newly created additional events involve a duplication step which copies the initial recording in the new media package. This has as a result that every new part of the same event contains media files of the same flavor (which causes problems in the processing/publishing workflows).
The initial media file(s) should not be included in any of the additional parts. Thus, I regard this as a bug. Right?
Should I reopen the original issue or create a new one?

lkiesow:
A second event will inherit all data already in the asset manager for that event.
We could only filter that to e.g. remove media if we wanted to.

Please file a new issue for that. I only noticed it by accident here.

ypatios:
Yes, it would make sense to omit the initial recording.
I'll create a new issue. Thanks

PR #3715, opened 2022-04-28
Unbreak the metadata editor included with the video editor and save events as well as the video editor state when save is pressed

Rillke:
### Your pull request should…

* [x] have a concise title
* include migration scripts and documentation, if appropriate
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/index.html
+++ b/modules/admin-ui-frontend/app/index.html
@@ -190,6 +190,7 @@
     <script src="scripts/shared/services/progressBarService.js"></script>
     <script src="scripts/shared/services/hotkeysService.js"></script>
     <script src="scripts/shared/services/restServiceMonitor.js"></script>
+    <script src="scripts/shared/services/metadataSaveService.js"></script>
     <script src="scripts/shared/resources/resources.js"></script>
     <script src="scripts/shared/resources/identityResource.js"></script>
     <script src="scripts/shared/resources/versionResource.js"></script>
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -29,12 +29,14 @@ angular.module('adminNg.controllers')
   'EventPublicationsResource', 'EventSchedulingResource','NewEventProcessingResource', 'CaptureAgentsResource',
   'ConflictCheckResource', 'Language', 'JsHelper', '$sce', '$timeout', 'EventHelperService', 'UploadAssetOptions',
   'EventUploadAssetResource', 'Table', 'SchedulingHelperService', 'StatisticsReusable', 'Modal', '$translate',
+  'MetadataSaveService',
   function ($scope, Notifications, EventTransactionResource, EventMetadataResource, EventAssetsResource,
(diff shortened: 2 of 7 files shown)


Rillke:
Before:

image

After:

image


What I did not check is if extended meta data is getting lost during save because we do not extend meta data.

gregorydlogan:
@mtneug you guys use extended metadata, you probably want to check this too.

lkiesow:
@gregorydlogan, right now, saving does not work at all. So, even if extended metadata would not work (and I'm not sure if they actually work in the editor) it is still a big improvement and we should merge it.

lkiesow:
My only concern is that the two “Save” buttons are pretty confusing and I'm wondering if you wouldn't want to save only if you hit the save button down below.

mliradelc:
Thanks for making the PR, 
I also have the same question as @lkiesow about the save button.

Maybe the one for the metadata should be called `Save metadata`, and now comes with another question, Should the Save button for the editor ask the user if there are changes to save them?

Rillke:
Since the PR improves the current situation (no metadata shown at all and no way to change them in the editor, which is a pain for all lecturers coming from LMS directly), I'd like to suggest to merge it so 11.7 can benefit from it, and if it becomes clear what the proper solution to the duplicate save button issue is, I can issue another PR.

From the UX perspective, I love minimal interfaces but with only one save button saving the meta data and starting the WF, the entire process might take too long for the user so we need a progress indicator (a simple throbber should be fine) and in case of errors, the user must know which step was that failed. Do you agree?

mliradelc:
Hi @Rillke, I agree with you that the PR improves to correct the situation and this  fix to merge it into 11.7.

We talked in the technical meeting about the UX, Can you only (for now) left the button to continue the publication and always save the metadata written in the fields?

Rillke:
Hi @mliradelc , I'm not sure I understood what you are requesting so I made a flow chart:

Untitled Diagram

Is this what the people at the technical meeting want me to do?

mliradelc:
Yes, you nailed it!

Rillke:
@mliradelc Here you go. Sorry for the delay. Had to learn Angular.JS (Services, Controllers, Scope, Dependency Injection and Q) and Jasmine Tests for this fix.

mliradelc (approved):
Tested and works, Excellent work!

Now I have a question, I was thinking to only add the "Save metadata" action to the general save button to make it. Was a constraint that you need to create a service?

Anyways, it is an outstanding PR 👏  👏  👏

Rillke:
> Was a constraint that you need to create a service?

Scope. The ToolsCtrl is the parent scope and the child scope is `EventCtrl`. I already extended `EventCtrl` to be the child scope where `submit()` is called; otherwise we would have no access to `commonMetadataCatalog`. From `ToolsCtrl::submit()` we can't call `EventCtrl::metadataSave()` due to how parent-child relationship is done in Angular.JS.

Reading multiple SO-questions about reusing code from 2 different controllers suggested to create services. If this is wrong or there is a nicer way, please advise.

Rillke:
Talking about the new service, I wonder if it would make sense to move some tests from the EventCtrl tests to a new MetadataSaveService test.

mliradelc:
> Scope

Yes, you are right, I'm not very well versed on that part of the code. For me, it's ok. 

About the tests, I would leave for another PR.

PR #3716, opened 2022-04-28
Add missing tags in encode WOH

### LINKED ISSUE #3639: Missing tags when encoding multiple qualities with the encode WOH
**Describe the bug**
When encoding multiple qualities using the encode WOH the media created is not tagged with the suffix. This is a regression introduced beteween Opencast 9 and Opencast 11.

**To Reproduce**
1. Go to https://stable.opencast.org/admin-ng/index.html#!/events/events
2. Login as admin
3. Start the Process upon Upload and Schedule WF on any media you like, and send the file straight to publishing
4. Multiple Qualities are created in one encode WO. [[1]](https://github.com/opencast/opencast/blob/r/11.x/etc/workflows/partial-publish.xml#L218) [[2]](https://github.com/opencast/opencast/blob/r/11.x/etc/encoding/opencast.properties#L34L38) but none of them are tagged as `480p-quality` or similar.

image

**Environment Information**
 - Opencast version with minor release Opencast 11.0.0.SNAPSHOT - 1c5fa68 - built on 2022-04-07, 11.4

Rillke:
Tags for different output files were missing in encode WOH, because
the comparison of the output file name to a specified suffix stopped
working after 19c0940c.
The reason for this is that the files were renamed not including
the specified suffix, and comparison for the suffix as a condition
for adding a tag is used.
Fixes the issue by remembering the original file name from encoding.

Resolves #3639

### Your pull request should…

* [x] have a concise title
* include migration scripts and documentation, if appropriate
* [x] pass automated tests

I might add additional test coverage in a separate PR against develop (as it doesn't seem to be covered yet, in spite of being documented).

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -494,6 +494,7 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
     source.put("video", mediaFile);
     List<File> outputFiles = encoderEngine.process(source, profile, properties);
     var returnURLs = new ArrayList<URI>();
+    var tagsForUrls = new ArrayList<List<String>>();
     activeEncoder.remove(encoderEngine);
     int i = 0;
     var fileMapping = new HashMap<String, String>();
@@ -515,10 +516,21 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
       }
 
       // Put the file in the workspace
+      final List<String> tags = profile.getTags();
       try (InputStream in = new FileInputStream(file)) {
-        var filename = fileMapping.get(file.getName());
-        var url = workspace.putInCollection(COLLECTION, filename, in);
+        var encodedFileName = file.getName();
+        var workspaceFilename = fileMapping.get(encodedFileName);
+        var url = workspace.putInCollection(COLLECTION, workspaceFilename, in);
         returnURLs.add(url);
+
+        var tagsForUrl = new ArrayList<String>();
+        for (final String tag : tags) {
+          if (encodedFileName.endsWith(profile.getSuffix(tag))) {
+            tagsForUrl.add(tag);
+          }
+        }
+        tagsForUrls.add(tagsForUrl);
+
         logger.info("Copied the encoded file to the workspace at {}", url);
       } catch (Exception e) {
         throw new EncoderException("Unable to put the encoded file into the workspace", e);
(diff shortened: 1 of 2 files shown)


Rillke:
```
Error:  testParallelEncode(org.opencastproject.composer.impl.ComposerServiceTest)  Time elapsed: 0.495 s  <<< ERROR!
org.opencastproject.composer.api.EncoderException: Unable to create a job
	at org.opencastproject.composer.impl.ComposerServiceTest.testParallelEncode(ComposerServiceTest.java:298)
Caused by: org.opencastproject.serviceregistry.api.ServiceRegistryException: Error handling operation 'ParallelEncode'
	at org.opencastproject.composer.impl.ComposerServiceTest.lambda$setUp$1(ComposerServiceTest.java:240)
	at org.opencastproject.composer.impl.ComposerServiceTest.testParallelEncode(ComposerServiceTest.java:298)
Caused by: java.lang.NullPointerException
	at org.opencastproject.composer.impl.ComposerServiceTest.lambda$setUp$1(ComposerServiceTest.java:240)
	at org.opencastproject.composer.impl.ComposerServiceTest.testParallelEncode(ComposerServiceTest.java:298)
```

Rillke:
Implemenation seems to be fine, but the [mock must be adjusted not to return the same URI](https://github.com/opencast/opencast/blob/r/11.x/modules/composer-ffmpeg/src/test/java/org/opencastproject/composer/impl/ComposerServiceTest.java#L277) for everything thrown into the workspace.

lkiesow:
I don't think the problem is that the workspace return the same path but rather that the URIs change in `inspect()`.
If you take a look at the track URIs and your mapping after `inspect()` is called, you get:

```
inspected uri: video.mp4                                                                                           
mapping: {file:/home/lars/dev/opencast/modules/composer-ffmpeg/target/video16440899319360639861.mp4=video16440899319360639861_0f80b27a-920e-4d67-bd05-fb8afa94
31b0-high.mp4}
```

Right now, `inspect()` does not necessary guarantee that URIs do not change.
An alternative may be to modify the signature of `insüect()` to allow passing on the tags directly.
That way, we could add them to the tracks right when we create the `Track` objects.
Something like:

```java
  protected List<Track> inspect(Job job, List<URI> uris, List<List<String>> tags) throws EncoderException {
```

or

```java
  protected List<Track> inspect(Job job, Map<URI, List<String>> uris) throws EncoderException {
```

Rillke:
> Right now, `inspect()` does not necessary guarantee that URIs do not change.

Quite surprising; like a customs officer messing with evidence :joy:

Rillke:
@lkiesow  Why do we need `List<List<String>> tags`? There can be only one tag per output file, right?

lkiesow:
> > Right now, `inspect()` does not necessary guarantee that URIs do not change.
> 
> Quite surprising; like a customs officer messing with evidence joy

Maybe, I didn't express this properly. I don't think the `inspect` operation changes the URI right now, and I don't expect this behavior to change any time soon. The ingest-mock does, and that is why the test failed. But in general, operations/jobs are allowed to do that and it wouldn't be “wrong” if in the future someone would change the code in a way so that the URI changes. And it would probably be unexpected to see that change to break something in the composer. That is why, since we have to change code anyway, we could use a more robust variant which does not rely on the internal implementation of the inspect operation.

lkiesow:
> @lkiesow Why do we need `List<List<String>> tags`? There can be only one tag per output file, right?

I think you are right. Passing a single tag would suffice. I was just thinking that in general, a list is more flexible. But I don't really care :D

lkiesow (approved):

PR #3718, opened 2022-04-29
fix transcription workflowoperation dependency

geichelberger:
This PR allows passing an arbitrary number of arguments to the startTranscription function in order to pass transcription service specific configurations.

 The transcription-service-workflowoperation bundle references the amberscript bundle, which is a weird dependency and causes can cause problems e.g., with the Plugin Manager [#3218](https://github.com/opencast/opencast/pull/3218).

* [x] have a concise title
* [x] pass automated tests


geichelberger:
Another possibility would be to split the workflow operation bundle by transcription service.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -272,29 +272,31 @@ public Job startTranscription(String mpId, Track track) throws TranscriptionServ
...
-  public Job startTranscription(String mpId, Track track, String language) throws TranscriptionServiceException {
-    return startTranscription(mpId, track, language, getAmberscriptJobType());
-  }
-
-  public Job startTranscription(String mpId, Track track, String language, String jobtype)
-          throws TranscriptionServiceException {
-    if (StringUtils.isBlank(language)) {
+  public Job startTranscription(String mpId, Track track, String... args) throws TranscriptionServiceException {
+    String language;
+    if (args.length > 0 && StringUtils.isBlank(args[0])) {
```
tyfranck: This doesn't look right. Do you expect an empty/blank language for Amberscript?

tyfranck: Hi @geichelberger, thanks for the PR. Are you able to provide me with an Amberscript client API key so that I can try the new changes.

geichelberger: The argument is a user input from the workflow definition. In theory, it could be blank. I don't have an amberscript API key.

tyfranck: `The argument is a user input from the workflow definition`
Yes but that argument is only assigned to language if it's blank see line 278.  That's the bit I don't get.
`I don't have an amberscript API key`
Hopefully someone with an amberscript client api will be able to test it.

geichelberger: Ah, the negotiation was missing. Fixed it.

tyfranck: Yep, makes sense now.
I've tested Google and IBM transcription and they work fine with new the changes. As for Amberscript the code looks fine but I've not been able to test the new functionality (`jobtype`) added.

gregorydlogan:
LGTM.  I looked at the jobtype bits, looks like it gets passed through directly from the workflow config, and has a reasonable default.  I'm going to assume that this works since we can't test it, and it's going to default to working.

PR #3741, opened 2022-05-05
ADD specialist worker nodes

### LINKED ISSUE #3740: Prefer GPU accelerated nodes when encoding
Opencast should have an easy way to prefer GPU accelerated nodes when encoding, since messing with JobLoads and MaxLoad values in the configuration files on several nodes can get messy quite fast.

berthob98:
### PR Description
This PR adds the ability to define a **list of worker nodes** that gets preferred when **dispatching compilation jobs** (Job Type: org.opencastproject.composer). This could for example be useful in a setup with one or more GPU accelerated nodes.
This feature is disabled by default and only activated when a list of specialized worker nodes is set. The comma-separated list of worker nodes is defined in the configuration file (etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg).
To ensure backwards compatibility not defining a list of specialized worker nodes is safe and leaves the behavior of the system unchanged.
When dispatching composer jobs, the defined worker nodes are chosen over every not specialized worker nodes if the LoadFactor of the specialized worker nodes is 0.4 or smaller and if the LoadFactor of the specialized worker node is smaller than 4 times the LoadFactor of the not specialized worker node. This mechanism is meant to prevent the overload of specialized nodes while ensuring to dispatch encoding jobs to the defined nodes if there is capacity. Since this is a (working) draft feedback to this idea is appreciated. It would also be possible to make the threshold for preferring specialized nodes configurable via the configuration file, if that is desirable.
If wanted, I could also add a definition of this feature to the admin configuration.

**Information for easier code review:**
- The variable org.opencastproject.encoding.workers is retrieved from the etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg configuration file
- When dispatching jobs a list of candidate nodes gets sorted by a comparator class. The dispatching attempts follow the order of the list.
- I made a copy of the existing comparator that only is used when dealing with a job of type composer. This comparator prefers the worker nodes retrieved form the configuration file.

This PR closes #3740.


mliradelc:
Hi @berthob98, thanks for your PR submission, in the next days I will try your PR in a cluster that has one worker with GPU and another without GPU.

Before to merge this PR you need to sign up the ICLA (In the bot comment).

And because it's a new feature you need to [add a txt file of your submission](https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs/releasenotes)

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -3419,6 +3437,72 @@ public int compare(ServiceRegistration serviceA, ServiceRegistration serviceB) {
...
+      String hostB = serviceB.getHost();
+      NodeLoad nodeA = loadByHost.get(hostA);
+      NodeLoad nodeB = loadByHost.get(hostB);
+
+
+
+      if (encodingWorkers != null) {
+
+        if (isEncodingWorker(hostA, encodingWorkers) && !isEncodingWorker(hostB, encodingWorkers)) {
+          if (nodeA.getLoadFactor() <= 0.4) {
```
gregorydlogan: Where did these magic load factors come from?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -2844,8 +2857,13 @@ protected List<ServiceRegistration> getServiceRegistrationsByLoad(String jobType
       filteredList.add(service);
     }
 
-    // Sort the list by capacity
-    Collections.sort(filteredList, new LoadComparator(systemLoad));
+    // Sort the list by capacity and distinguish between composer jobs and other jobs
+    if ("org.opencastproject.composer".equals(jobType)) {
+        Collections.sort(filteredList, new LoadComparatorEncoding(systemLoad));
```
gregorydlogan: I'm surprised that checkstyles isn't complaining here.  This line has at least two extra spaces.  We also do K&R bracing (this isn't checked afaik).

gregorydlogan (changes requested):

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: I thought there should be a threshold for loadFactors, below which a specialized worker (short SW.) always should be preferred against a non SW. This is meant to avoid a non SW with loadFactor 0 getting preferred over a SW that still has a low loadFactor of for example 0.1.

The value 0.4 was a first guess. Doing the math, 0.4 allows a SW with 16 cores to have 8 encoding jobs (à 0.8 JobLoad) before considering a non SW, which seems way to high. Maybe right now SW only should be preferred in idle situations.


I'm still thinking about how to better meet certain guarantees when dispatching, while also properly using the SW and reasonably spreading the load across the system.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: I will change this.

berthob98:
> Hi @berthob98, thanks for your PR submission, in the next days I will try your PR in a cluster that has one worker with GPU and another without GPU.
> 
> Before to merge this PR you need to sign up the ICLA (In the bot comment).
> 
> And because it's a new feature you need to [add a txt file of your submission](https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs/releasenotes)

Hi @mliradelc, thank you for testing. I will sign the ICLA and make the txt file.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
gregorydlogan: If the goal here is to load SWs with encoding jobs prior to spilling over to other nodes then how about

```
if nodeA and nodeB are both encoders
  nodeA load + job load >= nodeA's maxload ? nodeB : nodeA
elseif either node is an encoder and that node's load + job load <= node's maxload
  that node
else
  the other node
```

This prefers an encoder, unless all of them are full in which case it picks another.

mliradelc:
@berthob98 Today is the release cut, I need you to sign the ICLA so we can merge it. I will start testing right now

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: By my understanding the services accepting already check whether the MaxLoad is exceeded, by inheriting the isReadyToAccept function from the AbstractJobProducer. But I think filling the SW up to the MaxLoad before considering non SW, is a to high value when trying to evenly spread the load across the system. With the MaxLoad defaulting to the number of CPU cores that means a SW with 16 Cores could have 16 encoding jobs before considering non SW servers.

berthob98:
@mliradelc Ok I just signed the ICLA

mliradelc (changes requested):
Thanks for sign up!

I just realized that your PR is based of r/11.x instead of develop branch, can you change your base branch?

berthob98:
> Thanks for sign up!
> 
> I just realized that your PR is based of r/11.x instead of develop branch, can you change your base branch?

@mliradelc Ok I rebased the branch.

lkiesow:
A short notice that Apereo has received @berthob98's ICLA. As usual, it might take a bit of time for them to process it, meaning that you might still see complaining tests if you file new pull requests. Ignore that and point to this comment if necessary. There is now no reason to not accept patches from you based on the ICLA. Also, thanks for filing it!

mliradelc:
I tried the code, but I don't understand very well one thing,

When I select the worker using the new option in the service registry file, It's always choose as a first option. But when I have more than one encoding job It should queue those jobs to the GPU worker, but that doesn't happen. Is that the correct behavior?

Looking into the workers loads, they still select what worker to use with the profile jobload (Inside each profile), and at the end the non gpu workers also start encoding.

mliradelc:
> > Thanks for sign up!
> > I just realized that your PR is based of r/11.x instead of develop branch, can you change your base branch?
> 
> @mliradelc Ok I rebased the branch.

Thanks for signing up and look up for the base branch

berthob98:
> I tried the code, but I don't understand very well one thing,
> 
> When I select the worker using the new option in the service registry file, It's always choose as a first option. But when I have more than one encoding job It should queue those jobs to the GPU worker, but that doesn't happen. Is that the correct behavior?
> 
> Looking into the workers loads, they still select what worker to use with the profile jobload (Inside each profile), and at the end the non gpu workers also start encoding.
> 
> 

@mliradelc I am sorry for the late answer I missed your comment.
The behavior you described is expected. The code only prefers GPU workers in either of two cases:
- if the GPU workers NodeLoad is under 40%
- if there is no normal Node with 25% or less of the GPU workers NodeLoad

Sending every encoding job to a single GPU worker could lead to encoding jobs queuing up on a single GPU worker while other workers idle, producing unnesessary congestion.

In my tests modern GPUs where about 4 times faster than modern CPUs when encoding. It also should be considered that modern GPUs only have around 2-3 encoding units.

I am still testing how to improve this mechanism and interested in diffrent ideas.

mliradelc:
Hi @berthob98, I miss completely your answer. I read when you published, and then I forgot completely about it, my apologies.

How I think a Specialized worker should work.  (From my opinion)
- The Specialized worked *must* take priority all the encoding tasks
- If there is another specialized worker, they should balance their loads
- A non specialized worker *should only* take an encoding task if the ETA of the specialized workers is later than the ETA of the non-specialized worker.
  - Example, A specialized worker has a queue that will take 15 minutes to finish and there is a new video that will take 5 minutes to finish. So the ETA will be in 20 minutes. In the same cluster, you have a non-specialized worker free, it will take 15 minutes to encode that video. The admin should choose the non-specialized worker for that case.


Because of the nature of this feature, where the encoding hardware can be very different between them, I think that will be handy to have a benchmark tool inside opencast. A set of standardized test to try for now the raw power of the machine. With that, you can easily can configure the best load percentages for each machine using the slowest worker as baseline.

Another thing that needs to be clarified, is the capabilities of the machine to do parallel encoding, I've seen with CUDA, that even with 1 encoding engine, you can make 4 jobs with a little decrease of performance.


This topic is a huge one, but if it solved, can make the encoding tasks up to 10x faster for large organizations

gregorydlogan:
I'd be quite happy to see a benchmarking tool, for a number of reasons, but I suspect that's outside the bounds of this PR - unless @berthob98 *wants* to do it :)

What would improve the PR is allowing the user to adjust the limits where the special logic kicks in/out.

berthob98:
> What would improve the PR is allowing the user to adjust the limits where the special logic kicks in/out.

@gregorydlogan  I made the limit adjustable via the cfg file

berthob98:
>     * if there is no normal Node with 25% or less of the GPU workers NodeLoad
 
 I also removed this function, since encoding doesn't scale up like that for all GPUs. The number of parallel encodings is limited on GPUs, some have a pretty low (artificial) limit.

berthob98:
> Because of the nature of this feature, where the encoding hardware can be very different between them, I think that will be handy to have a benchmark tool inside opencast. A set of standardized test to try for now the raw power of the machine. With that, you can easily can configure the best load percentages for each machine using the slowest worker as baseline.

@mliradelc What would be interesting measurments?

- Encoding time of a standard video compared on the CPU and GPU?
- Encoding time of multiple (2, 5, 10) videos compared to the encoding time of a single video?

mliradelc:
I would keep it very simple from the start, Like, use the frames per second for the encoding task.

About the videos to encode and the FFmpeg command, you can leave that configurable to the user. What encoding profile to use and left the video to encode in a set folder.

About the limit problem from the GPU, that is very true, and you can have fail jobs because the GPU refused to encode more than the driver allows. That can be limited by configuration.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -197,6 +198,12 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
   /** Configuration key for the retrieval of service statistics: Do not consider jobs older than max_job_age (in days) */
   protected static final String OPT_SERVICE_STATISTICS_MAX_JOB_AGE = "org.opencastproject.statistics.services.max_job_age";
 
+  /** Configuration key for the encoding preferred worker nodes */
+  protected static final String OPT_ENCODINGWORKERS = "org.opencastproject.encoding.workers";
+
+  /** Configuration key for the encoding workers load threshold */
+  protected static final String OPT_ENCODINGTHRESHOLD = "org.opencastproject.encoding.workers.threshold";
+
```
ziegenberg:
```suggestion
  /** Configuration key for the encoding preferred worker nodes */
  protected static final String OPT_ENCODING_WORKERS = "org.opencastproject.encoding.workers";

  /** Configuration key for the encoding workers load threshold */
  protected static final String OPT_ENCODING_THRESHOLD = "org.opencastproject.encoding.workers.threshold";

```
Using snake case would be easier to read.

etc/org.opencastproject.serviceregistry.impl.ServiceRegistryJpaImpl.cfg
```diff
@@ -28,3 +28,13 @@
...
+
+
+# Comma-separated list of encoding specialized worker nodes. Specified workers are preferred when dispatching encoding jobs.
+# Default: empty
+#org.opencastproject.encoding.workers=http://example.com, http://example1.com:8080
+
+# Numeric value between 0 and 1. Sets an load value under which specified encoding workers are preferred over other workers when dispatching encoding jobs.
+# A value of 0.2 corresponds to a threshold of 20% worker load
+# Default: 0.0
+#org.opencastproject.encoding.workers.threshold=
```
ziegenberg: This is missing a new line at the end of the file.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: Thank you. I applied these changes.

gregorydlogan:
@berthob98 do you want to do this benchmarking as part of this pr, or a follow on? Happy either way, but this pr is starting to get pretty old.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
ziegenberg: Thx!

berthob98:
> @berthob98 do you want to do this benchmarking as part of this pr, or a follow on? Happy either way, but this pr is starting to get pretty old.

I will do the benchmarking as another PR. Is there something I can do to help merge this PR?

mliradelc:
For me, it's ok, works and now with the another PR that you made to check the CPU usage is a good solution. 

Now you need to solve the conflict that GitHub is flagging and for me is good to go.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -3419,6 +3458,68 @@ public int compare(ServiceRegistration serviceA, ServiceRegistration serviceB) {
...
+        //When we're comparing the load factors we want the node with the lowest factor to be first
+        //When we're comparing the maximum load value, we want the node with the highest max to be first
+        return Float.compare(nodeB.getMaxLoad(), nodeA.getMaxLoad());
+      }
+      return Float.compare(nodeA.getLoadFactor(), nodeB.getLoadFactor());
+
+    }
+
+    private boolean isEncodingWorker(String host, List<String> encodingWorkersList) {
+      Iterator<String> workerIterator = encodingWorkersList.iterator();
```
gregorydlogan: This could be an `encodingWorkersList.contains(host)` right?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -3419,6 +3458,68 @@ public int compare(ServiceRegistration serviceA, ServiceRegistration serviceB) {
...
+        }
+        if (isEncodingWorker(hostB, encodingWorkers) && !isEncodingWorker(hostA, encodingWorkers)) {
+          if (nodeB.getLoadFactor() <= encodingThreshold) {
+            return 1;
+          }
+          return Float.compare(nodeA.getLoadFactor(), nodeB.getLoadFactor());
+        }
+      }
+
+      //If the load factors are about the same, sort based on maximum load
```
gregorydlogan: I would be very, very tempted to subclass `LoadComparator`, and then replace this copy+pasted code with a call to `super.compare()` so that there's less chance of comparator implementation drift over time.

gregorydlogan (changes requested):
Very minor changes, and you'll probably want to rebase this anyway :)  Looks really good otherwise!

berthob98:
@gregorydlogan good points. I applied both things and also did some small changes to the parsing of the config file.
Should I squash my commits?

mliradelc:
Thanks for the changes, a squash is not necessary (Before it was but today GitHub squash and merges) but welcome because you can improve the commit description.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: I changed this

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
berthob98: I changed this

PR #3775, opened 2022-05-17
Indexing only latest workflows for all events

Alaschinger:
Added 2 named queries and the corresponding methods within the workflowDatabaseService to count and read the latest/active workflows of each event.

Replaced the old list of the repopulate function within WorkflowServiceImpl containing all workflows with a new one containing only the active/latest workflows.

The rest of the indexing procedure remains the same.

This should save a lot of time when updating the elasticsearch index because we update only one elasticsearch doc per event instead of as many docs as workflows per event.

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
@@ -82,6 +82,22 @@ import javax.xml.bind.annotation.adapters.XmlAdapter;
                 name = "Workflow.findAll",
                 query = "select w from WorkflowInstance w where w.organizationId=:organizationId order by w.dateCreated"
         ),
+        // Finding the latest or active workflow for all events
+        @NamedQuery(
+                name = "Workflow.findLatest",
+                query = "SELECT w FROM WorkflowInstance w "
+                      + "WHERE "
+                      + "w.organizationId=:organizationId AND "
+                      + "w.dateCreated = (SELECT MAX(w2.dateCreated) "
+                      + "FROM WorkflowInstance w2 "
+                      + "WHERE w2.mediaPackageId = w.mediaPackageId AND "
+                      + "w2.workflowId>:startToken) "
+                      + "ORDER BY w.workflowId ASC"
+        ),
+        @NamedQuery(
+                name = "Workflow.countLatest",
+                query = "SELECT COUNT(DISTINCT mediapackage_id) FROM oc_workflow"
+        ),
         @NamedQuery(
                 name = "Workflow.findAllOrganizationIndependent",
                 query = "select w from WorkflowInstance w"
@@ -513,5 +529,3 @@ public class WorkflowInstance {
     }
   }
 }
-
-
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
@@ -96,7 +96,7 @@ import javax.xml.bind.annotation.adapters.XmlAdapter;
         ),
         @NamedQuery(
                 name = "Workflow.countLatest",
-                query = "SELECT COUNT(DISTINCT mediapackage_id) FROM oc_workflow"
+                query = "SELECT COUNT(DISTINCT w.mediaPackageId) FROM WorkflowInstance w"
         ),
         @NamedQuery(
                 name = "Workflow.findAllOrganizationIndependent",
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
@@ -79,6 +79,29 @@ public interface WorkflowServiceDatabase {
   long countWorkflows(WorkflowInstance.WorkflowState state) throws WorkflowDatabaseException;
 
   /**
+   * Gets latest workflow instances for all events.
+   * Selects workflows which haven't been finished (DateCompleted IS NULL) or greatest DateCompleted
+   *
+   * @param limit
+   *          max number of workflows to be returned
+   * @param offset
+   *          only return workflows from this point onwards
+   * @return list of all {@link WorkflowInstance}s
+   * @throws WorkflowDatabaseException
+   *           if there is a problem communicating with the underlying data store
+   */
+  List<WorkflowInstance> getLatestWorkflowInstances(int limit, long token) throws WorkflowDatabaseException;
+
+  /**
+   * Returns the number of events workflows have been run on.
+   *
+   * @return the number of latest workflows
+   * @throws WorkflowDatabaseException
+   *           if there is a problem communicating with the underlying data store
+   */
+  int countMediaPackages() throws WorkflowDatabaseException;
+
+  /**
    * Gets all workflow instances for a mediapackage.
    *
    * @param mediaPackageId
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
@@ -172,7 +172,7 @@ public class WorkflowServiceDatabaseImpl implements WorkflowServiceDatabase {
     }
   }
 
-  /**
+    /**
    * {@inheritDoc}
    *
    * @see WorkflowServiceDatabase#countWorkflows(WorkflowInstance.WorkflowState state)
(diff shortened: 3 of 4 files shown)


modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -82,6 +82,32 @@
...
+                      + "OR "
+                      + "w.dateCompleted = (SELECT MAX(w2.dateCompleted) "
+                      + "FROM WorkflowInstance w2 "
+                      + "WHERE w2.mediaPackageId = w.mediaPackageId) "
+                      + ") "
+                      + "GROUP BY w.mediaPackageId"
+        ),
+        @NamedQuery(
+                name = "Workflow.countLatest",
+                query = "SELECT COUNT(w.mediaPackageId) FROM WorkflowInstance w "
```
lkiesow: This is a far slower and much more complex version of something like this:

```sql
select count(distinct mediapackage_id) from oc_workflow;
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -82,6 +82,32 @@
...
+                      + "WHERE "
+                      + "w.organizationId=:organizationId AND "
+                      + "((w.state = :stateInstantiated OR w.state = :statePaused OR w.state = :stateRunning OR w.state = :stateFailing) "
```
lkiesow: Why do we not count e.g. failed workflows?

lkiesow: Now that I realize what this check is for, why not just use `max(date_started)` instead? This will not necessarily even give you tha latest workflow? That should be simpler.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -82,6 +82,32 @@
...
+                query = "SELECT w FROM WorkflowInstance w "
+                      + "WHERE "
+                      + "w.organizationId=:organizationId AND "
+                      + "((w.state = :stateInstantiated OR w.state = :statePaused OR w.state = :stateRunning OR w.state = :stateFailing) "
+                      + "OR "
+                      + "w.dateCompleted = (SELECT MAX(w2.dateCompleted) "
+                      + "FROM WorkflowInstance w2 "
+                      + "WHERE w2.mediaPackageId = w.mediaPackageId) "
+                      + ") "
+                      + "GROUP BY w.mediaPackageId"
```
lkiesow: I don't think this `grooup by` statement works with PostgreSQL. Did you test that?
But then, I don't think it's actually necessary at all.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -82,6 +82,32 @@
                 name = "Workflow.findAll",
                 query = "select w from WorkflowInstance w where w.organizationId=:organizationId order by w.dateCreated"
         ),
+        // Finding the latest or active workflow for all events
+        @NamedQuery(
+                name = "Workflow.findLatest",
+                query = "SELECT w FROM WorkflowInstance w "
```
lkiesow, line 88: Looking at this, I'm wondering if we shouldn't go with native data types right away and just ditch the in database and just do something like 
```sql
SELECT id, mediapackage_id, state FROM oc_workflow order by date_created;
```
Getting these data would be simpler, faster and would probably even take less memory (compare with this approach):

```
opencast=# EXPLAIN ANALYSE
SELECT w.id, w.mediapackage_id, w.state FROM oc_workflow w
where w.date_created = (SELECT MAX(date_created) FROM oc_workflow w2 WHERE w2.mediapackage_id = w.mediapackage_id);
                                                                           QUERY PLAN                                                                           
----------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on oc_workflow w  (cost=0.00..64112.57 rows=38 width=49) (actual time=0.051..63.822 rows=7513 loops=1)
   Filter: (date_created = (SubPlan 1))
   Rows Removed by Filter: 44
   SubPlan 1
     ->  Aggregate  (cost=8.30..8.31 rows=1 width=8) (actual time=0.007..0.007 rows=1 loops=7557)
           ->  Index Scan using ix_oc_workflow_mediapackage_id on oc_workflow w2  (cost=0.28..8.30 rows=1 width=8) (actual time=0.005..0.006 rows=1 loops=7557)
                 Index Cond: ((mediapackage_id)::text = (w.mediapackage_id)::text)
 Planning Time: 0.230 ms
 Execution Time: 64.442 ms
(9 rows)

opencast=# EXPLAIN ANALYSE
SELECT id, mediapackage_id, state FROM oc_workflow order by date_created;
                                                     QUERY PLAN                                                      
---------------------------------------------------------------------------------------------------------------------
 Sort  (cost=1647.52..1666.45 rows=7571 width=57) (actual time=7.998..8.725 rows=7557 loops=1)
   Sort Key: date_created
   Sort Method: quicksort  Memory: 1255kB
   ->  Seq Scan on oc_workflow  (cost=0.00..1159.71 rows=7571 width=57) (actual time=0.011..4.259 rows=7557 loops=1)
 Planning Time: 0.084 ms
 Execution Time: 9.925 ms
(6 rows)
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
```diff
@@ -82,6 +82,32 @@
...
+                      + "WHERE "
+                      + "w.organizationId=:organizationId AND "
+                      + "((w.state = :stateInstantiated OR w.state = :statePaused OR w.state = :stateRunning OR w.state = :stateFailing) "
+                      + "OR "
+                      + "w.dateCompleted = (SELECT MAX(w2.dateCompleted) "
```
lkiesow: Same as with the other query: created instead of state and completed?

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
```diff
@@ -70,6 +70,29 @@ public interface WorkflowServiceDatabase {
...
+  List<WorkflowInstance> getLatestWorkflowInstances(int limit, int offset) throws WorkflowDatabaseException;
+
+  /**
+   * Gets the total number of latest workflows for all events.
+   *
+   * @return the number of latest workflows
+   * @throws WorkflowDatabaseException
+   *           if there is a problem communicating with the underlying data store
+   */
+  int getLatestWorkflowCount() throws WorkflowDatabaseException;
```
lkiesow: Maybe easier to understand?

```suggestion
   * Returns the number of events workflows have been run on.
   *
   * @return the number of latest workflows
   * @throws WorkflowDatabaseException
   *           if there is a problem communicating with the underlying data store
   */
  int countMediaPackages() throws WorkflowDatabaseException;
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
```diff
@@ -162,6 +162,67 @@ public List<WorkflowInstance> getWorkflowInstances(int limit, int offset) throws
...
+      var query = em.createNamedQuery("Workflow.findLatest", WorkflowInstance.class);
+
+      String orgId = securityService.getOrganization().getId();
+      query.setParameter("organizationId", orgId);
+      query.setParameter("stateInstantiated", WorkflowInstance.WorkflowState.INSTANTIATED);
+      query.setParameter("stateRunning", WorkflowInstance.WorkflowState.RUNNING);
+      query.setParameter("statePaused", WorkflowInstance.WorkflowState.PAUSED);
+      query.setParameter("stateFailing", WorkflowInstance.WorkflowState.FAILING);
+      query.setMaxResults(limit);
+      query.setFirstResult(offset);
```
lkiesow: Careful with limit and offset. The order may change while paging through this.
Token-based pagination may be better suited here.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
```diff
@@ -162,6 +162,67 @@ public List<WorkflowInstance> getWorkflowInstances(int limit, int offset) throws
...
+   *
+   * @see WorkflowServiceDatabase#getLatestWorkflowCount()
+   */
+  public int getLatestWorkflowCount() throws WorkflowDatabaseException {
+
+    EntityManager em = null;
+    try {
+      em = emf.createEntityManager();
+
+      Query query = em.createNamedQuery("Workflow.countLatest");
```
lkiesow: So you don't need type casting later:

```suggestion
      var query = em.createNamedQuery("Workflow.countLatest", Long.class);
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2316,11 +2316,10 @@ public Map<String, Map<String, String>> getWorkflowStateMappings() {
 
   @Override
   public void repopulate(final ElasticsearchIndex index) throws IndexRebuildException {
-    final String startWorkflow = Operation.START_WORKFLOW.toString();
     final int total;
     try {
-      total = serviceRegistry.getJobCount(startWorkflow);
-    } catch (ServiceRegistryException e) {
+     total = persistence.getLatestWorkflowCount();
```
lkiesow: indentation

lkiesow (changes requested):

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowInstance.java
lkiesow: As per discussion: For another pull request…

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
```diff
@@ -197,6 +197,56 @@ public long countWorkflows(WorkflowInstance.WorkflowState state) throws Workflow
...
+  public int countMediaPackages() throws WorkflowDatabaseException {
+
+    EntityManager em = null;
+    try {
+      em = emf.createEntityManager();
+
+      var query = em.createNamedQuery("Workflow.countLatest", Long.class);
+      logger.debug("Counting latest workflows using query: {}", query);
+      final Number countResult = query.getSingleResult();
+      return countResult.intValue();
```
lkiesow, line 241: When I suggested using a `long` I actually meant it to simplify the whole method:

```suggestion
  public long countMediaPackages() throws WorkflowDatabaseException {
    EntityManager em = null;
    try {
      em = emf.createEntityManager();

      var query = em.createNamedQuery("Workflow.countLatest", Long.class);
      logger.debug("Counting latest workflows using query: {}", query);
      return query.getSingleResult();
```

gregorydlogan:
I can test this locally, but I'd appreciate it if some of the adopters with more complex wf setups tested this too.

CC @mtneug @KatrinIhler

lkiesow (approved):
This works as expected and can reduce the time for workflow rebuilds quite dramatically depending on your set-up. The new time should basically be `min(previous_time, previous_time / n_workflows × n_events)`. For example, if you have 500 events and 1000 workflows, it will basically cut your workflow index rebuild time in half:

oc-3775-index-rebuild-time

From the responses we got so far, we know that the average ratio is 1.648 workflows per event, going up to 4.282 workflows per event. There are a few instances with less workflows than events but the patch should just do basically nothing for them and not hurt those.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
lkiesow: Missed that I still had a comment on this. It's not really important, though. I'll merge it like this.

PR #3797, opened 2022-05-25
Add missing elasticsearch dependencies to opencast-ingest feature

geichelberger:
The ingest service has a dependency on the elastic search bundles, but the feature does not contain those. This PR adds the elastic search bundles to the opencast-ingest feature.

It would be the first time that a non-admin node uses the index. Can this have side effects?

```
Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=opencast-ingest; type=karaf.feature; version="[12.0.0,12.0.0]"; filter:="(&(osgi.identity=opencast-ingest)(type=karaf.feature)(version>=12.0.0)(version<=12.0.0))" [caused by: Unable to resolve opencast-ingest/12.0.0: missing requirement [opencast-ingest/12.0.0] osgi.identity; osgi.identity=opencast-ingest-service-impl; type=osgi.bundle; version="[12.0.0,12.0.0]"; resolution:=mandatory [caused by: Unable to resolve opencast-ingest-service-impl/12.0.0: missing requirement [opencast-ingest-service-impl/12.0.0] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.opencastproject.elasticsearch.api)(version>=12.0.0)(!(version>=13.0.0)))"]]
```

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -575,6 +575,9 @@
   <feature name="opencast-ingest" version="${project.version}">
     <feature start-level="80">opencast-core</feature>
 
+    <bundle start-level="82">mvn:org.opencastproject/opencast-elasticsearch-api/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-elasticsearch-impl/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-elasticsearch-index/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-inspection-service-remote/${project.version}</bundle>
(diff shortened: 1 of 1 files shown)


lkiesow:
> It would be the first time that a non-admin node uses the index. Can this have side effects?

I think that will break the ingest distributions since the default setup doesn't allow them access to Elasticsearch.

Do you know where this dependency came from?

geichelberger:
It was introduced with the removal of the series service solr index. I deployed the fix on our cluster. The ingest node is working now, but I have to test it further.

lkiesow:
> It was introduced with the removal of the series service solr index. I deployed the fix on our cluster. The ingest node is working now, but I have to test it further.

The problem is that according to our docs, no node except admin needs to have access to Elasticsearch. So this may cause significant infrastructure trouble. At least, we need a major documentation update.

lkiesow:
Seems like this is used only for this new feature: https://github.com/opencast/opencast/pull/3586
Let's see if we cannot implement that differently

lkiesow:
Pull request #3810 contains an alternative approach to fixing the issue, removing the dependency altogether.

PR #3805, opened 2022-05-27
Set Title and start date and time for duplicate events using duplicate WFH

tyfranck:
For rolling forward Lectures from one year to the next we need to be able to change the Title and Start Date and Time of the duplicated mediapackage.

This PR is based on PR #3409 .

Fix 2 issues highlighted in the comments of the above PR
- Date changes not reflected
- issues with empty date

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -35,8 +35,8 @@ Parameter Table
 |max-number-of-events |`1000`                                     |How many events are allowed to be created at maximum.|
 |no-suffix            |`true`                                     |Boolean to decide if the number suffix (`copy #`) is not attached (true ... no attachment, false ...`copy #` is attached) |
 |set-series-id        |`3547a900-e0ee-4e3f-9e67-2157cd42c700`     |Id of the series the copied events should be added to |
-|set-title            |`copy of mediapackage`                     |Title of the newly created events. If `no-suffix` is set to `false` a `copy #` will be attached |
-|set-start-date-time  |`2021-12-02 22:22:22`  or `2021-12-02 22:22`  |Date and time to be set on the newly created events.|
+|set-title            |`copy of mediapackage`                     |Title of the newly created events. Default to original event title. If `no-suffix` is set to `false` a `copy #` will be attached |
+|set-start-date-time  |`2021-12-02T22:22:22`                      |Date and time to be set on the newly created events. Default to orignial event date. |
 
 
 Operation Example
@@ -55,7 +55,7 @@ Operation Example
         <configuration key="no-suffix">false</configuration
         <configuration key="set-series-id">${seriesId}</configuration>
         <configuration key="set-title">${mpTitle}</configuration>
-        <configuration key="set-start-date-time">${startDate} ${startTime}</configuration>
+        <configuration key="set-start-date-time">${startDateTime}</configuration>
         <configuration key="property-namespaces">
           org.opencastproject.assetmanager.security
         </configuration>
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -36,7 +36,7 @@ Parameter Table
 |no-suffix            |`true`                                     |Boolean to decide if the number suffix (`copy #`) is not attached (true ... no attachment, false ...`copy #` is attached) |
 |set-series-id        |`3547a900-e0ee-4e3f-9e67-2157cd42c700`     |Id of the series the copied events should be added to |
 |set-title            |`copy of mediapackage`                     |Title of the newly created events. If `no-suffix` is set to `false` a `copy #` will be attached |
-|set-start-date-time  |`2021-12-02 22:22:22`                      |Date and time to be set on the newly created events.|
+|set-start-date-time  |`2021-12-02 22:22:22`  or `2021-12-02 22:22`  |Date and time to be set on the newly created events.|
 
 
 Operation Example
--- a/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
@@ -16,8 +16,8 @@ For each duplicated event the new media package ID is stored as a workflow prope
 
 |Name                                    |Example                                                             |Description                                    |
 |----------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|
-|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
-|duplicate\_media\_package\_ids          |`duplicate_media_package_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
+|duplicate_media_package_*number*_id     |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
+|duplicate_media_package_ids             |`duplicate_media_package_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
 
 \* will be deprecated
 
@@ -33,6 +33,10 @@ Parameter Table
 |copy-number-prefix   |`copy`                                     |The prefix used for the number of the copy which is appended to the title of the new event.     |
 |number-of-events     |`2`                                        |How many events to create.|
 |max-number-of-events |`1000`                                     |How many events are allowed to be created at maximum.|
+|no-suffix            |`true`                                     |Boolean to decide if the number suffix (`copy #`) is not attached (true ... no attachment, false ...`copy #` is attached) |
+|set-series-id        |`3547a900-e0ee-4e3f-9e67-2157cd42c700`     |Id of the series the copied events should be added to |
+|set-title            |`copy of mediapackage`                     |Title of the newly created events. If `no-suffix` is set to `false` a `copy #` will be attached |
+|set-start-date-time  |`2021-12-02 22:22:22`                      |Date and time to be set on the newly created events.|
 
 
 Operation Example
@@ -47,7 +51,11 @@ Operation Example
         <configuration key="source-flavors">*/*</configuration>
         <configuration key="number-of-events">${numberOfEvents}</configuration>
         <configuration key="max-number-of-events">1000</configuration>
-        <configuration key="target-tags"></configuration>
+        <configuration key="target-tags"></configuration> 
+        <configuration key="no-suffix">false</configuration
+        <configuration key="set-series-id">${seriesId}</configuration>
+        <configuration key="set-title">${mpTitle}</configuration>
+        <configuration key="set-start-date-time">${startDate} ${startTime}</configuration>
         <configuration key="property-namespaces">
           org.opencastproject.assetmanager.security
         </configuration>
--- a/etc/workflows/duplicate-event.xml
+++ b/etc/workflows/duplicate-event.xml
@@ -81,7 +81,7 @@
         $(function () {
             $('input[type="datetime-local"]').setNow();
         });
-        
+
       </script>
     ]]>
   </configuration_panel>
--- a/etc/workflows/duplicate-event.xml
+++ b/etc/workflows/duplicate-event.xml
@@ -9,29 +9,24 @@
   <description/>
   <displayOrder>10</displayOrder>
   <configuration_panel>
-    <![CDATA[
+<![CDATA[
       <div id="workflow-configuration">
         <fieldset>
           <br/>
           <div>
-            <label for="mpTitle" style="width: 150px;" >Title of Duplicate:</label>
+            <label for="mpTitle" style="width: 150px;">Title of Duplicate:</label>
             <input id="mpTitle" name="mpTitle" type="text" style="width: 400px;" value=" " class="configField" />
           </div>
           <br/>
           <div>
-            <label for="seriesId" style="width: 150px;" >Series Id:</label>
+            <label for="seriesId" style="width: 150px;">Series Id:</label>
             <input id="seriesId" name="seriesId" type="text" maxlength="255" style="width: 400px;" value=" " class="configField" />
           </div>
           <br/>
           <div>
-            <label for="startDate" style="width: 150px;">StartDate (YYYY-MM-DD):</label>
-            <input id="startDate" name="startDate" type="text" style="width: 200px;" value=" " class="configField" />
+            <label for="startDateTime" style="width: 150px;">Date time:</label>
+            <input id="startDateTime" name="startDateTime" type="datetime-local" style="width: 300px;" class="configField" required />
           </div>
-          <div>
-            <label for="startTime" style="width: 150px;">startTime (HH:MM:SS):</label>
-            <input id="startTime" name="startTime" type="text" style="width: 200px;" value=" " class="configField" />
-          </div>
-          <br/>
           <br/>
           <div>
             <label for="numberOfEvents" style="width: 150px;">Number of Events</label>
@@ -65,6 +60,28 @@
           if (parseInt(value) < parseInt(min)) $('#numberOfEvents').val(min);
           if (parseInt(value) > parseInt(max)) $('#numberOfEvents').val(max);
         }
+
+        $.fn.setNow = function (onlyBlank) {
+          var now = new Date($.now()), year, month, date, hours, minutes, seconds, formattedDateTime;
+          year = now.getFullYear();
+          month = now.getMonth().toString().length === 1 ? '0' + (now.getMonth() + 1).toString() : now.getMonth() + 1;
+          date = now.getDate().toString().length === 1 ? '0' + (now.getDate()).toString() : now.getDate();
+          hours = now.getHours().toString().length === 1 ? '0' + now.getHours().toString() : now.getHours();
+          minutes = now.getMinutes().toString().length === 1 ? '0' + now.getMinutes().toString() : now.getMinutes();
+          seconds = now.getSeconds().toString().length === 1 ? '0' + now.getSeconds().toString() : now.getSeconds();
+          formattedDateTime = year + '-' + month + '-' + date + 'T' + hours + ':' + minutes + ':' + seconds;
+
+          if ((onlyBlank === true) && $(this).val()) {
+            return this;
+          }
+          $(this).val(formattedDateTime);
+          return this;
+        }
+
+        $(function () {
+            $('input[type="datetime-local"]').setNow();
+        });
+        
       </script>
     ]]>
   </configuration_panel>
@@ -85,7 +102,7 @@
         <configuration key="number-of-events">${numberOfEvents}</configuration>
         <configuration key="set-series-id">${seriesId}</configuration>
         <configuration key="set-title">${mpTitle}</configuration>
-        <configuration key="set-start-date-time">${startDate} ${startTime}</configuration>
+        <configuration key="set-start-date-time">${startDateTime}</configuration>
         <configuration key="target-tags"></configuration>
         <configuration key="property-namespaces">org.opencastproject.assetmanager.security,org.opencastproject.workflow.configuration</configuration>
         <configuration key="no-suffix">${noCopySuffix}</configuration>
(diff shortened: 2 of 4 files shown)


modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -227,7 +246,7 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
 
     SeriesInformation series = null;
     AccessControlList seriesAccessControl = null;
-    if (!seriesId.isEmpty()) {
+    if (!seriesId.isEmpty() && !seriesId.startsWith("${")) {
```
gregorydlogan: This doesn't check that the substitution terminates.  Is that expected?

gregorydlogan (approved):
LGTM, aside from my comment.

docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
```diff
@@ -16,8 +16,8 @@ For each duplicated event the new media package ID is stored as a workflow prope
 
 |Name                                    |Example                                                             |Description                                    |
 |----------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|
-|duplicate\_media\_package\_*number*\_id |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
-|duplicate\_media\_package\_ids          |`duplicate_media_package_ids=e72f2265-472a-49ae-bc04-8301d94b4b1a, a32e2265-472a-49ae-bc04-8351d94b4b1c` | comma separated list of Media package IDs of the duplicated event |
+|duplicate_media_package_*number*_id     |`duplicate_media_package_1_id=e72f2265-472a-49ae-bc04-8301d94b4b1a` |Media package ID of the duplicated event       | \*
```
lkiesow, line 19: Why did you remove the escape character on the underscore character?

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
lkiesow: The `.startsWith("${")` shouldn't be in here. If that's a value, it should be used

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+      final Date startDate;
       try {
         String newMpId = workflowInstance.getConfiguration("newMpId");
         if (newMpId == null) {
           newMpId = UUID.randomUUID().toString();
         }
         // Clone the media package (without its elements)
-        newMp = copyMediaPackage(mediaPackage, series, newMpId, noSuffix, i + 1, copyNumberPrefix);
+        String useTitle;
+        if (title.isEmpty() || (title.startsWith("${") && title.endsWith("}"))) {
```
lkiesow: Same here, the `(title.startsWith("${") && title.endsWith("}"))` is weird

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+          useTitle = mediaPackage.getTitle();
```
lkiesow: This method is deprecated. It's not guaranteed to yield the correct title. Get metadata from the DC catalog

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+        }
+        final String newTitle = noSuffix
+            ? useTitle
+            : String.format("%s (%s %d)", useTitle, copyNumberPrefix, i + 1);
+        Date mpDate = mediaPackage.getDate();
+        String date = new SimpleDateFormat("yyyy-MM-dd").format(mpDate);
+        String time = new SimpleDateFormat("HH:mm:ss").format(mpDate);
+        if (!startDateString.isEmpty()) {
+          String [] dt = startDateString.trim().split("\\s+");
+          if (dt.length >= 2) {
```
lkiesow: If `length > 2` that's clearly an error and you shouldn't guess values but throw an error

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+            if (!dt[0].startsWith("${")) {
+              date = dt[0];
+            }
+            if (!dt[1].startsWith("${")) {
+              time = (dt[1].split(":").length == 2) ? dt[1] + ":00" : dt[1];
```
lkiesow: The documentation above states `HH:MM:SS` as the format, this allows `HH:MM:SS` and `HH:MM`.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+            }
+          }
+        }
+        try {
+          mpDate = ADMIN_UI_DATE_FORMAT.parse(date + " " + time);
+        } catch (ParseException ex) {
+          logger.info("{} could not be parsed as Date", date + " " + time);
```
lkiesow: This should fail the operation instead of doing something unexpected

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+        }
+        logger.info("setting StartDate to {}", mpDate.toString());
```
lkiesow:
```suggestion
        logger.info("setting StartDate to {}", mpDate);
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
+        startDate = mpDate;
```
lkiesow: Do we need both variables?

etc/workflows/duplicate-event.xml
```diff
@@ -12,18 +12,48 @@
...
+          </div>
+          <br/>
+          <div>
+            <label for="seriesId" style="width: 150px;" >Series Id:</label>
+            <input id="seriesId" name="seriesId" type="text" maxlength="255" style="width: 400px;" value=" " class="configField" />
+          </div>
+          <br/>
+          <div>
+            <label for="startDate" style="width: 150px;">StartDate (YYYY-MM-DD):</label>
+            <input id="startDate" name="startDate" type="text" style="width: 200px;" value=" " class="configField" />
```
lkiesow: Use [`<input type="date">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)

etc/workflows/duplicate-event.xml
```diff
@@ -12,18 +12,48 @@
...
+          </div>
+          <div>
+            <label for="startTime" style="width: 150px;">startTime (HH:MM:SS):</label>
+            <input id="startTime" name="startTime" type="text" style="width: 200px;" value=" " class="configField" />
```
lkiesow: …and [`type=time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time)

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -309,14 +328,45 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, fi
...
-        newMp = copyMediaPackage(mediaPackage, series, newMpId, noSuffix, i + 1, copyNumberPrefix);
+        String useTitle;
+        if (title.isEmpty() || (title.startsWith("${") && title.endsWith("}"))) {
+          useTitle = mediaPackage.getTitle();
+        } else {
+          useTitle = title;
+        }
+        final String newTitle = noSuffix
+            ? useTitle
+            : String.format("%s (%s %d)", useTitle, copyNumberPrefix, i + 1);
```
lkiesow: I did not go through all the code but this seems a complex way of:

```suggestion
        if (title.isEmpty()) {
          title = mediaPackage.getTitle();
        }
        if (!noSuffix) {
          title = String.format("%s (%s %d)", title, copyNumberPrefix, i + 1);
        }
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -516,11 +565,16 @@ private MediaPackage copyDublinCore(
...
       final List<String> addTags,
       final List<String> overrideTags,
-      final List<URI> temporaryFiles) throws WorkflowOperationException {
+      final List<URI> temporaryFiles,
+      final Date creationDate
+  ) throws WorkflowOperationException {
     final DublinCoreCatalog destinationDublinCore = DublinCoreUtil.loadEpisodeDublinCore(workspace, source).get();
     destinationDublinCore.setIdentifier(null);
     destinationDublinCore.setURI(sourceDublinCore.getURI());
+    destinationDublinCore.set(DublinCore.PROPERTY_CREATED, OpencastMetadataCodec.encodeDate(creationDate,Precision.Second));
```
lkiesow:
```suggestion
    destinationDublinCore.set(DublinCore.PROPERTY_CREATED, OpencastMetadataCodec.encodeDate(creationDate, Precision.Second));
```

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
```diff
@@ -516,11 +565,16 @@ private MediaPackage copyDublinCore(
...
     destinationDublinCore.set(DublinCore.PROPERTY_TITLE, destination.getTitle());
+    DublinCoreValue eventTime = EncodingSchemeUtils.encodePeriod(new DCMIPeriod(creationDate, creationDate), Precision.Second);
```
lkiesow: This essentially sets the event date to 0.
If we don't have a period, we shouldn't set it

lkiesow (changes requested):

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: When a series is not provided by the user, `${seriesId}` in the workflow definition is not substituted. So without that check the series will literally be set as `${seriesId}` in such instance.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: Same answer as above

docs/guides/admin/docs/workflowoperationhandlers/duplicate-event-woh.md
tyfranck: No idea why Tobias removed the escape character but looking at it closing, I can see that escape characters are not really needed in the name. There's no need to escape underscore character in the name.  So I guess it's a bit of clean up.

etc/workflows/duplicate-event.xml
tyfranck: Yeah, I used date as a type before but the trouble is you can't really set a default value as `value =""` (empty). 
So when the date is not set, a null value is sent for that property which in turn throws an exception.
Properties values are enforced in `org.opencastproject.assetmanager.impl.persistence` package by the following code:

```
  public Property toProperty() {
    final PropertyId id = new PropertyId(mediaPackageId, namespace, propertyName);
    if (stringValue != null) {
      return new Property(id, Value.mk(stringValue));
    } else if (dateValue != null) {
      return new Property(id, Value.mk(dateValue));
    } else if (longValue != null) {
      return new Property(id, Value.mk(longValue));
    } else if (boolValue != null) {
      return new Property(id, Value.mk(boolValue));
    } else {
      throw new RuntimeException("Bug. At least one of the value columns must be non null.");
    }
  }
```
Basically input of type date or time can't be used unless a default value is set. Or maybe you have some suggestions.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: OK. That method needs @Deprecated Annotation if it can't been used anymore.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: Documentation updated to mention both format

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: code simplified :-)

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

tyfranck:
I've addressed most of the comments given my time. @lkiesow  I commented the one I didn't address. I don't see them as stopper but I let you judge.  Any suggestions welcome.

etc/workflows/duplicate-event.xml
mliradelc: In that case, where the time is null, I suggest using the current time and log a warning for the event.

etc/workflows/duplicate-event.xml
mliradelc: Same as above answer

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
mliradelc: Missing space here

lkiesow:
@mliradelc if you are happy with the current state, feel free to merge.
We start the release process right now but still need some time to prepare the release notes.

mliradelc (changes requested):
Hi @tyfranck I just reviewed your code, is good to merge.
 
There are some things that can be improved:
- Use the DC catalog instead of deprecated method
- When the time and date values are missing, use the actual time and date, then log a warning.
- And a missing space.

I added the comments that @lkiesow has pointed out.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed with code revamped

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

etc/workflows/duplicate-event.xml
tyfranck: fixed, current datetime set default in UI.

etc/workflows/duplicate-event.xml
tyfranck: fixed with code revamped.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/DuplicateEventWorkflowOperationHandler.java
tyfranck: fixed

tyfranck:
Hi @mliradelc, 
Thanks for taking a look. With the latest update to this PR, I've addressed your points raised above.
1- Using DC to get mediapackage Title. 
2- Current date time is set as default in UI
3- Space removed.
4- Documentation updated accordingly to reflect new changes.

tyfranck:
Any hint on why this is failing `Test configuration` run ?

etc/workflows/duplicate-event.xml
```diff
@@ -35,6 +60,28 @@
...
+            return this;
+          }
+          $(this).val(formattedDateTime);
+          return this;
+        }
+
+        $(function () {
+            $('input[type="datetime-local"]').setNow();
+        });
+
```
mliradelc: Here is the error, you need to delete the trailing spaces, that's why is failing the test

`etc/workflows/duplicate-event.xml:84`

mliradelc (changes requested):

etc/workflows/duplicate-event.xml
tyfranck: Thanks. Updated now.

mliradelc (approved):

PR #3810, opened 2022-05-30
No Elasticsearch Use In Ingest

lkiesow:
The ingest service searches for series titles starting with pull
request #3586.  This causes trouble when combined with the removal of
the Solr index which now makes this code fall back to the Elasticsearch
index instead.

Unfortunately, the Elasticsearch modules are not included in the ingest
distribution (#3797) which will cause the distribution to fail.

Simply including it works in theory, but causes trouble since our
infrastructure documentation states that Elasticsearch may be available
on the admin only and can thus, for example, be run with no encryption
or authentication to make deployments easier. All this would need to
change, making this quite messy.

This patch fixes the issue by using a generated series identifier for
which we need to full text search instead.

This closes #3797

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/releasenotes/auto-generated-capture-agent-series.txt
+++ b/docs/guides/admin/docs/releasenotes/auto-generated-capture-agent-series.txt
@@ -1,0 +1,2 @@
+Identifiers for auto-generated capture series are now generated slightly different. This may cause new series to be
+generated for capture agents in some cases.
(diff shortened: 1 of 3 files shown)


KatrinIhler (approved):
I like this version better, actually. But now I might have to migrate old ca series... T_T

gregorydlogan:
Seems to work, code looks good -> merged.

PR #3811, opened 2022-05-30
Remove red check mark when editing metadata

### LINKED ISSUE #3738: Red check mark when editing metadata is ambigious
**Describe the bug**
A clear and concise description of what you see:
When editing meta data in the admin UI, a red check mark is shown on changed (dirty/unwritten) fields. Red is not distinguishable from green for some people, and the check mark as an indicator for unsaved changes is not obvious.

image

**To Reproduce**
1. Go to https://stable.opencast.org/
2. Click on an episode's metadata icon image
3. Edit a metadata field
4. See red check mark image

**Environment Information**
 - Opencast version with minor release Opencast 12.0.0.SNAPSHOT - 2ce87d6 - built on 2022-05-04

lkiesow:
When editing meta data in the admin UI, a red check mark is shown on
changed (dirty/unwritten) fields. Red is not distinguishable from green
for some people, and the check mark as an indicator for unsaved changes
is not obvious.

Since the check marks do not mean anything anymore except that the user
finished their entry (which they should obviously know), this pull
request removes the check marks entirely.

If users try leaving the dialog with unsaved changes, a warning is
displayed anyway.

This fixes #3738

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/partials/editableBooleanValue.html
+++ b/modules/admin-ui-frontend/app/scripts/shared/partials/editableBooleanValue.html
@@ -1,4 +1,5 @@
-<div ng-form="innerForm">
+<div ng-form="innerForm"
+     ng-class="{ dirty: params.dirty }">
   <i class="edit fa fa-pencil-square"></i>
 
   <input type="checkbox"
--- a/modules/admin-ui-frontend/app/scripts/shared/partials/editableBooleanValue.html
+++ b/modules/admin-ui-frontend/app/scripts/shared/partials/editableBooleanValue.html
@@ -1,8 +1,9 @@
 <div ng-form="innerForm">
   <i class="edit fa fa-pencil-square"></i>
-  <i class="changed fa fa-check" ng-class="{ saved: params.saved, dirty: params.dirty }"></i>
 
   <input type="checkbox"
-         ng-model="params.value" ng-change="submit()"
-                                 ng-required="params.required" name="{{ params.name }}"/>
+         ng-model="params.value"
+         ng-change="submit()"
+         ng-required="params.required"
+         name="{{ params.name }}"/>
 </div>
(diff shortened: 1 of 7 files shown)


mtneug:
The opinion in the dev meeting seems to be to have some way of identifying fields that changed before hitting save. Is the color important here as this only indicates changed fields?

KatrinIhler (dismissed):
I actually don't agree with removing this entirely, I think it's good to know which fields are gonna be changed once I hit save, if I edited multiple. You're of course right that red and green are indistinguishable for some people though (although I don't think the color matters terribly in this case). For more accessibility, I would suggest to also change the icon for the red marks instead.

lkiesow:
Personally, I would also say that red is usually a color indicating an error or danger. That's not the case here.

And really, you cannot remember what fields you edited likely in the last 60 seconds? Plus, that's not what the checkbox does right now. If that was the intention for this, it's already completely broken since it e.g. doesn't indicate if you remove a value.

KatrinIhler:
I'm also fine with changing the color, I'm not at all attached to red, and you're right that it probably isn't the greatest color for this. Maybe blue instead?

I think it's nice if you edit a field by mistake, or you get interrupted while editing, or you do a lot of changes all at once...

This was definitely the intention for the check mark, and if I remember correctly, this used to work. It should definitely also show when removing values. If it doesn't, it is indeed broken. :/

lkiesow:
I've changed the patch. It does no longer remove the modification indicator from the metadata editor of the admin interface, but uses a neutral one instead. This uses the existing mechanism to detect changes – it doesn't improve the algorithm, nor does it make it worse – but switches to a neutral indicator for identifying changes from the red (Why red? Is it an error) check mark (Why a check mark? Do we want to indicate success?) it was before.

``
Screenshot from 2022-06-05 14-46-26

gregorydlogan:
This seems like a reasonable change to me.  Gets the point across without using colour, or checkmarks.

KatrinIhler (approved):
I'm fine with this change.

KatrinIhler:
> (Why red? Is it an error) check mark (Why a check mark? Do we want to indicate success?)

The answer to both of those questions is "Adding a save button for metadata was a very rushed last-minute change before a feature freeze iirc and I was so happy it worked that I didn't really consider the details" :P Keeping the check mark and simply changing the color (and red = unsaved) seemed like the easiest choice. So please, don't blame me anymore T_T

PR #3812, opened 2022-05-30
Fix rest services are unavailable under certain circumstances

geichelberger:
This PR fixes the problem that the JAX-RS rest services are not available after an endpoint got destroyed.

In the current implementation stopping an endpoint shuts down the bus and the server, which causes the rest services to be unavailable. (Check logs with: `log:set TRACE org.apache.cxf`)
The solution is to recreate the JAX-RS server after any changes (e.g., adding/removing endpoints or providers ...)

You can observe the same phenomena if you change the pax web config during runtime (e.g., changing Listen-IP from 127.0.0.1 to 0.0.0.0). In that case, the OSGi HttpService gets recreated and tears down all HttpServlets and the rest endpoints. This patch only fixes the recreation of the rest endpoints. The non-Declarative Services Servlets will still be unavailable. e. g. the OAI-PMH servlet.

By registering the bus as OSGi Service, we can use the Karaf cxf commands `cxf:list-busses` and `cxf:list-endpoints`. I think that's only useful for debugging :) 

Fixes rest endpoint problem of #3218

PS: OSGi R7 supports the JAX-RS whiteboard pattern http://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.jaxrs.html

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/rest/RestPublisher.java
@@ -62,6 +62,10 @@ import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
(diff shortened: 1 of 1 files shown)


lkiesow:
Seems to do the trick. Haven't gone through the code changes.

gregorydlogan:
Tests correctly here, code looks good -> merged.

PR #3816, opened 2022-05-31
Removing unnecessary workflow index data from rebuild

Alaschinger:
Instead of reading the whole workflow instance from database when rebuilding the elasticsearch index only four attributes will be used:
- workflow id (long)
- state (int)
- mediapackage id (String)
- organization id (String

They are stored within a new entity object "WorkflowIndexData".

To reduce the database access time the WorkflowIndexData is only sorted by date_created and not filtered within the sql query.
The latest workflows are selected by mapping the date-sorted WorkflowIndexData with the mediapackage id as key value within the repopulate function.

Some attributes and function calls which added additional metadata while rebuilding the workflow index have been removed from the workflowServiceImpl.

### DIFF
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
@@ -39,7 +39,7 @@ import javax.persistence.Table;
 @NamedNativeQueries({
         @NamedNativeQuery(
                 name = "WorkflowIndexData.getAll",
-                query = "SELECT id, state, mediapackage_id, organization_id FROM oc_workflow ORDER BY date_created ASC",
+                query = "SELECT id, state, mediapackage_id, organization_id FROM oc_workflow ORDER BY mediapackage_id, id DESC",
                 resultSetMapping = "DataResult"
         ),
 })
@@ -51,7 +51,7 @@ import javax.persistence.Table;
                         fields = {
                                   @FieldResult(name = "id",column = "id"),
                                   @FieldResult(name = "state", column = "state"),
-                                  @FieldResult(name = "mediaPackageId", column = "mediaPackage_id"),
+                                  @FieldResult(name = "mediaPackageId", column = "mediapackage_id"),
                                   @FieldResult(name = "organizationId", column = "organization_id")
                         }
                 )
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
@@ -1,0 +1,99 @@
+/**
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.opencastproject.workflow.api;
+
+import javax.persistence.Entity;
+import javax.persistence.EntityResult;
+import javax.persistence.FieldResult;
+import javax.persistence.Id;
+import javax.persistence.NamedNativeQueries;
+import javax.persistence.NamedNativeQuery;
+import javax.persistence.SqlResultSetMapping;
+import javax.persistence.Table;
+
+/**
+ * Summary of the workflow stats necessary for index rebuild,
+ * for use in repopulate and update (immutable)
+ */
+@Table(name = "oc_workflow")
+@Entity
+@NamedNativeQueries({
+        @NamedNativeQuery(
+                name = "WorkflowIndexData.getAll",
+                query = "SELECT id, state, mediapackage_id, organization_id FROM oc_workflow ORDER BY date_created ASC",
+                resultSetMapping = "DataResult"
+        ),
+})
+@SqlResultSetMapping(
+        name = "DataResult",
+        entities = {
+                @EntityResult(
+                        entityClass = WorkflowIndexData.class,
+                        fields = {
+                                  @FieldResult(name = "id",column = "id"),
+                                  @FieldResult(name = "state", column = "state"),
+                                  @FieldResult(name = "mediaPackageId", column = "mediaPackage_id"),
+                                  @FieldResult(name = "organizationId", column = "organization_id")
+                        }
+                )
+})
+
+
+public class WorkflowIndexData {
+  @Id
+  private Long id;
+  private int state;
+  private String mediaPackageId;
+  private String organizationId;
+
+
+  /**
+   * Default constructor without any import.
+   */
+  public WorkflowIndexData() {
+
+  }
+
+  public WorkflowIndexData(Long id, int state, String mediaPackageId, String organizationId) {
+    this.id = id;
+    this.state = state;
+    this.mediaPackageId = mediaPackageId;
+    this.organizationId = organizationId;
+  }
+
+  public Long getId() {
+    return id;
+  }
+
+  public int getState() {
+    return state;
+  }
+
+  public String getMediaPackageId() {
+    return mediaPackageId;
+  }
+
+  public String getOrganizationId() {
+    return organizationId;
+  }
+
+}
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
@@ -78,15 +78,19 @@ public interface WorkflowServiceDatabase {
    */
   long countWorkflows(WorkflowInstance.WorkflowState state) throws WorkflowDatabaseException;
 
-    /**
+  /**
    * Gets workflow index data for all events.
    * Selects only workflow id, state, mediapackage id and organization id
    *
-   * @return list of all {@link WorkflowIndexData}s
+   * @param limit
+   *          max number of data objects to be returned
+   * @param offset
+   *          only return data from this point onwards
+   * @return list of {@link WorkflowIndexData}s
    * @throws WorkflowDatabaseException
    *           if there is a problem communicating with the underlying data store
    */
-  List<WorkflowIndexData> getWorkflowIndexData() throws WorkflowDatabaseException;
+  List<WorkflowIndexData> getWorkflowIndexData(int limit, int offset) throws WorkflowDatabaseException;
 
   /**
    * Returns the number of events workflows have been run on.
--- a/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
+++ b/modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
@@ -78,19 +78,15 @@ public interface WorkflowServiceDatabase {
    */
   long countWorkflows(WorkflowInstance.WorkflowState state) throws WorkflowDatabaseException;
 
-  /**
-   * Gets latest workflow instances for all events.
-   * Selects workflows which haven't been finished (DateCompleted IS NULL) or greatest DateCompleted
+    /**
+   * Gets workflow index data for all events.
+   * Selects only workflow id, state, mediapackage id and organization id
    *
-   * @param limit
-   *          max number of workflows to be returned
-   * @param offset
-   *          only return workflows from this point onwards
-   * @return list of all {@link WorkflowInstance}s
+   * @return list of all {@link WorkflowIndexData}s
    * @throws WorkflowDatabaseException
    *           if there is a problem communicating with the underlying data store
    */
-  List<WorkflowInstance> getLatestWorkflowInstances(int limit, long token) throws WorkflowDatabaseException;
+  List<WorkflowIndexData> getWorkflowIndexData() throws WorkflowDatabaseException;
 
   /**
    * Returns the number of events workflows have been run on.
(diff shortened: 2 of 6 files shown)


modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2296,39 +2271,24 @@ private void removeWorkflowInstanceFromIndex(long workflowInstanceId, Elasticsea
...
     final String orgId = securityService.getOrganization().getId();
     final User user = securityService.getUser();
 
     logger.debug("Updating workflow instance {} of event {} in the {} index.", workflowInstanceId, eventId,
             index.getIndexName());
     Function<Optional<Event>, Optional<Event>> updateFunction = (Optional<Event> eventOpt) -> {
       Event event = eventOpt.orElse(new Event(eventId, orgId));
-      event.setCreator(user.getName());
       event.setWorkflowId(workflowInstanceId);
-      event.setWorkflowDefinitionId(workflowInstance.getTemplate());
```
lkiesow: Is it okay to not set these two?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2156,75 +2153,53 @@ public Map<String, Map<String, String>> getWorkflowStateMappings() {
...
-          // don't try this for terminated workflows since the ACLs are no longer in the working file repository and
-          // they will be overwritten later in the re-indexing process by ACLs from the asset manager anyway.
-          final AccessControlList accessControlList;
-          if (instance.getState().isTerminated()) {
-            accessControlList = new AccessControlList();
-          } else {
-            accessControlList = authorizationService.getActiveAcl(instance.getMediaPackage()).getA();
-          }
-
           SecurityUtil.runAs(securityService, organization,
```
lkiesow: If we don't pass modify the creator, I don't think we need to change the security context.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2156,75 +2153,53 @@ public Map<String, Map<String, String>> getWorkflowStateMappings() {
...
           current += 1;
           Organization organization = null;
           try {
-            organization = organizationDirectoryService.getOrganization(instance.getOrganizationId());
+            organization = organizationDirectoryService.getOrganization(data.getOrganizationId());
           } catch (NotFoundException e) {
-            logger.error("Found workflow with non-existing organization {}", instance.getOrganizationId());
+            logger.error("Found workflow with non-existing organization {}", data.getOrganizationId());
             continue;
           }
```
lkiesow: If we don't need to switch the security context below, we also wouldn't need to retrieve the organization here but could just pass on the organization identifier.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2156,75 +2153,53 @@ public Map<String, Map<String, String>> getWorkflowStateMappings() {
...
+        workflowIndexData = persistence.getWorkflowIndexData();
+      } catch (WorkflowDatabaseException e) {
+        logIndexRebuildError(logger.getSlf4jLogger(), index.getIndexName(), e);
+        throw new IndexRebuildException(index.getIndexName(), getService(), e);
+      }
+
+      Map<String, WorkflowIndexData> latestWorkflows = new HashMap<>();
+      for (WorkflowIndexData data : workflowIndexData) {
+        latestWorkflows.put(data.getMediaPackageId(), data);
+      }
```
lkiesow: This works assuming the newest workflow for a given media package comes last.
We could be smarter about it if we reverse the order and then only act whenever we get a new media package id without storing this in a map, overwriting the values again and again.

```
mpid  id
a     5
a     1
b     2
c     4
c     3
```

and the something like:

```py
last_mp = ""
for mp, id in data:
  if mp == last_mp:
    continue
  last_mp = mp
  update_index(data)
```

the SQL query for this would ge something like:

```sql
select id, mediapackage_id, … from oc_workflow order by mediapackage_id, id desc;
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Summary of the workflow stats necessary for index rebuild,
+ * for use in repopulate and update (immutable)
+ */
+@Table(name = "oc_workflow")
+@Entity
+@NamedNativeQueries({
+        @NamedNativeQuery(
+                name = "WorkflowIndexData.getAll",
+                query = "SELECT id, state, mediapackage_id, organization_id FROM oc_workflow ORDER BY date_created ASC",
```
lkiesow: Maybe instead something like:
```sql
select id, mediapackage_id, … from oc_workflow order by mediapackage_id, id desc;
```
See comment on code…

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
Alaschinger: Well as far as we know it is. The future will show if anything unexpected goes wrong.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowIndexData.java
```diff
@@ -0,0 +1,99 @@
...
+})
+@SqlResultSetMapping(
+        name = "DataResult",
+        entities = {
+                @EntityResult(
+                        entityClass = WorkflowIndexData.class,
+                        fields = {
+                                  @FieldResult(name = "id",column = "id"),
+                                  @FieldResult(name = "state", column = "state"),
+                                  @FieldResult(name = "mediaPackageId", column = "mediaPackage_id"),
```
lkiesow: Shouldn't matter on most databases, but…
```suggestion
                                  @FieldResult(name = "mediaPackageId", column = "mediapackage_id"),
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
```diff
@@ -78,19 +78,19 @@ List<WorkflowInstance> getWorkflowInstancesForCleanup(WorkflowInstance.WorkflowS
    */
   long countWorkflows(WorkflowInstance.WorkflowState state) throws WorkflowDatabaseException;
 
-  /**
-   * Gets latest workflow instances for all events.
-   * Selects workflows which haven't been finished (DateCompleted IS NULL) or greatest DateCompleted
+    /**
```
lkiesow: Indentation
```suggestion
  /**
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabase.java
```diff
@@ -78,19 +78,19 @@ List<WorkflowInstance> getWorkflowInstancesForCleanup(WorkflowInstance.WorkflowS
...
+   * Selects only workflow id, state, mediapackage id and organization id
    *
    * @param limit
-   *          max number of workflows to be returned
+   *          max number of data objects to be returned
    * @param offset
-   *          only return workflows from this point onwards
-   * @return list of all {@link WorkflowInstance}s
+   *          only return data from this point onwards
+   * @return list of all {@link WorkflowIndexData}s
```
lkiesow:
```suggestion
   * @return list of {@link WorkflowIndexData}s
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/WorkflowServiceDatabaseImpl.java
```diff
@@ -200,22 +200,17 @@ public long countWorkflows(WorkflowInstance.WorkflowState state) throws Workflow
...
 
-      var query = em.createNamedQuery("Workflow.findLatest", WorkflowInstance.class);
-
-      String orgId = securityService.getOrganization().getId();
-      query.setParameter("organizationId", orgId);
-      query.setParameter("startToken", token);
-      query.setMaxResults(limit);
-      logger.debug("Requesting latest workflows using query: {}", query);
-      return query.getResultList();
+      var nativeQuery = em.createNamedQuery("WorkflowIndexData.getAll",WorkflowIndexData.class);
```
lkiesow:
```suggestion
      var nativeQuery = em.createNamedQuery("WorkflowIndexData.getAll", WorkflowIndexData.class);
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1260,8 +1250,14 @@ public void update(final WorkflowInstance workflowInstance) throws WorkflowExcep
         // updates for running operations since we updated the metadata right before these operations and will do so
         // again right after those operations.
         if (op == null || op.getState() != OperationState.RUNNING) {
-          updateWorkflowInstanceInIndex(workflowInstance, accessControlList, episodeDublinCoreCatalog,
-                  elasticsearchIndex);
+          // Collect necessary information only for index update
+          long id = workflowInstance.getId();
+          int state = workflowInstance.getState().ordinal();
+          String mpId = workflowInstance.getMediaPackage().getIdentifier().toString();
+          String orgId = workflowInstance.getOrganizationId().toString();
```
lkiesow: Isn't this a string already?
```suggestion
          String orgId = workflowInstance.getOrganizationId();
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1260,8 +1250,14 @@ public void update(final WorkflowInstance workflowInstance) throws WorkflowExcep
...
+
+          WorkflowIndexData latestWorkflow = new WorkflowIndexData(id, state, mpId, orgId);
+          updateWorkflowInstanceInIndex(latestWorkflow, elasticsearchIndex);
```
lkiesow: Does it make sense to pass the four values to `updateWorkflowInstanceInIndex(…)` directly instead of wrapping this in a new Java object? The first thing which happens in that function is that all values are unpacked again. We wouldn't need to do that if we pass them on as arguments in the first place.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2163,68 +2159,44 @@ public void repopulate(final ElasticsearchIndex index) throws IndexRebuildExcept
...
-          try {
-            organization = organizationDirectoryService.getOrganization(instance.getOrganizationId());
-          } catch (NotFoundException e) {
-            logger.error("Found workflow with non-existing organization {}", instance.getOrganizationId());
+        for (WorkflowIndexData data : workflowIndexData) {
+          currentMediapackageId = data.getMediaPackageId();
+          if (currentMediapackageId.equals(lastMediapackageId)) {
             continue;
           }
+          else {
```
lkiesow: Else is unnecessary since you already terminate the loop with continue. Else invert the confition and do not use continue.
```suggestion
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2163,68 +2159,44 @@ public void repopulate(final ElasticsearchIndex index) throws IndexRebuildExcept
...
+            logIndexRebuildProgress(logger.getSlf4jLogger(), index.getIndexName(), total, current);
           }
-
-          SecurityUtil.runAs(securityService, organization,
-                  SecurityUtil.createSystemUser(componentContext, organization), () -> {
-                    updateWorkflowInstanceInIndex(instance, accessControlList, episodeDublinCoreCatalog, index);
-                  });
-          logIndexRebuildProgress(logger.getSlf4jLogger(), index.getIndexName(), total, current);
         }
       } while (current < total);
```
lkiesow: I just noticed that this is dangerous since new workflows could be added while this is running.
A better termination criteria would be something like:
```suggestion
      } while (workflowIndexData.size() <= 0);
```

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2296,39 +2268,24 @@ private void removeWorkflowInstanceFromIndex(long workflowInstanceId, Elasticsea
...
    */
-  private void updateWorkflowInstanceInIndex(WorkflowInstance workflowInstance, AccessControlList accessControlList,
-          DublinCoreCatalog episodeDublincoreCatalog, ElasticsearchIndex index) {
-    final long workflowInstanceId = workflowInstance.getId();
-    final String eventId = workflowInstance.getMediaPackage().getIdentifier().toString();
+  private void updateWorkflowInstanceInIndex(WorkflowIndexData latestWorkflow, ElasticsearchIndex index) {
+    final long workflowInstanceId = latestWorkflow.getId();
+    final WorkflowState workflowState = WorkflowState.values()[latestWorkflow.getState()];
+    final String eventId = latestWorkflow.getMediaPackageId();
     final String orgId = securityService.getOrganization().getId();
```
lkiesow: We removed `runAs(…)`. Therefore, this is no longer correct. Get the organization from `latestWorkflow`.

lkiesow (changes requested):
Found mostly minor things…

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
Alaschinger: Yes, it is. This type conversion was necessary in an earlier version of the code and should be removed now.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
Alaschinger: That's right. I have changed `updateWorkflowInstanceInIndex(…)` now, so that the values don't need to be wrapped and then unwrapped in an object anymore.

lkiesow:
Given 500 events and 1000 workflows, with this we drop from the 50% rebuild time we got from #3775 to only about 32% of the original time of the 12.x baseline. This speedup now also should be less affected by the ratio of workflows to events, but should be a mostly unconditional speedup:

rebuild

So, this looks very promising. I still want to do a final code review, though.

It would also be interesting to see how much difference there is if no workflows can be filtered at all. It's likely the 18% difference, but would be interesting to confirm that with a dataset of 500 events and one workflow per event.

PR #3832, opened 2022-06-02
Use Local Index References

lkiesow:
Services now track their index on their own and use it to push updates
into the index if things change. For example, when a new workflow is
started.

For reindexing, an external index was still passed to the service
though. That's pretty weird since that could mean that the same service
would write to completely different indexes in different situations,
but expecting that the data would end up at the same place.

In practice this did not really matter since the same index was passed
to the `repopulate()` method which was tracked internally anyway. Still,
it makes more sense to always use the tracked index.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
+++ b/modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
@@ -117,7 +117,7 @@ public class SeriesServiceImpl extends AbstractIndexProducer implements SeriesSe
   private String systemUserName;
 
   /** The API index */
-  private ElasticsearchIndex elasticsearchIndex;
+  private ElasticsearchIndex index;
 
   private AclServiceFactory aclServiceFactory;
 
@@ -158,7 +158,7 @@ public class SeriesServiceImpl extends AbstractIndexProducer implements SeriesSe
   /** OSGi callbacks for setting the API index. */
   @Reference
   public void setElasticsearchIndex(ElasticsearchIndex index) {
-    this.elasticsearchIndex = index;
+    this.index = index;
   }
 
   @Reference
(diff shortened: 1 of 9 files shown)


modules/themes/src/main/java/org/opencastproject/themes/persistence/ThemesServiceDatabaseImpl.java
```diff
@@ -329,12 +329,7 @@ private ThemeDto getThemeDto(long id, EntityManager em) {
   }
 
   @Override
-  public void repopulate(final ElasticsearchIndex index) {
-    if (index.getIndexName() != this.index.getIndexName()) {
-      logger.info("Themes are currently not part of the {} index, no re-indexing necessary.");
-      return;
-    }
```
Alaschinger, line 336: Will this case not happen any more so that themes are always written to the index or is it handled elsewhere so that we don't need this option here?

gregorydlogan (approved):
Looks good to me!

lkiesow:
Rebased to fix merge conflict

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -198,7 +198,7 @@ public DublinCoreCatalog updateSeries(DublinCoreCatalog dc) throws SeriesExcepti
 
         logger.debug("Updating series {}", id);
         // update API index
-        updateSeriesMetadataInIndex(id, elasticsearchIndex, dublinCore);
+        updateSeriesMetadataInIndex(id, index, dublinCore);
```
KatrinIhler: I guess we could kick the index out of these methods as well, but it's not strictly necessary.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2278,7 +2278,7 @@ private void removeWorkflowInstanceFromIndex(long workflowInstanceId, Elasticsea
    * @param index
    *         the index to update
    */
-  private void updateWorkflowInstanceInIndex(long id, int state, String mpId, String orgId, ElasticsearchIndex index) {
+  private void updateWorkflowInstanceInIndex(long id, int state, String mpId, String orgId) {
```
KatrinIhler, line 2280: This isn't really consistent with the other methods anymore (see `remove...`). I'd just remove the index from all of them.

KatrinIhler (approved):
Some nitpicking, but I'm fine with merging this.

lkiesow:
@KatrinIhler, updated the other methods as well.

KatrinIhler (approved):
Looks very nice now! :D

PR #3848, opened 2022-06-09
Adds a config key to set the number of speakers for amberscript trans…

marwyg:
Amberscripts default value of "number of speakers" is 2. In some cases, amberscripts recognizes 2 speakers, although there is only one, which results in a bad transcription file (1 word per line).

This PR adds the possibility to configure the "number of speakers". The configured value will be added to the REST call of the amberscript api.

### DIFF
--- a/modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
+++ b/modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
@@ -39,6 +39,7 @@ import org.opencastproject.mediapackage.MediaPackageException;
 import org.opencastproject.mediapackage.Track;
 import org.opencastproject.metadata.dublincore.DublinCore;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalog;
+import org.opencastproject.metadata.dublincore.DublinCoreValue;
 import org.opencastproject.metadata.dublincore.DublinCores;
 import org.opencastproject.security.api.DefaultOrganization;
 import org.opencastproject.security.api.Organization;
@@ -152,6 +153,12 @@ public class AmberscriptTranscriptionService extends AbstractJobProducer impleme
     StartTranscription
   }
 
+  // Enum for configuring which metadata field shall be used
+  // for determining the number of speakers of an event
+  private enum SpeakerMetadataField {
+    creator, contributor, both
+  }
+
   // service configuration keys
   private static final String ENABLED_CONFIG = "enabled";
   private static final String LANGUAGE = "language";
@@ -161,6 +168,7 @@ public class AmberscriptTranscriptionService extends AbstractJobProducer impleme
   private static final String DISPATCH_WORKFLOW_INTERVAL_CONFIG = "workflow.dispatch.interval";
   private static final String MAX_PROCESSING_TIME_CONFIG = "max.overdue.time";
   private static final String CLEANUP_RESULTS_DAYS_CONFIG = "cleanup.results.days";
+  private static final String SPEAKER_METADATA_FIELD = "speaker.metadata.field";
   private static final String LANGUAGE_FROM_DUBLINCORE = "language.from.dublincore";
   private static final String LANGUAGE_CODE_MAP = "language.code.map";
 
@@ -173,6 +181,8 @@ public class AmberscriptTranscriptionService extends AbstractJobProducer impleme
   private long workflowDispatchIntervalSeconds = 60;
   private long maxProcessingSeconds = 8 * 24 * 60 * 60; // maximum runtime for jobType perfect is 8 days
   private int cleanupResultDays = 7;
+  private final int defaultNumberOfSpeakers = 1;
+  private SpeakerMetadataField speakerMetadataField = SpeakerMetadataField.creator;
 
   /**
    * Contains mappings from several possible ways of writing a language name/code to the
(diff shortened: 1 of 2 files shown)


lkiesow:
Instead of having a global setting for the service, wouldn't it be better to have this as a configuration option at the workflow operation?

But even better than that, why not just use the number of creators/presenters set in the metadata of Opencast? In there you should already have the amount of participants or, if none are set, you can still always fall back to setting one.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -234,6 +236,16 @@ public void activate(ComponentContext cc) {
     }
     logger.info("Cleanup result files after {} days.", cleanupResultDays);
 
+    Option<String> numberOfSpeakersOpt = OsgiUtil.getOptCfg(cc.getProperties(), NUMBER_OF_SPEAKERS_CONFIG);
+    if (numberOfSpeakersOpt.isSome()) {
+      try {
+        numberOfSpeakers = Integer.parseInt(numberOfSpeakersOpt.get());
+      } catch (NumberFormatException e) {
+        logger.warn("Configured '{}' is invalid. Using default.", NUMBER_OF_SPEAKERS_CONFIG);
```
lkiesow: If the configuration is actually really broken, I think the service should refuse to accept it and fail and not fall back to a probably unexpected value.

Though my suggestion is to not have this setting anyway :D
See above

lkiesow (changes requested):

gregorydlogan:
Agree with Lars.  I find it profoundly odd that the *default* is two speakers.  Especially for our usecase, the majority of the time there will only be one speaker.  I would expect a default of one speaker, and then to use the number of presenters/creators to override.  You *might* need to add a config field to the amberscript config to be able to set which DC field to count though - since some institutions are probably using the creator field, and some are using presenter.  And no doubt, somewhere, someone is using *both* :)

marwyg:
I didn't want to change the Amberscript default, because I don't want people to run into some kind of problem 😨 But I like the idea of taking the number of speakers from the metadata. I will do it that way and set the fallback to 1.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
mliradelc: In my opinion, fail is too harsh for a problem like this, I agree with @marwyg to let it as default. From what I understand, it is a metadata value.

mliradelc:
I think it is a great improvement, but because the number of speakers can change from each event, Would it better to get that number from the number of presenters in the metadata? With "1" being the default.

lkiesow:
@marwyg, I converted this to a draft to make it easier to spot what to review for the upcoming release. Please convert it back once you made your change and it is ready to review again.

marwyg:
There is a configuration key now, where someone can define in which metadata field the operation shall look up the amount of speakers. You can define "presenter", "contributor", "both". If "both" is configured, the presenter and contributor list will be put together in a set, which removes duplicate Strings.

If there are no entries in the configured field, the default of '1' will be transmitted to the amberscript service.

marwyg:
https://github.com/opencast/opencast/pull/3848#discussion_r893973429

The new configuration will also use the default value, if the configured value is not right. I did it that way, because all other configurations in this operation work like this. If someone wants me to change this, maybe we should consider changing the other configurations as well? So the service won't start at all, if one of the values are not configured properly.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -271,16 +291,33 @@ public Job startTranscription(String mpId, Track track, String language, String
       jobtype = getAmberscriptJobType();
     }
 
+    int numberOfSpeakers = defaultNumberOfSpeakers;
+    if (track.getMediaPackage() != null) {
+      Set<String> speakers = new HashSet<>();
+      if (speakerMetadataField.equals(SpeakerMetadataField.presenter)
+          || speakerMetadataField.equals(SpeakerMetadataField.both)) {
+        speakers.addAll(Arrays.asList(track.getMediaPackage().getCreators()));
```
lkiesow: Weird mixture between terms. If you count the creators, the setting should also be named creator and not presenter.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -271,16 +291,33 @@ public Job startTranscription(String mpId, Track track, String language, String
...
+      }
+      if (speakerMetadataField.equals(SpeakerMetadataField.contributor)
+          || speakerMetadataField.equals(SpeakerMetadataField.both)) {
+        speakers.addAll(Arrays.asList(track.getMediaPackage().getContributors()));
```
lkiesow: The methods `getCreators` and `getContributors` are deprecated by now, since they may not necessarily yield accurate results (they will do so most of the time). It's potentially better to use the values from the dublincore catalog.

lkiesow (changes requested):

PR #3849, opened 2022-06-10
Amberscript configuration fix

CGreweling:
The configuration is set to the workflow filename but the id is needed by the code.

Change configured workflow filename to id in the configuration.


CGreweling:
I actually do not know if theres is a pr needed for this?

lkiesow:
Yes, but you could have merged it yourself if you argue that it's documentation… which I think it is.
Anyway… looks good merging.

PR #3851, opened 2022-06-10
Fix submitter not being displayed in the event workflow details

### LINKED ISSUE #3831: Submitter missing from workflow details
The submitter of a workflow (should be a username) seems to be missing from the workflow details in the admin interface with the current `r/12.x` and `develop`:

Screenshot from 2022-06-02 16-41-41

Arnei:
Resolves #3831.

The change to workflow storage introduced with OC 12 caused workflow instances to only store the username of a user instead of the full user object. This caused a display error in the event workflow details. This PR fetches the additional user information based on the available username.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -27,16 +27,16 @@ angular.module('adminNg.controllers')
   'EventAssetCatalogsResource', 'CommentResource', 'EventWorkflowsResource', 'EventWorkflowActionResource',
   'EventWorkflowDetailsResource', 'ResourcesListResource', 'RolesResource', 'EventAccessResource',
   'EventPublicationsResource', 'EventSchedulingResource','NewEventProcessingResource', 'CaptureAgentsResource',
-  'ConflictCheckResource', 'Language', 'JsHelper', '$sce', '$timeout', 'EventHelperService', 'UploadAssetOptions',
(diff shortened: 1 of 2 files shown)


lkiesow:
Any opinion on if we actually want to show the user's full name instead of using the username?

I don't have a strong preference, but I would slightly prefer the username since it actually identified the user, while the displayed data do not necessarily do that. On top of that, it's easier.

gregorydlogan:
Does what it says on the tin.

I'd be tempted to show usernames since those will likely also be unique, whereas names might be less so?  No strong preference.

gregorydlogan (approved):

Arnei:
I've got no strong preference either. Having a unique identifier here is probably not very important here. Name + email should be unique enough and give a better indication on who to contact about the workflow than the username. On the other hand I agree that just using the username would be easier :)

Rillke (approved):
Looks good to me considering https://github.com/opencast/opencast/commit/a55686beba099b0d173b9183751297f408cb7e42#diff-3d8288974076a5de3165ea04bce42e7caff46736eab6bc684e9a913ace6cff57R1946

I prefer human-readable names with email (because it allows me to just throw a message to that user) but if most of the developers opted for user names, it shall be. We're not storing those WF details for a long time, either so not of much importance to me (and for asking our customer, it's too late).

lkiesow:
Works → merging

Screenshot from 2022-06-14 21-31-48

PR #3865, opened 2022-06-14
Rename yaml workflow title

geichelberger:
Changes the workflow title, mainly to get rid of the question mark.

* [x] have a concise title
* [x] pass automated tests


lkiesow:
Funny, that's something I wanted to do today. But I wanted to go one step further and drop one of the workflows. If we want to use YAML, can we maybe just use YAML for “Fast Testing” and then remove the XML variant of that one? If people prefer XML and want an example, they can still look at the default workflow. Just having the same workflow twice in the UI is confusing.

Screenshot from 2022-06-14 15-34-27

geichelberger:
That makes sense. I updated the PR.

Rillke (approved):

PR #3903, opened 2022-06-26
Common persistence util classes that also implement transaction retries

mtneug:
This took way longer than anticipated, but here we go:

This started with exceptions we were seeing operating Opencast with CockroachDB, a distributed DB that aims for being compatible with PostgreSQL. Other than PostgreSQL, however, CockroachDB only supports the strict `SERIALIZABLE` SQL consistency level for transactions resulting in transactions that are rolled back by CockroachDB. The application, i.e. Opencast, should then retry the transaction. Nearly every DB access in Opencast does not do this.

This PR implements common classes in functional style for DB access basically a wrapper around `EntityManager`. Transaction errors can then be handled in a common way allowing to retry by executing the passed lambda again. Working in a functional way, this PR rewrites may helper methods in DTOs that execute queries to methods that return a `Function` that can be executed in a DB exec context. Some DTOs even asked for an `EntityManagerFactory` or opened their own transaction which I feel strongly against.

As it turns out, Opencast already had common classes for DB access (or I should better say multiple sets including some not even in use). These are now removed in favour of this implementation that includes some of the ideas.

I still need to test this with our staging system and CockroachDB. A local setup with H2 is working normally when performing quick ingests.

As this PR touches a lot of places, a quick review would be welcomed 😅

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan:
A quick review he says, of 5k new lines and 7.5k deletions. A quick review this is not, young ~~padawan~~developer.

mtneug:
I only want to avoid merge conflicts and new DB access code that needs to be adopted :D

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
```diff
@@ -78,6 +86,7 @@ public class UserSettingsService {
   @Activate
   public void activate(ComponentContext cc) {
     logger.info("Activating persistence manager for user settings");
+    db = dbSessionFactory.createSession(emf);
```
lkiesow, line 89: Why initializing this in here? Couldn't we just turn this into an OSGi service which binds itself to the EntityManagerFactory?

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentStateServiceImpl.java
```diff
@@ -198,19 +206,19 @@ protected AgentImpl getAgent(String name, String org) throws NotFoundException {
...
-      q.setParameter("id", name);
-      q.setParameter("org", organization);
-      return (AgentImpl) q.getSingleResult();
-    } catch (NoResultException e) {
-      throw new NotFoundException(e);
-    }
+  protected ThrowingFunction<EntityManager, AgentImpl, NotFoundException> getAgentEntityQuery(String name, String organization) {
+    return em -> {
+      try {
+        TypedQuery<AgentImpl> q = em.createNamedQuery("Agent.get", AgentImpl.class);
```
lkiesow, line 214: What's the advantage of this compared to

```java
  protected AgentImpl getAgentEntityQuery(EntityManager em, String name, String organization) {
    try {
      TypedQuery<AgentImpl> q = em.createNamedQuery("Agent.get", AgentImpl.class);
      …
```

The latter form is far more common in the Java world and will probably mean that new developers are more comfortable with the code, especially if you then combine it with calls like this where I'm wondering why we would construct this function to just call the function?

```java
AgentImpl existing = getAgentEntityQuery(agent.getName(), agent.getOrganization()).apply(em);
```

lkiesow (changes requested):
After a quick glance, I don't quite know how I feel about this and would at least like to take a closer look.

I see the intention, but this feels like all the other layers of abstractions which were added over the years and which ended up causing problems. This patch means that new Opencast developers couldn't any longer just Google “java database query” and get a decent result which tells them more or less everything they need to know. They would need to learn to deal with the additional layer of abstraction.

This also seems to introduce unnecessary packing and more or less instant unpacking of code in functions, which is weird. Looking at the calls, just using that code would be so much more intuitive.

This also introduces a number of new functions all over the place using the old Java `Function` objects, which always feel very clunky and not very intuitive to me.

A hard `-1` for me is also the current lack of any documentation. What does `execTxChecked` do? Who knows. While I don't usually care much about JavaDocs, here they are essential because there is no other way of finding out about how to use it apart from reading the code.  But that's obviously easy to change.

To end on a more positive note, this does seem to replace a few previous attempts at doing something similar. I like that.

I'll try having a closer look at this next week.

mtneug:
> After a quick glance, I don't quite know how I feel about this and would at least like to take a closer look.
> 
> I see the intention, but this feels like all the other layers of abstractions which were added over the years and which ended up causing problems. This patch means that new Opencast developers couldn't any longer just Google “java database query” and get a decent result which tells them more or less everything they need to know. They would need to learn to deal with the additional layer of abstraction.

I disagree. In fact, searching for "java database query" will actually show results to directly connect to a database using raw JDBC. So users at least have to get familiar with `EnitityManager` and using these abstraction is actually very close to using `EntityManager`. E.g. the following code:

```java
EntityManager em = null;
List<FooDto> fooList;
try {
  TypedQuery<FooDto> query = em.createNamedQuery("Foo.selectAllByBar", FooDto.class)
          .setParameter("bar", myBar);
  fooList = query.getResultList();
} finally {
  if (em != null)
    em.close();
  }
}
```

could become

```java
List<FooDto> fooList = db.exec(em -> {
  TypedQuery<FooDto> query = em.createNamedQuery("Foo.selectAllByBar", FooDto.class)
          .setParameter("bar", myBar);
  return query.getResultList();
});
```

So the main parts (the query definition and execution) will stay the same and the abstraction will reduce the boilerplate with using `EntityManager`. You probably have seen that there are some special methods defining common queries that may read nicer IMO. So the code could also be written as

```java
List<FooDto> fooList = db.exec(namedQuery.findAll(
  "Foo.selectAllByBar",
  FooDto.class,
  Pair.of("bar", myBar)
));
```

This, of course, is very different to using `EntityManager`, but you don't have to use it in this way and just use the second option.

> This also seems to introduce unnecessary packing and more or less instant unpacking of code in functions, which is weird. Looking at the calls, just using that code would be so much more intuitive.
> 
> This also introduces a number of new functions all over the place using the old Java `Function` objects, which always feel very clunky and not very intuitive to me.

`Function` is old? The main point is that transactions should be retried. How would I do that other than using lambdas?

The idea behind returning `Function` in DTOs is that the method can then be used directly with the `exec...` functions, e.g.:

```java
db.exec(FooDto.getAllByBarQuery(myBar));

...

db.execTx(FooDto.create(myFoo));
```

> A hard `-1` for me is also the current lack of any documentation. What does `execTxChecked` do? Who knows. While I don't usually care much about JavaDocs, here they are essential because there is no other way of finding out about how to use it apart from reading the code. But that's obviously easy to change.

Fair enough. I will add more documentation.

> To end on a more positive note, this does seem to replace a few previous attempts at doing something similar. I like that.
> 
> I'll try having a closer look at this next week.

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
mtneug: I thought about that, but I wasn't sure if every instance needs a different persistent unit, e.g. `AwsAssetDatabaseImpl`:

```java
@Reference(target = "(osgi.unit.name=org.opencastproject.assetmanager.aws.persistence)")
public void setEntityManagerFactory(EntityManagerFactory emf) {
  this.emf = emf;
}
```

Are `EntityManagerFactory` actually the same between modules? Wouldn't this require a `persistence.xml` in one module that contains every DTO?

modules/capture-admin-service-impl/src/main/java/org/opencastproject/capture/admin/impl/CaptureAgentStateServiceImpl.java
mtneug: The idea is that you can directly pass this to `exec...` functions, e.g.

```java
AgentImpl existing = db.exec(getAgentEntityQuery(String name, String organization));
```

In more complex scenarios with multiple queries using the same `EntityManager` (e.g. a larger transactions), this is a little bit awkward, but I wanted to keep it consistent.

mtneug: The idea is that you can directly pass this to `exec...` functions, e.g.

```java
AgentImpl existing = db.exec(getAgentEntityQuery(String name, String organization));
```

In more complex scenarios with multiple queries using the same `EntityManager` (e.g. a larger transactions), this is a little bit awkward, but I wanted to keep it consistent.

mtneug:
@lkiesow I added documentation for the util classes.

mtneug:
Fixed the merge conflicts.

As a side note: we run this in educast.nrw with CockroachDB; now with 12 before with a patched 11 version. That been said, we don't run every code path.

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
gregorydlogan: Whether a unique unit is *needed* is unclear (and potentially something that someone could be tasked with testing), but let's leave that alone for now.  AFAIK the emfs are different for each service, but I haven't actually tested.

modules/kernel/src/main/java/org/opencastproject/kernel/security/persistence/OrganizationDatabaseImpl.java
```diff
@@ -128,38 +132,28 @@ public void deleteOrganization(String orgId) throws OrganizationDatabaseExceptio
...
-                org.getAdminRole(), org.getAnonymousRole(), org.getProperties());
-        em.persist(organization);
-      } else {
-        organizationEntity.setName(org.getName());
-        organizationEntity.setAdminRole(org.getAdminRole());
-        organizationEntity.setAnonymousRole(org.getAnonymousRole());
-        for (Map.Entry<String, Integer> servers : org.getServers().entrySet()) {
-          organizationEntity.addServer(servers.getKey(), servers.getValue());
+      db.execTx(em -> {
+        Optional<JpaOrganization> organizationEntity = getOrganizationEntityQuery(org.getId()).apply(em);
```
gregorydlogan, line 137: why not do a `.get()` here and check against null below?

gregorydlogan (approved):
Approved so that it can go into 13.  That being said we need to *very quickly* find a way to enforce usage of this new paradigm, or we'll end up with random bits of code without the retry and error handling mechanisms.

Issue #3768, opened 2022-05-14
Applied user filter not shown in AdminUI; can't apply filter for users in filter dropdown anymore


Rillke:
**Describe the bug**
A clear and concise description of what you see:
*When filtering for a user by clicking their name in the Admin UI, e.g. in the presenter(s) column, the filter is correcty applied but is not shown in the field listing applied filters.*

**To Reproduce**
1. Go to https://stable.opencast.org or https://legacy.opencast.org and log in.
2. Make sure there are episodes shown, if not upload some with presenters set.
3. Clear all the filters and searches by pressing the x in the filter field and blanking the search field.
4. Click on the name of a presenter in the main event table.
5. The filter is applied (filtered by the name you just clicked) but there is no indication in the filter field that there a filter currently applied
6. Click the filter icon in the filter field.  <img width="29" alt="Screenshot 2022-05-14 at 03 59 42" src="https://user-images.githubusercontent.com/2311611/168406713-e3167c61-3ff0-46bd-8e35-d0b011868b2c.png">
7.  Whereas you previously had some user filters, they are gone by now.

Situation now (OC versions: see below):

<img width="342" alt="Screenshot 2022-05-14 at 03 59 21" src="https://user-images.githubusercontent.com/2311611/168406700-b766c6e3-bdfb-47d5-9941-93d1cc24a1ac.png">

Situation previously (9.12):

<img width="268" alt="Screenshot 2022-05-14 at 04 02 16" src="https://user-images.githubusercontent.com/2311611/168406785-79ca57ef-aa29-430a-a98e-6e8b3fc85b85.png">

https://user-images.githubusercontent.com/2311611/168406681-0444cfdf-deb2-4ce2-9ab7-e1c91efd89de.mp4

Expected result (this was like it's shown in OC9.12):

<img width="253" alt="Screenshot 2022-05-14 at 03 50 03" src="https://user-images.githubusercontent.com/2311611/168406445-d0934f05-3d62-4611-97f7-f3c6277818a8.png">

**Environment Information**
 - Browser Firefox 100 (MacOS 10.15.7) Chrome 101 (Windows 10)
 - Opencast version with minor release Opencast 11.0.0.SNAPSHOT - 46ad60e - built on 2022-05-14 and Opencast 10.0.0.SNAPSHOT - 7e37474 - gebaut am 2022-05-14

lkiesow:
This is an issue related to the old (now removed) admin interface. Closing this.
[...]

PR #3909, opened 2022-06-27
Adds the bibliographic presenter filter back to the Admin UI

Arnei:
Resolves (the very detailed and nicely written) issue #3768.

Does what is described in the title.

@KatrinIhler, you removed the presenter filter (among others) some months ago for performance reasons I think. Would you be opposed to adding it back in? I'd argue that it makes sense to have it, since it's effectively still usable (by clicking on a presenter name in the events table) and it is confusing to not have the filter pop up in the filter box in that case. Or do you think it be better to remove the ability to click on presenters in the events table?

### DIFF
--- a/modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/query/EventListQuery.java
+++ b/modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/query/EventListQuery.java
@@ -90,6 +90,7 @@ public class EventListQuery extends ResourceListQueryImpl {
   public EventListQuery() {
     super();
     this.availableFilters.add(createSeriesFilter(Option.none()));
+    this.availableFilters.add(createPresentersFilter(Option.none()));
     this.availableFilters.add(createLocationFilter(Option.none()));
     this.availableFilters.add(createAgentFilter(Option.none()));
     this.availableFilters.add(createStartDateFilter(Option.none()));


KatrinIhler:
Just from memory, so take with a grain of salt: Afair the problem with the filter was that it loads a list of users you can select in a drop down menu, and if you have a lot of users, that can take... a while. Iirc there was also more than one filter who does this, so it loads the users more than once. As long as the available filters aren't loaded, you can't see any active filters either, which is a real pain. It's possible you can deactivate this in some other way (list provider config maybe, just exclude all users?), but it seems unwise to let people run into this by default if the filter isn't used much.
I didn't know you can still use this filter in another way, to be honest. Maybe there's a way to show it without adding the user loading back in? But I don't know how much work that would be.

gregorydlogan:
Can we get a perf check on this with a cluster with lots of users? It's hard to make a call on whether this is reasonable without knowing if we're going to sink adopter clusters.

KatrinIhler:
@gregorydlogan I currently don't have access to a client system with that many users. We deactivated this for SWITCH because for them, it took minutes to load the filter view.

gregorydlogan:
Would it be reasonable to make this filter (and *only* this filter) something that can be easily turned off?

I suppose we could add an 'expensive' option to turn all of the expensive ones on and off, assuming there is more than one like that.

Arnei:
Not necessarily against marking some filters as "can slow your system to a crawl", but just noting that making the filter configurable would require to solve the initially linked issue some other way.

gregorydlogan:
@Arnei Agreed.  Do you think you (or someone else) would have time to address this?

Arnei:
> @Arnei Agreed. Do you think you (or someone else) would have time to address this?

I don't have time for that currently.

gregorydlogan:
So, is this still relevant in Picard?  I'm assuming so.  But this has been hanging around so long I suspect that it's not going to be merged, especially at the end of the support window.

Arnei:
Is indeed relevant for picard (although picard will actually display that a filter was set, just not which one, so it's a little less bad there). This is definitely not gonna get merged as is, the performance dip would just not be worth it, so I'd be fine with closing it.

gregorydlogan:
Done

PR #3912, opened 2022-06-28
Always store ACE entries in JSON as array

### LINKED ISSUE #3904: Admin interface does not accurately display access control list
Create a series with a very simple access control list like this:

```
❯ curl -u admin:opencast http://localhost:8080/series/ \
  -F title="I 🖤 Opencast" \
  -F acl='{"acl": {"ace": [{"allow": true,"role": "ROLE_USER","action": "read"}]}}'
```

The ACL will bet set correctly and be sent to the frontend correctly, but will not at all be displayed in the admin interface:

Screenshot from 2022-06-26 23-01-39

Arnei:
Removes an (imo) unnecessary special case when storing ACLs as JSON.

Resolves #3904.

When storing ACLs as JSON, it seems much nicer to always have an array instead of having a special case for single entries. This avoids problems like in the issue mentioned above, where the frontend code is only equipped to handle arrays.

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/security/api/AccessControlParser.java
+++ b/modules/common/src/main/java/org/opencastproject/security/api/AccessControlParser.java
@@ -223,20 +223,16 @@ public final class AccessControlParser {
     JSONObject json = new JSONObject();
     JSONObject jsonAcl = new JSONObject();
     List<AccessControlEntry> entries = acl.getEntries();
-    int numEntries = entries.size();
-    switch (numEntries) {
-      case 0:
-        break;
-      default:
-        JSONArray jsonEntryArray = new JSONArray();
-        jsonAcl.put(ACE, jsonEntryArray);
-        for (AccessControlEntry entry : entries) {
-          JSONObject jsonEntry = new JSONObject();
-          jsonEntry.put(ACTION, entry.getAction());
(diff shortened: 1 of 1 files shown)


modules/common/src/main/java/org/opencastproject/security/api/AccessControlParser.java
```diff
@@ -227,14 +227,6 @@ public static String toJson(AccessControlList acl) throws IOException {
...
-        AccessControlEntry singleEntry = entries.get(0);
-        JSONObject singleJsonEntry = new JSONObject();
-        jsonAcl.put(ACE, singleJsonEntry);
-        singleJsonEntry.put(ACTION, singleEntry.getAction());
-        singleJsonEntry.put(ROLE, singleEntry.getRole());
-        singleJsonEntry.put(ALLOW, singleEntry.isAllow());
-        break;
       default:
         JSONArray jsonEntryArray = new JSONArray();
         jsonAcl.put(ACE, jsonEntryArray);
```
geichelberger: I think replacing the switch statement with an if statement would be better.

geichelberger (changes requested):

geichelberger (approved):
LGTM

PR #3915, opened 2022-06-29
Fix Ingest Service Download Feature

KatrinIhler:
I actually wanted to make this a bit prettier, but I can't find the time and people keep running into this, so here it is. Basically, the feature which allows the ingest service to download files from a different Opencast instance via digest was broken by some refactoring a while ago, and this fixes it.

@mtneug Have you tried this out already? If so, you could review this. :D

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/impl/IngestServiceImpl.java
@@ -1582,6 +1582,7 @@ public class IngestServiceImpl extends AbstractJobProducer implements IngestServ
         if (uri.toString().matches(downloadSource)) {
           //NB: We're creating a new client here with *different* auth than the system auth creds
           externalHttpClient = getAuthedHttpClient();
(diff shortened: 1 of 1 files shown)


gregorydlogan (approved):

mtneug:
I tried this now, and why exactly are we assuming digest auth? My assumption would be that basic auth is used since this is more common (assuming the source system is not Opencast) and we also encourage the usage of basic auth. In my case, I use a normal JPA user in the source system that cannot log in using digest auth.

gregorydlogan:
The [config for this](https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.ingest.impl.IngestServiceImpl.cfg#L48) makes it looks like it's exclusively for fetching from *another* OC install.  I can see defaulting to digest auth in that case, although I can see how someone might want auth but not digest - especially if they're downloading from something other than OC.

mtneug:
I have a patch for basic auth support (configured in the config file) as well as forced pre-authenticated basic auth (otherwise, requests would only be authenticated if required). This is currently running for a migration in educast.nrw. Will post a PR in the coming days.

PR #3920, opened 2022-06-30
Automate the language selection in the transcription workflow of amberscript

marwyg:
Adds 2 new configurations for the amberscript transcription workflow. First, you can activate this feature, so the language for the transcription will be taken from the dublincore catalog. If no language is given or the feature is deactivated, the default process will be executed. 

A new class will handle the many ways a language can be represented. It takes the different fields of the java 'Locale' class and maps them to the corresponding amberscript language code. The second configuration variable allows someone to add custom mappings. If somebody knows, what language codes they are using internally, they can add these mappings manually. The custom mappings will overwrite existing mappings in case the key already exist.

### DIFF
--- a/etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
+++ b/etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
@@ -8,6 +8,24 @@
 # Can be one of: nl, en, de, fr, da, sv, fi, no, es
 # language=en
 
+# Set to true if the transcription language should be taken from the dublincore catalog
+# If the dublincore catalog doesn't contain a language, the default process for determine
+# the language will be taken. Note, that the language field in the workflow operation
+# will be overwritten, in case the dublincore catalog has a language
+# Default=false
+#language.from.dublincore=false
+
+# Manual mapping for language codes
+# Here you can set which language spellings/codes from your system shall be
+# mapped to the corresponding amberscript language. In general the amberscript workflow
+# should figure it out by itself. Use this, if you have some special language strings and run into errors
+# You can map a language to the amberscript code with a Colon ':'
+# You can separate multiple mappings with a comma ','
+#example1=english:en
+#example2=german:de,french:fr,spanish:es
+#example3=SomeWeirdL4nguageC0de:en
+#language.code.map=
+
 # Default Job Type to use
 # You can override this within you workflow (see workflowoperationhandler amberscript-start-transcription).
 # Default: direct
--- a/etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
+++ b/etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
@@ -21,9 +21,11 @@
 # should figure it out by itself. Use this, if you have some special language strings and run into errors
 # You can map a language to the amberscript code with a Colon ':'
 # You can separate multiple mappings with a comma ','
-#example1=english:en
-#example2=german:de,french:fr,spanish:es
-#example3=SomeWeirdL4nguageC0de:en
+# Examples:
+# language.code.map=english:en
+# language.code.map=german:de,french:fr,spanish:es
+# language.code.map=SomeWeirdL4nguageC0de:en
+# These are three different examples, please note that you can specify 'language.code.map' only once
 #language.code.map=
 
 # Default Job Type to use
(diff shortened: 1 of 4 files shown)


modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptLangUtil.java
```diff
@@ -0,0 +1,210 @@
...
+   */
+  private void addAmberLangToMap(String amberLang) {
+    Locale locale;
+    if (amberLang.contains("-")) {
+      locale = Locale.forLanguageTag(amberLang);
+    } else {
+      locale = Locale.forLanguageTag(amberLang + "-" + amberLang);
+    }
+    if (locale.getDisplayCountry().equals("")) {
+      logger.warn("Didn't found locale for code '{}'", amberLang);
```
gregorydlogan: `Locale not found for code '{}'`

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -247,6 +265,35 @@ public void activate(ComponentContext cc) {
     }
     logger.info("Cleanup result files after {} days.", cleanupResultDays);
 
+    Option<String> languageFromDublinCoreOpt = OsgiUtil.getOptCfg(cc.getProperties(), LANGUAGE_FROM_DUBLINCORE);
+    if (languageFromDublinCoreOpt.isSome()) {
+      try {
+        languageFromDublinCore = Boolean.parseBoolean(languageFromDublinCoreOpt.get());
+      } catch (Exception e) {
+        logger.warn("Configured '{}' is invalid. Set to 'false'.", LANGUAGE_FROM_DUBLINCORE);
```
gregorydlogan: `Configuration value for '{}' is invalid, defaulting to false.`

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -247,6 +265,35 @@ public void activate(ComponentContext cc) {
...
+      }
+    }
+    logger.info("Configuration '{}' set to '{}'.", LANGUAGE_FROM_DUBLINCORE, languageFromDublinCore);
```
gregorydlogan: Similar wording to above.

etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
```diff
@@ -8,6 +8,24 @@
...
+
+# Manual mapping for language codes
+# Here you can set which language spellings/codes from your system shall be
+# mapped to the corresponding amberscript language. In general the amberscript workflow
+# should figure it out by itself. Use this, if you have some special language strings and run into errors
+# You can map a language to the amberscript code with a Colon ':'
+# You can separate multiple mappings with a comma ','
+#example1=english:en
+#example2=german:de,french:fr,spanish:es
+#example3=SomeWeirdL4nguageC0de:en
```
gregorydlogan: These examples are confusing.  Suggest something like
`#Examples:`
`# language.code.map=english:en`
`# language.code.map=german:de,french:fr,spanish:es`
`# language.code.map=SomeWeirdL4nguageC0de:en`

The goal here would be to ensure that it's obvious that the `example` part isn't supposed to be there, and that you can only specify `language.code.map` once.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -273,11 +320,32 @@ public Job startTranscription(String mpId, Track track) throws TranscriptionServ
...
-    String language;
-    if (args.length > 0 && StringUtils.isNotBlank(args[0])) {
-      language = args[0];
-    } else {
-      language = getLanguage();
+    String language = null;
+
+    if (languageFromDublinCore) {
+      for (Catalog catalog : track.getMediaPackage().getCatalogs(MediaPackageElements.EPISODE)) {
+        if (language != null) {
```
gregorydlogan: Why not move the `break` down to after the (hopefully) successful language fetch?  Ie, stick it on line 336, and skip re-looping.

gregorydlogan (commented):
Nits picked.

marwyg:
Thank you for the review.

PR #3921, opened 2022-06-30
Remove Solr from Search Service: Elasticsearch

lkiesow:
> __Note:__ This is prototype 1 of 2 for removing Solr from the search service you will find a pure database-based implementation in pull request #3922

This is a prototype for an Elasticsearch-based search service. All major functionality is working just fine. For a list of what is left to do, see below.

Major changes:
- Removed Solr completely!
- New `/search/{episode,series}.json` API
  - This breaks Player and media module, but is very similar and should be easy to fix
  - The new API gives direct access to all metadata
  - All metadata are treated as array, not jumping between different types
  - I've tried to avoid duplications
  - I've refused to provide XML endpoints: Build a wrapper around Opencast if you want it so desperately ;-P
  - `series.json` and `episode.json` are very similar. We could even merge them into a single `/search.json`?
  - Media package serialization is still the exact same stupid thing from before.
  - Examples of search responses can be found below.
- The old feed service got completely removed
  - No one seems to use it (talked about it on the adopters meeting) and it's very Solr specific
  - We think it's better to create something new if we need feeds
- This replaces all remote implementations
  - We have no local bits anymore; therefore we can just replicate the service
  - We can now have multiple admin and presentation nodes
- This requires all admins and presentation nodes to have access to Elasticsearch
  - That's a new requirement
  - We also probably want to encrypt that communication and enable authentication by default

As mentioned, this is a prototype and there are still things to sort out (often indicated by `// TODO: …` in the code). The major points missing are:

- [ ] Rearrange assemblies: We need the Elasticsearch client on more nodes
- [ ] Sign URLs doesn't work right now, and I tend to say that we should leave that to clients if they want to use that. It's a pretty costly  operation.
- [ ] At a few places, we should mark entities as deleted instead of actually deleting them.
- [ ] Permission checks were (and still are) shaky at some places
- [ ] Index rebuilds are missing
- [ ] This needs a player/media module update
- [ ] Optional: The way we update events if a series is updated is very ineffective

All in all, this is a good option if we need full text search or even want to expand it in the future. It removes a lot of our old stack, but does add a new runtime service dependency for the presentation nodes. It still needs index rebuilds and has the usual problems of having a cache. Most of the backend calls have been replaced with database calls directly. Using Solr for that was weird in the first place.

### Series

```
❯ curl -s -u admin:opencast 'http://localhost:8080/search/series.json?limit=1&offset=0' | jq
```
```json
{
  "limit": "1",
  "result": [
    {
      "org": "mh_default_org",
[...]

PR #3922, opened 2022-06-30
 Remove Solr from Search Service: Database

lkiesow:
> __Note:__ This is prototype 2 of 2 for removing Solr from the search service you will find an Elasticsearch-based implementation in pull request #3921

This is a technology demonstration (it's probably too early to call this a prototype) for a pure database-based search service. This shows that for most of the functionality, we could just drop the search index entirely. It's not feature complete, and we would lose a few features, though. But maybe it's still enough, since few people seem to actually use full text search all that much, and it means even less infrastructure requirements.

- Removed Solr and Elasticsearch completely!
- We search in the database directly
- We can add more fields if we want more specific search options (e.g. series titles)
- For searching through series (do we need that?) we could
  - either add a `search(…)` method to the series service
  - or also add series to the search database, similar to how we do that in Elasticsearch or Solr
- Most of the search options work like the y did before

For this tech demo, I completely ignored the series portion and also didn't implement the same nice JSON response I implemented in the Elasticsearch prototype. But we could do that if we choose to go this route easily enough.

All in all, this is a good option if we think of this more as the service which mostly delivers data to our players or lists a series. It removes even more old weird stuff than the other one did. It also minimizes our service dependencies. We also do not need index rebuilds at all, and we don't have to worry about our data being out of sync.

_Note: Use `-DskipTests` when building this._


lkiesow:
As we discussed, we will go with the Elasticsearch based variant.

PR #3923, opened 2022-06-30
Handle tracks with multiple videos in engage player

KatrinIhler:
We have some events that have tracks with multiple videos, and those can't be played in the engage player (since track.video.resolution is undefined). This is a small change to make this work also for non-HLS-VOD.


mliradelc:
Katrin, do you refer to Paella as the engage-player?

KatrinIhler:
@mliradelc Kind of, but this does not affect the paella player directly, but the integration code for the paella player in opencast that parses the data for the player. :)

miesgre (approved):
It's fine for me

mliradelc:
All good, also is ok for me

PR #3954, opened 2022-07-06
Handle process stdout and stderr

geichelberger:
If the buffer allocated for standard output/error is full,
the started process will wait forever if the buffer's content is never consumed.
On Linux-based systems, the pipe has a buffer size of 65,536 bytes.

We encountered the error in the inspect FFmpegAnalyzer step. It was caused by ffprobe which was complaining on stderr about skipping HLS tags. The m3u8 file contained enough of those tags to fill up the buffer.
`[hls @ 0x33c4900] Skip ('#EXT-X-PROGRAM-DATE-TIME:2022-07-05T17:32:43.386+0200')`

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
+++ b/modules/crop-ffmpeg/src/main/java/org/opencastproject/crop/impl/CropServiceImpl.java
@@ -319,8 +319,11 @@ public class CropServiceImpl extends AbstractJobProducer implements CropService,
     logger.info("Running {}", cropCommandline);
 
     try {
-      pbuilder = new ProcessBuilder(cropCommandline.split(" "));
-      Process process = pbuilder.start();
+      Process process = new ProcessBuilder(cropCommandline.split(" "))
+          .redirectError(ProcessBuilder.Redirect.DISCARD)
+          .redirectOutput(ProcessBuilder.Redirect.DISCARD)
+          .start();
+
       //wait until the task is finished
       exitCode = process.waitFor();
     } catch (InterruptedException | IOException e) {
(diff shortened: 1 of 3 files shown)


gregorydlogan:
I'd be tempted to have some way to view this data in the OC log as previously.  Some kind of debug flag maybe?  Detect based on the loglevel for the class?

geichelberger:
I don't think that it was visible in opencast logs. At least for the FFmpegAnalyze stuff, nothing got logged.

geichelberger:
But yeah, it would be nice to have the possibility to see the error log.

gregorydlogan:
I'm sure I've seen raw ffmpeg output during compose steps, and thought this was applying to that too.  Which, maybe it should?  Regardless, I can see a usecase for seeing the raw output when required.

geichelberger:
The easy way to achieve that is to implement an OutputStream that directly writes each line to the Slf4j debug output.

lkiesow:
That's more or less what the composer does.

geichelberger:
Never mind, my idea is not working. I thought it was possible to set an OutputStream as a redirect target, but it's only possible to write to a File.

lkiesow:
What the composer (`EncoderEngine`) does is to read the output and then log most of it using different log levels (that's what `handleEncoderOutput` does:

https://github.com/opencast/opencast/blob/2765b736904f8f38d742526cdbfff69c77ea7bc4/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/EncoderEngine.java#L227-L232

We could do something similar here, reading the output and then just logging it using `logger.debug(…)`. But this is already an improvement like this and I don't see any reason not to merge it like this. We can always add the logging later. It's not that we removed that feature.

lkiesow (approved):

PR #3960, opened 2022-07-07
Make Allowed Filenames Configurable

### LINKED ISSUE #3621: Class handout notes does not support non-latin characters as filename
I added an event in the admin ui, uploading a PDF with Chinese characters in the filename as class handout notes. When the event is finished, the Chinese characters are all replaced by _ in the handout file name as the screenshot below:
image

lkiesow:
This patch allows administrators to gain more control of how files are named in Opencast and what name is deemed safe by modifying the previously very restrictive file name pattern for ensuring safe file names.

This also moves the code for generating safe file names from the static PathSupport class to the working file repository.
Developers can now access this functionality either via the working file repository or via the workspace, although most of the working file repository and workspace methods ensure safe file names anyway and there is often no need to call this separately.

Most of the changes in the first commit stem from changes to unit tests which mock the workspace and now needs to be mocked differently and from removing the unused remote implementation of the working file repository, so that I do not have to modify it.

This should likely also fix #3621

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


etc/org.opencastproject.workingfilerepository.impl.WorkingFileRepository.cfg
```diff
@@ -0,0 +1,21 @@
...
+# extension. In any case, both base name and extension are limited to 255 characters each.
+#
+# WARNING: This setting is dangerous. It lets you configure Opencast in a way that poses a security risk to your system
+#          by e.g. allowing for path traversal attacks.
+# BETA:    For a long time, this was a very restricted, fixed value. Consider this to be an unstable configuration which
+#          might lead to problems since some component somewhere assumes the old restricted behavior. Test your system
+#          very carefully after changing this. Feedback about problems and successful changes are welcome.
+#
+# Default: (^\\W|[^\\w-_.])
+#filename.pattern = (^\\W|[^\\w-_.])
```
Rillke, line 21: Can we make it more clear from its name that it's a pattern for **forbidden characters** and not a pattern for those that are okay?

Rillke: Can we make it more clear from its name that it's a pattern for **forbidden characters** and not a pattern for those that are okay?

modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
```diff
@@ -115,6 +117,16 @@ public boolean accept(File dir, String name) {
   /** The URL path for the services provided by the working file repository */
   protected String servicePath = null;
 
+  /** The default pattern for characters allowed in filenames */
```
Rillke:
```suggestion
  /** The default pattern for characters forbidden in filenames */
```

etc/org.opencastproject.workingfilerepository.impl.WorkingFileRepository.cfg
```diff
@@ -0,0 +1,21 @@
+# Filename pattern
+#
+# Opencast will filter out dangerous pattern from file names in order to make these compatible with different operating
+# and file systems and to prevent local security risks like path traversal. For example, the default conversion would
+# look like this:
+#
+#   original:  -problematic-file/../name&&&%.ext
+#   converted: _problematic-file____name____.ext
```
Rillke:
```
converted: _problematic-file_.._name____.ext
```

modules/working-file-repository-service-impl/src/main/java/org/opencastproject/workingfilerepository/impl/WorkingFileRepositoryImpl.java
```diff
@@ -208,6 +225,29 @@ public void deactivate() {
...
+   *
+   * @param fileName
+   *          The file name
+   * @return the safe version
+   */
+  @Override
+  public String toSafeName(String fileName) {
+    var extension = FilenameUtils.getExtension(fileName)
+        .replaceAll(filenameRegex, "_");
+    var baseName = FilenameUtils.getBaseName(fileName)
```
Rillke, line 242: This doesn't appear to do what's documented in the .properties configuration file (i.e. the baseName does not contain the path part):

```
#   original:  -problematic-file/../name&&&%.ext
#   converted: _problematic-file____name____.ext
```

which is okay but documentation might need an update.

etc/org.opencastproject.workingfilerepository.impl.WorkingFileRepository.cfg
Rillke: The underscore `_` is part of `\w`, according to [Oracle](https://docs.oracle.com/javase/tutorial/essential/regex/pre_char_classes.html).

lkiesow:
I have hopefully addressed all your comments. It would be great if you could re-review this

Rillke:
Tested this locally with the fast testing WF and the outcome was as expected.

image

image

To my defense, I have to say it's really hot today (and was during the past days).

Rillke (approved):

Rillke:
@lkiesow What was the Remote WorkingFileRepository used for/ with; was it ever in use?

lkiesow:
The `WorkingFileRepositoryRemoteImpl` seems to have been a remote implementation of the working file repository, like other remotes we have. It takes requests from other OSGi services and then relays these requests to the actual service via HTTP. However, the working file repository lives on all nodes anyway and this service was never in use. At least not since we switched to using Karaf. Before that, I didn't check. So, maybe it was shortly in use in Matterhorn 0.x or 1.x, but I doubt it and it definitely hasn't been used for ages. It's just that no one noticed.

Since I had to modify the working file repository, I noticed and instead of updating the dead code to match the new interface, I decided to remove it and not spend any time there.

lkiesow:
> Rillke approved these changes 14 days ago

@Rillke, just to make sure: Do you think this can be merged like this? Or do you want me to change anything?

Rillke:
Please merge.

Rillke:
This is now causing #4649. Previously, all dots (and other characters matched by `\W`) in the base file name [were converted to `_`](https://github.com/opencast/opencast/pull/3960/files#diff-777dbd40c77b8eab9f93a0fa6cf20b8683bb29d199a94ca2a58a8c0b2ce67238L56) (underscore).

PR #4026, opened 2022-07-27
Make LtiServiceImpl read config file properly

Gao-Jun:
Opencast 12.x cannot properly read config from `etc/org.opencastproject.lti.service.impl.LtiServiceImpl.cfg`.

According to `LtiServiceImpl.java`, the config is expected be loaded in `updated` method. Yet, it seems the method never be invoked.

After some digging of the code, I added `service.pid` property to the `@Component` annotation and the config file can be loaded properly now. I'm not familiar with OSGi, but I guess may be there are some changes have been made to force explicitly define `service.pid` property.

This PR can make LtiServiceImpl read config file properly. But I guess there are some other components have the same problem and should be modified accordingly.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
+++ b/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java
@@ -111,7 +111,8 @@ import java.util.stream.Collectors;
     immediate = true,
     service = { LtiService.class,ManagedService.class },
     property = {
-        "service.description=LTI Service"
+        "service.description=LTI Service",
+        "service.pid=org.opencastproject.lti.service.impl.LtiServiceImpl"
     }
 )
 public class LtiServiceImpl implements LtiService, ManagedService {


Arnei:
Could you elaborate on what exactly cannot be properly read from the config file? In my local 12.x installation, both the `activate` and the `updated` method are called on startup and when I make changes to `etc/org.opencastproject.lti.service.impl.LtiServiceImpl.cfg`. No `service.pid` needed.

Whether that is how `activate` and `updated` should behave is another topic...

Gao-Jun:
Actually, this happens randomly. But set `service.pid` explicitly will overcome this problem, which ensure `update` method be invoked properly. In other words, without `service.pid` property, system seems that occasionally omit the `update` method.

For example, if add the property with `service.pid=foo`, and add a `foo.cfg` file in `etc` folder, the config in `foo.cfg` will loaded.

As I said, I'm not family with OSGi, so I cannot tell why sometimes it works and sometimes it doesn't.

Also, I face some similar issues these days, all about Components with `update` method and `service.pid` not set. 

I'm not sure whether I described this issue clearly enough.

Arnei:
Thanks for elaborating, I think I understand your problem better now. Not that I know why it happens, I'm not an OSGI expert either :')

Instead of adding the `service.pid`, could you try replacing the `ManagedService` interface with OSGI annotations, like in for example #3431, and test if that fixes your problem? Iirc the general consensus in the community is to move away from `ManagedService` implementations anyway, so if replacing it with OSGI annotations also fixes your problem that's two birds with one stone.

It might require slightly rewriting the `activate` and `updated` methods, since methods annotated with `@Activate` are ONLY called when the service first registers and methods annotated with `@Modified` are ONLY called on subsequent changes to the configuration.

Gao-Jun:
Thank you for your reply. I'll try your method later. It it works, I'll make another PR.

Gao-Jun:
@Arnei You're right. I create a new PR #4031 by removing `ManagedService` from `LtiServiceImpl`. It works fine so far.
This PR would be closed.

PR #4032, opened 2022-07-29
Clarify migration script dependencies and make aware of the mysql_connector_python error

marwyg:
I updated the docs, so people hopefully don't run into the weird "UTF-8" error, when executing the 11-to-12 database migration script.

Issue #4028, opened 2022-07-27
MariaDB upgrade script for 12.x fails over Character Set Issue


mliradelc:
**Describe the bug**
When try to upgrade Opencast MariaDB script with a database with a Collation set to `utf8_general_ci`, the script fails with the error:

```
# python3 /usr/share/opencast/docs/upgrade/11_to_12/workflows_mariadb.py
Creating connection to database...
Traceback (most recent call last):
  File "/usr/share/opencast/docs/upgrade/11_to_12/workflows_mariadb.py", line 121, in <module>
    connection = create_connection(host, user, password, database)
  File "/usr/share/opencast/docs/upgrade/11_to_12/workflows_mariadb.py", line 39, in create_connection
    database=db_name
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/pooling.py", line 287, in connect
    return MySQLConnection(*args, **kwargs)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/connection.py", line 137, in __init__
    self.connect(**kwargs)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/abstracts.py", line 1095, in connect
    self._open_connection()
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/connection.py", line 544, in _open_connection
    self.set_converter_class(self._converter_class)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/abstracts.py", line 1271, in set_converter_class
    self.converter = convclass(charset_name, self._use_unicode)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/conversion.py", line 136, in __init__
    MySQLConverterBase.__init__(self, charset, use_unicode, str_fallback)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/conversion.py", line 59, in __init__
    self.set_charset(charset)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/conversion.py", line 73, in set_charset
    self.charset_id = CharacterSet.get_charset_info(self.charset)[0]
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/constants.py", line 775, in get_charset_info
    info = cls.get_default_collation(charset)
  File "/home/ladmin/venv/lib64/python3.6/site-packages/mysql/connector/constants.py", line 746, in get_default_collation
[...]

PR #4033, opened 2022-07-29
Update mariadb migration script to make aware of the mysql_connector_python error

marwyg:
I updated the docs in the mariadb migration script from 11 to 12. It makes aware of the mysql_connector_python error that currently occurs, when the version 8.0.30 is used.


mliradelc:
Hi, Thanks for the nice update, did you have the same error as in #4028 ? If yes, can you please add a try-catch over that error. If you have problems, I'll be happy to add it. 😄

marwyg:
The change is included in https://github.com/opencast/opencast/pull/4032

PR #4083, opened 2022-08-16
Use bulk inserts for series index rebuild

Alaschinger:
Instead of rebuilding the series index entries with single index requests, now a configurable number of index request are summarized to a bulk request.
Therefor the repopulate function of the series service groups multiple series objects to a list now and new bundle update functions have been added to the series and the elastic search services.


Alaschinger:
This PR is superseded by #4120 which contains bulk inserts for all services.

PR #4086, opened 2022-08-17
Add thumbnail support to new editor backend

Arnei:
Modifies the editor-service to support thumbnails.

The editor service sends exisiting thumbnails to the frontend, and gets newly generated thumbnails from the frontend to save in the archive. It does not do any thumbnail generation itself.

As of now, thumbnails received from the frontend are not added to any publication (maybe they should?), just the archive. When using the thumbnail editor, Opencast workflows will need to be modified in such a way that they do not accidentally overwrite thumbnails.

Related editor PR: https://github.com/opencast/opencast-editor/pull/760

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
+++ b/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
@@ -53,6 +53,7 @@ import org.opencastproject.mediapackage.MediaPackageElementBuilderFactory;
 import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.Publication;
 import org.opencastproject.mediapackage.Track;
+import org.opencastproject.mediapackage.attachment.AttachmentImpl;
 import org.opencastproject.metadata.dublincore.DublinCoreMetadataCollection;
 import org.opencastproject.metadata.dublincore.EventCatalogUIAdapter;
 import org.opencastproject.metadata.dublincore.MetadataJson;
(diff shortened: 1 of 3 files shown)


modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -715,8 +842,17 @@ public EditingData getEditData(final String mediaPackageId) throws EditorService
       final TrackSubData video = new TrackSubData(track.hasVideo(), videoPreview,
                         videoEnable);
 
+      final String thumbnailURI = Arrays.stream(internalPub.getAttachments())
+              .filter(attachment -> attachment.getFlavor().getType().equals(track.getFlavor().getType()))
+              .filter(attachment -> attachment.getFlavor().getSubtype().equals(getThumbnailSubtype()))
+              .map(MediaPackageElement::getURI).map(this::signIfNecessary)
+              .findAny()
+              .orElse(null);
```
gregorydlogan, line 850: I'm concerned this `null` will cause issues in the future, but this is *probably* fine for now since nothing else is using this.  Yet.

gregorydlogan (changes requested):
LGTM, needs the release notes file still.

Arnei:
@gregorydlogan Just noticed that I filed this against develop. Do you think it would be fine to go into r/12.x?

gregorydlogan:
I think so, given that the code should work without config changes.

PR #4089, opened 2022-08-17
Add optional key "skip-if-target-flavor-has-elements" to Image WOH

Arnei:
Adds an optional key to the image workflow operation handler that skips the operation if the specified target flavor contains at least one element. This can for example be useful if, for some events, you provide the images in some other form and don't want them to be overwritten by your image operation.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/ImageWorkflowOperationHandler.java
+++ b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/ImageWorkflowOperationHandler.java
@@ -115,7 +115,7 @@ public class ImageWorkflowOperationHandler extends AbstractWorkflowOperationHand
   public static final String OPT_TARGET_BASE_NAME_FORMAT_SECOND = "target-base-name-format-second";
   public static final String OPT_TARGET_BASE_NAME_FORMAT_PERCENT = "target-base-name-format-percent";
   public static final String OPT_END_MARGIN = "end-margin";
-  public static final String OPT_SKIP_IF_TARGET_FLAVOR_HAS_ELEMENTS = "skip_if_target_flavor_has_elements";
+  public static final String OPT_SKIP_IF_TARGET_FLAVOR_HAS_ELEMENTS = "skip-if-target-flavor-has-elements";
 
   private static final long END_MARGIN_DEFAULT = 100;
   public static final double SINGLE_FRAME_POS = 0.0;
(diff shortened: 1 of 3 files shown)


modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/ImageWorkflowOperationHandler.java
```diff
@@ -113,6 +115,7 @@ public class ImageWorkflowOperationHandler extends AbstractWorkflowOperationHand
   public static final String OPT_TARGET_BASE_NAME_FORMAT_SECOND = "target-base-name-format-second";
   public static final String OPT_TARGET_BASE_NAME_FORMAT_PERCENT = "target-base-name-format-percent";
   public static final String OPT_END_MARGIN = "end-margin";
+  public static final String OPT_SKIP_IF_TARGET_FLAVOR_HAS_ELEMENTS = "skip_if_target_flavor_has_elements";
```
geichelberger: All the other keys use dashes. Changing the style to underscores could be unexpected and error-prone.

geichelberger (changes requested):

lkiesow:
Assuming you have `presenter/img` and `presentation/img` as target flavor, the new option is set to `true` and someone previously uploaded `presenter/img`, would the operation now also skip generating `presentation/img`?

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/ImageWorkflowOperationHandler.java
```diff
@@ -176,6 +179,10 @@ WorkflowOperationResult main(final MediaPackage mp) throws WorkflowOperationExce
         logger.info("No source tracks found in media package {}, skipping operation", mp.getIdentifier());
         return handler.createResult(mp, Action.SKIP);
       }
+      if (cfg.skipIfTargetFlavorHasElements && mp.getElementsByFlavor(cfg.targetImageFlavor).length > 0) {
```
lkiesow, line 182: Couldn't the target flavor be `*/*`? In which case it would end up being the source flavor? This check wouldn't catch that, would it?

mtneug:
I don't like this very specific option for just a single operation. We actually have a more generic solution for this with the `AnalyzeMediapackageWorkflowOperationHandler` operation.

mtneug:
@Arnei do you think this can be dropped in favor of `AnalyzeMediapackageWorkflowOperationHandler`?

Arnei:
It completely slipped my mind that `AnalyzeMediapackageWorkflowOperationHandler` is a thing, I agree that the more generic solutions is better, so I'll close this PR.

PR #4098, opened 2022-08-22
Change default resolution of live schedule impl to standard 16:9

LukasKalbertodt:
I'm not sure why this default was set to a suuper-widescreen resolution. I think
it makes sense to change that as many Opencast installations don't change 
this, resulting in potentially weird effects (e.g. wrong player aspect ratio).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.liveschedule.impl.LiveScheduleServiceImpl.cfg
+++ b/etc/org.opencastproject.liveschedule.impl.LiveScheduleServiceImpl.cfg
@@ -9,8 +9,13 @@
 # The base URL that the player will use to play the live stream
 #live.streamingUrl=http://streaming.server/hls/
 
-# If a comma-separated list is provided, several resolutions will be generated for each flavor
-live.resolution=1920x540,960x270
+# If a comma-separated list is provided, several resolutions will be generated for each flavor.
+#
+# The default resolution is 16:9, a very typical aspect ratio nowadays. If your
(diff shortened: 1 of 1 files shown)


etc/org.opencastproject.liveschedule.impl.LiveScheduleServiceImpl.cfg
```diff
@@ -10,7 +10,7 @@
 #live.streamingUrl=http://streaming.server/hls/
 
 # If a comma-separated list is provided, several resolutions will be generated for each flavor
-live.resolution=1920x540,960x270
+live.resolution=1920x1080,1280x720
```
karendolan, line 18: Hi @LukasKalbertodt, it might help adopters to have a comment regarding the background of these numbers... something like
```
# The default resolution is 16:9, a typical single movie aspect ratio
# If your capture agent's default is to stream two 16:9 images stitched together, as an extra wide presenter and presentation video, parted later in an Opencast workflow, it may help to expand the resolution to 32:9: 1920x540,960x270
```
Or something shorter. Feel free to improvise.

LukasKalbertodt: Oh, now I finally understand where this weird resolution comes from :) Makes sense! I amended my commit to contain (almost) the comment you suggest.

mtneug (approved):

PR #4102, opened 2022-08-23
Move thumbnail preview generation to the client

darolfes:
This pullrequest allows for the editor to generate thumbnail previews on the client instead of generating the thumbnails on the opencast admin-node.
This reduces load on the admin-node, which previously was responsible for the preview generation.

Some smaller problems still exist, but overall this is probably much more usable than the current state.

Smaller problems:
- When cutting the default preview won't be refreshed (this is only really a problem if the default preview was already created, which only happens after generating another preview before switching back to the default preview).
- The text indicating a outdated preview is inaccurate and isn't shown in all cases.


mtneug:
I think the new editor already works on a new thumbnail feature (CC @Arnei)?

Arnei:
Yes, there is thumbnail feature in the works for the "new editor", but this PR concerns the "old editor".

PR #4119, opened 2022-08-26
Respect system settings when sending adopter data

gregorydlogan:
As reported on list, the adopter registration data is sent disregarding the system proxy settings.  This should resolve that, copying the methods from the version endpoint.

This also fixes [this](https://github.com/opencast/opencast/pull/4119/files#diff-8943442c01293b164d1c5ef7a1217b55208536757a3604d12040f21d10c0de17L115), which I'm 99% sure works even without this fix, but I'm not 100% sure since we haven't actually deployed the server side of this yet.  Hence the inclusion in `r/11.x`.

### DIFF
--- a/modules/adopter-registration-impl/pom.xml
+++ b/modules/adopter-registration-impl/pom.xml
@@ -59,6 +59,14 @@
       <artifactId>jakarta.ws.rs-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient-osgi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpcore-osgi</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.eclipse.persistence</groupId>
       <artifactId>jakarta.persistence</artifactId>
     </dependency>
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/Sender.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/Sender.java
@@ -21,6 +21,13 @@
 
 package org.opencastproject.adopter.statistic;
 
+import org.apache.http.HttpResponse;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.HttpDelete;
+import org.apache.http.client.methods.HttpPost;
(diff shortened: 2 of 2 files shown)


modules/adopter-registration-impl/pom.xml
```diff
@@ -58,6 +58,14 @@
       <groupId>jakarta.ws.rs</groupId>
       <artifactId>jakarta.ws.rs-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient-osgi</artifactId>
```
lkiesow, line 63: Why use an external library now that [Java has a built-in HTTP client](https://www.baeldung.com/java-9-http-client)? Does it not support HTTP forward proxies?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/Sender.java
```diff
@@ -112,7 +117,7 @@ public void deleteStatistics(String json) throws IOException {
    * @throws IOException General exception that can occur while processing the POST request.
    */
   private void send(String json, String urlSuffix) throws IOException {
-    send(json, urlSuffix, "GET");
+    send(json, urlSuffix, "POST");
```
lkiesow, line 120: This does not seem to have anything to do with forward proxy support. Is this change accidental?

gregorydlogan: No, it's fixing a bug.  We're sending data, via GET, to an endpoint expecting a POST.  We'll talk about this in the technical meeting today.

modules/adopter-registration-impl/pom.xml
gregorydlogan: I was cribbing from known working code (or at least code that no one's complaining about :smile:), which pulled this in.  I don't really care, but I also don't have an institutional proxy to test with...

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/Sender.java
gregorydlogan: I've since confirmed that this gets sent as a POST regardless - probably the library is smart enough to correct dumb mistakes like this.  I'd still like to get this into 11.x in case we ever do get another release done there, but it's no longer driving a potential minor release on its own.

PR #4120, opened 2022-08-29
Use bulk inserts for all services during index rebuild

Alaschinger:
Instead of rebuilding the elastic search index entries with single index requests, now a configurable number of index request are summarized to a bulk request.
Therefor the repopulate functions of all services group multiple documents to a list now and new bundle update functions have been added to the corresponding services and to the elastic search service.

### DIFF
--- a/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
+++ b/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
@@ -42,6 +42,8 @@ import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.elasticsearch.ElasticsearchException;
 import org.elasticsearch.ElasticsearchStatusException;
 import org.elasticsearch.action.admin.indices.delete.DeleteIndexRequest;
+import org.elasticsearch.action.bulk.BulkRequest;
+import org.elasticsearch.action.bulk.BulkResponse;
 import org.elasticsearch.action.delete.DeleteRequest;
 import org.elasticsearch.action.delete.DeleteResponse;
 import org.elasticsearch.action.get.GetRequest;
@@ -285,6 +287,57 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
     } while (indexResponse == null);
 
     return indexResponse;
+  }
+
+  /**
+   * Posts the input documents to the search index.
+   *
+   * @param maxRetryAttempts
+   *          How often to retry update in case of ElasticsearchStatusException
+   * @param retryWaitingPeriod
+   *          How long to wait (in ms) between retries
+   * @param documents
+   *          The Elasticsearch documents
+   * @return the query response
+   *
+   * @throws IOException
+   *         If updating the index fails
+   * @throws InterruptedException
+   *         If waiting during retry is interrupted
+   */
+  protected BulkResponse bundleUpdate(int maxRetryAttempts, int retryWaitingPeriod,
+      List<ElasticsearchDocument> documents)
+          throws IOException, InterruptedException {
+    BulkRequest bulkRequest = new BulkRequest().setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
+
+    for (ElasticsearchDocument document: documents) {
+      bulkRequest.add(new IndexRequest(getSubIndexIdentifier(document.getType())).id(document.getUID())
+          .source(document));
+    }
+
+    BulkResponse bulkResponse = null;
+    int retryAttempts = 0;
+    do {
+      try {
+        bulkResponse = client.bulk(bulkRequest, RequestOptions.DEFAULT);
+      } catch (ElasticsearchStatusException e) {
+        retryAttempts++;
+
+        if (retryAttempts <= maxRetryAttempts) {
+          logger.warn("Could not update documents in index {} because of {}, retrying in {} ms.", getIndexName(),
+                  e.getMessage(), retryWaitingPeriod);
+          if (retryWaitingPeriod > 0) {
+            Thread.sleep(retryWaitingPeriod);
+          }
+        } else {
+          logger.error("Could not update documents in index {} because of {}, not retrying.", getIndexName(),
+                  e.getMessage());
+          throw e;
+        }
+      }
+    } while (bulkResponse == null);
+
+    return bulkResponse;
   }
 
   /**
--- a/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
+++ b/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
@@ -273,15 +273,14 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
         retryAttempts++;
 
         if (retryAttempts <= maxRetryAttempts) {
-          logger.warn("Could not update documents in index {} because of {}, retrying in {} ms.", getIndexName(),
-                  e.getMessage(), retryWaitingPeriod);
+          logger.warn("Could not update documents in index {}, retrying in {} ms.", getIndexName(),
+                  e, retryWaitingPeriod);
           if (retryWaitingPeriod > 0) {
             Thread.sleep(retryWaitingPeriod);
           }
         } else {
-          logger.error("Could not update documents in index {} because of {}, not retrying.", getIndexName(),
-                  e.getMessage());
-          throw e;
+          logger.error("Could not update documents in index {}, not retrying.", getIndexName(),
+                  e);
         }
       }
     } while (indexResponse == null);
@@ -305,7 +304,7 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
    * @throws InterruptedException
    *         If waiting during retry is interrupted
    */
-  protected BulkResponse bundleUpdate(int maxRetryAttempts, int retryWaitingPeriod,
+  protected BulkResponse bulkUpdate(int maxRetryAttempts, int retryWaitingPeriod,
       List<ElasticsearchDocument> documents)
           throws IOException, InterruptedException {
     BulkRequest bulkRequest = new BulkRequest().setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
@@ -330,9 +329,8 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
             Thread.sleep(retryWaitingPeriod);
           }
         } else {
-          logger.error("Could not update documents in index {} because of {}, not retrying.", getIndexName(),
-                  e.getMessage());
-          throw e;
+          logger.error("Could not update documents in index {}, not retrying.", getIndexName(),
+                  e);
         }
       }
     } while (bulkResponse == null);
@@ -374,9 +372,8 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
             Thread.sleep(retryWaitingPeriod);
           }
         } else {
-          logger.error("Could not remove documents from index {} because of {}, not retrying.", getIndexName(),
-                  e.getMessage());
-          throw e;
+          logger.error("Could not remove documents from index {}, not retrying.", getIndexName(),
+                  e);
         }
       }
     } while (deleteResponse == null);
@@ -669,9 +666,8 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
             Thread.sleep(retryWaitingPeriod);
           }
         } else {
-          logger.error("Could not query documents from index {} because of {}, not retrying.", getIndexName(),
-                  e.getMessage());
-          throw e;
+          logger.error("Could not query documents from index {}, not retrying.", getIndexName(),
+                  e);
         }
       }
     } while (searchResponse == null);
--- a/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
+++ b/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
@@ -58,6 +58,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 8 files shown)


modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
```diff
@@ -284,6 +286,57 @@ protected IndexResponse update(int maxRetryAttempts, int retryWaitingPeriod, Ela
...
+   *
+   * @throws IOException
+   *         If updating the index fails
+   * @throws InterruptedException
+   *         If waiting during retry is interrupted
+   */
+  protected BulkResponse bundleUpdate(int maxRetryAttempts, int retryWaitingPeriod,
+      List<ElasticsearchDocument> documents)
+          throws IOException, InterruptedException {
+    BulkRequest bulkRequest = new BulkRequest().setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE);
```
lkiesow, line 310: Using [`WriteRequest.RefreshPolicy.NONE` is recommended](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html) and I don't see a reason why we would need `IMMEDIATE` here.

modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
```diff
@@ -284,6 +286,57 @@ protected IndexResponse update(int maxRetryAttempts, int retryWaitingPeriod, Ela
...
+        if (retryAttempts <= maxRetryAttempts) {
+          logger.warn("Could not update documents in index {} because of {}, retrying in {} ms.", getIndexName(),
+                  e.getMessage(), retryWaitingPeriod);
+          if (retryWaitingPeriod > 0) {
+            Thread.sleep(retryWaitingPeriod);
+          }
+        } else {
+          logger.error("Could not update documents in index {} because of {}, not retrying.", getIndexName(),
+                  e.getMessage());
+          throw e;
```
lkiesow: Please don't separate error message and error. Throw an error with the appropriate message.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -765,4 +790,40 @@ private  Optional<Series> updateSeriesInIndex(String seriesId, String orgId,
...
+    try {
+      List<Optional<Series>> seriesOptBundle = index.addOrUpdateSeriesBundle(seriesIds, updateFunctions, orgId, user);
+      for (String seriesId: seriesIds) {
+        logger.debug("Series {} updated in the {} index", seriesId, index.getIndexName());
+      }
+      return seriesOptBundle;
+    } catch (SearchIndexException e) {
+      for (String seriesId: seriesIds) {
+        logger.error("Series {} couldn't be updated in the {} index.", seriesId, index.getIndexName(), e);
+      }
```
lkiesow: This will print the same error over and over again. Why not just print it once?

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
-                  }
-
-                  String aclStr = series.getAccessControl();
-                  if (StringUtils.isNotBlank(aclStr)) {
-                    try {
-                      AccessControlList acl = AccessControlParser.parseAcl(aclStr);
-                      updateFunctions.add(getAclUpdateFunction(seriesId, acl, organization.getId()));
-                    } catch (Exception ex) {
-                      logger.error("Unable to parse ACL of series {}.", seriesId, ex);
+      if (total > 0) {
```
lkiesow: This check is unnecessary, isn't it? If `total = 0` we will never run the following loop anyway.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
+        String lastOrg = databaseSeries.get(0).getOrganization();
+        int current = 0;
+        int n = 16;
+        while (current < total) {
+          List<SeriesEntity> seriesBundle = new ArrayList<>();
```
lkiesow: Naming: Please don't use names like series bundle. This is highly confusing since we work in an OSGi context and the series bundle could also mean the OSGi bundle of the series service. Maybe better to talk about a range of series, or a slice of the set of series or just series?

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
+            if (current + i >= total) {
+              break;
+            }
+            String currentOrg = databaseSeries.get(current + i).getOrganization();
+            if (currentOrg.equals(lastOrg)) {
+              seriesBundle.add(databaseSeries.get(current + i));
+            } else {
+              lastOrg = currentOrg;
+              break;
+            }
```
lkiesow: I think you are trying to skip series from different organizations here to only make bulk inserts for series from the same organization. But if you have series with `org1`, `org2`, `org1`, `org1` you may end up skipping the `org2` series entirely while inserting another one twice.

Plus, this also seems like a very complex way of doing this. You can probably just pass on the organization in the inserts.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -475,6 +555,85 @@ private void update(Series series) throws SearchIndexException {
...
+      lockBundle.get(lockBundle.size() - 1).lock();
+      logger.debug("Locked series '{}'", id);
+    }
+    try {
+      List<Optional<Series>> updatedSeriesOptBundle = new ArrayList<>();
+      List<Series> updatedSeriesBundle = new ArrayList<>();
+      boolean allPresent = true;
+      for (int i = 0; i < ids.size(); i++) {
+        Optional<Series> seriesOpt = getSeries(ids.get(i), orgId, user,
+                maxRetryAttemptsUpdate, retryWaitingPeriodUpdate);
```
lkiesow: To me, this seems really weird. We set out to rebuild the index from data we have in our database, which I think we assume to be our source of truth (compared to the index which acts as a cache). But instead of taking the data we got from the database, we start building a hyper complex way of merging database data with potentially incorrect data from the index itself?

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
+                    for (SeriesEntity series: seriesBundle) {
+                      String seriesId = series.getSeriesId();
+                      seriesIds.add(seriesId);
+                      logger.trace("Adding series {} for organization {} to the {} index.", seriesId,
+                              series.getOrganization(), index.getIndexName());
+                      List<Function<Optional<Series>, Optional<Series>>> updateFunctions = new ArrayList<>();
+                      DublinCoreCatalog catalog;
+                      try {
+                        catalog = DublinCoreXmlFormat.read(series.getDublinCoreXML());
+                        updateFunctions.add(getMetadataUpdateFunction(seriesId, catalog, organization.getId()));
```
lkiesow: Not sure what to do about that, and it was broken before, but this code will always set the creator to the system user, overwriting whoever actually created the series in the index.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
lkiesow: This is actually more about the complex way of passing series along to the actual update code, but this would also solve the organization issue why I put it here. Instead of passing along lists of identifiers and update functions, why not just pass on a list of data to update? That simplifies the code quite a bit and you could have something like:

```java

  var seriesBlock = new ArrayList<Series>();
  for (SeriesEntity series: seriesBundle) {
    String seriesId = series.getSeriesId();
    logger.trace("Adding series {} for organization {} to the {} index.", seriesId,
        series.getOrganization(), index.getIndexName());
    Organization organization = orgDirectory.getOrganization(series.getOrganization());
    User systemUser = SecurityUtil.createSystemUser(systemUserName, organization);
    SecurityUtil.runAs(securityService, organization, systemUser,
          () -> {
              var seriesData = Optional.of(new Series(seriesId, organization.getId()));
              try {
                DublinCoreCatalog catalog = DublinCoreXmlFormat.read(series.getDublinCoreXML());
                seriesData = getMetadataUpdateFunction(seriesId, catalog, organization.getId())
                    .apply(seriesData);
              } catch (IOException | ParserConfigurationException | SAXException e) {
                logger.error("Could not read dublincore XML of series {}.", seriesId, e);
                return;
              }

              String aclStr = series.getAccessControl();
              if (StringUtils.isNotBlank(aclStr)) {
                try {
                  AccessControlList acl = AccessControlParser.parseAcl(aclStr);
                  seriesData = getAclUpdateFunction(seriesId, acl, organization.getId())
                      .apply(seriesData);
                } catch (Exception ex) {
                  logger.error("Unable to parse ACL of series {}.", seriesId, ex);
                }
              }

              try {
                Map<String, String> properties = persistence.getSeriesProperties(seriesId);
                seriesData = getThemePropertyUpdateFunction(seriesId,
                        Optional.ofNullable(properties.get(THEME_PROPERTY_NAME)), organization.getId())
                    .apply(seriesData);
              } catch (NotFoundException | SeriesServiceDatabaseException e) {
                logger.error("Error reading properties of series {}", seriesId, e);
              }
              seriesBlock.add(seriesData.get());
          });
  }

  // do the actual index update
  updateSeriesBundleInIndex(seriesBlock);
```

You could even just run this loop over all data and then call update index if `seriesBlock.size() >= n`.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
+                      }
+
+                      try {
+                        Map<String, String> properties = persistence.getSeriesProperties(seriesId);
+                        updateFunctions.add(getThemePropertyUpdateFunction(seriesId,
+                                Optional.ofNullable(properties.get(THEME_PROPERTY_NAME)), organization.getId()));
+                      } catch (NotFoundException | SeriesServiceDatabaseException e) {
+                        logger.error("Error reading properties of series {}", seriesId, e);
+                      }
+                      updateFunctionBundle.add(updateFunctions.toArray(new Function[0]));
```
lkiesow: <del>You pass on these three update functions as three entries. Later on, that means, you get the series from the index multiple times to then run an update instead of having one update per series.</del>

Never mind. I misread that as `addAll`, you do exactly that in the next function. But why not do it in the first place? Why the array structure in between?

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -504,53 +504,78 @@ public void repopulate() throws IndexRebuildException {
...
+
+                    // do the actual index update
+                    updateSeriesBundleInIndex(seriesIds, organization.getId(),
+                            updateFunctionBundle);
+
+                  });
+          for (int i = 1; i <= seriesBundle.size(); i++) {
+            int logEntry = i + (current - seriesBundle.size());
+            logIndexRebuildProgress(logger, index.getIndexName(), total, logEntry);
+          }
```
lkiesow: I wouldn't generate a bunch of logs simultaneously, but just log the current state instead.

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
```diff
@@ -765,4 +790,40 @@ private  Optional<Series> updateSeriesInIndex(String seriesId, String orgId,
...
+      Function<Optional<Series>, Optional<Series>> updateFunction =
+              Arrays.stream(updateFunctionBundle.get(i)).reduce(Function.identity(), Function::andThen);
+      updateFunctions.add(updateFunction);
+    }
+
+    try {
+      List<Optional<Series>> seriesOptBundle = index.addOrUpdateSeriesBundle(seriesIds, updateFunctions, orgId, user);
+      for (String seriesId: seriesIds) {
+        logger.debug("Series {} updated in the {} index", seriesId, index.getIndexName());
+      }
```
lkiesow:
```suggestion
      logger.debug("Series {} updated in the {} index", seriesIds, index.getIndexName());
```

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -475,6 +555,85 @@ private void update(Series series) throws SearchIndexException {
     }
   }
 
+  /**
+   * Adds or updates the series in the search index. Uses a locking mechanism to avoid issues like Lost Update.
```
lkiesow: > Uses a locking mechanism to avoid issues like Lost Update.

Can you elaborate on that? What do we need this locking for?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -475,6 +555,85 @@ private void update(Series series) throws SearchIndexException {
...
+    }
+    try {
+      List<Optional<Series>> updatedSeriesOptBundle = new ArrayList<>();
+      List<Series> updatedSeriesBundle = new ArrayList<>();
+      boolean allPresent = true;
+      for (int i = 0; i < ids.size(); i++) {
+        Optional<Series> seriesOpt = getSeries(ids.get(i), orgId, user,
+                maxRetryAttemptsUpdate, retryWaitingPeriodUpdate);
+        updatedSeriesOptBundle.add(updateFunctionBundle.get(i).apply(seriesOpt));
+        if (!updatedSeriesOptBundle.get(updatedSeriesOptBundle.size() - 1).isPresent()) {
```
lkiesow: Isn't this basically guaranteed to be `false` all the time, since we just run a function on this which will update or create that series? And assuming this could be `true`, that would mean, we would just skip updating the whole block of series?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -475,6 +555,85 @@ private void update(Series series) throws SearchIndexException {
...
+      List<SearchMetadata<?>> resourceMetadata = inputDocument.getMetadata();
+      docs.add(new ElasticsearchDocument(inputDocument.getIdentifier(),
+              inputDocument.getDocumentType(), resourceMetadata));
+    }
+    try {
+      bundleUpdate(maxRetryAttemptsUpdate, retryWaitingPeriodUpdate, docs);
+    } catch (Throwable t) {
+      for (Series series: seriesList) {
+        throw new SearchIndexException("Cannot write series " + series + " to index", t);
+      }
```
lkiesow: This will throw an exception for exactly the first entry in this list. Then, throwing that exception will cause the program to immediately jump out of that loop.

lkiesow (changes requested):
I just did a review of on the series part of this pull request, but since I added quite a number of comments already, it might make sense to sort that out first and take a look at the other parts after that.

Some of my complaints are based on existing code that got reused or reworked, and we could certainly argue about if we actually want to change that or not. But I would like to (at least partly), because I find the current code extremely hard to read and understand. Mostly because of the data structures. Sometimes you have multiple lists, all containing parts of the data you need, sometimes you have the same data in multiple places at the same time, sometimes, you have lists in lists containing data and everything is named very similar (which kind of makes sense, since the variables do contain similar data).

modules/series-service-impl/src/main/java/org/opencastproject/series/impl/SeriesServiceImpl.java
Alaschinger: I just modified the original process, so that we can use multiple elements for bulk ingests. But dealing with the updated elements instead of the update functions and all the other stuff will also remove at least two additional abstraction layers. I really would prefer that and I'll try to integrate this approach in the other services, too.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
Alaschinger: I agree with you and even if that was the original way how the index has been rebuilt I'll remove that whole step, working with updated elements for the rebuild instead of old elements and update functions.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
Alaschinger: (same as above)

modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
Alaschinger: It seems that opencast doesn't display the actual version of a new created event in the adminUI if we use NONE instead of IMMEDIATE as refresh policy. The meta data fields remain empty in this case even if the ES doc has been submitted. Than an complete index rebuild would be necessary.

gregorydlogan:
IMO this looks good to go @lkiesow.  Your remaining complaints are pre-existing (and probably out of scope here anyway), or have been rebased away.  Planning on merging this prior to branch cut unless you object.

gregorydlogan (approved):

PR #2762, opened 2021-06-21
Small additions and improvements in search service (for the upcoming Tobira module)

LukasKalbertodt:
Like #2643 and #2644 before, this PR contains modifications there were required (or just useful) to write the "tobira" module. The module will enable the new video portal Tobira to synchronize with Opencast somewhat efficiently. It's still in development in https://github.com/elan-ev/opencast-tobira/. We want to push all tobira-unrelated improvements to the community as early as possible though. Hence this PR.

These changes should be completely backwards compatible and rather minor.

CC @JulianKniephoff @lkiesow 

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4129, opened 2022-08-29
Add Tobira module for Synchronization with an external Tobira application

LukasKalbertodt:
This PR adds a new module `tobira`. This mainly offers the `/tobira/harvest` API that is used to synchronize Opencast data with an external Tobira. Using the `since` parameter, it is possible to get incremental updates, i.e. all data that changed since the given timestamp. Currently, the API only returns information about episodes and series.

There have been some previous PRs that changed existing code to meet the requirements of the Tobira module: #2762 #3075 #3076 #3119

For reviewers: I kept all original commits as I think they provide some useful context. However, it's probably most useful to just review the resulting code, instead of each commit individually. 

### Why merge Tobira-related code into upstream (this community repo)?

Unfortunately, maintaining and including external Opencast modules is a hassle and has a number of disadvantages. We think it is worth having the Tobira module in the community-Opencast to make sure it's always working and to reduce the hassle for adopters who want to use Tobira.

There is quite a bit of interest in Tobira from various universities (including ETH Zürich, Bern, TU Wien, Osnabrück, Siegen, and more). Especially compared with some other modules included in the community version, we think the Tobira module will be used quite a lot. It also makes it a lot easier for organizations to quickly test Tobira to evaluate whether they want to use it.

Finally, this Opencast module doesn't hurt any adopter not using it, except probably eating a tiny amount of RAM. It just offers an API, but otherwise is completely idle. Also, the API can only be used by users with admin privileges.

### Why merge this into `r/12.x` instead of `develop`?

Minor Opencast releases can include new features, but no breaking changes. This is very clearly just a new feature and doesn't break anything. It is practically useful as at least three universities are planning on using Tobira with Opencast 12. Thus, we can avoid some work there, if this already lands in `12.x`.

Also, as mentioned in the previous section, this module is completely passive and won't bother anyone not using it.

### How is the API versioned?

We have `/tobira/version` returning a `x.y` version number. That's just used for Tobira internally and doesn't mean the API may be used externally. The API's docs contain a note about that. For information on how the versioning works, see the comment in `TobiraApi.java`.

### Why does Tobira have to synchronize with Opencast this way?

One important requirement towards Tobira was that it still works even if Opencast is down (assuming static files are still served). Thus, Tobira needs a copy/cache of the relevant Opencast data. Additionally, if Tobira would always talk to Opencast APIs every time it needed some information, it would slow down Tobira considerably. 

### Why should this module run on the presentation node?

This module requests quite a bit of data from the search service, which runs on presentation nodes. To reduce overhead, the Tobira module should run on the same node.

---

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraApi.java
```diff
@@ -0,0 +1,207 @@
...
+    property = {
+        "service.description=Tobira-related APIs",
+        "opencast.service.type=org.opencastproject.tobira",
+        "opencast.service.path=/tobira",
+        "opencast.service.jobproducer=false"
+    },
+    immediate = true,
+    service = TobiraApi.class
+)
+public class TobiraApi {
```
geichelberger: I would rename the class to match the other "REST"endpoints. e.g. TobiraEndpoint

modules/tobira/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+            <Import-Package>
+              javax.ws.rs;version=2.0.1,
+              javax.ws.rs.core;version=2.0.1,
+              *
+            </Import-Package>
+            <Export-Package>
+              org.opencastproject.tobira;version=${project.version}
+            </Export-Package>
+            <!-- Remove once all used services use OSGi annotations -->
+            <_dsannotations-options>norequirements</_dsannotations-options>
```
geichelberger: It should be possible to remove this.

geichelberger (changes requested):
An `intercept-url` entry is missing for the Tobira endpoint in the security.xml.

modules/tobira/pom.xml
LukasKalbertodt: Oh true, somehow I thought it was necessary :shrug:

geichelberger (approved):
LGTM

It feels like every bundle uses a different way to build JSON Responses ;)

PR #4131, opened 2022-08-29
Add Tobira functionality in the admin UI

JulianKniephoff:
This PR integrates the new video portal Tobira in the Opencast admin UI. This integration takes a form of new "Tobira" tabs in the two preexisting modals related to series: the series details modal and the new series wizard.

image

image

What these allow you to do is:

* "Mount" a series directly from the admin UI into an existing or newly created page of the Tobira page tree.
* See all the places a series is mounted in said page tree.
* Get a direct link to the series in Tobira, regardless of whether or not it was mounted in this or any other way.

This functionality will likely be expanded in the future; this is kind of the MVP we wanted for the "pilot phase" in Bern and at the ETH.

For further explanations of Tobira-specific concepts, please refer to our documentation or just ask. :smile: The same goes for setting up a Tobira instance which you will need to test this.

For reviewers: I kept all original commits as I think they provide some useful context. However, it's probably most useful to just review the resulting code, instead of each commit individually. 

Also note that this PR is relatively independent of the other big Tobira PR currently open. (#4129)

### Why merge Tobira-related code into upstream (this community repo)?

#4129 has some explanation why this makes sense for an additional Opencast module, and all of that goes doubly so for this case. There is no easy way to drop this integration into an existing, vanilla Opencast installation. You have to be running custom patches. And the same basically goes for developers. This integration can't ever be an external repository like the Annotation Tool for example without basically being a fork of (at least parts of) Opencast. No one wants to maintain that.

### Why merge this into `r/12.x` instead of `develop`?

Minor Opencast releases can include new features, but no breaking changes. This is very clearly just a new feature and doesn't break anything. It is practically useful as at least three universities are planning on using Tobira with Opencast 12. Thus, we can avoid some work there, if this already lands in `12.x`.

Without explicitly configuring your Tobira instance (see the included changes to the config) this patch also shouldn't change anything. You won't even see the tabs.

#### The Tobira-Part

The admin UI (backend) currently doesn't do any handshake/version negotiation/... with Tobira. I plan on looking into this in the next couple of days. If the PR isn't merged by then, I will just amend it here and strike out this section. Otherwise I will make a follow up PR.

#### The admin UI part

Like all the other `/admin-ng`-endpoints, the ones added here aren't versioned at all. This is an internal API, after all.

### Why not wait for admin-ui-picard?

Well, since that is not done yet, we really couldn't, since we want to take this to production in same places soon-ish. And so, with the work already done contributing it is really just the next logical step. See also the first point.

We plan on porting this to the new admin UI when it is merged, which should probably happen sooner rather than later.

---

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


etc/org.opencastproject.adminui.endpoint.SeriesEndpoint.cfg
```diff
@@ -14,3 +14,13 @@
...
 #eventsFilter.onlySeriesWithWriteAccess=true
+
+# The base URL of the Tobira instance you want to be able to publish series to.
+# You need to configure this for the Tobira panel to show up in the series creation wizard.
+#tobira.origin=http://localhost:3080
+
+# A shared secret between Tobira and Opencast to get authenticated access to the
+# `mount`-API. See also the `auth.trusted_external_key` configuration option in your
+# Tobira configuration.
+# You need to configure this for the Tobira panel to show up in the series creation wizard.
```
lkiesow, line 42: Maybe add something like:

> Be aware that configuring this means that anyone with write access to the Tobira editors in the admin interface will have access to do modifications to any part of the Tobira instance configured here. If you want to be more restrictive, you can also always add content within Tobira itself.

mtneug (changes requested):
This implementation does not support multi-tenancy.

JulianKniephoff:
> This implementation does not support multi-tenancy.

It should now. :wink:

etc/org.opencastproject.adminui.endpoint.SeriesEndpoint.cfg
JulianKniephoff: I took the liberty of adapting your suggestion a bit, but something to this effect is now added as well. :+1:

LukasKalbertodt (approved):
It's worth noting that all of these changes already went through a round of reviews over at [our fork](https://github.com/elan-ev/opencast-tobira/pulls?q=is%3Apr+is%3Aclosed) where we developed this stuff. I don't know a lot about the admin UI code, but I can confirm again that I already approved these changes once.

Issue #4084, opened 2022-08-16
Can't delete scheduled events


mliradelc:
**Describe the bug**
In Opencast 12, if you schedule an event, then try to delete, the event it will be deleted, but it will stay in the search service. If you try to delete again, the admin ui it will show the message `The event(s) could not be deleted, because you don't have the permission to do so.` even if you are the admin.

This problem was verified on stable.opencast.org

**To Reproduce**
1. Schedule a new event*
2. Delete it

The event will be `Finished` instead of being deleted

\* You need first a capture agent registered in opencast to schedule first. The easiest way to do it is registering a dummy capture agent through the API.
    1. Go to the API panel
    2. In the `/capture-admin` endpoint register a dummy with `POST /agents/{name}/configuration` add a name and configuration
        You can copy/paste this JSON to register the dummy:
        ```json
        {
        "key": "capture.device.names",
        "value": "presentation,presenter"
        }
        ```

**Workaround**

Reindexing removes the entry from Elasticsearch

**Environment Information**
 - OS: [Mac OS
 - Browser [Chrome]
 - Opencast version with minor release [12.x]

mliradelc:
Adding more information:

Specifically, when through the Admin UI the endpoint `DELETE /admin-ng/event/{event-id}` this gives `Error 401 (Unauthorizided)` and changes the state to `Finished`

mtneug:
Is this also happening for events that were recorded and then deleted?

mliradelc:
Hi Mathias, only happens with scheduled events.

I've updated the first post on how to easily create a dummy CA to try it.

mliradelc:
I've made more testing and debbuging, and the error triggers in `WorkflowServiceImpl.java:1137`
```java
  protected boolean assertMediaPackagePermission(String mediaPackageId, String action) {
    Opt<MediaPackage> assetMediaPackage = assetManager.getMediaPackage(mediaPackageId);
    if (assetMediaPackage.isSome()) {
      var currentUser = securityService.getUser();
      var mediaPackage = assetMediaPackage.get();
      return currentUser.hasRole(GLOBAL_ADMIN_ROLE)
          || currentUser.hasRole(securityService.getOrganization().getAdminRole())
          || authorizationService.hasPermission(mediaPackage, action);
    }
    return false;
  }
```

Specifically, the `assetMediaPackage` gets empty, thus can't enter to the logic. If I force entering to the If bracket, I get error getting the media package, because there is no one.

[...]

PR #4151, opened 2022-09-01
Allow global admins to always get workflow instances for a media package

KatrinIhler:
... even if that media package is already deleted from the asset manager. (We can't do the same for org admins because we can no longer make sure that the organization of the user and the media package are the same.)

This should allow global admins to delete events that were broken by #4084 (see also the PR #4137).

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -1135,15 +1135,15 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
   }
 
   protected boolean assertMediaPackagePermission(String mediaPackageId, String action) {
-    Opt<MediaPackage> assetMediaPackage = assetManager.getMediaPackage(mediaPackageId);
-    if (assetMediaPackage.isSome()) {
-      var currentUser = securityService.getUser();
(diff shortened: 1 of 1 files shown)


mliradelc:
Tried this patch and works, the best thing is complements #4137 in the way that the bug #4084 if you try to delete is not getting deleted, like an incomplete deleted.  This patch, you can now delete those incomplete deletions without re-index.

The only thing was that the admin-ui showed me the message "this event can't be deleted" but I think is not-an-issue because those events were caused for a malformed deletion in the first place.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1135,15 +1135,15 @@ protected void assertPermission(WorkflowInstance workflow, String action, String
   }
 
   protected boolean assertMediaPackagePermission(String mediaPackageId, String action) {
-    Opt<MediaPackage> assetMediaPackage = assetManager.getMediaPackage(mediaPackageId);
-    if (assetMediaPackage.isSome()) {
-      var currentUser = securityService.getUser();
-      var mediaPackage = assetMediaPackage.get();
```
mliradelc, line 1141: I was trying fixing this method, but now I know that the problem was mainly that line. It will fail with packages that doesn't have media package because there wasn't any.  I was not bold enough to delete it (And also I didn't understand the code completely).

mliradelc (approved):
Works! approved.

PR #4154, opened 2022-09-02
Fix workflow config handover from ingest to scheduler service

KatrinIhler:
When scheduling via the ingest service, the workflow config is not handed over correctly. This will result in the config keys either having the workflow config prefix ("org.opencastproject.workflow.config") twice or not at all in the scheduler config. Neither of that will work obviously. :D 
Stripping the prefix already happens in other parts of the ingest service, it was just forgotten here.


Arnei (approved):

Arnei:
I wasn't even aware you could pass workflow configs there. Is that documented anywhere?

KatrinIhler:
> I wasn't even aware you could pass workflow configs there. Is that documented anywhere?

The rest docs don't mention it, and don't allow you to add workflow config either. I don't know if they would be capable of doing that tbh. 
It's mentioned somewhere in the docs for capture agents, though that part says that the prefix should be stripped prior to ingesting, which at least for scheduling via the ingest service, isn't true. 
So that's a bit of a mess right now...

PR #4158, opened 2022-09-02
Give internal opencast editor a unique external URL for Shibboleth

KatrinIhler:
I'm not entirely sure if this will be helpful, but I thought I'd at least put it out there.

So what we're trying to do is link to the internal Opencast editor (_not_ the new editor) in Ilias. That URL is protected by Shibboleth, so you will be prompted to log in if you haven't already. However, after login, you will be redirected to the Admin UI, not the editor UI. This is because the part that's telling Opencast we want to go the editor is only in the query string, and this information gets lost. (If you have an idea about how to get this to work with Shibboleth and Nginx, I'd love to hear it, I currently don't think it's possible.)
So we stole an idea from SWITCH, which is to create a new endpoint under a unique URL that simply redirects to the internal one.

Ideally this would probably be part of Opencast, but not active by default, because most people will never need it, but I'm too tired to understand how this is currently done for other modules (e.g. the ones for Shibboleth). It looks complicated?
On the other hand... it's an endpoint.

Just tell me what you think. ;)

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -182,6 +182,7 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-distribution-service-streaming-wowza/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-distribution-workflowoperation/${project.version}</bundle>
     <bundle start-level="83">mvn:org.opencastproject/opencast-editor/${project.version}</bundle>
+    <bundle start-level="82">mvn:org.opencastproject/opencast-editor-external-url/${project.version}</bundle>
     <bundle start-level="83">mvn:org.opencastproject/opencast-editor-service/${project.version}</bundle>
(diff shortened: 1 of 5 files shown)


lkiesow:
While I remember this and might even have been the one writing this(?), this is a very specific edge case with no flexibility.
More than that, I think you could solve the same thing using a simple [Nginx rewrite rule](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite) like:

```
server {
    # ...
    rewrite ^/admin-ng/editor/(.+)$ /admin-ng/index.html#!/events/events/$1/tools/editor;
    return  403;
    # ...
}
```

That's probably more flexible.

geichelberger:
If shibboleth allows anonymous access and spring is handling the login redirect, it should be possible to use the `RedirectQueryParamAuthenticationEntryPoint` https://github.com/opencast/opencast/blob/3d36158cd5b8345926b5e4c772d963d280f925d1/etc/security/mh_default_org.xml#L489

PR #4172, opened 2022-09-07
Allow upload of multiple assets with the same flavor (based on config)

### LINKED ISSUE #4170: Uploading multiple tracks with same flavor when event is created does not work anymore
**Describe the bug**
We upload video assist files that consist of several parts when creating an event. We then concatenate them using our workflow.
This is not working anymore. The configuration _multiple=true_ available in _etc/listproviders/event.upload.asset.options.properties_ has no effect.

**To Reproduce**
1. In etc/listproviders/event.upload.asset.options.properties, add a source track with multiple=true e.g. EVENTS.EVENTS.NEW.SOURCE.UPLOAD.MULTIPLE_PARTS={"id": "track_parts","type":"track", "flavorType": "presenter","flavorSubType": "source", **"multiple":true**, "displayOrder":11, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv,.flac,.m4a,.mp3,.ogg,.wav,.wma"}
2. Select multiple files when uploading presenter
3. Only the last track is added to the media package.

**Environment Information**
 - Opencast 11

rute-santos:
Fixes #4170

### DIFF
--- a/docs/guides/admin/docs/configuration/admin-ui/asset-upload.md
+++ b/docs/guides/admin/docs/configuration/admin-ui/asset-upload.md
@@ -74,12 +74,10 @@ How to Enable Preconfigured Asset Options
 
 Catalogs and attachments can be added to new and existing events. New source tracks can only be defined for the upload of new events. Some
 predefined catalog and attachment examples are commented out in the properties file. You can uncomment any of these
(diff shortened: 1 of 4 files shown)


Arnei (approved):
Looks good to me.

PR #4175, opened 2022-09-08
Add Theodul Deprecation Notice

gregorydlogan:
~This pull request removes Theodul from Opencast 13 per the passed proposal.~
This pull request adds a deprecation warning interstitial for those who enable Theodul in Opencast 13.  There will be a follow-on PR *removing* Theodul once `r/13.x` is cut.

It also adds a redirect which sends requests to the Theodul player URL *back* to the normal /play redirect.  This may cause issues for those who have manually configured the player to be Theodul.  There is no handling for those cases where Theodul has been set as the default player - this results in a loop which should hopefully alert adopters quickly enough that something is awry.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/developer/docs/proposal-log.md
+++ b/docs/guides/developer/docs/proposal-log.md
@@ -9,6 +9,27 @@ The following list contains a list of passed proposals for reference.
 Passed Proposals
 ----------------
 
+### Deprecate Theodul with Opencast 12
+Proposed by Greg Logan <gregorydlogan@gmail.com>, passed on Fri, July 1 2022
+```no-highlight
+Hi all,
+
+After a discussion today at the technical meeting, I'm going to propose
+deprecating Theodul with Opencast 12.  It's no longer the default, and actually
+causes problems in some situations[1].  It's also no longer being developed.
+If this passes then Theodul could be removed from the codebase with Opencast
+13.  If this is a problem for your installation, now would be the time to speak
+up!
+
+Proposal closes EOD 2022-07-01
+
+G
+
+1: https://github.com/opencast/opencast/issues/1749
+```
+
+Note: Deprecation actually happened in Opencast 13, with removal in Opencast 14.
+
 ### Opensearch Support
 Proposed by Greg Logan <gregorydlogan@gmail.com>, passed on Wed, June 22 2022
 ```no-highlight


docs/guides/admin/docs/configuration/user-statistics.and.privacy.md
```diff
@@ -1,8 +1,6 @@
 User Statistics and Privacy
 ===========================
 
-*There exists a newer and more complete tracking service using Matomo included [as a module](../modules/player.matomo.tracking.md).*
```
gregorydlogan: This file went away, but as far as I know Paella can/does use the old tracking tool.  If this is no longer true maybe this file should go away entirely?  Or be replaced with Paella specific docs?

mtneug:
We should not forget to update the `NOTICES` file. @lkiesow, @gregorydlogan was this ever automated? I remember that this was discussed at some point.

gregorydlogan:
Derp, thanks for the reminder.  AFAIK NOTICES has not been automated, although it definitely should be at some point.

geichelberger:
An alternative approach to the redirect would be to display a placeholder with deprecation information and present a link to the video. This has the benefit that it is more evident for admins that something changed or that they still have links not using the /play functionality. Something like that is widespread if the content gets moved to a new location. Admins could still use a reverse proxy to have an automatic redirect.

PS: Losing the focus of the input field and pressing m has side-effects ;)

gregorydlogan:
@DZenker you asked for the redirect.  Would an optional placeholder be something useful?  Right now it anything linking to Theodul will redirect back to Paella via the `/play` endpoint

modules/engage-ui/src/main/java/org/opencastproject/engage/ui/TheodulRedirect.java
```diff
@@ -0,0 +1,120 @@
...
+        "service.description=Theodul Redirect Endpoint",
+        "opencast.service.type=org.opencastproject.engage.theodul.player.redirect",
+        "opencast.service.path=/engage/theodul/ui"
+    }
+)
+public class TheodulRedirect {
+
+  private static final Logger logger = LoggerFactory.getLogger(PlayerRedirect.class);
+
+  private static final String dep_warning = "<!DOCTYPE html>\n" + "<html lang=\"en\">\n" + "  <head>\n"
```
gregorydlogan: Since this is going to be a one release thing, and *hopefully* quickly replaced with Paella, I left this (lovingly written) hardcoded html in here rather than its own file.  Downside being that there's no translation available at all.

Issue #4076, opened 2022-08-14
Differences between generated and manual DB schema


mtneug:
We have observed long query times when accessing job tables which prompted a look at the DB schema. For the job issue, we noticed a missing index that was previously present with the custom DDL script. Adding that significantly reduced our query times and just job scheduling. For that, I will create a PR. However, I also found other differences. This issue is mostly there to document them. Many probably will not lead to performance issues. However, there are many foreign keys missing previously with the `ON DELETE CASCADE` options. There may therefore be records in tables that are no longer relevant. Additionally, some tables no longer have primary keys (others never had them to begin with).

Other things we noticed:
* `oc_oaipmh_harvesting` was never dropped in a migration although no longer used in the code
* `oc_user_settings` is probably never generated as the service behind it is never used by the frontend. We may want to drop this feature altogether. In any case, the schema has an error in the unique constraint that will prevent users from storing more than one setting.

I tried to represent the differences in a diff like manner in SQL syntax:

```diff
CREATE TABLE oc_organization_node (
-  PRIMARY KEY (organization, port, name),
)

CREATE TABLE oc_organization_property (
-  PRIMARY KEY (organization, name),
-  CONSTRAINT FK_oc_organization_property_organization FOREIGN KEY (organization) REFERENCES oc_organization (id) ON DELETE CASCADE
)

- CREATE INDEX IX_oc_organization_property_pk ON oc_organization_property (organization);

CREATE TABLE oc_capture_agent_role (
-  PRIMARY KEY (id, organization, role),
-  CONSTRAINT FK_oc_capture_agent_role_organization FOREIGN KEY (organization) REFERENCES oc_organization (id) ON DELETE CASCADE
+  UNIQUE INDEX unq_capture_agent_role_id_org_role (id ASC, organization ASC, "role" ASC)
+  CONSTRAINT FK_oc_capture_agent_role_organization FOREIGN KEY (organization, id) REFERENCES oc_capture_agent_state(organization, id),
)

CREATE TABLE oc_capture_agent_state (
-  CONSTRAINT FK_oc_capture_agent_state_organization FOREIGN KEY (organization) REFERENCES oc_organization (id) ON DELETE CASCADE
)

CREATE TABLE oc_job (
-  CONSTRAINT FK_oc_job_organization FOREIGN KEY (organization) REFERENCES oc_organization (id) ON DELETE CASCADE
)

CREATE TABLE oc_job_argument (
# No PK
)

[...]

PR #4179, opened 2022-09-09
Add index for oc_job_argument table

mtneug:
JPA previously did not create an index for the job_argument table (that has no primary key). This differs compared to the custom DDL script we used to have until Opencast 8. We have seen very slow query times resulting in slow scheduling.

Also see #4076.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/upgrade/12_to_13/postgresql.sql
+++ b/docs/upgrade/12_to_13/postgresql.sql
@@ -1,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id ON oc_job_argument (id);
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id_argument ON oc_job_argument (id, argument);
--- a/docs/upgrade/12.2_to_12.3_optional/mariadb.sql
+++ b/docs/upgrade/12.2_to_12.3_optional/mariadb.sql
@@ -1,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id ON oc_job_argument (id);
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id_argument ON oc_job_argument (id, argument);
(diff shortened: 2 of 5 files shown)


gregorydlogan:
NB: This needs testing to see if that the index is autogenerated on existing schemas.  This may affect mergability depending on how expensive that operation is...

mtneug:
Just tested this locally by deleting the index and restarting Opencast. This did **not** recreate the index. So I think we need a migration.

mtneug:
This now includes the migrations.

lkiesow (commented):
I'm fine with the automated part going into `r/12.x` since it will be automatically created and will not cause any manual steps, but the migration scripts should probably go into `12_to_13` only, there shouldn't be one for minor versions. Especially if this is already in Opencast since Opencast 8. I'm fine with adding a notice that people can apply this in advance, but I wouldn't do that as a script directory but in the release notes, since otherwise people will wonder what's the difference.

On that note: This should go into the release notes directory anyway.

I'm also wondering if we could always auto-create this by setting the table to create or extend:

https://github.com/opencast/opencast/blob/6900011a4a37bd22d8daf7d7421794e65661ae00/modules/adopter-registration-impl/src/main/resources/META-INF/persistence.xml#L14

gregorydlogan:
@mtneug ping on this.

docs/upgrade/12_to_13/postgresql.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id ON oc_job_argument (id);
+CREATE INDEX IF NOT EXISTS IX_oc_job_argument_id_argument ON oc_job_argument (id, argument);
```
ziegenberg, line 2: This is not going to work. I receive the following error message from our PostgreSQL:

```
index row requires 10032 bytes, maximum size is 8191.
```
This seems to be a hard limit on PostgreSQL.

ziegenberg (changes requested):

docs/upgrade/12_to_13/postgresql.sql
ziegenberg: Others seem to have stumbled across this before: https://postgrespro.com/list/thread-id/1506742

mtneug:
Closed in favor of #4450

PR #4180, opened 2022-09-09
Add basic auth support ingest download

mtneug:
The ingest service allows downloads from external HTTP servers. However, currently, only digest auth is supported. This introduces a new setting to switch this to basic authentication. By default, requests are only authenticated if the HTTP server requires this during a previous request attempt. An additional option allows forcing authenticating requests for basic auth.

Opencast allows forcing requests to be made with digest auth when setting the `X-Requested-Auth: Digest` header. This PR also allows the same for basic auth with `X-Requested-Auth: Basic`.

Also see #3915

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
@@ -23,6 +23,7 @@ package org.opencastproject.kernel.security;
 
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.web.AuthenticationEntryPoint;
+import org.springframework.security.web.authentication.www.BasicAuthenticationEntryPoint;
 import org.springframework.security.web.authentication.www.DigestAuthenticationEntryPoint;
 
 import java.io.IOException;
@@ -38,10 +39,12 @@ import javax.servlet.http.HttpServletResponse;
 public class DelegatingAuthenticationEntryPoint implements AuthenticationEntryPoint {
   public static final String REQUESTED_AUTH_HEADER = "X-Requested-Auth";
   public static final String DIGEST_AUTH = "Digest";
+  public static final String BASIC_AUTH = "Basic";
   public static final String INITIAL_REQUEST_PATH = "initial_request_path";
 
   protected AuthenticationEntryPoint userEntryPoint;
   protected DigestAuthenticationEntryPoint digestAuthenticationEntryPoint;
+  private BasicAuthenticationEntryPoint basicAuthenticationEntryPoint;
 
   /**
    * {@inheritDoc}
@@ -54,6 +57,8 @@ public class DelegatingAuthenticationEntryPoint implements AuthenticationEntryPo
           throws IOException, ServletException {
     if (DIGEST_AUTH.equals(request.getHeader(REQUESTED_AUTH_HEADER))) {
       digestAuthenticationEntryPoint.commence(request, response, authException);
+    } else if (BASIC_AUTH.equals(request.getHeader(REQUESTED_AUTH_HEADER))) {
+      basicAuthenticationEntryPoint.commence(request, response, authException);
     } else {
       // if the user attempted to access a url other than /, store this in the session so we can forward the user there
       // after a successful login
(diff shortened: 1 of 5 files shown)


modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
```diff
@@ -54,6 +57,8 @@ public void commence(HttpServletRequest request, HttpServletResponse response, A
           throws IOException, ServletException {
     if (DIGEST_AUTH.equals(request.getHeader(REQUESTED_AUTH_HEADER))) {
       digestAuthenticationEntryPoint.commence(request, response, authException);
+    } else if (BASIC_AUTH.equals(request.getHeader(REQUESTED_AUTH_HEADER))) {
+      basicAuthenticationEntryPoint.commence(request, response, authException);
```
lkiesow, line 61: Isn't this class also used by Opencast's internal node communication?

mtneug: Yes. OC nodes use the Digest path above.

lkiesow: In that case, this is very likely to re-introduce a severe security issue where anyone could basically ask Opencast to send its admin password to external servers in plain text. See for details:

https://groups.google.com/a/opencast.org/g/committers/c/4U_cJ-lCbNM/m/Mj_TfiRDBwAJ

mtneug: Wait, if I understand this correctly this is the incoming site. So `basicAuthenticationEntryPoint` would force requests to Opencast to use basic auth.

gregorydlogan: Here's how to trigger the above CVE with the code as it stands right now:

- Set things up to force basic auth, but don't change the user and password
- Ingest a maliciously crafted mediapackage

*hint: the default user and password are the system user and password*

IMO, what you want to do is add another case where the ingest service is checking that basic auth is forced __and__ that the user and password are set.  Ideally it would do an idiot check to require that the user and password *don't* match the global digest auth data, just to be safe.  Then, worst case, you've disclosed your download digest creds which are _
__required__ to be different than the system ones.  Those creds shouldn't get you much in terms of permissions I think...

mtneug: @lkiesow: Greg and I discussed this after the dev meeting. The code you commented on is not the requesting side but the source Opencast system receiving the request. The purpose is to force the use of basic auth using the `X-Requested-Auth` header:

```
$ curl -s -D - -o /dev/null http://opencast.my.edu/info/me.json
HTTP/2 200
date: Mon, 14 Nov 2022 23:38:28 GMT
content-type: application/json
content-length: 4030
expires: Thu, 01 Jan 1970 00:00:00 GMT
x-content-type-options: nosniff
x-xss-protection: 1; mode=block

$ curl -s -D - -o /dev/null -H 'X-Requested-Auth: Digest' http://opencast.my.edu/info/me.json
HTTP/2 401
date: Mon, 14 Nov 2022 23:38:45 GMT
content-type: text/html;charset=iso-8859-1
content-length: 657
www-authenticate: Digest realm="Opencast", qop="auth", nonce="MTY2ODQ2OTQyNTgwMjo0OWMzNzY2ZTg3NzNlNGY2MjE3ZmM4NGMwZjQxYTQzYw=="
cache-control: must-revalidate,no-cache,no-store

$ curl -s -D - -o /dev/null -H 'X-Requested-Auth: Basic' http://opencast.my.edu/info/me.json
HTTP/2 401
date: Mon, 14 Nov 2022 23:38:10 GMT
content-type: text/html;charset=iso-8859-1
content-length: 657
www-authenticate: Basic realm="Opencast"
cache-control: must-revalidate,no-cache,no-store
```

As for the requesting side: this is using a custom HTTP client and explicitly not the trusted OC system client.

I therefore think this does not bring back the CVE.

mtneug:
@gregorydlogan I added a check that disables the download feature if no user/password is given.

gregorydlogan (approved):
There we go, this will prevent sending auth data accidentally.  I guess it doesn't protect against an admin who chooses to reuse credentials, but there's not much we can do about that :D

mtneug:
ok seems there a failed tests. Will check.

PR #4182, opened 2022-09-09
Fix race condition when loading workflow definitions

mtneug:
There is a race between the OrganizationDirectoryService and WorkflowDefinitionScanner: workflow definitions that set the organization field must be loaded after the organization is loaded. If the org is not loaded beforehand the workflow definition would error out and not be available. This registers an organization listener. If a new organization is loaded previously error out workflow definitions are tried to be loaded in again.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java
@@ -25,6 +25,7 @@ import static org.opencastproject.security.api.SecurityConstants.GLOBAL_ADMIN_RO
 import static org.opencastproject.util.ReadinessIndicator.ARTIFACT;
 
 import org.opencastproject.security.api.Organization;
+import org.opencastproject.security.api.OrganizationDirectoryListener;
 import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.User;
 import org.opencastproject.util.ReadinessIndicator;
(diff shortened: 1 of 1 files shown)


geichelberger:
Wouldn't it be easier to remove the exception which causes the problem?

https://github.com/opencast/opencast/blob/89909d2eff63587599253c2242351295a2404952/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowDefinitionScanner.java#L211

A workflow with an organization that does not (yet) exist will be installed without effect. This function `getAvailableWorkflowDefinitions` can only return workflows for existing organizations. If there is a workflow with a faulty organization, nothing happens. 
If the organization is registered, later on, everything is fine. The function will return it.

Arnei:
I would prefer going with Matthias solution and keeping the exception. In case you made a mistake in the organization field (e.g. due to a typo), it gives you quick feedback. And if the faulty organization is intentional (because you know it will be added later on), you can just ignore the exception.

geichelberger:
> I would prefer going with Matthias solution and keeping the exception. In case you made a mistake in the organization field (e.g. due to a typo), it gives you quick feedback. And if the faulty organization is intentional (because you know it will be added later on), you can just ignore the exception.

You can still log an error that the workflow organization does not exist, which would have the same effect.

gregorydlogan:
I would not expect to be able to register workflow definitions for an organization that does not exist, so @mtneug's approach seems like a better option.  There's nothing wrong with the other way, but it's inconsistent with the rest of Opencast's approach to things.

PR #4187, opened 2022-09-12
Dont run gh actions on dbot multiple times

### LINKED ISSUE #4171: Don't run tests on Dependabot push
Right now, tests run on `pull_request` and on `push`. This means that we have the same tests run multiple times with a higher likelyhood of one failing on Dependabot pull requests, since Dependabot creates a pull request from an upstream repository branch.

It would be great if we could not run the tests on `push` if the author is Dependabot.

Current configuration:
https://github.com/opencast/opencast/blob/d4d5c8094d49a47657a3582d213ece14f4320344/.github/workflows/test-opencast-build.yml#L3-L19

Arnei:
Aims to resolve #4171 by checking if the author is dependabot and if the event is a push. See #4171 for more details.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -20,6 +20,8 @@ on:
 
 jobs:
   build:
+    # Don't run dependabot pull requests multiple times
+    if: ${{ !(github.event_name == 'push' && github.actor == 'dependabot[bot]')  }}
     strategy:
       matrix:
         java:


.github/workflows/test-opencast-build.yml
```diff
@@ -20,6 +20,8 @@ on:
```
ziegenberg, line 20: Why not filter at the event level and just ignore the dependabot branches?

```yaml
on:
  push:
    branches-ignore:    
      - 'dependabot/**'
```

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-branches-and-tags

ziegenberg:
Guess, I'm too late :(

gregorydlogan:
You wanted to merge?

ziegenberg:
Na, I suggested an alternative approach.

.github/workflows/test-opencast-build.yml
Arnei: I did not go with that approach because I was under the impression that `branches-ignore` filters by target branch (e.g. develop instead of dependabot/**). But maybe I'm wrong about that, not a github actions expert at all. I would also prefer a filter at event level, so if @ziegenberg 's proposal works let's go with that.

ziegenberg: As we are filtering here on push events and not on pull request events, the target branch is `dependabot/**` and it should work. As long as I'm not completely mistaken...

ziegenberg: I filed an alternative PR here: #4207

PR #4206, opened 2022-09-14
Changed rebuild order and added rebuild indicator

Alaschinger:
The rebuild order has been changed, so that opencast can be used while the workflows are still updating. Still active workflows won't be updated during the rebuild, because they will automatically be updated when finished.

Since not all information in the admin UI is displayed correctly until the workflow rebuild has finished an index rebuild indicator has been added to the admin UI frontend. This indicator is getting the states of the repopulation services from the IndexRebuildService using a new endpoint. They are shown to the user within the header of the admin UI. If no rebuild is running the indicator is hidden.


gregorydlogan:
Looks like GH missed this conflict.

mtneug:
Since this PR requires changes to `etc/security/mh_default_org.xml`, it can probably no longer go into 13.0.

gregorydlogan:
Alternatively, as long as the counter doesn't break the ui if it can't get at the endpoint it still could...

Arnei:
Since Greg suggested this might still make it into 13.0 if the index rebuild indicator does not break the UI if the endpoint is unreachable (it does not), I took the liberty of rebasing this onto r/13.x while fixing the merge conflict. @mtneug what do you think?

mtneug:
If there is no error when running with an old `etc/security/mh_default_org.xml` than this is fine for me 👍

Arnei (approved):

modules/admin-ui-frontend/.jshintrc
```diff
@@ -1,16 +0,0 @@
-{
```
gregorydlogan: This file is required by the current state of `r/13.x`

gregorydlogan (changes requested):

modules/admin-ui-frontend/.jshintrc
gregorydlogan: To be clear:
```
[INFO] 16 01 2023 12:33:13.247:INFO [framework.detect-browsers]: The following browsers will be used: [ 'ChromeHeadless', 'FirefoxHeadless' ]
[INFO] 16 01 2023 12:33:13.437:WARN [filelist]: Pattern "/home/greg/opencast/upstream/modules/admin-ui-frontend/test/test/unit/**/*Helper.js" does not match any file.
```

modules/admin-ui-frontend/app/scripts/shared/services/indexRebuild.js
```diff
@@ -0,0 +1,115 @@
...
+      angular.forEach(data.data.services.service, function (service) {
+        var name = service.type;
+        var state = service.state;
+        var order = service.executionOrder;
+        RebuildStates.setState(name, state, order);
+        if (state != COMPLETED_STATE) {
+          rebuildServices.showRebuildStates = true;
+        }
+      });
+    }).catch(function (err) {
```
gregorydlogan, line 82: This exception will cause the UI element to incorrectly indicated that there's an issue if an old security xml is in use.  Two choices here:

1) Parse out the error, if it's a 503 then it's *probably* just old security and now you get to figure out how to tell the user that.
2) Move this PR into `develop`

gregorydlogan (changes requested):

modules/admin-ui-frontend/app/scripts/shared/services/indexRebuild.js
Arnei: Well then, rebasing on `develop` it is.

modules/admin-ui-frontend/.jshintrc
Arnei: Added it back, tests should run again now.

PR #3607, opened 2022-03-31
Extract JobDispatcher into its own class

gregorydlogan:
This pull request:
 - Extracts the JobDispatcher from inside the ServiceRegistry into its own class
 - Make the JobDispatcher a configurable OSGi service, such that dispatching can be enabled or disabled during runtime
 - Changes the *default* job dispatch behaviour to not dispatch jobs, and alters the node profiles to enable it by default on the expected nodes

This should *not* change current job dispatcher behaviour, aside from being able to enable or disable job dispatch on the fly.
Further work would likely use some kind of database locking strategy to negotiate between (surviving) nodes in a failover situation for who should be doing job dispatching.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] ~include migration scripts and documentation, if appropriate~
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4212, opened 2022-09-15
Usage of CPU load for job dispatching

berthob98:
This Pull Request adds the usage of the operating systems load value as metric for scheduling jobs. The hardware load values are a more realistic representation of a systems occupation then the current load values and there usage for managing loads promises more precise scheduling and better system usage.

The usage of hardware loads is an optional-in, disabled by default. If enabled the usage of hardware loads means two things:

1. When two nodes offer the the same service, the node with the lower system load is preferred

2. A maximum hardware load can be configured in the configuration file for each node. If a nodes maximum hardware load is reached this server stops accepting jobs till the nodes load gets below the maximum.

This mechanism is implemented by introducing two new configurable values managed in the service registry. A flag to enable the hardware load and a maximum hardware load value, defaulting to the number CPU of cores.
If the hardware load flag is set, the already existing logic of the service registry switches to using the hardware load values instead of the current load values. The existing logic uses the load values for two purposes:

1. Ordering the potential services a job could be dispatched to

2. Accepting a job on the service side, if the load doesn’t exceed the maximum load

The new load values are obtained from the OperatingSystemMXBean, if this bean is for some reasons not obtainable a log warning advices the user to switch to non hardware load scheduling.

Outlook:
With the usage of hardware loads nodes tend to decline jobs more reliable, if they are fully occupied. Together with an upcoming pull request, which introduces the preference of jobs from older workflows, this could stop Opencasts habit to try to work on all workflows at the same time under heavy loads.

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
@@ -333,7 +333,12 @@ public class JobDispatcher {
      */
     private void dispatchDispatchableJobs(EntityManager em, List<JpaJob> jobsToDispatch) {
       //Get the current system load
-      SystemLoad systemLoad = serviceRegistry.getHostLoads(em);
+      SystemLoad systemLoad;
+      //if enabled use the hardware loads instead
+      if (serviceRegistry.isHardwareLoadEnabled())
+        systemLoad = serviceRegistry.getSystemHardwareLoad();
+      else
+        systemLoad = serviceRegistry.getHostLoads(em);
 
       for (JpaJob job : jobsToDispatch) {
 
@@ -411,12 +416,29 @@ public class JobDispatcher {
             candidateServices = serviceRegistry.getServiceRegistrationsByLoad(jobType, services, hosts, systemLoad);
           }
 
+          //if enabled use the hardware loads to exclude fully occupied services
+          if (serviceRegistry.isHardwareLoadEnabled()) {
+            final List<ServiceRegistration> servicesToRemove = new ArrayList<ServiceRegistration>();
+
+            for (ServiceRegistration serviceReg : candidateServices) {
+              if (systemLoad.get(serviceReg.getHost()).getCurrentLoad() > systemLoad.get(serviceReg.getHost()).getMaxLoad())
+                servicesToRemove.add(serviceReg);
+            }
+            candidateServices.removeAll(servicesToRemove);
+            if (candidateServices.isEmpty())
+              logger.info("Dispatching of job {} delayed, all potential services of type '{}' are occupied", job, jobType);
+          }
+
           // Try to dispatch the job
           String hostAcceptingJob = null;
           try {
             hostAcceptingJob = dispatchJob(em, job, candidateServices);
             try {
-              systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              //if hardware load is enabled update the hardware load value for the host
+              if (!serviceRegistry.isHardwareLoadEnabled())
+                systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              else
+                systemLoad.get(hostAcceptingJob).setCurrentLoad(serviceRegistry.getHardwareLoadbyHost(hostAcceptingJob));
             } catch (NotFoundException e) {
               logger.info("Host {} not found in load list, cannot dispatch {} to it", hostAcceptingJob, job);
             }
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
@@ -70,6 +70,9 @@ import org.apache.http.HttpStatus;
 import org.apache.http.StatusLine;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpHead;
+import org.json.simple.JSONObject;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.InvalidSyntaxException;
 import org.osgi.framework.ServiceReference;
@@ -202,6 +205,8 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
 
   /** The http client to use when connecting to remote servers */
   protected TrustedHttpClient client = null;
+
+  private static final JSONParser jsonParser = new JSONParser();
 
   /** Default jobs limit during dispatching
    * (larger value will fetch more entries from the database at the same time and increase RAM usage) */
@@ -419,38 +424,42 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
     return hardwareLoad;
   }
 
+  /**
+   * Returns maximum hardware load for this host
+   */
   @Override
   public double getMaxHardwareLoad() throws ServiceRegistryException {
     return maxHardwareLoad;
   }
 
+  /**
+   *  Returns SystemLoad object with hardware load values for all hosts in the cluster (used by JobDispatcher class)
+   */
   protected SystemLoad getSystemHardwareLoad() {
 
     SystemLoad loads = new SystemLoad();
 
     List<HostRegistration> hostsHW = getHostRegistrations();
     for (HostRegistration hostHW : hostsHW) {
-      float load = getHardwareLoadbyHost(hostHW.toString());
-      float maxLoad = getMaxHardwareLoadbyHost(hostHW.toString());
-
+      // If we can't reach the hardware load endpoint return -1 as load and maxLoad
+      float load = (float)-1;
+      float maxLoad = (float)-1;
+      JSONObject jsonObject = reachHardwareLoadEndpoint(hostHW.toString());
+      if (jsonObject != null) {
+        load = ((Double)jsonObject.get("load")).floatValue();
+        maxLoad = ((Double)jsonObject.get("maxLoad")).floatValue();
+      }
       loads.addNodeLoad(new NodeLoad(hostHW.toString(), load, maxLoad));
     }
     return loads;
   }
 
-  protected float getHardwareLoadbyHost(String host) {
-    return reachHardwareLoadEndpoint(host, "/services/hardwareload");
-  }
+  /**
+   *  Helper function for getSystemHardwareLoad
+   */
+  protected JSONObject reachHardwareLoadEndpoint(String host) {
 
-  protected float getMaxHardwareLoadbyHost(String host) {
-    return reachHardwareLoadEndpoint(host, "/services/maxhardwareload");
-  }
-
-  //Helper function for getHardwareLoadbyHost and getMaxHardwareLoadbyHost
-  protected float reachHardwareLoadEndpoint(String host, String endpoint) {
-
-    float load = Float.parseFloat("-1.0");
-    String serviceUrl = UrlSupport.concat(host , endpoint);
+    String serviceUrl = UrlSupport.concat(host , "/services/hardwareload");
     HttpGet post = new HttpGet(serviceUrl);
     HttpResponse response = null;
 
@@ -463,14 +472,30 @@ public class ServiceRegistryJpaImpl implements ServiceRegistry, ManagedService {
         response.getEntity().writeTo(out);
         String responseString = out.toString();
         out.close();
-        load = Float.parseFloat(responseString);
+        JSONObject jsonObject = (JSONObject) jsonParser.parse(responseString);
+        return jsonObject;
       } else
         logger.warn("HTTP Status {} trying to reach {}", statusLine.getStatusCode(), serviceUrl);
     } catch (IOException e) {
       logger.warn("IOException trying to reach {}", serviceUrl);
       e.printStackTrace();
+    } catch (ParseException p) {
+      logger.warn("JSON ParseException trying to parse {}", response.toString());
+      p.printStackTrace();
     }
-    return load;
+    return null;
+  }
+
+  /**
+   *  Returns the current hardware load for a host (used by JobDispatcher class)
+   */
+  protected float getHardwareLoadbyHost(String host) {
+    JSONObject jsonObject = reachHardwareLoadEndpoint(host);
+    if (jsonObject != null) {
+      float load = ((Double)jsonObject.get("load")).floatValue();
+      return load;
+    }
+    return (float)-1;
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


berthob98:
I just saw Greg moved the job dispatcher into its own class 14 days ago. I will update this PR to resolve the merge conflict.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
```diff
@@ -687,6 +687,20 @@ public Response getMaxLoadOnNode(@QueryParam("host") String host) throws NotFoun
     }
   }
 
+  @GET
+  @Path("maxhardwareload")
+  @Produces(MediaType.TEXT_PLAIN)
```
geichelberger: Shouldn't it be better to return JSON?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
```diff
@@ -717,6 +731,20 @@ public Response getOwnLoad() {
     }
   }
 
+  @GET
+  @Path("hardwareload")
+  @Produces(MediaType.TEXT_PLAIN)
```
geichelberger, line 741: Same here.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -2196,10 +2315,13 @@ public List<ServiceRegistration> getServiceRegistrationsByLoad(String serviceTyp
...
       em = emf.createEntityManager();
-      SystemLoad loadByHost = getHostLoads(em);
+      SystemLoad loads = getHostLoads(em);
       List<HostRegistration> hostRegistrations = getHostRegistrations();
       List<ServiceRegistration> serviceRegistrations = getServiceRegistrationsByType(serviceType);
-      return getServiceRegistrationsByLoad(serviceType, serviceRegistrations, hostRegistrations, loadByHost);
+      // if enabled use the hardware loads instead
+      if (hardwareLoadEnabled)
+        loads = getSystemHardwareLoad();
+      return getServiceRegistrationsByLoad(serviceType, serviceRegistrations, hostRegistrations, loads);
```
geichelberger: As the loads variable gets overwritten, you could use the if statement to retrieve only the used load.

geichelberger (changes requested):

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
berthob98: I adapted this changes:

`
-  @Produces(MediaType.TEXT_PLAIN)
   @RestQuery(name = "maxhardwareload", description = "Returns the maximum hardware load this service registry can execute concurrently.",
       returnDescription = "The  maximum hardware load this service registry can execute", restParameters = {},
       responses = { @RestResponse(responseCode = SC_OK, description = "Maximum hardware load this service registry can execute.") })
   public Response getMaxHardwareLoad() {
     try {
-      return Response.ok(serviceRegistry.getMaxHardwareLoad()).build();
+      JSONArray list = new JSONArray();
+
+      list.add(serviceRegistry.getMaxHardwareLoad());
+
+      return Response.ok(list.toString()).build();
`

berthob98:
@geichelberger Thank you for your review!

I adapted all the requested changes.

The endpint now returns JSON...

```diff
   @GET
   @Path("maxhardwareload")
-  @Produces(MediaType.TEXT_PLAIN)
+  @Produces(MediaType.APPLICATION_JSON)
   @RestQuery(name = "maxhardwareload", description = "Returns the maximum hardware load this service registry can execute concurrently.",
       returnDescription = "The  maximum hardware load this service registry can execute", restParameters = {},
       responses = { @RestResponse(responseCode = SC_OK, description = "Maximum hardware load this service registry can execute.") })
   public Response getMaxHardwareLoad() {
     try {
-      return Response.ok(serviceRegistry.getMaxHardwareLoad()).build();
+      JSONArray list = new JSONArray();
+
+      list.add(serviceRegistry.getMaxHardwareLoad());
+
+      return Response.ok(list.toString()).build();
```

...and only the used load is now retrieve:
```diff
-      SystemLoad loads = getHostLoads(em);
-      List<HostRegistration> hostRegistrations = getHostRegistrations();
-      List<ServiceRegistration> serviceRegistrations = getServiceRegistrationsByType(serviceType);
+      SystemLoad loadByHost;
       // if enabled use the hardware loads instead
       if (hardwareLoadEnabled)
-        loads = getSystemHardwareLoad();
-      return getServiceRegistrationsByLoad(serviceType, serviceRegistrations, hostRegistrations, loads);
+        loadByHost = getSystemHardwareLoad();
+      else
+        loadByHost = getHostLoads(em);
+      List<HostRegistration> hostRegistrations = getHostRegistrations();
+      List<ServiceRegistration> serviceRegistrations = getServiceRegistrationsByType(serviceType);
+
+      return getServiceRegistrationsByLoad(serviceType, serviceRegistrations, hostRegistrations, loadByHost);
```

berthob98:
I also added a commit that got lost when I adopted the PR to changes brought by greg in #3607

```diff
     private void dispatchDispatchableJobs(EntityManager em, List<JpaJob> jobsToDispatch) {
       //Get the current system load
-      SystemLoad systemLoad = serviceRegistry.getHostLoads(em);
+      SystemLoad systemLoad;
+      //if enabled use the hardware loads instead
+      if (serviceRegistry.isHardwareLoadEnabled())
+        systemLoad = serviceRegistry.getSystemHardwareLoad();
+      else
+        systemLoad = serviceRegistry.getHostLoads(em);
 
       for (JpaJob job : jobsToDispatch) {
 
@@ -411,12 +416,29 @@ public class JobDispatcher {
             candidateServices = serviceRegistry.getServiceRegistrationsByLoad(jobType, services, hosts, systemLoad);
           }
 
+          //if enabled use the hardware loads to exclude fully occupied services
+          //if enabled use the hardware loads to exclude fully occupied services
+          if (serviceRegistry.isHardwareLoadEnabled()) {
+            final List<ServiceRegistration> servicesToRemove = new ArrayList<ServiceRegistration>();
+
+            for (ServiceRegistration serviceReg : candidateServices) {
+              if (systemLoad.get(serviceReg.getHost()).getCurrentLoad() > systemLoad.get(serviceReg.getHost()).getMaxLoad())
+                servicesToRemove.add(serviceReg);
+            }
+            candidateServices.removeAll(servicesToRemove);
+            if (candidateServices.isEmpty())
+              logger.info("Dispatching of job {} delayed, all potential services of type '{}' are occupied", job, jobType);
+          }
+
           // Try to dispatch the job
           String hostAcceptingJob = null;
           try {
             hostAcceptingJob = dispatchJob(em, job, candidateServices);
             try {
-              systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              //if hardware load is enabled update the hardware load value for the host
+              if (!serviceRegistry.isHardwareLoadEnabled())
+                systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              else
+                systemLoad.get(hostAcceptingJob).setCurrentLoad(serviceRegistry.getHardwareLoadbyHost(hostAcceptingJob));
```

and added the handling of an edge case that never occured during testing and is not related to the changes in this PR but I thought this should be handeld since we can't assume that NodeLoads will be found here:
```Diff
       NodeLoad nodeA = loadByHost.get(hostA);
       NodeLoad nodeB = loadByHost.get(hostB);
...
+      if (nodeA != null && nodeB != null) {
...
+      } else
+        return 0;  // undecided
```

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/endpoint/ServiceRegistryEndpoint.java
ziegenberg: @berthob98 I think you missed this instance... :)

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -333,7 +333,12 @@ public void run() {
      */
     private void dispatchDispatchableJobs(EntityManager em, List<JpaJob> jobsToDispatch) {
       //Get the current system load
-      SystemLoad systemLoad = serviceRegistry.getHostLoads(em);
+      SystemLoad systemLoad;
+      //if enabled use the hardware loads instead
+      if (serviceRegistry.isHardwareLoadEnabled())
```
gregorydlogan, line 338: K&R brackets please.  Surprised checkstyles isn't complaining about this.  This also applies in other spots for this PR, not going to comment on this further.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -411,12 +416,29 @@ private void dispatchDispatchableJobs(EntityManager em, List<JpaJob> jobsToDispa
...
+          if (serviceRegistry.isHardwareLoadEnabled()) {
+            final List<ServiceRegistration> servicesToRemove = new ArrayList<ServiceRegistration>();
+
+            for (ServiceRegistration serviceReg : candidateServices) {
+              if (systemLoad.get(serviceReg.getHost()).getCurrentLoad() > systemLoad.get(serviceReg.getHost()).getMaxLoad())
+                servicesToRemove.add(serviceReg);
+            }
+            candidateServices.removeAll(servicesToRemove);
+            if (candidateServices.isEmpty())
+              logger.info("Dispatching of job {} delayed, all potential services of type '{}' are occupied", job, jobType);
```
gregorydlogan, line 429: So we say all potential services are occupied, but then... continue and try and dispatch the job?  If `candidateServices` is empty, why not just `continue` here?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -384,6 +415,94 @@ public float getOwnLoad() {
...
+    return loads;
+  }
+
+  /**
+   *  Helper function for getSystemHardwareLoad
+   */
+  protected JSONObject reachHardwareLoadEndpoint(String host) {
+
+    String serviceUrl = UrlSupport.concat(host , "/services/hardwareload");
+    HttpGet post = new HttpGet(serviceUrl);
```
gregorydlogan, line 463: You'll want to use `TrustedHttpClient` for this, otherwise this should fail to connect since you haven't exposed the rest endpoint with ROLE_ANONYMOUS in the security xml

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -2957,15 +3107,18 @@ public int compare(ServiceRegistration serviceA, ServiceRegistration serviceB) {
...
-      //If the load factors are about the same, sort based on maximum load
-      if (Math.abs(nodeA.getLoadFactor() - nodeB.getLoadFactor()) <= 0.01) {
-        //NOTE: The sort order below is *reversed* from what you'd expect
-        //When we're comparing the load factors we want the node with the lowest factor to be first
-        //When we're comparing the maximum load value, we want the node with the highest max to be first
-        return Float.compare(nodeB.getMaxLoad(), nodeA.getMaxLoad());
-      }
-      return Float.compare(nodeA.getLoadFactor(), nodeB.getLoadFactor());
 
+      if (nodeA != null && nodeB != null) {
```
gregorydlogan, line 3111: I'm struggling to imagine why we need this check here.  Were you able to trigger a case where `nodeN` was null?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -411,12 +416,29 @@ private void dispatchDispatchableJobs(EntityManager em, List<JpaJob> jobsToDispa
...
           String hostAcceptingJob = null;
           try {
             hostAcceptingJob = dispatchJob(em, job, candidateServices);
             try {
-              systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              //if hardware load is enabled update the hardware load value for the host
+              if (!serviceRegistry.isHardwareLoadEnabled())
+                systemLoad.updateNodeLoad(hostAcceptingJob, job.getJobLoad());
+              else
+                systemLoad.get(hostAcceptingJob).setCurrentLoad(serviceRegistry.getHardwareLoadbyHost(hostAcceptingJob));
```
gregorydlogan, line 441: This is (IMO) quite risky.  You're making the assumption that the dispatched job starts imposing its load immediately, which is not true.  You're also now making a (slow!) HTTP request as part of the dispatch loop for *every single dispatched job*.  I suspect this will have drastic perf implications when deployed at scale.

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/ServiceRegistryJpaImpl.java
```diff
@@ -384,6 +415,94 @@ public float getOwnLoad() {
     return localSystemLoad;
   }
 
+  @Override
+  public double getHardwareLoad() {
+    OperatingSystemMXBean osBeanJava = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
```
gregorydlogan, line 420: So how does this handle cases where my current container context doesn't have access to all the host CPUs, but its load values still show the full system load?  Ie, my host right now has a load of 2.4, but my LXC container only gets two cores, and isn't actually doing anything right now anyway.  In theory it should accept jobs, but the code here wouldn't allow that if it's detecting two cores, and would allow far too much load if it's detecting all 16 cores.

gregorydlogan (changes requested):
A few issues spotted, but a good first crack at this.

mtneug:

berthob98:
I'm sorry for the late response, I didn't find time during the last weeks to further develope this PR. Thank you for the review and the motivating evaluation. I will take time on the weekend to adapt the requested changes and resolve the newly emerged conflicts.

gregorydlogan:
Is this going to move forward?  If not, next I'm passing through the PRs I'll close this.

KatrinIhler:
Same here - gonna close for now, this does not constitute a rejection, feel free to re-open at a later time.

PR #4232, opened 2022-09-21
Do not update default player components on legacy branches

lkiesow:
This patch removes the configuration that tells Dependabot to update components of Paella Player 6 – Opencast's default player – on legacy branches.

The reasoning is that we know that dependency update pull requests are often not well tested but get merged nonetheless. Doing so on `develop` gives us time to react if something breaks. Doing this on a release branch comes with a high risk of breaking the default player in a release. Doing so in a legacy branch even increases the risk since less people will test these.

There is usually no need for updating dependencies at all in the release branches unless it is a security update or the dependency is the Player itself, which should get proper attention and manual testing.

I don't mind auto-updating Paella Player 7, but only since that one is not the default and specifically marked as “beta”.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,7 +34,6 @@ updates:
   directory: "/modules/engage-paella-player"
   schedule:
     interval: daily
-  target-branch: "r/11.x"
   open-pull-requests-limit: 10
 - package-ecosystem: npm
   directory: "/modules/engage-paella-player-7"


mliradelc (approved):
The reasoning I've added `r/11.x` it was because it still "supported", but Lars is right in PP 6 Dependabot updates are not deeply tested for a legacy release. 

The legacy release should only update, as Lars says, bug fixes and security updates only.

I approve this PR.

PR #4237, opened 2022-09-23
Added maximum line length configuration option to Azure Transcription Service

Arnei:
In the Azure Transcription service, the text received from a recognized event can be rather long, longer than a subtitle line should reasonably be. This PR adds new config options to the service that allow you to limit the size of a line to a number of characters. Post processing in the service will then try to adhere to limit by splitting long lines into multiple shorter ones (i.e. without breaking up words), and dividing the display times between them evenly.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
+++ b/etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
@@ -32,6 +32,17 @@ region=
 # Default: ANY
 #encoding.extension=ANY
 
+# Set this to true if you want to limit the maximum size of a line in the subtitle file.
+# Azure does not do this automatically, and sometimes produces very long lines.
+# Default: false
+split.text=false
+
+# The maximum amount of characters a single line in the subtitle file should contain.
+# Can go over the set number to avoid cutting words in half.
+# Only takes effect if "split.text" is set to true.
+# Default: 100
+split.text.line.size=100
+
 # Workflow to be executed when results are ready to be attached to media package.
 #workflow=microsoft-azure-attach-transcripts
 
--- a/etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
+++ b/etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
@@ -35,13 +35,13 @@ region=
 # Set this to true if you want to limit the maximum size of a line in the subtitle file.
 # Azure does not do this automatically, and sometimes produces very long lines.
 # Default: false
-split.text=false
+#split.text=false
 
 # The maximum amount of characters a single line in the subtitle file should contain.
 # Can go over the set number to avoid cutting words in half.
 # Only takes effect if "split.text" is set to true.
 # Default: 100
-split.text.line.size=100
+#split.text.line.size=100
 
 # Workflow to be executed when results are ready to be attached to media package.
 #workflow=microsoft-azure-attach-transcripts
--- a/modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
+++ b/modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
@@ -292,7 +292,6 @@ public class MicrosoftAzureTranscriptionService extends AbstractJobProducer impl
     if (isSplitTextOpt.isSome()) {
       splitText = Boolean.valueOf(isSplitTextOpt.get());
     }
-    logger.info("Long text will be split at {} characters", splitText);
 
     // When splitting text into multiple cues, how many characters should each cue have at most
     Option<String> splitTextLineSizeOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT_LINE_SIZE);
@@ -303,6 +302,10 @@ public class MicrosoftAzureTranscriptionService extends AbstractJobProducer impl
         throw new ConfigurationException("Invalid configuration for split text line size. Please check your"
                 + "configuration");
       }
+    }
+
+    if (splitText) {
+      logger.info("Long text will be split at {} characters", splitTextLineSize);
     }
 
     // Workflow to execute when getting callback (optional, with default)
--- a/modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
+++ b/modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
@@ -87,11 +87,13 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.URI;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -178,6 +180,8 @@ public class MicrosoftAzureTranscriptionService extends AbstractJobProducer impl
   public static final String NOTIFICATION_EMAIL_CONFIG = "notification.email";
   public static final String CLEANUP_RESULTS_DAYS_CONFIG = "cleanup.results.days";
   public static final String ENCODING_EXTENSION = "encoding.extension";
+  public static final String SPLIT_TEXT = "split.text";
+  public static final String SPLIT_TEXT_LINE_SIZE = "split.text.line.size";
 
   /**
    * Service configuration values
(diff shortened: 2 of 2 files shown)


modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -280,6 +286,24 @@ public void activate(ComponentContext cc) {
...
+    logger.info("Long text will be split: {}", splitText);
+
+    // When splitting text into multiple cues, how many characters should each cue have at most
+    Option<String> splitTextLineSizeOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT_LINE_SIZE);
+    if (splitTextLineSizeOpt.isSome()) {
+      try {
+        splitTextLineSize = Integer.parseInt(splitTextLineSizeOpt.get());
+      } catch (NumberFormatException e) {
+        // Use default
+        logger.warn("Invalid configuration for split text line size. Default used instead: {}", splitTextLineSize);
```
lkiesow: Please don't do that. Please let the service fail instead. If an admin misconfigures a service, the service should not just do something else. The admin clearly trued to set a specific setting and would probably assume everything is well if things seem to work. That is, until he later notices that the code ignored his setting.

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -520,12 +544,15 @@ void recognizeContinuous(SpeechRecognizer speechRecognizer, String jobId, String
     speechRecognizer.recognized.addEventListener((s, e) -> {
       if (ResultReason.RecognizedSpeech == e.getResult().getReason() && e.getResult().getText().length() > 0) {
         sequenceNumber[0]++;
-        String text = captionFromSpeechRecognitionResult(sequenceNumber[0], e.getResult());
+        List<String> texts = captionFromSpeechRecognitionResult(sequenceNumber[0], e.getResult());
+        sequenceNumber[0] += texts.size() - 1;
```
lkiesow, line 552: Why increment the value above only to subtract 1 down here?

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -520,12 +544,15 @@ void recognizeContinuous(SpeechRecognizer speechRecognizer, String jobId, String
...
 
-        try {
-          writeToTmpFile(text, jobId);
-        } catch (IOException ioException) {
-          logger.error(String.format("Could not write to tmp file: %s", text));
+        for (String text : texts) {
+          try {
+            writeToTmpFile(text, jobId);
+          } catch (IOException ioException) {
+            logger.error(String.format("Could not write to tmp file: %s", text));
```
lkiesow: Don't use `format` in a logger statement
```suggestion
            logger.error("Could not write to tmp file: {}", text);
```

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
Arnei: We are never really substracting 1, since `text.size() - 1` is guaranteed to be at least 0 (there will always be at least one result).

`sequenceNumber` is supposed to enumerate the cues in the subtitle file. Previously `captionFromSpeechRecognitionResult(int, SpeechRecognitionResult)` would always return exactly one line, but now it can return any positive number of lines. So to keep `sequenceNumber` correct, the number of additional lines needs to be added.

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
Arnei: Thanks!

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
Arnei: Did not think about, thanks for pointing it out!

etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
```diff
@@ -32,6 +32,17 @@ region=
...
+# Set this to true if you want to limit the maximum size of a line in the subtitle file.
+# Azure does not do this automatically, and sometimes produces very long lines.
+# Default: false
+split.text=false
+
+# The maximum amount of characters a single line in the subtitle file should contain.
+# Can go over the set number to avoid cutting words in half.
+# Only takes effect if "split.text" is set to true.
+# Default: 100
+split.text.line.size=100
```
gregorydlogan: Defaults are usually commented out.

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -280,6 +287,24 @@ public void activate(ComponentContext cc) {
       logger.info("Default '{}' audio encoding will be used", compressedAudioFormat);
     }
 
+    // If text should be split into multiple cues
+    Option<String> isSplitTextOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT);
+    if (isSplitTextOpt.isSome()) {
+      splitText = Boolean.valueOf(isSplitTextOpt.get());
+    }
+    logger.info("Long text will be split: {}", splitText);
```
gregorydlogan:
```suggestion
    logger.info("Long text will be split at {} characters", splitText);
```

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -280,6 +287,24 @@ public void activate(ComponentContext cc) {
...
+
+    // When splitting text into multiple cues, how many characters should each cue have at most
+    Option<String> splitTextLineSizeOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT_LINE_SIZE);
+    if (splitTextLineSizeOpt.isSome()) {
+      try {
+        splitTextLineSize = Integer.parseInt(splitTextLineSizeOpt.get());
+      } catch (NumberFormatException e) {
+        throw new ConfigurationException("Invalid configuration for split text line size. Please check your"
+                + "configuration");
```
gregorydlogan, line 303:
```suggestion
                + "configuration.  Using default split length of " + splitTextLineSize + " characters");
```

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -280,6 +287,24 @@ public void activate(ComponentContext cc) {
...
+    // If text should be split into multiple cues
+    Option<String> isSplitTextOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT);
+    if (isSplitTextOpt.isSome()) {
+      splitText = Boolean.valueOf(isSplitTextOpt.get());
+    }
+    logger.info("Long text will be split: {}", splitText);
+
+    // When splitting text into multiple cues, how many characters should each cue have at most
+    Option<String> splitTextLineSizeOpt = OsgiUtil.getOptCfg(cc.getProperties(), SPLIT_TEXT_LINE_SIZE);
+    if (splitTextLineSizeOpt.isSome()) {
```
gregorydlogan, line 298: IMO, if we're not splitting lines we shouldn't bother logging the split length either.  I would probably move :295 to :302.  Then we're either:
- not splitting at all -> no logging of split length
- successfully set the split length -> new length logged
- unsuccessfully set -> error message, with default split length.

gregorydlogan (changes requested):
Looks good aside from minor nits.

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
Arnei: But we are not using the default value, we're throwing an exception, right?

And throwing an exception make sense, since if the user went out of their way to modify the default they probably don't want the default, so we should notify them at the earliest opportunity?

gregorydlogan: Derp, yeah you're right. This exception should also prevent the service from activating, so there's no default to use anyway ;D

PR #4239, opened 2022-09-23
Added automatic language detection to Azure Transcription Service

Arnei:
Added automatic language detection to Azure Transcription Service

The Speech-to-Text Service by Microsoft Azure offers to detect the spoken language for you, allowing you to request transcription without setting a specific language. This PR allows users to use that functionality through global configuration or WOH keys.

The detected language is also stored in a file in the workspace, so the Attach WOH can later know which language was detected and set the file flavor to include the language code if the user so wishes. It does not set it as a workflow property because the transcription (and thus the language detection) run asynchronously, so technically it is uncertain whether the workflow which started the transcription is still active when the language detection is done. Still, this is quite a roundabout way of getting the language to the Attach WOh, so if there are better ways to solve this problem (or if my explanation is just wrong), please let me know.

(Also includes #4238 because I'm too lazy remove it from the commit history)

* [ ] have a concise title
* [ ] pass automated tests


etc/org.opencastproject.transcription.microsoftazure.MicrosoftAzureTranscriptionService.cfg
```diff
@@ -7,9 +7,21 @@ region=
...
+# Setting this option will prompt Azure to detect the spoken language by itself. Language is detected once at the
+# start of the media file. Auto detection will take precedence over a defined language.
+# Default: false
+#auto.detect.language=true
+
+# A list of language codes. The Azure language auto detection will pick it's detected language from one of these.
+# The language auto detection cannot detect any languages not specified in this list.
+# The list needs to have at least one element and can have at most four elements.
+# List of supported languages: https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/language-support?tabs=language-identification
+# Default: None
```
gregorydlogan, line 23: Stated default does not match default below.

docs/guides/admin/docs/workflowoperationhandlers/microsoft-azure-start-transcription-woh.md
```diff
@@ -18,20 +18,40 @@ Parameter Table
 |source-tag|The flavor of the file to be sent for translation.|EMPTY|transcript|
 |skip-if-flavor-exists|If this flavor already exists in the media package, skip this operation.<br/>To be used when the media package already has a transcript file. Optional|false|captions/vtt+en-us|
 |language-code|The language code to use for the transcription. Optional. If set, it will override the configuration language code|EMPTY|en-US, [supported languages](https://docs.microsoft.com/de-de/azure/cognitive-services/speech-service/language-support?tabs=speechtotext#speech-to-text)|
+|auto-detect-language|Activate automatic language detection by Azure. Optional. Overrides the language set in language-code. If set, will override the value in the configuration|false|true
```
gregorydlogan:
```suggestion
|auto-detect-language|Activate automatic language detection by Azure. Optional. Overrides the language set in \`language-code\`. If set, will override the value in the configuration|false|true
```

Unsure of the correct Markdown, but `language-code` should look different than normal text here.

docs/guides/admin/docs/workflowoperationhandlers/microsoft-azure-start-transcription-woh.md
```diff
@@ -18,20 +18,40 @@ Parameter Table
...
+|auto-detect-languages|A list of language codes. The Azure language auto detection will pick it's detected language from one of these. The language auto detection cannot detect any languages not specified in this list. The list needs to have at least one element and can have at most four elements.|EMPTY|en-US,de-DE,it-IT
```
gregorydlogan:
```suggestion
|auto-detect-languages|A list of language codes. Restricts Azure's language autodetection to this list of languages. The list needs to have at least one element and can have at most four elements.|EMPTY|en-US,de-DE,it-IT
```

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/MicrosoftAzureAttachTranscriptionOperationHandler.java
```diff
@@ -93,21 +97,41 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, Jo
...
+        Map<String, Object> returnValues = service.getReturnValues(mediaPackage.getIdentifier().toString(), jobId);
+        autoDetectedLanguage = (String) returnValues.get("autoDetectedLanguage");
+      } catch (NullPointerException e) {
+        logger.warn("Key missing in return values", e);
+      } catch (TranscriptionServiceException e) {
+        logger.warn("Something went wrong when trying to receive return values", e);
+      }
 
       // Set the target flavor
+      if (language != null) {
```
gregorydlogan, line 121: In the case(s) where the language is blank, or missing, it's going to create flavours like `foo/bar+`.  I'm assuming Opencast is smart enough not to break on that, but I personally would expect the `+` character to disappear.  Is `parseFlavor` going to strip that trailing `+` for us?

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
```diff
@@ -280,6 +296,27 @@ public void activate(ComponentContext cc) {
...
+    // If language should be detected automatically
+    Option<String> isAutoDetectLanguageOpt = OsgiUtil.getOptCfg(cc.getProperties(), IS_AUTO_DETECT_LANGUAGE);
+    if (isAutoDetectLanguageOpt.isSome()) {
+      defaultIsAutoDetectLanguage = Boolean.valueOf(isAutoDetectLanguageOpt.get());
+    }
+    logger.info("Automatically detecting language is globally enabled: {}", defaultIsAutoDetectLanguage);
+
+    // Which languages should be used in automatic language detection
+    Option<String> autoDetectLanguagesOpt = OsgiUtil.getOptCfg(cc.getProperties(), AUTO_DETECT_LANGUAGES);
+    if (autoDetectLanguagesOpt.isSome()) {
```
gregorydlogan, line 314: I would be tempted to hide this behind a check on `defaultIsAutoDetectLanguage` since we don't care about these checks unless the detection is turned on.

docs/guides/admin/docs/workflowoperationhandlers/microsoft-azure-start-transcription-woh.md
Arnei: I don't agree with this rewording. To me, "restricts" implies that the autodetection would be "unrestricted" if `auto-detect-languages` is not specified, which is incorrect. If no languages are specified, the auto detection (and thus, the workflow operation) will fail. (...unless there is a list defined in the global config I believe)

How about:
"A list of language codes. MUST be set if not set already in configuration and will override configuration. The Azure language auto detection chooses from the given list and cannot detect any language not in the given list. The list needs to have at least one element and can have at most four elements."

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/MicrosoftAzureAttachTranscriptionOperationHandler.java
Arnei: Afaik `parseFlavor` only cares about `/` and just ignores other special characters.

Personally, I woudn't expect the `+` to dissappear. The whole replacing `___` with something else already feels pretty hacky and not great to work with from a workflow creator perspective, implicitly removing other characters would probably cause them even more headache?

But maybe there is a reason a flavor cannot end in a `+` (or other special charactrs?) that I'm forgetting?

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
Arnei: I think currently the code is trying to accommodate for a scenario where the list of languages is defined in the global config, and the workflow operation later only decides whether automatic detection should be used or not, but does not specify the list of languages (in which case the list of languages from the global config would be used). 

Maybe having the same config in the global config and as a workflow operation key is just a bad idea.

docs/guides/admin/docs/workflowoperationhandlers/microsoft-azure-start-transcription-woh.md
gregorydlogan: Ok, I misunderstood what this config item meant.  Yours is better.  What about "MUST be set if not set already in **service** configuration"?  I was confused which configuration you meant, and I can see adopters being confused as well.

modules/transcription-service-workflowoperation/src/main/java/org/opencastproject/transcription/workflowoperation/MicrosoftAzureAttachTranscriptionOperationHandler.java
gregorydlogan: AFAIK there's no reason it can't.  I guess my approach would be that seeing a flavour of `foo/` would mean that something is wrong and there's a subflavour missing, so seeing `foo/bar+` would be equally an error case.

I suspect you're right in that it won't *actually* break anything though, and I agree that the current approach isn't great - although I don't have a better suggestion.  Let's just leave this as is.

modules/transcription-service-microsoft-azure/src/main/java/org/opencastproject/transcription/microsoftazure/MicrosoftAzureTranscriptionService.java
gregorydlogan: No, I just missed the part where it's settable in both.  I can *absolutely* see a use in setting this in both places.  One's a default, one is for handling a hypothetical overseas tenant where they speak an entirely different set of languages.

PR #4240, opened 2022-09-23
Group dependabot Paella 7 commits

mliradelc:
This is a GitHub action that I created to group all Paella 7 dependency commits.

Because of the way that Paella 7 is build, Dependabot makes a new PR for each new module instead to make a digest PR with all the modules. This makes difficult for testing.

This GitHub action uses https://github.com/Legal-and-General/dependabot-batcher

To be run, needs a GitHub API secret with `repo` and `workflow` rights. 

This is my first GitHub action that I develop for Opencast, please let me know any improvements that can be done.

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
In general this looks good to me. Does it close the PRs it's batching as it goes?

.github/workflows/dependabot-batcher.yml
```diff
@@ -0,0 +1,38 @@
+# Dependabot Batcher v1.0.1
+
+name: 'Dependabot Batcher'
+
+# Controls when the workflow will run
+on:
+  push:
+    branches: [ "r/12.x", "develop" ]
+  pull_request:
+    branches: [ "r/12.x", "develop" ]
```
lkiesow: This would run on all pushes to the repository and opened pull requests. You probably just want to react to either of

- pushed to `dependabot/**` branches
- pull requests created or updated by dependabot
For example, something like:

```yaml
push:
  branches:
    - 'dependabot/**'
```

.github/workflows/dependabot-batcher.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  dependabot-batcher:
+    name: 'Combine dependabot PRs'
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repo'
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.DEPENDABOT_BATCHER }}
```
lkiesow: If you react to pushes, you don't need this.

.github/workflows/dependabot-batcher.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: 'Set Node Version'
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.8.0
```
lkiesow: Does `dependabot-batche` need this?

.github/workflows/dependabot-batcher.yml
```diff
@@ -0,0 +1,38 @@
...
+
+  # Batches Dependabot PRs into one by merging them into a combined branch, then raising a new PR
+      - name: 'Dependabot Batcher'
+        if: ${{ contains(github.head_ref, '/modules/engage-paella-player-7') }}
```
lkiesow: Instead of checking this here, you can also only run on pushes or pull requests which change this patch:

https://github.com/opencast/opencast/blob/34cf94c108afa249862916829aa8b4085388b60c/.github/workflows/test-opencast-build.yml#L12-L15

.github/workflows/dependabot-batcher.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: Legal-and-General/dependabot-batcher@v1.0.1
+        with:
+          token: ${{ secrets.DEPENDABOT_BATCHER }} #required
+          baseBranchName: 'r/12.x' #optional
+          batchBranchName: 'dep-batcher-updates' #optional
+          batchPullTitle: 'Dependabot Batcher Updates' #optional
```
lkiesow: Maybe include Paella, since it's for Paella 7 specifically?

```suggestion
          batchBranchName: 'paella7-batcher-updates'
          batchPullTitle: 'Batch Dependabot Updates for Paella 7'
```

lkiesow:
> In general this looks good to me. Does it close the PRs it's batching as it goes?

Dependabot would close the ones automatically if the update is no longer necessary

.github/workflows/dependabot-batcher.yml
mliradelc: To be honest, I'm not very sure, I based my code in the implementation from the creators of this action https://github.com/Legal-and-General/dependabot-batcher/blob/main/.github/workflows/dependabot.scheduled.yml

.github/workflows/dependabot-batcher.yml
mliradelc: I like that change, makes easier to know for what is the action. Since is only for PP 7 makes sense

.github/workflows/dependabot-batcher.yml
mliradelc: That is a more efficient approach, this action is only for dependabot, thanks, I just changed it.

.github/workflows/dependabot-batcher.yml
mliradelc: From the developers of this action. They run in a scheduled way instead of for every push.

```
on:
  schedule:
    - cron:  '0 9 * * 1,3'
```

I think that is the reason to use a token. I will try with main repo on how behaves without this `token`

.github/workflows/dependabot-batcher.yml
mliradelc: I changed to only check where paella 7 resides

mliradelc:
~Please don't merge yet, I'm testing right now this code~, don't hesitate to write suggestions on how to improve it :smile:


Works like a charm:
https://github.com/mliradelc/opencast/pull/607

image

.github/workflows/dependabot-batcher.yml
mliradelc: You were right!, it's not necessary for the action to work.

mliradelc:
I made simpler the action now is faster.

.github/workflows/dependabot-batcher.yml
mliradelc: Works well without the `checkout repo` step.

mliradelc:
I changed the secret from PAT to GitHub token, also, I added pushes as an event to trigger the action.

gregorydlogan:
Not to detrail this whole PR, but [this thread](https://twitter.com/bitandbang/status/1576161582899286016) brought up [Renovate](https://www.mend.io/free-developer-tools/renovate/), which sounds like it does a lot of this and might be generally better than our Dependabot. Thoughts?

gregorydlogan:
I'm going to merge this since I'm tired of looking at the bloated PR list.

PR #4248, opened 2022-09-27
Join index rebuild of scheduler to assetmanager

Alaschinger:
Builds on PR #4120

- The repopulate function from the scheduler service has been removed.
- The index rebuild of scheduled events will now be proceessed as a part of the asset manager rebuild.
- A new SchedulerIndexData object will be created, selecting the relevant data from the scheduler database table.
- Within the rebuild of the assetmanager the scheduler data is checked for the given mediapackage id.
- If the event is a scheduled event it will by extended by the scheduler data, if not it will be processed as before.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -58,6 +58,7 @@ import org.opencastproject.assetmanager.api.storage.Source;
 import org.opencastproject.assetmanager.api.storage.StoragePath;
 import org.opencastproject.assetmanager.impl.persistence.AssetDtos;
 import org.opencastproject.assetmanager.impl.persistence.Database;
+import org.opencastproject.assetmanager.impl.persistence.SchedulerIndexData;
 import org.opencastproject.assetmanager.impl.persistence.SnapshotDto;
 import org.opencastproject.assetmanager.impl.query.AQueryBuilderImpl;
 import org.opencastproject.assetmanager.impl.query.AbstractADeleteQuery;
@@ -97,6 +98,7 @@ import org.opencastproject.security.api.User;
 import org.opencastproject.security.util.SecurityUtil;
 import org.opencastproject.util.Checksum;
 import org.opencastproject.util.ChecksumType;
+import org.opencastproject.util.DateTimeSupport;
 import org.opencastproject.util.MimeTypes;
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.RequireUtil;
@@ -134,9 +136,11 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -949,6 +953,10 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
 
       final AQueryBuilder q = createQuery();
       final RichAResult r = enrich(q.select(q.snapshot()).where(q.version().isLatest()).run());
+
+      final Map<String, SchedulerIndexData> schedulerData;
+      schedulerData = getDatabase().getSchedulerIndexData();
+
       final int total = r.countSnapshots();
       logIndexRebuildBegin(logger, index.getIndexName(), total, "snapshot(s)");
       int current = 0;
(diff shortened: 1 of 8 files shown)


etc/custom.properties
```diff
@@ -113,16 +113,16 @@ org.opencastproject.download.directory=${org.opencastproject.storage.dir}/downlo
...
 # Defaults:
 #  .url=jdbc:h2:${org.opencastproject.storage.dir}/db
 #  .user=sa
 #  .pass=sa
-#org.opencastproject.db.jdbc.url=jdbc:mariadb://localhost/opencast?useMysqlMetadata=true
-#org.opencastproject.db.jdbc.user=opencast
-#org.opencastproject.db.jdbc.pass=dbpassword
+org.opencastproject.db.jdbc.url=jdbc:mariadb://localhost/opencast?useMysqlMetadata=true
+org.opencastproject.db.jdbc.user=opencast
+org.opencastproject.db.jdbc.pass=dbpassword
```
mtneug: This probably needs to be reverted?

lkiesow: Probably not, what you wanted to commit.

modules/asset-manager-impl/pom.xml
```diff
@@ -66,6 +66,11 @@
       <artifactId>opencast-authorization-manager</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-scheduler-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
lkiesow: We should think about possible side effects here.
Thinking on potential for circular dependencies

lkiesow (changes requested):

etc/custom.properties
Alaschinger: Oh. Yes, this should be reverted. It was left from debugging.

modules/asset-manager-impl/pom.xml
Alaschinger: I could easily remove this dependency. It was only used to get the string value of the the recording state 'unknown', which isn't expected to change.

mtneug:

Arnei:
Index rebuild is throwing an error if there never was a scheduled event. Unsure on how to best handle this.

```
2022-12-05 08:50:36,928 | WARN  | (Slf4JLogger:100) - Error: 1146-42S02: Table 'opencast.oc_scheduled_extended_event' doesn't exist
2022-12-05 08:50:36,931 | ERROR | (IndexEndpoint:163) - Repopulating the index failed
javax.persistence.PersistenceException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.8.v20201217-ecdf3c32c4): org.eclipse.persistence.exceptions.DatabaseException
Internal Exception: java.sql.SQLSyntaxErrorException: (conn=4) Table 'opencast.oc_scheduled_extended_event' doesn't exist
Error Code: 1146
Call: SELECT mediapackage_id, capture_agent_id, recording_state, presenters, start_date, end_date FROM oc_scheduled_extended_event
Query: ResultSetMappingQuery(name="SchedulerIndexData.getAll" sql="SELECT mediapackage_id, capture_agent_id, recording_state, presenters, start_date, end_date FROM oc_scheduled_extended_event")
	at org.eclipse.persistence.internal.jpa.QueryImpl.getDetailedException(QueryImpl.java:391) ~[?:?]
	at org.eclipse.persistence.internal.jpa.QueryImpl.executeReadQuery(QueryImpl.java:264) ~[?:?]
	at org.eclipse.persistence.internal.jpa.QueryImpl.getResultList(QueryImpl.java:482) ~[?:?]
	at org.opencastproject.db.Queries$TypedQueriesBase.lambda$findAll$4(Queries.java:150) ~[?:?]
	at org.opencastproject.db.DBSessionImpl.execChecked(DBSessionImpl.java:79) ~[?:?]
	at org.opencastproject.db.DBSessionImpl.exec(DBSessionImpl.java:66) ~[?:?]
	at org.opencastproject.assetmanager.impl.persistence.Database.getSchedulerIndexData(Database.java:366) ~[?:?]
[...]
```

gregorydlogan:
Looking at [Database:366](https://github.com/opencast/opencast/blob/a7bccdd47f91ebbc147a6533693a96c0ca82d8b3/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java#L366), I'm not *entirely* sure what `insidePersistenceContextCheck` is supposed to be doing, but I do know that [the SnapshotDTO](https://github.com/opencast/opencast/blob/a7bccdd47f91ebbc147a6533693a96c0ca82d8b3/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java#L165) is the only place it's used upstream.  And what the check in `Database` is doing is ensuring that the DTO is being used within a JPA context.  Looking at the diff, you're touching DTO fields outside of an obvious JPA context.  Try that?

Arnei:
For posterity: I've talked with a Greg a bit and in the end he suggested that `getSchedulerIndexData()` should maybe not live where it lives right now, but ultimately didn't arrive at a solution.

KatrinIhler:
Talked to Julian and as there are no current plans to do work on this, I'm gonna close this for now.

PR #4295, opened 2022-10-07
Improve trigger on Dependabot Batcher for Paella7

### LINKED ISSUE #4279: Paella 7 Dependabot batcher is not getting triggered automatically
**Describe the bug**
The new Dependabot batcher #4240, is not getting automatically triggered when a new Dependabot paella 7 bump PR is created.

**To Reproduce**
1. Wait for any Dependabot PRs that is aimed to Paella 7
2. There is no automatic trigger on the automation list

**More information**

When I was working into this, the action was successfully triggered when the push or pull request it was being targeted to the desired branch, but it was aiming all PRs, this is the code.  Could it be another way to aim to the exact path of a triggering?

```
# Controls when the workflow will run
on:
  push:
    branches: [ "r/12.x", "develop" ]
  pull_request:
    branches: [ "r/12.x", "develop" ]
```

mliradelc:
After the set in action of the new GitHub action to group all automatic PRs for Paella 7, we notice that they are not being triggered.

After looking into another actions, I notice that the trigger path doesn't start with a `/` at the beginning of each path. This PR amends that.

Closes #4279 

* [x] have a concise title
* [X] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] pass automated tests

### DIFF
--- a/.github/workflows/dependabot-batcher.yml
+++ b/.github/workflows/dependabot-batcher.yml
@@ -6,10 +6,10 @@ name: 'Dependabot Batcher for Paella 7'
 on:
   pull_request:
     paths:
-      - '/modules/engage-paella-player-7/**'
+      - 'modules/engage-paella-player-7/**'
   push:
     paths:
-      - '/modules/engage-paella-player-7/**'
+      - 'modules/engage-paella-player-7/**'
 
 
   # Allows you to run this workflow manually from the Actions tab
--- a/.github/workflows/dependabot-batcher.yml
+++ b/.github/workflows/dependabot-batcher.yml
@@ -15,6 +15,10 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
   dependabot-batcher-paella-7:
     name: 'Combine dependabot PRs'


.github/workflows/dependabot-batcher.yml
```diff
@@ -6,10 +6,10 @@ name: 'Dependabot Batcher for Paella 7'
 on:
   pull_request:
     paths:
-      - '/modules/engage-paella-player-7/**'
+      - 'modules/engage-paella-player-7/**'
```
lkiesow, line 9: This should always fail since you cannot access write secrets when acting on a `pull_request` trigger. This makes sense since the workflow is executed on whatever any user changes. Allowing something like that would mean that someone could just modify this workflow in a pull request and e.g. delete all of Opencast's code (or do other nasty things).

But you don't actually need `pull_request` at all, since whenever Dependabot updates a pull request, it will also push code to the repository. That's why push should be sufficient.

lkiesow (changes requested):
The changes make sense and are good, but the `pull_request` should still cause trouble.

.github/workflows/dependabot-batcher.yml
mliradelc: That's right, I was yesterday talking about it in Matrix over this problem. When you manually execute the action, the token gets write-all permissions, but not when is automatically triggered.

Fortunately, there is a fix: I can add the permissions' parameter in with I can add the writing rights to the token. Still, I haven't found what right is the correct one to enable. (I'm not comfortable in enable writing rights to all scopes)

mliradelc:
I added the write permissions that GitHub token needs. Now works when is called by dependabot.

I figured out what causes the merge conflict every time. Is the way that PP7 is being developed. Each repository of paella calls the another repository when builds the package.json, instead to be like a tree schema, where you have a root repo that is first updated and then update the rest.

The ideal should be have a merge conflict solver for the package.json whose strategy should be to have the higher version of everything. I've found some projects [like this one](https://github.com/enricoaleandri/merge-package-json) that could it be useful to integrate to this action but I don't know how.

.github/workflows/dependabot-batcher.yml
mliradelc: Solved!

mliradelc:
Now is working better, in my personal repository, the action is automatically merging into the paella branch only and closing those PRs from dependabot.

gregorydlogan:
I'm going to merge this to see if it handles the current batch of PRs.

PR #4309, opened 2022-10-10
Add captions to Tobira harvest API (plus: avoid using deprecated method for creators.)

LukasKalbertodt:
See both commits! The main goal of this PR is to expose caption/subtitle information so that Tobira can show those in the player. Unfortunately, there are no strict rules how to attach captions to events, so this current code just checks multiple sources and takes all captions it can find. This is not optimal and it would be better if there were established best practices or even rules. This code is a bit flying blind. However, this harvest API does not guarantee any stability, so we can just improve the logic in the future.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
@@ -48,6 +48,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
```diff
@@ -241,6 +250,56 @@ private static List<Jsons.Val> assembleTracks(SearchResultItem event, MediaPacka
...
+        .flatMap(e -> {
+          final var languages = e.getValue();
+          final var uri = e.getKey();
+
+          // If this caption URL had no language associated with it, we push a
+          // `null` language to emit this caption at all. Otherwise, all
+          // languages are output as a separate track.
+          if (languages.isEmpty()) {
+            languages.add(null);
+          }
```
lkiesow, line 293: If `lang` is a set anyway, why add `null` to the set and not just transmit an empty set?

LukasKalbertodt: I `flatMap` the languages below that and if the set were empty, the subtitle wouldn't be sent at all. I have to admit, I'm not too happy with this whole code, but I didn't find a better way to code this logic.

gregorydlogan:
> this harvest API does not guarantee any stability, so we can just improve the logic in the future.

I'm going to merge this since it looks like it works and there's upcoming discussion on how to do captions going forward.  Let's get this in for now, and fix it when it breaks as the captions changes come in.

PR #4317, opened 2022-10-13
Fix NullPointerException when editor does not send subtitles

Arnei:
Does not try to add subtitles to the mediapackage if the post request did not contain any.


lkiesow (approved):
Fixes the issue.

PR #4320, opened 2022-10-14
Add summary of sent data during registration

### LINKED ISSUE #4254: Adopter Registration: Preview Data before submission
**Description of Feature Request / Enhancement**

_Location: Admin UI > "?" > Adopter Registration_

In the registration dialogue it would be great to be able to get precise information about what will be sent. In addition to the more general descritpion within the "privacy policy and terms of use" it would be valuable to see e.g.

a) a JSON "sceleton" file that shows the data fields
or better yet:
b) the exact statistical data that would be sent at the time of registration

This could be offered as downloadable JSON file in the dialogue.

gregorydlogan:
This pull request adds a summary screen of the data sent to our registration server.  ~It pulls the current data, only shortening the services list on the nodes.~
While I was in there, I also fixed a few UI shortcomings, and upgrading the data deletion workflow to properly reflect what is going on.

Fixes #4254

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/index.html
+++ b/modules/admin-ui-frontend/app/index.html
@@ -258,6 +258,7 @@
     <script src="scripts/shared/resources/newThemeResource.js"></script>
     <script src="scripts/shared/resources/taskResource.js"></script>
     <script src="scripts/shared/resources/adopterRegistrationResource.js"></script>
+    <script src="scripts/shared/resources/adopterStatisticSummaryResource.js"></script>
     <script src="scripts/shared/resources/countryResource.js"></script>
     <script src="scripts/shared/resources/termsOfUseResource.js"></script>
     <script src="scripts/shared/services/wizards/new-event/access.js"></script>
--- a/modules/admin-ui-frontend/app/scripts/modules/registration/controllers/registrationController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/registration/controllers/registrationController.js
@@ -24,15 +24,17 @@
 angular.module('adminNg.controllers')
 .controller('RegistrationCtrl', ['$scope', '$timeout', 'Table', 'AdopterRegistrationStates',
   'AdopterRegistrationResource', 'TermsOfUseResource', 'CountryResource', 'NewEventStates', 'NewEventResource',
-  'EVENT_TAB_CHANGE', 'Notifications', 'Modal', 'AuthService',
+  'AdopterStatisticSummaryResource', 'EVENT_TAB_CHANGE', 'Notifications', 'Modal', 'AuthService',
   function ($scope, $timeout, Table, AdopterRegistrationStates, AdopterRegistrationResource, TermsOfUseResource,
-    CountryResource, NewEventStates, NewEventResource, EVENT_TAB_CHANGE, Notifications, Modal, AuthService) {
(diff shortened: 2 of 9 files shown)


lkiesow:
Interesting. Locally, this fails for me with:

```
> eslint                                                                                                                                                                                                                                      
> eslint --config ../../docs/checkstyle/eslintrc.js --rule 'no-unused-vars: off' --rule 'no-empty: off' app/ --resolve-plugins-relative-to .                                                                                                  
                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                              
/home/lars/dev/opencast/modules/admin-ui-frontend/app/scripts/shared/resources/adopterStatisticSummaryResource.js                                                                                                                             
  30:32   error  Missing semicolon             semi                                                                                                                                                                                           
  32:100  error  Strings must use singlequote  quotes                                                                                                                                                                                         
                                                                                                                                                                                                                                              
✖ 2 problems (2 errors, 0 warnings)                                                                                                                                                                                                           
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

Maybe because I built with my local nodejs which is slightly newer.
I will look into it…

gregorydlogan:
There was a syntax error in the branch prior to my force-push, so remerge and it'll work :)

Turns out the proxy-admin thing doesn't run eslint, so I didn't catch the error until after I finished for the night and CI got to it.

lkiesow:
I would suggest putting the dialog showing the data to send after the registration.
Maybe even as an optional dialog, people can get to from the registration with an additional button like this:

Screenshot from 2022-10-15 00-05-41

There are two reasons for this suggestion. First, when the dialog first pops up, it's not very helpful with many of the information people will then fill out actually missing:

Screenshot from 2022-10-14 23-53-46

Second, if we show this afterwards, we can actually react to the user's choices and for example not show the statistical information at all if the user did not select this. Otherwise, people do not choose to send that information, open the dialog again and are told that this information might be sent after all:

Screenshot from 2022-10-15 00-20-44

What do you think?

gregorydlogan:
Sigh, I wondered if you'd catch that... :)

Adidng that button will be a lot more work, but you're right in that it's a better approach. I'll get back at this next week.

lkiesow:
Instead of adding that button, you could also have the review view as a widget step after entering the data, showing only a “Next” button on the form view and then having the “Save” button on the data review step.

gregorydlogan:
Changes:
- UI no longer truncates the service list
- Summary screen is now after the initial data gathering, and reflects user input.  Also shows/hides stats data as appropriate
- Deleting your data is both possible(!) now, and immediate rather than having to wait until the next registration cycle.

lkiesow:
Generally looks good, except that the button labels aren't correct. The form view should have “Next” and the review data view should have “Create”. I first review the data to then register.

Screenshot from 2022-10-19 16-53-00

Screenshot from 2022-10-19 16-54-30

Actually using “Register” as button label in the final step would be even better.

gregorydlogan:
So technically these labels *are* correct - the record is indeed created as soon as 'Create' is clicked in the form view.  It *has* to be since that's what's used to generate the following Summary screen.  I an rejigger the labels though, sure.

gregorydlogan:
@lkiesow The labels have been fixed with #4321.  To fix it here would have caused conflicts in that one.  TBH, review #4321 since it now contains this one anyway :D

PR #4321, opened 2022-10-14
Annoy adopters who haven't registered

gregorydlogan:
This pull request makes a lack of up-to-date registration annoying, by raising an error in the admin UI service monitor bell.  Also added a missing rest doc in the same endpoint class.

Pull request now includes #4320 

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/index.html
+++ b/modules/admin-ui-frontend/app/index.html
@@ -59,12 +59,12 @@
           <span id="error-count" class="badge" ng-show="services.error">{{ services.numErr }}</span>
           <ul class="dropdown-ul">
             <li ng-repeat="(key, value) in services.service"><!-- service directive -->
-              <a ng-if="key == 'Latest Version' && (value.error || value.warning)" href="{{ value.docs_url }}" service-name="{{ key }}">
-                <span>Opencast</span>
+              <a ng-if="key == 'Latest Version' || key == 'Registration' && (value.error || value.warning)" href="{{ value.docs_url }}" service-name="{{ key }}">
+                <span>{{ key }}</span>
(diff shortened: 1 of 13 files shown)


lkiesow:
Can I still decide not to share my information and get rid of the warning?

gregorydlogan:
There isn't a way to record that currently. Guess that should be added.

lkiesow:
The problem I see with this is, that, for example, it may be hard in some cases to get the okay to officially put your organization on the Opencast map. Or there could be several other reasons that you don't want to register.

In these cases, Opencast will now constantly tell you that something is wrong. That's not really true.

In general, the registration form already pops up every time an admin opens the admin interface, unless you tell Opencast you don't want to register. That's why I'm not really sure about this.

If we want more reminders, could it be better for the registration form to pop up again, maybe after each update? E.g. storing that the user hid this when running 11.1, but I'm now running 12.3 → Let's show it again.

gregorydlogan:
Error has been demoted to a warning, so now the bell isn't constantly red.  Same logic as having an out-of-date Opencast install.  Clicking on the warning now opens the modal as well.  Regarding repeated popups, the adopter isn't marked as registered until they actually agree to send things now, so the popup will continue to appear at 30-day intervals until that happens.  I feel like that's close enough to on-new-release that we don't need any additional logic there?

gregorydlogan:
It occurs to me that one thing that's *not* being checked here is whether the communication with the registration server is actually succeeding. I'll add that soon.

gregorydlogan:
Done.  Requested @lkiesow review since he was looking at #4320, which was superseded by this one.

gregorydlogan:
Merging this so that it gets into the next 12.x release.

PR #4332, opened 2022-10-17
CVE-2022-42889 library upgrade

gregorydlogan:
This pull request upgrades commons-text to handle [CVE-2022-42889](https://nvd.nist.gov/vuln/detail/CVE-2022-42889).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <commons-fileupload.version>1.4</commons-fileupload.version>
     <commons-io.version>2.8.0</commons-io.version>
     <commons-lang3.version>3.12.0</commons-lang3.version>
-    <commons-text.version>1.9</commons-text.version>
+    <commons-text.version>1.10.0</commons-text.version>
     <cxf.version>3.4.3</cxf.version>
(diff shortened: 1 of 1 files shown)


lkiesow:
Does that problem affect/endanger Opencast? IF so, we probably need to write a security advisory. Or is this not really affecting Opencast and it's more about getting rid of the warnings?

gregorydlogan:
It's a good question. It's only in animate, but I don't know that module well enough to know how's its used. It would only affect users running specific WOHs which is why I thought this was more of a dep update than a security issue. I'm fine with doing a security notice though.

gregorydlogan:
Something brought up in the meeting today: Is Karaf itself affected by this, and which version(s) have they patched?  I'll track this.

lkiesow:
In animate, we only use `StringEscapeUtils.escapeXml11` to escape strings we then put into XML. That shouldn't be affected. Hence, I don't think Opencast is affected, and we don't need a security advisory. That is, if Karaf isn't affected.

PR #4352, opened 2022-10-20
Automate Release Process

lkiesow:
Right now, cutting a new release is mostly just copy and pasting commands from the developer documentation, hoping that you aren't so bored that you miss changing a version number in those commands.

This patch completely automates the release process. To cut a new release, a release manager can now go to GitHub Actions and trigger the “Cut Release Tag” workflow.

The GitHub Actions user interface will ask for a branch to run this on and the release manager would select the release branch, e.g. `r/13.x`. The workflow will check this and automatically fail on non-release branches.

Based on the release branch, the workflow will then extract the major version and based on existing tags automatically determine the next version in line.

For example, selecting `r/13.x`, with no 13.x versions tagged right now, the workflow will determine that the 13.0 release needs to be cut.

The workflow will then update the `pom.xml` versions, build Opencast and check it starts up properly, commit the changes, create a git tag and finally push the tag to GitHub. In other words, it does all the annoying stuff a release manager would usually have to do manually.

Screenshot from 2022-10-20 17-55-44
*Selecting the branch to run the workflow on*

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/.github/workflows/cut-release-tag.yml
+++ b/.github/workflows/cut-release-tag.yml
@@ -1,0 +1,19 @@
+name: Cut Release Tag
+
+on:
+  workflow_dispatch:
+
+jobs:
+  cut-release-tag:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: prepare git
+        run: |
+          git config --global user.email 'cloud@opencast.org'
+          git config --global user.name 'Release Bot'
+
+      - name: make release
+        run: |
+          ./.github/make-release
--- a/.github/make-release
+++ b/.github/make-release
@@ -1,0 +1,56 @@
+#!/bin/sh
+
+set -eux
+
+# fail if we are not on a release branch
+git rev-parse --abbrev-ref HEAD | grep -e 'r/[0-9]*\.x'
+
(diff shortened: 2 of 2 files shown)


gregorydlogan:
We probably need to update the RM docs now, correct?

.github/workflows/cut-release-tag.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: prepare git
+        run: |
+          git config --global user.email 'cloud@opencast.org'
+          git config --global user.name 'Release Bot'
+
+      - name: make release
+        run: |
+          ./.github/make-release
+
+      - name: just for testing
```
gregorydlogan: Is this still needed?

lkiesow: That can go → removed.

lkiesow:
> We probably need to update the RM docs now, correct?

I deliberately kept them for now since
- This doesn't work on the 11.x releases
- I thought it's a good idea to let me (as OC 12 RM) test this in action before we make it official. I can check if everything actually works and have the necessary rights to fix things if this breaks something.

But long-term, this should go into the docs.

PR #4376, opened 2022-10-25
Allow use of extended metadata in send-email WOH

mtneug:
Currently, the send-email WOH simply searches for catalogs with the flavor `dublincore/*` to populate the flavor. This circumvents the catalog configuration and does not allow for flavors with a different main type.

This now uses the index service that already has logic for filtering for configured catalogs of the current tenant. Only configured catalogs are considered, i.e. this could be a breaking change if adopters use e.g. `dublincore/foo` in the mediapackage but never configure it as a catalog. I think this is a configuration error.

To allow for multiple catalogs with the same subflavor, the templating now uses the full flavor name, e.g. `${catalogs['dublincore/episode']['title']}`. As a special case for `dublincore/*`, it is still posible to only use the subtype. I would declare this as deprecated.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/email-template-service-impl/src/main/java/org/opencastproject/email/template/impl/EmailTemplateServiceImpl.java
+++ b/modules/email-template-service-impl/src/main/java/org/opencastproject/email/template/impl/EmailTemplateServiceImpl.java
@@ -21,11 +21,14 @@
 package org.opencastproject.email.template.impl;
 
 import org.opencastproject.email.template.api.EmailTemplateService;
+import org.opencastproject.index.service.api.IndexService;
 import org.opencastproject.job.api.Incident;
 import org.opencastproject.job.api.IncidentTree;
 import org.opencastproject.mediapackage.Catalog;
 import org.opencastproject.mediapackage.EName;
 import org.opencastproject.mediapackage.MediaPackage;
+import org.opencastproject.mediapackage.MediaPackageElementFlavor;
+import org.opencastproject.metadata.dublincore.CatalogUIAdapter;
 import org.opencastproject.metadata.dublincore.DublinCore;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalog;
 import org.opencastproject.metadata.dublincore.DublinCores;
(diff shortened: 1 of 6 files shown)


mtneug:
As discussed, I changed the scope to `test`.

modules/email-template-service-impl/src/main/java/org/opencastproject/email/template/impl/EmailTemplateServiceImpl.java
```diff
@@ -130,25 +140,43 @@ public String applyTemplate(String templateName, String templateContent, Workflo
...
         String name = ename.getLocalName();
         catalogHash.put(name, dc.getAsText(ename, DublinCore.LANGUAGE_ANY, delimiter));
       }
-      catalogs.put(catalogFlavor, catalogHash);
+
+      catalogs.put(c.getFlavor().toString(), catalogHash);
+      // Backwards compatibility: use only subtype of the flavor as key
+      if (c.getFlavor().getType().equals("dublincore")) {
+        catalogs.put(c.getFlavor().getSubtype(), catalogHash);
+      }
```
geichelberger, line 179: Maybe mention this behavior in the documentation as deprecated but still supported for backward compatibility reasons?

gregorydlogan: IMO not mentioning it is better.  We want to do away with the old style ideally, so not mentioning it would encourage people to upgrade.

gregorydlogan (approved):

modules/email-template-service-impl/src/main/java/org/opencastproject/email/template/impl/EmailTemplateServiceImpl.java
geichelberger: If you want to encourage people to upgrade, it's better to fail fast/early and not have backward compatibility instead of having undocumented behavior. It's still working, but nobody knows why.
You usually document that a method is deprecated, and a few releases later, you remove the stuff.

mtneug: I agree with @geichelberger and added a notice to the docs.

PR #4387, opened 2022-10-30
Preview of img attachments

ShivamGarg2910:
### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
+++ b/modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
@@ -1181,7 +1181,7 @@
                 <td translate="EVENTS.EVENTS.DETAILS.ASSETS.ATTACHMENTS.DETAILS.URL">
                   <!-- Link -->
                 </td>
-                <td><a href="{{ subNavData.url }}" class="fa fa-external-link"></a></td>
+                <td><a href="{{ subNavData.url}}" target="_blank"><img src="{{ subNavData.url }}" alt="Image cannot be loaded"></img></a></td>
               </tr>
             </table>
           </div>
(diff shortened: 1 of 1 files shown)


ShivamGarg2910:
@lkiesow, please review the changes.

lkiesow (changes requested):
Thanks for your contribution, @ShivamGarg2910,

Unfortunately, there are a few problems with this patch.
First, you don't check if the attachment actually is an image. An easy way to do that would be tocheck if the mime type starts with `image/`. This leads to the admin interface now trying to display e.g. CML files as an image as well:

Screenshot from 2022-10-30 13-20-12

Problem number two is that if you have an image and it's a larger one, it will break the dialog and extend over the edge of the dialog. Something like `style="max-width: 100%"` should fix that problem.

Screenshot from 2022-10-30 13-17-29

Number three is that for consistency, it would be great to have the preview in the same place you would find the video preview. This would also fix the problem that the metadata fields says URL, but no URL is shown in the right column any longer. So, my suggestion would be to leave the URL as it is and place the image below the table like the videos (we don't need all the stream metadata):

Screenshot from 2022-10-30 13-17-18

lkiesow:
I've also just noticed that the html validator fails (see CI tests). You can also run that locally:

```
❯ npm run html-validate

> html-validate
> html-validate 'app/**/*.html'


/home/lars/dev/opencast/modules/admin-ui-frontend/app/scripts/shared/partials/modals/event-details.html
  1184:130  error  End tag for <img> must be omitted  void-content

✖ 1 problem (1 error, 0 warnings)
```

mtneug:

mtneug:
Sorry for the close. I misclicked.

lkiesow:
Closing this since this seems to be abandoned.
But feel free to re-open this if you want to continue.

PR #4393, opened 2022-10-31
Prioritization of jobs from older workflows

berthob98:
**This PR modifies the job dispatching to prioritize jobs from older workflows.**

### Technical explanation:
The job dispatcher uses three lists where the queued jobs are stored. 1) Restarted jobs 2) "normal" jobs 3) jobs from the workflow service. The lists are handled in this order.
This PR adds a comparator to order the last two lists by the time, when there related workflow has been started. By changing the order inside this lists the jobs from older workflows get handled earlier.

To be able to sort the jobs by the starting date of the related workflows, it's necessary to be able to match jobs against there related workflow. To achieve that the rootJob field every job object already has was used. Previously this field wasn't really set. To activate this field, the workflow service was modified in three places of job creation, to set this field.
The START_WORKFLOW job of each workflow is used as the rootJob for other jobs.
In cases, where a root job is missing the comparator leaves the order of the list untouched.

### TL;DR
The job queues are FIFO on a workflow level.

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
@@ -97,6 +97,9 @@ public class JobDispatcher {
   /** JPA persistence unit name */
   public static final String PERSISTENCE_UNIT = "org.opencastproject.common";
 
+  /** Identifier for the workflow service */
+  public static final String TYPE_WORKFLOW = "org.opencastproject.workflow";
+
   /** Configuration key for the dispatch interval, in seconds */
   protected static final String OPT_DISPATCHINTERVAL = "dispatch.interval";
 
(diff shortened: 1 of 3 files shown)


modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -306,10 +306,14 @@ public void run() {
             continue;
           }
 
+          //Prioritization of jobs from older workflows
+          dispatchableJobs.sort(new WorkflowPriorizationComparator());
           dispatchDispatchableJobs(em, dispatchableJobs);
         } while (jobsFound);
```
geichelberger, line 312: You will only have the effect of prioritizing older workflows if you have fewer than 100 jobs in the queue. The job dispatcher uses pagination and only grabs the following 100 Jobs sorted by creation date.

You are sorting only the pages and not the complete resultset.

mtneug:

berthob98:
I'm sorry for the late response, I didn't find time during the last weeks to further develope this PR. Thank you for the review. I didn't knew about the pagination in the dispatcher. I will try to test the behaviour for more than 100 parallel jobs and look out for a solution to sort the complete resultset.

gregorydlogan:
Is this going to move forward?  If not, next I'm passing through the PRs I'll close this.

KatrinIhler:
Since this hasn't been getting attention in a while, I'm gonna close this for now. Feel free to re-open once it has been updated. :)

Issue #3538, opened 2022-03-17
Standalone Editor CORS error in opencast 11.4


AL-THD:
**Describe the issue**
In opencast 11.4 the standalone Editor cannot load the waveform due to a CORS error:
XMLHttpRequest blocked by CORS policy as no header is present on the requested ressource.

Browser - console output:
```
Access to XMLHttpRequest at 'https://<presentation-node>/static/mh_default_org/internal/69f554e6-98ad-4e60-ad3b-ff8cdc00a887/7d9e16a2-114a-4bf9-86fc-cc506236659b/composite.mp4' from origin 'https://<admin-node>' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
composite.mp4:1          Failed to load resource: net::ERR_FAILED
composite.mp4:1          Failed to load resource: the server responded with a status of 403 (Forbidden)
composite.mp4:1          Failed to load resource: the server responded with a status of 403 (Forbidden)
/admin-ng/index.html;jsessionid=node0ubq5wme8kduk148dsxtqqovuu2517.node0:1          Failed to load resource: net::ERR_TOO_MANY_REDIRECTS
```

**To Reproduce**
1. open Admin UI
2. use "open video editor" button on any video
3. waveform gets the "wheel of death"
4. open browser console for errors

**Environment Information**
- Cluster: 1 admin vm, 1 worker vm, 1 presentation vm
- OS: CentOS 7 (Red Hat 4.8.5-39)
- MariaDB Ver 10.7.3
- Shibboleth Auth
- Apache httpd
- Browser: Chrome, Firefox, Edge
- Opencast 11.4

**Our Fix:**
Add a line to the apache ssl file `/etc/httpd/conf.d/<presentation-node-url>.ssl.conf`:
```
Header add Access-Controll-Allow-Origin <admin – server – node – adress>
```

Rillke:
> CORS missing allow origin

Is the .4 version cursed? I can see this issue in 12.4 now, too although I am sure, I didn't experience that in 12.3 where we ran the editor without any -UI-configuration. Wondering if something changed in the editor or if the added configuration did that.

![Screenshot from 2022-10-21 17-16-29](https://user-images.githubusercontent.com/2311611/197231364-f050590b-1f92-4fe7-b8cd-0f4d451b987b.png)

The waveform is not an issue because it's fetched as an img src.

Rillke:
[...]

PR #4411, opened 2022-11-02
Let Editor know if files are available locally

lkiesow:
This patch is in preparation to fixing #3538: The editor is deployed on both the admin and the presentation server. But the publication only points to one of these servers by default. This means that one of the servers needs to return special HTTP headers so that the other one can load data or cross-origin requests will fail.

In reality though, most installations will either have the distribution on a network file system, meaning it is available from all servers and we could easily circumvent the issue, or they have dedicated and hopefully properly configured download servers anyway.

In case the latter exist, there shouldn't be an issue in the first place. This patch now indicates to the editor if the former is the case and the editor can load files from the local server.

This should always be true in the recommended installation of Opencast, but this actually lets us react to the cases where it is not true.

The status is indicated in the `/editor/<id>/edit.json`:

```json
{
  "local": true
}
```

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
+++ b/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
@@ -92,6 +92,7 @@ import com.entwinemedia.fn.data.Opt;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
@@ -173,6 +174,8 @@ public class EditorServiceImpl implements EditorService {
   private MediaPackageElementFlavor captionsFlavor;
   private String thumbnailWfProperty;
   private List<MediaPackageElementFlavor> thumbnailSourcePrimary;
+  private String distributionDirectory;
+  private Boolean localPublication = null;
 
   private static final String DEFAULT_PREVIEW_SUBTYPE = "prepared";
   private static final String DEFAULT_PREVIEW_TAG = "editor";
@@ -197,6 +200,7 @@ public class EditorServiceImpl implements EditorService {
   public static final String OPT_THUMBNAILSUBTYPE = "thumbnail.subtype";
   public static final String OPT_THUMBNAIL_WF_PROPERTY = "thumbnail.workflow.property";
   public static final String OPT_THUMBNAIL_PRIORITY_FLAVOR = "thumbnail.priority.flavor";
+  public static final String OPT_LOCAL_PUBLICATION = "publication.local";
 
   private final Set<String> smilCatalogTagSet = new HashSet<>();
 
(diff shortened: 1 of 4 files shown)


modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -962,8 +997,12 @@ public EditingData getEditData(final String mediaPackageId) throws EditorService
 
       final int priority = thumbnailSourcePrimary.indexOf(track.getFlavor());
 
+      if (localPublication == null) {
+        localPublication = isLocal(track.getURI());
```
gregorydlogan, line 1001: This is calculated for each track, and would affect other tracks, right?

Cases:
- Initially null, afterwards it's not
- Initially *not* null, is not recalculated
- Next time someone calls `getEditData` the initial state of this would depend on what happened *last* time, right?

I mean, I guess it shouldn't change that often, but still...

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -972,7 +1011,7 @@ public EditingData getEditData(final String mediaPackageId) throws EditorService
             .collect(Collectors.toList());
 
     return new EditingData(segments, tracks, workflows, mp.getDuration(), mp.getTitle(), event.getRecordingStartDate(),
-            event.getSeriesId(), event.getSeriesName(), workflowActive, waveformList, subtitles);
+            event.getSeriesId(), event.getSeriesName(), workflowActive, waveformList, subtitles, localPublication);
```
gregorydlogan, line 1014: This use of `localPublication` is dependent on the state of the *last* call to `isLocal`.  This is *probably* fine, but what if one file isn't local somehow, and the rest are?

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lkiesow: What is a case where this would change? Wither you have a shared file system in which case you have all the files on all servers, or you don't in which case all distribution files should be on the presentation server.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
lkiesow: Then, you don't have a shared file system but made a hell of mess on your server since you manually copied your distribution files to other servers where they aren't used. For me, that's a bit like asking, “but what if someone manually wrote stuff into the database”? The answer: Things break. However, you can configure this away from `auto` in these cases if you want to.

We could check, but to be sure, we would need to check all the files and then, things might still break if you find and flag one as not-local because of https://github.com/opencast/opencast/issues/3538

gregorydlogan: Yeah, ok fair enough.  I could see this biting us in the future if we somehow do tiered storage for delivery, but that's a problem for future us and can be addressed at that point.

PR #4412, opened 2022-11-02
JWTs in request parameters

JulianKniephoff:
This extends the existing JWT machinery to allow for passing a JWT as a (configurable) URL/query parameter.

This is useful for example if you are an external system wanting to link to Opencast services like Studio or the new editor passing along an authenticated context. You know, [like Tobira](https://github.com/elan-ev/tobira/issues/363) for example. :wink: 

This is completely backwards compatible. You can just ignore/not configure the newly added authentication filter.

### Notes for reviewers

It probably makes sense to look at the individual commits, instead of just the changed files. Also note the commit messages, which might have more details/justifications. Some general notes:

* I'm not super sure about the first commit, but as far as I understand, these declarations are redundant. And in fact, as they were written, the Shibboleth and JWT filter wouldn't work together, because they can't both be mounted at the same `position`.
* The next two commits are just preparatory. Some refactorings that make introducing another filter to the module a bit easier and avoid some duplication. I put those in their own commit to simplify the review.
* The last commit is where the actual sausage is made.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


etc/security/mh_default_org.xml
```diff
@@ -818,8 +829,8 @@
                   interface="org.opencastproject.security.api.SecurityService" />
 
   <!-- Uncomment to enable external users e.g. used together with shibboleth and JWT -->
-  <!-- <osgi:reference id="userReferenceProvider" cardinality="1..1"
-                  interface="org.opencastproject.userdirectory.api.UserReferenceProvider"  /> -->
+  <osgi:reference id="userReferenceProvider" cardinality="1..1"
+                  interface="org.opencastproject.userdirectory.api.UserReferenceProvider"  />
```
LukasKalbertodt: This seems like a leftover from testing?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
```diff
@@ -0,0 +1,128 @@
...
+import org.springframework.util.Assert;
+
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import javax.servlet.http.HttpServletRequest;
+
+// TODO Amend documentation
```
LukasKalbertodt: This is done right?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTRequestHeaderAuthenticationFilter.java
```diff
@@ -27,7 +27,7 @@
 import javax.servlet.http.HttpServletRequest;
 
 /**
- * Authentication filter for JWTs.
+ * Authentication filter for JWTs in request headers
```
LukasKalbertodt: You stole a period. Give it back.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
```diff
@@ -0,0 +1,128 @@
...
+  protected Object getPreAuthenticatedPrincipal(HttpServletRequest request) {
+    if (debug) {
+      Util.debug(logger, request);
+    }
+
+    String queryString = Objects.requireNonNullElse(
+            request.getQueryString(), "");
+
+    Map<String, String> queryParams = Arrays.stream(
+                    queryString.split("&"))
```
LukasKalbertodt: I think both of these don't quite match the Opencast style. Putting each of those in one line is still under 80 chars, so I don't think these should be broken into two lines each. Again, the rest of the codebase doesn't break that early, with 120 being the hard limit.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
```diff
@@ -0,0 +1,128 @@
...
+            .map(kv -> kv.split("="))
+            .collect(Collectors.toMap(
+                    kv -> kv[0],
+                    kv -> kv.length < 2 ? "" : kv[1]));
+
+    String token = queryParams.get(parameterName);
```
LukasKalbertodt: Do we really need to collect all parameters into a map to then retrieve only one? Can't this be done with `.filter().findFirst()` on the stream?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
```diff
@@ -0,0 +1,128 @@
...
+
+    if (token == null && exceptionIfParameterMissing) {
+      throw new PreAuthenticatedCredentialsNotFoundException(this.parameterName + " parameter not found in request.");
+    } else if (token == null || token.isBlank()) {
+      return null;
+    }
```
LukasKalbertodt: Is that intended that `isBlank` is only checked in the second condition?

LukasKalbertodt (changes requested):
I tested this patch and it does indeed work! It is worth noting that if a login session already exists, that takes precedence and is not replaced by the JWT session. 

I also did a brief review and noticed some things. Apart from this minor stuff it looks good to me! However, my expertise in this area is rather limited, so someone else should take a look as well. Maybe @joneug or @mtneug?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
JulianKniephoff: Yes, the exception should be thrown exactly when the parameter is not provided at all, not like `?jwt=` where it would be blank.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTRequestHeaderAuthenticationFilter.java
JulianKniephoff: :ninja:

ziegenberg:
Are you sure having JWT tokens exposed in URL parameters is a good idea? Logged URLs can leave those tokens exposed if those logs were to become compromised. Now also end users have easy access to tokens. This could mean those tokens are shared accidentally by an unknowing user via copy/pasting. It might be okay if they are single-use tokens and there is absolutely no other way to solve the problem.

JulianKniephoff:
@ziegenberg we talked about this internally as well and our conclusion was something like: It's fine if the JWTs are short-lived enough, which is a best practice anyway.

If an attacker has access to server logs, they can probably also read the request headers, where the JWTs are "normally" stored, so in comparison to that, the only real additional vector is sharing links accidentally, or spying a JWT from a users URL/browser history somehow. And by that I mean manually, because (apart from browser bugs of course) you can't do that from within the browser engine. (Extensions could do it, but these can, again, also read headers.)

In Tobira for example (~PR still pending unfortunately, if I remember, I will link it here~ see elan-ev/tobira#587) we generate the token on the fly, for example when the user clicks on a link to Studio, and you need it only once to establish the session, so you can really make this live only a few seconds. 🤷‍♀️ 

Really, the security profile is about the same as for the URL signing mechanism that's always floating around somewhere in the Opencast sphere. Not that that's necessarily good or bad; just so people are aware.

All that being said: I'm open to alternative ideas fro the use case posed in this PR that doesn't cross any of the ever stricter boundaries browsers draw. I thought some of them through in the issue I linked in the description. Feel free to add to that discussion as well.

ziegenberg:
Can I securely pass a JWT in the URL query parameters of a GET request?
Answer: https://community.auth0.com/t/can-i-securely-pass-a-jwt-in-the-url-query-parameters-of-a-get-request/65678

> Really, the security profile is about the same as for the URL signing mechanism that's always floating around somewhere in the Opencast sphere. Not that that's necessarily good or bad; just so people are aware.

The argument that because it's not better elsewhere, it doesn't have to be done well here is not optimal. Also, they are not really comparable.
The URL signing mechanism only does authorize you to watch a specific resource. It operates in a read-only context.
This proposed JWTs in URL parameters is meant to handle an authenticated context, which logs you into another service. And it comes with the capabilities to not only read but also do write and change information. It has to have a higher security profile.

I suggest using the _Post/Redirect/Get (PRG)_ Pattern https://en.wikipedia.org/wiki/Post/Redirect/Get.

JulianKniephoff:
> Can I securely pass a JWT in the URL query parameters of a GET request? Answer: https://community.auth0.com/t/can-i-securely-pass-a-jwt-in-the-url-query-parameters-of-a-get-request/65678

The length argument is interesting, I hadn't considered that. I feel like we are far away from what the admin UI does for example, passing what feels like the entire application state in the URL. Which is not to say that we couldn't do it better, you are right. I'm simply saying: I don't think it's a problem for us, because we know our clients' browsers can deal with way more, and the only thing that could reasonably "blow up" our tokens are ridiculously long role lists. :shrug:

The other points were covered in my previous answer, I think, but for the record again in brief: The tokens live for a few seconds, so we feel the risk is tolerable.

> The argument that because it's not better elsewhere, it doesn't have to be done well here is not optimal. Also, they are not really comparable. The URL signing mechanism only does authorize you to watch a specific resource. It operates in a read-only context. This proposed JWTs in URL parameters is meant to handle an authenticated context, which logs you into another service. And it comes with the capabilities to not only read but also do write and change information. It has to have a higher security profile.

You are right, the comparison is slightly misleading, I also realized that shortly after writing it. URL signing also stops you from tampering with the URL for example. However, you **can** use it in a similar fashion, granting privileges beyond just viewing resources. SWITCH for example used this mechanism to allow people to annotate videos in the Annotation Tool for a long time. :shrug: 

It's also fair to say that "it's done poorly elsewhere, so it's fine to do it here" is a bad argument. It is. For the record, though, that's not what I'm saying. I'm merely pointing out a precedent which could raise the question of whether it's actually such a poor decision after all, if it's "common practice" in this community. :shrug: 

> I suggest using the _Post/Redirect/Get (PRG)_ Pattern https://en.wikipedia.org/wiki/Post/Redirect/Get.

Anyway, all that being said, as I indicated in the technical already, I'm willing to look into alternatives. PRG is something I considered already, but this was simpler for now. Thanks for the link anyway. :D Before I look into this further, though, I would like to know whether that's an acceptable solution for @mtneug as well, since he indicated that he wants something like this for other use cases as well.

ziegenberg:
"common practice" should not be used as a justification for the continued use of "not best practice." We generally should strive for "better practice", especially regarding security, so we raise the bar in the community.

JulianKniephoff:
Common practice indicates *something* though, and that's worth discussing. I'm not using it as a justification, I'm merely pointing it out, as I said. The result of that discussion can totally be that it's justifiable, though. Again: It's common practice for *some* reason. Whether or not it's a good one in this case I personally withhold judgement over for now. :shrug:

Most of these things can be said aobut "best practices" as well, by the way. Blindly following them is not a good idea either. They don't apply in all cases, and might even be downright harmful in yet others. Also, they might not even be the best in *most* cases. Remember the time where it was a best practice to force your employees to change their passwords every week? Now you could say that was never a best practice, but I ask you who determines what is best and what is merely common. :shrug: 

Anyway, again, I don't even disagree with you in principle! :D I'm just saying: I think this is fine. Some people agree. I'm totally willing to investigate/build alternatives if the community decides it's not. But honestly I would like to see some more voices in this thread before I do, I think. :thinking: Alternatively we can bring it up in next week's technical again.

mtneug:
How JWTs can be stored/passed is indeed interesting. The actual implementation should keep in mind "how much security is needed". I previously read through Auth0 docs on storing JWT for browser use (https://auth0.com/docs/secure/security-guidance/data-security/token-storage). Here are some random thoughts coming from that and other places:

* Storing JWTs in-memory of the JavaScript context gives you a strong security option but has significant drawbacks. Having the user authenticate on every page refresh makes for a very poor user experience for most pages. This also does not help in passing along JWT between systems. This could be "better practice", but only for certain apps needing these security guarantees.

* Storing JWTs in local storage is the second option Auth0 discussed. This could be worse as this storage location can be accessed from any JavaScript running in the same context. So if the app is vulnerable to XSS, the attacker could read out the JWT. Using local storage between systems is also subject to the same-origin policy, e.g. even subdomains don't share this store. So not a good solution for passing JWTs to any other system. Although, in general, I think this is a valid approach for auth within one system and something I want to explore as an alternative to cookies.

* Cookies are not really discussed in the link above but are also an option for storing JWTs. The advantages are that you can hinder JavaScript from accessing these, the browser will send them automatically, and you can share cookies between subdomains.  However, you still have the problem of sharing cookies between different sites. Also, cookies have a very bad reputation leading to multiple legal and technical policy decisions. Browsers more and more restrict the use of third-party cookies. This is mainly the reason we do not use LTI with Opencast for educast.nrw. Safari will simply not send the cookie if Opencast is embedded in an LMS that is a different site. For now, there are probably ways of allowing that with the `SameSite` cookie flag, but if you ask me, I see a clear trend to outright ban third-party cookies.

* As a side note: we don't store the JWT coming out of the OIDC auth flow directly in the educast.nrw SSO cookie. Instead, a random string is stored in the cookie as a key into a Redis store that actually stores the JWT. I have also seen the suggestion to have an external JWT that the front-end sees and an internal JWT with more claims fetched by a central component and passed along to the back-end services.

* Signing/securing URLs with JWTs is not a new idea and something we want to do for video delivery as well. Cloudflare allows adopters to secure access to HLS manifests and media using JWTs that are encoded as part of the URL path (see https://developers.cloudflare.com/stream/viewing-videos/securing-your-stream/). [RFC9246](https://datatracker.ietf.org/doc/html/rfc9246) (part of the CDNi family, but I also think valid for simple media delivery scenarios) describes a mechanism of passing JWTs as URL parameters between users and CDNs. The Streaming Video Technology Alliance proposes a different signing format in the URL, but uses JWTs in HTTP headers (https://www.svta.org/project/open-caching-relayed-token-authentication/). These JWTs do not contain typical user data and are used to identify the video + further authorization constraints. If the application does not rotate the key, these JWT could also have a longer `exp` duration on the order of hours instead of seconds/minutes. Using POST requests is not possible here as the streaming formats don't allow that. You could argue that users typically don't see those URLs directly, but the issue with web logging still remains. If you ask me, I very much like this approach, as JWTs allow edge nodes to be fully autonomous when delivering files. At the very least, these examples show that passing JWTs in URLs is becoming a common approach for media delivery (might I say "best practice" 😉).

* As @JulianKniephoff already mentioned, signed URLs were also used to authenticate users to tools like the Opencast Annotation Tool. We would like to do the same mainly due to the mentioned problems with LTI. The other reason is that in our role model, students do not have roles corresponding to videos/series ACLs, i.e. Opencast would not give read/write access to students. Instead, the decision if a user should be able to access a video is delegated to external systems like the LMS (this model was used by SWITCH as well). This, in turn, means that the authorization decision for giving access to the OAT also should come from the external system. We wanted to model this intent using JWT.

etc/security/mh_default_org.xml
```diff
@@ -368,18 +368,10 @@
...
-    -->
-
     <!-- CAS Auth:  Uncomment this if using CAS authentication
     <sec:custom-filter position="FORM_LOGIN_FILTER" ref="casFilter" />
     -->
 
-    <!-- JWT header authentication filter
-    <sec:custom-filter ref="jwtHeaderFilter" position="PRE_AUTH_FILTER"/>
-    -->
-
```
geichelberger, line 382: The `shibbolethHeaderFilter` and the `jwtHeaderFilter` are both `RequestHeaderAuthenticationFilter` which extends `AbstractPreAuthenticatedProcessingFilter`. The correct filter position is `PRE_AUTH_FILTER`.

etc/security/mh_default_org.xml
```diff
@@ -450,6 +442,8 @@
 
         <!-- Uncomment the line below to support JWT.
         <ref bean="jwtHeaderFilter" /> -->
+        <!-- Additionally/alternatively uncomment this to support passing a JWT in a URL parameter.
+        <ref bean="jwtQueryParameterFilter" /> -->
```
geichelberger: They should be part of a CompositeFilter for the PRE_AUTH_FILTER position.

Section 3.3.5 https://docs.spring.io/spring-security/site/docs/3.1.x/reference/springsecurity-single.html
> **Table 3.1 Standard Filter Aliasing and Ordering**
PRE_AUTH_FILTER | AstractPreAuthenticatedProcessingFilter Subclasses
BASIC_AUTH_FILTER | BasicAuthenticationFilter

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
```diff
@@ -0,0 +1,125 @@
...
+            .filter(kv -> kv[0].equals(parameterName))
+            .findAny()
+            .map(kv -> kv.length < 2 ? "" : kv[1])
+            .orElseGet(() -> {
+              if (exceptionIfParameterMissing) {
+                throw new PreAuthenticatedCredentialsNotFoundException(
+                        parameterName + " parameter not found in request.");
+              }
+              return "";
+            });
```
geichelberger: The HttpServletRequest has a `getParameter` method if the parameter has only one value and a `getParameterValues` method if it's a multivalued parameter.

geichelberger (changes requested):

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTQueryParameterAuthenticationFilter.java
JulianKniephoff: How did I miss this?! Anyway, using it now, significantly shortening the implementation. Thanks! :D

etc/security/mh_default_org.xml
JulianKniephoff: I didn't even consider having another composite filter for that position. Thanks for the idea! :D

I pushed another commit restructuring this a bit further. I left the OAuth/LTI filter where it is now because I wasn't sure how to classify it. You think this is more correct now? :thinking:

etc/security/mh_default_org.xml
JulianKniephoff: See your other, related comment for a reply to this. :)

gregorydlogan (approved):
Per discussion in the technical meeting, there is no particular objection to merging this in, despite the number of comments and remaining (potential) issues.  Since this is inactive by default it poses no risk to existing deployments, and future adopters should be aware of the risks should they enable it.

I'm going to mark this as approved so that this is clear, but I have not actually reviewed this code.  If I get around to this I'll remove this part of the comment, or reapprove clarifying the situation.

tl;dr: We took a voice vote and it amounted to a +0 for proceeding with this, with a minor preference for Post/Redirect/Get to be implemented later.

JulianKniephoff:
OK so, in the interest of moving this forward I made a few more changes:

* Due to @geichelberger's suggestion [here](https://github.com/opencast/opencast/pull/4412#discussion_r1020106181), the filter as it was implemented can actually already read the JWT from a form parameter of a `POST`-request as well, already implementing half of P/R/G.
* The other half is given by the new `redirect` module which contains endpoints to redirect users to a specified URL. Well, at the moment this is only one, implementing exactly the P/R/G pattern, but I could imagine people extending this with further schemes in the future if they are needed. :shrug: 

For an example of how to use this combination you can look at elan-ev/tobira#587. (~NOT YET, THOUGH! I will update it tomorrow, probably.~ Took a bit longer but it's done now.) So yeah, Tobira will now also use P/R/G.

All that being said, I am not going to disable/prevent people from using this to pass JWTs in query parameters. As @mtneug pointed out there are actually valid use cases for it.

We hope that takes a bit of the controversy out of this PR so that it can land soon. :wink:

modules/redirect/src/main/java/org/opencastproject/redirect/RedirectEndpoint.java
```diff
@@ -0,0 +1,102 @@
...
+    if (target == null) {
+      throw new WebApplicationException("missing `target` URL", Status.BAD_REQUEST);
+    }
+
+    URI targetUri;
+    try {
+      targetUri = URI.create(target);
+    } catch (IllegalArgumentException e) {
+      throw new WebApplicationException("invalid `target` URL", e, Status.BAD_REQUEST);
+    }
```
geichelberger: We need a configurable allowlist, or it's possible to redirect the user to an arbitrary URL.

etc/security/mh_default_org.xml
```diff
@@ -712,6 +723,15 @@
     <property name="debug" value="false" />
   </bean>-->
 
+  <!-- General JWT request parameter extraction filter
+  <bean id="jwtRequestParameterFilter" class="org.opencastproject.security.jwt.JWTRequestParameterAuthenticationFilter">
+    <property name="parameterName" value="jwt" />
+    <property name="authenticationManager" ref="authenticationManager" />
+    <property name="loginHandler" ref="jwtLoginHandler" />
+    <property name="exceptionIfParameterMissing" value="false" />
+    <property name="debug" value="false" />
```
geichelberger, line 729: What do you think of using `logger.isDebugEnabled()` instead of the property? Because the Util.debug will only print output if debugging is enabled, you can kill two birds with one stone (The german version of that saying is nicer,  🪰🪰🎬).

modules/redirect/src/main/java/org/opencastproject/redirect/RedirectEndpoint.java
JulianKniephoff: Also added now.

etc/security/mh_default_org.xml
JulianKniephoff: :shrug: :+1:

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -76,8 +75,16 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
   <property name="loginHandler" ref="jwtLoginHandler" />
   <!-- Throws an exception if a request is missing the configured header (default: true) -->
   <property name="exceptionIfHeaderMissing" value="false" />
-  <!-- Logs the header contents for debug purposes (default: false) -->
-  <property name="debug" value="false" />
```
geichelberger: I just realized that the old filter had the same property, and as this PR goes into 12, it should not contain breaking changes. If you want that both filters behave similarly, revert the changes, or else do not remove the property from the existing filter.

geichelberger (changes requested):
LGTM

docs/guides/admin/docs/configuration/security.jwt.md
JulianKniephoff: Good point. I will extract the change into a new PR. That has to target `develop` then, right? 🤔

geichelberger: :+1:

JulianKniephoff: Done: https://github.com/opencast/opencast/pull/4511

JulianKniephoff:
We had another internal talk about this and felt like it was better to not have a completely open redirection facility, so I changed the `POST /redirect/get` endpoint to only allow redirects to paths (+ queries + fragments) on the same scheme + authority. With that, we shouldn't need the allow list anymore and also feel safe to open the endpoint for everyone. This makes the whole thing way easier to configure (it just works by default) and harder to mis-configure (you can't, basically).

geichelberger (approved):
LGTM

PR #4416, opened 2022-11-06
Allow dots in template file names

mtneug:
This adds dots to allowed characters in template files names (e.g. email templates). This also replaces the custom `isBlank` implementation.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/RestDocData.java
+++ b/modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/RestDocData.java
@@ -21,6 +21,8 @@
 
 package org.opencastproject.runtimeinfo.rest;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
 import org.opencastproject.util.doc.DocData;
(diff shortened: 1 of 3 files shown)


modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/RestDocData.java
```diff
@@ -21,6 +21,8 @@
 
 package org.opencastproject.runtimeinfo.rest;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
```
gregorydlogan, line 24: Not to be picky, but this isn't needed right?  I'm surprised that checkstyles isn't complaining here.

gregorydlogan (approved):
Approve, pending resolution on comment

modules/runtime-info/src/main/java/org/opencastproject/runtimeinfo/rest/RestDocData.java
mtneug: It is as `DocData` no longer has a `isBlank` implementation. This class uses it.

PR #4417, opened 2022-11-06
Improve event catalog handling in External API

mtneug:
This is based on #4181

This adds four improvements to event catalog handling in the External API

1. Avoid DB query is not necessary.
2. Cache map of organization to list of catalogs instead of always calculating it.
3. Avoid calling list providers to fetch collection data if this is later on not used (e.g. all series).
4. Remove collection data from responses for extended metadata.

The last point can be seen as a change to the External API. However, one can also see this as a bug, as other methods in the class hide the collection data from extended metadata.

All changes can add up when requesting multiple events with `/events` especially with many events and series.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -63,6 +63,7 @@ import org.opencastproject.index.service.util.RequestUtils;
 import org.opencastproject.index.service.util.RestUtils;
 import org.opencastproject.ingest.api.IngestException;
 import org.opencastproject.ingest.api.IngestService;
+import org.opencastproject.list.impl.EmptyResourceListQuery;
 import org.opencastproject.mediapackage.Attachment;
 import org.opencastproject.mediapackage.AudioStream;
 import org.opencastproject.mediapackage.Catalog;
@@ -1370,8 +1371,8 @@ public class EventsEndpoint implements ManagedService {
         if (fields != null) metadataList.add(catalogUIAdapter, fields);
       }
     }
-    // TODO: This is slow:
-    DublinCoreMetadataCollection collection = EventUtils.getEventMetadata(event, eventCatalogUIAdapter);
+    DublinCoreMetadataCollection collection = EventUtils.getEventMetadata(event, eventCatalogUIAdapter,
+        EmptyResourceListQuery.get());
     ExternalMetadataUtils.changeSubjectToSubjects(collection);
     ExternalMetadataUtils.removeCollectionList(collection);
     metadataList.add(eventCatalogUIAdapter, collection);
@@ -1400,7 +1401,8 @@ public class EventsEndpoint implements ManagedService {
       // Try the main catalog first as we load it from the index.
       EventCatalogUIAdapter eventCatalogUIAdapter = indexService.getCommonEventCatalogUIAdapter();
       if (flavor.get().equals(eventCatalogUIAdapter.getFlavor())) {
-        DublinCoreMetadataCollection collection = EventUtils.getEventMetadata(event, eventCatalogUIAdapter);
+        DublinCoreMetadataCollection collection = EventUtils.getEventMetadata(event, eventCatalogUIAdapter,
+            EmptyResourceListQuery.get());
         ExternalMetadataUtils.changeSubjectToSubjects(collection);
         ExternalMetadataUtils.removeCollectionList(collection);
         convertStartDateTimeToApiV1(collection);
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -159,6 +159,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletRequest;
@@ -242,6 +243,7 @@ public class EventsEndpoint implements ManagedService {
   private IngestService ingestService;
   private SecurityService securityService;
   private final List<EventCatalogUIAdapter> catalogUIAdapters = new ArrayList<>();
+  private final HashMap<String, List<EventCatalogUIAdapter>> orgCatalogUIAdaptersMap = new HashMap<>();
   private UrlSigningService urlSigningService;
   private SchedulerService schedulerService;
   private CaptureAgentStateService agentStateService;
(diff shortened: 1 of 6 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -336,19 +347,16 @@ public void setWorkflowService(WorkflowService workflowService) {
...
-        adapters.add(adapter);
-      }
-    }
-    return adapters;
+  private List<EventCatalogUIAdapter> getEventCatalogUIAdapters(String organization) {
+    List<EventCatalogUIAdapter> cachedCatalogUIAdapters = orgCatalogUIAdaptersMap.computeIfAbsent(organization,
+        org -> new ArrayList<>(catalogUIAdapters.stream()
+            .filter(a -> a.handlesOrganization(org))
+            .collect(Collectors.toList())));
+    return new CopyOnWriteArrayList<>(cachedCatalogUIAdapters);
```
geichelberger, line 359: Why return a CopyOnWriteArrayList? Is the Instance used with multithreading?

modules/list-providers-service/src/main/java/org/opencastproject/list/impl/EmptyResourceListQuery.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public final class EmptyResourceListQuery implements ResourceListQuery {
+  private static final EmptyResourceListQuery instance = new EmptyResourceListQuery();
+
+  private EmptyResourceListQuery() {
+  }
+
+  public static ResourceListQuery get() {
+    return instance;
+  }
```
geichelberger, line 42: Why are you implementing a Singleton?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -312,11 +307,27 @@ public void setCommonEventCatalogUIAdapter(CommonEventCatalogUIAdapter eventCata
   )
   public void addCatalogUIAdapter(EventCatalogUIAdapter catalogUIAdapter) {
     catalogUIAdapters.add(catalogUIAdapter);
+    invalidateOrgCatalogUIAdaptersMapFor(catalogUIAdapter);
   }
```
geichelberger, line 311: Instead of just invalidating the cache, you could add the adapter. What's the reason for doing it lazily?

geichelberger (changes requested):
Couldn't the caching logic and managing the adapters be extracted to the IndexService? It already contains the adapters and has functions that return a list equivalent to that from the EventEndpoints getEventCatalogUIAdapters.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
mtneug: The list might be changed (some callers remove stuff). This should not affect other clients.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
mtneug: I find this mentally easier to reason about such a system: only one place where the cache is filled and one place where we invalidate the cache. In practice, this should also nearly never happen. Usually, the number of adapters is fixed, and these methods are called only during boot.

modules/list-providers-service/src/main/java/org/opencastproject/list/impl/EmptyResourceListQuery.java
mtneug: Do we need multiple instances of this class?

mtneug:
> Couldn't the caching logic and managing the adapters be extracted to the IndexService? It already contains the adapters and has functions that return a list equivalent to that from the EventEndpoints getEventCatalogUIAdapters.

I agree that it would be good to extract that logic into a common service, but I would not use the IndexService. Not sure if this is in the scope of this PR. Do you want me to do that here?

modules/list-providers-service/src/main/java/org/opencastproject/list/impl/EmptyResourceListQuery.java
geichelberger: Maybe not but it does not have to ensure that only one instance exists.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
geichelberger: But are those callers from different threads? Maybe I have overseen something, but each thread has its list and does not need a thread-safe version.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
geichelberger: Do you have to invalidate the cache on adding an adapter, wouldn't just adding the adapter be sufficient? Then you would still have only one place where the cache is filled and one place where it's invalidated.

geichelberger:
I was just wondering because you implemented caching, but you are still using the function `getCommonEventCatalogUIAdapter()` of the index service, which does not benefit from it. Moving the logic into the IndexService would deduplicate code, and every consumer of the index service would benefit from that. Extracting the handling in a common service would be a step further and out of the scope of that PR.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -242,8 +243,8 @@ private enum CommentResolution {
   private IndexService indexService;
   private IngestService ingestService;
   private SecurityService securityService;
-  private CommonEventCatalogUIAdapter eventCatalogUIAdapter;
   private final List<EventCatalogUIAdapter> catalogUIAdapters = new ArrayList<>();
+  private final HashMap<String, List<EventCatalogUIAdapter>> orgCatalogUIAdaptersMap = new HashMap<>();
```
geichelberger, line 247:
```suggestion
  private final List<EventCatalogUIAdapter> catalogUIAdapters = new CopyOnWriteArrayList<>();
  private final HashMap<String, List<EventCatalogUIAdapter>> orgCatalogUIAdaptersMap = new ConcurrentHashMap<>();
```
Many implementations are using a thread-safe list in combination with multiple, dynamic references https://enroute.osgi.org/FAQ/300-declarative-services.html. I once read an article about why and when you should do that but I cannot find it anymore.

gregorydlogan:
Any news here @mtneug ?

mtneug:
13 is EOL. Closing this for now until I find time to adopt this for newer versions.

PR #4423, opened 2022-11-07
Update Issue Template

lkiesow:
This patch updates the issue temple and adds an overview page for different types of issues.

Screenshot from 2022-11-07 16-38-56

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):

lkiesow:
As discussed in Matrix, I added a link to the mailing list in the following form:

- **Community Support**
Join the mailing list to get help running and configuring Opencast.
[Open](https://groups.google.com/a/opencast.org/forum/#!forum/users)

@mliradelc, @gregorydlogan, if you are happy, please merge ;-)

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,14 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Opencast Studio
+    url: https://github.com/elan-ev/opencast-studio/issues
+    about: Report issues with Opencast Studio in its own repository.
+  - name: Opencast Editor
```
mliradelc, line 6: We have still the old editor as default, should we point that (Until the new editor be the new default for the current and legacy releases).

```suggestion
  - name: Opencast Editor (New version)
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,14 @@
...
+    url: https://github.com/opencast/opencast-editor/issues
+    about: Report issues with Opencast Editor in its own repository.
+  - name: Security issue report
+    url: https://github.com/opencast/opencast/blob/develop/SECURITY.md
+    about: Please report security vulnerabilities here.
+  - name: Community Support
```
mliradelc, line 12: Maybe is better to have this part as the first on the list. Lot of the problems that face new users are normally configuration issues than bugs. What do you think?

mliradelc (approved):
I approve the PR, the things I made a comment are only suggestions.

.github/ISSUE_TEMPLATE/config.yml
lkiesow: I thought about that. The problem with “new version” is that users may not know which one is new and which one is not. You know that because you have seen the development, but imagine you are new to Opencast. You would have no idea. I also thought about adding as comment that it's the one not in the admin interface. But since it's configurable, even that is not universally true. That's why I ended up using what is the official name of the subproject, hoping that we get rid of the old one soon.

That said, I'm open for better suggestions because I definitely see the issue.

.github/ISSUE_TEMPLATE/config.yml
lkiesow: We cannot push it to the top. The external links are always below the issue templates. That would mean we would end up with something like:

- report bug
- report bug
- mailing list
- report bug
- report bug

That's a bit weird.

But I also don't see this as a big problem because right now, we don't really see the issue tracker being used for configuration questions. Just adding the templates shouldn't change that. So, most people coming do want to report a bug and for them the other options make more sense and should therefor be at the top.

.github/ISSUE_TEMPLATE/config.yml
mliradelc: You're right! The new user don't know what editor is.
Have an idea, can you make another step and show the UI of each? Depending on what UI they select you can redirect to the correct page.

Something like this:

```
Select how your editor looks like:

[Screenshot old editor] [Screenshot new editor]
```

.github/ISSUE_TEMPLATE/config.yml
mliradelc: In that case, yes, is it better to left it where it is.

.github/ISSUE_TEMPLATE/config.yml
mliradelc: Simpler approach. 
```suggestion
  about: Report issues with Opencast Editor in its own repository [If the UI looks like this](https://editor.opencast.org/)
```

If its troublesome is better left like originally you wrote, at the end we are going into that direction and if someone publishes an issue about the old editor, one can guide the user to the correct repo.

gregorydlogan: Would it be reasonable to list the version(s) the new one appeared in?  I would expect someone reporting a bug to know which Opencast version is in use...

PR #4434, opened 2022-11-09
Add Studip User Provider

Arnei:
Adds a user provider for the LMS StudIP, similar to other user providers e.g. Moodle.

Sponsored and tested by the University of Osnabrück.


modules/userdirectory-studip/src/main/java/org/opencastproject/userdirectory/studip/StudipUserProviderFactory.java
```diff
@@ -0,0 +1,226 @@
...
+    try {
+      if (properties.get(CACHE_EXPIRATION) != null) {
+        Integer configuredCacheExpiration = Integer.parseInt(properties.get(CACHE_EXPIRATION).toString());
+        if (configuredCacheExpiration != null) {
+          cacheExpiration = configuredCacheExpiration.intValue();
+        }
+      }
+    } catch (Exception e) {
+      logger.warn("{} could not be loaded, default value is used: {}", CACHE_EXPIRATION, cacheExpiration);
+    }
```
lkiesow: Don't fall back if things are misconfigured. Please fail.

etc/org.opencastproject.userdirectory.studip-default.cfg.template
```diff
@@ -0,0 +1,17 @@
...
+
+# The URL and token for the Studip REST webservice
+org.opencastproject.userdirectory.studip.url=https://my-studip.de/studip5/plugins.php/opencast/api/
+org.opencastproject.userdirectory.studip.token=mytoken1234abcdef
+
+# The maximum number of users to cache
+#org.opencastproject.userdirectory.studip.cache.size=1000
+
+# The maximum number of minutes to cache a user
+#org.opencastproject.userdirectory.studip.cache.expiration=60
```
lkiesow:
```suggestion
# The maximum number of users to cache
# Default: 1000
#org.opencastproject.userdirectory.studip.cache.size=1000

# The maximum number of minutes to cache a user
# Default: 60
#org.opencastproject.userdirectory.studip.cache.expiration=60
```

etc/org.opencastproject.userdirectory.studip-default.cfg.template
```diff
@@ -0,0 +1,19 @@
...
+
+# This is an an optional service which is not enabled by default. To enable it,
+# edit etc/org.apache.karaf.features.cfg and add opencast-studip to the featuresBoot option.
+
+# The organization for this provider
+org.opencastproject.userdirectory.studip.org=mh_default_org
+
+# The URL and token for the Studip REST webservice
+org.opencastproject.userdirectory.studip.url=https://my-studip.de/studip5/plugins.php/opencast/api/
+org.opencastproject.userdirectory.studip.token=mytoken1234abcdef
```
gregorydlogan, line 11: These should be unset I'd think?  Take a look at the LDAP config file, equivalent keys are present, but unset.

gregorydlogan (changes requested):
LGTM aside from the config file.  Obviously can't test this since I don't have an StudIP to test against, but I'm assuming this is in use somewhere so it can't be *too* broken :D

etc/org.opencastproject.userdirectory.studip-default.cfg.template
Arnei: I was going off of the Moodle and Sakai config files, which also have those set. 

I think it would be a good idea to at least keep the url in some shape or form, so that the admin doesn't need to look up which endpoint this should target. If you want, I'll unset the url and token and make mention of the endpoint in a comment.

gregorydlogan: Fair enough.  It's not going to cause significant confusion IMO since it *has* to be reset for it to work at all :)

PR #4383, opened 2022-10-27
Simplify ldap user directory implementation

geichelberger:
The previous user directory implementation was reimplementing parts of the OpencastLdapAuthoritiesPopulator.
This PR removes the same logic from the user directory and uses the OpencastLdapAuthoritiesPopulator instead.

We only use LDAP for authorization, so I could only test it with Shibboleth for authentication. The behavior did not change for us.

This PR supersede #4377

**Follow up:**

Reusing the implementation throws an IllegalStateException caused by a recursive load under certain circumstances. This is caused by enriching the LDAP roles with opencast-managed roles.
`AuthoritiesPopulator.getGrantedAuthorities->SecurityService.getUser->UserAndRoleDirectory.loadUser->AuthortitiesPopulator.getGrantedAuthorities`

```
Caused by: java.lang.IllegalStateException: Recursive load of: (mh_default_org,test)
        at com.google.common.base.Preconditions.checkState(Preconditions.java:591) ~[!/:?]
        at com.google.common.cache.LocalCache$Segment.waitForLoadingValue(LocalCache.java:2172) ~[!/:?]
        at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2162) ~[!/:?]
        at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2045) ~[!/:?]
        at com.google.common.cache.LocalCache.get(LocalCache.java:3962) ~[!/:?]
        at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3985) ~[!/:?]
        at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:4946) ~[!/:?]
        at com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache.java:4952) ~[!/:?]
        at org.opencastproject.userdirectory.UserAndRoleDirectoryServiceImpl.loadUser(UserAndRoleDirectoryServiceImpl.java:265) ~[?:?]
        at org.opencastproject.kernel.security.SecurityServiceSpringImpl.getUser(SecurityServiceSpringImpl.java:127) ~[?:?]
        at org.opencastproject.userdirectory.ldap.OpencastLdapAuthoritiesPopulator.getGrantedAuthorities(OpencastLdapAuthoritiesPopulator.java:274) ~[?:?]
        at org.springframework.security.ldap.userdetails.LdapUserDetailsService.loadUserByUsername(LdapUserDetailsService.java:40) ~[?:?]
[...]

PR #4440, opened 2022-11-10
Enrich LDAP users with name and mail address

geichelberger:
For some metadata fields (e.g., presenters), it's possible to enter usernames, and they will be mapped to their corresponding names. If LDAP is enabled, Opencast will dismiss the looked-up users because the name attribute is not set.

This PR enables Opencast to retrieve user attributes like name and mail from LDAP.

The attributes are only mapped if the corresponding config keys are set, else it will fall back to the default behavior.

Example:
```ini
org.opencastproject.userdirectory.ldap.userattributes.name=givenName,sn
org.opencastproject.userdirectory.ldap.userattributes.mail=mail
```

Based on: [#4383](https://github.com/opencast/opencast/pull/4383)

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.userdirectory.ldap.cfg.template
+++ b/etc/org.opencastproject.userdirectory.ldap.cfg.template
@@ -50,6 +50,19 @@ org.opencastproject.userdirectory.ldap.roleattributes=
 #org.opencastproject.userdirectory.ldap.userDn=
 #org.opencastproject.userdirectory.ldap.password=
 
+## The attributes used to construct the user details.
+## If left empty, the user details wont be mapped.
+## The name config key is a comma-separated list of attributes that will be joined
+##   with a whitespace character in the order they appear.
+## Example:
+##  org.opencastproject.userdirectory.ldap.userattributes.name=givenName,sn
+##  org.opencastproject.userdirectory.ldap.userattributes.mail=mail
+## Result:
+##   name: John Doe
+##   mail: john.doe@example.org
+#org.opencastproject.userdirectory.ldap.userattributes.name=
+#org.opencastproject.userdirectory.ldap.userattributes.mail=
+
 ## The maximum number of users to cache
 ## Default: 1000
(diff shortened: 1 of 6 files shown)


mliradelc:
Hi, should be tested with #4383 ? Or are totally independent?

geichelberger:
You can, but you don't have to. It's a different part of the LDAP implementation.

etc/org.opencastproject.userdirectory.ldap.cfg.template
```diff
@@ -50,6 +50,19 @@ org.opencastproject.userdirectory.ldap.roleattributes=
...
+## The name config key is a comma-separated list of attributes that will be joined
+##   with a whitespace character in the order they appear.
+## Example:
+##  org.opencastproject.userdirectory.ldap.userattributes.name=givenName,sn
+##  org.opencastproject.userdirectory.ldap.userattributes.mail=mail
+## Result:
+##   name: John Doe
+##   mail: john.doe@example.org
+#org.opencastproject.userdirectory.ldap.userattributes.name=
+#org.opencastproject.userdirectory.ldap.userattributes.mail=
```
mliradelc, line 64: Tried and works.


One question, can you add custom ldap fields? Because we have another attributes to identify if is a student or a employee. And also if belongs to the uni.

Anyways, works well.

mliradelc (approved):
Works well this PR

etc/org.opencastproject.userdirectory.ldap.cfg.template
geichelberger: What do you want to achieve with the LDAP fields?

mliradelc: Can you add custom fields? The current LDAP you can add roles from a custom parameter is set in LDAP, but not save it like is with the name or email.

PR #4478, opened 2022-11-18
Bug fix: publish engage woh with merge SKIP the operation if media package not in search index

### LINKED ISSUE #4477: PublishEngage WOH with the "merge" strategy fails if the media package does not exist in search instead of SKIPping
**Describe the bug**
A clear and concise description of what you see:

PublishEngage WOH with the "merge" strategy fails if the media package does not exist in the search index. It should just SKIP the operation.
Also, it first distributes all the elements before checking the existence of the media package in the search index, which potentially leaves files distributed that are not referenced by anything. 

**To Reproduce**
1. Publish a media package
2. Retract it
3. Publish again with the "merge" strategy
4. WOH fails instead of SKIPping

**Environment Information**
 - Opencast 11.x, 12.x, 13.x

rute-santos:
Fixes #4477 

The fix is simple, but I also added some unit tests because there were almost none :)

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
+++ b/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java
@@ -115,26 +115,29 @@ public class PublishEngageWorkflowOperationHandler extends AbstractWorkflowOpera
   private static final Logger logger = LoggerFactory.getLogger(PublishEngageWorkflowOperationHandler.class);
 
   /** Configuration properties id */
-  private static final String ENGAGE_URL_PROPERTY = "org.opencastproject.engage.ui.url";
-  private static final String STREAMING_PUBLISH_PROPERTY = "org.opencastproject.publish.streaming.formats";
+  static final String ENGAGE_URL_PROPERTY = "org.opencastproject.engage.ui.url";
+  static final String STREAMING_PUBLISH_PROPERTY = "org.opencastproject.publish.streaming.formats";
 
   /** Workflow configuration option keys */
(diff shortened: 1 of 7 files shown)


mtneug:
Without looking at the code: is this also fixing the dangling files when republishing using the merge strategy? We switched to the retract strategy because of this bug. When I noticed that problem, I was not sure if this is easily fixable with Opencast since elements in the publication have no reference to elements in the media package (and if they had, they could get lost when the element is removed from the archive).

rute-santos:
Hi @mtneug ,
No, it does not include that fix. I can try to fix that, but it will take me a few days. Is there a ticket with details? I vaguely remember seeing this bug...

mtneug:
Is this not the issue you also describe in #4477? Also, I don't know how we could fix this since keeping track of which published element should be merged with which element in the archive seems hard/impossible to know in every case.

rute-santos:
Yes, it fixes the issue I described in #4477 , which is very specific to the case where the strategy is "merge" and the media package does not  previously exist in the search index.
I believe you were asking about a more complex bug: strategy "merge" leaving files around when 1) it forces their removal (elements are never deleted from the distribution directory) and 2) replacing the same flavor with a different file (same). I agree with you that the fix may not be simple and not sure if even possible... Please correct me if I'm wrong!

rute-santos:
Hmm Just thinking... For cases above: 1), when forcing the removal, we can just retract the element; for 2) we could compare the current mp element IDs to the previously distributed element urls (they have the original element IDs) for each flavor set and retract elements where the IDs don't match (same IDs will be overwritten anyway). Would that work?

Arnei (approved):
Might not fix all the issues with the merge strategy, but small improvements are also helpful and this PR achieves what it set out to do.

PR #4482, opened 2022-11-21
Add silient detection based on subtitles (webvtt-to-cutmarks woh)

darolfes:
This pullrequests adds a workflow operation, which transforms a WebVTT Subtitle file into a CutMarkJson, which can then be used for cutting.
Primary use case is silence detection based on subtitle information.

It also moves the WebVTT parser into its own module, so that the parser can be included in different modules.

Documentation: `docs/guides/admin/docs/workflowoperationhandlers/webvtt-to-cutmarks.md`


Arnei (approved):
Works as advertised, code looks fine, documentation is there, LGTM!

mtneug:
@Arnei this could go into r/13.x right?

Arnei:
Doesn't change existing behaviour, just moves some code around and adds a new WOH, so it should be fine to go into r/13.x I think.

darolfes:
rebased to r/13.x

PR #4485, opened 2022-11-22
Skip publication channel to workspace operation if media package has no publication

geichelberger:
Check if the media package has a publication for the configured channel and skip instead of failing.
Contains a minor refactoring.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/PublicationChannelToWorkspace.java
+++ b/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/PublicationChannelToWorkspace.java
@@ -92,7 +92,7 @@ public class PublicationChannelToWorkspace extends AbstractWorkflowOperationHand
 
     // Skip if the media package does not contain a matching publication
     if (publication.isEmpty()) {
-      logger.debug("The media package '{}' contains no publications on channel '{}', skipping.",
+      logger.warn("The media package '{}' contains no publications on channel '{}', skipping.",
           mediaPackage.getIdentifier(), publicationChannel);
       return createResult(mediaPackage, Action.SKIP);
     }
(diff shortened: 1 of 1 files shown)


modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/PublicationChannelToWorkspace.java
```diff
@@ -90,67 +89,62 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
-        tracks.stream().forEach(track -> {
-          configuredTargetTags.stream().forEach(targetTag -> track.addTag(targetTag.toString()));
-        });
-      }
-      tracks.stream().forEach(track -> mediaPackage.add(track));
+
+    // Skip if the media package does not contain a matching publication
+    if (publication.isEmpty()) {
+      logger.debug("The media package '{}' contains no publications on channel '{}', skipping.",
+          mediaPackage.getIdentifier(), publicationChannel);
```
mliradelc:
```suggestion
      logger.warn("The media package '{}' contains no publications on channel '{}', skipping.",
          mediaPackage.getIdentifier(), publicationChannel);
```

I like this more  `warn` than `debug`, because if there is an error on the workflow, I want to be warned that the MP doesn't contain the media to be published.

mliradelc (changes requested):

mliradelc:
It is a nice PR, 

What is the difference between your PR and use `fail-on-error= false`

Anyway, the debug message is something that was missing, not knowing what match fail it is very exhausting.

geichelberger:
We have no rules on how a workflow operation should be implemented. For me, a workflow is only valid to fail if it throws a `WorkflowOperationException`, else it's just a programming error.

> What is the difference between your PR and use fail-on-error= false
You can not distinguish between a publication is not present, and an error is happening.

And having no publication is like trying to copy a flavor from the publication channel which is not present, which does not fail either.

Arnei (approved):
Code looks good to me and works as advertised.

I'm not sure whether it is better to skip or throw an error in case of an empty publication. However, I do agree that explicitly handling that case is an improvement regardless, so I'm inclined to approve this.

PR #4498, opened 2022-12-01
Fix possible duplicate notes in Admin UI

Arnei:
There is a bug in the admin UI where you can accidentally duplicate the notes for an event if you behave a certain way. This fixes that.

How to reproduce the bug this fixes in detail:

0. Enable the notes column in the event view of the Admin UI (disabled by default)
1. Write some text in an empty notes input fields.
2. Click somewhere else on the page, so the input field loses focus.
3. Click back into the input field you just wrote into, and write some more.
4. Reload the page
5. See duplicated notes input field.

 Repeating step 2 and 3 before going to step 4 increases the number of
 duplications.


lkiesow:
@Arnei, can you fix the test failure:

```
[INFO] /home/runner/work/opencast/opencast/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventsController.js
[INFO]   224:9  error  Missing semicolon  semi
```

PR #4513, opened 2022-12-06
Adds Whisper STT to SpeechToText WoH

mliradelc:
This PR adds [Whisper](https://www.github.com/openai/whisper) as a new STT engine. The new engine comes with additional configuration options. These options will not break any current setup as the previous settings remain as default.

The new engine also comes with a configuration file to choose what model to use and the CLI command.

#### TO-DO:
* Write documentation
    * [x] How to install whisper
    * [x] About the new engine
* [x] Solve “Not file found” after transcription
* [x] Clean PR from files and annotations for debugging
* [x] Add language model as configuration parameter
* [x] Change Target branch to `r/13.x`

#### Next PRs planed:
* Translation to english option to WoH
* Automatic language recognition option

#### Current caveats:
~It is not possible to change engine without an opencast restart. If the parameter changes while OC is executing, it will be a `null pointer error`when the STT WoH runs.~ Now works as expected (Thanks @geichelberger !)

____
#### Standard PR checklist

* [x] have a concise title
* [x] pass automated tests


modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -214,11 +220,23 @@ protected OrganizationDirectoryService getOrganizationDirectoryService() {
...
-  @Reference
+  //TODO: Improve OSGI wiring to change STT Engine implementation during OC execution
+  @Reference(
+          policy = ReferencePolicy.DYNAMIC,
+          policyOption = ReferencePolicyOption.GREEDY,
+          cardinality = ReferenceCardinality.MULTIPLE,
+          target = "(engineType=vosk)",
+          unbind = "unsetSpeechToTextEngine",
+          bind = "setSpeechToTextEngine"
+  )
```
geichelberger: It would be best if you enforced the Reference name with `name = "SpeechToTextEngine"`. You do not need cardinality `MULTIPLE`, `MANDATORY` should be sufficient. Use the policy `STATIC` instead of `DYNAMIC`. This will ensure that the component gets deactivated if the bound service is unavailable.

mliradelc: Now works! No more `nullpointer error` when the parameter changes during execution. Thanks for the tip!

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -214,11 +220,24 @@ protected OrganizationDirectoryService getOrganizationDirectoryService() {
...
   public void setSpeechToTextEngine(SpeechToTextEngine engine) {
     this.speechToTextEngine = engine;
   }
 
+  public void unsetSpeechToTextEngine(SpeechToTextEngine engine) {
+    this.speechToTextEngine = null;
+  }
+
```
geichelberger: That's not necessary.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -214,11 +220,24 @@ protected OrganizationDirectoryService getOrganizationDirectoryService() {
...
+  //TODO: Improve OSGI wiring to change STT Engine implementation during OC execution
+  @Reference(
+          name = "SpeechToTextEngine",
+          policy = ReferencePolicy.STATIC,
+          policyOption = ReferencePolicyOption.GREEDY,
+          cardinality = ReferenceCardinality.MANDATORY,
+          target = "(engineType=vosk)",
+          unbind = "unsetSpeechToTextEngine",
+          bind = "setSpeechToTextEngine"
+  )
```
geichelberger:
```suggestion
  @Reference(
          name = "SpeechToTextEngine",
          target = "(engineType=vosk)"
  )
```

The default values for policy and cardinality are `STATIC` and `MANDATORY`. Instead of the policy option `GREEDY`, I would use the default `RELUCTANT`; if a new service becomes available, it will bind to it. This should not happen because we should have only one service that matches the target filter.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java
```diff
@@ -124,6 +127,9 @@ public void activate(ComponentContext cc) {
   }
 
 
+
+
+
```
geichelberger: Remove the blank lines

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -0,0 +1,156 @@
...
+
+/** Whisper implementation of the Speech-to-text engine interface. */
+@Component(
+    immediate = true,
+    service = {
+        SpeechToTextEngine.class
+    },
+    property = {
+        "service.description=Whisper implementation of the SpeechToTextEngine interface",
+        "service.pid=org.opencastproject.speechtotext.impl.engine.WhisperEngine",
```
geichelberger: The property `service.pid` is not needed.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -0,0 +1,156 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+
+/** Whisper implementation of the Speech-to-text engine interface. */
+@Component(
+    immediate = true,
+    service = {
+        SpeechToTextEngine.class
+    },
```
geichelberger: By default, implemented interfaces will be used as services. You can remove that part.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -0,0 +1,156 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+
+/** Whisper implementation of the Speech-to-text engine interface. */
+@Component(
+    immediate = true,
```
geichelberger: You can remove that. If true, the component will immediately be activated upon becoming satisfied. The default value is false, meaning the component will only be activated if it's satisfied and another OSGi service needs it. Delayed activation.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -0,0 +1,156 @@
...
+
+public class WhisperEngine implements SpeechToTextEngine {
+
+  private static final Logger logger = LoggerFactory.getLogger(WhisperEngine.class);
+
+  /** Name of the engine. */
+  private static final String engineName = "Whisper";
+
+  /** Config key for setting the path to Whisper. */
+  private static final String WHISPER_EXECUTABLE_PATH_CONFIG_KEY = "whisper.root.engine.path";
```
geichelberger: In the config, you use `whisper.root.path` as the key.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
geichelberger: The same is valid for the VoskEngine

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
```diff
@@ -1,3 +1,7 @@
 # Job load for a speech-to-text job
 # Default: 0.8
 #job.load.speechtotext=0.8
+
+# Select STT engine.
+# Available engines: vosk, whisper | Default: vosk
+SpeechToTextEngine.target=(engineType=vosk)
```
geichelberger: The config key is case-sensitive. Maybe you want to replace the reference name with something just lowercase, like all our config keys.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/VoskEngine.java
```diff
@@ -47,7 +47,8 @@
     },
     property = {
         "service.description=Vosk implementation of the SpeechToTextEngine interface",
-        "service.pid=org.opencastproject.speechtotext.impl.engine.VoskEngine"
+        "service.pid=org.opencastproject.speechtotext.impl.engine.VoskEngine",
```
geichelberger: The property service.pid is not needed.

geichelberger (changes requested):
These are just a few remarks and OSGi-specific optimizations.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
mliradelc: ~True for whisper, but I left configurable because there are several whisper forks right now to try,  One of the most interesting is [Whisper's code translation to C++](https://github.com/ggerganov/whisper.cpp)~

Misread, thanks for noting that 😉

mliradelc:
Documentation is ready for review. I grouped all transcription modules in only one folder in the documentation. If you think is out of context please let me know and I will do another PR.

geichelberger:
Isn't that the wrong documentation location? The strange thing is we have transcription services and the new speech-to-text stuff. The vosk and whisper engines are part of the speech-to-text service, not the transcription service.

mliradelc:
The Transcription services and the STT make the same goal, transcript audio to generate a VTT file, for me it was very odd that these where separate and for the user perspective, one can think that the only way you have to make transcriptions is using a cloud service.

But I can make a subsection for "on premise" and "Cloud" solutions.

Under the hood are different services, but for daily use make the same task.

geichelberger (approved):
LGTM

PR #4515, opened 2022-12-08
Allow hotkeys in create dialogs in input elements

Arnei:
In the Admin UI, the "Create event" and "Create series" dialogs both have hotkeys with which one can easily jump to the previous/next tab of the dialog (provided the requirements to go to the next tab are met of course). However, one can get "stuck" in the metadata views if an input field or dropdown is in focus, as the hotkeys won't trigger in that case. This fixes that by allowing them to trigger.

Behaviour for any other hotkeys is unchanged.

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/directives/wizardDirective.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/directives/wizardDirective.js
@@ -269,6 +269,18 @@ angular.module('adminNg.directives')
 
       HotkeysService.activateHotkey(scope, 'add_media.next_tab', (event) => {
         if (!scope.wizard.isLast()) {
+          // Trigger submit function on form elements, thereby updating the requiredMetadata array of the form.
+          // This allows the hotkey to trigger the "next" step while the form element is still in focus, if all
+          // requirements are met.
+          // It would  be preferable to just trigger 'blur'/'change' on the element, but since the execution of their
+          // callbacks is asynchronous, there is no guarantee that the callbacks will have run before the check for
+          // the "next" step is run.
+          var targetElement = angular.element(event.target);
+          if (targetElement.parents('form').length > 0) {
+            var targetScope = targetElement.scope();
+            targetScope.submit();
+          }
+
           scope.wizard.toTab(event, 'next');
         } else {
           scope.submit();
--- a/modules/admin-ui-frontend/app/scripts/shared/directives/wizardDirective.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/directives/wizardDirective.js
@@ -274,12 +274,14 @@ angular.module('adminNg.directives')
           scope.submit();
         }
         event.preventDefault();
-      });
+      },
+      ['INPUT', 'TEXTAREA', 'SELECT']);
 
       HotkeysService.activateHotkey(scope, 'add_media.previous_tab', (event) => {
         scope.wizard.toTab(event, 'previous');
         event.preventDefault();
-      });
+      },
+      ['INPUT', 'TEXTAREA', 'SELECT']);
     }
   };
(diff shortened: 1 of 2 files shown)


lkiesow (changes requested):
This makes things better and we can merge it like this since it's still an improvement, but it doesn't always work.
Specifically (tested in Firefox):

- Open admin interface
- Click on “Add Event”
   - The title field is focused automatically
 - Enter title, e.g. “asdf”
 - Hit `Alt` + `Return`
   - The shortcut is not invoked

Going back to the dialog later works.
Maybe it does not work because before setting the title, the next step is not reachable?

Arnei:
> This makes things better and we can merge it like this since it's still an improvement, but it doesn't always work. Specifically (tested in Firefox):
> 
>     * Open admin interface
> 
>     * Click on “Add Event”
>       
>       * The title field is focused automatically
> 
>     * Enter title, e.g. “asdf”
> 
>     * Hit `Alt` + `Return`
>       
>       * The shortcut is not invoked
> 
> 
> Going back to the dialog later works. Maybe it does not work because before setting the title, the next step is not reachable?

Your last assumption is correct. The Create dialogs will sometimes not let the user advance to the next tab unless certain requirements are met (i.e. a non-empty title field that has been confirmed by pressing `Return` at least once). The shortcuts respect these restrictions, and this PR should not have changed their behaviour in this regard what-so-ever.

But maybe pressing `Alt` + `Return` *should* also confirm the current action (i.e. confirm the entered title)?

lkiesow:
> But maybe pressing Alt + Return should also confirm the current action (i.e. confirm the entered title)?

I think it would make sense to evaluate the hotkey after the value has been added. After all, from a user's perspective, the complaint that no title is set is not correct. I just set one in my example. And clicking the “Next” button also actually works.

Arnei:
> I think it would make sense to evaluate the hotkey after the value has been added.

I believe you are suggesting this, because currently, when you type into the title field and press `Alt` + `Enter`, it will add the value (because `Alt`+`Enter` as a shortcut is ignored, and only the `Enter` is parsed, thus adding the value). Since this PR would change that behaviour, evaluating the hotkey after the value has been added does not make sense I think, because the value will not be added by pressing `Alt`+`Enter` anymore (because `Alt`+`Enter` is recognized as a shortcut, there is no single `Enter` parsed and thus no value is added).

I hope that was understandable. Pretty sure we mean the same thing anyway, and this just implementation detail ^^

> After all, from a user's perspective, the complaint that no title is set is not correct. I just set one in my example. And clicking the “Next” button also actually works.

Agreed. So pressing `Alt`+`Enter should work the same in that regard as clicking on Next (even though the Next button is marked as inactive when entering a title into a previously empty title field, but I digress).

lkiesow:
> I believe you are suggesting this, because currently, when you type into the title field and press Alt + Enter, it will add the value

Honestly, no. That never crossed my mind.  While your explanation makes sense from a technical point of view if you are very familiar with the input system of the admin interface, I don't think this is very helpful from a user's perspective.

- As a user I understand that a title is required since it's marked with a red asterisk and the button “next” is disabled when I enter the view.
- As soon as I enter anything in the title field, I expect to have fulfilled the requirements. I provided a title. I don't care about internal states of the web app.
- The current user experience supports this explanation: Once I have entered anything in the title field, I can click next. I don't have to confirm any input.

To argue that a shortcut should work different and that a user must know of and deal with a different behavior is really confusing. As a developer, I understand that clicking outside the input field will confirm the input and that is why I can click “Next”. But if that's the case, gitting `Alt` + `Enter` should work the same, confirming the input and then executing the shortcut's functionality of going to the next view.

lkiesow (approved):

PR #4519, opened 2022-12-09
Reduce number of snapshots taken in the new editor backend

Arnei:
In the new editor backend, various functions would create snapshots if run, e.g. the "Save subtitles"-function would create one and then the "Save thumbnails"-function would create one as well. This changes behaviour so that a snapshot is only taken once after all the potentially snapshot worthy changes have been made.

Does not solve the problem of metadata having their own endpoint for saving, and thus causing another snapshot.

### DIFF
--- a/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
+++ b/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
@@ -364,7 +364,7 @@ public class EditorServiceImpl implements EditorService {
   /**
    * Check if a media URL can be served from this server.
    *
-   * @param mediaUrl
+   * @param uri
    *      URL locating a media file
    * @return
    *      If the file is available locally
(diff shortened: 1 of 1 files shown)


lkiesow:
Looks reasonable, but what's interesting is that if you click save, you still end up with multiple snapshots:

```
2022-12-13 20:46:41,659 | INFO  | (AssetManagerImpl:1321) - Creating new version 4 of media package 01
2022-12-13 20:46:41,729 | INFO  | (AssetManagerImpl:439) - Trigger update handlers for snapshot 01, version 4
2022-12-13 20:46:42,052 | INFO  | (AssetManagerImpl:1321) - Creating new version 5 of media package 01
2022-12-13 20:46:42,102 | INFO  | (AssetManagerImpl:439) - Trigger update handlers for snapshot 01, version 5
```

lkiesow:
> Does not solve the problem of metadata having their own endpoint for saving, and thus causing another snapshot.

Actually reading the description of this pull request solves that riddle :D
Let's get this merged like this for now.

PR #4521, opened 2022-12-09
Support AWS S3 Glacier tiering

gregorydlogan:
This pull request adds AWS S3 storage tiering to the existing AWS AssetManager S3 module.

This is a forward port of production code at Manchester, with the addition of some bug fixes, style changes, and a bunch of mostly-done unit tests.  Initially developed against `r/12.x`, I don't think there's a particular requirement that it go there vs `r/13.x`.

Note: There is *no* UI for this - you need to use the REST docs to trigger things.  I'm open to adding this as later work, but especially with the new new admin UI in the near-term pipeline I didn't see a lot of point of putting a ton of effort into the old new admin UI.

This work was paid for by the University of Manchester.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
+++ b/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
@@ -12,9 +12,9 @@ org.opencastproject.assetmanager.aws.s3.enabled=false
 # Example: cluster-name-file-assetmanager
 #org.opencastproject.assetmanager.aws.s3.bucket=
 
-+# The number of days which to restore objects in the Glacier storage class for.
-+# Default: 2
-+#org.opencastproject.assetmanager.aws.s3.glacier.restore=2
+# The number of days which to restore objects in the Glacier storage class for.
+# Default: 2
+#org.opencastproject.assetmanager.aws.s3.glacier.restore=2
 
 # The AWS access key ID to use for connecting to S3.
 #org.opencastproject.assetmanager.aws.s3.access.id=
--- a/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
+++ b/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
@@ -12,6 +12,10 @@ org.opencastproject.assetmanager.aws.s3.enabled=false
 # Example: cluster-name-file-assetmanager
 #org.opencastproject.assetmanager.aws.s3.bucket=
 
++# The number of days which to restore objects in the Glacier storage class for.
++# Default: 2
++#org.opencastproject.assetmanager.aws.s3.glacier.restore=2
+
 # The AWS access key ID to use for connecting to S3.
 #org.opencastproject.assetmanager.aws.s3.access.id=
 
--- a/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
+++ b/etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
@@ -14,7 +14,7 @@ org.opencastproject.assetmanager.aws.s3.enabled=false
 
 # The number of days which to restore objects in the Glacier storage class for.
 # Default: 2
-#org.opencastproject.assetmanager.aws.s3.glacier.restore=2
+#org.opencastproject.assetmanager.aws.s3.glacier.restore.days=2
 
 # The AWS access key ID to use for connecting to S3.
 #org.opencastproject.assetmanager.aws.s3.access.id=
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
@@ -20,6 +20,11 @@
  */
 package org.opencastproject.assetmanager.api;
 
+import static org.opencastproject.util.data.functions.Misc.chuck;
+
+import org.opencastproject.security.api.UnauthorizedException;
+import org.opencastproject.util.NotFoundException;
+
 /**
  * A common exception indicating various issues.
  */
@@ -38,4 +43,26 @@ public class AssetManagerException extends RuntimeException {
   public AssetManagerException(Throwable cause) {
     super(cause);
   }
+
+  /** Returns true if the exception is caused by a {@link org.opencastproject.security.api.UnauthorizedException}. */
+  // todo is an authorization failure really unrecoverable?
+  public boolean isCauseNotAuthorized() {
+    return getCause() instanceof UnauthorizedException;
+  }
+
+  /** Returns true if the exception is caused by a {@link org.opencastproject.util.NotFoundException}. */
+  public boolean isCauseNotFound() {
+    return getCause() instanceof NotFoundException;
+  }
+
+  /**
+   * If the exception is caused by an {@link org.opencastproject.security.api.UnauthorizedException}
+   * rethrow it, otherwise do nothing.
+   */
+  public void rethrowUnauthorizedException() {
+    if (isCauseNotAuthorized()) {
+      chuck(getCause());
+    }
+  }
+
 }
--- a/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
+++ b/modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
@@ -20,8 +20,6 @@
  */
 package org.opencastproject.assetmanager.api;
 
-import static org.opencastproject.util.data.functions.Misc.chuck;
-
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.util.NotFoundException;
 
@@ -53,16 +51,6 @@ public class AssetManagerException extends RuntimeException {
   /** Returns true if the exception is caused by a {@link org.opencastproject.util.NotFoundException}. */
   public boolean isCauseNotFound() {
     return getCause() instanceof NotFoundException;
-  }
-
-  /**
-   * If the exception is caused by an {@link org.opencastproject.security.api.UnauthorizedException}
-   * rethrow it, otherwise do nothing.
-   */
-  public void rethrowUnauthorizedException() {
-    if (isCauseNotAuthorized()) {
-      chuck(getCause());
-    }
   }
 
 }
--- a/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
+++ b/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
@@ -336,9 +336,9 @@ public class AwsS3AssetStore extends AwsAbstractArchive implements RemoteAssetSt
   }
 
   /**
-   +   * Return the object key of the asset in S3
-   +   * @param storagePath asset storage path
-   +   */
+   * Return the object key of the asset in S3
+   * @param storagePath asset storage path
+   */
   public String getAssetObjectKey(StoragePath storagePath) throws AssetStoreException {
     try {
       AwsAssetMapping map = database.findMapping(storagePath);
@@ -390,7 +390,7 @@ public class AwsS3AssetStore extends AwsAbstractArchive implements RemoteAssetSt
       StorageClass objectStorageClass = StorageClass.fromValue(getObjectStorageClass(objectName));
 
       if (storageClass != objectStorageClass) {
-        /* objects can only be retrived from Glacier not moved */
+        /* objects can only be retrieved from Glacier not moved */
         if (objectStorageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) {
           logger.warn("S3 Object {} can not be moved from storage class {}", objectStorageClass);
           return objectStorageClass;
--- a/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
+++ b/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
@@ -21,6 +21,8 @@
 
 package org.opencastproject.assetmanager.aws.s3;
 
+import static java.lang.String.format;
+
 import org.opencastproject.assetmanager.api.storage.AssetStore;
 import org.opencastproject.assetmanager.api.storage.AssetStoreException;
 import org.opencastproject.assetmanager.api.storage.RemoteAssetStore;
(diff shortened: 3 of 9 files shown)


gregorydlogan:
Requesting review from the other known S3 folks (@rute-santos, @karendolan, @mtneug).
Pinging @JamesUoM 'cause I'm *finally* done until someone finds the bugs :D

etc/org.opencastproject.assetmanager.aws.s3.AwsS3AssetStore.cfg
```diff
@@ -12,6 +12,10 @@ org.opencastproject.assetmanager.aws.s3.enabled=false
 # Example: cluster-name-file-assetmanager
 #org.opencastproject.assetmanager.aws.s3.bucket=
 
+# The number of days which to restore objects in the Glacier storage class for.
+# Default: 2
+#org.opencastproject.assetmanager.aws.s3.glacier.restore=2
```
mtneug: Is `.days` missing?

gregorydlogan: Yep, fixed.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
+        if (objectStorageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) {
+          logger.warn("S3 Object {} can not be moved from storage class {}", objectStorageClass);
+          return objectStorageClass;
+        }
+
+        /* Only put suitable objects in Glacier */
+        if (storageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) {
+          GetObjectTaggingRequest gotr = new GetObjectTaggingRequest(bucketName, objectName);
+          GetObjectTaggingResult objectTaggingRequest = s3.getObjectTagging(gotr);
+          logger.info("S3 object {} is not suitable for storage class {}", objectName, storageClass);
```
rute-santos: Shouldn't this log msg be inside the "if" below?

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
```diff
@@ -38,4 +43,26 @@ public AssetManagerException(String message, Throwable cause) {
...
+  /**
+   * If the exception is caused by an {@link org.opencastproject.security.api.UnauthorizedException}
+   * rethrow it, otherwise do nothing.
+   */
+  public void rethrowUnauthorizedException() {
+    if (isCauseNotAuthorized()) {
+      chuck(getCause());
+    }
+  }
+
```
rute-santos: isCauseNotAuthorized(), isCauseNotFound(), rethrowUnauthorizedException() don't seem to be used anywhere?

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -270,6 +304,24 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
       ObjectMetadata objMetadata = s3.getObjectMetadata(bucketName, objectName);
       logger.trace("Got object metadata for: {}, version is {}", objectName, objMetadata.getVersionId());
+
+      // Tag objects that are suitable for Glacier storage class
+      // NOTE: Use of S3TransferManager means that tagging has to be done as a separate request
+      if (mimeType.isSome()) {
+        switch (mimeType.get().getType()) {
+          case "audio":
+          case "image":
+          case "video":
```
rute-santos, line 314: Just curious why only these types are being sent to Glacier? Is it because they tend to be larger? Maybe this should be configurable?

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
     }
   }
 
+  /**
+   +   * Return the object key of the asset in S3
+   +   * @param storagePath asset storage path
+   +   */
```
rute-santos: Remove the "+"s :)

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
+    }
+  }
+
+  private StorageClass modifyObjectStorageClass(String objectName, StorageClass storageClass)
+          throws AssetStoreException {
+    try {
+      StorageClass objectStorageClass = StorageClass.fromValue(getObjectStorageClass(objectName));
+
+      if (storageClass != objectStorageClass) {
+        /* objects can only be retrived from Glacier not moved */
```
rute-santos: Typo: retrived :)

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/endpoint/AwsS3RestEndpoint.java
```diff
@@ -0,0 +1,450 @@
...
+              type = RestParameter.Type.STRING,
+              description = "The media package indentifier.")
+      },
+      restParameters = {
+          @RestParameter(
+              name = "storageClass",
+              isRequired = true,
+              type = RestParameter.Type.STRING,
+              description = "The S3 storage class, valid terms STANDARD, STANDARD_IA, INTELLIGENT_TIERING, ONEZONE_IA,"
+                          + "REDUCED_REDUNDANCY and GLACIER. See https://aws.amazon.com/s3/storage-classes/")
```
rute-santos: Shouldn't this be DEEP_ARCHIVE Instead of REDUCED_REDUNDANCY?

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
+        if (objectStorageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) {
```
rute-santos, line 394: Can't an object be moved from Glacier to Deep Archive?

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
+
   /**
    *
    */
   @Override
-  protected InputStream getObject(AwsAssetMapping map) throws AssetStoreException {
+  protected InputStream getObject(AwsAssetMapping map) {
+    String storageClassId = getObjectStorageClass(map.getObjectKey());
+
+    if (StorageClass.Glacier.equals(storageClassId)) {
```
rute-santos: What about DeepArchive? We do not restore from it?
My IDE gives me a warning on this line. Maybe use StorageClass.Glacier.name() to compare to string?

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -303,6 +447,90 @@ protected InputStream getObject(AwsAssetMapping map) throws AssetStoreException
...
+      // Wait min restore time and then poll ofter that
+      try {
+        // Check as restore might have already been initiated
+        if (prevOngoingRestore != null && !prevOngoingRestore) {
+          Thread.sleep(RESTORE_MIN_WAIT);
+        }
+
+        while (s3.getObjectMetadata(bucketName, objectName).getOngoingRestore()) {
+          Thread.sleep(RESTORE_POLL);
+        }
```
rute-santos, line 535: This can potentially wait for hours, right? So an AssetManager getAsset() could time out and throw an exception, maybe even fail a workflow... Is there a better way to handle this? Maybe make use of the "availability" flag in the snapshot or something like that. I know this is a lot more work... :)

rute-santos (commented):
Hi @gregorydlogan,
Some comments:

- My biggest concern is the waiting for potentially hours when restoring from Glacier and Glacier Deep Archive and not handling this situation in the code that requests it.
- It would be nice to have some documentation on this (with pointers to aws documentation about costs, etc)
- Why not support Glacier Instant Retrieval?
- Code formatting seems to be off in some files.

JamesUoM:
Finally responding to my summons and will be working on this till done.

JamesUoM:
A little background: This code was original written before S3 added instant access glacier, and so just GLACIER was deep storage. It was is worth storaging small files in glacier. Also catalogs such as series could be objects shared between many mediapackages (assest mapping via checksums) which may or may not be in glacier.  

Initially we intended classic GLACIER storage to be a manual process for deep archiving of old materials and did write code to enable the archive service (3.x) to wait for GLACIER to "restore" objects - however we never put this in production as at the same time the PR was ready S3 added instant access GLACIER. They even added it to Intelligent Tiering and over night our storage costs halved so  we deemed it not necessary to use the classic GLACIER storage class.

Moving object too increasingly deeper storage is best handled but S3 lifecycle policies. These endpoints were a manual way of retrieving old mediapackages to instantly accessible levels in rare the event they were needed.

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/AssetManagerException.java
gregorydlogan: These are used in `AwsS3RestEndpoint`.  I thought that too, until I deleted them an the compiler complained :D

Edit: Actually, `rethrowUnauthorizedException` doesn't seem to be used on second inspection.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/endpoint/AwsS3RestEndpoint.java
gregorydlogan: I mean, it's not actually checked on the service, so as long as AWS has a matching storage class it'll do it.  Will still change RR to DA.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
gregorydlogan: Yep, good catch.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
gregorydlogan: This came from Manchester, but I would think that 99% of the Glacier-able files are going to fall into audio, video, or images.  I'm not opposed to adding configurable support here, but I'm struggling to imagine the usecase.

JamesUoM: See constraints here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-transition-general-considerations.html

File smaller than 128k are not transitioned to Glacier. Also see my comment below about series or other policy catalogues that may be mapped (simulated softlinks) for multiple  mediapackages

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/endpoint/AwsS3RestEndpoint.java
JamesUoM: The current list of values StorageClass ID strings is `STANDARD, STANDARD_IA, ONEZONE_IA, INTELLIGENT_TIERING, GLACIER_IR, GLACIER, DEEP_ARCHIVE` .

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
JamesUoM: @rute-santos yes your right. The logic needs to be changed to at least allow glacier -> deeparchive. I can find answer for deeparchive -> glacier so will test it- (the console UI will only let you restore an object in glacier or deeparchive so you can't transition it directly by that means)

NOTE  Transition lifecycle policies objects can only move objects to deeper storage, but you can move objects between all classes above Glacier.

JamesUoM: OK, to change the storage class of an object in glacier or deep archive you must restore it first. You can them change it's class to whatever you like. The logic should be change to something like:
```
if ( (objectStorageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) && ! objectRestored )
  {
     log.warn("S3 Object {} must be restored before changing it's storage class");
   ...
```

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -283,11 +335,103 @@ protected AwsUploadOperationResult uploadObject(File origin, String objectName)
...
+
+  private StorageClass modifyObjectStorageClass(String objectName, StorageClass storageClass)
+          throws AssetStoreException {
+    try {
+      StorageClass objectStorageClass = StorageClass.fromValue(getObjectStorageClass(objectName));
+
+      if (storageClass != objectStorageClass) {
+        /* objects can only be retrieved from Glacier not moved */
+        if (objectStorageClass == StorageClass.Glacier || objectStorageClass == StorageClass.DeepArchive) {
+          logger.warn("S3 Object {} can not be moved from storage class {}", objectStorageClass);
```
JamesUoM: message is missing the objectName argument

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
gregorydlogan: Done.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
gregorydlogan: Fixed!

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
gregorydlogan: Yes, absolutely will wait for hours.  Will take a look at the availability flag, but I'm not sure how useful that would be - the things trying to use the file(s) have to know about them too probably.  This may be a case of requiring business/institution logic (an admin rule that says 'don't do that') rather than properly supporting this in Opencast.  At least until the new new admin UI hits and someone funds the UI work to make this visible...

JamesUoM: Yes, you are correct. the `wait=true` option is not helpful if initiated by a http request, but this is meant to be an internal switch feature for code that knows what it's doing and is false when using the endpoints. I'm ok with it being removed. It is expected the  REST endpoints in this PR are to used manually ie someone initiates a restore and then checks whether it is available before initiating a workflow on the mediapackage. See my explanation here: https://github.com/opencast/opencast/pull/4521#issuecomment-1384166361

I have worked on code that added the wait-until-available concept that is in the workspace to our OC 3.x archive service, this could be ported to the asset manager so you could initiate a WF on a GLAICER mediapackage and the asset manager would request it from the workspace. The workspace would initiate the restore and tell the asset manager it's unavailable and to poll again. Repeated until the mediapackage is restored.

PR #4560, opened 2022-12-22
chapter extraction as segmentation

darolfes:
This pr extends the VideoSegmenter with the ability to extract available chapter information from video tracks and use this information for the track segmentation.

### DIFF
--- a/modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
+++ b/modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
@@ -111,7 +111,7 @@ public class VideoSegmenterServiceImpl extends AbstractJobProducer implements
   private class Chapter {
     protected double start;
     protected double end;
-    protected String title;
+    protected Optional<String> title;
   };
 
   /** Path to the executable */
@@ -538,17 +538,20 @@ public class VideoSegmenterServiceImpl extends AbstractJobProducer implements
     double timebase = defaultTimebase;
     long start = -1;
     long end = -1;
-    StringBuilder title = new StringBuilder();
+    Optional<StringBuilder> title = Optional.empty();
 
     String line = reader.readLine();
     int lineNumber = 1;
-    while (null != line) {
+    if (line == null) {
+      return chapters;
+    }
+    while (true) {
       // begin parsing
       if (state == 0 && ";FFMETADATA1".equals(line)) {
         state++;
       }
       // ignore comments, empty lines
-      else if (line.startsWith(";") || line.startsWith("#") || line.isEmpty()) { }
+      else if (line != null && (line.startsWith(";") || line.startsWith("#") || line.isEmpty())) { }
       // search for chapter begin
       else if (state == 1 && "[CHAPTER]".equals(line)) {
         state++;
@@ -634,12 +637,15 @@ public class VideoSegmenterServiceImpl extends AbstractJobProducer implements
       // Being processing of title
       else if (state == 5) {
         if (!line.startsWith("title=")) {
-          throw new ParseException("Failed to parse FFMETADATA: CHAPTER title field missing", lineNumber);
+          state = 7;
+          continue;
         }
 
         String fakeLine = Arrays.stream(line.split("="))
                 .skip(1)
                 .collect(Collectors.joining());
+
+        title = Optional.of(new StringBuilder());
 
         // Process title further in next state
         line = fakeLine;
@@ -652,7 +658,7 @@ public class VideoSegmenterServiceImpl extends AbstractJobProducer implements
         boolean isEscaped = false;
         for (int codePoint : codePoints) {
           if (isEscaped) {
-            title.appendCodePoint(codePoint);
+            title.get().appendCodePoint(codePoint);
 
             isEscaped = false;
           }
@@ -669,30 +675,53 @@ public class VideoSegmenterServiceImpl extends AbstractJobProducer implements
                         + " CHAPTER title field '=' ';' '#' '\\' '\\n' have to be escaped", lineNumber);
             }
             else {
-              title.appendCodePoint(codePoint);
+              title.get().appendCodePoint(codePoint);
             }
           }
         }
 
         if (!isEscaped) {
-          state = 1;
-
-          Chapter chapter = new Chapter();
-          chapter.title = title.toString();
-          chapter.start = timebase * start;
-          chapter.end = timebase * end;
-
-          chapters.add(chapter);
-
-          timebase = defaultTimebase;
-          start = -1;
-          end = -1;
-          title = new StringBuilder();
+          state++;
         }
+      }
+      else if (state == 7) {
+        state = 1;
+
+        Chapter chapter = new Chapter();
+        chapter.title = title.map((t) -> t.toString());
+        chapter.start = timebase * start;
+        chapter.end = timebase * end;
+
+        chapters.add(chapter);
+
+        timebase = defaultTimebase;
+        start = -1;
+        end = -1;
+        title = Optional.empty();
+
+        if (line == null) {
+          break;
+        }
+        continue;
       }
 
       line = reader.readLine();
       lineNumber++;
+
+      if (line == null) {
+        if (state <= 1) {
+          // Haven't found a chapter yet or searching for next chapter,
+          // just finish and return current chapter list
+          break;
+        }
+        // state 5 and 7 can finish up a chapter, continue processing state 7 a last time
+        else if (state == 5 || state == 7) {
+          state = 7;
+        }
+        else {
+          throw new ParseException("Failed to parse FFMETADATA: Unexpected end of file", lineNumber);
+        }
+      }
     }
 
     return chapters;
--- a/modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
+++ b/modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
@@ -52,6 +52,8 @@ import org.opencastproject.videosegmenter.api.VideoSegmenterException;
 import org.opencastproject.videosegmenter.api.VideoSegmenterService;
 import org.opencastproject.workspace.api.Workspace;
 
+import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedService;
 import org.osgi.service.component.ComponentContext;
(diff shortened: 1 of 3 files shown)


modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -318,6 +346,39 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
       }
     }
 
+    if (properties.get(OPT_USE_CHAPTER_IF_AVAILABLE) != null) {
+      String value = (String) properties.get(OPT_USE_CHAPTER_IF_AVAILABLE);
+      try {
+        useChapterIfAvailable = Boolean.parseBoolean(value);
+        logger.info("Use Chapters if available is set to {}", useChapterIfAvailable);
+      } catch (Exception e) {
+        logger.warn("Found illegal value '{}' for videosegmenter's option: {}",
```
Arnei: I think it would be better to log an error here and throw a ConfigurationException.
I am aware the current code is in line with the rest of the `updated()` method, maybe there is a reason for why throwing a ConfigurationException is problematic somehow. But if it isn't, i'd throw it. Really let the user know about their typo.

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -318,6 +346,39 @@ public void updated(Dictionary<String, ?> properties) throws ConfigurationExcept
...
+        String[] values = value.split(",");
+
+        for (String mimeString : values) {
+          MimeType mt = MimeTypes.parseMimeType(mimeString);
+          mts.add(mt);
+        }
+
+        useChapterMimeTypes = mts;
+      } catch (Exception e) {
+        logger.warn("Found illegal value '{}' for videosegmenter's option: {}",
```
Arnei: See comment about ConfigurationException above.

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -379,252 +438,498 @@ protected Catalog segment(Job job, Track track)
...
+        "-nostats", "-nostdin",
+        "-i", mediaFile.getAbsolutePath(),
+        "-f", "FFMETADATA",
+        "-"
+    };
 
-      // optimization loop to get a segmentation with a number of segments close
-      // to the desired number of segments
-      while (!endOptimization) {
+    logger.info("Detecting chapters using command: {}", (Object) command);
```
Arnei:
```suggestion
    logger.debug("Detecting chapters using command: {}", (Object) command);
```

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -379,252 +438,498 @@ protected Catalog segment(Job job, Track track)
...
-        tmpSegments = new LinkedList<Segment>();
-        filterSegmentation(segments, track, tmpSegments, i);
-        float newError = OptimizationStep.calculateErrorAbs(tmpSegments.size(), prefNumberLocal);
-        if (newError < smallestError) {
-          smallestError = newError;
-          bestI = i;
+      // Being processing of title
+      else if (state == 5) {
+        if (!line.startsWith("title=")) {
+          throw new ParseException("Failed to parse FFMETADATA: CHAPTER title field missing", lineNumber);
```
Arnei: Is the title field required? I can't find anything about it in the documentation.

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -45,11 +45,15 @@
...
 import org.opencastproject.serviceregistry.api.ServiceRegistryException;
 import org.opencastproject.util.LoadUtil;
+import org.opencastproject.util.MimeType;
+import org.opencastproject.util.MimeTypes;
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.videosegmenter.api.VideoSegmenterException;
 import org.opencastproject.videosegmenter.api.VideoSegmenterService;
 import org.opencastproject.workspace.api.Workspace;
 
+import com.entwinemedia.fn.data.Opt;
```
Arnei: We would like to eventually get rid of the entwine classes. Please use `java.util.Optional` or something alike.

Arnei (changes requested):
I imagine a unit test for this would be really nice.

I wonder how this is expected to be used? Are there capture agents that include chapter information in their video files? Depending on who this is relevant to, the documentation should be amended.

Code works from what I tested, the comments below are minor stuff.

gregorydlogan:
Any progress on this?

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
Arnei: Couldn't find anything on whether the title field is required or not. I am willing to give David the benefit of the doubt here and just accept this.

Arnei (approved):
Fixed my review myself. Should be good to go now.

darolfes:
@Arnei Sorry, I had little time to come back to this.
I still have a patch that makes the title field optional (but parses it if it is present). I could push it if you want. However, it's still completely untested.

Arnei:
> @Arnei Sorry, I had little time to come back to this. I still have a patch that makes the title field optional (but parses it if it is present). I could push it if you want. However, it's still completely untested.

Oh hey, nice hearing from you. Didn't expect you to come back to this :D
Good to know that the title field should be optional after all. If you have code to that effect, please do contribute it. Shouldn't be that much work to test it.

darolfes:
Pushed it. Hopefully, it works. Looking it over, I am unsure of the `continue`s of the changed states. The force push adds one to state 7.
Edit: Wrong line, now it does.

Arnei:
Sooo, I tried to create a FFMETADATA file with a chapter without a title, looking like this:
```
;FFMETADATA1
major_brand=isom
minor_version=512
compatible_brands=isomiso2avc1mp41
encoder=Lavf58.76.100

[CHAPTER]
TIMEBASE=1/1000
START=1
END=6000
title=The Pledge

[CHAPTER]
TIMEBASE=1/1000
START=6001
END=12000

[CHAPTER]
TIMEBASE=1/1000
START=12001
END=20000
title= The Prestige
```

Trying to add this file to a video results in wrong chapter information:

```
;FFMETADATA1
major_brand=isom
minor_version=512
compatible_brands=isomiso2avc1mp41
encoder=Lavf58.76.100
[CHAPTER]
TIMEBASE=1/1000
START=0
END=12001
title=The Pledge
[CHAPTER]
TIMEBASE=1/1000
START=12001
END=20000
title= The Prestige
[CHAPTER]
TIMEBASE=1/10000000
START=120010000
END=647000000
title= The Prestige
```

So maybe the title field is required after all? :')

darolfes:
Interesting, I looked at my test files, and one file produces the following ffmetadata:

```
;FFMETADATA1
encoder=Lavf60.3.100
[CHAPTER]
TIMEBASE=1/1000000000
START=1000000
END=10000000000
[CHAPTER]
TIMEBASE=1/1000000000
START=10001000000
END=40000000000
title=Middle\
Test
[CHAPTER]
TIMEBASE=1/1000000000
START=40001000000
END=60000000000
title=End
```

The title field is missing. Maybe it depends on the chosen container format (this one is webm), or it only "works" for the first chapter, which would indicate a bug in FFmpeg (either by accepting/processing corrupt ffmetadata files or by incorrectly setting the chapter information in the video files when missing titles are present). VLC displays the first chapter as `Chapter 0`.

darolfes:
After reproducing the video file:
 Are there already chapters defined in your input file?
 This led to problems for me. A mapping with map_chapters is probably needed in this case (did not test this). With input files without chapters, I can reproduce the video file without a problem.

Arnei:
> Maybe it depends on the chosen container format (this one is webm)

Interesting. My original test was with a `mp4` container. Another test I did with `mp4` also failed. But with a `webm` it works (at least the title information is preserved correctly, webm seems to default to changing the timebase back to nanoseconds for whatever reason)

> Are there already chapters defined in your input file?

All files I tried to add chapter information to did not previously have chapter information.

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
```diff
@@ -379,252 +453,506 @@ protected Catalog segment(Job job, Track track)
...
+    // Nanoseconds are the default timebase
+    final double defaultTimebase = 1e-9f;
+    double timebase = defaultTimebase;
+    long start = -1;
+    long end = -1;
+    Optional<StringBuilder> title = Optional.empty();
+
+    String line = reader.readLine();
+    int lineNumber = 1;
+    while (null != line) {
```
Arnei: Since you mentioned being unsure about `continue`s, I checked the loop again. This can cause the last chapter in the file to not be added, if the last line in the FFMETADATA file belongs to the last chapter. The loop will parse the last line, read the next line which is null (as it does not exist), and exit the loop before it can get to the `state 7` part. Do you want to fix this or should I?

Arnei (changes requested):

modules/videosegmenter-ffmpeg/src/main/java/org/opencastproject/videosegmenter/ffmpeg/VideoSegmenterServiceImpl.java
darolfes: A similar problem exists with state 5 as it can now directly lead into the new state 7. Additionally a ParseException would probably be better instead of a NullPointerException for unexpected file ends.

darolfes: Will push an update.

darolfes: Hopefully this fixes the problem: [55163d..bd6b50](https://github.com/opencast/opencast/compare/55163d681e85628a8645d636adfd00dd3cd44321..bd6b50c24c2372e8c20f6d2234dd03fb32fdee84)

Arnei: This opens up the possibility of `line` being `null` during the loop, which can now cause NullPointerExceptions with e.g. `startsWith("")`

darolfes: This only affects the comments/empty line if-branch or did I overlook something else (as all states except 7 are never visited if line == null)?
[Fix](https://github.com/opencast/opencast/compare/bd6b50c24c2372e8c20f6d2234dd03fb32fdee84..6cd02b092cc1286f3b9d13ee9b2293f3f8f7a68b)

Arnei: Seems so, I just saw that it would be a problem in at least one case ^^.

Arnei (approved):

PR #4590, opened 2023-01-09
Add Tag-Engage Workflow Operation Handler

KatrinIhler:
This adds a new workflow operation handler with which you can update tags and flavors of media package elements directly in the engage publication. :tada: This means that for such changes, it's no longer necessary to pull the files from the publication into the workspace, saving lots of time. The operation works almost exactly the same as the tag WOH except that it looks at published elements and does not support `copy`. The publication is updated in the archive as well as the search service.

This could probably be generalized to work for other publications, but this is currently out of scope.

### DIFF
--- a/docs/guides/admin/docs/releasenotes/tag-engage-woh
+++ b/docs/guides/admin/docs/releasenotes/tag-engage-woh
@@ -1,0 +1,2 @@
+Added a tag-engage-WOH which can be used to update tags and flavors directly
+in the engage publication. This change was initially targeted at 12.x.
--- a/docs/guides/admin/docs/workflowoperationhandlers/index.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/index.md
@@ -21,98 +21,99 @@ definitions consist of a list of workflow operations, which basically map a piec
 
 The following table contains the workflow operations that are available in an out-of-the-box Opencast installation:
 
-|Operation Handler   |Description                                                    |Details|
-|--------------------|---------------------------------------------------------------|------------------------------------|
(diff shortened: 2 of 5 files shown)


mtneug:
This should go into 13.

KatrinIhler:
> This should go into 13.

Any particular reason for this? There's not a rule against features going into a legacy release, is there? One could argue for develop, maybe, but I think it's unlikely this will affect anything else.

I would rather this go into 12 because this will mean less custom code for us sooner. ;)

mtneug:
mmh in my head, I always thought that legacy releases should not get new features. But looking at the mailing list's last messages on that topic, I don't see that anymore. BTW our dev docs still say that features should only go into `develop` 🙈.

Regarding the operation: I'm not a fan of highly specialized operations, but I think I have kind of given up on that.

gregorydlogan:
> mmh in my head, I always thought that legacy releases should not get new features.

Personally, I would agree with that approach - legacy doesn't see anywhere near as much testing. That being said, this is a WOH, and therefore should (at least in theory) only affect those who choose to use it. This code is also backed by a large, stable consultancy who will be stuck maintaining this regardless. I would accept it in this case, maintaining an approach of not merging features into legacy otherwise. 

> BTW our dev docs still say that features should only go into `develop` 🙈.

Whoops, need to fix that.

KatrinIhler:
> Regarding the operation: I'm not a fan of highly specialized operations, but I think I have kind of given up on that.

I agree with you there, I think it would be better if this operation could handle arbitrary publications.

But since I wrote this in one day to quickly fix thousands of published videos with the wrong tag, I'm kind of happy with it anyway. ;D

Maybe in the future...

gregorydlogan (approved):

PR #4609, opened 2023-01-17
Fix documentation syntax error

gregorydlogan:
This PR:
- Fixes the syntax error
- Adds a check to ensure that the `opencast_major_version` macro in the documentation has appropriate spacing around it

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/checkstyle/check-docs.sh
+++ b/docs/checkstyle/check-docs.sh
@@ -7,6 +7,13 @@ cd docs/guides
 
 for docs in admin developer; do
   cd $docs
+  if test "$( grep -R opencast_major_version | grep '\w}\|{\w' )" != ""; then
+    echo "Error, $docs has a syntax error related to opencast_major_version:"
+    grep -R opencast_major_version | grep '\w}\|{\w'
+    ret=1
+    continue
+  fi
+
   echo "Building $docs documentation…"
   if mkdocs build &> mkdocs.log; then
     if grep \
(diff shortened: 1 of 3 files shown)


gregorydlogan:
Discussion question: Should the check be backported to `r/12.x` or `r/13.x`?

docs/checkstyle/check-docs.sh
```diff
@@ -7,6 +7,13 @@ cd docs/guides
 
 for docs in admin developer; do
   cd $docs
+  if test "$( grep -R opencast_major_version | grep '\w}\|{\w' )" != ""; then
```
lkiesow: Not that important, but maybe replace `-R` with `-r` to not follow symlinks?

Also, this would fail for something like:

```
{{ opencast_major_version }} some text `{xy}`
```

Maybe check for `{opencast_major_version` instead?

lkiesow: Not that important, but maybe replace `-R` with `-r` to not follow symlinks?

Also, this would fail for something like:

```
{{ opencast_major_version }} some text `{xy}`
```

Maybe check for `{opencast_major_version` instead?

gregorydlogan: Good point!  Fixed.  Also added checks for `opencast_major_version()}`, which would have been missed previously.

lkiesow (approved):

Issue #4407, opened 2022-11-01
How subtitles should work in Opencast


lkiesow:
As a more interactive version, I've also added this to a HackMD where you can add comments on certain lines. Use whatever you prefer to add comments:

- [@lkiesow/oc-subtitles](https://hackmd.io/@lkiesow/oc-subtitles)

---

We have a lot of implicit and often slightly different assumptions of how we handle subtitles. In here, I'll try to document a best-practice approach. After some discussion, we can maybe add this to the Opencast documentation and make it the official approach.

I'll try to suggest common ways of storing and identifying the subtitles from upload, through processing and finally for showing them in the player.

## TL;DR

Let's make subtitles a first-class citizen in Opencast, storing them as tracks alongside audio and video streams and letting workflows handle them by default.

Let us use flavors of the form `captions/<processing>`, e.g. `captions/source` and `captions/delivery`. This makes it easy to identify subtitles and also easy to publish everything ready for publication.

Let us specify tags holding additional information of the form `lang:<lang>`, `generator:<type>[:<id>]` and `type:<caption-type>` to each subtitle track.

## History

While there are many components in the history of Opencast dealing with subtitles (Matterhorn admin interface, Engage player, …), the most relevant ones are Opencast's HTML5 players: Theodul and Paella.

> Since it is still beta, I leave Paella Player 7 out of the historical consideration.

### Theodul

Theodul supports WebVTT files and considers both tracks and attachments when loading subtitles. It only supports a single subtitle stream. If both exist, it prefers tracks over attachments. The only selection criteria is, that the media package element must have the mime type `text/vtt`.

Examples:

- type `track`, flavor `captions/delivery`, no tags, mime type `text/vtt`
    - will be selected for the player
- type `track`, flavor `captions/vtt`, tag `lang:en`, mime type `text/plain`
    - will __not__ be selected

See `loadAndAppendCaptions(…)` in `engage-theodul-plugin-video-videojs/src/main/resources/static/main.js` for more details.

### Paella

Paella supports dxfp, WebVTT and SubRip files (not entirely sure). It supports loading subtitles from attachments, catalogs and (since ≥ 12.x) tracks. It supports multiple subtitle streams. Subtitles selection happens by selecting all media package elements (only attachments in Opencast ≤ 11) with main flavor `caption`.

[...]

PR #4622, opened 2023-01-25
Treat captions as tracks in standard workflows

Arnei:

Instead of treating captions as something special, this commit attempts to change the workflows to treat them as tracks, subjecting them to the same operations.
Actually has to skip some operations in partial-publish that will cause an exception if given something that is not a media file, but that should be fine since those operations are irrelevant to subtitles.


KatrinIhler:
Changes look reasonable. Will report once I have tested this a bit.

KatrinIhler:
A couple of problems I noticed during testing:

1. The explicit snapshotting of captions/* should also be removed from publish.xml  and partial-error.xml.
2. Publication of captions doesn't work for a republish because captions/prepared isn't archived the first time because it is changed (without copy) to presenter/work, and will not created during the second run because presenter and/or presentation/prepared _was_ archived. 
For this to work properly, the easiest way is to use the copy option when changing captions/prepared to captions/work (although the whole */prepared stuff isn't terribly useful right now).

KatrinIhler (changes requested):
A couple of problems I noticed during testing:

1. The explicit snapshotting of captions/* should also be removed from publish.xml and partial-error.xml.
2.  Publication of captions doesn't work for a republish because captions/prepared isn't archived the first time because it is changed (without copy) to presenter/work, and will not created during the second run because presenter and/or presentation/prepared was archived.
    For this to work properly, the easiest way is to use the copy option when changing captions/prepared to captions/work (although the whole */prepared stuff isn't terribly useful right now).

mliradelc:
@KatrinIhler I don't have strong opinions with the standard workflows (We use custom ones) Are the latest changes OK with you?

etc/workflows/fast.yaml
```diff
@@ -44,6 +44,16 @@ operations:
...
 
+  - id: tag
+    fail-on-error: true
+    exception-handler-workflow: "partial-error"
+    description: "Tag captions for publication"
+    configurations:
+      - source-flavor: "captions/source"
+      - target-flavor: "captions/preview"
+      - target-tags: "+engage-download"
+      - copy: true
```
mliradelc, line 55: Should you add?
`generator:unknown`
`generator-type:manual`
`lang:unknown`

mliradelc (approved):
For me is ok, only one question regarding the tags.

etc/workflows/fast.yaml
Arnei: It is not necessary per se, tags are not variables that need to be initalized or anything like that. 

I don't think the workflows generally should make general assumptions about subtitle tags (was the caption written manually or automatically, the workflow does not know). Although since this is a workflow used primarily for testing, maybe the tags are nice for "demonstrating" how tags for captions are used in Opencast?

Bottom line, for the fast workflow in particular I don't really care. If you think it's a good idea, let's do it.

mliradelc: Fair enough. I'm only thinking in your changes in Paella, that can detect those tags for the subtitle selection.

PR #4623, opened 2023-01-25
Add captions as standard upload option in Admin UI

Arnei:

This changes subtitles in the asset upload from attachments to tracks and adds a subtitle upload to the source tab in the Admin UIs "Create Event" dialog.

Furthermore, adds tag support to the asset upload. One can now add multiple tags to their asset upload configuration.

Known issues:
- Adding multiple options for uploading subtitles (for e.g. multiple languages) does not currently work. This is due to tracks in the same flavor being overwritten.


etc/listproviders/event.upload.asset.options.properties
```diff
@@ -31,16 +31,16 @@ list.name=eventUploadAssetOptions
...
+# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.SUBTITLES={"id": "track_subtitles_option", "type":"track", "flavorType": "captions","flavorSubType": "source", "tags": "generator:unknown", "displayOrder":3, "accept": ".vtt"}
 EVENTS.EVENTS.NEW.UPLOAD_ASSET.WORKFLOWDEFID=publish-uploaded-assets
 
 # The video source track upload options are only for new events.
+EVENTS.EVENTS.NEW.SOURCE.UPLOAD.NON_SEGMENTABLE={"id": "track_presenter","type":"track", "flavorType": "presenter","flavorSubType": "source", "multiple":false, "displayOrder":1, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv,.flac,.m4a,.mp3,.ogg,.wav,.wma,.f4v"}
+EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SEGMENTABLE={"id": "track_presentation","type":"track", "flavorType": "presentation","flavorSubType": "source", "multiple":false, "displayOrder":2, "accept": ".avi,.flv,.m4v,.mkv,.mov,.mp4,.mpeg,.mpg,.ogv,.webm,.wmv,.flac,.m4a,.mp3,.ogg,.wav,.wma,.f4v"}
+EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES={"id": "track_subtitles", "type":"track", "flavorType": "captions", "flavorSubType": "source", "tags": "generator:false", "displayOrder":3, "accept": ".vtt"}
```
mtneug: Suggestion: use `NEW.UPLOAD_ASSET` instead of `NEW.SOURCE` for subtitles. This would allow you to upload files afterward + during the new event dialog. Also stops new events that only contain subtitles.

Arnei: Wanted to reply to this earlier and then forgot. But now, finally, it is time.

I got two reasons for using `NEW.SOURCE` instead of `NEW.UPLOAD_ASSET`. The first is that the admin ui does not display assets of type `track` in the `NEW.UPLOAD_ASSET` tab of the new event dialog. Relevant code here: https://github.com/opencast/opencast/blob/a84dd5f0fac4f9883c9c621c7b82e07c11f96a51/modules/admin-ui-frontend/app/scripts/shared/services/wizards/new-event/uploadAssetNewEvent.js#L112. Maybe that is a design decision we want to rethink, maybe not, I did not want to open that can of worms.

The second reason is that grouping subtitle upload together with presenter and presentation is in my opinion less likely to annoy users who will not upload subtitles with every new event. Since I am proposing these settings as the new default, using `NEW.UPLOAD_ASSET` would require users to click through an additional tab everytime the want to create a new event, a tab which is useless to them if they do not wish to upload subtitles. I wanted to avoid this dent in the user experience (well, or they could just remove subtitles form this file, but UX is all about sensible defaults, right?).

As for new events that only contain subtitles, I am hopefully that that is a minor issue at best. Adding a validator that checks if the source uploads contain at least one media file would hopefully be fairly easy to achieve.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -44,7 +44,7 @@
      "VIDEOFILE": "Videofile",
      "VIDEOFILE_DESCRIPTION": "Video file for upload",
      "CAPTION": "Captions",
-     "CAPTION_DESCRIPTION": "Caption file; VTT or dfxp format is supported",
+     "CAPTION_DESCRIPTION": "Caption file in VTT format",
```
KatrinIhler, line 47: I think changing the content of an existing translation will not result in Crowdin prompting to re-translate this into other languages - we might want to change the key here?

etc/listproviders/event.upload.asset.options.properties
KatrinIhler: I agree with the UX comment from Arne, I would definitely prefer this view over a new tab. If I understood you correctly, you could activate NEW.UPLOAD_ASSET additionally to also allow adding subtitles to an existing event?

KatrinIhler:
Tag support seems like a great idea. Changes look reasonable.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
Arnei: Not sure about prompting, but afaik it at least marks the string as requiring a new translation. And if it doesn't, it really should. Changing the translation key each time you change the translation seems ... not great.

KatrinIhler: Okay, then I take that back.

KatrinIhler:
The new tags option should probably be listed at the top of `event.upload.asset.options.properties` with the rest of the config keys. Also as far as I can tell it's not documented anywhere that the tags are comma-separated without whitespaces in between? That is something I got wrong at first, which will lead to the language tag not being recognized in the player. (And personally, I like ", " better, but maybe that's a matter of taste.)

etc/listproviders/event.upload.asset.options.properties
```diff
@@ -31,16 +33,16 @@ list.name=eventUploadAssetOptions
...
-# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.CAPTIONS_DFXP={"id":"attachment_captions_dfxp", "type": "attachment", "flavorType": "captions", "flavorSubType": "dfxp", "displayOrder":2, "accept": ".ttml, .dfxp, .xml"}
-# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.CAPTIONS_WEBVTT={"id":"attachment_captions_webvtt", "type": "attachment", "flavorType": "text", "flavorSubType": "vtt", "displayOrder":3, "accept": ".vtt"}
-# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.PREVIEW_IMAGE={"id":"attachment_preview_image", "type":"attachment", "flavorType": "presenter","flavorSubType": "search+preview", "displayOrder":4, "accept": ".bmp,.gif,.jpeg,.jpg,.png,.tif,.tiff"}
+# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.PREVIEW_IMAGE={"id":"attachment_preview_image", "type":"attachment", "flavorType": "presenter","flavorSubType": "search+preview", "displayOrder":2, "accept": ".bmp,.gif,.jpeg,.jpg,.png,.tif,.tiff"}
+# EVENTS.EVENTS.NEW.UPLOAD_ASSET.OPTION.SUBTITLES={"id": "track_subtitles_option", "type":"track", "flavorType": "captions","flavorSubType": "source", "tags": "generator:unknown", "displayOrder":3, "accept": ".vtt"}
```
KatrinIhler, line 37: Another question: Why is the tag here `generator: unknown`, but for source uploads it's `generator: false`? Technically the subtitles could have been automatically generated in both cases, so I think we should pick one and stick with it (I'm fine with either).

Arnei: Agreed, that's just me being inconsistent. Thanks for the catch.

mliradelc (commented):
Works as expected. 

I have a question, there is a way to add more list providers `event.upload.asset.options.properties` without modifying the language file in the admin UI module?

Arnei:
Iirc you should be able to add more entries for subtitles to `etc/listproviders/event.upload.asset.options.properties`, but the labels displayed in the admin ui won't be properly translated if you don't also modify the translations files.

mliradelc:
Unfortunately, it didn't work for me. Try adding a custom upload field.

Arnei:
Did you add `multiple:true`? That's required for uploading multiple files to the same flavor.  
So you should be able to do something like
```
EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES={"id": "track_subtitles", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:one", "displayOrder":3, "accept": ".vtt"}
EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES_TWO={"id": "track_subtitles_two", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:two", "displayOrder":4, "accept": ".vtt"}
```

mliradelc:
> Did you add `multiple:true`? That's required for uploading multiple files to the same flavor. So you should be able to do something like
> 
> ```
> EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES={"id": "track_subtitles", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:one", "displayOrder":3, "accept": ".vtt"}
> EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES_TWO={"id": "track_subtitles_two", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:two", "displayOrder":4, "accept": ".vtt"}
> ```

Do you mean upload several files in the same field, or upload a single file in different fields at the same time?

Arnei:
> > Did you add `multiple:true`? That's required for uploading multiple files to the same flavor. So you should be able to do something like
> > ```
> > EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES={"id": "track_subtitles", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:one", "displayOrder":3, "accept": ".vtt"}
> > EVENTS.EVENTS.NEW.SOURCE.UPLOAD.SUBTITLES_TWO={"id": "track_subtitles_two", "type":"track", "flavorType": "captions", "flavorSubType": "source", "multiple":true, "tags": "generator:two", "displayOrder":4, "accept": ".vtt"}
> > ```
> 
> Do you mean upload several files in the same field, or upload a single file in different fields at the same time?

The latter. The config I posted should display two fields in the Create Event dialog, allowing uploading one file per field.

mliradelc:
I see. Have you ever tried the `title` option? Like is shown in the docs:

> - Modify etc/listproviders/event.upload.asset.options.properties
>- For the title of the options displayed in the admin interface, either:
>   - add a translation for the new asset name to modules/admin-ui-> frontend/resources/public/org/opencastproject/adminui/languages/...
>   - **or add a title field to the upload specification**
>- Add a workflow to publish the uploaded assets to etc/workflows
>Test your changes

https://docs.opencast.org/r/13.x/admin/#configuration/admin-ui/asset-upload/

Arnei:
The one you marked bold? That kinda works, but you can't set `SHORT` and `DETAIL` specifically (at least I don't know how to). So it displays something like "Your title here.SHORT".

mliradelc (approved):
Works everything.

There is one thing that needs to be addressed in another PR. If you want to add other fields it's imposible if you don't compile  opencast again (Or at least the admin-ui module).

The "title" option won't work as described in the docs.

PR #4624, opened 2023-01-25
Update tag support for subtitles as tracks in Paella Player

Arnei:

Since the Paella Player already supports subtitles as tracks, this merely adds some support for tags that will be used with subtitles in the future.

This also drops anything dfxp related, support for subtitles as catalogs, and some unused
translation strings.


mliradelc:
@miesgre have you some time to look into this PR?

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -347,37 +361,6 @@ function getCaptions(episode) {
   // Read the tracks
   readCaptions(tracks, captions);
 
-  // Read the catalogs
```
miesgre, line 357: There are some institutions using subtitles from catalogs. See (https://github.com/opencast/opencast/pull/5049#issuecomment-1599126063).

I believe that we should not remove this support until a way to transfer these subtitles to tracks is provided. Or at least add a note that this feature will be removed in future versions to give the institutions time to upgrade.

Perhaps we could add a configuration variable in the config.json: `deprecatedCaptionsAsCatalog: false`. By default it could be set to false and if any institution needs it, they can activate it and realize that it is a deprecated feature and in future versions it will be removed.

@karendolan?

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/03_oc_search_converter.js
```diff
@@ -437,38 +448,6 @@ class OpencastToPaellaConverter {
     // Read the tracks
     this.readCaptions(tracks, captions);
 
-    // Read the catalogs
```
miesgre, line 440: The same comment here that in paella 7

miesgre (commented):
I think that subtitle support from captions should not be removed yet.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: For release notes and migrations notes see this PR: https://github.com/opencast/opencast/pull/4844 I've split the subtitle-rework into multiple PRs to make them easier to work on and hopefully also easier to review.

If enough people are in favor of it, I don't mind keeping the existing "captions as catalog" functionality and marking it as deprecated.

miesgre: @Arnei. Is it possible to use the `changetype` operation to change the type of subtitles from catalogs to tracks (from PR: #4844)?

Arnei: That's the intention, yes.

miesgre: Release notes from PR #4844 are ok for me. 
If nobody complains about removing the subtitles from the catalogs in this release, it's fine with me (there is an operation so that the type can be changed).

The support of subtitles in the catalogs has been removed, but the support of subtitles in the attachments is kept. Why is it not removed as well?

https://github.com/opencast/opencast/blob/0de9f6118f55390169fbdfa7b5d62ef41aa9bcb4/modules/engage-paella-player-7/src/js/EpisodeConversor.js#L358-L362

modules/engage-paella-player/src/main/paella-opencast/plugins/es.upv.paella.opencast.loader/localization/da-DK.json
```diff
@@ -1,8 +1,4 @@
 {
 	"Error loading video! No video tracks found": "Videoindlæsningsfejl! Igen videospor fundet",
-	"Error loading video {0}": "Videoindlæsningsfejl {0}",
```
KatrinIhler, line 3: Are these translations not managed by Crowdin? Otherwise you would need to change only en_US.

Arnei: Before you asked that I was sure that these translations were not managed by crowdin, but you made me check and they indeed are :o

modules/engage-paella-player-7/src/js/EpisodeConversor.js
Arnei: That is a good question. `readCaptions(attachments, captions); ` was probably left in also in the name of backwards compatibility? I'd be fine with removing that too.

KatrinIhler:
Changes look reasonable at first glance.

KatrinIhler:
I have a question about how the language tags work:

If I understand this correctly, I have to add the tags to the config for the asset upload if I want manually uploaded captions to get the tag? So I have to make an assumption about what kind of language captions will be uploaded? Or is there a way to get this information from the file? (I don't even know if vtt supports this kind of metadata tbh.) 

And without the language tag, the subtitle option will appear as "undefined" in the player? Can we maybe get a better default for this, like "Unknown language" or something?

Arnei:
> If I understand this correctly, I have to add the tags to the config for the asset upload if I want manually uploaded captions to get the tag? So I have to make an assumption about what kind of language captions will be uploaded? Or is there a way to get this information from the file? (I don't even know if vtt supports this kind of metadata tbh.)

To my knowledge VTT does not support that, so Opencast unfortunately cannot just derive the language from the subtitles file.

Correct, one way would be to make assumptions in the asset upload config file about what languages your subtitle can be in. If you are certain they will always be in e.g. English, French and Spanish, that might be the simplest solution for you.

Otherwise you will probably have to resolve language selection in your upload workflow(s), using workflow configuration and the `tag` operation. Maybe `tag-by-DCTerm` is also an option, but I have no experience with that one.

mliradelc:
The standard while allows some metadata as a JSON objects, doesn't have a proper header section. 

You can add a JSON metadata at the beginning of the subtitle, but in our case, we don't need to. We can add that metadata to our media package.

mliradelc (approved):
I approve the Changes, and deprecating the captions in catalogues with a PR to migrate the old media to tracks seems reasonable.

I will wait what is the outcome of #4844 before to merge it.

PR #4626, opened 2023-01-25
Refit editor endpoint for subtitles as tracks

Arnei:

*This PR will break the all subtitle functionality in the frontend, and thus should optimally be merged together with opencast/opencast-editor#958.*

Does away with subtitle flavors, since track identifiers are now used
for matching subtitles between back- and frontend.

Adds tags for display and potential language matching purposes in the
frontend.


mliradelc:
Reminder for myself: Need to implement the editor PR in my test environment

PR #4627, opened 2023-01-25
Add documentation on subtitles

Arnei:

Adds a new documentation file about subtitles.


docs/guides/admin/docs/modules/subtitles.md
```diff
@@ -0,0 +1,79 @@
...
+be stored in the same flavor, while being optionally differentiated by [tags](#tags).
+
+### Tags
+
+Tags are used to hold additional information about individual subtitle tracks. All tags are optional and components
+should work without them being present, falling back to generic displays, or not showing information at all.
+
+The following is a list of tags that Opencast components may recognize:
+- `lang:<language>`
+	- `<language>` represents a language code of the form TODO: WHICH LANGUAGE CODE?
```
KatrinIhler: I assume these Todos should be done before we merge this?

Arnei: Thanks for the reminder.

Regarding language codes: I don't have a solid opinion on which to use. Do you? I should probably bring this is up in the Technical Meeting.

Regarding no major language: Might just have to ignore that for now and cross that bridge when we come to it?

KatrinIhler: > Regarding language codes: I don't have a solid opinion on which to use. Do you?

I think I prefer including the region in the language code, so `en-GB` instead of `en`. Most people commenting on the issue also seemed to think so.

> I should probably bring this is up in the Technical Meeting.

Good idea.

> Regarding no major language: Might just have to ignore that for now and cross that bridge when we come to it?

Agreed.

docs/guides/admin/docs/modules/subtitles.md
```diff
@@ -0,0 +1,81 @@
+Subtitles
+=========
+
+This document records guidelines and conventions on how subtitles should be handled in Opencast.
+
+<div class=warn>
+These guidelines are new to Opencast 14. Not all parts of Opencast may yet adhere to them!
```
mliradelc:
```suggestion
These guidelines are new to Opencast 15. Not all parts of Opencast may yet adhere to them! 
```

mliradelc (approved):
Docs are ok and very easy to undestand. Only a little update regarding opencast version

docs/guides/admin/docs/modules/subtitles.md
mliradelc: Or maybe rewording to avoid update every time OC upgrades.

Arnei: Thanks for catching that again :see_no_evil:  
Maybe it's actually good having to look at this every major OC upgrade, so we're forced the re-evaluate if this warning is still appropriate?

mliradelc: I don't know, It's straightforward easy to forget to update the tiny details in the docs. I just found in the dev docs some repos to Bitbucket that are from almost 5 years deprecated.

Arnei: Well, in that case the version number at least serves as an indicator that this part of the docs is likely outdated, and as a user you should proceed with care :D

mliradelc: Fair Enough :)

PR #4651, opened 2023-01-26
Add studio series dropdown selection

geichelberger:
This PR adds a dedicated studio API and changes the studio UI path from `/studio` to `/studio-ui`, consistent with the editor UI and API path.

Related Studio PR https://github.com/elan-ev/opencast-studio/pull/992

This PR depends on merging the related studio PR and releasing a new version.

* [ ] have a concise title
* [ ] pass automated tests


lkiesow:
> This PR adds a dedicated studio API and changes the studio UI path from `/studio` to `/studio-ui`, consistent with the editor UI and API path.

Is it really necessary to do that? That will break a bunch of integrations.

I haven't actually looked at any of these changes, but I'm wondering why not just getting the series information from the search service or the external API?

geichelberger:
Using the search service will only work if the series is already published, and until now, the external API is not used by any frontend, and the user needs a role_api* to access the external API.

lkiesow:
Can we then make the API available at `/studio-api`? Changing the path will break the integrations in Tobira, Stud.IP, Moodle, ILIAS and likely more and is really unnecessary.

geichelberger:
Theoretically, we should be consistent with naming and follow specific conventions, even if that means we have a breaking change. But I have changed my mind about how we should proceed.
During the migration to Karaf 4.4, I noticed several areas for improvement in our JAX-RS implementation. Therefore I suggest combining the APIs under two standard servlets anyway. The external API can be based on /api, and the others on /rest.

Therefore, /rest/studio is a possible choice for the studio API.

I will propose my Idea on the dev list.

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,111 @@
...
+    abstractText = "This service creates, edits and retrieves and helps managing series.", notes = {
+    "All paths above are relative to the REST endpoint base (something like http://your.server/files)",
+    "If the service is down or not working it will return a status 503, this means the the "
+        + "underlying service is not working and is either restarting or has failed",
+    "A status code 500 means a general failure has occurred which is not recoverable and was "
+        + "not anticipated. In other words, there is a bug! You should file an error report "
+        + "with your server logs from the time when the error occurred: "
+        + "<a href=\"https://github.com/opencast/opencast/issues\">Opencast Issue Tracker</a>" })
+@Component(immediate = true, service = StudioEndpoint.class, property = { "service.description=Studio REST Endpoint",
+    "opencast.service.type=org.opencastproject.studio", "opencast.service.path=/studio" })
```
LukasKalbertodt: This should be formatted differently. For excample, check [this file](https://github.com/opencast/opencast/blob/29a58866e8523f0ba69b68d54c61f82eb4aedb27/modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java#L56). Also see [this issue](https://github.com/opencast/opencast/issues/2238) for information on the "new style" that Opencast code should follow. Unfortunately, I haven't made much progress in reformatting the remaining modules recently :/

But yeah, from the file above and the linked PRs, it should be clear how these annotations should be formatted.

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Path("/")
+@RestService(name = "seriesservice", title = "Series Service",
+    abstractText = "This service creates, edits and retrieves and helps managing series.", notes = {
```
LukasKalbertodt: > This service creates, edits and retrieves and helps managing series.

Copy & paste bug

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+  private SecurityService securityService;
+
+  {
+    gson = new GsonBuilder().serializeNulls().create();
+  }
+
+  @GET
+  @Path("/series.json")
+  @Produces(MediaType.APPLICATION_JSON)
+  @RestQuery(name = "getSeries", description = "Returns series where the current user has write permissions.",
```
LukasKalbertodt:
```suggestion
  @RestQuery(name = "getSeries", description = "Returns all series for which the current user has write permissions.",
```

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+      returnDescription = "A list of JSON series objects",
+      restParameters = {
+         @RestParameter(name = "filter", isRequired = false,
+          description = "Usage <Filter Name>:<Value to Filter With>. Filters can combine using a comma \",\"."
+            + " Available Filters: title, textFilter.", type = STRING),
+      },
+      responses = {
+      @RestResponse(description = "Returns a list of series.",
+          responseCode = HttpServletResponse.SC_OK)
+  })
```
LukasKalbertodt: As above, this should be formatted differently. Something like this:

```java
@RestQuery(
    name = "getSeries", 
    description = "Returns series where the current user has write permissions.",
    returnDescription = "A list of JSON series objects",
    restParameters = {
        @RestParameter(
            name = "filter", 
            isRequired = false,
            description = "Usage <Filter Name>:<Value to Filter With>. Filters can combine using"
                    + " a comma \",\". Available Filters: title, textFilter.", 
            type = STRING
        ),
    },
    responses = {
        @RestResponse(
            description = "Returns a list of series.",
            responseCode = HttpServletResponse.SC_OK
        ),
    }
)
```

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+  public Response getSeries(@QueryParam("filter") String filter) {
+    SeriesSearchQuery query = new SeriesSearchQuery(securityService.getOrganization().getId(),
+        securityService.getUser());
```
LukasKalbertodt, line 120: You said

> Using the search service will only work if the series is already published

So this also uses the elasticsearch, but is returning series for which episodes have been published?

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+    query.withoutActions();
+    query.withAction(Permissions.Action.WRITE);
+
+    if (filter != null && !filter.isBlank()) {
+      for (String f : filter.split(",")) {
+        String[] filterTuple = f.split(":",2);
```
LukasKalbertodt:
```suggestion
        String[] filterTuple = f.split(":", 2);
```

Weird, the style checker should catch these things :thinking:

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+        if (filterTuple.length < 2) {
+          logger.info("Filter {} not valid: {}", filterTuple[0], filter);
+          continue;
```
LukasKalbertodt: Why not return BAD REQUEST here?

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+        String name = filterTuple[0];
+        String value = filterTuple[1];
+
+        if ("title".equals(name)) {
+          query.withTitle(value);
+        } else if ("textFilter".equals(name)) {
+          query.withText("*" + value + "*");
+        } else {
+          logger.warn("Unknown filter criteria {}", name);
+          return Response.status(SC_BAD_REQUEST).build();
```
LukasKalbertodt: I find it rather useful to also send the error message as body in the 400 response. That's way easier for people working with the API to understand what the error is. E.g. check [this](https://github.com/opencast/opencast/blob/29a58866e8523f0ba69b68d54c61f82eb4aedb27/modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java#L203-L206)

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+      }
+    }
+
+    try {
+      return Response.ok(gson.toJson(Arrays.stream(elasticsearchIndex.getByQuery(query).getItems())
+              .map(SearchResultItem::getSource)
+              .map(SeriesAssembler::toDto)
+              .collect(Collectors.toList())
+          )).build();
```
LukasKalbertodt: Minor nit, but I would probably separate the actual searching, the gson serialization and the response building  to visually separate them from one another:

```suggestion
      var items = Arrays.stream(elasticsearchIndex.getByQuery(query).getItems())
              .map(SearchResultItem::getSource)
              .map(SeriesAssembler::toDto)
              .collect(Collectors.toList());
      var json = gson.toJson(items);
      return Response.ok(json).build();
```

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/assembler/SeriesAssembler.java
```diff
@@ -0,0 +1,37 @@
...
+
+public final class SeriesAssembler {
+
+  private SeriesAssembler() {
+
+  }
+
+  public static SeriesDto toDto(Series series) {
+    return new SeriesDto(series.getIdentifier(), series.getTitle());
+  }
```
LukasKalbertodt: Does this really need to live in its own class? It seems like it would just be a short line when inserted into the `StudioEndpoint.java`. Or alternatively, the constructor of `SeriesDto` could just be this function, i.e. take one `Series` argument.

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/dto/SeriesDto.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public String getId() {
+    return id;
+  }
+
+  public String getTitle() {
+    return title;
+  }
+
+}
```
LukasKalbertodt: It would be great, if `SeriesDto` wouldn't need to live in its own file and if it didn't need a constructor and getters. Unfortunately, public fields are currently rejected by the style checker, but this is something I wanted to get changed for a long time already. Maybe I'll finally get to bringing this up in the community.

LukasKalbertodt (changes requested):
Thanks for tacking this feature. I know many people are interested in this. 

I agree with Lars regarding the path: changing `/studio` to `/studio-ui` would lead to years of broken Opencast-integrations. This advantage of having some consistency in this regard is outweighed very much by the breakage it will cause. But you already suggested keeping `/studio` and instead using a different path for this new Studio API.

~~Regarding adding a new API: while I trust you that the existing APIs might not be sufficient yet, I very much dislike adding yet another ad-hoc API. But this is not your fault, but rather the fault of the general Opencast API situation. Honestly, I have not really a better idea. Maybe the upcoming (?) LMS GraphQL API would solve something, but no idea if/when that will land. Was this discussed in the technical meeting? If not, maybe we again should bring this up and see what others think.~~ See [this comment](https://github.com/opencast/opencast/pull/4651#issuecomment-1559747934)

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/dto/SeriesDto.java
geichelberger: Java has the record keyword for that since Java 14

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
geichelberger: The search service is based on Solr.

geichelberger:
I am implementing a GraphQL API for Opencast for a project based on Opencast; the current state is a pre-alpha.

modules/studio-service/src/main/java/org/opencastproject/studio/endpoint/StudioEndpoint.java
LukasKalbertodt: Oh woops, somehow I was already living in the future when solr had been removed :P

LukasKalbertodt:
From the technical meeting just now:
- It would of course be better to use an existing API. But the search API and the external API indeed both have problems for usage in Studio right now. So work would be required to use them, if it's possible at all.
- Of course, we all want a new general, perfect API, but we don't have it yet. 
- No one had any objections to just merging this tiny new API as we can easily remove it in the future. 

So, let's go  ahead with this approach.

LukasKalbertodt (changes requested):
Most of my previous comments were resolved. If someone could go ahead and mark the review comment (which I already hid) as "resolved", that would be nice. Only thing that remains is a tiny style nit and the comment about sending the error message as body. But neither is really important!

Two things I noticed while testing though:

- I think this PR should already add an entry into the security config. Something like `<sec:intercept-url pattern="/studio-api/**" access="ROLE_ADMIN, ROLE_STUDIO" />`? I think it makes sense to give `ROLE_STUDIO` access to that route, no? I don't see any problems with that.
- Series to which the user has write access via `ROLE_ANONYMOUS` are not returned by the API. It's a weird situation though, I agree. And maybe it's even better to NOT list those series in Studio? Just wanted to point this out, I have no opinion on whether to keep that behavior or change it.

gregorydlogan:
> * I think this PR should already add an entry into the security config. Something like `<sec:intercept-url pattern="/studio-api/**" access="ROLE_ADMIN, ROLE_STUDIO" />`? I think it makes sense to give `ROLE_STUDIO` access to that route, no? I don't see any problems with that.

+1

> * Series to which the user has write access via `ROLE_ANONYMOUS` are not returned by the API. It's a weird situation though, I agree. And maybe it's even better to NOT list those series in Studio? Just wanted to point this out, I have no opinion on whether to keep that behavior or change it.

Sorry, wat?  Write access via `ROLE_ANONYMOUS`?  So anyone can write to this series?

LukasKalbertodt:
> Sorry, wat? Write access via `ROLE_ANONYMOUS`? So anyone can write to this series?

Yeah, series and events with `ROLE_ANONYMOUS` write permission exist in the wild. Don't ask me how or why, but they do exist. I also can't think of a reason why it would be a good idea to do that. But: they exist so I brought up that this PR does not list these kinds of series. Not sure if that's an oversight or intentional.

gregorydlogan:
@geichelberger If you can get that security change made we cna probably get this into the next `r/14.x`

geichelberger:
I have already prepared the changes, and I will push them today.

geichelberger:
It doesn't make sense to merge it into 14.x as the studio changes won't be part of 14.

etc/security/mh_default_org.xml
```diff
@@ -296,14 +296,15 @@
     <sec:intercept-url pattern="/static/**" method="GET" access="ROLE_ANONYMOUS" />
     <sec:intercept-url pattern="/export/**" method="GET" access="ROLE_ANONYMOUS" />
 
-    <!-- Enable access to Opencast Studio -->
+    <!-- Enable access to Opencast Studiqo -->
```
LukasKalbertodt: Typo, but don't want to block this PR only due to this :D

LukasKalbertodt (approved):
All my comments were resolved!

PR #4658, opened 2023-01-30
 Editor service mediapackage locking

JamesUoM:
To prevent multiple users from editing the same recording simultaneously some form of locking is required in the editor. This PR added the functionality to the editor service to maintain in memory a temporal cache of locks for any valid mediapackage. The lock will expire after a set amount of time and need to be re-obtained. 

This PR doesn't add the locking feature to the built-in editor as it is intended to be used with the standalone editor.

This PR is a heavily reworked version of the original submitted by Tobias for UoM.


etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -79,3 +79,15 @@
 #
 # Default: auto
 #publication.local = auto
+
+# Editor lock Configuration:
```
Arnei: Should mention that this will only work with the new editor

Arnei (approved):
Looks good to me and works as far as my testing goes.

etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -79,3 +79,15 @@
...
+#
+# If this property is true the Editor will lock mediapackages to avoid concurrent editing
+#lock.enable=false
+
+# This Property sets the timeout of the editor lock (in seconds) after which the lock on a mediapackage is released as it is
+# assumed the lock is stale (if this is set to 0 locks will not be released unless the editor is closed properly)
+#lock.release.after.seconds=300
```
gregorydlogan, line 92:
```suggestion
#Default: 300
#lock.release.after.seconds=300
```

etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -79,3 +79,15 @@
...
+
+# This Property sets the time (in seconds) after which the editor should refresh the lock on a mediapackage.
+#lock.refresh.after.seconds=60
```
gregorydlogan, line 96:
```suggestion
#Default: 60
#lock.refresh.after.seconds=60
```

etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -79,3 +79,15 @@
 #
 # Default: auto
 #publication.local = auto
+
+# Editor lock Configuration:
+#
+# If this property is true the Editor will lock mediapackages to avoid concurrent editing
+#lock.enable=false
+
+# This Property sets the timeout of the editor lock (in seconds) after which the lock on a mediapackage is released as it is
```
gregorydlogan, line 89: Config keys should have included defaults, suggestions below.

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -359,6 +371,25 @@ public void activate(ComponentContext cc) {
...
     }
     logger.debug("Thumbnail track priority set to '{}'", thumbnailSourcePrimary);
+
+    lockingActive = Boolean.parseBoolean(StringUtils.trimToEmpty((String) properties.get(OPT_LOCK_ENABLED)));
+
+    try {
+      lockTimeout = Integer.parseUnsignedInt(
+           Objects.toString(properties.get(OPT_LOCK_TIMEOUT)));
+    } catch (NumberFormatException e) {
+      logger.warn("Configuration {} contains invalid value", OPT_LOCK_TIMEOUT);
```
gregorydlogan:
```suggestion
      logger.INFO("Configuration {} contains invalid value, defaulting to {}", OPT_LOCK_TIMEOUT, lockTimeout);
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -359,6 +371,25 @@ public void activate(ComponentContext cc) {
...
+    }
+
+    try {
+      lockRefresh = Integer.parseUnsignedInt(
+            Objects.toString(properties.get(OPT_LOCK_REFRESH)));
+    } catch (NumberFormatException e) {
+      logger.warn("Configuration {} contains invalid value", OPT_LOCK_REFRESH);
```
gregorydlogan:
```suggestion
      logger.info("Configuration {} contains invalid value, defaulting to {}", OPT_LOCK_REFRESH, lockRefresh);
```

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -359,6 +371,25 @@ public void activate(ComponentContext cc) {
       }
     }
     logger.debug("Thumbnail track priority set to '{}'", thumbnailSourcePrimary);
+
+    lockingActive = Boolean.parseBoolean(StringUtils.trimToEmpty((String) properties.get(OPT_LOCK_ENABLED)));
+
+    try {
```
gregorydlogan, line 377: These are logging unclear `WARN`s with the default config, suggestions below:
```
2023-04-18 11:33:34,184 | WARN  | (EditorServiceImpl:381) - Configuration lock.release.after.seconds contains invalid value
2023-04-18 11:33:34,185 | WARN  | (EditorServiceImpl:388) - Configuration lock.refresh.after.seconds contains invalid value
```

modules/editor-service-remote/src/main/java/org/opencastproject/editor/remote/EditorServiceRemoteImpl.java
```diff
@@ -82,13 +86,45 @@ public void setMetadata(String mediaPackageId, String metadata) throws EditorSer
...
+
+  protected String doDeleteForMediaPackage(String mediaPackageId, final String urlSuffix)
+          throws EditorServiceException {
+    logger.debug("Editor Remote Lock POST Request for mediaPackage : '{}'", mediaPackageId);
+    HttpDelete delete = new HttpDelete(mediaPackageId + urlSuffix);
+    HttpResponse response = null;
+    try {
+      response = getResponse(delete, HttpStatus.SC_OK, HttpStatus.SC_NOT_FOUND, HttpStatus.SC_CONFLICT);
+      if (response == null || response.getStatusLine() == null) {
+        throw new EditorServiceException("Editor Remote call failed to repsond", ErrorStatus.UNKNOWN);
```
gregorydlogan: Spelling :)

```suggestion
        throw new EditorServiceException("Editor Remote call failed to respond", ErrorStatus.UNKNOWN);
```

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/EditorRestEndpointBase.java
```diff
@@ -87,6 +91,63 @@ public Response getEditorData(@PathParam("mediaPackageId") final String mediaPac
...
+          },
+          responses = {
+            @RestResponse(description = "Lock obtained", responseCode = SC_CREATED),
+            @RestResponse(description = "Lock not obtained", responseCode = SC_CONFLICT),
+            @RestResponse(description = "Mediapackage not found", responseCode = SC_NOT_FOUND)
+          })
+  public Response lockMediapackage(@PathParam("mediaPackageId") final String mediaPackageId,
+         @Context HttpServletRequest request) {
+    try {
+      String details = readInputStream(request);
```
gregorydlogan: `readInputStream`, at least when testing with the rest docs, always returns an empty string.  It's unclear what `details` is *supposed* to be, but because it's empty `lockData` is null, which causes the lock process to fail [here](https://github.com/opencast/opencast/pull/4658/files#diff-22ae4a40c4d0c69ab7f0385e1853c759f72647351f5b58866fd90eb97d3ae475R59).

gregorydlogan (changes requested):
I'm guessing this is due to the major version difference between your internal code and ours, but this no worky.

gregorydlogan:
@JamesUoM if you want this in 12.N, this needs an update soonish :)

modules/editor-service-api/src/main/java/org/opencastproject/editor/api/EditorRestEndpointBase.java
JamesUoM: This was only intended to complete the remote API. Locking is POST operation and to make re-use of the existing doPostForMediaPackage() function the lock object is converted to JSON and sent as the request body. 
Like the other POST's in this API it should return an 400 error if no body is present -  same as POST /{mediaPackageId}/edit.json ( metadata.json returns 500!)

I guess it should create a default lock if none is supplied though.

JamesUoM: I relented and changed to using form data so the REST docs make sense. I'd gone down the route of trying to use the existing request code here and in the opencast-editor which just has a simple post(json) call. That PR https://github.com/opencast/opencast-editor/pull/1000 will need additional work now too.

gregorydlogan:
@JamesUoM when I look at the output of `/editor/{mediaPackageId}/edit.json`, I see a number of lock related fields, including `lock_uuid`.  Except every time I refetch that endpoint, I see a *different* uuid and none of them match the uuid I sent when locking the mp.

Further behaviour question:
- I seem to be able to create locks as both the global admin user, and test user at the same time as long as the UUID is the same.  Is this expected?
- The UUID I mention above is present (and constantly changing) regardless of whether there are any current locks.  Should this be there?

JamesUoM:
> @JamesUoM when I look at the output of `/editor/{mediaPackageId}/edit.json`, I see a number of lock related fields, including `lock_uuid`. Except every time I refetch that endpoint, I see a _different_ uuid and none of them match the uuid I sent when locking the mp.

This is correct, we'd expect `/edit.json` to be called once per editing session. The UUID is not stored and the mediapackage is not locked with it when editor.js is called. It is up to the editor to use this UUID or to create its own when it requests the lock. 

If the user reloads the editor application, the editor should release either release the lock on page unload (this is what the PR to opencast-editor does) or maintain the UUID between reloads.

In the case the editor is restarted without releasing the lock or retaining the UUID, then the user will have to wait until the timeout has elapsed (which is why it's short) as this just looks like a new session.

> Further behaviour question:
> 
> * I seem to be able to create locks as both the global admin user, and test user at the same time as long as the UUID is the same.  Is this expected?

Yes, the "user" is just metadata. Though there would be no harm in testing it.

> * The UUID I mention above is present (and constantly changing) regardless of whether there are any current locks.  Should this be there?

Essentially this asking whether `/edit.js` should check if the mediapackage is currently locked. I'd not considered that, but checking the lock status is what the new REST call is for and makes for a cleaner API

gregorydlogan:
Ok, sounds good!

PR #4670, opened 2023-02-06
Add a configuration file for the WebVTTCaptionConverter

marwyg:
With this change, you can set how subtitle files will be appended to a mediapackage after a conversion.

### DIFF
--- a/modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
+++ b/modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
@@ -27,7 +27,9 @@ import org.opencastproject.caption.util.TimeUtil;
 import org.opencastproject.mediapackage.MediaPackageElement;
 import org.opencastproject.mediapackage.MediaPackageElement.Type;
 import org.opencastproject.util.OsgiUtil;
+import org.opencastproject.util.data.Option;
 
+import org.apache.commons.lang3.EnumUtils;
 import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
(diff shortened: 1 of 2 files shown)


modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
```diff
@@ -107,7 +122,28 @@ public String getExtension() {
...
+   * @param cc ComponentContext
+   * @return the configured mediapackage element type
+   */
+  private MediaPackageElement.Type getConfiguredMediapackageElementType(ComponentContext cc) {
+    String mediapackageElementType = null;
+    try {
+      mediapackageElementType = OsgiUtil.getOptCfg(cc.getProperties(), MEDIAPACKAGE_ELEMENT_TYPE_CONFIG_KEY).get();
+      // capitalize the first letter
+      mediapackageElementType = mediapackageElementType.substring(0, 1).toUpperCase()
+          + mediapackageElementType.substring(1);
```
lkiesow: If you want to ignore casing, you need to convert this as well:

```suggestion
          + mediapackageElementType.substring(1).toLowerCase();
```

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
```diff
@@ -107,7 +122,28 @@ public String getExtension() {
...
+      return MediaPackageElement.Type.valueOf(mediapackageElementType);
+    } catch (Exception e) {
+      logger.warn("Couldn't convert configuration '{}'='{}' into enum. Using default '{}'.",
+          MEDIAPACKAGE_ELEMENT_TYPE_CONFIG_KEY, mediapackageElementType, DEFAULT_MEDIAPACKAGE_ELEMENT_TYPE);
+      return DEFAULT_MEDIAPACKAGE_ELEMENT_TYPE;
```
lkiesow: Please don't do that!
As an administrator, if I accidentally misconfigure this, don't just choose something.
Use the default __only__ if nothing is set.
If I set something that does not make any sense, please fail, so I realize that immediately and don't realize after a few weeks that Opencast did always silently do the wrong thing.

lkiesow (changes requested):

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
marwyg: For this, I switched to org.apache.commons.lang3.EnumUtils.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
marwyg: Sorry for that. Now it will fail if something wrong was configured and a detailed message will be printed.

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
```diff
@@ -107,7 +124,41 @@ public String getExtension() {
...
+  private Type convertStringToEnum(String mediaPackageElementType) {
+    if (EnumUtils.isValidEnumIgnoreCase(MediaPackageElement.Type.class, mediaPackageElementType)) {
+      return EnumUtils.getEnumIgnoreCase(MediaPackageElement.Type.class, mediaPackageElementType);
+    }
+    // Conversion didn't work. Print an error and throw an exception
+    String additionalErrorInfo = String.format("Please check the config file '%s'.", this.getClass().getName());
+    String errorMessage = String.format("Couldn't convert configuration '%s'='%s' into enum. " + additionalErrorInfo,
+        MEDIAPACKAGE_ELEMENT_TYPE_CONFIG_KEY, mediaPackageElementType);
+    logger.error(errorMessage);
+    throw new IllegalArgumentException(errorMessage);
```
lkiesow: Why log this if you throw this as an error, which will then be logged as an error?
Just throwing the error should be sufficient and should also make the `format()` unnecessary.

lkiesow (changes requested):

modules/caption-impl/src/main/java/org/opencastproject/caption/converters/WebVttCaptionConverter.java
marwyg: I removed the logging line and thinned out the error message.

lkiesow (approved):

PR #4684, opened 2023-02-09
Ensure workflows have an associated org after upgrade

gregorydlogan:
This PR ensures that all workflows have an organization associated with them after an upgrade.  This was reported on list by Sven Laudel, along with the fix by Tilman

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -2198,8 +2198,18 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
               }
               current++;
 
-              var updatedWorkflowData = index.getEvent(indexData.getMediaPackageId(), indexData.getOrganizationId(),
-                        securityService.getUser());
+              String orgid = indexData.getOrganizationId();
+              if (null == orgid) {
+                //NB: This version of getWorkflow takes the org id, which in this case is null
+                // Using the normal version filters by org, and since this workflow has a NULL org it can't be found
+                WorkflowInstance instance = persistence.getWorkflow(indexData.getId(), orgid);
+                //We're assuming here that mediapackages don't change orgs
+                orgid = assetManager.getSnapshotsById(indexData.getMediaPackageId())
+                    .getSnapshots().head2().getOrganizationId();
+                instance.setOrganizationId(orgid);
+                persistence.updateInDatabase(instance);
+              }
+              var updatedWorkflowData = index.getEvent(indexData.getMediaPackageId(), orgid, securityService.getUser());
               updatedWorkflowData = getStateUpdateFunction(indexData).apply(updatedWorkflowData);
               updatedWorkflowRange.add(updatedWorkflowData.get());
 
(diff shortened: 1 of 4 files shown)


gregorydlogan:
Discussion questions:

[Approach 1](https://github.com/opencast/opencast/pull/4684/commits/503f8566004da68e0aa2718ee5d476e2ba42ba7c): Is [this](https://github.com/opencast/opencast/pull/4684/commits/503f8566004da68e0aa2718ee5d476e2ba42ba7c#diff-80eb4e148bd7d6c02871d1ba3f104b935ac544bec652d225f700384befb07602R212) a safe assumption?  The rebuild is called in a per-org context, so I think so?  Otherwise I need to wire in the AM to the WF database and that seems like a bad idea.
[Approach 2](https://github.com/opencast/opencast/pull/4684/commits/fcc0229f749a7365bf2a084453ad83397520d14f): Has access to the AM, yay.  But now I need a bunch more code, and it's spread out.

KatrinIhler:
I don't think the first approach is safe. If I understand this correctly, the security service will return whatever organization the index rebuild was triggered from, not the organization the workflow actually belongs to. This will be a problem for multi-tenant systems, and the resulting data corruption will be much harder to fix than the initial problem.

Approach 2 needs to be able to handle the case that the media package doesn't exist anymore, I think? But then it should be fine.

gregorydlogan:
Approach 1 removed, updated approach 2 with proper handling for the missing edgecase.

gregorydlogan:
Ping @KatrinIhler

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2198,8 +2198,24 @@ public void repopulate() throws IndexRebuildException {
...
+                //We try-catch here since it's possible for the WF to exist in the *index* but not in the *DB*
+                // It probably shouldn't be, but that won't keep it from happening anyway.
+                try {
+                  //NB: This version of getWorkflow takes the org id, which in this case is null
+                  // Using the normal version filters by org, and since this workflow has a NULL org it can't be found
+                  WorkflowInstance instance = persistence.getWorkflow(indexData.getId(), null);
+                  instance.setOrganizationId(orgid);
+                  persistence.updateInDatabase(instance);
+                } catch (NotFoundException e) {
+                  logger.warn("Workflow {} not found in the database, but present in the index", indexData.getId());
```
KatrinIhler: I don't think this can ever happen, because the info that this workflow exists came from the database, not the index (even though it's called indexData).

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -2198,8 +2198,24 @@ public void repopulate() throws IndexRebuildException {
               }
               current++;
 
-              var updatedWorkflowData = index.getEvent(indexData.getMediaPackageId(), indexData.getOrganizationId(),
-                        securityService.getUser());
+              String orgid = indexData.getOrganizationId();
+              if (null == orgid) {
+                //We're assuming here that mediapackages don't change orgs
+                orgid = assetManager.getSnapshotsById(indexData.getMediaPackageId())
```
KatrinIhler: We should check if the result is empty here, in case the media package associated with the workflow doesn't exist anymore.

KatrinIhler (changes requested):
Slight misunderstanding here - I meant that we need to handle that the _mediapackage_ might not exist, we know that the workflow exists.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
gregorydlogan: Ok, will remove the logging line.  `getWorkflow` throws this so we have to catch it eventually :)

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
gregorydlogan: Good catch, added.

KatrinIhler (approved):

PR #3376, opened 2022-01-26
Change the way workflows are stored in the database from XML to tables

Arnei:
This PR is in preperation for:
- #3377!

<hr>

Currently, workflows are stored as a single XML object in the database. This makes it difficult/inperformant to use more complex queries, which is why workflow information is also stored in a Solr Index. In preparation for removing Solr from the Workflow Service, this PR creates new database tables for workflow information.

Key changes:
- Workflow information is now stored in tables named "oc_workflow*"
- Workflow information is no longer stored in the job tables "payload" column. Instead, it stores the workflow_id there (which is the same id as the corresponding "START_WORKFLOW" job, so it is actually kinda superfluous)
- Therefore, the WorkflowInstance and WorkflowOperationInstance classes are now using JPA annotations instead of JAXB annotations.
- Workflows still require XML representations for e.g. REST endpoints, so corresponding JAXB classes for that purposes are added
- Requires a db upgrade script. The script is a python script, placed in the usual sport for db upgrade scripts.
- Adds an endpoint that allows to query for workflow instances by mediapackage id

Work sponsored by our generous Crowdfunding initiatives!

<hr>

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4693, opened 2023-02-14
Fix Workflow Condition Parsing

KatrinIhler:
This restores the behavior of 11.x where the parsing of the workflow execution condition is concerned. Otherwise, one would now need to add ' to all workflow variables in the execution condition that resolve to strings. This was changed - probably unintentionally - by #3376.

### DIFF
--- a/docs/guides/admin/docs/releasenotes/workflow-condition-fix-12.x
+++ b/docs/guides/admin/docs/releasenotes/workflow-condition-fix-12.x
@@ -1,0 +1,5 @@
+Fix workflow conditions with string variables
+
+Re-enable parsing of workflow conditions with string variables where the variable is unquoted, as was possible
+in 11.x, while still keeping the option to quote string variables as was introduced in 12.0.
+This change was initially aimed at 12.x, 12.8 or higher.
--- a/modules/workflow-condition-parser/src/main/java/org/opencastproject/workflow/conditionparser/WorkflowConditionInterpreter.java
+++ b/modules/workflow-condition-parser/src/main/java/org/opencastproject/workflow/conditionparser/WorkflowConditionInterpreter.java
@@ -97,7 +97,14 @@ public final class WorkflowConditionInterpreter {
               result.append(toAppend);
             }
           } catch (NumberFormatException e) {
-            result.append("'").append(toAppend.replace("''", "'")).append("'");
+            // quote string value if variable not already quoted
+            if (matchStart == 0 || source.charAt(matchStart - 1) != '\'') {
+              result.append("'");
+            }
+            result.append(toAppend.replace("''", "'"));
+            if (matchEnd == source.length() || source.charAt(matchEnd) != '\'') {
+              result.append("'");
+            }
           }
         }
       } else {
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowOperationWorker.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowOperationWorker.java
@@ -176,7 +176,7 @@ final class WorkflowOperationWorker {
       return null;
     };
     final String executionCondition = WorkflowConditionInterpreter.replaceVariables(
-        operation.getExecutionCondition(), variables, null, false);
+        operation.getExecutionCondition(), variables, null, true);
     operation.setExecutionCondition(executionCondition);
     operation.setDescription(WorkflowConditionInterpreter.replaceVariables(
         operation.getDescription(), variables, null, false));
--- a/modules/workflow-service-impl/src/test/resources/workflow-definition-skipping-intricate.xml
+++ b/modules/workflow-service-impl/src/test/resources/workflow-definition-skipping-intricate.xml
@@ -5,7 +5,7 @@
   <operations>
     <operation
       id="op1"
-      if="'${executecondition}' == 'a'"
+      if="${executecondition} == 'a'"
       fail-on-error="true"
       description="operation 1"/>
   </operations>


Arnei:
> This was changed - probably unintentionally - by https://github.com/opencast/opencast/pull/3376.

*Looks at code.* [I have no memory of this place.](https://i.kym-cdn.com/photos/images/newsfeed/001/071/150/034.jpg)

If this fixes things, I'm totally fine with it :D

lkiesow:
This is likely my change when I took over the pull request from you.
We also had a long discussion about security implications of the old mechanism at some point in the past and I remember that we found several problems with it.

I found the intermediate patch that made this change but got combined with the big change in the end. I'm not sure if this was the only one in this direction, but was certainly the main one:

https://github.com/opencast/opencast/pull/3376/commits/e26b9336af0daa71c2168d79fba6ffdb9b720bcc#diff-f9296c149dcdf3bfe423a3c6e15348390dba8d7e99aa092078e141fab2982ed1

Since I've also updated the test and not only the code, my guess would be that this change was intentional. But I really don't remember the details any longer. Maybe it had security implications? But maybe it's also just consistency so that replacements in different places work the same way.

lkiesow (changes requested):
Thinking about this a bit more, this patch will actually break configuration compatibility between minor versions. If we merge this into 12.x and someone applies the upgrade, their workflows may actually fail due to this. In other words, this patch needs to go into `develop`.

Which brings me to the question if you actually think we should change this back or leave it like this. Whatever we decide, I agree that we need to document this properly. But do you see an advantage in the old syntax over the new one?

Not digging in too deep, I did not remember any security problem in this specific case, but I think this is actually nice for consistency. Right now, the condition would be identical for both these workflows:

```
❯ curl -u admin:opencast http://localhost:8080/ingest/addMediaPackage/fast -F 'flavor=presenter/source' -F mediaUri=https://data.lkiesow.io/opencast/test-media/goat.mp4 \
    -F title=title -F x=yes
❯ curl -u admin:opencast http://localhost:8080/ingest/addMediaPackage/fast -F 'flavor=presenter/source' -F mediaUri=https://data.lkiesow.io/opencast/test-media/goat.mp4 \
    -F title=title -F x=true
```

It would always be `if="'${x}' == 'VAL'". Changing this back, you would have to treat these cases differently based on their implicit types.

I don't have a strong opinion wither way. Apart from this going into 12.x as is, that is.

KatrinIhler:
~~I agree with you that this patch needs to go into develop to avoid further complications. I'm gonna add a line to the future release notes as well. :)~~ (Or maybe not, see end of comment.)

I disagree with you on the consistency thing. I did some testing to figure out what workflow conditions currently work in develop, and what changed since 11.x. I found the following:

**Boolean:**
Both quoting or not-quoting (`${a} == true` and `'${a} == 'true'`) works when comparing explicitly. However, mostly you will simply want to write `${a}`, as in our current workflows. Here, quoting doesn't work. The current docs treat `true` and `false` as constants that don't require quoting.

**Numbers:**
Both quoting and not-quoting (`${a} == 5` vs `'${a}' == '5'` works. The docs don't require numbers to be quoted, and I would personally also prefer the first variation.

**Strings:**
Now here it gets interesting. Strings are the only values that are required to be wrapped into single-quotes as per our documentation. So `${a} == test` doesn't work and never has. In 11.x and earlier, you simply wrote `${a} == 'test'`. Since 12.x, you have to quote the variable as well: `'${a}' == 'test'`.

This feels unintuitive to me. In my opinion, the quotes are part of the string value, so they should be substituted when replacing the variable with its value. We also don't require variable quoting for the other cases described above (although they still work). So string variables are still a special case, albeit differently than before. And if that can't be helped, I like the old way better.

There is a third option, however. The code substituting the variable could recognize if the variable is already quoted, and only quote the value if that isn't the case, thus avoiding quoting twice. This would also be backwards compatible with the prior change, meaning this patch could still go into 12.x, _and_ improve consistency with the other cases as well.

I think I'm going to look into that...

lkiesow:
Sounds reasonable

KatrinIhler:
@lkiesow Ready for review again! :)

mtneug:
> There is a third option, however. The code substituting the variable could recognize if the variable is already quoted, and only quote the value if that isn't the case, thus avoiding quoting twice. This would also be backwards compatible with the prior change, meaning this patch could still go into 12.x, and improve consistency with the other cases as well.

I really dislike this approach. I think the variable substitution logic should be very simple. This logic is unexpected for me. What if I want to have quotes as part of my string?

mtneug:
I did some digging but only found this discussion:

quotes

I definitely remember that there were two code paths for variable substitution: one replaced variables before/after(?) each workflow operation if this variable is available, and the other replaced variables during condition evaluation which would replace non-existing variables with `false`. Making quotes not part of the variable would expose cases where a variable is missing during the condition evaluation. Maybe this is also part of the reason why this was changed.

lkiesow:
> > There is a third option, however. The code substituting the variable could recognize if the variable is already quoted, and only quote the value if that isn't the case, thus avoiding quoting twice. This would also be backwards compatible with the prior change, meaning this patch could still go into 12.x, and improve consistency with the other cases as well.
> 
> I really dislike this approach. I think the variable substitution logic should be very simple. This logic is unexpected for me. What if I want to have quotes as part of my string?

Do you have a specific suggestion for going forward? Is that keeping the current state and not applying this patch? Or is it to never quite strings?

mtneug:
> Do you have a specific suggestion for going forward? Is that keeping the current state and not applying this patch? Or is it to never quite strings?

I would suggest leaving it as is. Adopting this for custom workflows is quite easy and this is the syntax for some releases now. We should only add a notice in the docs (and maybe on list as well).

gregorydlogan:
We're basically out of time here, is this going to happen?

gregorydlogan:
Closing this since 12.x is about to be over.  If further work gets done on this please feel free to open a new PR.

PR #4695, opened 2023-02-16
Add new admin UI as beta

Arnei:
This PR adds the new Admin UI as optional feature alongside the old Admin UI. This allows users to the test out the new Admin UI and easily switch back the old Admin UI if they find the new one to be lacking in some regard.

Much like Studio and the Editor, the code for the new Admin UI lives in its own repository https://github.com/opencast/opencast-admin-interface (well not quite yet, https://github.com/opencast/opencast-admin-interface/pull/134 still needs to be merged). The new Admin UI will be built and released in the new repository as well, Opencast will just need to download and unpack the build. This should make development of the new Admin UI easier, and the build process for Opencast shorter.

This PR also which modules are responsible for login. `runtime-info-ui` is now the single module responsible for deploying the login page, the page in `admin-ui-frontend` goes unused. This allows us to get rid of the `runtime-info-ui-ng` module and hopefully lessen confusion regarding the login page. After logging in, a user will be redirected to the Admin UI if on the admin node and to the Rest Docs if on presentation or worker node.

**UPDATE: https://github.com/opencast/opencast-admin-interface/pull/134 has been merged, this PR is ready now!**


etc/security/mh_default_org.xml
```diff
@@ -29,12 +29,13 @@
...
     <sec:intercept-url pattern="/" access="ROLE_ANONYMOUS" />
-    <sec:intercept-url pattern="/admin-ng/login.html" access="ROLE_ANONYMOUS" />
+    <sec:intercept-url pattern="/login.html" access="ROLE_ANONYMOUS" />
     <sec:intercept-url pattern="/sysinfo/bundles/version" method="GET" access="ROLE_ANONYMOUS" />
 
     <!-- Allow anonymous access to resources from runtime info ui -->
     <sec:intercept-url pattern="/img/**" access="ROLE_ANONYMOUS" />
-    <sec:intercept-url pattern="/js/**" access="ROLE_ANONYMOUS" />
+    <sec:intercept-url pattern="/scripts/**" access="ROLE_ANONYMOUS" />
+    <sec:intercept-url pattern="/styles/**" access="ROLE_ANONYMOUS" />
```
mtneug, line 38: Why the new paths? Can everything beloging to the admin UI be prefixed with `/admin-ui/...`?

Reason: simple loadbalancer rules when routing HTTP requests. I plan to run the admin UI as separate containers.

Arnei: Those are for the `runtime-info-ui` module, which contains the login page and the rest docs, and is also responsible for redirecting to the `/admin-ui/index.html` after login.

I tried to explain why the login page is in a different module from the admin ui in the initial post. Also, I got very confused by Spring Security authentication and redirect methods and therefore opted what seemed to me like the simplest solution, resulting in these paths.

modules/admin-ui-interface/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <parent>
+        <groupId>org.opencastproject</groupId>
+        <artifactId>base</artifactId>
+        <version>14-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+    <properties>
+        <opencast.basedir>${project.basedir}/../..</opencast.basedir>
+        <checkstyle.skip>false</checkstyle.skip>
+        <interface.url>https://github.com/opencast/opencast-admin-interface/releases/download/2023-04-12-2/oc-admin-ui-2023-04-12.tar.gz</interface.url>
```
gregorydlogan, line 16: How is this kept up to date as the other repo changes?  Same way as Studio - manually?

etc/workflows/duplicate-event.xml
```diff
@@ -85,6 +85,21 @@
       </script>
     ]]>
   </configuration_panel>
+  <configuration_panel_json>
```
gregorydlogan, line 88: There are a bunch of fields which show up in the old admin ui, but not the new one for this workflow (which only shows the number of dups, as defined here).  This causes things to be unhappy, but work:
```
2023-04-20 15:23:01,481 | INFO  | (DuplicateEventWorkflowOperationHandler:281) - Creating 4 new media packages from media package with id 0e583a0f-d1bc-450d-8f85-23aa5d95d176.
2023-04-20 15:23:01,487 | WARN  | (DuplicateEventWorkflowOperationHandler:371) - Could not parse: ${startDateTime} as date time
2023-04-20 15:23:01,488 | WARN  | (DuplicateEventWorkflowOperationHandler:373) - Using original event date Thu Apr 20 14:58:00 CST 2023 as default
```

gregorydlogan (changes requested):
Very, very minor changes.  Otherwise looks good to me!

Edit: Also, make a note of this in the release notes as well

modules/admin-ui-interface/pom.xml
Arnei: Pretty much, yeah.

Arguably it would be better to have a branch and release structure more similar to Opencast, so you could more easily backport changes, but for now I went with a way I know how to set up ^^'

etc/workflows/duplicate-event.xml
Arnei: It seems the new Admin UI currently only supports workflow config fields of types `radio`, `checkbox`, `number` and `text`. So I can't just add field of type `datetime-local` to the `configuration_panel_json`, as it is unsupported (I mean, I tried anyway, but of course it didn't work).

I'll open an issue for that. Do you want this fixed before merging?

gregorydlogan: Nope, this is a beta so the occasional issue is to be expected.  This would block promotion of Picard to the default ui though.

modules/admin-ui-interface/pom.xml
gregorydlogan: That's fair.  There's a conversation to be had in the community in terms of how this *should* work for all three modules.

PR #4711, opened 2023-02-22
Fix caption creation in lti for subtitles as tracks

Arnei:

Changes how a caption file is added to an event upserted through lti by the new standards.

This can likely be improved upon by someone who actually knows how lti and the ltitools are used in practice. For now the goal was to simply keep everything in working order.


mliradelc:
Can't try this because we don't use LTI, codewise seems good.

Issue #4704, opened 2023-02-20
S3 documentation mentions nonexistent workflow


JulianKniephoff:
The [Using S3 Archiving](https://docs.opencast.org/develop/admin/#modules/awss3archive/#using-s3-archiving) section of the [AWS S3 Archive Configuration](https://docs.opencast.org/develop/admin/#modules/awss3archive) documentation page still references the `offload.xml` workflow configuration file which was removed in 4087ed032e6152702ee6cae4c94dd0481b921b48.

I would have just stricken the reference from the document but I don't know how to do what it is describing without that workflow. :thinking:

PR #4721, opened 2023-02-28
Clarify S3 docs

gregorydlogan:
This pull request removes a reference to a now-deleted workflow from the AWS S3 archiving docs.  I've replaced the link with example WOH definitions instead.

Fixes: #4704

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/modules/awss3archive.md
+++ b/docs/guides/admin/docs/modules/awss3archive.md
@@ -49,10 +49,10 @@ Using S3 Archiving
 S3 archiving is done on a Snapshot level, that is a mediapackage ID + version.  Because of the way that the Asset
 Manager handles snapshots, all newly created snapshots are *always* local.  Creating a snapshot of a mediapackage with
 non local data will download *all* related snapshots for that mediapackage which can incur significant costs.  S3
-archiving is meant to be a cost reduction, and storage expansion tool, rather than then primary storage for your
-asset manager.  Therefore, most adopters do not want to immediately (ie, at the end of your default workflow) offload
-your recordings to S3!  Instead, we suggest using the TimedMediaArchiver
-(org.opencastproject.assetmanager.impl.TimedMediaArchiver.cfg) to offload your recordings after sufficient time that
+archiving is meant to be a cost reduction, and storage expansion tool, rather than hot storage where lots of reads and
+writes will occur.  Therefore, most adopters do not want to immediately (ie, at the end of your default workflow)
+offload your recordings to S3!  Instead, we suggest using the `TimedMediaArchiver` as configured in
+`/etc/org.opencastproject.assetmanager.impl.TimedMediaArchiver.cfg` to offload your recordings after sufficient time that
 further modification of the recording is unlikely.
 
 If you do need to create an additional workflow, a substantially better approach than restoring snapshots involves
--- a/docs/guides/admin/docs/modules/awss3archive.md
+++ b/docs/guides/admin/docs/modules/awss3archive.md
@@ -46,9 +46,25 @@ The Opencast AWS S3 Archive service configuration can be found in the
 Using S3 Archiving
 ------------------
 
-Moving assets to S3 is done via a workflow operation handler added as part of a workflow.  This can be part of your
-main workflow that runs on all of your recordings, or a small manually activated workflow accessed on-demand from the
-tasks menu.  The workflow operation handler definition looks like this
+S3 archiving is done on a Snapshot level, that is a mediapackage ID + version.  Because of the way that the Asset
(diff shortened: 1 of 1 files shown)


docs/guides/admin/docs/modules/awss3archive.md
```diff
@@ -46,13 +46,33 @@ The Opencast AWS S3 Archive service configuration can be found in the
...
+changing the `target-storage` configuration value from `aws-s3` to `local-filesystem` like so
+```
+    <operation
+      id="move-storage"
+      description="Restoring from AWS S3">
+      <configurations>
+        <configuration key="target-storage">local-filesystem</configuration>
+      </configurations>
+    </operation>
+```
```
mtneug, line 89: I'm unsure if this "restore paragraph" is a good advice without more context. "Moving" is always based on snapshots. The `target-version` is used to specify what snapshot to move. It not specified (as done here) all snapshots are moved. This may be what is desired, but not necessarily.

The issue is that new snapshots are *always* created locally. You, therefor, mention that `move-storage` could be part of all workflows. Instead, I suggest using the tiered storage feature that moves older snapshots automatically to S3, e.g., after two weeks. Especially during the time of the ingest, there could be quite a lot of new snapshots because of metadata/ACL changes or workflows. The problem is that shapshots always contain all selected data, resulting in many downloads from S3.

When you know that a certain workflow will create a new shapshot, I would also not use `move-storage` as this would restore existing snapshots. Instead I would run `ingest-download` at the start of the workflow, which will put S3 files into the workspace. The `snapshot` operation is then very quick, but I think some operations also assume that everything is available locally (at least, I remember having problems with `sanitize-adaptive` when I did not download everything beforehand).

gregorydlogan: Do you by any chance have a workflow that's already set up to do that?  That makes a ton of sense.

mtneug: We still publish a reference to our workflows here (https://github.com/opencast/community-workflows/tree/master/13.x/educast.nrw), but they are completely different from upstream workflows and assume certain changes. However, the `ingest-download` workflow is rather simple: https://zivgitlab.uni-muenster.de/educast-nrw/opencast/workflows/-/blob/educast/13/src/workflows/educast-partial-setup-workspaces.yaml. This is included in every main workflow that will eventually create a snapshot or otherwise needs files locally.

docs/guides/admin/docs/modules/awss3archive.md
```diff
@@ -46,20 +46,48 @@ The Opencast AWS S3 Archive service configuration can be found in the
...
 ------------------
 
-There are two major methods to access S3 archiving features: manually, and via a workflow.  Amazon S3 archiving is not
-part of the default workflows and manual S3 offload is disabled by default.  To enable manual S3 offload you must edit
-the `offload.xml` workflow configuration file and change `var s3Enabled = false;` to `var s3Enabled = true;`.  To
-manually offload a media package follow the directions in the user documentation.
+S3 archiving is done on a Snapshot level, that is a mediapackage ID + version.  Because of the way that the Asset
+Manager handles snapshots, all newly created snapshots are *always* local.  Creating a snapshot of a mediapackage with
+non local data will download *all* related snapshots for that mediapackage which can incur significant costs.  S3
+archiving is meant to be a cost reduction, and storage expansion tool, rather than then primary storage for your
```
mtneug:
```suggestion
archiving is meant to be a cost reduction, and storage expansion tool, rather than the primary storage for your
```

docs/guides/admin/docs/modules/awss3archive.md
```diff
@@ -46,20 +46,48 @@ The Opencast AWS S3 Archive service configuration can be found in the
...
+asset manager.  Therefore, most adopters do not want to immediately (ie, at the end of your default workflow) offload
+your recordings to S3!  Instead, we suggest using the TimedMediaArchiver
+(org.opencastproject.assetmanager.impl.TimedMediaArchiver.cfg) to offload your recordings after sufficient time that
```
mtneug: Is this picked up by the JavaScript config linking bit? Does that need `etc/` as prefix?

mtneug (changes requested):
I think this reads nicely now. Although, I'm not sure what the definition of "primary storage" is. Main working space, sure. Place where most files are stored, probably not with S3. But I think the message comes across.

docs/guides/admin/docs/modules/awss3archive.md
gregorydlogan: It's not.  But I've reworked that bit, since it didn't read nicely.

mtneug (approved):

PR #4745, opened 2023-03-06
Include Amberscript-Transcription Documentation in Module Overview 

lars-amberscript:
Updating the overview file to also include a link to the Amberscript Transcription module documentation

- While the documentation is already available, it is not yet mentioned on the Overview Page regarding the modules
- This is simply just adding the link to it to the page


KatrinIhler:
If you're planning to contribute in the future, you should probably file that ICLA - but since this is just a minor docs change, I'm gonna merge this as is.

PR #4791, opened 2023-03-16
fixed faulty regex on analyze-mediapackage

### LINKED ISSUE #4790: analyze-mediapackage replaces every symbol that doesn't match [^a-z0-9] regex to underscore ("_")
analyze-mediapackage will REPLACE every symbol that doesn't match [^a-z0-9] regex to underscore ("_"): capital letters, plus sign ("+"), etc.
example: analyze-mediapackage on captions/*: captions/vtt+auto -> captions_vtt_auto_exists

**To Reproduce**

1. Run analyze-mediapackage on an event on a track which has a flavor that contains capital letter, or plus sign
2. Check out created variables, everything that doesn't match [^a-z0-9] is replaced with "_"

**Expected behavior**

For example "analzye-tracks" will leave '+' (and probably other symbols) untouched when creating new workflow variables. Example: track with audio and flavor presenter/trimmed+audio+auph -> presenter_trimmed+audio+auph_audio

**Server environment:**
 - Version 12.4

aliyevTU:
Closes #4790
 
fixed faulty regex on analyze-mediapackage
analyze-mediapackage will REPLACE every symbol that doesn't match [^a-z0-9] regex to underscore ("_"): capital letters, plus sign ("+"), etc.
example: analyze-mediapackage on captions/*: captions/vtt+auto -> captions_vtt_auto_exists


aliyevTU:
> Hi @aliyevTU Thank you for contributing to Opencast. We noticed that you have not yet filed an [Individual Contributor License Agreement](https://www.apereo.org/licensing/agreements/icla). Doing that (once) helps us to ensure that Opencast stays free for all. If you make your contribution on behalf of an institution, you might also want to file a [Corporate Contributor License Agreement](https://www.apereo.org/licensing/agreements/ccla) (giving you as individual contributor a bit more security as well). It can take a while for this bot to find out about new filings, so if you just filed one or both of the above do not worry about this message! Please let us know if you have any questions regarding the CLA.

I have filled and agreed to CLA. Is it an issue with CLA checking script or do I have to refill CLA?

lkiesow:
> I have filled and agreed to CLA. Is it an issue with CLA checking script or do I have to refill CLA?

It's neither. I can confirm that you submitted your ICLA on Thu, 23 Feb 2023. But Apereo has not processed it yet. That can be a bit slow at times. Just ignore the test for now and point people to this comment if someone asks about it.

lkiesow:
You also don't need to file an issue if you immediately file a pull request for that. Spare yourself the work an just file a pull request.

lkiesow:
Looking at the failing test, the behavior seems to be by design. Your patch causes the tests to fail since they expect an underscore instead of a plus sign. I'm also not sure how the workflow conditions handle something like a `+` in variable names. Does that work?

aliyevTU:
> Looking at the failing test, the behavior seems to be by design. Your patch causes the tests to fail since they expect an underscore instead of a plus sign. I'm also not sure how the workflow conditions handle something like a `+` in variable names. Does that work?

Of course, every single WOH works with + sign. The thing is that this implementation that I am trying to fix is not only replacing +, but also capital letters. We use + sign on many parts of our different workflows.
For example "analzye-tracks" will leave '+' (and probably other symbols) untouched when creating new workflow variables. Example: track with audio and flavor presenter/trimmed+audio+auph -> presenter_trimmed+audio+auph_audio

aliyevTU:
Should I also change the tests?

lkiesow:
If you think the behavior is a bug, then the test is also buggy and needs fixing. Your current patch cannot be merged like this. It would break Opencast builds like this due to the failing test. Additionally, you need to adopt the workflow operation's documentation, which also states the current behavior. Finally, it would be great if you could put a note in the releasenotes folder. A change like this should be pointed out since many users will need to adapt their workflows due to the change in behavior.

> Of course, every single WOH works with + sign. 

The question was about workflow conditions specifically. For example, ingesting something like `presenter/source+a+b` will result in a variable `presenter_source_a_b_type` right now. That is safe to use in a workflow condition:

```yaml
- id: inspect
  if: "${presenter_source_a_b_type} == 'Track'"
```

My question was if this is still true if you allow special characters in variables? Your code would require writing something like:

```yaml
- id: inspect
  if: "${presenter_source+a+b_type} == 'Track'"
```

Worse, what if the flavor contains something like `}`? This would probably fail:

```yaml
- id: inspect
  if: "${presenter_source}a}b_type} == 'Track'"
```

That's why I'm wondering if this should really be changed like this. That said, I didn't take a close look at the code or actually tested it.

If you start allowing special characters, I'm also wondering why you are still replacing the `/` character? If they are not a problem, why not keep all of them?

If, on the other hand, this is really only about the uppercase letters being replaced, I would suggest a change like this instead:

```java
String flavorPrefix = mpe.getFlavor().toString().toLowerCase().replaceAll("[^a-z0-9]", "_");
```

aliyevTU:
These are all very good points. I will look into these more in depth and get back to you.

aliyevTU:
As we discussed it with @gregorydlogan  at Opencast technical meeting on 28.03.2023:
The main issue is that `analyze-track` WOH behaves differently to `analyze-mediapackage`. This discrepancy can create problems for adopters with custom workflows. We as adopter are using `+` in our flavors without any issue for many years now. I will update the tests and update the PR.

lkiesow:
> We as adopter are using + in our flavors without any issue for many years now.

This doesn't really answer any of the question from me, does it? The `+` sign is used in flavors by default. No doubt about that. But you are trying to create variables with arbitrary characters in them. That hasn't anything to do with flavors any longer. Or at least only indirectly. I also don't really understand why this would lead to any error for adopters since the behavior is well-defined.

Is your argument is usability in the sense, that it would be easier or less confusing for users if both workflow operation handler would use the same variable names? If so, I'm fine with adjusting that. But __only__ if you then actually use the same algorithm for generating variable names. Right now, your replacement is generating slightly different variable names.

This is the algorithm used by `analyze-tracks`:
https://github.com/opencast/opencast/blob/0f6b717aa99866f41c360622361ce5e17f84a64e/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AnalyzeTracksWorkflowOperationHandler.java#L160-L163

You also need to update this operation's documentation to reflect the different behavior and should add [a note to go into the release notes](https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs/releasenotes).

And finally, it would be great if you could update this pull request's title and description to reflect your intent. That's less important, but would still be great as it can be really useful if you try to find patches in the future.

aliyevTU:
> But you are trying to create variables with arbitrary characters in them. 

How you came to this conclusion?

>I also don't really understand why this would lead to any error for adopters since the behavior is well-defined.

It is clear that you have not understood what the bug is or what the changes I introduced are meant to do. Please attend the technical meeting this week, and I will go over it once again just for you. I already explained everything in detail to @gregorydlogan .

lkiesow:
I should likely be at the meeting tomorrow. If I cannot make that, we can hop into a video call anytime, if you think that will help to explain.

aliyevTU:
Can we get it merged so that I can close this issue also from my internal issue list? @lkiesow will send an email for a video call appointment.

lkiesow:
Listing the points from the meeting earlier again.
I'm definitely not opposed to making `analyze-mediapackage` generate the same variables `analyze-trakcs` generates.
We just should make sure that people are aware of the change since they may need to adapt their workflows.

Based on this goal and the current state of this pull request, here is a todo list:

- [ ] Make sure  `analyze-mediapackage` actually generate the same variables `analyze-trakcs` does. For example, given a flavor `a/B` your code would right now generate variables like `a_b_…` while `analyze-tracks` would generate `a_B_…`. If you want them to work the same, I suggest simply using the same algorithm for both. This is the algorithm used by `analyze-tracks`:
    https://github.com/opencast/opencast/blob/0f6b717aa99866f41c360622361ce5e17f84a64e/modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AnalyzeTracksWorkflowOperationHandler.java#L160-L163
- [ ] You need to [update this operation's documentation](https://github.com/opencast/opencast/blob/510ce7e65a7504965c17f139fb77c8fa32314b27/docs/guides/admin/docs/workflowoperationhandlers/analyze-mediapackage-woh.md) to reflect the different behavior. Right now, it states e.g. that it would generate a variable like `security_xacml_episode_exists=true` which (assuming the default flavors) would be `security_xacml+episode_exists=true` after your patch is applied.
- [ ]  You should add [a note to go into the release notes](https://github.com/opencast/opencast/tree/develop/docs/guides/admin/docs/releasenotes). Adopters are using this in its current form at the moment. We should point out that something has changes. See the readme in that folder for details, but essentially, just place a text file with a few lines of explanation in that directory. This should not be long.
- [ ] (optional, but would be great) Update this pull request's title and description to reflect your intent (make both operations work alike). That's less important, but would still be great as it can be really useful if you try to find patches in the future and you don't have to dig through a lengthy discussion down here ;-) Also, the Title will be what is shown in the changelogs.

I hope this makes things clearer.

aliyevTU:
@lkiesow after changing docs, mkdown style makes codeblock properties not render + correctly as in:
image

How can I fix this?

aliyevTU:
Updated algo, and docs, added release note. Thanks for tips @lkiesow Happy to be finally be done with this.

aliyevTU:
@lkiesow you there?

docs/guides/admin/docs/releasenotes/analyze-mediapackage
```diff
@@ -0,0 +1,4 @@
+Changed behaviour of analyze-mediapackage
+-----------------------------------------
+
+Behaviour of analyze-mediapackage has been changed with this fix. The buggy behaviour was replacing every character that doesn't match a-z or 0-9 with underscore character. This fix only adds underscore character to the already existing flavor only to remove "/" character, just like analyze-tracks does it.
```
lkiesow:
```suggestion
The behaviour of the `analyze-mediapackage` workflow operation has been changed. Instead of replacing every character that doesn't match `a-z` or `0-9` with an underscore character, the operation now only replaces the `/` separating flavor and subflavor. This makes it behave identical to the `analyze-tracks` operation.
```

docs/guides/admin/docs/workflowoperationhandlers/analyze-mediapackage-woh.md
```diff
@@ -46,11 +46,11 @@ Operation Example
 
 The operation will create workflow instance variables like this:
 
-```properties
+```
```
lkiesow, line 49: Why remove the syntax highlighting?

lkiesow (changes requested):
Generally looks good. Two small requests for the docs.

docs/guides/admin/docs/workflowoperationhandlers/analyze-mediapackage-woh.md
aliyevTU: "+" Doesnt work with properties style of codeblock. On official docs this doesnt make any difference anyway.

PR #4792, opened 2023-03-17
Fixed long WF description 11_to_12 DB upgrade script bug

aliyevTU:
11_to_12 DB Upgrade script fails, if there exists a WF with a description longer than 255 chars. Added a check for description field, and changed the script to cut description near 255 limit, and add three dots - "..."


aliyevTU:
It seems like 11_to_12 script has port implemented on OC branch 13.x, but not in 12.x. I used the script from 13.x so in my PR port implementation also exists. If it should be changed let me know.

aliyevTU:
> Hi @aliyevTU Thank you for contributing to Opencast. We noticed that you have not yet filed an [Individual Contributor License Agreement](https://www.apereo.org/licensing/agreements/icla). Doing that (once) helps us to ensure that Opencast stays free for all. If you make your contribution on behalf of an institution, you might also want to file a [Corporate Contributor License Agreement](https://www.apereo.org/licensing/agreements/ccla) (giving you as individual contributor a bit more security as well). It can take a while for this bot to find out about new filings, so if you just filed one or both of the above do not worry about this message! Please let us know if you have any questions regarding the CLA.

https://github.com/opencast/opencast/pull/4791#issuecomment-1472469073

aliyevTU:
I haven't tested PSQL script, should I also implement this for PSQL upgrade script?

gregorydlogan:
Yep, if we're doing one we should do both.

PR #4487, opened 2022-11-23
Seperate user and non-user roles in the access policy modal in various details views in the Admin UI

Arnei:
To make access policies a little more intuitive to Admin UI users, this commit changes the way roles are displayed in views where roles with additional actions can be set (e.g in the event details view). It splits the role selection section into two parts. The first part is for selecting "users" (i.e user roles), the second part is for more technical non-user roles.

Furthermore, additional roles are provided to allow users to see either of the two role selection sections, to e.g. hide the non-user roles to certain users.

Furthermore, additional list provider settings allow for customizing default read/write rights when adding a new role, and for making the read/write rights read-only.

Example image:
Bildschirmfoto vom 2022-11-23 10-51-15

Sponsored by the University of Bern.

This PR is going against r/12.x solely because it was developed against that branch. If it is determined to be too breaking for 12, I would have no qualms rebasing it on 13 or develop.

@naweidner could you check if the changes you made in #4308 are still working? I might have broken them while resolving merge conflicts. 

* [~] have a concise title
* [~] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4799, opened 2023-03-20
Sets values in specific config file to default

Arnei:
This sets the config values in the listproviders file `acl.default.create.properties` to their default values. This should already have happened in the PR associated with the files creation, but was simply overlooked. With these values set to their defaults, the ACL tab in the event and series details view in the Admin UI should function again.

### DIFF
--- a/etc/listproviders/acl.default.create.properties
+++ b/etc/listproviders/acl.default.create.properties
@@ -4,19 +4,19 @@ list.name=ACL.DEFAULTS
 
 # New access policy automatically have the read attribute set.
 # Default: true
-read_enabled = true
+#read_enabled = true
 
 # New access policy automatically have the write attribute set.
 # Default: false
-write_enabled = true
+#write_enabled = false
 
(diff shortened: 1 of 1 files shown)


lkiesow:
> With these values set to their defaults, the ACL tab in the event and series details view in the Admin UI should function again.

Can you briefly describe what the issue is and when it was introduced?

Arnei:
> > With these values set to their defaults, the ACL tab in the event and series details view in the Admin UI should function again.
> 
> Can you briefly describe what the issue is and when it was introduced?

Right. The issue was introduced with PR #4487. It prevented role assignment to users in role tab in the event and series details view of the admin ui, as described in this comment: https://github.com/opencast/opencast/pull/4487#issuecomment-1419632301

mtneug:
So both configs work fine for me in the sense that the configs do what is promised. Don't see anything "broken" just not what is the default for Opencast.

mtneug (approved):

PR #4783, opened 2023-03-15
Add paging to asset manager index rebuild

gregorydlogan:
This PR adds paging to the asset manager index rebuild to prevent loading *all* of the snapshots into ram at the same time.

Fixes #4748

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] ~include migration scripts and documentation, if appropriate~
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4843, opened 2023-04-04
Fixed pagination when reindexing asset manager

### LINKED ISSUE #4748: Slow query during asset manager reindex
**Describe the bug**

During the asset manager reindex, all snapshots are loaded in memory from the DB. This is done in a single query resulting in a very slow DB query (and possibly in a heap overflow). I think this should be the culprit:

https://github.com/opencast/opencast/blob/cf744c6632e94c47b9adbbb31c3d609986aa01f1/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java#L951

In our CockroachDB cluster, this is reported as this (simplified) query:

```sql
SELECT DISTINCT
       t0.id,
       t0.archival_date,
       t0.availability,
       t0.mediapackage_id,
       t0.mediapackage_xml,
       t0.organization_id,
       t0.owner,
       t0.series_id,
       t0.storage_id,
       t0.version
  FROM oc_assets_snapshot AS t0
 WHERE t0.version
       = (
      SELECT max(t1.version)
        FROM oc_assets_snapshot AS t1
       WHERE t1.mediapackage_id = t0.mediapackage_id
    )
```

**To Reproduce**

1. Ingest videos
2. Perform a reindex

**Expected behavior**

The asset manager reindex uses pagination.

**Server environment:**
- Installation method container
- Version 13.x

JamesUoM:
#4783 which attempted to fix #4748 had some issues.

- To get consistent results the DB query results need to be explicitly ordered for pagination to work.
- Calls to logging progress changed frequency after first page.
- Last page was never process 

This should fix those issues. 
Ordering on mpId might not be the most efficient, however the row id is not exposed by the asset query API.
The page size has been increased to reduce the number of expense requests with ordered results.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -165,7 +165,7 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
 
   private static final Logger logger = LoggerFactory.getLogger(AssetManagerImpl.class);
 
-  private static final int PAGE_SIZE = 100;
+  private static final int PAGE_SIZE = 1000;
 
   enum AdminRole {
     GLOBAL, ORGANIZATION, NONE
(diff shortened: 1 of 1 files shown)


JamesUoM:
NOTE this doesn’t speed up the overall indexing but reduces the resources needed to read the entire asset table in one go. The slow down in tests with 20k assets, H2 database, local disk were significant almost doubling the time from ~10m to ~20m. I think the ordering is an expensive DB constant for each request. I think increasing the page size should help as 100 is quite small.

JamesUoM:
Upping page size to 1000 reduced the indexing time to 11m30s which I think is reasonable

PR #4868, opened 2023-04-17
Extend configuration options of Amberscript integration

JulianKniephoff:
This adds some global and per-workflow options to the Amberscript documentation. The options should be more or less on par with https://amberscript.github.io/api-docs/#uploading-and-status now. (Except for `turnaroundTime` which I honestly don't understand. ¯\\\_(ツ)_/¯)

It also includes some cleanup, moves code around, nitpicks some stuff; there is more to do here, but not enough time, and in the end I had to **not** do a lot of (in my eyes) improvements in the interest of consistency. Anyway, I think this still improves a few things.

For reviewers: The commits can and should probably be reviewed in isolation. Commit messages might contain traces of further information. :wink: The improvements I mentioned above are in their own commits; the new/extended config stuff is mostly in the last one.

Note: Includes #4864.

Work sponsored by Amberscript themselves, by the way. :smile:

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
@@ -14,13 +14,13 @@ an audio or video file to be transcribed and captioned.
 Parameter Table
 ---------------
 
-| configuration keys    | description                                                                | default      | example            |
-|-----------------------|----------------------------------------------------------------------------|--------------|--------------------|
-| source-tag            | A tag selecting the audio or video file to be sent for translation/transcription. | -     | engage-download    |
-| source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription. | -  | \*/themed           |
-| jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow) | direct       | perfect            |
-| language              | The assumed language for transcription.                                    | en           | nl                 |
-| skip-if-flavor-exists | If this flavor already exists in the media package, skip this operation.   | captions/vtt | captions/timedtext |
+| configuration keys    | description                                                                          | default      | example            |
+|-----------------------|--------------------------------------------------------------------------------------|--------------|--------------------|
+| source-tag            | A tag selecting the audio or video file to be sent for translation/transcription.    | -            | engage-download    |
+| source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription. | -            | \*/themed          |
+| jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow).          | direct       | perfect            |
+| language              | The assumed language for transcription.                                              | en           | nl                 |
+| skip-if-flavor-exists | If this flavor already exists in the media package, skip this operation.             | captions/vtt | captions/timedtext |
 
 ### Supported Languages
 
--- a/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
@@ -19,7 +19,7 @@ Parameter Table
 | source-tag            | A tag selecting the audio or video file to be sent for translation/transcription. | -     | engage-download    |
 | source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription. | -  | \*/themed           |
 | jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow) | direct       | perfect            |
-| language              | The target language for transcription.                                     | en           | nl                 |
+| language              | The assumed language for transcription.                                    | en           | nl                 |
 | skip-if-flavor-exists | If this flavor already exists in the media package, skip this operation.   | captions/vtt | captions/timedtext |
 
 ### Supported Languages
--- a/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
@@ -14,18 +14,18 @@ an audio or video file to be transcribed and captioned.
 Parameter Table
 ---------------
 
-| configuration keys    | description                                                                              | default       | example                  |
-|-----------------------|------------------------------------------------------------------------------------------|---------------|--------------------------|
-| source-tag            | A tag selecting the audio or video file to be sent for translation/transcription.        | -             | engage-download          |
-| source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription.     | -             | \*/themed                |
-| jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow).              | direct        | perfect                  |
-| language              | The assumed language for transcription.                                                  | en            | nl                       |
-| speaker               | The number of speakers in the recording.                                                 | 1             | 2                        |
-| transcriptiontype     | transcription, captions or translatedSubtitles                                           | transcription | captions                 |
-| glossary              | An ID of a custom glossary to use. See https://amberscript.github.io/api-docs/#glossary. | -             | 643966c3f3e91e0c96e9e060 |
-| transcriptionstyle    | cleanread or verbatim                                                                    | cleanread     | verbatim                 |
-| targetlanguage        | The language of the resulting transcriptions.                                            | -             | en                       |
-| skip-if-flavor-exists | If this flavor already exists in the media package, skip this operation.                 | captions/vtt  | captions/timedtext       |
+| configuration keys    | description                                                                          | default       | example                  |
+|-----------------------|--------------------------------------------------------------------------------------|---------------|--------------------------|
+| source-tag            | A tag selecting the audio or video file to be sent for translation/transcription.    | -             | engage-download          |
+| source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription. | -             | \*/themed                |
+| jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow).          | direct        | perfect                  |
+| language              | The assumed language for transcription.                                              | en            | nl                       |
+| speaker               | The number of speakers in the recording.                                             | 1             | 2                        |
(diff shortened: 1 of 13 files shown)


modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -379,38 +442,77 @@ public Job startTranscription(String mpId, Track track, String... args) throws T
...
+        } catch (IOException | NotFoundException e) {
+          logger.error("Unable to load dublin core catalog for event '{}'",
+                  track.getMediaPackage().getIdentifier(), e);
         }
+      }
 
-      } catch (IOException | NotFoundException e) {
-        logger.error("Unable to load dublin core catalog for event '{}'",
-            track.getMediaPackage().getIdentifier(), e);
+      if (speakers.size() > 1) {
```
Arnei: Shouldn't this be `>=` to cover cases where the dublincore specifies `1` speaker? 

Code will still work in most cases, because `getNumberOfSpeakers()` returns `1` per default. But if `numberOfSpeakers` is set to something else, this might behave weirdly.

Arnei (commented):
Looks reasonable to me. Found one small thing, but nothing major. 
It is nice to see such a well structured PR with well commented commits :)

I am also really not a fan of this services "If configuration is invalid, fall back to default instead of throwing an error"-approach, but since it has been like this since forever I don't think it's a blocker.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
JulianKniephoff: That sounds right to me. :thinking: ~~I'ma push a fix in a sec.~~ I amended the last commit and force-pushed.

JulianKniephoff:
> I am also really not a fan of this services "If configuration is invalid, fall back to default instead of throwing an error"-approach, but since it has been like this since forever I don't think it's a blocker.

I know what you mean and I'm mostly with you. In this case it's necessary, though, since Amberscript want this in 13.x. Without the fallbacks, this would break existing installations. :slightly_frowning_face:

Arnei (approved):

etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
```diff
@@ -2,63 +2,96 @@
...
-# after the transcription job is completed.
-# (in seconds) Default is 1 minute.
+# after the transcription job is completed in seconds.
+# Default: 1 minute
 #workflow.dispatch.interval=60
 
 # How long to wait after a transcription is supposed to finish before marking the job as
-# canceled in the database. Default is 48 hours.
-# (in seconds)
+# canceled in the databas in seconds
```
gregorydlogan:
```suggestion
# cancelled in the database in seconds
```

etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
```diff
@@ -2,63 +2,96 @@
...
+
+# Default glossary to use.
+# You can override this within you workflow (see workflow operation handler `amberscript-start-transcription`).
+#glossary=
+
+# Default transcription style to use.
+# You can override this within you workflow (see workflow operation handler `amberscript-start-transcription`).
+# Can be one of: cleanread, verbatim
+# Default: cleanread
+#transcriptiontype=transcription
```
gregorydlogan: Duplicate key?

etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
```diff
@@ -2,63 +2,96 @@
...
+
+# Default transcription type to use.
+# You can override this within you workflow (see workflow operation handler `amberscript-start-transcription`).
+# Can be one of: transcription, captions, translatedSubtitles
+# Default: transcription
+#transcriptiontype=transcription
+
+# Default glossary to use.
+# You can override this within you workflow (see workflow operation handler `amberscript-start-transcription`).
+#glossary=
```
gregorydlogan, line 87: Missing default, and explanation of logic below where `"" != <empty>`

docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
```diff
@@ -14,13 +14,18 @@ an audio or video file to be transcribed and captioned.
...
+| source-tag            | A tag selecting the audio or video file to be sent for translation/transcription.        | -             | engage-download          |
+| source-flavor         | A flavor selecting the audio or video file to be sent for translation/transcription.     | -             | \*/themed                |
+| jobtype               | direct (automated, fast) or perfect (additional manual improvements, slow).              | direct        | perfect                  |
+| language              | The assumed language for transcription.                                                  | en            | nl                       |
+| speaker               | The number of speakers in the recording.                                                 | 1             | 2                        |
+| transcriptiontype     | transcription, captions or translatedSubtitles                                           | transcription | captions                 |
+| glossary              | An ID of a custom glossary to use. See https://amberscript.github.io/api-docs/#glossary. | -             | 643966c3f3e91e0c96e9e060 |
```
gregorydlogan: Behaviour of unset vs `""` should be documented here, and/or in the config.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -264,56 +303,78 @@ public void activate(ComponentContext cc) {
...
+            speakerMetadataFieldOpt.get(), SPEAKER_METADATA_FIELD, speakerMetadataField);
       }
     }
-    logger.info("Language code map was set. Added '{}' additional entries.", customMapEntriesCount);
+    logger.info("Default metadata field for calculating the amount of speakers is set to '{}'.", speakerMetadataField);
+
+    Option<String> transcriptionTypeOpt = OsgiUtil.getOptCfg(cc.getProperties(), TRANSCRIPTIONTYPE);
+    if (transcriptionTypeOpt.isSome()) {
+      transcriptionType = transcriptionTypeOpt.get();
+      logger.info("Default transcription type is set to '{}'.", transcriptionType);
```
gregorydlogan: You've got a list of valid types in the config file, shouldn't the input be checked against that?

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -264,56 +303,78 @@ public void activate(ComponentContext cc) {
...
+      glossary = glossaryOpt.get();
+      logger.info("Default glossary is set to '{}'.", glossary);
+    } else {
+      logger.info("No glossary will be used by default");
+    }
+
+    Option<String> transcriptionStyleOpt = OsgiUtil.getOptCfg(cc.getProperties(), TRANSCRIPTIONSTYLE);
+    if (transcriptionStyleOpt.isSome()) {
+      transcriptionStyle = transcriptionStyleOpt.get();
+      logger.info("Default transcription style is set to '{}'.", transcriptionStyle);
```
gregorydlogan: Likewise here, we should be checking the style against a valid list.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
```diff
@@ -264,56 +303,78 @@ public void activate(ComponentContext cc) {
...
+    } else {
+      logger.info("Default transcription style '{}' will be used.", transcriptionStyle);
+    }
+
+    Option<String> targetLanguageOpt = OsgiUtil.getOptCfg(cc.getProperties(), TARGETLANGUAGE);
+    if (targetLanguageOpt.isSome()) {
+      targetLanguage = targetLanguageOpt.get();
+      logger.info("Default target language is set to '{}'.", targetLanguage);
+    } else {
+      logger.info("Transcriptions won't be translated");
```
gregorydlogan, line 386: Hold on now, I read the docs to mean that we're looking for transcriptions in `targetLanguage`, not that Amberscript would be running the transcriptions through some kind of translation tool.  Which is it?

gregorydlogan (changes requested):
A few things to fix prior to merge.

etc/org.opencastproject.transcription.amberscript.AmberscriptTranscriptionService.cfg
JulianKniephoff: <del>Well there is no default, it's unset by default. The other point is fair, though; will add something. And maybe that will clear up the default situation as well. :thinking: </del>

Nevermind, I'm dumb. The default here is `""`, and this (i.e. the "global" setting) can't be unset. I will add the default, but the logic you mention only applies to the workflow parameter, where I will add a note in the docs.

modules/transcription-service-amberscript/src/main/java/org/opencastproject/transcription/amberscript/AmberscriptTranscriptionService.java
JulianKniephoff: `language` is the language Amberscript expects the audio to be in. `targetLanguage` is the language of the captions that you get. If they differ there will be some kind of translation AFAIU, yes. That's what I tried to clarify in 5d05741b1d1fd18c20582e8b5269a99af3491ca6. :thinking:

docs/guides/admin/docs/workflowoperationhandlers/amberscript-start-transcription-woh.md
JulianKniephoff: Properly explaining this would blow up the table; I will add a small section at the bottom, I hope that's okay.

PR #4878, opened 2023-04-20
Make Composite Ffmpeg Command Configurable

KatrinIhler:
If we want to use GPU's to encode composite videos, we need to change the part of the ffmpeg command that is currently being constructed by the workflow operation handler. This makes most of that configurable (except for the audio mapping), by making use of conditional encoding profile parameters similarly to the encode operation. It also adds a bunch of other variables that the workflow operation handler fills with information needed in the ffmpeg command.

The `composite` encoding profile makes full use of the new configuration options, while the `mp4-preview.dual` profile was simplified.

Additionally, this also reformats a table for aesthetic purposes only.


Arnei (approved):
Didn't test all the ways this could be used (in particular I did not test GPU encoding), but the default workflows at least still run as expected.
Also looks reasonable and is documented, so I think this can be approved :)

mliradelc:
Nice, let me try this, we use composite with our GPUs using the execute WoH, I will try with our encoding ladder.

mliradelc:
Looks good to me

mliradelc (approved):
Looks good for me

PR #4921, opened 2023-04-25
Fix REST docs login problem

lkiesow:
The REST docs of Opencast have always been publicly available, but the recently added new admin interface seems to have broken this. Going to the REST docs overview page and not being logged in causes an empty page to be displayed.

This patch fixes the issue by moving from using jQuery to implementing the whole functionality in pure JavaScript which isn't really any harder with modern JavaScript.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #4928, opened 2023-04-26
Fix rest docs forms

### LINKED ISSUE #4922: REST docs broken
The new admin interface/login page seems to have caused the removal of some JavaScript and CSS files essential for the REST docs which are now broken:

Screenshot from 2023-04-25 20-20-29

Arnei:
After Lars fixed one part in https://github.com/opencast/opencast/pull/4921, this PR fixes the other by correcting the path to script and css files.

~Only fixes things WHEN LOGGED IN. When not logged in, there is still a bug that prevents jquery from getting loaded properly.~
EDIT: Now also allows loading jquery when not logged in, by allowing anonymous access to the resource (same as for the other js and css files). 

Includes #4921.
~Does not yet completely resolve #4922.~
Edit: Now resolves #4922.

### DIFF
--- a/docs/checkstyle/eslintrc.js
+++ b/docs/checkstyle/eslintrc.js
@@ -5,7 +5,7 @@ module.exports = {
   },
   'extends': 'eslint:recommended',
   'parserOptions': {
-    'ecmaVersion': 2015
+    'ecmaVersion': 2017
   },
   'plugins': [
     'header'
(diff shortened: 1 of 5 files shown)


gregorydlogan (approved):
Awesome, fixed the whole thing.  TYVM.

lkiesow (approved):
Looks good

PR #4951, opened 2023-05-10
Allow opencast to share the oc-remember-me cookie between nodes

marwyg:
In a multi node scenario, if you configure a different cookie-key than 'opencast', the same cookie-key will be used for all nodes. Therefore, you can log into the admin node for example and open up a published video without logging into the presentation server again.

You also have to configure the subdomain for the cookie in your nginx configuration: https://docs.opencast.org/develop/admin/#configuration/https/nginx/

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -421,8 +421,11 @@
     <!-- All following settings are optional -->
     <property name="tokenValiditySeconds" value="1209600"/>
     <property name="cookieName" value="oc-remember-me"/>
-    <!-- The following key will be augmented by system properties. Thus, leaving this untouched is okay -->
-    <!-- This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below) -->
+    <!-- The following key will be augmented by system properties. Thus, leaving this untouched is okay.
+         This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below)
+         'Augmented' means, leaving this as 'opencast' will result in a random key. So if you log into the
+         admin server for example and switch to the presentation server, you will have to log in again.
+         The key won't be augmented/randomized if you use something different than 'opencast'. -->
     <property name="key" value="opencast"/>
   </bean>
 
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -421,11 +421,10 @@
     <!-- All following settings are optional -->
     <property name="tokenValiditySeconds" value="1209600"/>
     <property name="cookieName" value="oc-remember-me"/>
-    <!-- The following key will be augmented by system properties. Thus, leaving this untouched is okay.
-         This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below)
-         'Augmented' means, leaving this as 'opencast' will result in a random key. So if you log into the
-         admin server for example and switch to the presentation server, you will have to log in again.
-         The key won't be augmented/randomized if you use something different than 'opencast'. -->
+    <!-- The following key will be augmented by system properties if left at the default value.
+         Thus, leaving this untouched is okay. This key must be equal to the key passed to
+         rememberMeAuthenticationProvider (s. below). To generate cookies which are valid for the whole cluster,
+         set this manually. The key won't be augmented/randomized if you use something different than 'opencast'. -->
     <property name="key" value="opencast"/>
   </bean>
 
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
@@ -44,10 +44,19 @@ public final class SystemTokenRememberMeUtils {
 
   private static final Logger logger = LoggerFactory.getLogger(SystemTokenRememberMeUtils.class);
 
+  /** This is the default cookie key, that is configured in Opencast **/
+  private static final String defaultCookieKey = "opencast";
+
   private SystemTokenRememberMeUtils() {
   }
 
   public static String augmentKey(String key) {
+
+    if (!defaultCookieKey.equals(key)) {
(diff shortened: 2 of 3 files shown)


etc/security/mh_default_org.xml
```diff
@@ -421,8 +421,11 @@
     <!-- All following settings are optional -->
     <property name="tokenValiditySeconds" value="1209600"/>
     <property name="cookieName" value="oc-remember-me"/>
-    <!-- The following key will be augmented by system properties. Thus, leaving this untouched is okay -->
-    <!-- This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below) -->
+    <!-- The following key will be augmented by system properties. Thus, leaving this untouched is okay.
+         This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below)
+         'Augmented' means, leaving this as 'opencast' will result in a random key. So if you log into the
+         admin server for example and switch to the presentation server, you will have to log in again.
```
lkiesow: The augmentation is a problem for a domain-wide cookie, but it's not the configuration of such. Meaning that not augmenting the key is not sufficient for configuring a domain-wide cookie. Unfortunately, this documentation makes it look like it is.

I suggest removing this hint and just stating what it does:

```suggestion
    <!-- The following key will be augmented by system properties if left at the default value.
         Thus, leaving this untouched is okay.
         This key must be equal to the key passed to rememberMeAuthenticationProvider (s. below).
         To generate cookies which are valid for the whole custer, set this manually.
```

modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
```diff
@@ -44,10 +44,19 @@ public final class SystemTokenRememberMeUtils {
 
   private static final Logger logger = LoggerFactory.getLogger(SystemTokenRememberMeUtils.class);
 
+  /** This is the default cookie key, that is configured in Opencast **/
+  private static final String defaultCookieKey = "opencast";
```
lkiesow: As a class-wide constant I suggest using an uppercase name as usual:

```suggestion
  private static final String DEFAULT_COOKIE_KEY = "opencast";
```

modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
```diff
@@ -44,10 +44,19 @@ public final class SystemTokenRememberMeUtils {
...
+
   private SystemTokenRememberMeUtils() {
   }
 
   public static String augmentKey(String key) {
+
+    if (!defaultCookieKey.equals(key)) {
+      logger.debug("The default cookie key 'opencast' is not in use. The given key won't be augmented.");
```
lkiesow: Since `defaultCookieKey` exists, we shouldn't hard-code the value:

```suggestion
      logger.debug("The default cookie key '{}' is not in use. The given key won't be augmented.", defaultCookieKey);
```

lkiesow (changes requested):
Generally looks good. This is more or less nitpicking.

etc/security/mh_default_org.xml
marwyg: Done

modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
marwyg: Done

modules/kernel/src/main/java/org/opencastproject/kernel/security/SystemTokenRememberMeUtils.java
marwyg: Done

marwyg:
@lkiesow Thank you for the review, I implemented your suggestions.

mtneug:
Just as a note since I brought that subject up a couple of days ago in the Matrix chat: the remember me cookie is never set for Opencast installations that use SSO. I looked into sharing the actual sessions tracked by the `JSESSIONID` cookie. This is set and managed by Jetty, and there are some configs you can probably set using a Jetty config (OPS4j Pax Web does not directly expose all of the settings). This is called [persistent or clustered session](https://www.eclipse.org/jetty/documentation/jetty-9/index.html#session-management). I have not followed up on this since I fixed the root cause of the bug I faced, and having SSO creating a new session is done automatically, but maybe this is worth looking further into.

lkiesow:
Works :+1: 

With two servers `a.oc.local` and `b.oc.local` and an Nginx configuration containing:

```js
proxy_cookie_path / "/; HTTPOnly; Domain=oc.local";
```

I get a response like this

```http
❯ curl -I a.oc.local
HTTP/1.1 302 Found
Server: nginx/1.22.1
Date: Mon, 15 May 2023 18:04:54 GMT
Connection: keep-alive
Set-Cookie: JSESSIONID=node0vthrmsgnc1hb1xkc13v2448tv114.node0; Path=/; HTTPOnly; Domain=oc.local; HttpOnly
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Location: http://a.oc.local/index.html
```

Which cause me to be automatically logged into `b.oc.local` if the remember me option is set.

Also as expected, this only works if I manually set the remember me secret.

mtneug:
@lkiesow: you can set the cookie domain with Pax in your `org.ops4j.pax.web.cfg`:

```
org.ops4j.pax.web.session.domain=oc.local
```

marwyg:
@mtneug That would be wonderful! I looked for this option, but I tried `org.ops4j.pax.web.session.cookie.domain` instead, which didn't work. I wanted this to be in the PR as well. If I find the time, I will test and add this config.

mtneug:
@marwyg we use this in prod right now. If I remember correctly, this option was not documented in OPS4J Pax last time I looked. I think I had to look into the code.

aliyevTU:
This PR did complete behavioral change in our system. We use SSO.

PR #4955, opened 2023-05-15
Asset manager multiple storage

Arnei:
> The Problem
>
>The storage of Opencast systems can easily be split between the asset manager/archive and the distributions. But splitting the archive itself is extremely hard and tricky.
>
>Unfortunately, some storage systems have limitations on how much data they can store. This means that you may have a limitation on how large your system can get based on the underlying storage system.
>
>That is why it would be incredibly helpful if Opencast could use more than one storage location/storage system for storing the asset manager data.

> Solution: Tracking Media Packages
> For this, we don’t need any migration since we would keep the same storage path structure we have in the asset manager right now:
>
> `.../archive[1..N]/<organization>/<mediapackage>/<version>/...`
>
>But to allow for dynamically adding new storage systems, we would have Opencast keep track of which media package goes to which storage, building a map like this:
>
>```
># storage(<mediapackage>) -> location
>storage("ID1") -> ".../storage/a/"
>storage("ID2") -> ".../storage/b/"
>```
>
>This means that for adding new media packages, Opencast could automatically select the storage with the lowest usage and that users could add new storage to Opencast at any time.

(lkiesow, 2022)

This PR attempts to solve the problem mentioned above by implementing the solution mentioned above.

### DIFF
--- a/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
+++ b/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
@@ -47,6 +47,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
@@ -158,7 +159,7 @@ public class OsgiFileSystemAssetStore extends AbstractFileSystemAssetStore {
   }
 
   private List<String> getRootDirectories() {
-    return rootDirectories;
+    return Collections.unmodifiableList(rootDirectories);
   }
 
   protected void setupCache() {
--- a/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
+++ b/modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
@@ -146,6 +146,17 @@ public class OsgiFileSystemAssetStore extends AbstractFileSystemAssetStore {
     return null;
   }
 
+  protected void setupCache() {
+    cache = CacheBuilder.newBuilder().maximumSize(cacheSize).expireAfterWrite(cacheExpiration, TimeUnit.MINUTES)
+            .build(new CacheLoader<String, Object>() {
+              @Override
+              public Object load(String orgAndMpId) throws Exception {
+                String rootDirectory = getRootDirectoryForMediaPackage(orgAndMpId);
+                return rootDirectory == null ? null : rootDirectory;
+              }
+            });
+  }
+
   /**
    * OSGi DI.
    */
@@ -201,13 +212,6 @@ public class OsgiFileSystemAssetStore extends AbstractFileSystemAssetStore {
 
     // Setup rootDirectory cache
     // Remembers the root directory for a given mediapackage
-    cache = CacheBuilder.newBuilder().maximumSize(cacheSize).expireAfterWrite(cacheExpiration, TimeUnit.MINUTES)
-            .build(new CacheLoader<String, Object>() {
-              @Override
-              public Object load(String orgAndMpId) throws Exception {
-                String rootDirectory = getRootDirectoryForMediaPackage(orgAndMpId);
-                return rootDirectory == null ? null : rootDirectory;
-              }
-            });
+    setupCache();
   }
 }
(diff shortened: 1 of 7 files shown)


gregorydlogan:
I'm assigning myself here but if anyone else wants to be coassigned go ahead.

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
```diff
@@ -71,8 +92,84 @@ public class OsgiFileSystemAssetStore extends AbstractFileSystemAssetStore {
...
+      if (Files.exists(dirPath) && Files.isDirectory(dirPath)) {
+        return path;
+      }
+    }
+
+    return null;
+  }
+
+  private List<String> getRootDirectories() {
+    return rootDirectories;
```
gregorydlogan: At the risk of being annoying, shouldn't this be an immutable list?

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
```diff
@@ -103,6 +204,61 @@ public void activate(final ComponentContext cc) throws IllegalStateException, IO
...
+
+      if (directory != null) {
+        rootDirectories.add(directory);
+      } else {
+        isRootDirectory = false;
+      }
+      index++;
+    }
+    for (String directory: rootDirectories) {
+      mkDirs(file(directory));
```
gregorydlogan, line 238: Hoo boy, lotsa questions :D

- What if one or more defined directories
  - Cannot be created (permissions)
  - Are *secretly* a device (cf `/dev/null`)
  - Exist but aren't writable
- Should this 
  - Does it throw an error and die? (by far the simpliest)
  - Does it continue, but not use that storage target? (ideal)
- What if one or more defined directories are *subdirectories* of one (or more!) of the others

gregorydlogan (changes requested):
I like the idea here, and to me it reads cleanly.  I can see adopters wanting an operation to balance *current* mediapackages across their newly defined storage targets in OC 15, but that's probably outside the scope.  I have questions about the current implementation, see the review :)

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
Arnei: That makes sense.

modules/asset-manager-storage-fs/src/main/java/org/opencastproject/assetmanager/storage/impl/fs/OsgiFileSystemAssetStore.java
Arnei: Those ... are good questions D:

> Cannot be created (permissions)

That throws an exception

> Are secretly a device (cf /dev/null)

I don't know. And I'm not sure how to check for or handle that.

> Exist but aren't writable

That won't lead to an exception. Should/Can we check this?

> Should this [...]

Throw an error and die, no? Users probably want to know if they configured unusable storage. Or if their storage has become unusable for some reason.

> What if one or more defined directories are subdirectories of one (or more!) of the others

Oh no. Why would anyone do this. Please.  
I believe this actually should work fine for the most part, but we should probably disallow it anyway? To avoid things like someone deleting their storage and accidentally also deleting all their other storages in subdirectories

gregorydlogan: > > Are secretly a device (cf /dev/null)
> 
> I don't know. And I'm not sure how to check for or handle that.

See below? I guess this isn't really likely to happen, and at a certain point PEBKAC and there's nothing we can do... 

> > Exist but aren't writable
> 
> That won't lead to an exception. Should/Can we check this?

Maybe as part of startup we try and write something to each storage? It won't guard against storages going away during runtime, but it would ensure they start out being present at least. 

> > Should this [...]
> 
> Throw an error and die, no? Users probably want to know if they configured unusable storage. Or if their storage has become unusable for some reason.

Agree, especially for something as important as this. 

> > What if one or more defined directories are subdirectories of one (or more!) of the others
> 
> Oh no. Why would anyone do this. Please.  
> I believe this actually should work fine for the most part, but we should probably disallow it anyway? To avoid things like someone deleting their storage and accidentally also deleting all their other storages in subdirectories

Because they're ~~evil~~DevOps and made an oopsie with their ansible script. I would hope that no one would do this on purpose, but I can see it happening accidentally and then not getting noticed until things get _messy_.

gregorydlogan:
This, at minimum, doesn't break anything.  It also works for me (when I remember to actually mount the alternative storage directories...), so let's get this in.

Arnei:
Thanks for merging, but I've been informed yesterday that this (when actually used) might break the snapshot operation for older mediapackages. Currently looking into whether this is actually at fault, but seems likely so far :')

Arnei:
Further investigation revealed that not ALL older mediapackages were broken, just a number in the single digits. Whether that was do to this PR or some custom migration scripts remains unclear.

PR #4969, opened 2023-05-25
Automate PR comments with built tarball links

gregorydlogan:
This PR automates creation of a comment with both the build log, and a download link for all relevant PRs.  As the PR gets updated new builds will *update* the existing comment.  I think this is a reasonable approach, but we can just have it post a new comment if that's better for our uses.

See https://github.com/gregorydlogan/opencast/pull/176 for an example (although the 'tarball' is just junk data).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/comment-on-prs.yml
+++ b/.github/workflows/comment-on-prs.yml
@@ -1,0 +1,77 @@
+name: Comment on Pull request
+#Cribbed from https://github.com/orgs/community/discussions/51403#discussioncomment-5535167
+
+on:
+  workflow_run:
+    workflows: [Test Opencast]
+    types:
+      - completed
+
+jobs:
+  comment:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get Artifact and Pull request info
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          WORKFLOW_RUN_EVENT_OBJ: ${{ toJSON(github.event.workflow_run) }}
+          OWNER: ${{ github.repository_owner }}
+          REPO: ${{ github.event.repository.name }}
+        run: |
+          PREVIOUS_JOB_ID=$(jq -r '.id' <<< "$WORKFLOW_RUN_EVENT_OBJ")
+          echo "Previous Job ID: $PREVIOUS_JOB_ID"
+          echo "PREVIOUS_JOB_ID=$PREVIOUS_JOB_ID" >> "$GITHUB_ENV"
+          
+          SUITE_ID=$(jq -r '.check_suite_id' <<< "$WORKFLOW_RUN_EVENT_OBJ")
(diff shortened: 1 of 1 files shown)


.github/workflows/comment-on-prs.yml
```diff
@@ -0,0 +1,77 @@
...
+
+on:
+  workflow_run:
+    workflows: [Test Opencast]
+    types:
+      - completed
+
+jobs:
+  comment:
+    if: github.event.workflow_run.conclusion == 'success'
```
lkiesow: Does it make sense to limit this to the `opencast/opencast` repo?

.github/workflows/comment-on-prs.yml
```diff
@@ -0,0 +1,77 @@
...
+            --jq ".artifacts.[] |
+            select(.workflow_run.id==${PREVIOUS_JOB_ID}) |
+            select(.expired==false) |
+            .id")
+          
+          echo "Artifact ID: $ARTIFACT_ID"
+          echo "ARTIFACT_ID=$ARTIFACT_ID" >> "$GITHUB_ENV"
+          
+          PR_NUMBER=$(jq -r '.pull_requests[0].number' \
+            <<< "$WORKFLOW_RUN_EVENT_OBJ")
```
lkiesow, line 50: Does this fail if the workflow `Test Opencast` was triggered by a `push` event?

.github/workflows/comment-on-prs.yml
```diff
@@ -0,0 +1,77 @@
...
+          
+          echo "Head SHA: $HEAD_SHA"
+          echo "HEAD_SHA=$HEAD_SHA" >> "$GITHUB_ENV"
+      - name: Find Comment
+        uses: peter-evans/find-comment@v2
+        id: find-comment
+        with:
+          issue-number: ${{ env.PR_NUMBER }}
+          comment-author: 'github-actions[bot]'
```
lkiesow, line 66: Given this condition, this would probably also find (and overwrite) the comment about missing ICLAs, wouldn't it?

lkiesow (changes requested):

.github/workflows/comment-on-prs.yml
gregorydlogan: Yep, probably.  The `find-comment` action has other filtering params, will add.

gregorydlogan: Fixed!

.github/workflows/comment-on-prs.yml
gregorydlogan: Fixed!

.github/workflows/comment-on-prs.yml
gregorydlogan: Good lord does GHA make this difficult.  Finally did get this done.

PR #4972, opened 2023-05-30
Fix directory cleanup for symlinks

marwyg:
This PR fixes the directory cleanup regarding symbolic links. Currently, if your workspace (for example) is a symlink, the cleanup process will delete it, instead of iterating through the files within the directory it's referencing.

To test this, you can set a lower cleanup period in your custom.properties (like 30 seconds, for example) and the value for the maximum age of a file (5 seconds, for example).

In the following example, the period and maximum age are already set to a lower value:
```
######### Workspace Cleanup #########

# The scheduled period in seconds, at which a workspace cleanup operation is performed.
# 86400 seconds equals 24 hours.
# Default value: -1 (Disable cleanup scheduler)
org.opencastproject.workspace.cleanup.period=30

# The maximum age a file must reach in seconds before a deletion of the file in the workspace cleanup operation is
# performed. 2592000 seconds equals 30 days.
# Default value: -1 (max age will never be reached)
org.opencastproject.workspace.cleanup.max.age=5
```

In the second step, you have to delete your workspace directory (or any directory the cleanup service is cleaning up)  and create a symlink instead. Absolute paths as well as relative paths should work.

### DIFF
--- a/modules/common/src/main/java/org/opencastproject/cleanup/RecursiveDirectoryCleaner.java
+++ b/modules/common/src/main/java/org/opencastproject/cleanup/RecursiveDirectoryCleaner.java
@@ -76,7 +76,7 @@ public final class RecursiveDirectoryCleaner implements FileVisitor<Path> {
   @Override
   public FileVisitResult visitFileFailed(Path path, IOException e) throws IOException {
     if (!(e instanceof NoSuchFileException)) {
-      logger.warn("Visiting file {} failed with exception {}", path, e);
+      logger.warn("Visiting file {} failed", path, e);
     }
     return FileVisitResult.CONTINUE;
   }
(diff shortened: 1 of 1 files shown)


modules/common/src/main/java/org/opencastproject/cleanup/RecursiveDirectoryCleaner.java
```diff
@@ -95,14 +95,20 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
   }
 
   public static boolean cleanDirectory(Path startingDir, Duration duration) {
-    if (!Files.exists(startingDir)) {
```
lkiesow: This seems a very complex approach. Why not just add as the first line in this method something like:

```java
var absStartDir = new File(startingDir).getCanonicalPath();
```

The `Files::getCanonicalPath` method will resolve symlinks if necessary or just return a path otherwise.
That's much simpler.

Example:

```
❯ ll
total 4.0K
drwxr-xr-x. 2 lars lars 4.0K May 30 16:55 a
lrwxrwxrwx. 1 lars lars    1 May 30 16:55 b -> a

❯ jshell 
jshell> new File("a").getCanonicalPath();
$1 ==> "/home/lars/a/a"

jshell> new File("b").getCanonicalPath();
$2 ==> "/home/lars/a/a"
```

marwyg: Oh man, thank you, that's clearly simpler. I have to practice my Google-Fu 😂 
I force pushed the new approach.

mtneug:
What is the use case for this? Can this be a potential security risk?

marwyg:
Use case: You create a symlink to your "true" workspace that is located on a NFS share. That's the situation, I saw on a production system. 

There are other ways to solve this, but this seemed like a bug. Regarding security, if you point the symlink to another directory, the cleanup would delete the files in it. But now you also could edit the custom.properties and set another directory for the cleanup(?) If someone sets the permissions of the custom.properties and the symlink correctly, it should be okay, I guess(?)

PR #4973, opened 2023-05-31
Add WhisperC++ engine to speech-to-text-impl

majosch:
> ~~Caveat: This PR is not ready to merge and needs to be debugged - command execution still fails - see this [post](https://groups.google.com/a/opencast.org/g/users/c/AzJYC0QhOz0/m/Pk3QujakAQAJ) in Opencast users group.~~ As suggested by @mliradelc I opened this PR for easier reviewing.

This PR adds a WhisperC++ engine to the speech-to-text-module.

[WhisperC++](https://github.com/ggerganov/whisper.cpp) is a C/C++ implementation of OpenAI's Whisper automatic speech recognition (ASR) model.

Many thanks to @mliradelc and @geichelberger for any help so far!

### DIFF
--- a/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
+++ b/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
@@ -1,0 +1,71 @@
+WhisperC++ Transcription Engine
+===============================
+
+Overview
+--------
+
+Opencast can take advantage of [Open AI's Whisper Project](https://github.com/openai/whisper) to generate automatic
+transcriptions on premise through [SpeechToText WoH](../../workflowoperationhandlers/speechtotext-woh.md).
+
+WhisperC++ is a C/C++ implementation of OpenAI's Whisper automatic speech recognition (ASR) model.
+
+Advantages
+----------
+- Transcription on more than 80 languages
+- Translation to English
+- Automatic language detection
+- Faster processing than whisper when using CPUs
+- Run locally, no data sent to third parties.
+
+Enable WhisperC++ engine
+------------------------
+
+To enable WhisperC++ as for the `SpeechToText` WoH, follow these steps.
+
+1. Install whispercpp binary and language models following the instruction on
+[whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or build and use your own RPMs for RHEL-based
+distributions from [whispercpp-rpmbuild](https://github.com/elearning-univie/whispercpp-rpmbuild).
+2. Enable whispercpp engine and set job load in `org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg`.
+3. Set the target model to use in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine`.
+4. WhisperC++ processes only PCM16 (wav) audio files. Therefore you probably have to add an `encode`-operation before
+running `speechtotext` in your workflow and an encoder profile:
+
+```
+  - id: encode
+    description: "Extract audio for processing with whispercpp"
+    configurations:
+      - source-flavor: "*/themed"
+      - target-flavor: "*/audio+stt"
+      - encoding-profile: audio-whisper
+
+  - id: speechtotext
+    description: "Generates subtitles with whispercpp"
+    configurations:
+      - source-flavor: "*/audio+stt"
+      - target-flavor: captions/vtt+auto-#{lang}
+      - target-element: attachment
+      - target-tags: archive,subtitle,engage-download
+      # - translate: true
+```
+
+```
+# Audio-only encoding format used for whispercpp
+profile.audio-whisper.name = whispercpp wav
+profile.audio-whisper.input = audio
+profile.audio-whisper.output = audio
+profile.audio-whisper.suffix = -stt.wav
+profile.audio-whisper.ffmpeg.command = -i #{in.video.path} -vn -ar 16000 -ac 1 -c:a pcm_s16le #{out.dir}/#{out.name}#{out.suffix}
+profile.audio-whisper.jobload = 1.0
+```
+
+
+Additional Notes
+----------------
+
+- Considerations about model usage and job load configuration in the [Additional notes](whisper.md#additional-notes)
+  from the `Whisper Transcription Engine` are also valid for this engine.
+- WhisperC++ can be run on CPU only or (partially) on dedicated hardware (Nvidia GPU, Apple Core ML) - if whispercpp
+  binary/library is built with appropriate options/modules.
+- Language models are based on Whisper and are transformed to a custom ggml format. You can download already converted
+  models or build and tweak your own models. For more informations visit
+  [models/README.md](https://github.com/ggerganov/whisper.cpp/blob/master/models/README.md).
--- a/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
+++ b/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
@@ -26,7 +26,7 @@ To enable WhisperC++ as for the `SpeechToText` WoH, follow these steps.
 [whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or build and use your own RPMs for RHEL-based
 distributions from [whispercpp-rpmbuild](https://github.com/elearning-univie/whispercpp-rpmbuild).
 2. Enable whispercpp engine and set job load in `org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg`.
-3. Set the target model to use in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine`.
+3. Set the binary and target model path to use in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine`.
 4. WhisperC++ processes only PCM16 (wav) audio files. Therefore you probably have to add an `encode`-operation before
 running `speechtotext` in your workflow and an encoder profile:
 
@@ -36,26 +36,31 @@ running `speechtotext` in your workflow and an encoder profile:
     configurations:
       - source-flavor: "*/themed"
       - target-flavor: "*/audio+stt"
-      - encoding-profile: audio-whisper
+      - encoding-profile: audio-whispercpp
 
   - id: speechtotext
     description: "Generates subtitles with whispercpp"
     configurations:
       - source-flavor: "*/audio+stt"
-      - target-flavor: captions/vtt+auto-#{lang}
-      - target-element: attachment
-      - target-tags: archive,subtitle,engage-download
+      - target-flavor: "captions/delivery"
+      - target-element: track
+      - target-tags: >-
+          archive,
+          engage-download,
+          generator-type:auto,
+          generator:whispercpp,
+          type:subtitle
       # - translate: true
 ```
 
 ```
 # Audio-only encoding format used for whispercpp
-profile.audio-whisper.name = whispercpp wav
-profile.audio-whisper.input = audio
-profile.audio-whisper.output = audio
-profile.audio-whisper.suffix = -stt.wav
-profile.audio-whisper.ffmpeg.command = -i #{in.video.path} -vn -ar 16000 -ac 1 -c:a pcm_s16le #{out.dir}/#{out.name}#{out.suffix}
-profile.audio-whisper.jobload = 1.0
+profile.audio-whispercpp.name = whispercpp wav
+profile.audio-whispercpp.input = audio
+profile.audio-whispercpp.output = audio
+profile.audio-whispercpp.suffix = -stt.wav
+profile.audio-whispercpp.ffmpeg.command = -i #{in.video.path} -vn -ar 16000 -ac 1 -c:a pcm_s16le #{out.dir}/#{out.name}#{out.suffix}
+profile.audio-whispercpp.jobload = 1.0
 ```
 
 
--- a/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
+++ b/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
@@ -46,10 +46,7 @@ running `speechtotext` in your workflow and an encoder profile:
       - target-element: track
       - target-tags: >-
           archive,
-          engage-download,
-          generator-type:auto,
-          generator:whispercpp,
-          type:subtitle
+          engage-download
       # - translate: true
 ```
 
--- a/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
+++ b/docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
@@ -23,8 +23,8 @@ Enable WhisperC++ engine
 To enable WhisperC++ as for the `SpeechToText` WoH, follow these steps.
 
 1. Install whispercpp binary and language models following the instruction on
-[whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or build and use your own RPMs for RHEL-based
-distributions from [whispercpp-rpmbuild](https://github.com/elearning-univie/whispercpp-rpmbuild).
+[whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or install the `whisper.cpp` package from the
+Opencast repository.
 2. Enable whispercpp engine and set job load in `org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg`.
 3. Set the binary and target model path to use in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine`.
 4. WhisperC++ processes only PCM16 (wav) audio files. Therefore you probably have to add an `encode`-operation before
@@ -34,7 +34,7 @@ running `speechtotext` in your workflow and an encoder profile:
   - id: encode
     description: "Extract audio for processing with whispercpp"
     configurations:
-      - source-flavor: "*/themed"
+      - source-flavor: "*/source"
       - target-flavor: "*/audio+stt"
       - encoding-profile: audio-whispercpp
 
@@ -47,14 +47,13 @@ running `speechtotext` in your workflow and an encoder profile:
       - target-tags: >-
           archive,
           engage-download
-      # - translate: true
 ```
 
 ```
 # Audio-only encoding format used for whispercpp
 profile.audio-whispercpp.name = whispercpp wav
-profile.audio-whispercpp.input = audio
-profile.audio-whispercpp.output = audio
+profile.audio-whispercpp.input = stream
+profile.audio-whispercpp.output = stream
 profile.audio-whispercpp.suffix = -stt.wav
 profile.audio-whispercpp.ffmpeg.command = -i #{in.video.path} -vn -ar 16000 -ac 1 -c:a pcm_s16le #{out.dir}/#{out.name}#{out.suffix}
 profile.audio-whispercpp.jobload = 1.0
@@ -68,6 +67,6 @@ Additional Notes
   from the `Whisper Transcription Engine` are also valid for this engine.
 - WhisperC++ can be run on CPU only or (partially) on dedicated hardware (Nvidia GPU, Apple Core ML) - if whispercpp
   binary/library is built with appropriate options/modules.
-- Language models are based on Whisper and are transformed to a custom ggml format. You can download already converted
-  models or build and tweak your own models. For more informations visit
+- Language models are based on Whisper and are transformed to a custom ggml format. You can install them from Opencast
+  repository, download already converted models or build and tweak your own models. For more informations visit
   [models/README.md](https://github.com/ggerganov/whisper.cpp/blob/master/models/README.md).
--- a/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
@@ -31,7 +31,7 @@ Parameter Table
 Requirements
 ------------
 
-In order for it to work, you have to install the vosk-cli or whisper package
+In order for it to work, you have to install the vosk-cli, whisper or whispercpp package.
 
 
 Operation Examples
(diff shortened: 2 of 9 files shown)


KatrinIhler:
Is this ready yet? If not, this should probably be a draft.

mliradelc:
It was ready, but I had problems testing it, at the end Ctranslate2 PR arrived and overwrited some things regarding this. The autor needs to rebase.

majosch:
Thanks to @sulyoklev for rebasing the PR.

@KatrinIhler PR should be ready for review now.

Works fine on our test installation (15-SNAPSHOT).

If someone else wants to test:
- static binary: [whispercpp.zip](https://github.com/opencast/opencast/files/13112158/whispercpp.zip)
- tiny model: https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-tiny.bin

mliradelc:
Hi, I started today to review this PR, but the dependencies are playing against me 😅 . At least I've got anaconda started to compile the coreml models but another missing dependency started to make problems 😓 .

lkiesow:
Hi @majosch, I did a few tests and did a rebase of your code. You will find the rebase at https://github.com/lkiesow/opencast/tree/whisper-cpp
Can you update this pull request? I'll then do a proper review and we can try getting this in.

lkiesow:
Thanks for updating this, @majosch. @mwuttke tested your patch from yesterday and found some issues. If you throw in long recordings, the whole process can hang indefinitely. I pushed a commit to https://github.com/lkiesow/opencast/tree/whisper-cpp (specifically https://github.com/lkiesow/opencast/commit/7be1348fe8cfdaeeb7f1dd74a7f6d6629bbe34d5) which should fix the issue.

lkiesow:
Looking at the failed tests, you are also missing updates to your docs which I already included in my rebase at https://github.com/lkiesow/opencast/commit/5ea4aadef1ad538c56415ade87beefa7dbc62b1c. Not aure how you updated your branch, but you will need something similar.

majosch:
@lkiesow Thank you for looking into this PR and for the fix. Also thanks to @mwuttke for testing.

I just updated r/15.x on our Opencast fork and did a `git rebase r/15.x" and fixed the merge conflicts - so, if I made no mistake, links should hopefulle already be correct.

Would you prefer updating my branch with yours, Lars?

majosch:
@lkiesow I just push your branch.

docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
```diff
@@ -0,0 +1,73 @@
...
+- Run locally, no data sent to third parties.
+
+Enable WhisperC++ engine
+------------------------
+
+To enable WhisperC++ as for the `SpeechToText` WoH, follow these steps.
+
+1. Install whispercpp binary and language models following the instruction on
+[whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or build and use your own RPMs for RHEL-based
+distributions from [whispercpp-rpmbuild](https://github.com/elearning-univie/whispercpp-rpmbuild).
```
lkiesow:
```suggestion
[whisper.cpp](https://github.com/ggerganov/whisper.cpp) repository or install the `whisper.cpp` package from the Opencast repository.
```

docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
```diff
@@ -0,0 +1,73 @@
...
+2. Enable whispercpp engine and set job load in `org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg`.
+3. Set the binary and target model path to use in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine`.
+4. WhisperC++ processes only PCM16 (wav) audio files. Therefore you probably have to add an `encode`-operation before
+running `speechtotext` in your workflow and an encoder profile:
+
+```
+  - id: encode
+    description: "Extract audio for processing with whispercpp"
+    configurations:
+      - source-flavor: "*/themed"
```
lkiesow: Most of the time, you probably want to operate on `*/source`:
```suggestion
      - source-flavor: "*/source"
```

docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
```diff
@@ -0,0 +1,73 @@
...
+  - id: speechtotext
+    description: "Generates subtitles with whispercpp"
+    configurations:
+      - source-flavor: "*/audio+stt"
+      - target-flavor: "captions/delivery"
+      - target-element: track
+      - target-tags: >-
+          archive,
+          engage-download
+      # - translate: true
```
lkiesow:
```suggestion
```

docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+```
+# Audio-only encoding format used for whispercpp
+profile.audio-whispercpp.name = whispercpp wav
+profile.audio-whispercpp.input = audio
+profile.audio-whispercpp.output = audio
```
lkiesow: iirc, this won't work with a video as input.
I think this should be:
```suggestion
profile.audio-whispercpp.input = stream
profile.audio-whispercpp.output = stream
```

docs/guides/admin/docs/configuration/transcription.configuration/whispercpp.md
```diff
@@ -0,0 +1,73 @@
...
+Additional Notes
+----------------
+
+- Considerations about model usage and job load configuration in the [Additional notes](whisper.md#additional-notes)
+  from the `Whisper Transcription Engine` are also valid for this engine.
+- WhisperC++ can be run on CPU only or (partially) on dedicated hardware (Nvidia GPU, Apple Core ML) - if whispercpp
+  binary/library is built with appropriate options/modules.
+- Language models are based on Whisper and are transformed to a custom ggml format. You can download already converted
+  models or build and tweak your own models. For more informations visit
+  [models/README.md](https://github.com/ggerganov/whisper.cpp/blob/master/models/README.md).
```
lkiesow, line 72: Note to self: Add a point about packages and the packaged version of the download script

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
```diff
@@ -6,23 +6,25 @@ ID: `speechtotext`
 Description
 -----------
 
-The speech to text operation can be used to generate subtitles for Videos or Audio files. Currently, there are two STT
-engines available, [Whisper](../configuration/transcription.configuration/whisper.md) and 
-[Vosk](../configuration/transcription.configuration/vosk.md). The subtitles file format ist WebVTT.
+The speech to text operation can be used to generate subtitles for Videos or Audio files. Currently, there are three STT
```
lkiesow:
```suggestion
The speech to text operation can be used to generate subtitles from video or audio files. Currently, there are three STT
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+    whispercppNoFallback = OsgiUtil.getOptCfgAsBoolean(cc.getProperties(), WHISPERCPP_NO_FALLBACK_CONFIG_KEY);
+    if (whispercppNoFallback.isSome()) {
+      logger.debug("WhisperC++ no fallback set to {}", whispercppNoFallback);
+    }
+
+    logger.debug("Finished activating/updating speech-to-text service");
+  }
+
+  //TODO: Add method for language detection
+
```
lkiesow: Do it or remove this ;-)
```suggestion
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.speechtotext.api.SpeechToTextEngine#generateSubtitlesFile(File, File, String, Boolean)
+   */
+
+  @Override
```
lkiesow:
```suggestion
   */
  @Override
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+  public Map<String, Object> generateSubtitlesFile(File mediaFile,
+      File preparedOutputFile, String language, Boolean translate)
+          throws SpeechToTextEngineException {
+
+    if (!FilenameUtils.isExtension(mediaFile.getPath(), "wav")) {
```
lkiesow: Not a big deal, but maybe to be a bit more flexible:
```suggestion
    if (!FilenameUtils.isExtension(mediaFile.getPath(), "wav")) {
    if (!mediaFile.getPath().toLowerCase().endsWith(".wav")) {
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    String[] baseCommands = { whispercppExecutable,
+        mediaFile.getAbsolutePath(),
+        "--model", whispercppModel,
+        "-ovtt",
+        "-oj",
+        "--output-file", preparedOutputFile.getAbsolutePath().replaceFirst("[.][^.]+$", "")};
+
+    List<String> command = new ArrayList<>(Arrays.asList(baseCommands));
```
lkiesow: Instead of creating an array, converting that to a list which is then used to initialize an `ArrayList` below, you could directly work with a list. Something like:
```suggestion
    List<String> baseCommand = List.of(whispercppExecutable,
        mediaFile.getAbsolutePath(),
        "--model", whispercppModel,
        "-ovtt",
        "-oj",
        "--output-file", preparedOutputFile.getAbsolutePath().replaceFirst("[.][^.]+$", ""));

    List<String> command = new ArrayList<>(baseCommand);
```

lkiesow: Or even:
```suggestion
    var command = new ArrayList<String>(List.of(
        whispercppExecutable,
        mediaFile.getAbsolutePath(),
        "--model", whispercppModel,
        "-ovtt",
        "-oj",
        "--output-file", preparedOutputFile.getAbsolutePath().replaceFirst("[.][^.]+$", "")));
```

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+    if (whispercppSplitOnWord.isSome() && whispercppSplitOnWord.get()) {
+      command.add("-sow");
+    }
+    if (whispercppBestOf.isSome()) {
+      command.add("-bo");
+      command.add(Integer.toString(whispercppBestOf.get()));
+    }
+    if (whispercppWordThreshold.isSome()) {
+      command.add("-wt");
+      command.add(String.format("%f", whispercppWordThreshold.get()));
```
lkiesow: This depends on the system's language settings and may break.
If the system language is English, this will result in something like `1.1`:

```
❯ echo $LANG
en_US.UTF-8

❯ jshell    
jshell> String.format("%f", 1.1)
$1 ==> "1.100000"
```

But switching to German, for example, this will be `1,1` which I assume will cause a problem?
```
❯ export LANG=de_DE.UTF-8

❯ jshell                 
jshell> String.format("%f", 1.1)
$1 ==> "1,100000"
```

This is used at more places down below.

You are also converting the configuration values to numbers, only to convert them back here. You could just never convert them.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+
+    // set language of the source audio if known
+    if (!language.isBlank()) {
+      logger.info("Using language {} from workflows", language);
+      command.add("--language");
+      command.add(language);
+    } else {
+      logger.debug("Auto-detecting language");
+      command.add("--language");
+      command.add("auto");
```
lkiesow, line 334: I think this is superfluous since auto-detection is the default anyway. There is no need to pass on this parameter, is there?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
```diff
@@ -0,0 +1,417 @@
...
+    } finally {
+      IoSupport.closeQuietly(process);
+    }
+
+    // Detect language if not set
+    if (subtitleLanguage.isBlank()) {
+      JSONParser jsonParser = new JSONParser();
+      try {
+        FileReader reader = new FileReader(preparedOutputFile.getAbsolutePath().replaceFirst("[.][^.]+$", "")
+            + ".json");
```
lkiesow: I'm writing this down here since this operated on the JSON file, although this is a more general problem. This code will cause Whisper to generate a JSON file which is never cleaned up. The changes made necessary by pull request #5472 will also fix this and you wouldn't need to clean up every file, but waiting for that PR also means some minor changes to this methods interface.

lkiesow (changes requested):
Sorry for the delay. I finally found time to do a proper review of this.
This generally looks good. I just found a few minor issues.

Depending on pull request #5472 being merged before/after this one of the pull requests needs to fix some conflicts as well.

Some of the comments are just suggestions. Feel free to argue against any of these changes ;-)

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
sulyoklev: Actually no, the default in whispercpp is English, so it's needed for auto detection

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
sulyoklev: You're right. Converting the configuration values to numbers, although somewhat unnecessary, is useful to prevent unwanted behavior when incompatible config values are passed to whispercpp.  This way it is ensured that whispercpp will behave as expected. I will fix the format issue.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
sulyoklev: I added the cleanup, so this PR can be merged anytime before or after #5472

gregorydlogan:
Trying to take a look at this, but I'm seeing compilation issues:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project opencast-speech-to-text-impl: Compilation failure: Compilation failure: 
[ERROR] /home/greg/opencast/upstream/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java:[62,8] org.opencastproject.speechtotext.impl.engine.WhisperCppEngine is not abstract and does not override abstract method generateSubtitlesFile(java.io.File,java.io.File,java.lang.String,java.lang.Boolean) in org.opencastproject.speechtotext.api.SpeechToTextEngine
[ERROR] /home/greg/opencast/upstream/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java:[260,30] generateSubtitlesFile(java.io.File,java.io.File,java.lang.String,java.lang.Boolean) in org.opencastproject.speechtotext.impl.engine.WhisperCppEngine cannot implement generateSubtitlesFile(java.io.File,java.io.File,java.lang.String,java.lang.Boolean) in org.opencastproject.speechtotext.api.SpeechToTextEngine
[ERROR]   return type java.util.Map<java.lang.String,java.lang.Object> is not compatible with org.opencastproject.speechtotext.api.SpeechToTextEngine.Result
[ERROR] /home/greg/opencast/upstream/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java:[259,3] method does not override or implement a method from a supertype

```

sulyoklev:
Should compile now, I've implemented the changes in #5472.

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
```diff
@@ -3,5 +3,5 @@
 #job.load.speechtotext=0.8
 
 # Select STT engine.
-# Available engines: vosk, whisper | Default: vosk
-SpeechToTextEngine.target=(enginetype=whisper)
+# Available engines: vosk, whisper, whispercpp | Default: vosk
+#SpeechToTextEngine.target=(enginetype=vosk)
```
gregorydlogan: So, this config doesn't actually *do* anything, since `STTSImpl` does not parse it, or change its hardcoded (vosk) binding.

What this does do is cause `STTSImpl` to hang waiting for its bindings to finish :)

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
gregorydlogan: So my setup is defaulting to 
` --language eng`
which spits out 
```
whisper_lang_id: unknown language 'eng'
error: unknown language 'eng'
```

By default, if I did not set the language in the workflow I would get `eng`, rather than `english`, or `auto`.

gregorydlogan (commented):
In testing this works well, aside from my comments.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
sulyoklev: It seems the workflow defaults to using `eng` if the language code isn't given, eventhough the speechtotext-woh documentation says that language codes are the two latter codes defined in ISO 639-1. Since whispercpp only accepts either the two latter codes (e.g. `en`) or the full lowercase names (e.g. `english`) it cannot recognize `eng` and thus throws an error.

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
sulyoklev: You're right, I pushed a fix

etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
```diff
@@ -4,4 +4,4 @@
 
 # Select STT engine.
 # Available engines: vosk, whisper, whispercpp | Default: vosk
-#SpeechToTextEngine.target=(enginetype=vosk)
+SpeechToTextEngine.target=(enginetype=whisper)
```
gregorydlogan, line 7: A few of issues here:
- You meant `whispercpp` here I'm assuming?
- This changes the default config, and thus can't go into `r/15.x`
- This config change still does not work how you want it to.

`SpeechToTextServiceImpl` does not read this config and thus can't actually change its binding filter.  You want to add logic to its `activate` method, where both the `@activate` and `@modified` annotations live.  And probably rip out/replace `setSpeechToTextEngine` in the same class, since it's where the hardcoded binding lives.

majosch: This PR does NOT intend to change the default SpeechToText-Engine - the changed line just reflects the changes from 14 to 15.

gregorydlogan: I think it's probably https://github.com/opencast/opencast/pull/4973/commits/014827467f70d8a650cdefc6063dc2327fea8d6f that's causing the problem.  I saw that listed in this PR, and assumed it was an attempt to resolve the binding issue from the commit log entry.  Your commit seems to duplicate 0e2c9ba3086a2d6093a5ed3f28709cfb149de1e3 (which probably shouldn't have touche this file, but I digress), which is already in r/15.x

mwuttke:
I have successfully tested the current status of this patch with the Opencast Version 15.2 together with the language packs that @gregorydlogan has built.

Would it be possible to merge the patch quickly so that it is included in the next release?

Or what else speaks against a merge?

gregorydlogan:
@mwuttke Did you have to modify the patch at all?  Notably, `etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg`'s target config does not seem to work for me.

Otherwise I think this is reasonable, and I suspect that the config item I'm talking about it just straight up *broken* for everything so I don't see that preventing this from being merged.  @lkiesow short notice, but any objections?

mwuttke:
I have only changed the enginetype in `/etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg` to whispercpp. That was all. After that, the transcription ran without errors.

gregorydlogan:
Retested, and the engine config does actually work.

I did see a failure when the language was not set for my workflow but that easily cleared up, a default can be set, and this engine isn't even active by default anyway.  Merging.

PR #5023, opened 2023-06-05
Remove Track Selection Restrictions from the Old Editor

KatrinIhler:
Currently you can only save/start a workflow in the old video editor if the following requirements are fulfilled:

1. You have selected at least one video stream, if one exists
2. You have selected at least one audio stream, if one exists
3. You have not selected more audio than video stream

This is, in my opinion, too restrictive and not really helpful. It might be useful to publish a video without an audio, or audio without a video, and the latter use case is actually frequently requested. Additionally, why can I only select two audio streams if I also have two video streams?

In an ideal world, this would be configurable. But giving that the old editor is nearing the end of its life span, that doesn't seem worth the effort. So I suggest that we drop these restrictions altogether, and allow people to save/publish whatever they want, as long as they select _at least one stream_.  The user will still be warned about unusual selections via notifications, but if they decide to proceed, they now can.

---

I made the following changes to the notifications in the track selection tab:

1. Demote current errors to warnings
2. Add new error when no stream, audio or video, is selected
3. Don't show warnings when error is shown
4. Change wording of warnings (I created new translated labels for this because as far as I know, it's not possible to trigger the re-translation for existing labels in Crowdin.)

I also decided to only show the "Too many audio streams" warning if there are two or more audio streams selected. I'm not very happy with this warning as a whole because I'm uncertain what exactly it is supposed to achieve, but I'm gonna leave it for now.

### DIFF
--- a/docs/guides/admin/docs/releasenotes/old-editor-restrictions
+++ b/docs/guides/admin/docs/releasenotes/old-editor-restrictions
@@ -1,0 +1,3 @@
+Restrictions in regards to what kind of streams can be selected in the old integrated video editor were eased to allow
+for more diverse use cases. Previous errors were demoted to warnings. The user still has to select at least one stream,
+but publishing e.g. of an audio stream without a video stream is now allowed. For more details see PR 5023.
--- a/modules/admin-ui-frontend/app/scripts/modules/events/subresources/controllers/toolsController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/subresources/controllers/toolsController.js
@@ -155,23 +155,30 @@ angular.module('adminNg.controllers')
       $scope.area = area;
     };
 
+    $scope.anyTrackPresent = function (type) {
+      if ($scope.video.source_tracks === undefined) {
+        return false;
+      }
+      for(var i = 0; i < $scope.video.source_tracks.length; i++) {
+        var t = $scope.video.source_tracks[i][type];
+        if (t.present === true) {
+          return true;
+        }
+      }
+      return false;
+    };
+
     $scope.anyTrackSelected = function (type) {
       if ($scope.video.source_tracks === undefined) {
         return false;
       }
-      var present = false;
       for(var i = 0; i < $scope.video.source_tracks.length; i++) {
         var t = $scope.video.source_tracks[i][type];
-        if (t.present === true) {
-          present = true;
-          // track selected?
-          if (t.hidden === false) {
-            return true;
-          }
+        if (t.present === true && t.hidden === false) {
+          return true;
         }
       }
-      // If we don't have any tracks at all, selecting none is valid
-      return !present;
+      return false;
     };
 
     $scope.trackClicked = function(index, type) {
@@ -199,11 +206,11 @@ angular.module('adminNg.controllers')
           videoTrackCount++;
         }
       }
-      return audioTrackCount > videoTrackCount;
+      return audioTrackCount >= 2 && audioTrackCount > videoTrackCount;
     };
 
-    $scope.sanityCheckFlags = function() {
-      return (!$scope.anyTrackSelected('video') || $scope.tooManyAudios()) || !$scope.anyTrackSelected('audio');
+    $scope.isValid = function() {
+      return $scope.anyTrackSelected('video') || $scope.anyTrackSelected('audio');
     };
 
     // TODO Move the following to a VideoCtrl
--- a/modules/admin-ui-frontend/app/scripts/modules/events/subresources/controllers/toolsController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/subresources/controllers/toolsController.js
@@ -156,25 +156,22 @@ angular.module('adminNg.controllers')
     };
 
     $scope.anyTrackSelected = function (type) {
-      var selected = false;
-      var present = false;
       if ($scope.video.source_tracks === undefined) {
         return false;
       }
+      var present = false;
       for(var i = 0; i < $scope.video.source_tracks.length; i++) {
         var t = $scope.video.source_tracks[i][type];
         if (t.present === true) {
           present = true;
-        }
-        if (t.present === true && t.hidden === false) {
-          selected = true;
+          // track selected?
+          if (t.hidden === false) {
+            return true;
+          }
         }
       }
       // If we don't have any tracks at all, selecting none is valid
-      if (present === false) {
-        return true;
-      }
-      return selected;
+      return !present;
     };
 
     $scope.trackClicked = function(index, type) {
--- a/modules/admin-ui-frontend/app/scripts/modules/events/subresources/partials/tools.html
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/subresources/partials/tools.html
@@ -302,13 +302,20 @@
           <div ng-if="area === 'tracks'" class="editor-tracks">
             <div class="area-body">
               <div data-admin-ng-notification="" data-type="error"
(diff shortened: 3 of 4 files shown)


snoesberger:
This change makes perfect sense for our use cases. I was able to successfully test the change in our environment.

KatrinIhler:
Further testing required to ensure that the backend can handle the different input.

Arnei (commented):
Reason for wanting this seems reasonable to me and testing locally the code seems to work fine. 

The warning messages fade out after a couple of seconds. If you let them, they cannot be triggered again until you refresh the page. I'd say the warning messages shouldn't fade out.

Having no audio tracks should be completely fine.

Having no video tracks is a problem. Not all Opencast operations support audio-only, the default publish workflow will fail in that case. Maybe the warning could make note of that, i.e. "[...] this will likely fail unless you have a special audio-only workflow" or something?

Having more audio tracks than video tracks ... I don't even know what one should expect to happen there. Mixing them? With the default publish workflow it does not matter, it will make sure that only of the audio tracks is published either way.
Maybe this is just about protecting the user from bad assumptions. I.e. they deselect the `presenter` track, expecting the remaining `presentation` track to have `presentation` audio, but then it actually has `presenter` audio because they did not deselect `presenter` audio. Or something like that.
Maybe this case should stay an error? Or do you have a use case for it?

Also this can go against an earlier branch than `develop` I'd say.

snoesberger:
> Having no video tracks is a problem. Not all Opencast operations support audio-only, the default publish workflow will fail in that case. Maybe the warning could make note of that, i.e. "[...] this will likely fail unless you have a special audio-only workflow" or something?

In my opinion, such a message would confuse the end user, who is usually unaware of the existence of such a workflow. Ideally, the behaviour would be configurable, depending on whether audio-only is generally supported by the operator.


> Having more audio tracks than video tracks ... I don't even know what one should expect to happen there. Mixing them? With the default publish workflow it does not matter, it will make sure that only of the audio tracks is published either way.
> Maybe this is just about protecting the user from bad assumptions. I.e. they deselect the `presenter` track, expecting the remaining `presentation` track to have `presentation` audio, but then it actually has `presenter` audio because they did not deselect `presenter` audio. Or something like that.
> Maybe this case should stay an error? Or do you have a use case for it?

So far I don't know of any case at our university where the use of two audio tracks has been requested. I can only imagine a case where someone has audio tracks in different languages in their video.
To keep it simple, in my opinion this can stay an error for the moment.

KatrinIhler:
In my opinion, as long as the errors in the video editor aren't configurable (which would arguably be better), then everything that can be handled in _a_ workflow should be allowed in the editor - just because something isn't possible with the _default_ workflows, it shouldn't be forbidden. It shouldn't be necessary to patch Opencast to make alternative use cases work, because this is not practical for a lot of institutions. As long as the backend of Opencast can handle it, the editor should only warn about unusual input, and this should suffice. (It's also not a tragedy if a publication isn't successful or has unexpected results, a user can easily go back into the editor, fix the warning and trigger a re-publish.)

Regarding the "Two audio tracks and one or no video tracks" - I agree that this is an unusual combination. _But_: Currently we allow two videos which both have audio tracks. With this PR, we're attempting to also allow one track audio without a video. What would be the reasoning for not allowing two audio tracks without video? Just because it's "weird"? If the workflow can handle a video track being missing (which we actually just substitute with a still image), it can do the same thing for two. And then it's no different from two videos with audio, which we allow. You get what I mean? :sweat_smile: 

Regarding the warnings fading: If you want, I can try to patch that. But it's possible that the errors were also fading, would need to test to confirm.

I would not extend the warning because a user might have to idea if the workflows can handle audio-only or not, as Sascha said, and also if the warning is too long, this will break the UI.

Since this changes behavior in a way that might be controversial, I'm gonna keep aiming this against develop so that people can still complain about it even after the merge.

If I failed to address anything, please let me know. :)

Arnei:
> Regarding the "Two audio tracks and one or no video tracks" - I agree that this is an unusual combination. But: Currently we allow two videos which both have audio tracks. With this PR, we're attempting to also allow one track audio without a video. What would be the reasoning for not allowing two audio tracks without video? Just because it's "weird"? If the workflow can handle a video track being missing (which we actually just substitute with a still image), it can do the same thing for two. And then it's no different from two videos with audio, which we allow. You get what I mean? sweat_smile

My main reason for not allowing this would be that there is no single intuitive expectation to as to what will happen.

Situation 1: No audio tracks are selected
Expectation 1a: No audio plays in the published event

Situation 2: One audio track selected
Expectation 2a: The selected audio track plays in the published event

Situation 3: Both audio tracks are selected, and both video tracks are selected
Expectation 3a: I didn't change anything, so it will magically play the "correct" audio track in the published event

Situation 4: Both audio tracks are selected, and one video track is selected
Expectation 4a: I didn't change anything about the audio tracks, so it will magically play the "correct" audio track in the published event 
Expectation 4b: It will play the audio track that corresponds to the selected video
Expectation 4c: Something else entirely

So my gripe is that in situation 4 there are multiple valid intuitive expectations, which will lead to confusion among the users with the "wrong" expectation. Forcing users who select a single video to also select the audio track that should go along with it would hopefully help avoiding such "wrong" expectations. 

But then again I'm no UX designer, have run no case studies and no real-life feedback, so this isn't exactly a scientific opinion or anything.

> Regarding the warnings fading: If you want, I can try to patch that. But it's possible that the errors were also fading, would need to test to confirm.

Would by nice if it's not too much trouble. If it was like this before already (didn't check) then it's whatever.

> I would not extend the warning because a user might have to idea if the workflows can handle audio-only or not, as Sascha said, 

Fine be me, but I still want to somehow make it clear to the user that if they don't know if audio-only will work or not, it likely won't. And using the same wording as in the video-only case (which likely will work) seems misleading.

> and also if the warning is too long, this will break the UI.

Love it.

> (It's also not a tragedy if a publication isn't successful or has unexpected results, a user can easily go back into the editor, fix the warning and trigger a re-publish.)

In my head this sounds like a lot of work for the technical support, since the user will likely have no idea what went wrong or what they should to fix what they broke, so they'll have to seek help. But again, that's just me guessing, I don't have a lot of direct contact to end users at all.

KatrinIhler:
Quick answer to this point:

> My main reason for not allowing this would be that there is no single intuitive expectation to as to what will happen.

The editor can't guarantee that anyway, as what will happen relies completely on the workflow, in which different institutions do a number of different things in varying degrees of craziness. I don't think we should try to solve this in the editor, because the only really valid way would be to make this configurable, which is a great idea, but doesn't seem worth it anymore. So as an alternative, I vote for picking the lowest common denominator, so to speak.

Arnei:
I'm conceding this discussion on the grounds that
- At the very least, creating an audio only event through the editor is useful for multiple institutions, so forcing them to custom patch the editor is not very nice.
- The old editor will (hopefully) soon cease to be used anyway.

Will have to come back to this discussion for the new editor, by the time the new editor can also differentiate between video and audio tracks (currently it can't).

Arnei (approved):

KatrinIhler:
Rebased on current develop and added releasenotes. @gregorydlogan Could you take a look at this and confirm it's generally sane before merging?

gregorydlogan:
~Is audio only supposed to work without any warnings?  I'm fine with that (PEBKAC), but did not see a warning/notification and ended up with a failed workflow.  As you mention it's not worth dumping time into this to make it properly configurable.~

Edit: Sigh.  Nevermind.  P truly exists between KAC.

Issue #5031, opened 2023-06-07
Adopter registration does not report correct user counts


gregorydlogan:
**Describe the bug**

Registered adopter has 20 users reported, with 17k+ user references in the db.

JulianKniephoff:
This was resolved by #5032, right? It's not automatically closed because that PR only mentioned this issue in the description and was not targeting our default branch.

Of course anyone is invited to reopen this if it is still an issue. 😄

PR #5032, opened 2023-06-08
Count user references

gregorydlogan:
This PR fixes (in theory) the user count reported to the adoption registration server.  The existing implementation counts the raw JPA users, but does not count the user *references*.  This PR should hopefully count the references.

Fixes: #5031

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
@@ -40,11 +40,13 @@ import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.security.api.User;
+import org.opencastproject.security.api.UserProvider;
 import org.opencastproject.security.util.SecurityUtil;
 import org.opencastproject.series.api.SeriesService;
 import org.opencastproject.serviceregistry.api.ServiceRegistry;
 import org.opencastproject.serviceregistry.api.ServiceRegistryException;
 import org.opencastproject.userdirectory.JpaUserAndRoleProvider;
+import org.opencastproject.userdirectory.JpaUserReferenceProvider;
 
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Version;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
@lkiesow You pointed this out, can you test this against your systems?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -333,6 +335,10 @@ private String collectStatisticData(String adopterKey, String statisticKey) thro
           logger.warn("Unable to calculate total minutes, unauthorized");
         }
         statisticData.setTotalMinutes(current + orgDuration);
+
+        //Add the users for each org
+        long currentUsers = statisticData.getUserCount();
+        statisticData.setUserCount(currentUsers + userProvider.countUsers() + userRefProvider.countUsers());
```
Arnei, line 341: Why would `statisticData.getUserCount();` ever return a number greater 0? And in the cases in which would, can we be sure that we don't count them twice when adding the user counts from the two user providers?

I don't think is a blocker, since I don't expect `statisticData.getUserCount();` to ever return anything other than 0, and even if ended up counting a few users twice it likely does not matter as we don't needs these stats to be 100% exact.  I'm just curious. PR looks good and works otherwise.

gregorydlogan: Take look at line 307, we're iterating across all possible orgs :)

In terms of counting them twice, good point.  Let me give it a think.

Arnei: ... Reading becomes hard at the end of the day. Thanks for pointing me to correct line, makes sense to me now.

Then no user should be counted twice, right? I don't know too much about tenant stuff, but two users with the same uid in different tenants should be considered different users I'd expect.

Arnei (approved):

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: I'm not 100% sure, but I'm pretty sure the only ones that'd get counted multiple times could be things that are common across tenants. How often that would happen I'm not sure.

What I am sure of is that the number of users who get counted multiple times is going to be insignificant compared to the full body of user references. You'd have to count your faculty/admin people a whole bunch of times to even approach your student/user count and thus I wasn't too worried about it.  Whether it's worth it/possible to figure out the dups I'm not sure...

Issue #1673, opened 2020-07-03
External api search filter can't handle ":", and admin ui can't handle "," in their filters


kristofkeppens:
**Describe the bug**
When you have for instance serie titles that include a ":" in it's name then it's impossible to retrieve this serie via the external api. 

When using the external api and using /api/series?filter=title:test:test for instance the api returns ALL series. 
I've tried escaping the ":" character with \: or %3A. \: results in all series retrieved, %3A returns no results.

Edit: the admin UI has the same issue, but the delimiter in use is ",".  Likely very similar logic though. --@gregorydlogan

**Environment Information**
 - opencast 8.x

mliradelc:
Hi, I will take on this, let me check it.

mliradelc:
Ok, I made some tests and I found interesting results. First of all, I can confirm, **It's a bug**

The other thing I found is that the internal API doesn't have this problem, if you make the same query, you get the correct result, for example using:

 ```python
{{admin-url}}/admin-ng/series/series.json?filter=title:test:test
```

gives you the correct answer. In the meantime, if you can, please use the internal API as a temporary workaround.

mliradelc:
Update: 
Also affects version 9.x (Develop)

miesgre:
We have the same issue but with the comma character `,`
All our presenters are identified as `lastname, name` and we can not filter by presenters

> 
> ```python
> {{admin-url}}/admin-ng/series/series.json?filter=title:test:test
> ```
> 
> gives you the correct answer. In the meantime, if you can, please use the internal API as a temporary workaround.

@mliradelc The internal API does **NOT** work with the comma character

gregorydlogan:
I just ran into this testing some admin UI changes.  A location value of "DE, Onsabruck..." finds no workflows, even when you use the filter dropdown to select your value.

What's happening internally (at least in the admin ui endpoints) is that it's splitting on ",", which is why @miesgre's stuff isn't working either.  I suspect the external API is using the same logic, but a different delimiter for its parsing.

I'm going to modify the initial report here rather than creating an (almost) duplicate issue.

PR #5033, opened 2023-06-08
Escape ES query string in external API endpoints

Arnei:
Some external API endpoints have optional query parameters that are internally treated as ElasticSearch queries. Endpoint users are not informed about this, due to which they may include characters in their parameter that are actually reserved in Elasticsearch without escaping them. At best, the endpoint then might return unexpected results, at worst it goes into an endless loops that spams the Opencast log with error messages until the disk is full.

Example: The `/api/series/series.json` endpoint. Set the parameter `q` to `T9-VB-20-Online:`.

This commit suggests fixing the problem by adding a method for sanitizing parameters. It simply escapes ElasticSearch reserved characters. 
Another way of fixing this would be to explicitly inform users that a certain parameter will be an ElasticSearch query and let them decide which reserved characters they want to escape in their query.

If people are happy with simply sanitizing incoming strings, the code suggested here could likely be approved upon by burying it deeper in the Elasticsearch modules. But that's a pain so I refrained from doing that for now ^^'.

### DIFF
--- a/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
+++ b/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
@@ -839,4 +839,27 @@ public class ElasticsearchIndex extends AbstractElasticsearchIndex {
       throw new SearchIndexException("Error querying theme index", t);
     }
   }
+
+  /**
+   * Escapes all reserved Elasticsearch characters in a given string
+   * Useful for when a given query string does not know about Elasticsearch query syntax
+   * @param text String to escape reserved characters in
+   * @return the given string with escaped characters
+   */
+  public String escapeQuery(String text) {
+    StringBuilder sb = new StringBuilder();
+    for (int i = 0; i < text.length(); i++) {
+      char c = text.charAt(i);
+      // These characters are part of the query syntax and must be escaped
+      // CHECKSTYLE:OFF
+      if (c == '\\' || c == '+' || c == '-' || c == '!' || c == '(' || c == ')' || c == ':'
+              || c == '^' || c == '[' || c == ']' || c == '\"' || c == '{' || c == '}' || c == '~'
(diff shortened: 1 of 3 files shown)


modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -839,4 +839,27 @@ private SearchResult<IndexTheme> getByQuery(ThemeSearchQuery query, int maxRetry
...
+      if (c == '\\' || c == '+' || c == '-' || c == '!' || c == '(' || c == ')' || c == ':'
+              || c == '^' || c == '[' || c == ']' || c == '\"' || c == '{' || c == '}' || c == '~'
+              || c == '*' || c == '?' || c == '|' || c == '&' || c == '/') {
+        sb.append('\\');
+      }
+      // CHECKSTYLE:ON
+      sb.append(c);
+    }
+    return sb.toString();
+  }
```
geichelberger:
```suggestion
public String escapeQuery(String text) {
  Set<Character> specialChars = Set.of('\\', '+', '-', '!', '(', ')', ':', '^', '[', ']', '\"', '{', '}', '~', '*', '?', '|', '&', '/');
  
  StringBuilder sb = new StringBuilder(text.length());
  
  for (char c : text.toCharArray()) {
    if (specialChars.contains(c)) {
      sb.append('\\');
    }
    sb.append(c);
  }
  
  return sb.toString();
}
```

JulianKniephoff: Or just `Set.of(...)` instead of `new HashSet<>(Arrays.asList(...))` :D

mliradelc:
This PR is a possible solution for #4969

wsmirnow:
@mliradelc you probably mean https://github.com/opencast/opencast/issues/1673

gregorydlogan:
Works as expected!

PR #5035, opened 2023-06-10
Publish and update maven cache

lkiesow:
We have a lot of automated build downloading dependencies from Maven Central and mvn.opencast.org over and over again. This takes time and occasionally causes build failures.

This patch adds a workflow building up a repository cache once a week and making it publically available, so that you can just get most of the dependencies in just a single tarball. This should make things faster and easier.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/build-maven-cache.yml
+++ b/.github/workflows/build-maven-cache.yml
@@ -1,0 +1,76 @@
+name: Build Maven Cache
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 6'
+
+jobs:
+  build:
+    name: build maven cache
+    runs-on: ubuntu-latest
+    if: github.repository == 'opencast/opencast'
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: setup java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: clone mvn.opencast.org
+        run: |
+          git clone --depth 1 https://github.com/opencast/mvn.opencast.org.git
+
+      - name: prepare additional maven repository
+        run: |
+          sed -i "s#https://mvn.opencast.org/#file://$(readlink -f mvn.opencast.org)#" pom.xml
+
+      - name: install dependencies
+        run: |
+          sudo apt update -q
+          sudo apt install -y -q \
+            bzip2 \
+            ffmpeg \
+            gzip \
+            s3cmd \
+            sox \
+            tar \
+            unzip
+
+      - name: prepare build
+        run: |
+          sed -i 's/build --skipTests=${skipTests}/build --skipTests=true/' modules/admin-ui-frontend/pom.xml
+
+      - name: build opencast
+        run: |
+          mvn clean install -Pdev \
+            --batch-mode \
+            -Dsurefire.rerunFailingTestsCount=2 \
+            -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
+            -Dhttp.keepAlive=false \
+            -Dmaven.wagon.http.pool=false \
+            -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-maven-cache.yml
```diff
@@ -0,0 +1,76 @@
...
+            tar \
+            unzip
+
+      - name: prepare build
+        run: |
+          sed -i 's/build --skipTests=${skipTests}/build --skipTests=true/' modules/admin-ui-frontend/pom.xml
+
+      - name: build opencast
+        run: |
+          mvn clean install -Pdev \
```
gregorydlogan, line 47: Does it make sense to use `dependency:go-offline` instead?  Then we're not packaging the resulting OC jars as well, which we (probably) don't actually want?

.github/workflows/build-maven-cache.yml
```diff
@@ -0,0 +1,76 @@
...
+          echo "host_bucket = ${S3_HOST}" >> "$HOME/.s3cfg"
+          echo "bucket_location = us-east-1" >> "$HOME/.s3cfg"
+          echo "use_https = True" >> "$HOME/.s3cfg"
+          echo "access_key = ${S3_ACCESS_KEY}" >> "$HOME/.s3cfg"
+          echo "secret_key = ${S3_SECRET_KEY}" >> "$HOME/.s3cfg"
+          echo "signature_v2 = False" >> "$HOME/.s3cfg"
+
+      - name: upload assets
+        run: |
+          s3cmd put -P dot-m2.tar.xz s3://opencast-maven-cache/
```
gregorydlogan, line 74: This workflow doesn't explicitly say which branch(es) it runs on, but they're all going to write to the same place.  I'm assuming this only runs against the main branch when it's unspecified?

Also, we're building this cache, should we use it in the builds?

.github/workflows/build-maven-cache.yml
lkiesow: I didn't know about `dependency:go-offline`. That sounds like a good idea.
Would the command you are suggesting be:
```
          mvn dependency:go-offline -Pdev \
            --batch-mode \
            -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
            -Dhttp.keepAlive=false \
            -Dmaven.wagon.http.pool=false \
            -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
```

.github/workflows/build-maven-cache.yml
lkiesow: Scheduled workflows always run on the default branch. But even using that, it should provide most dependencies for release branches.

I'd like to use this in builds in the long run. But I thought it makes sense to get this uploading first, and then we can play with using this for different builds.

.github/workflows/build-maven-cache.yml
gregorydlogan: I would normally have left the `-Pdev` off, but realistically it's not going to matter either way.

.github/workflows/build-maven-cache.yml
gregorydlogan: Fair.  Builds in a followup PR then.

.github/workflows/build-maven-cache.yml
lkiesow: I used `-Pdev` instead of `-Pnone` so that maven will fetch Karaf, its plugins and dependencies which I think it otherwise wouldn't do, would it?

gregorydlogan: No idea, but including the dev stuff makes sense to me.

lkiesow: Unfortunately, `mvn dependency:go-offline` fails. I've worked around that by now by using the previous command but removing the Opencast JARs from the cache before creating the tarball.

lkiesow:
This should be ready to go now

PR #5040, opened 2023-06-14
Enable Tobira adopter stats tracking

gregorydlogan:
This PR adds a REST endpoint for Tobira to store usage statistics in.  This data will then be sent to the adopter registration server so that we can estimate how many tobira adopters there are.

Of note:
- This data is stored in memory only.  Tobira would repopulate this field on next stats post (so, daily, just like the adopter registration?).
- This data is only sent if it is present, and a lack of adopter-side data would not overwrite existing recorded data on the registration server.
- This data is passed through more or less unmodified.  The statistics UUID is added to whatever json structure is sent by Tobira so that a given Tobira instance can be associated with an adopter.

### Open Questions
- What does the data structure sent from Tobira look like?  Something like this?
```json
{
  "version": "1.0.0",
  "blocks": 4,
  ...
}
```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -224,8 +231,13 @@ public void run() {
 
       if (adopter.allowsStatistics()) {
         try {
-          String statisticDataAsJson = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
-          sender.sendStatistics(statisticDataAsJson);
+          StatisticData statisticData = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
+          sender.sendStatistics(statisticData.jsonify());
+          if (null != tobiraEndpoint) {
+            JsonObject tobiraJson = tobiraEndpoint.getStats();
+            tobiraJson.addProperty("statistic_key", statisticData.getStatisticKey());
```
LukasKalbertodt: One could instead construct a JSON `{ "statistic_key": ..., data: { /* whatever tobira sent */ } }` so that we never accidentally have two fields of the same name, if Tobira adds `statistic_key` in the future. That said, if it will only be `statistic_key` forever, I don't think this scenario is terribly likely :P Wouldn't know why Tobira would send such a field.

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -204,4 +214,31 @@ private static Response badRequest(String msg) {
...
+      },
+      responses = {
+          @RestResponse(description = "Stats parsed", responseCode = HttpServletResponse.SC_ACCEPTED)
+      },
+      returnDescription = "No data returned, just a 204 on success"
+  )
+  public Response acceptStats(@FormParam("stats") String stats) {
+    this.stats = new Gson().fromJson(stats, JsonElement.class).getAsJsonObject();
+    return Response.noContent().build();
+  }
```
LukasKalbertodt: This should be protected somehow. Don't know if that should happen in this Java code or in the security config. Tobira has a user with `ROLE_ADMIN` anyway to contact the harvest API, so here we can also just require `ROLE_ADMIN`. Otherwise random people could just post garbage here.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
LukasKalbertodt: If Tobira hasn't posted data yet `tobiraJson` is `null` here, right?

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -204,4 +214,31 @@ private static Response badRequest(String msg) {
...
+      name = "stats",
+      description = "Accepts a json blob of statistical data about Tobira.",
+      restParameters = {
+          @RestParameter(
+              name = "stats",
+              isRequired = true,
+              description = "The statistics data generated by Tobira",
+              type = Type.STRING
+          )
+      },
```
LukasKalbertodt: So this means using `application/x-www-form-urlencoded` correct? Would it be possible to accept an `application/json` body? I.e. the whole HTTP requests body is just JSON? Would be easier for Tobira and I don't see the need to put another encoding on top, i.e. I don't see why we would add additional form parameters. We can just add additional fields in the JSON instead.

LukasKalbertodt (changes requested):
I noticed a few minor things in code. I ran it locally and already built the Tobira side (see https://github.com/elan-ev/tobira/pull/878). Well, almost. There is one open question regarding the HTTP content type (see below). So I couldn't fully test it yet.

I think the approach with just storing it in memory is absolutely valid IMO. Also, passing the JSON through completely unmodified is a good idea I think, as we then only have to touch two, and not three, places if we want to add new data.

> What does the data structure sent from Tobira look like? Something like this?

With the linked PR, Tobira would currently send this:

```json
{
    "numRealms": 3631,
    "numBlocks": 5558,
    "version": {
        "identifier": "v2.0",
        "buildTimeUtc": "Thu, 15 Jun 2023 11:26:19 +0000",
        "gitCommitHash": "18521b24e34cd11990f6b33913fb0d7a854ab0ec",
        "gitWasDirty": true
    },
    "config": {
        "downloadButtonShown": true,
        "authMode": "login-proxy",
        "loginLinkOverwritten": false,
        "logoutLinkOverwritten": false,
        "usesPreAuth": true,
        "hasNarrowLogo": true
    }
}
```

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: Tobira might not need the key but it does make the separation-of-data clear: This is the stuff that Tobira sent, and this is the OC cluster's ID.  Will implement it this way I think, makes more sense.

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
gregorydlogan: The security XML already restricts the general `/tobira` namespace to `ROLE_ADMIN`, aside from `/tobira/version`.  I... think that's enough?  Or are you thinking of further future granularity?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: Ha, yes!  Good catch!

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
LukasKalbertodt: Oh right. That's sufficient indeed!
(Feel free to press the resolve button... I don't have the rights to do so -.-)

lkiesow:
> This data is only sent if it is present, and a lack of adopter-side data would not overwrite existing recorded data on the registration server.

That means that we will never know if people stop using Tobira. 

If we overwrite the data, it will get corrected eventually? Or, we don't overwrite just the first time we send data?

gregorydlogan:
> That means that we will never know if people stop using Tobira.

Lars is right, we'd have to do some reasoning about how they're using it from the stats...

@LukasKalbertodt Can we slap a last-harvested date into the structure?  Then if that stops updating then we know it's not being used anymore.

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
gregorydlogan: Try it now?  I *think* I've got it right...

LukasKalbertodt:
Mh I suppose the Opencast side can just put a `date` field next to `statistic_key`? I.e. `blob.addProperty("date", now())` or something like that? Sure, Tobira could also report "last harvested". Or maybe both? Though... if Tobira reports statistics, it is very very likely also harvesting, so that the "last harvested" would always be approximately `now()` from when the statistic data is received. I don't think "sends statistics data but doesn't harvest" is something that happens.

EDIT: ah, I just noticed that `statistic_key` is added when sending out the data, not when receiving Tobira data. The `now()` needs to be executed when receiving the data, of course. So yeah, I don't care. If you prefer I can easily add a date or "last harvested" field to the blob Tobira sends. Or Opencast can just store the date of when the data was received. :shrug:

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -224,8 +232,19 @@ public void run() {
...
-          String statisticDataAsJson = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
-          sender.sendStatistics(statisticDataAsJson);
+          StatisticData statisticData = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
+          sender.sendStatistics(statisticData.jsonify());
+          if (null != tobiraEndpoint) {
+            JsonObject tobiraJson = tobiraEndpoint.getStats();
+            // This is null in the case that Tobira hasn't sent any stats yet.
+            // This could be due to Tobira not existing, or because we've just rebooted.
+            if (null != tobiraJson) {
+              JsonObject blob = new Gson().fromJson("{}", JsonElement.class).getAsJsonObject();
```
LukasKalbertodt: Shouldn't `new JsonObject()` do the same thing? At least the docs I found say that the constructor returns an empty object.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -224,8 +232,19 @@ public void run() {
...
+              blob.addProperty("statistic_key", statisticData.getStatisticKey());
```
LukasKalbertodt: Not a comment about your code but I just noticed the `snake_case`. And checking that other data that is sent, everything seems to be snake case. So then I will change Tobira to also send snake case!

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -204,4 +219,35 @@ private static Response badRequest(String msg) {
...
+  @Path("/stats")
+  @Consumes(APPLICATION_JSON)
+  @RestQuery(
+      name = "stats",
+      description = "Accepts a json blob of statistical data about Tobira.",
+      restParameters = {},
+      bodyParameter = @RestParameter(description = "The job blob",
+                                     isRequired = true,
+                                     name = "BODY",
+                                     type = Type.STRING),
```
LukasKalbertodt: I think the latest style guide would say this (same format as in your prior commits).

```suggestion
      bodyParameter = @RestParameter(
            description = "The job blob",
            isRequired = true,
            name = "BODY",
            type = Type.STRING
      ),
```

Also, "the job blob" is likely a copy and paste error?

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -204,4 +219,35 @@ private static Response badRequest(String msg) {
...
+      responses = {
+          @RestResponse(description = "Stats parsed", responseCode = HttpServletResponse.SC_ACCEPTED)
+      },
+      returnDescription = "No data returned, just a 204 on success"
+  )
+  public Response acceptStats(@Context HttpServletRequest request) {
+    try (InputStream is = request.getInputStream()) {
+      String json = IOUtils.toString(is, request.getCharacterEncoding());
+      stats = new Gson().fromJson(json, JsonElement.class).getAsJsonObject();
```
LukasKalbertodt: I suppose you wanted to use `gson` here? Alternatively, remove the class field.

```suggestion
      stats = gson.fromJson(json, JsonElement.class).getAsJsonObject();
```

LukasKalbertodt (changes requested):
I tested it now locally with https://github.com/elan-ev/tobira/pull/878 and it seems to work! I haven't yet tested whether Opencast sends out the correct data again, though.

One thing worth noting: if Tobira send data to Opencast first and then you open the admin UI and see the adopter registration. In the step where it shows the JSON, I don't see the Tobira data anywhere, despite Opencast already having it. Maybe that should still be added there?

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: This is an (unfortunate) artifact of the Java<->Python stuff integration.  The sever's Python is written in proper (more or less) `snake_case`, but that's obviously not going to match Opencast's codebase :(

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: Derp, thanks, good catch.

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
gregorydlogan: Yep

gregorydlogan:
I'm going to mark this as ready for review since *this* part is done. We still need to hash out the final json structure, but Opencast doesn't need to care much about that :D

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -415,8 +416,15 @@ public void setOrganizationDirectoryService(OrganizationDirectoryService orgDirS
...
-  @Reference(cardinality = ReferenceCardinality.OPTIONAL)
+  @Reference(
+      cardinality = ReferenceCardinality.OPTIONAL,
+      policy = ReferencePolicy.DYNAMIC,
+      unbind = "unsetTobiraEndpoint")
   public void setTobiraEndpoint(TobiraEndpoint endpoint) {
     this.tobiraEndpoint = endpoint;
   }
+
+  public void unsetTobiraEndpoint(TobiraEndpoint endpoint) {
```
LukasKalbertodt, line 427: Is this argument here actually useful? :thinking: Is it always `null`?

LukasKalbertodt (commented):
Apart from the small question/comment, this looks fine to me.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
gregorydlogan: It's not actually null, it's the service being unregistered :)

LukasKalbertodt: But like, what's the value of `endpoint` when the function is called?

gregorydlogan: `endpoint` is the service instance which is shutting down - `TobiraEndpoint` in this case.  This is a simplification of the `1..n` relationship where *multiple* `TobiraEndpoint`s could be defined.  We'd need to pass the one which is going away so that the binding service (`ScheduledDataCollector`) can do whatever it needs to shut down its end of the connection.

LukasKalbertodt: Thanks for the explanation! This can be resolved then.

PR #5049, opened 2023-06-20
Paella 7: Enable dfxp captions support

miesgre:
### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/ui-config/mh_default_org/paella7/config.json
+++ b/etc/ui-config/mh_default_org/paella7/config.json
@@ -522,7 +522,9 @@
         "es.upv.paella.vttManifestCaptionsPlugin": {
             "enabled": true
         },
-
+        "es.upv.paella.dfxpManifestCaptionsPlugin": {
+            "enabled": true
+        },
         "es.upv.paella.hlsCaptionsPlugin": {
             "enabled": true
         },
--- a/modules/engage-paella-player-7/package-lock.json
+++ b/modules/engage-paella-player-7/package-lock.json
@@ -6,8 +6,8 @@
     "": {
       "name": "paella-7-opencast",
       "dependencies": {
-        "paella-basic-plugins": "^1.31.2",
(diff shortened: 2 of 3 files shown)


Arnei:
I don't think we need to have a rather old and probably rarely used subtitle format enabled by default. Those who need it can still enable it of course.

Also if we ever get #4407 done, the only subtitle format we are going to officially support will be WebVTT. Having dfxp enabled per default would contradict that.

karendolan:
> Also if we ever get #4407 done, the only subtitle format we are going to officially support will be WebVTT. Having dfxp enabled per default would contradict that.

Hi Arnei, Our site has a lot of old mediapackages published with DFXP captions. The very old ones are attached as catalogs, and the less old ones are attached as attachments.

Even if we switch to WebVTT tracks in the new events, it will take a while to upgrade the old mediapackages, if we are required to do that when migrating to a newer version of Opencast. I agree that our site, and other sites with  the old style captions, can enable the DFXP parser plugin as needed through the Paella 7 config. It might help to mention this in Opencast upgrade documentation if the DFXP parser will not be enabled during the upgrade.

miesgre:
> I don't think we need to have a rather old and probably rarely used subtitle format enabled by default. Those who need it can still enable it of course.
> 
> Also if we ever get #4407 done, the only subtitle format we are going to officially support will be WebVTT. Having dfxp enabled per default would contradict that.

There are several institutions that are using the dfxp format, including us.

Enabling it in the player only simplifies its use, since in paella 6 this support is active by default.
If any institution switches from paella 6 to 7, the subtitles will not work automatically.

I believe that enabling dfxp support in the player does not hurt, but if the community wants to disable it, there are two options:
- Leave the plugin in the config.json with enabled:false (this would simplify activating it if someone wants to).
- Remove the plugin from config.json

PR #5051, opened 2023-06-20
Paella7 backwards support for old captions/dfxp flavored xml files

karendolan:
Issue: DFXP flavored xml files are given am 'xml' format instead of a 'dfxp' caption format.

This pull patches the EpisodeConversor for Paella7 to support backwards compatibility for old mediapackagesa with the historic "captions/dfxp" flavored XML attachments and catalogs.


karendolan:
I changed the target branch to be 14.x to match https://github.com/opencast/opencast/pull/5049

miesgre (approved):
For me it is ok! 👍

PR #5055, opened 2023-06-22
Dont copy media files to new scheduled event

### LINKED ISSUE #4312: Inbox behaviour: additional recordings in scheduled events should not copy initial media files
**Describe the bug**
This bug report concerns the following:
https://github.com/opencast/opencast/pull/3687

When ingesting multiple events within the time frame of one scheduled event (which can happen in case of breaks, test recordings etc.), all additional media packages contain the initial media file(s) from the first part of the event. This should not be the case as copying those media files is superfluous. Instead each additional media package should only include its respective media file(s).

**To Reproduce**
1. Activate and setup feature in the inbox-scanner config
2. Schedule an event
3. Ingest multiple events within the time scope of the scheduled event

**Environment Information**
 - Opencast version 10.8

Arnei:
Resolves #4312

In case of adding new media files to a scheduled event through the InboxScanner, Opencast creates a copy of the scheduled event with the new media files. Unfortunately, this means it also copies the old media files from the scheduled event, which could lead to unwanted media files in the new event. Therefore, this commit drops all media files after copying the mediapackage.

Copying and then dropping the media files, while simple, is arguably a pretty wasteful solution to the problem. Another approach would be to define which files need to be copied and then only copy those (i.e. if we actually only need the dublincore, we could simply create a new mediapackage and copy the dublincore over).

### Your pull request should…

* [x] have a concise title
~* [ ] include migration scripts and documentation, if appropriate~
* [ ] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/scanner/Ingestor.java
@@ -30,6 +30,7 @@ import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.MediaPackageElements;
 import org.opencastproject.mediapackage.MediaPackageException;
+import org.opencastproject.mediapackage.Track;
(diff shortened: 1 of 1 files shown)


aliyevTU:
We tested it, looks good.

PR #5056, opened 2023-06-27
Allow episode ID based access control via roles

lkiesow:
If activated, users with a role like `ROLE_EPISODE_<ID>_<ACTION>` will have access to the episode with the given identifier, without this having to be specifically specified in the ACL attached to the episode.

For example, `ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ` will allow the user to get read access to episode `872dc4ec-ca8a-4e12-8dac-ce99784d6d29`.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


lkiesow:
(To-do) What is missing?

ACL file permission checks already work, and checks via Solr work as well. But Elasticsearch also holds a representation of the ACL and needs to honor the new ID roles.

In the best case, it does that without having to modify the stored ACLs in Elasticsearch. Unfortunately, that seems to be hard.

An alternative would be to write modified ACLs to the search index. This would already work when doing a full re-indexing of asset manager data, but that's probably an unreasonable amount of work. Acceptable would be, however, if we could trigger a re-index/update of only the ACLs from the asset manager: Essentially, run through all events in the asset manager, take the ACL from the latest snapshot and write that (including the ID roles) to the correct field in the search index.

Arnei:
Attempted to implement the to-do lkiesow mentioned above, following his suggested alternative (because I agree with his assertion that the implementation for the "best case" would be very hard).

~Also I don't understand why the docs are not compiling.~

etc/org.opencastproject.adminui.endpoint.OsgiEventEndpoint.cfg
gregorydlogan, line 1: Was this included by accident?  If not then this needs to be in `develop`, right?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1611,4 +1626,45 @@ private Function<Optional<Event>, Optional<Event>> getEventUpdateFunction(Snapsh
...
+
+  private Event updateAclInEvent(Event event, MediaPackage mp, String eventId) {
+    AccessControlList acl = authorizationService.getActiveAcl(mp).getA();
+    List<ManagedAcl> acls = aclServiceFactory.serviceFor(securityService.getOrganization()).getAcls();
+
+    if (episodeIdRole) {
+      // Add custom roles to the ACL
+      // This allows users with a role of the form ROLE_EPISODE_<ID>_<ACTION> to access the event through the index
+      List<AccessControlEntry> customEntries = new ArrayList<>();
+      for (AccessControlEntry entry : acl.getEntries()) {
```
gregorydlogan: Why are we doing this once for every ACL entry that's already existing?

Say I've got an event with this ACL:
```
ROLE_USER: READ
ROLE_COURSE_ADMIN: READ,WRITE
```

Wouldn't `customEntries` have
```
ROLE_EPISODE_abc123_READ,
ROLE_EPISODE_abc123_READ,
ROLE_EPISODE_abc123_WRITE
```

in it?  Or am I completely out to lunch on this logic?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
```diff
@@ -626,13 +629,22 @@ private static String mkString(Object f) {
    */
   private SolrQuery getForAction(SearchQuery q, String action, boolean applyPermissions) throws SolrServerException {
     StringBuilder sb = new StringBuilder();
+    logger.error("Called getForAction");
```
gregorydlogan: This should be removed :D

gregorydlogan (changes requested):
Found a few things which need work.  Otherwise looks good?  If this is for a client, you're going to run into issues once #5597 goes into develop, one of us is going to have to add support for this.

etc/org.opencastproject.adminui.endpoint.OsgiEventEndpoint.cfg
Arnei: Yup, an accident. Will remove this change.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
Arnei: The intention here is to also have custom entries for other actions. So you can for example have

```
ROLE_EPISODE_abc123_ANNOTATE,
ROLE_EPISODE_abc123_ANNOTATE_ADMIN,
```

as well. But maybe the code is flawed anyway?

gregorydlogan: I think so, because this *disregards* the role name from `entry` to generates the custom roles.  `ROLE_ANNOTATE_ADMIN: READ` becomes `ROLE_EPISODE_abc123_READ: READ` because it's using `entry`'s action (READ/WRITE/WHATEVER), rather than using some kind of substring of the role name.

Arnei: The role name would be something like `ROLE_USER_ADMIN`, right? That does not contain any info relevant to our custom roles. `ROLE_ANNOTATE_ADMIN: READ` is not really something we have in our ACLs, rather `ROLE_USER_ADMIN: ANNOTATE`, which becomes `ROLE_EPISODE_abc123_ANNOTATE`.

Little excerpt from an Elasticsearch doc for how this looks like for an event I tested with:
```
  "acl_permission_annotate-admin": [
    [
      "ROLE_EPISODE_a112b2c5-f20f-4275-b953-2510a09f4957_ANNOTATE-ADMIN",
      "ROLE_USER_ADMIN"
    ]
  ],
    "acl_permission_read": [
    [
      "ROLE_EPISODE_a112b2c5-f20f-4275-b953-2510a09f4957_READ",
      "ROLE_USER_ADMIN",
      "ROLE_USER_BOB"
    ]
  ],
    "acl_permission_write": [
    [
      "ROLE_EPISODE_a112b2c5-f20f-4275-b953-2510a09f4957_WRITE",
      "ROLE_USER_ADMIN",
      "ROLE_USER_BOB"
    ]
  ],
    "acl_permission_annotate": [
    [
      "ROLE_EPISODE_a112b2c5-f20f-4275-b953-2510a09f4957_ANNOTATE",
      "ROLE_USER_ADMIN",
      "ROLE_USER_BOB"
    ]
  ],
  ```

Comment (Arnei) [reply modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java]: That being said,

> Why are we doing this once for every ACL entry that's already existing?

is probably a valid question anyway. With how I wrote this, I rely on all the actions I want `ROLE_EPISODE_[...]` for to be present in the ACL already. Maybe that's bad? Maybe I want to generate `ROLE_EPISODE_[...]` for all possible actions defined in the system instead? Is there a list of those somewhere?

Comment (gregorydlogan) [reply modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java]: Not that I'm aware of, and certainly if someone were to add a new one that list would likely be incomplete.  I think depending on the incoming ACL seems reasonable, since it should contain everything that's needed.  I guess the question becomes how many is too many to process.  I hope people don't have events with hundreds of ACL entries, but it's technically possible.

I would also be tempted to either check if the generated role is already in `customEntries`, or make `customEntries` a `Set` so that we don't generate a ton of dup roles.  Past that I'm not sure how else to do what you're doing.

File: docs/guides/admin/docs/configuration/role-based-event-access.md [hunk 3]
```diff
@@ -0,0 +1,19 @@
+Role based event access
```
KatrinIhler: <del>I think calling it role based event access might be confusing to some people - since other access is also role-based. The difference here is that the user having the role is enough, without needing to amend the ACL. Not sure how to shorten this to something snappy though... Magic user roles? :D Superpower user roles? :laughing:</del>

<del>Technically ROLE_ADMIN is also such a role.</del>

Since the text calls them episode id roles, which is much more precise, maybe just include that in the title as well?

docs/guides/admin/docs/configuration/role-based-event-access.md
```diff
@@ -0,0 +1,19 @@
...
+Setup
+--------------------
+
+Enable `episode.id.role.access` in `etc/org.opencastproject.assetmanager.impl.AssetmanagerImpl.cfg` and
+`etc/org.opencastproject.authorization.xacml.XACMLAuthorizationService.cfg`.
+
+To make this work for the Admin UI and External API, the Elasticsearch Index needs to be updated with modified
+ACLs. You can achieve this by calling the /index/rebuild/AssetManager/ACL endpoint AFTER enabling this feature
+in the aforementioned configuration files.
+The endpoint will reindex only event ACLs.
```
KatrinIhler: Maybe mention here that only custom actions currently configured will work, and when adding a new custom action, this needs to be done again?

docs/guides/admin/docs/releasenotes/role-based-event-access.txt
```diff
@@ -0,0 +1,2 @@
+Adds the feature "Role Based Event Access". For more information see the doc file
```
KatrinIhler: Same complaint here.

docs/guides/admin/docs/configuration/index.md
```diff
@@ -31,6 +31,7 @@ General Configuration
     - [JWT-based Authentication and Authorization (e.g. for OIDC)](security.jwt.md)
     - [Access Control Lists](acl.md)
     - [Stream Security](stream-security.md)
+    - [Role Based Event Access](role-based-event-access.md)
```
KatrinIhler: Same complaint here.

docs/guides/admin/mkdocs.yml
```diff
@@ -91,6 +91,7 @@ nav:
       - Sakai User Provider: 'configuration/security.user.sakai.md'
       - Brightspace User Provider: 'configuration/security.user.brightspace.md'
       - Canvas User Provider: 'configuration/security.user.canvas.md'
+      - Role Based Event Access: 'configuration/role-based-event-access.md'
```
KatrinIhler: Aand here.

etc/org.opencastproject.assetmanager.impl.AssetManagerImpl.cfg
```diff
@@ -31,3 +31,18 @@
...
+# 872dc4ec-ca8a-4e12-8dac-ce99784d6d29.
+#
+# Don't forget to also activate this feature in org.opencastproject.authorization.xacml.XACMLAuthorizationService.cfg
+#
+# To make this work for the Admin UI and External API, the Elasticsearch Index needs to be updated with modified
+# ACLs. You can achieve this by calling the /index/rebuild/AssetManager/ACL endpoint AFTER activating this feature.
+# The endpoint will reindex only event ACLs.
+#
+# Default: false
+episode.id.role.access = true
```
KatrinIhler: Not set to default :D

etc/org.opencastproject.assetmanager.impl.AssetManagerImpl.cfg
```diff
@@ -31,3 +31,18 @@
...
 #includeUIRoles = false
 
+# Allow episode ID based access control via roles.
+# If activated, users with a role like ROLE_EPISODE_<ID>_<ACTION> will have access to the episode with the given
+# identifier, without this having to be specifically specified in the ACL attached to the episode.
+#
+# For example, ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ will allow the user to get read access to episode
+# 872dc4ec-ca8a-4e12-8dac-ce99784d6d29.
+#
+# Don't forget to also activate this feature in org.opencastproject.authorization.xacml.XACMLAuthorizationService.cfg
```
KatrinIhler: Maybe naive question - wouldn't it be better to have a global config for this instead of having to turn it on for multiple services? E.g. in custom.properties?

etc/org.opencastproject.assetmanager.impl.AssetManagerImpl.cfg
```diff
@@ -31,3 +31,18 @@
 #
 #includeUIRoles = false
 
+# Allow episode ID based access control via roles.
```
KatrinIhler: Good explanation here imo.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexRebuildService.java
```diff
@@ -153,10 +157,15 @@ public synchronized void rebuildIndex(ElasticsearchIndex index)
    */
   public synchronized void rebuildIndex(ElasticsearchIndex index, String serviceName)
           throws IllegalArgumentException, IndexRebuildException {
-    IndexRebuildService.Service service = IndexRebuildService.Service.valueOf(serviceName);
+    rebuildIndex(index, serviceName, null);
+  }
+
+  public synchronized void rebuildIndex(ElasticsearchIndex index, String serviceName, ServicePart part)
+          throws IllegalArgumentException, IndexRebuildException {
+    IndexRebuildService.Service service = IndexRebuildService.Service.valueOf("AssetManager");
```
KatrinIhler: I think this isn't going to work, because it hardcodes the AssetManager also for the method above, which calls this one. Maybe this is a leftover from testing?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexRebuildService.java
```diff
@@ -72,6 +72,10 @@ public enum State {
     PENDING, RUNNING, OK, ERROR
   }
 
+  public enum ServicePart {
+    ACL
```
KatrinIhler: I really like this idea in general, giving us the option to only update certain data, which could maybe be expanded later. This is something that has really been missing so far.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
```diff
@@ -152,6 +153,57 @@ public Response partiallyRebuildIndex(@PathParam("service") final String service
     return R.ok();
   }
 
+  @POST
+  @Path("rebuild/AssetManager/{part}")
+  @RestQuery(name = "partiallyRebuildIndexByPart",
```
KatrinIhler, line 167: I think I would like a more general endpoint here, even though right now the AssetManager is the only one who really supports this feature. Maybe we could have `rebuild/{service}/{part}` and then either

a) have the service itself throw an error if the part is not supported, letting the rebuild service handle that error and return a reasonable return code, or
b) add a getSupportedParts() method to the interface all services implement and have the rebuildservice check that before calling repopulate()

The result is the same, so I don't really care. But this will make this more extendable later and should hopefully not be that much more work.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -941,6 +947,11 @@ public IndexRebuildService.Service getService() {
 
   @Override
   public void repopulate() throws IndexRebuildException {
+    repopulate(null);
```
KatrinIhler: Technically we don't need this method anymore, right? The rebuild service might as well pass the null everytime.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -976,7 +987,15 @@ public void repopulate() throws IndexRebuildException {
 
                 var updatedEventData = index.getEvent(snapshot.getMediaPackage().getIdentifier().toString(), orgId,
                     snapshotSystemUser);
-                updatedEventData = getEventUpdateFunction(snapshot, orgId, snapshotSystemUser).apply(updatedEventData);
+                if (IndexRebuildService.ServicePart.ACL.equals(type)) {
+                  // Only reindex ACLs
+                  updatedEventData = getEventUpdateFunctionOnlyAcl(snapshot, orgId, snapshotSystemUser)
+                      .apply(updatedEventData);
+                } else {
+                  // Reindex everything (default)
```
KatrinIhler: IMO all parts other than null and ACL should be rejected here. Otherwise a typo in a part will cause everything to be re-indexed, which might be... bad :D

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1611,4 +1626,45 @@ private Function<Optional<Event>, Optional<Event>> getEventUpdateFunction(Snapsh
       return Optional.of(event);
     };
   }
+
+  private Function<Optional<Event>, Optional<Event>> getEventUpdateFunctionOnlyAcl(Snapshot snapshot,
+      String orgId, User user) {
+    return (Optional<Event> eventOpt) -> {
+      MediaPackage mp = snapshot.getMediaPackage();
+      String eventId = mp.getIdentifier().toString();
+      Event event = eventOpt.orElse(new Event(eventId, orgId));
```
KatrinIhler, line 1641: Creating a new event with just the ACL is probably not a reasonable thing to do. Maybe we should better skip those and warn.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
```diff
@@ -631,8 +634,16 @@ private SolrQuery getForAction(SearchQuery q, String action, boolean applyPermis
       sb.append(q.getQuery());
     }
 
+    User user = securityService.getUser();
     String solrIdRequest = StringUtils.trimToNull(q.getId());
     if (solrIdRequest != null) {
+
+      // If user has ROLE_EPISODE_<ID>_READ/WRITE, no further permission checks are necessary
+      for (var suffix: ACL_ID_SUFFIX) {
+        if (user.hasRole(ACL_ID_PREFIX_EPISODE + solrIdRequest + suffix)) {
```
KatrinIhler: Shouldn't we check here only for the action getForAction was called with, instead of allowing both?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
```diff
@@ -760,13 +770,34 @@ private SolrQuery getForAction(SearchQuery q, String action, boolean applyPermis
...
         StringBuilder roleList = new StringBuilder();
+        var ids = new ArrayList<String>(roles.size());
         for (Role role : roles) {
-          if (roleList.length() > 0) {
-            roleList.append(" OR ");
-          }
-          roleList.append(Schema.OC_ACL_PREFIX).append(action).append(":").append(SolrUtils.clean(role.getName()));
-          if (role.getName().equalsIgnoreCase(securityService.getOrganization().getAnonymousRole())) {
-            userHasAnonymousRole = true;
+          var roleName = role.getName();
```
KatrinIhler: I'm confused. Why do we need to do this? Didn't we already set applyPermissions to false in this case further above?

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -308,8 +317,18 @@ private Optional<AccessControlList> loadAcl(final URI uri) {
   @Override
   public boolean hasPermission(final MediaPackage mp, final String action) {
     AccessControlList acl = getActiveAcl(mp).getA();
-    boolean allowed = false;
     final User user = securityService.getUser();
+    var allowed = false;
+
+    // Check special ROLE_EPISODE_<ID>_<ACTION> permissions
```
KatrinIhler, line 322: No idea how much this really matters performance-wise, but since we iterate over all roles of the user for the ACL check anyway, it might be better to compare with the id role there?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
KatrinIhler: Okay, the first concern was already brought up: I think this would generate duplicate roles, because we generate the new roles for each entry, and don't check whether any id roles already exists either (although the latter is probably unlikely?) This can create interesting effects for example in the Admin UI.

Second, granting read access for write access is something I feel unsure about. I know we do this in a couple of places, but for these roles, the ES index would be the only place where they are being handled this way, right? I don't know if we want that.

As Greg said, this simply won't work for any new actions that are introduced afterwards, unless an index update is triggered somehow. That's not great, but it is the way it is and as long as we document this, I think it's okay.

What I really don't like is depending on the existing ACL to set these new id roles. This will lead to inconsistencies since not every action needs to be present in every ACL, which will mean that the new roles will work for some events and not others. That's a support nightmare.

Custom actions are defined in the listprovider `acl.additional.actions.properties` so we should probably pull them from there. Read and write roles can always be generated. And we could add a comment to that config file to tell people they have to re-index the ACLs if they add a new action and use the id roles.

Also 

> I hope people don't have events with hundreds of ACL entries

Oh, some definitely do. :upside_down_face:

KatrinIhler (changes requested):
Alright, I had a couple of complaints, but all-in-all, I'm pretty happy with this approach. However, there might be a couple of things we haven't considered yet:

One thing that occurred to me: Did you test this with static file protection on? Because there are more permission checks for this, for example in AssetManagerStaticFileAuthorization and the SearchService that might not work with this new feature.

Additionally, the AssetManager checks permission internally via isAuthorized (might be relevant for snapshots). The WorkflowService checks permissions based on roles as well (relevant if users want to trigger a workflow from the editor maybe?). There is also AccessControlUtil used by the SearchService for... stuff??

Maybe some of this is not relevant for this or out of scope for this PR, but we should at least take a moment to think about these things.

...And we should really harmonize permission checking some time, bc yikes.

etc/org.opencastproject.assetmanager.impl.AssetManagerImpl.cfg
Arnei: Absolutely agree. If `custom.properties` works (and I'm allowed to clutter it with this), I'll gladly switch over to that.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
Arnei: While I generally agree, since this is not the main focus of this PR and I'm limited on time, I tried to not put too much effort into this while hopefully keeping it easily extendable for the future. But if you say your suggestion should really be part of this PR (for example to avoid changing the endpoint path down the line), I'll get to it.

etc/org.opencastproject.assetmanager.impl.AssetManagerImpl.cfg
KatrinIhler: I think in this case it's warranted. Another possibility would be the org config (so each tenant can turn that off and on), but I'm a bit fuzzy on the differences right now. (Iirc one gets hotloaded on change and the other requires a restart, maybe? But that might be out-of-date knowledge.)

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
KatrinIhler: I think it would be good to get this right from the get-go. Even though we don't guarantee anything for these endpoints, changing them sometime down the line is always a hassle.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
Arnei: I'm not sure what you want to skip. The event created here is for updating the event in the index, and since we only want to change the ACL, it only has the ACL.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
Arnei: Maybe? I'm not terribly clear on how Solr queries are supposed to work tbh.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
Arnei: `applyPermissions` can still be `true` if the query did not contain an id , see line 639.

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
Arnei: The check only iterates over the user roles once, instead of over all user roles for every ACE like usual, so I'd wager there should not be much of a performance difference. But I honestly have no idea, would have to run benchmarks.

KatrinIhler: Yeah, no hard feelings here, feel free to skip.

Arnei:
Aaand I just noticed that new events are not getting the custom roles >.< Will look into that.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
Arnei: Tried to rewrite the endpoint. Currently confused by https://github.com/opencast/opencast/blob/ccfe8dd415bbeea3b1f8a8ab1bda45b4ff2b54d2/modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexRebuildService.java#L212, which eats up exceptions during index rebuilds for reasons unknown to me. Will try to look further into this some other time. Furthermore, checking ServicePart in IndexEndpoint is not really possible anymore if ServicePart is moved to the actual service due to code architecture, so currently the endpoint always returns 200, no matter if the part was valid or not.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
KatrinIhler: Instead of creating a new event in that case, you could return an empty Optional. That's the way these methods are supposed to work. But apparently the repopulate() method doesn't handle that case correctly. :(

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
KatrinIhler: Took a further look at the code, and yeah, I think so. Otherwise a user with read permissions might gain write permissions here, which might be bad.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
KatrinIhler: Ah, I see, I overlooked that part, thx. In that case, I only think that we need to consider the action parameter here as well, instead of allowing both every time.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -862,4 +896,41 @@ public String escapeQuery(String text) {
...
     return sb.toString();
   }
+
+  private Event updateEventAclWithCustomRoles(Event event) {
+    AccessControlList acl = new AccessControlList();
+
+    try {
+      acl = AccessControlParser.parseAcl(event.getAccessPolicy());
+    } catch (Exception e) {
+      logger.error("Unable to parse access policy", e);
```
KatrinIhler: I wouldn't swallow this here. Better to let the whole thing fail imo. Maybe wrap it in SearchIndexException.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/ElasticsearchIndex.java
```diff
@@ -862,4 +896,41 @@ public String escapeQuery(String text) {
...
+    customEntries.add(new AccessControlEntry("ROLE_EPISODE_" + event.getIdentifier()
+        + "_" + "WRITE", "write", true));
+
+    ResourceListQuery query = new ResourceListQueryImpl();
+    if (listProvidersService.hasProvider("ACL.ACTIONS")) {
+      Map<String, String> actions = new HashMap<>();
+      try {
+        actions = listProvidersService.getList("ACL.ACTIONS", query, true);
+      } catch (ListProviderException e) {
+        throw new RuntimeException("Listproviders not loaded. " + e);
```
KatrinIhler: This should be something catchable imo, like SearchIndexException.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
KatrinIhler: Looks like this is resolved, too? @Arnei ?

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
KatrinIhler: I just took a look at this. Swallowing the exception there is a bug imo, that was introduced by #4206. All the surrounding code expects that exception to bubble up. Just throwing that exception again after setting the rebuild state should solve your problem.

KatrinIhler (commented):
New changes look good except for minor things. I also looked a bit further into the existing code. 

I guess we could've also added the id roles for each snapshot to handle new and updated events, but I don't have a clear preference for where this code should live.

In either case, the id roles will show up for existing events in the Admin UI and API, bc they're in the ES index, right? I guess we could filter them out prior, but this is probably true for ROLE_ADMIN as well, right now... Let's see if anyone is confused by this, first. Especially bc you can't really delete them. :sweat_smile: 

Some stuff from my old review is still outstanding. (This is more a reminder for me than for you, so I remember what to look at when I review the next time.)

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
Arnei: I can certainly throw an exception there, but this will not solve the problem that the index rebuild is running async (I think?), so I can't have it influence the return code.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
Arnei: Well, in case that that's how they are supposed to work, it wasn't working like that before, so I didn't break anything :x. So for the sake of consistency, leave it as it is for now? And address it in another PR?

KatrinIhler: Yeah, I think these functions don't actually make any sense the way they are used here, but that's not your fault. I'll add it to the pile...

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
KatrinIhler: Ah, of course, you're right about that. Sorry, that slipped my mind. Well, at least the UI will show you if it went wrong... IMO that exception should still be thrown, though. Adding to the list... I should stop looking at that code.

Arnei:
> Looks like this is resolved, too? @Arnei ?

Uuh, remind we what we were talking about here? ^^'

KatrinIhler:
> > Looks like this is resolved, too? @Arnei ?
> 
> Uuh, remind we what we were talking about here? ^^'

Never mind, I'll check for the next review pass. :D

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -940,7 +945,13 @@ public IndexRebuildService.Service getService() {
   }
 
   @Override
-  public void repopulate() throws IndexRebuildException {
+  public void repopulate(String type) throws IndexRebuildException {
```
gregorydlogan, line 948: This spits out an NPE when you trigger a rebuild using `http://localhost:8080/index/rebuild`.  This is due to [this line](https://github.com/opencast/opencast/pull/5056/files#diff-3cb53a562e43f510142236fdfc62041d78ebfdb445656a1c99f187de65738b08R156), which is passing a `null` which isn't properly handled here.

This should read
```suggestion
  public void repopulate(String type) throws IndexRebuildException {
    ServicePart parsedType = null;
    if (type != null && EnumUtils.isValidEnum(ServicePart.class, type)) {
      parsedType = ServicePart.valueOf(type);
    else {
      throw new IndexRebuildException("The given type " + type + " was not valid. Should be null or "
          + ServicePart.values());
    }
```

Caution: Above *probably* has spacing issues, this UI is not helpful with multiline suggestions.  Also don't forget the next suggestion below!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -976,7 +987,18 @@ public void repopulate() throws IndexRebuildException {
 
                 var updatedEventData = index.getEvent(snapshot.getMediaPackage().getIdentifier().toString(), orgId,
                     snapshotSystemUser);
-                updatedEventData = getEventUpdateFunction(snapshot, orgId, snapshotSystemUser).apply(updatedEventData);
+                if (ServicePart.ACL.equals(parsedType)) {
```
gregorydlogan: Reverse the order of checks here to adapt to the suggestion about.  If `type` is `null`, we don't need to worry about `parsedType` at all anymore.  If it's not then `parsedType` won't be `null`, at the very least.

gregorydlogan (changes requested):
Caught a last minute bug.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -940,7 +945,13 @@ public IndexRebuildService.Service getService() {
   }
 
   @Override
-  public void repopulate() throws IndexRebuildException {
+  public void repopulate(String type) throws IndexRebuildException {
+    if (type != null && !EnumUtils.isValidEnum(ServicePart.class, type)) {
+      throw new IndexRebuildException("The given type " + type + " was not valid. Should be null or "
+          + ServicePart.values());
+    }
+    ServicePart parsedType = ServicePart.valueOf(type);
```
gregorydlogan: This spits out an NPE when you trigger a rebuild using http://localhost:8080/index/rebuild. This is due to [this line](https://github.com/opencast/opencast/pull/5056/files#diff-3cb53a562e43f510142236fdfc62041d78ebfdb445656a1c99f187de65738b08R156), which is passing a null which isn't properly handled here.

Suggested change
```suggestion
    ServicePart parsedType = null;
    if (type != null && EnumUtils.isValidEnum(ServicePart.class, type)) {
      parsedType = ServicePart.valueOf(type);
    }
```

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/solr/SolrRequester.java
```diff
@@ -631,8 +634,14 @@ private SolrQuery getForAction(SearchQuery q, String action, boolean applyPermis
       sb.append(q.getQuery());
     }
 
+    User user = securityService.getUser();
     String solrIdRequest = StringUtils.trimToNull(q.getId());
     if (solrIdRequest != null) {
+
+      // If user has ROLE_EPISODE_<ID>_READ/WRITE, no further permission checks are necessary
+      if (user.hasRole(ACL_ID_PREFIX_EPISODE + solrIdRequest + action.toUpperCase())) {
```
gregorydlogan: Oof.  Subtle bug is subtle.  And annoying.

```suggestion
      if (user.hasRole(ACL_ID_PREFIX_EPISODE + solrIdRequest + "_" + action.toUpperCase())) {
```

gregorydlogan (changes requested):

Arnei:
Uff, thanks for catching those Greg. Should all be fixed now.

KatrinIhler:
To do: Re-target to 16

Arnei:
Now points to 16. Means we can drop changes to Solr, since Solr is gone. Instead, this now updates the Search Index in Elasticsearch in a similar fashion to the event index. Not yet sure how to make an ACL-update-only version for fast reindexing of the Search index, since the current code is very "all or nothing" with its indexing approach.

Also currently cannot view publications or start workflows on events (that the current user has an episode role id for). Looking into that.

Arnei:
Can now view publications, but not view assets in event details or start workflows. Both seem to be related to an issue with ASelectQuery generating different queries for episode id roles users, but I am having trouble figuring out the details.

Arnei:
So ASelectQuery generates an authorization predicate prior to generating the query, based on whether the user is admin, their organization and roles. Unfortunately, it does so before any episode id comes into play (and for some queries an episode id never comes into play), so we can't *just* circumvent it. To circumvent it, we would have to rewrite the query logic somewhat, but that would be a major pain as ASelectQuery is used in quite a few different places and based off of entwine code. Not sure how to proceed here.

KatrinIhler:
> So ASelectQuery generates an authorization predicate prior to generating the query, based on whether the user is admin, their organization and roles. Unfortunately, it does so before any episode id comes into play (and for some queries an episode id never comes into play), so we can't _just_ circumvent it. To circumvent it, we would have to rewrite the query logic somewhat, but that would be a major pain as ASelectQuery is used in quite a few different places and based off of entwine code. Not sure how to proceed here.

My impression is that for some of these things, extending `isAuthorized()` in AssetManagerImpl should already be enough, for example getting assets? And that should be do-able. 

As for the authPredicate, one could extend it, I _think_ similarly to 
```
.map((role) -> q.property(Value.BOOLEAN, SECURITY_NAMESPACE, mkPropertyName(role.getName(), action))
                    .eq(true))
```
but by mapping the special roles to `q.mediapackageId().eq(<the id extracted from the special role>)` and connect that with an "or" to the existing conditions. It's pretty nasty work because of the custom query language, but if I understand things correctly, it should be possible.

(The other way would be to add the special roles to the properties when a snapshot is created which the authPredicate then compares to, similar to the way we "solved" Elasticsearch. It has the same disadvantage though - it only works for new snapshots, and there's no "re-indexing" functionality for this so far.)

Once this is working, we'll see if workflows then work, and static file protection (but I guess since it's on by default, you kinda already tested this). And then I'm actually gonna do some testing and not just stare at code. :D

And I was also wondering whether we can put this into 16.x anymore, but that's a discussion for later.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1047,6 +1056,9 @@ public HttpAssetProvider getHttpAssetProvider() {
    * Security handling
    */
 
+  //  .map((role) -> q.mediapackageId().eq(StringUtils.substringBetween(role.getName(), ACL_ID_PREFIX_EPISODE, "_"))
```
KatrinIhler: Leftover.

KatrinIhler (commented):
Change looks reasonable, just one itty-bitty thing. Did you already test this? Does this make workflows work as well? :D

Arnei:
My episode id role user was able to view and download Assets, as well as start a workflow.

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexProducer.java
```diff
@@ -27,9 +27,10 @@
 public interface IndexProducer {
 
   /**
-   * Re-add all data of this service to the index.
+   * Re-add the data of this service to the index for the given type.
+   * R-add ALL data of this service if null is passed instead.
```
JulianKniephoff:
```suggestion
   * Re-add ALL data of this service if null is passed instead.
```

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexRebuildService.java
```diff
@@ -206,10 +216,11 @@ private void rebuildIndex(ElasticsearchIndex index, IndexRebuildService.Service
     logger.info("Starting to rebuild the {} index", service);
     setRebuildState(service, IndexRebuildService.State.RUNNING);
     try {
-      indexProducer.repopulate();
+      indexProducer.repopulate(part);
```
JulianKniephoff: Why is this called `part` here and `type` everywhere else? 🤔

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexProducer.java
JulianKniephoff: But actually probably more like

```suggestion
   * Re-add the data of this service to the index.
   * @param type Limit the data added to the index to the given type. Re-add all data if <code>null</code> is given.
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
+Episode ID roles
+=================================
+
+If activated, users with a role like ROLE_EPISODE_<ID>_<ACTION> will have access to the episode with the given
+identifier, without this having to be specifically specified in the ACL attached to the episode.
+
+For example, ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ will allow the user to get read access to
```
JulianKniephoff:
```suggestion
For example, `ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ` will allow the user to get read access to
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
+Episode ID roles
+=================================
+
+If activated, users with a role like ROLE_EPISODE_<ID>_<ACTION> will have access to the episode with the given
```
JulianKniephoff:
```suggestion
If activated, users with a role like `ROLE_EPISODE_<ID>_<ACTION>` will have access to the episode with the given
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
...
+identifier, without this having to be specifically specified in the ACL attached to the episode.
```
JulianKniephoff:
```suggestion
identifier, without this having to be explicitly stated in the ACL attached to the episode.
```

Replacing both instances of a repetition just because I can 😎

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
...
+
+For example, ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ will allow the user to get read access to
+episode 872dc4ec-ca8a-4e12-8dac-ce99784d6d29.
```
JulianKniephoff:
```suggestion
episode `872dc4ec-ca8a-4e12-8dac-ce99784d6d29`.
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
...
+
+Setup
+--------------------
+
+Enable `episode.id.role.access` in `etc/custom.properties`.
+
+To make this work for the Admin UI and External API, the Elasticsearch Index needs to be updated with modified
+ACLs. You can achieve this by calling the /index/rebuild/AssetManager/ACL endpoint AFTER enabling this feature
```
JulianKniephoff:
```suggestion
ACLs. You can achieve this by calling the `/index/rebuild/AssetManager/ACL` endpoint **after** enabling this feature
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
...
+in the aforementioned configuration files.
+The endpoint will reindex only event ACLs.
+
+In case that you have custom actions configured, this will only work for the actions that were configured during the
```
JulianKniephoff:
```suggestion
In case you have custom actions configured, this will only work for the actions that were configured during the
```

docs/guides/admin/docs/configuration/episode-id-roles.md
```diff
@@ -0,0 +1,21 @@
...
+reindex of the Elasticsearch Index. If you later add custom actions, you will have to reindex again.
```
JulianKniephoff:
```suggestion
reindex of the Elasticsearch index. If you later add custom actions, you will have to reindex again.
```

etc/custom.properties
```diff
@@ -356,3 +356,17 @@ karaf.pid.file=${karaf.data}/pid
...
+# Allow episode ID based access control via roles.
+# If activated, users with a role like ROLE_EPISODE_<ID>_<ACTION> will have access to the episode with the given
+# identifier, without this having to be specifically specified in the ACL attached to the episode.
+#
+# For example, ROLE_EPISODE_872dc4ec-ca8a-4e12-8dac-ce99784d6d29_READ will allow the user to get read access to episode
+# 872dc4ec-ca8a-4e12-8dac-ce99784d6d29.
+#
+# To make this work for the Admin UI and External API, the Elasticsearch Index needs to be updated with modified
+# ACLs. You can achieve this by calling the /index/rebuild/AssetManager/ACL endpoint AFTER activating this feature.
+# The endpoint will reindex only event ACLs.
```
JulianKniephoff: Since this is the same content as in the guide, the same suggestions apply.

JulianKniephoff (changes requested):
Mini drive-by review, maybe more later. 🙈

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/rebuild/IndexRebuildService.java
Arnei: No reason I suppose.

KatrinIhler (changes requested):
Alright, finally my updated review. First the things I tested that work:

- Full and partial index rebuild updates ACLs in index correctly
- New event gets new roles
- User can play the video in Engage
- User can cut the video in the new Editor
- User can start a workflow on an existing event
- User can edit ACLs and metadata

Potential stumbling block we might want to document:

The - in an action are not changed to _ for the role, so it's `ROLE_EPISODE_<ID>_ANNOTATE-ADMIN`.

Also we still don't get an error for a wrong part for the partial index rebuild, but if I remember correctly that was complicated to do?

Biggest remaining problem: 

The new roles are shown in the Admin UI (because they are in the index, the same is true for `ROLE_ADMIN`), and while deleting them doesn't really do something because they will reappear when opening the event details again, one _can_ change their actions, which is potentially very confusing.  (Edit: This is probably caused by the ACL merging.

So either we make sure the actions are always correct when updating the ACL in the index, _or_, maybe my preferred solution, we always filter out those roles when fetching the event from the index because they're supposed to be implicit (and for example also aren't part of the archived ACL). (Edit: Although this might have other consequences, because it could be that some things e.g. in the Workflow Service working right now is precisely _because_ the roles are in the ACL for the indexed event :scream: )



Edit: This should be re-pointed to develop. Also conflicts. :D

KatrinIhler:
Oh, and in regards to testing: I tested this by creating an unprivileged user and giving it the new roles for an event by either configuring those roles in `custom.properties` first so I can select them in the UI or by creating a group via the API so I can set arbitrary roles (probably also possible for the user directly).

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -348,14 +352,18 @@ public void storeMediaPackage(MediaPackage mediaPackage, AccessControlList acl,
...
           // Ensure this user is allowed to update this media package
-          String accessControlXml = entity.get().getAccessControl();
-          if (accessControlXml != null && entity.get().getDeletionDate() == null) {
-            AccessControlList accessList = AccessControlParser.parseAcl(accessControlXml);
-            User currentUser = securityService.getUser();
-            Organization currentOrg = securityService.getOrganization();
-            if (!AccessControlUtil.isAuthorized(accessList, currentUser, currentOrg, WRITE.toString())) {
-              throw new UnauthorizedException(currentUser + " is not authorized to update media package "
-                  + mediaPackageId);
+          // If user has ROLE_EPISODE_<ID>_READ/WRITE, no further permission checks are necessary
```
KatrinIhler: This only checks write.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -84,6 +84,8 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
   /** Logging utilities */
   private static final Logger logger = LoggerFactory.getLogger(SearchServiceDatabaseImpl.class);
 
+  private static final String ACL_ID_PREFIX_EPISODE = "ROLE_EPISODE_";
```
KatrinIhler: If you still have time, maybe shove those somewhere with the other constants like ROLE_ADMIN and maybe even create a method to build such a role instead of every service having to concatenate strings together (but the latter might be too much work).

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
```diff
@@ -278,6 +298,31 @@ private void indexMediaPackage(MediaPackage mediaPackage, AccessControlList acl,
...
+      customEntries.add(new AccessControlEntry("ROLE_EPISODE_" + mediaPackageId + "_" + "READ", "read", true));
+      customEntries.add(new AccessControlEntry("ROLE_EPISODE_" + mediaPackageId + "_" + "WRITE", "write", true));
+
+      ResourceListQuery query = new ResourceListQueryImpl();
+      if (listProvidersService.hasProvider("ACL.ACTIONS")) {
+        Map<String, String> actions = new HashMap<>();
+        try {
+          actions = listProvidersService.getList("ACL.ACTIONS", query, true);
+        } catch (ListProviderException e) {
+          throw new RuntimeException("Listproviders not loaded. " + e);
```
KatrinIhler: This should be an exception that will be caught like SearchException.

KatrinIhler (commented):
Some nitpicks after staring at the code again.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
Arnei: Do you perhaps have a recommendation on where the constants/methods could live? I'd like to avoid creating unnecessary dependencies between asset manager and search if possible. Maybe common?

Arnei: Put them in common.

KatrinIhler:
Re: no errors from the index endpoint when the wrong data type for re-indexing is entered, see https://github.com/KatrinIhler/opencast/commit/ac5f74a2f9882c03b664bb6b897615f19461a215. This compiles fine, but i didn't test it (because this took way longer than I anticipated...)  Could you take a look at this and pull it in if it works?

Basically what this does is check the supported services and data types before entering the asynchronous part of the code (and some very minor cleanup that I couldn't resist).

KatrinIhler (changes requested):
Short version of my messages: I think the ACL that we get from the index should not contain the new roles. If that leads to breakage in e.g. the Workflow Service, we should extend those permission checks in the same way we did for e.g. the AssetManager and similarly to how we treat ROLE_ADMIN.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -172,12 +173,14 @@ public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws
...
-            throw new UnauthorizedException(currentUser + " is not authorized to delete media package "
-                + mediaPackageId);
+        User user = securityService.getUser();
+        if (!user.hasRole(getEpisodeRoleId(mediaPackageId, WRITE.toString()))) {
+          if (!(searchMp.isLive() && currentUser.hasRole(GLOBAL_CAPTURE_AGENT_ROLE)) && accessControlXml != null) {
+            AccessControlList acl = AccessControlParser.parseAcl(accessControlXml);
+            if (!AccessControlUtil.isAuthorized(acl, currentUser, currentOrg, WRITE.toString())) {
+              throw new UnauthorizedException(
+                  currentUser + " is not authorized to delete media package " + mediaPackageId);
+            }
```
LukasKalbertodt: This check and the ones below don't seem to check the config value that controls whether these special roles are activated?

Also: can this extra logic not move into `AccessControlUtil.isAuthorized`? Seems like thats the perfect place for it and would make all changes in this file unnecessary I think.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
```diff
@@ -274,6 +296,31 @@ private void indexMediaPackage(MediaPackage mediaPackage, AccessControlList acl,
...
+          throw new SearchException("Listproviders not loaded. " + e);
+        }
+        for (String action : actions.keySet()) {
+          customEntries.add(
+              new AccessControlEntry(getEpisodeRoleId(mediaPackageId, action), action, true));
+        }
+      }
+
+      AccessControlList customRoles = new AccessControlList(new ArrayList<>(customEntries));
+      acl = customRoles.merge(acl);
```
LukasKalbertodt, line 321: This looks very very similar to the code in `ElasticsearchIndex::updateEventAclWithCustomRoles`

LukasKalbertodt (commented):
I haven't followed this PR in detail, but I have a comment regarding the "implicit roles in index cause trouble".

I don't think implicit roles should be stored in the index at all. Especially if this would require us to touch multiple places to deal with those implicit roles. Shouldn't it be possible to *only* change one or very few central methods that would adjust the query sent to ES? (Also see the one inline comment about `AccessControlUtil.isAuthorized`.

This would get rid of the explicitly stored roles and also the need to have a list provider with all actions that should be written to the index. Adding new actions later would require an index rebuild as far as I understand.

I haven't figured all of this out yet, but I wanted to quickly drop this comment here.

LukasKalbertodt:
Just for clarification: feel free to ignore my previous comments, it might be harder than I think and I really don't know a lot about this part of Opencast. I just want this feature in OC :P

Arnei:
>  I think the ACL that we get from the index should not contain the new roles. If that leads to breakage in e.g. the Workflow Service, we should extend those permission checks in the same way we did for e.g. the AssetManager and similarly to how we treat ROLE_ADMIN.

Latest commit should achieve that. It still puts the roles in the index, but not in the ACL in the index, but in other ACL-related fields that are used for permission checks (I think).

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
Arnei: Good point, we could also do that. ([ff61800](https://github.com/opencast/opencast/pull/5056/commits/ff61800f8c2014558062e9dceb256ca0d8a73314))

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
Arnei: `ElasticsearchIndex::updateEventAclWithCustomRoles` is gone now, so this is arguably outdated.

Arnei:
> Shouldn't it be possible to only change one or very few central methods that would adjust the query sent to ES?

That would be very nice, but I could not come up with a solution on how to achieve that.

gregorydlogan:
This works as expected, so I'm going to merge it.  Found https://github.com/opencast/opencast-admin-interface/issues/903 in the process of testing this (again), but it does not block the backend feature's use.

JulianKniephoff:
#6500 should have been raised during review, but better late than never! 😅

PR #5100, opened 2023-07-04
docs: Fix NGINX configuration template preventing alias traversals

ziegenberg:
We provide a documentation example of a potentially insecure Nginx configuration that may allow alias traversal. According to https://labs.hakaioffsec.com/nginx-alias-traversal/ when there is an `alias` directive present within the `location` directive, then the `location` directive should end with a trailing slash in its path and the `alias` directive should end with a trailing slash as well.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests


lkiesow (approved):
While not really security-relevant in this case since it is an `internal` location and the URL is generated by Opencast, it's still an interesting find and there is no reason to not fix this. Merging.

Issue #5099, opened 2023-07-03
Feature request: Endpoint to get Licenses & Languages in metadata


ferishili:
**Description**
So far all major Opencast Plugins in LMSes provide available licenses and languages themselves (Hardcoded) and basically is a copy and paste of those in Opencast server, which is not a good practice.

**Feature**
That would be really great, if we could have a listing endpoint under `/api` to get the list of listables like languages and licenses.

**Expected**
I should be able to call to an endpoint like `/api/listing/{value}` where possible values could be `languages` or `licenses` and get the exact listing as provided in adminUI.
Of course in `key:value` pairs

**Nice to have**
Would be much nicer if it could handle the localization as well
E.g. passing 'de' returns the list in Germen.

mtneug:
@Arnei can this be closed?

Arnei:
Yup, can be closed

PR #5102, opened 2023-07-06
Add endpoint to get languages & licenses

Arnei:
Attempts to resolve https://github.com/opencast/opencast/issues/5099.

Adds two new endpoints to the external API. They return the languages/licenses configured in the listproviders.

The endpoint behaviour is fairly straightly copied from their Admin UI endpoint counterpart in `admin-ng/resources`. The key difference is that `admin-ng/resources` allows for request for any provider, whereas this commit created two explicit endpoints for two explicit providers. The reasoning is that users of the external api should not have access to every provider, and that most providers only make sense in the context of the Admin UI anyway. I'd be happy to hear other opinions on this.

Another issue is that both the language & license endpoint return translation identifiers that are only useful to the Admin UI instead of more useful labels (e.g. `LANGUAGE.ARABIC` instead of `Arabic`). And I don't see a clean way of getting those translations from the Admin UI, short of copying the entire translation to the external-api module (like the lti module did).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/developer/docs/api/listprovider-api.md
+++ b/docs/guides/developer/docs/api/listprovider-api.md
@@ -1,0 +1,52 @@
+[TOC]
+
+# Information
+
+The ListProvider API is available since API version 1.10.0.
+
+### GET /api/listproviders/languages
+
+Returns a list of languages as configured in the listproviders.
+
+__Response__
+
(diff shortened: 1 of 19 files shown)


Arnei:
Technical meeting concluded that it would be good if this could be extended to all providers. No privacy concerns.

PR #5115, opened 2023-07-14
Fix index rebuild

geichelberger:
This PR fixes a bug in that the last eight events of the current rebuild batch were not indexed because it never reached the required bulk update size and got reset afterward.

>1000 contains 62 bulk updates of size 16, but it only updates if the updatedEventRange size is greater than equals 16, which means the last batch of size 8 will be skipped. 

Maybe controversial: I changed all the bulk updates to 20 for consistency.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -955,12 +955,12 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
       RichAResult r;
       int current = 0;
       logIndexRebuildBegin(logger, index.getIndexName(), total, "snapshot(s)");
+      var updatedEventRange = new ArrayList<Event>();
       do {
         r = enrich(q.select(q.snapshot()).where(q.version().isLatest()).orderBy(q.mediapackageId().desc())
           .page(offset, PAGE_SIZE).run());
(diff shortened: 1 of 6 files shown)


Arnei:
Can't reproduce the issue. In my test installation, all of my 42 events get indexed. Probably because it not only checks that `updatedEventRange.size() >= n`, but also that `current >= total` which should always eventually occur since all events get indexed? Though the whole `current` counting seems funky if you have multiple tenants, maybe that's an issue? (Admittedly haven't looked at the code too closely).

geichelberger:
You need more than 1000 events.

ziegenberg:
This is a problem with only huge OC instances. For every 1000 events, 8 events go missing. This bug is 4 months old. See #4783.

KatrinIhler:
So do I understand this correctly that while you change the patch size for every service, the actual bug only occurred in the AssetManager? (Asking because I'm thinking about how to fix the broken events without doing a whole index rebuild.)

geichelberger:
Yes, the actual bug is only in the AssetManager. The other service that uses pagination is not affected.

KatrinIhler:
> The other service that uses pagination is not affected.

I'm pretty sure the Workflow Service is also affected. It also uses pagination with pages of 1000, and doesn't carry over the instances contained in the last batch that doesn't match the batch size.

Additionally setting the batch size to 20 doesn't fix this for the Workflow Service because not all instances of a page are actually indexed, so the last batch can be less than 20 as well.

Could you take a look at #5226 to confirm this? :)

PR #5116, opened 2023-07-17
Only show Asset Upload for options of correct type

KatrinIhler:
Currently the Asset Upload link in the event details is always shown because ng-show also considers the source options, even though these are then not displayed, leading to an empty form with a non-functional button. If we don't have options configured for the actual asset upload, we shouldn't show the link in the first place.


Arnei (approved):

KatrinIhler:
Just noticed I used the wrong source branch and since I can't change that, i created #5116 instead. @Arnei Would you mind approving that one?

PR #5117, opened 2023-07-18
Only show Asset Upload for options of correct type

KatrinIhler:
Currently the Asset Upload link in the event details is always shown because ng-show also considers the source options, even though these are then not displayed, leading to an empty form with a non-function button. If we don't have options configured for the actual asset upload, we shouldn't show the link in the first place.

Edit: This was unintentionally broken by #3150.


KatrinIhler:
Same as #5116 but this time with the correct source branch...

Arnei (approved):

PR #5121, opened 2023-07-20
Paella 7: Add default layout for unknown track flavours and multiple streams

### LINKED ISSUE #5107: The default player should gracefully handle multiple streams
Ingesting multiple streams or streams with unknown flavors, the player shouldn't just give up, but try to nicely render the existing streams. Right now. Paella will just show a massive error.

### Example

Ingest a new event like this:

```
SERVER=https://develop.opencast.org

curl -u admin:opencast "${SERVER}/ingest/addMediaPackage/fast" \
        -F 'flavor=a/source' \
        -F 'flavor=b/source' \
        -F 'flavor=c/source' \
        -F title="Three Video Streams" \
        -F creator="Test"
```

Wait for it to be processed and go to the player. You get nothing but a big error message:

Screenshot from 2023-07-12 14-31-38

The player does that even though valid video streams are published:

```
❯ curl -s -u admin:opencast 'https://develop.opencast.org/search/episode.json?id=0660c600-78cd-43a9-a771-cc4e7770353c' | jq -r '.["search-results"].result.mediapackage.media.track[] | .type + "  ->  " + .url'
c/preview  ->  https://develop.opencast.org/static/mh_default_org/engage-player/0660c600-78cd-43a9-a771-cc4e7770353c/a73fabaf-9f44-4877-8399-7a07a1ffa2e6/goat.mp4
b/preview  ->  https://develop.opencast.org/static/mh_default_org/engage-player/0660c600-78cd-43a9-a771-cc4e7770353c/96aa5320-7f3f-4179-918b-05d6fb6f229a/goat.mp4
a/preview  ->  https://develop.opencast.org/static/mh_default_org/engage-player/0660c600-78cd-43a9-a771-cc4e7770353c/9208f55d-88e9-411c-93a7-59b07354c559/goat.mp4
```

### Improvement

As a last resort, before giving up, the player should engage an auto mode:

- Get all tracks
- Filter our audio-only tracks
- Group tracks by flavor/type
- Show one video of each flavor in the player

Long term, we might even make this the default

dennis531:
The changes weaken the requirement of predefined flavours by providing a default layout named auto. The default layout works with any number of tracks and arbitrary flavours. The default layout is always selectable in the layout selector.

The fix is more a hack, as the flavors are hardcoded and required in the layout configuration. Thus, the paella player is not designed for arbitrary flavors and multiple streams. In my opinion, this problem could be solved by removing the restriction of fixed/valid flavours in the paella core. 

Fix #5107


modules/engage-paella-player-7/src/plugins/org.opencast.paella.multiVideoDynamicLayout.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  getValidContentIds(streamData) {
+    // Ignore content of streamData
+    return this.validContentIds;
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  getLayoutStructure(streamData, contentId) {
```
lkiesow: Can't you just not define unnecessary parameters? I remember that you can have more parameters in a JS function than are defined. So, this should work:

```suggestion
  getLayoutStructure(streamData) {
```

etc/ui-config/mh_default_org/paella7/config.json
```diff
@@ -228,6 +222,18 @@
...
+            "enabled": true,
+            "validContent": [
+                {
+                    "id": "multi-video",
+                    "content": [],
+                    "icon": "present-mode-4.svg",
+                    "title": "Auto"
+                }
+            ],
+            "tabIndexStart": 40
```
lkiesow, line 235: I don't really understand these configuration options.
Icon doesn't seem to be used. Is it necessary?

The title seems to be what's being shown in the video layout selector. That may make sense to configure:



But everything else could be just hard-coded configuration. Or does it make sense to configure this array differently at some point?

So, you could just have something like this as configuration which is less confusing for users:
```json
{
  "org.opencast.paella.multiVideoDynamicLayout": {
    "enabled": true,
    "title": "Auto",
    "icon": "present-mode-4.svg"  ← not sure about this?
  },
  "tabIndexStart": 40
}
```

lkiesow (changes requested):
At first glance, this already looks better than what we have right now :+1: 

One thing which is still a bit weird is that sometimes you get a video selector and sometimes you don't. This may be an old problem, and it may sense to address this in a separate pull request instead? What do you think?

Screenshot from 2023-08-08 15-24-15

Also, if we change this, instead of generating all possible combinations like this:

- A, B, C
- A, B
- A, C
- B, C
- A
- B
- C

(And now imagine that with four videos :grin:)
It would be far easier to just be able toselect the different videos:

- [x] A
- [ ] B
- [x] C

…but that really might be out of scope for this pull request.

etc/ui-config/mh_default_org/paella7/config.json
miesgre: I'm on vacation right now, I'll review this PR when I get back.

> Icon doesn't seem to be used. Is it necessary?
> 
> The title seems to be what's being shown in the video layout selector. That may make sense to configure:

Just to clarify, icons are not used in default settings, but this can be changed in the `config.json` file
https://github.com/opencast/opencast/blob/f4454f61f2f66c9166384fee13b27d0e86df42cb/etc/ui-config/mh_default_org/paella7/config.json#L347-L357

dennis531: I kept `content` because I am not sure if this could break something. In https://github.com/polimediaupv/paella-core/blob/ea62b0b9a5a99d8344069d20eb42d03e143d35d5/src/js/core/VideoLayout.js#L38C80-L38C80 this config is used. Without this configuration the player with the default settings seems to work.

dennis531:
> One thing which is still a bit weird is that sometimes you get a video selector and sometimes you don't. This may be an old problem, and it may sense to address this in a separate pull request instead? What do you think?

The layout selector is visible if the provided videos match the configured flavours/contents of at least two layout configs. I am not sure if the current behavior is a problem.

dennis531:
> Also, if we change this, instead of generating all possible combinations like this:
> 
>     * A, B, C
> 
>     * A, B
> 
>     * A, C
> 
>     * B, C
> 
>     * A
> 
>     * B
> 
>     * C
> 
> 
> (And now imagine that with four videos grin) It would be far easier to just be able toselect the different videos:
> 
>     * [x]  A
> 
>     * [ ]  B
> 
>     * [x]  C
> 
> 
> …but that really might be out of scope for this pull request.

If I understand it correctly, the layout selector should be replaced by a video flavor selector?
In this case, we need to move away from the layout idea and possibly need to adapt the paella-core.

lkiesow:
Okay, I dug a bit deeper and I now (partially) understand the layout selector problem.
While I don't like it, I acknoledge that it exceeds the bounds of this pull request.
This is an improvement over the current situation and I think we can merge it.

IThere is a test which fails. I think that is because @gregorydlogan slightly changed the default headers.
Can you do a quick rebase and fix the header?

For the other player-complaints, I will make sure to file a few separate issues :grin:

lkiesow:
Thanks for fixing the merge conflict.
Tested again and works as expected (with the known caveats)
Merging

karendolan:
@dennis531  what is the expected behavior of the "auto" selection at the end of the generated layout list?

```
* A, B, C
* A, B
* A, C
* B, C
* A
* B
* C
* Auto

dennis531:
> @dennis531 what is the expected behavior of the "auto" selection at the end of the generated layout list?
> 
> ```
> * A, B, C
> * A, B
> * A, C
> * B, C
> * A
> * B
> * C
> * Auto
> ```

The `auto` layout is the same as A, B, C. Therefore, all available videos/flavours should be displayed. I hope that this answers your question.

PR #5125, opened 2023-07-28
Logging nitpicks

JulianKniephoff:
Our build currently produces the following warning (among others ... 😉):

```
[WARNING] The artifact org.slf4j:slf4j-log4j12:jar:1.7.36 has been relocated to org.slf4j:slf4j-reload4j:jar:1.7.36
```

I don't like warnings, so I used this as an excuse to learn a little bit about our logging setup. In the end it turned out that we can really just replace the former dependency with the latter to get rid of the warning, because that is literally already what was happening anyway.

This PR also includes some other cleanup around this particular dependency, neatly packaged into individual commits should one of them prove controversial. Specifically f019efd311655f7368ed785a849384dd55f41d52 might spark some discussion, but we can also just remove it if the community isn't comfortable with it.

See the individual commit messages as well for some more details.

### DIFF
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -353,7 +353,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
-      <scope>test</scope>
     </dependency>
   </dependencies>
   <build>
--- a/modules/admin-ui/pom.xml
+++ b/modules/admin-ui/pom.xml
@@ -352,7 +352,7 @@
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
+      <artifactId>slf4j-reload4j</artifactId>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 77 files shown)


lkiesow:
> This PR also includes some other cleanup around this particular dependency, neatly packaged into individual commits should one of them prove controversial. 

It would be far easier to discuss individual patches if the patches, in particular the potentially controversial ones, would have been separate pull requests. Ignore that for now, but would be nice for next time.

>  Remove redundant dependency scope specifications
>
> At least for `slf4j-log4j12`.
>
> We might want to do this for other dependencies as well, or we might not want to do it at all, we'll see what people say about this, I guess.

Why is the scope `test` redundant? Isn't `compile` the default scope?

JulianKniephoff:
> It would be far easier to discuss individual patches if the patches, in particular the potentially controversial ones, would have been separate pull requests. Ignore that for now, but would be nice for next time.

I see what you're saying, and I agree in general, but not in this case, because these patches are all related to some extent. (Said extent is debatable of course, granted. 🤷‍♀️). To wit:

> Why is the scope test redundant? Isn't compile the default scope?

It's **made** redundant by specifying the scope globally in the `dependencyManagement`-section of the base POM by the previous commit. I split this into two commits because the second part might be controversial/inconsistent/non-idiomatic/... The first one isn't. All of our test dependencies have their default scope set to `test` in this way.

lkiesow (approved):

PR #5150, opened 2023-08-01
Fix Endless Loop on Elasticsearch Exception

KatrinIhler:
Instead of stopping once maxRetryAttempts has been reached, Opencast will try to query or update Elasticsearch endlessly when encountering an ElasticsearchStatusException. This was broken by #4120.


lkiesow:
Rebased this to fix the zip issue

gregorydlogan:
Works as expected!

Issue #5151, opened 2023-08-03
Installing Karaf features completely breaks Opencast


JulianKniephoff:
Trying to `feature:install` any feature, or chucking it into `deploy`---even if it is empty!---currently completely brings Opencast down.

Spending way too much time on finding out why I found that this is not actually an Opencast problem. I can reproduce it in a fresh copy of Karaf. It still concerns us, of course, so I'm making this issue to track it.

I will link the appropriate upstream issue here as soon as I'm able to create it. But for those who can't wait: It's basically because Karaf wants to refresh the `javax.persistence` bundle because different parts of the system don't agree on the wiring it's supposed to get. Basically every bundle of ours (transitively) depends on it so everything is restarted which Opencast can't handle. (I guess that part **is** an Opencast problem, but it's not what this issue is concerned with AFAIAC.

**To Reproduce**

`feature:install basically-whatever-you-want` or chuck a `features.xml` thingy into `deploy`.

**Expected behavior**

The feature is installed and I can use it and can continue to use Opencast undisturbed.

**Stack trace**

The log shows Opencast basically shutting down and restarting, but it doesn't quite make it; it will continue to spam the log with a wide range of exceptions, too many to list here.

**Server environment:**
 - OS: tested on latest macOS and Arch Linux, but likely doesn't matter.
 - Distribution: dev/allinone, but it shouldn't matter.
- Installation method: from source, but see above
 - Version: tested with `develop` but I also tested Karaf versions as far back as 4.1, so everything that uses that will likely be affected as well.

**Additional context**

🤷

mliradelc:
Related to #4610 

What shows you if you do `feature:install -v <what-you-want>`?

JulianKniephoff:
I saw #4610. Maybe it's a special case of this issue? Didn't really look too closely into it, tbh. 🤷‍♀️

For an empty feature I get this:

```
karaf@root()> feature:install -t -v foo
Adding features: foo/[1.0.0,1.0.0]
No deployment change.
  Bundles to refresh:
    com.mysema.querydsl.jpa/3.7.4 (Wired to javax.persistence/2.2.0.v201708071007 which is being refreshed)
    javax.persistence/2.2.0.v201708071007 (Should be wired to: org.apache.aries.jpa.container/2.7.3 (through [javax.persistence/2.2.0.v201708071007] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.osgi.service.jpa)(version>=1.1.0))"))
[...]

PR #5153, opened 2023-08-03
Disable auto refresh on feature installation

JulianKniephoff:
Due to some wiring inconsistencies (and not all Opencast bundles being able to hot-restart properly) this currently breaks feature installation, because Karaf weants to refresh `javax.persistence`, which more or less everything depends on.

/cc #5151

### DIFF
--- a/assemblies/resources/build.xml
+++ b/assemblies/resources/build.xml
@@ -10,6 +10,8 @@
     <replaceregexp file="target/assembly/bin/karaf" match=" init$" replace=" [ -z &quot;${TERM##*256color}&quot; ] &amp;&amp; export TERM=xterm-color&#x0A;    init" byline="true"/>
     <!-- Make sure bundle cache gets cleared by default -->
     <replaceregexp file="target/assembly/etc/system.properties" match="^karaf.clean.cache .*=.*$" replace="karaf.clean.cache = true" byline="true"/>
+    <!-- Disable feature auto refresh (it's broken rn, see #5151) -->
+    <replaceregexp file="target/assembly/etc/org.apache.karaf.features.cfg" match="^autoRefresh=\s*.*$" replace="autoRefresh=false" byline="true"/>
     <!-- Additional config lines to system.properties -->
     <concat destfile="target/assembly/etc/system.properties" append="true">
       <filelist dir="../resources" files="system.properties.append"/>


lkiesow:
What does this pull request do and why? You just described a problem of Opencast. How does disabling auto refresh on feature installations help, and what side effect does it have?

JulianKniephoff:
Is there anything in particular that's unclear? Because I feel like most of your question is answered in the title and description: This PR disables Karaf's auto-refresh of bundles on feature installation, because said auto-refresh currently breaks Opencast when you install a feature. I would like to be able to install a feature (the Annotation Tool in particular, but that shouldn't matter), so disabling this would allow me to do so.

You can also skip auto-refresh using `feature:install -r/--no-auto-refresh`, but there is no way to skip it for features you drop into `deploy`, for example, which I would like to do. Plus: With the current settings you can't install features at all rn, so this is strictly an improvement, IMO. 🤷‍♀️

Side-note: Our plugin system is based on Karaf features, and [lo and behold](https://github.com/opencast/opencast/blob/14.1/modules/plugin-manager/src/main/java/org/opencastproject/plugin/impl/PluginManagerImpl.java#L134), the plugin manager does this as well.

As for the side-effects: That's fair. So what does auto-refresh even do? It rewires bundles, which you might want to do after a feature installation because the feature could contain new bundles/services/... that satisfy preexisting requirements. So the side-effect of this PR is obviously that this doesn't happen, which could lead to inconsistencies in your wiring, **I think**, but I'm not 100% sure. There is a whole swath of use cases where this does not apply, though. The Annotation Tool as an additional module that the rest of the system is completely agnostic of is one such example, apparently all of our plugins are, too.

And again, this strictly enables more stuff, only breaks in cases that didn't even work/exist before (if at all), and can be reverted if/when this is fixed in Karaf.

lkiesow:
Huh, I wrote that plugin stuff :grin: 
But that doesn't mean I remember what Karaf's auto-refresh does.
Thanks for the explanation. That's what was missing for me.

lkiesow:
Is this intended for `develop`? Or do you want this fix to go into a release branch?

JulianKniephoff:
Fair. 😁 Sorry if that came across a bit salty. 🧂

Re. branch: For me personally it would be useful in the release branches as well, but this **is** a behavior changing config change, so I'm not sure if it's allowed to go in there? On the other hand, as said, the behavior it changes was broken to begin with, soooo ... 🤔🤷‍♀️?

lkiesow (approved):
Seems reasonable and seems to work.
I don't care about the branch.
I'll see where it ends up when I merge it after the technical meeting.

JulianKniephoff:
In that case: Rebased on and retargeted to `r/13.x`. 😊👍

**EDIT**: The build fails due to the corrupted `jar` issue now, sadly. ☹️

PR #5157, opened 2023-08-07
Test for broken JAR (zip) files

lkiesow:
This ensures that our JAR files are actually valid zip files since we seem to have broken files headers in several releases and this is causing issues with recent JVM updates.

This is not a very nice fix, but it should do the trick. It seems like Opencast 15 no longer suffers from this bug. We can probably remove this workaround in that branch. Though I would maybe keep the integrity test around for a bit longer.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


.github/fix-jars
```diff
@@ -0,0 +1,33 @@
...
+for tarball in opencast*.tar.gz
+do
+  tar xf "${tarball}"
+  distname="$(echo "${tarball}" | sed 's/-[0-9].*//')"
+
+  cd "${distname}/system/org/opencastproject/"
+
+  find . -name '*jar' | while read i
+  do
+    if ! zip --test "$i" &> /dev/null
```
ziegenberg: ShellCheck says: In POSIX sh, `&>` is undefined. Use the long form instead of the shorthand:
```suggestion
    if ! zip --test "$i" > /dev/null 2>&1
```

`&>` is a bash and ksh extension for redirecting both stdout and stderr. In dash and POSIX sh, use `2>&1` explicitly.

.github/fix-jars
```diff
@@ -0,0 +1,33 @@
...
+    then
+      echo "$i is broken -> Fixing"
+      tmpdir="$(mktemp --directory)"
+      jarpath="$(readlink -f "$i")"
+      cd "${tmpdir}"
+      unzip "${jarpath}"
+      zip -r --compression-method store fix.zip *
```
ziegenberg: Using `./*glob*` looks a bit safer.

```suggestion
      zip -r --compression-method store fix.zip ./*
```

.github/fix-jars
```diff
@@ -0,0 +1,33 @@
...
+
+cd build
+for tarball in opencast*.tar.gz
+do
+  tar xf "${tarball}"
+  distname="$(echo "${tarball}" | sed 's/-[0-9].*//')"
+
+  cd "${distname}/system/org/opencastproject/"
+
+  find . -name '*jar' | while read i
```
ziegenberg: ShellCheck says: `read` without `-r` will mangle backslashes.


```suggestion
  find . -name '*jar' | while read -r i
```

.github/fix-jars
ziegenberg, line 1: `fix-jars` does what says it should do. Run it locally to verify.

.github/test-jars
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eu
+
+find build/opencast-*/system/org/opencastproject -name '*.jar' | while read jarfile
```
ziegenberg: ShellCheck says: `read` without `-r` will mangle backslashes.
```suggestion
find build/opencast-*/system/org/opencastproject -name '*.jar' | while read -r jarfile
```

.github/test-jars
ziegenberg, line 1: `test-jars` does what says it should do. Run it locally to verify.

ziegenberg (changes requested):
I guess it looks fine, just a minor suggestion. Would be nice to have others' approval too.

lkiesow:
Thanks for the review. Applied all changes.

ziegenberg:
Maybe PR #5158 makes this obsolete?

ziegenberg (approved):

ziegenberg:
@lkiesow As most certainly #5158 will find it's into the codebase I suggest throwing `github/fix-jars` overboard and just keeping the verify integrity bits. That shouldn't hurt us and might prevent this from happening again in the future.

lkiesow:
> @lkiesow As most certainly #5158 will find it's into the codebase I suggest throwing `github/fix-jars` overboard and just keeping the verify integrity bits. That shouldn't hurt us and might prevent this from happening again in the future.

Done. This pull request now just contains the test. Nothing else. Can you do the review/merge?

ziegenberg (approved):

ziegenberg:
Looks good to me

JamesUoM:
As java use to happily parse these jar files in the previous point release it is considered a bug that has been fixed in 11 and 17 https://bugs.openjdk.org/browse/JDK-8313765

PR #5164, opened 2023-08-10
Fix editing custom actions in the ACL editor

JulianKniephoff:
Right now you can't edit custom actions of individual policies in the ACL editor; all the actions of all policies are kind of tied together.

This is due to a bug in the JavaScript caused by the select components all operating on the same shared array. This PR fixes that bug by just making copies in appropriate places to break that link.

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -53,7 +53,7 @@ angular.module('adminNg.controllers')
             write : write !== undefined ? write : false,
             actions : {
               name : 'event-acl-actions',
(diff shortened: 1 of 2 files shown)


JulianKniephoff:
Requesting a review by @Arnei since he did the ACL editor overhaul. Of course I welcome any other reviewers as well! 😀

Arnei (approved):

mliradelc:
I would try this with LDAP this week. Since the ACL UI overhaul has broken our installation, we are using an OC with the PR retracted.

JulianKniephoff:
@mliradelc did you have a chance to look at this? Would be nice to get this in, soon. :) FWIW I doubt that the choice of LDAP over other methods of authentication has any bearing on this code. 🤔

PR #5165, opened 2023-08-10
Fix crowdin package name

geichelberger:
This PR fixes the crowding deb package install command.


lkiesow (approved):
Makes sense

gregorydlogan (approved):
Dumb mistake was dumb. Thanks for the catch.

PR #5169, opened 2023-08-14
Copy active inputs between CAs if they have the same set of inputs

Arnei:
For a scheduled event in the admin ui, the capture agent for the event can be changed in the event details. However, when changing the capture agent, the input channels for the capture agent will always be unset (except for the capture agent the event was created with). As most institutions usually use multiple capture agents of the same brand, with the same set of input channels, having to manually reactivate every input after switching capture agents is quite cumbersome.

This PR changes the default behaviour for switching capture agents. If both capture agents have the same set of input channels, which inpus are active will now be copied over from the previously selected capture agent to the newly selected one.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -651,7 +651,7 @@ angular.module('adminNg.controllers')
             id: $scope.resourceId,
             entries: $scope.source,
             previousId: oldObj ? oldObj.id : undefined,
-            previousEntries: oldObj? oldObj.inputMethods : undefined
+            previousEntries: oldObj ? oldObj.inputMethods : undefined
           }, function () {
             fetchChildResources($scope.resourceId);
           });
@@ -661,10 +661,10 @@ angular.module('adminNg.controllers')
           if (oldObj && $scope.source.device.inputs && oldObj.inputs) {
             var sourceInputs = $scope.source.device.inputs.map(function(input){
               return input.id;
-            }).join(",");
+            }).join(',');
             var oldObjInputs = oldObj.inputs.map(function(input){
               return input.id;
-            }).join(",");
+            }).join(',');
             if (sourceInputs === oldObjInputs) {
               $scope.source.device.inputMethods = oldObj.inputMethods;
             }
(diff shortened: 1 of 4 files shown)


snoesberger:
The change basically works, the SMP input channel settings are now copied over. However, there is still one flaw: at the first moment, after changing the CA, the two checkboxes are deactivated. After the change has been processed (takes in our environment a few seconds), the checkboxes are active again (as they were before).

Is it possible to adjust this behaviour so that the checkboxes remain activated after the CA change until processing is complete?

Arnei:
@snoesberger Thanks for the feedback, should now behave like you want it to.

snoesberger:
@Arnei Thanks for the quick fix!
I just found another issue. On an existing event with an Extron SMP as CA, I can't change the CA to a pyca. I assume this is because pyca's do not have channel attributes like the Extron SMP's do?
In the broser console I can see that the following request leads to a 500 Internal Server Error:
https://admin.opencast-test.unibe.ch/admin-ng/event/65f3a50a-bf2f-4d35-8ae4-10bd47fe257d/scheduling
With data:
```
{
  "agentId":"pyca-Sascha",
  "start":"2023-08-31T15:00:00Z",
  "end":"2023-08-31T15:10:00Z",
  "agentConfiguration":
  {
    "event.series":"c23a27cc-8eb9-41e7-8965-d01d76b81794",
    "capture.device.names":"",
    "org.opencastproject.workflow.config.publishLive":"true",
    "org.opencastproject.workflow.definition":"schedule-and-upload",
    "org.opencastproject.workflow.config.straightToPublishing":"false",
    "event.location":"WTI-Hal6-020-ext351L",
    "event.title":"Live Stream 2023-08-31 17:00"
  },
  "previousAgentId":"WTI-Hal6-020-ext351L",
  "previousEntries":
  {
    "Channel A":true,"Channel B":true
  }
}
```

In the opencast server logs I have the following warning at the same time:
`2023-08-31T16:23:05,170 | WARN  | (LogUtils:468) - Application {http://endpoint.external.opencastproject.org/}ListProviderEndpoint has thrown exception, unwinding now
org.apache.cxf.interceptor.Fault: null must not be wrapped in a some`

Arnei:
Thanks for the detailed report, should be fixed now.

PR #5170, opened 2023-08-16
Fix double subtitle when a flavor has several videos

mliradelc:
This PR addresses an issue when an event has a flavour with multiple videos on it. Whisper or Vosk will try to create subtitles for each video, instead of making only one. Now will select the first track from the targeted flavor.

Furthermore, adds the option `source-tags` to any user who wants to select a specific track from the targeted flavor.

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
@@ -10,6 +10,8 @@ The speech to text operation can be used to generate subtitles for Videos or Aud
 engines available, [Whisper](../modules/transcription.modules/whisper.md) and 
 [Vosk](../modules/transcription.modules/vosk.md). The subtitles file format ist WebVTT.
 
+This operation is designed to only work with one source file per flavor. If you want to use multiple source files in one
+flavor, you need to set the same tag to each element.
 
 Parameter Table
 ---------------
@@ -17,6 +19,7 @@ Parameter Table
 | configuration keys | required | Example           | description                                                                                                                                        |
 |--------------------|----------|-------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | source-flavor      | yes      | source/presenter  | The source media package to use                                                                                                                    |
+| source-tags        | no       | 1080p             | The tags that the source track uses                                                                                                                   |
 | target-flavor      | yes      | archive           | Flavor of the produced subtitle file. The subflavor supports the language-code placeholder `#{lang}`                                               |
 | target-element     | no       | attachment        | Define where to append the subtitles file. Possibilities are: as a 'track' or as an 'attachment'. The default is "attachment".                     |
 | language-code      | no       | de                | The language of the video or audio source (default is "eng"). Vosk only: It has to match the name of the language model directory. See 'vosk-cli'. |
--- a/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
+++ b/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
@@ -29,6 +29,7 @@ import org.opencastproject.mediapackage.MediaPackageElementFlavor;
 import org.opencastproject.mediapackage.MediaPackageElementParser;
 import org.opencastproject.mediapackage.Track;
 import org.opencastproject.mediapackage.attachment.AttachmentImpl;
+import org.opencastproject.mediapackage.elementbuilder.TrackBuilderPlugin;
 import org.opencastproject.mediapackage.track.TrackImpl;
 import org.opencastproject.metadata.api.MediaPackageMetadata;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalogService;
@@ -56,7 +57,10 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.UUID;
 
 /**
@@ -125,23 +129,53 @@ public class
           throws WorkflowOperationException {
 
     MediaPackage mediaPackage = workflowInstance.getMediaPackage();
+    WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
     logger.info("Start speech-to-text workflow operation for media package {}", mediaPackage);
+
 
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(workflowInstance,
             Configuration.none, Configuration.one,
             Configuration.many, Configuration.one);
-    MediaPackageElementFlavor sourceFlavor = tagsAndFlavors.getSingleSrcFlavor();
+    List<MediaPackageElementFlavor> sourceFlavor = tagsAndFlavors.getSrcFlavors();
+    List<String> sourceTagList = tagsAndFlavors.getSrcTags();
 
-    Track[] tracks = mediaPackage.getTracks(sourceFlavor);
-    if (tracks.length == 0) {
+
+    MediaPackageElementFlavor matchingFlavor = null;
+    if (!sourceFlavor.isEmpty()) {
+      matchingFlavor = sourceFlavor.get(0);
+    }
+
+    // Select the tracks based on source flavors and tags
+    Set<MediaPackageElement> inputSet = new HashSet<>();
+    for (MediaPackageElement element : mediaPackage.getElementsByTags(sourceTagList)) {
+      MediaPackageElementFlavor elementFlavor = element.getFlavor();
+      if (sourceFlavor == null || (elementFlavor != null && elementFlavor.matches(matchingFlavor))) {
+        inputSet.add(element);
+      }
+    }
+
+    if (inputSet.size() == 0) {
       throw new WorkflowOperationException(
-              String.format("No tracks with source flavor '%s' found for transcription", sourceFlavor));
+              String.format("No elements with source flavor(s) {} and tag(s)"
+                  + " {} found for transcription", sourceFlavor, sourceTagList));
     }
 
-    if (tracks.length > 1) {
-      logger.warn("Found {} track(s) with source flavor '{}'.", tracks.length, sourceFlavor);
-      logger.warn("Only 1 subtitle will be generated for the flavor");
+    // Convert MediaPackageElements to Tracks
+    TrackBuilderPlugin trackBuilder = new TrackBuilderPlugin();
+    List<Track> tracksList = new ArrayList<>();
+
+    for (MediaPackageElement element : inputSet) {
+      try {
+        URI elementUri = element.getURI();
+        Track trackElement = (Track)  trackBuilder.elementFromURI(elementUri);
+        tracksList.add(trackElement);
+      } catch (ClassCastException e) {
+        logger.error("Element {} is not a track", element);
+      }
     }
+
+    Track[] tracks = tracksList.toArray(new Track[tracksList.size()]);
+
 
     // Get the information in which language the audio track should be
     String languageCode = getMediaPackageLanguage(mediaPackage, workflowInstance);
@@ -151,11 +185,15 @@ public class
 
     // Translate to english
     Boolean translate = getTranslationMode(mediaPackage, workflowInstance);
-
-    for (Track track : tracks) {
-      createSubtitle(track, languageCode, mediaPackage, tagsAndFlavors, appendSubtitleAs, translate);
+    if (sourceTagList.isEmpty() && (tracks.length > 1)) {
+      logger.info("Multiple tracks found but no source tags specified");
+      logger.info("Using the first track in the flavor list");
+      createSubtitle(tracks[0], languageCode, mediaPackage, tagsAndFlavors, appendSubtitleAs, translate);
+    } else {
+      for (Track track : tracks) {
+        createSubtitle(track, languageCode, mediaPackage, tagsAndFlavors, appendSubtitleAs, translate);
+      }
     }
-
     logger.info("Text-to-Speech workflow operation for media package {} completed", mediaPackage);
     return createResult(mediaPackage, WorkflowOperationResult.Action.CONTINUE);
   }
--- a/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
+++ b/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
@@ -129,7 +129,6 @@ public class
           throws WorkflowOperationException {
 
     MediaPackage mediaPackage = workflowInstance.getMediaPackage();
-    WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
     logger.info("Start speech-to-text workflow operation for media package {}", mediaPackage);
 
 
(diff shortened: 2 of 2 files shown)


Arnei:
It feels like this PR is missing at least one commit. As of now, all it does is adding a warning that (probably wrongly?) claims that "Only 1 subtitle will be generated for the flavor". It does not change the behaviour of the SpeechToText WoH.

Furthermore, the warning suggests that the SpeechToText WoH would pick a track from the flavor at its own discretion. How would it decide which track to pick? To me, that is something the user should decide, so I'd be in favor of either keeping the current behaviour or throwing an error if there is more than one track.

mliradelc:
Thanks Arne, for noticing that is missing actual the code. That happens when you cherry-pick from another repo 😅 

Well, the logic is, while opencast allows you to set different videos in the same flavour. The scenario involves using a flavour for one type of video (presenter, slides, etc.) and later tags for other variations (codec, resolution, etc.)

I asked in the last summit, and there is no use case where someone put different types of videos in the same flavour. 

~Still, if you want to choose the exact video, you can use the source tags alongside the flavour.~
I just realized that there is no `source-tag` mechanism in the WoH, I will patch it.

mliradelc:
@Arnei Now I added a way to select specific tracks using tags, Can you try the new mechanism?

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+      matchingFlavor = sourceFlavor.get(0);
+    }
 
-    Track[] tracks = mediaPackage.getTracks(sourceFlavor);
-    if (tracks.length == 0) {
+    // Select the tracks based on source flavors and tags
+    Set<MediaPackageElement> inputSet = new HashSet<>();
+    for (MediaPackageElement element : mediaPackage.getElementsByTags(sourceTagList)) {
+      MediaPackageElementFlavor elementFlavor = element.getFlavor();
+      if (sourceFlavor == null || (elementFlavor != null && elementFlavor.matches(matchingFlavor))) {
```
Arnei:
```suggestion
      if (sourceFlavor.isEmpty() || (elementFlavor != null && elementFlavor.matches(matchingFlavor))) {
```

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+    TrackBuilderPlugin trackBuilder = new TrackBuilderPlugin();
+    List<Track> tracksList = new ArrayList<>();
+
+    for (MediaPackageElement element : inputSet) {
+      try {
+        URI elementUri = element.getURI();
+        Track trackElement = (Track)  trackBuilder.elementFromURI(elementUri);
+        tracksList.add(trackElement);
+      } catch (ClassCastException e) {
+        logger.error("Element {} is not a track", element);
```
Arnei: I think this error should be thrown, not just logged. I don't see a use case were a user has multiple elements with the same tag in the same flavor, where some are tracks and some aren't, and still wants those tracks translated. It's more likely that something is wrong in the workflow.

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
```diff
@@ -10,13 +10,16 @@ The speech to text operation can be used to generate subtitles for Videos or Aud
 engines available, [Whisper](../modules/transcription.modules/whisper.md) and 
 [Vosk](../modules/transcription.modules/vosk.md). The subtitles file format ist WebVTT.
 
+This operation is designed to only work with one source file per flavor. If you want to use multiple source files in one
+flavor, you need to set the same tag to each element.
```
Arnei, line 14: This seems overcomplicated. In particular, forcing users to add tags to tracks if they want to translate multiple at the same time. Why not just "This operation will attempt to translate all tracks in the given `source-flavor`. To restrict which tracks to translate, add tags to the respective tracks and specify them in `source-tags`"?

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
```diff
@@ -10,13 +10,16 @@ The speech to text operation can be used to generate subtitles for Videos or Aud
...
 
 Parameter Table
 ---------------
 
 | configuration keys | required | Example           | description                                                                                                                                        |
 |--------------------|----------|-------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | source-flavor      | yes      | source/presenter  | The source media package to use                                                                                                                    |
+| source-tags        | no       | 1080p             | The tags that the source track uses                                                                                                                   |
```
Arnei, line 22: > The tags that the source track uses

Sounds like I need to specify ALL tags of a particular track, but actually I just need to specify one of them, correct?

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
           throws WorkflowOperationException {
 
     MediaPackage mediaPackage = workflowInstance.getMediaPackage();
+    WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
```
Arnei:
```suggestion

```
Unused

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
     logger.info("Start speech-to-text workflow operation for media package {}", mediaPackage);
 
+
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(workflowInstance,
             Configuration.none, Configuration.one,
             Configuration.many, Configuration.one);
-    MediaPackageElementFlavor sourceFlavor = tagsAndFlavors.getSingleSrcFlavor();
+    List<MediaPackageElementFlavor> sourceFlavor = tagsAndFlavors.getSrcFlavors();
```
Arnei, line 138: Can use `.getSingleSourceFlavor()` here

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
           throws WorkflowOperationException {
 
     MediaPackage mediaPackage = workflowInstance.getMediaPackage();
+    WorkflowOperationInstance operation = workflowInstance.getCurrentOperation();
     logger.info("Start speech-to-text workflow operation for media package {}", mediaPackage);
 
+
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(workflowInstance,
             Configuration.none, Configuration.one,
```
Arnei:
```suggestion
            Configuration.many, Configuration.one,
```

Arnei (changes requested):

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
mliradelc: thats correct, you need only to check one tag, not all.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
mliradelc: Thanks for noting that, initially I was to use that variable, but later I changed my idea.

It surprises me that Check Style didn't warn me about.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
mliradelc: I thought the same as you, then later I said, "maybe is better to keep the workflow running but log the error with that element". 

Viewing that again, perhaps is better to throw and not allow bat Workflows.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
lkiesow: Agree. Seems like the next lines check for and only take the first flavor anyway.
I think `Configuration.one` should also make the check for one flavor unnecessary, shouldn't it?

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
lkiesow: When using `Configuration.one` and `getSingleSourceFlavor`, isn't there a guarantee for a single valid flavor? That would make the check on `sourceFlavor` unnecessary.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+      if (sourceFlavor == null || (elementFlavor != null && elementFlavor.matches(matchingFlavor))) {
+        inputSet.add(element);
+      }
+    }
+
+    if (inputSet.size() == 0) {
       throw new WorkflowOperationException(
-              String.format("No tracks with source flavor '%s' found for transcription", sourceFlavor));
+              String.format("No elements with source flavor(s) {} and tag(s)"
+                  + " {} found for transcription", sourceFlavor, sourceTagList));
```
lkiesow: You switched from `%s` to `{}`. That doesn't work with `String.format`:

```java
jshell> String.format("No elements with source flavor(s) {} and tag(s) {} found for transcription", 1, 2);
$2 ==> "No elements with source flavor(s) {} and tag(s) {} found for transcription"

jshell> String.format("No elements with source flavor(s) %s and tag(s) %s found for transcription", 1, 2);
$3 ==> "No elements with source flavor(s) 1 and tag(s) 2 found for transcription"
```

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -149,10 +186,16 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
     // Translate to english
     Boolean translate = getTranslationMode(mediaPackage, workflowInstance);
 
-    for (Track track : tracks) {
-      createSubtitle(track, languageCode, mediaPackage, tagsAndFlavors, appendSubtitleAs, translate);
+    // Subtitles creation
+    if (sourceTagList.isEmpty() && (tracks.length > 1)) {
+      logger.info("Multiple tracks found but no source tags specified");
+      logger.info("Using the first track in the flavor list");
```
lkiesow: If it's one message, combine it into one logger statement. Otherwise, you might end up with other log statements in between these two.

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
mliradelc: Yes, you are right. That line is confusing. This PR aims to work like any other Workflow.
If source flavors and tags are meet, it should select the correct files.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
mliradelc: I don't understand what you mean with using `.getSingleSourceFlavor()`  is not a valid method for that type of object

The method is (Workflow instance, Src Tags, Src Flavors, Target Tags, Target Flavors). 

`Configuration.one` takes the first Flavor only.

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-woh.md
mliradelc: For example. If you have 2 tracks with tags `"one","two"` and another with the tags `"one","three"`

If you target only `one`, it will process both.
If you target `one,two`, it will process the first.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -125,20 +129,53 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
             Configuration.many, Configuration.one);
-    MediaPackageElementFlavor sourceFlavor = tagsAndFlavors.getSingleSrcFlavor();
+    List<MediaPackageElementFlavor> sourceFlavor = tagsAndFlavors.getSrcFlavors();
+    List<String> sourceTagList = tagsAndFlavors.getSrcTags();
+
+
+    MediaPackageElementFlavor matchingFlavor = null;
+    if (!sourceFlavor.isEmpty()) {
+      matchingFlavor = sourceFlavor.get(0);
+    }
```
Arnei, line 145: How I think `getSingleSrcFlavor` could be used

```suggestion
     MediaPackageElementFlavor sourceFlavor = tagsAndFlavors.getSingleSrcFlavor();
    List<String> sourceTagList = tagsAndFlavors.getSrcTags();
```

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
Arnei: My bad, it's called `getSingleSrcFlavor`. See other comment below

mliradelc:
Moved to draft because I need to rethink better if this fix worth it as is.

mtneug:
13.13 was the last 13.x release. @mliradelc, please redirect this to `r/14.x` or newer.

mtneug:
I will close this as 13 is EOL now. Feel free to refile for newer versions or redirect and reopen.

PR #5172, opened 2023-08-18
Update metadata publication when updating asset manager

Arnei:
Automatically republish catalogs and ACLs to existing publication channels, as well as the search service. The intention here is to avoid having to run a "republish-workflow" when updating metadata or access rights of an event.

Marking this as draft because I'm really unsure if this should go on as is, as I'm not deeply knowledgable on Opencasts publication mechanisms. Also this just kinda feels like a hacky, side-effect solution.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


mliradelc:
What happens if you run `republish-workflow` with this PR? Will you have repeated data?

I'm asking to know if is ok to go into 13.x. There is a potential risk to breaking workflows with this PR.

If not the case, I think is good to go.

modules/index-service/pom.xml
```diff
@@ -130,6 +130,16 @@
...
+    <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-distribution-service-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-search-service-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
JulianKniephoff: This is a nitpick, but these should probably have `<scope>provided</scope>`

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
     return metadataList;
   }
 
+  /**
+   * Updates the publications of a mediapackage with new metadata and access rights.
+   * @param mediaPackage
```
JulianKniephoff: Another nitpick: Empty JavaDoc tags like this are unnecessary. If you don't want to add a comment, remove them. The parameter will still end up correctly in the JavaDoc.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+      }
+
+      // Remove publication from mediapackage
+      for (Publication publicationElement : mediaPackage.getPublications()) {
+        if (channelId.equals(publicationElement.getChannel())) {
+          mediaPackage.remove(publicationElement);
+        }
+      }
+
+      List<MediaPackageElement> downloadElements = new ArrayList<>();
```
JulianKniephoff: The initialization here is redundant.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+
+      // Get metadata and acls
+      SimpleElementSelector elementSelector = new SimpleElementSelector();
+      // This does not necessarily work for extended metadata, which can have completely different flavors.
+//      elementSelector.addFlavor("dublincore/*");
+      // Relies on ACLs having this particular flavor
+      elementSelector.addFlavor("security/*");
+      Collection<MediaPackageElement> elements = elementSelector.select(mediaPackage, false);
+
+      Set<String> elementIds = new HashSet<String>();
```
JulianKniephoff:
```suggestion
      Set<String> elementIds = new HashSet<>();
```

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+    for (Publication publication : mediaPackage.getPublications()) {
+      String channelId = publication.getChannel();
+      String publicationId = publication.getIdentifier();
+      URI publicationURI = publication.getURI();
+      MimeType publicationMimeType = publication.getMimeType();
+
+      if (channelId == null || publicationURI == null || publicationMimeType == null || publicationId == null) {
+        // No publication, do nothing
+        continue;
+      }
```
JulianKniephoff: I mean this doesn't hurt, but can this situation actually happen? Why would there be a publication in the list of publications that lacks any of these essential fields? 👀 Or like, lack one of them, but not the others, even?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+
+      // Get metadata and acls
```
JulianKniephoff:
```suggestion
      // Get metadata and ACLs
```

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
     return metadataList;
   }
 
+  /**
+   * Updates the publications of a mediapackage with new metadata and access rights.
+   * @param mediaPackage
+   */
+  private void updatePublication(MediaPackage mediaPackage) {
```
JulianKniephoff:
```suggestion
  private void updatePublications(MediaPackage mediaPackage) {
```

seems more accurate

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+      if (channelId == null || publicationURI == null || publicationMimeType == null || publicationId == null) {
+        // No publication, do nothing
+        continue;
+      }
+
+      // Get metadata and acls
+      SimpleElementSelector elementSelector = new SimpleElementSelector();
+      // This does not necessarily work for extended metadata, which can have completely different flavors.
+//      elementSelector.addFlavor("dublincore/*");
+      // Relies on ACLs having this particular flavor
```
JulianKniephoff: Dead code?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+        if (channelId.equals(publicationElement.getChannel())) {
+          mediaPackage.remove(publicationElement);
+        }
+      }
+
+      List<MediaPackageElement> downloadElements = new ArrayList<>();
+      List<MediaPackageElement> streamingElements = new ArrayList<>();
+      try {
+        downloadElements = downloadDistributionService.distributeSync(channelId, mediaPackage, elementIds, false);
+        if (streamingDistributionService != null && streamingDistributionService.publishToStreaming()) {
```
JulianKniephoff: The latter check seems to be redundant, since the single streaming distribution service in existence already checks this before actually distributing, which is what I would expect from every sensible implementation, tbh.

JulianKniephoff: The former check might actually be redundant as well, as it is written right now, since with the new `Reference` to a `StreamingDistributionService`, the index service won't shouldn't even start up when this is `null`.

Whether or not that is a good thing is a different question, though. 🤔 

**EDIT**: Well, Opencast just seems to assume such a service to be there in other places as well, so 🤷‍♀️ for this PR, I guess, but it's still questionable, IMO.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
     return metadataList;
   }
 
+  /**
+   * Updates the publications of a mediapackage with new metadata and access rights.
+   * @param mediaPackage
+   */
+  private void updatePublication(MediaPackage mediaPackage) {
+    for (Publication publication : mediaPackage.getPublications()) {
```
JulianKniephoff: I have a lot of questions about the control flow in this function:

* This outer loop loops over the publication, but the body also changes the publications, (why) is that okay? 🤔
* There is another inner loop over the publications that seems kind of pointless, but I might be wrong. See below.
* Large parts of the body, like the gathering of the ACLs/catalogues seems to be agnostic of the current publication, so that could be pulled out at the least, but one should probably also think about whether that might be a logic bug? See further comments down below.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+        // Nothing to republish, do nothing
+        continue;
+      }
+
+      // Remove publication from mediapackage
+      for (Publication publicationElement : mediaPackage.getPublications()) {
+        if (channelId.equals(publicationElement.getChannel())) {
+          mediaPackage.remove(publicationElement);
+        }
+      }
```
JulianKniephoff: Is this just looking for the right publication to remove? But shouldn't you already have this in `publication`? OK, fair, this will find others with the same channel ID. (Can there even be multiple publications with the same channel ID?) But those would be "found" by the outer loop eventually anyway, no? 🤔

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1347,6 +1374,7 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
         break;
       case ARCHIVE:
         assetManager.takeSnapshot(mediaPackage);
+        updatePublication(mediaPackage);
```
JulianKniephoff: Is it correct to take the snapshot **before** updating the publications? 🤔

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+        element.setIdentifier(null);
+        PublicationImpl.addElementToPublication(newPublication, element);
+      }
+
+      mediaPackage.add(newPublication);
+    }
+
+    // Also search service
+    try {
+      searchService.addSynchronously(mediaPackage);
```
JulianKniephoff: The PublishEngageWOH does a bunch of transformations to the media package before adding it to search, don't we need to do any of those here?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+        if (streamingDistributionService != null && streamingDistributionService.publishToStreaming()) {
+          streamingElements = streamingDistributionService.distributeSync(channelId, mediaPackage, elementIds);
+        }
+      } catch (DistributionException e) {
+        throw new RuntimeException(e);
+      }
+      logger.debug("Distribute of mediapackage {} to channel {} completed", mediaPackage, channelId);
+
+      // Re-add publication
+      Publication newPublication = PublicationImpl.publication(publicationId, channelId, publicationURI, publicationMimeType);
```
JulianKniephoff: Is it okay to reuse the publication ID? 🤔

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -1363,6 +1391,88 @@ public MetadataList updateEventMetadata(String id, MetadataList metadataList, El
...
+
+      // Add published elements
+      for (MediaPackageElement element : downloadElements) {
+        element.setIdentifier(null);
+        PublicationImpl.addElementToPublication(newPublication, element);
```
JulianKniephoff: It seems to me that the new publication just gets **every** media package element with flavor `security/*` and every catalog, which might be fine, I'm not sure, but seems a bit excessive? Shouldn't we check which elements the publication had before and only redistribute and attach those? 🤔

JulianKniephoff (changes requested):
OK so there is a bunch in this review: Some stylistic nitpicks, some more serious problems with the code, but also a lot of open questions of a more conceptual nature. The latter in particular are not really change requests, since I don't know the answers, either, unfortunately. Rather, they are invitations to @Arnei and/or the community to discuss and/or educate me.

Same goes for most of the following broader issues that didn't really belong to any line of code in particular:

* This should probably not go into 13.x.
* @mtneug mentioned in the technical that he would like to be able to deactivate this feature, so we might need a config option.
* AWS was also mentioned in the technical as a potential problem with this PR. I don't know enough to really say anything useful here. There is another distribution service for that. People with more knowledge need to chip in here, I just didn't want this to get lost.
* This updates **all** publications, which is probably fine most of the time, but I don't know how the YouTube stuff works, for example. I assume that more stuff has to happen to properly update that publication than to just update the MP.
  * Kind of related: @Arnei brought up the question of how to find out which publication channels exist internally, which I think is not something Opencast can answer right now. We **could** just have a config option specifying which channels are updated, though. Deactivating the feature could also be covered by this: Just leave that list empty.
* AFAICS this mostly imitates what the PublishWOHs do, which makes sense, but also makes me wonder: Should we duplicate that logic in the IndexService? Can we somehow share code here, where share != "copy&paste"?
* Do we need to think about transactions here? Like, what if any of these steps fail? What about race conditions? 🤔 
* Is `updatePublication` called in all the right places? I'm not sure, tbh. Random example: What about `removeCatalogByFlavor`. Does that need to republish? (Or maybe does even already and I'm just blind?)
  * But even if it does, this kind of makes me uneasy, that we now have to remember to do this everywhere. Isn't there a more central place to put this logic? When I looked at this problem briefly a while back (remember, we want this for Tobira for a long time already), I found [`AssetManagerUpdateHandler`](https://github.com/opencast/opencast/blob/14.3/modules/message-broker-api/src/main/java/org/opencastproject/message/broker/api/update/AssetManagerUpdateHandler.java#L26), maybe that's worth investigating? 🤔  IIRC that's also used to (partially) implement this feature for series, already, which would also nicely "colocate" these things.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Sure. Still prefer to not care about what a sensible implementation for the streaming service is and rely on `publishToStreaming()` to return a sensible value instead.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: > This outer loop loops over the publication, but the body also changes the publications, (why) is that okay? 🤔

Good question, maybe it is not.

> There is another inner loop over the publications that seems kind of pointless, but I might be wrong. See below.

Is indeed pointless, should be simplified.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Right, this is pointless, should be simplified.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: No idea. It seemed reasonable to first make sure data is saved before distributing it. But the default workflows do it the other way around it seems.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Maybe. Seemed superfluous to me.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Hopefully. It's definitely easier than generating a new one and having to figure out how to generate a publication uri with the new id for each publication.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Sounds reasonable. Would fail in case there is a new file though?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Don't know.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
Arnei: Made republish flavors configurable.

modules/index-service/pom.xml
```diff
@@ -226,6 +226,12 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-message-broker-api</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
```
JulianKniephoff: I think `provided` is more correct/idiomatic here. If `compile` is what you really want, it should already be the default (unless we override it in some `dependencyManagement` section, which I don't think we do) anyway, but of course it's not the worst thing to be explicit about it. 🤷‍♀️

etc/org.opencastproject.index.service.impl.util.PublicationEventUpdateHandler.cfg
```diff
@@ -0,0 +1,9 @@
+# A comma separated list of publication channel identifiers to automatically republish to when saving to the asset
+# manager.
+# The available publication channel identifiers depends on your system configuration.
+# Common publication channels are:
+#   - engage-player: Standard publication for players
+#   - interal: Opencast internal publication (mainly for the editor)
+#   - api: for the external api
+# Default value: <empty>
+# publication.channel.ids =
```
JulianKniephoff: No newline at end of file.

I don't even know what Opencast's convention is here, but it bugs me. xP

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -85,6 +91,16 @@ protected void setSearchService(SearchService searchService) {
...
+    Object pubChannelIdsValue = cc.getProperties().get(CONFIGURATION_PUBLICATION_CHANNEL_IDS);
+    configurationPublicationChannelIds.clear();
+    if (pubChannelIdsValue != null) {
+      for (String publicationChannel : StringUtils.split(pubChannelIdsValue.toString(), ',')) {
+        if (StringUtils.trimToNull(publicationChannel) != null) {
+          configurationPublicationChannelIds.add(StringUtils.trimToNull(publicationChannel));
+        }
+      }
+    }
+    logger.debug("Excluded user providers: {}", configurationPublicationChannelIds);
```
JulianKniephoff: User providers? 🤔

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -119,7 +120,7 @@ public void execute(AssetManagerItem messageItem) {
   }
 
   private void updatePublications(MediaPackage mediaPackage) {
-    for (Publication publication : mediaPackage.getPublications()) {
+    for (Publication publication : new ArrayList<>(Arrays.asList(mediaPackage.getPublications()))) {
```
JulianKniephoff: There are a few ways to make this shorter and probably more efficient. I think my fav. so far is

```suggestion
    for (Publication publication : mediaPackage.getPublications().clone()) {
```

JulianKniephoff (changes requested):
A few more nitpicks from me, but overall I don't think I have a lot to add here anymore.

I voiced all my main concerns in the last review, but as I said there already, I'm not super sure about this stuff either, so I would love to see more 👀 from the community on this PR.

etc/org.opencastproject.index.service.impl.util.PublicationEventUpdateHandler.cfg
```diff
@@ -0,0 +1,14 @@
+# A comma separated list of publication channel identifiers to automatically republish to when saving to the asset
+# manager.
+# The available publication channel identifiers depends on your system configuration.
+# Common publication channels are:
+#   - engage-player: Standard publication for players
+#   - interal: Opencast internal publication (mainly for the editor)
+#   - api: for the external api
+# Default value: <empty>
```
gregorydlogan, line 9: So by default we don't automatically republish?  This should be made clear here.

etc/org.opencastproject.index.service.impl.util.PublicationEventUpdateHandler.cfg
```diff
@@ -0,0 +1,14 @@
...
+#publication.channel.ids =
+
+# A comma separated list of flavors of the metadata and ACLs to republish.
+# While you can in theory specify any flavor and thus republish anything you want, it is recommended to only specify
+# the flavors of your metadata catalogs and ACL files. Anything else may lead to unexpected results.
+publication.flavors = "dublincore/*, security/*"
```
gregorydlogan: What happens if I leave this blank? :D

Also, are the quotes necessary?  My read of the code says that
`"dublincore/*, security/*"` becomes
```
[
  "dublincore/*,
  security/*"
]
```
Which is probably not what we're looking for?  Unless I missed the quote-stripping code somewhere...

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
gregorydlogan: Also, `excluded`?  These are the channels you're automatically republishing, right?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
gregorydlogan: Double better, `MediaPackageImpl` does this when you `.getPublications()` ;)
```
return mlist(elements).bind(presentations).value().toArray(new Publication[0]);
```

So just use the raw `Publication[]` you're getting back.

gregorydlogan (approved):
In general this looks good to me.  There are a few very minor rough edges to sand down but otherwise I'm happy.

etc/org.opencastproject.index.service.impl.util.PublicationEventUpdateHandler.cfg
Arnei: > What happens if I leave this blank? :D

Then you publish all elements in the `""` flavor. Will fix that, thanks.

> Also, are the quotes necessary?

No, but it works with or without quotes (No, I can't point to the function that does this, sorry). It still makes sense to remove them for consistency across config options I suppose, so I'll do that.

JulianKniephoff (commented):
@Arnei asked me to summarize my remaining concerns, especially those I would like to have input from the community about, so here goes:

* This should probably not go into `r/13.x`. Although with it being configurable now and off by default, maybe it doesn't matter, actually. 🤷‍♀️
* Somethingsomething S3 and the way we get a hold of the distribution services. Is this kosher? Will it just work in every configuration? Maybe, I just can't judge it.
* This duplicates (but not really) code from the publishing WOHs. I have two concerns here:
  * Is the behavior of this new mechanism sufficiently equivalent to me hitting "republish metadata"? Does it need to be?
  * And if so, can we maybe share code somehow?
* What about race conditions, concurrent updates, etc. Do we need to worry about and/or mitigate stuff like this?
* One other thing that came to mind while thinking about the point above some more **after** my review actually: Don't series already use this asset manager event mechanism to update their events under certain conditions? Does that mechanism also republish? So do we republish twice now in that case? 🤔
  * And yeah, such a batch mechanism is the perfect breeding ground for race conditions, and there even is one right now already, even without this PR, IIRC.
  * And also is this a performance problem maybe?
  * I know that this is out of scope of Lars' original "order" with us, but from a Tobira standpoint I can already announce that we want this for series as well, so one could also look at said existing series mechanism, and see how it could be integrated with/adapted for this one. But that's probably out of scope for this PR. 🤷‍♀️

Changing my "approval state" to comment (**EDIT** meh that's actually not a thing, you either approve or request changes; but I think that shouldn't block anything); I don't want to block this PR with this, but I personally could sleep better when someone with the proper knowledge would say a few words about these, and why I maybe shouldn't worry. 🙂

I'll also go ahead and request a review from @lkiesow, because he was the sponsor in this case, and he already said he wanted to look at it as well, still. No pressure, just so this doesn't get lost.

karendolan:
@Arnei what is the issue with running republish-workflow (with "merge" strategy) that this pull resolves?
For example, at our site, when we update event metadata and click submit for the changes, we automatically run a merge workflow. How will this pull impact that?

This is the merge strategy in the PublishEngage...
https://github.com/opencast/opencast/blob/develop/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java#L339-L345
https://github.com/opencast/opencast/blob/develop/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java#L387-L394

KatrinIhler:
I talked about this PR with Julian on Saturday (yes, we talk about work when not at work...) and I had some thoughts. Will try to review this this week.

LukasKalbertodt:
@karendolan I don't know anything about this merge strategy you mentioned, but I can describe the advantages of this PR from my perspective, a Tobira dev. 

There are several apps that want to allow users to update video metadata (e.g. Tobira, LMS plugins, ...). Doing that right now is not very robust and does not make for a good user experience. The UX should be: user presses "save", they see a spinner for a few hundred milliseconds, and then the metadata is updated. Doing that right now is impossible, not only due to the long time the workflow takes to finish. The workflow can randomly fail and it can randomly take a lot longer. To the best o my knowledge, external applications have to busy poll Opencast to know when the workflow is done. During the workflow, you can't use the editor AFAIK. 

But yes, also just the time it takes for the workflow to finish. You can't show a spinner for 20 seconds (or often even more!). Users will think something is broken after 3 seconds. So you have to show a text box explaining that the operation will take a while. The external application also has to somehow "lock" the events from being edited further during the time the workflow runs. That adds lots of complexity. For Tobira it gets even more complicated as we would have to make this work nicely with our normal synchronization code.

I do know that some institutions do non-trivial things on metadata update, like re-generating thumbnails (I ranted about this on the mailing list already) or updating that metadata in external services. But most institutions use Opencast in a way where updating the metadata only updates the Opencast database (and maybe a few files on the NFS). And for those cases, I must say, I find it quite laughable that Opencast takes that long to perform that operation. That's not acceptable IMO. 

In summary: vastly improved user experience, a lot less complexity in external applications, and just because it should not take that long.

(I should add: this is purely from a downstream user perspective. How I want the API to work. This is in no way a comment about how any of this is done internally in Opencast.)

mtneug:
Updating metadata will create a new snapshot. If all current snapshots live in S3, all files (including videos) will first be downloaded to the local asset manager repository, "a few lines" will be changed and the new snapshot will be stored on the local repository. This can take more than 3s. I think this was one of the reasons why Manchester stores the metadata / ACLs in the DB, but this is not upstream right now.

Also, I like the staging workflow: changing metadata / ACLs does not directly change stuff publically and I can later on continue working on metadata or whatever.

LukasKalbertodt:
Thanks for the explanation/insight Matthias. Are you saying that with this PR as is, that would happen? I.e. changing the metadata via API (for example) would trigger what you described and the API would only return a response once those are done?

Note though: from the point of an API consumer (Tobira), that's just a flaw in Opencast's implementation and has nothing to do *whether* changing metadata instantaneously should be possible or not. 
I've also been advocating to store all metadata, ACLs and similar data in the database (not as XML!). So yeah, that would be a better solution. And yes, I am aware that changing this isn't trivial (nor in scope of this PR).

I can also see the benefits of the staging workflow as you describe! But I would strongly suspect that its advantages are used rather rarely. I certainly have seen how we had to explain this "you have to start this workflow after changing metadata" to every new client of ours and every new colleague. Strong indicator that this is not what's intuitive or what users would expect.
And again: the fact that the staging workflow can be useful is no argument whether or not "changing metadata immediately" should be possible.

etc/org.opencastproject.index.service.impl.util.PublicationEventUpdateHandler.cfg
```diff
@@ -0,0 +1,15 @@
+# A comma separated list of publication channel identifiers to automatically republish to when saving to the asset
+# manager.
+# Per default no channel identifiers are listed, which means that no automatic republication happens.
+# The available publication channel identifiers depends on your system configuration.
+# Common publication channels are:
+#   - engage-player: Standard publication for players
+#   - interal: Opencast internal publication (mainly for the editor)
```
rute-santos: Typo here: internal

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -0,0 +1,203 @@
...
+  private DownloadDistributionService downloadDistributionService = null;
+  private StreamingDistributionService streamingDistributionService = null;
+  private SearchService searchService = null;
+
+  @Reference(target = "(distribution.channel=streaming)")
+  protected void setStreamingDistributionService(StreamingDistributionService streamingDistributionService) {
+    this.streamingDistributionService = streamingDistributionService;
+  }
+
+  @Reference(target = "(distribution.channel=download)")
```
rute-santos, line 91: This does not take into account the s3 distribution service; there should be a way to configure which download distributions service to use. We have done this at Harvard-DCE with a configuration in custom.properties that a couple of services use (we can provide more details  if there's interest...)

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -0,0 +1,203 @@
...
+        element.setIdentifier(null);
+        PublicationImpl.addElementToPublication(newPublication, element);
+      }
+
+      mediaPackage.add(newPublication);
+    }
+
+    // Also search service
+    try {
+      searchService.addSynchronously(mediaPackage);
```
rute-santos: This looks weird to me. AFAICT the media package was gotten from the asset manager snapshot so all element urls are archive urls, not the distribution urls???
Also, the previously distributed files are never deleted from the distribution bucket/file system.

karendolan:
> @karendolan I don't know anything about this merge strategy you mentioned, but I can describe the advantages of this PR from my perspective, a Tobira dev.

The merge publish-engage strategy keeps the existing published mediapackage in-tack and updates only the items that have changed or a are being added or removed. The rest of the media and other items remain distributed. We use this to update the ACL and episode catalog of a distributed mediapackage. For example, when changing the series, or updating the title, or adjusting access rights to the published event.

> 
> There are several apps that want to allow users to update video metadata (e.g. Tobira, LMS plugins, ...). Doing that right now is not very robust and does not make for a good user experience. The UX should be: user presses "save", they see a spinner for a few hundred milliseconds, and then the metadata is updated. Doing that right now is impossible, not only due to the long time the workflow takes to finish. The workflow can randomly fail and it can randomly take a lot longer. To the best o my knowledge, external applications have to busy poll Opencast to know when the workflow is done. During the workflow, you can't use the editor AFAIK.

Wow, our site's merge publish-engage workflows complete very quickly. We tied it into the "submit" event of the admin-UI metadata page so that it starts automatically when you click "submit". Then you see the running workflow that ends quickly for us. If your workflows randomly fail, does it mean your regular processing workflows fail a lot as well? It sounds like a bigger issue.

> (I should add: this is purely from a downstream user perspective. How I want the API to work. This is in no way a comment about how any of this is done internally in Opencast.)

It also makes me wonder if media could use this route to be auto-published, as well. Instead of being distributed in the usual way via a workflow which conducts all the correct steps.

karendolan:
FYI - An example of using merge strategy for publish engage
```
    <!-- Publish to engage -->
    <!-- Use "merge" strategy to add new elements or overwrite existing flavors on an existing publication. SKIPPED, if existing pub is not found. -->

    <operation
      id="publish-engage-aws"
      exception-handler-workflow="dce-par-error"
      fail-on-error="true"
      retry-strategy="hold"
      max-attempts="10"
      description="Re-publish to engage">
      <configurations>
        <configuration key="download-source-flavors">dublincore/*,security/*</configuration>
        <configuration key="strategy">merge</configuration>
        <configuration key="check-availability">false</configuration>
      </configurations>
    </operation>
```

Arnei:
> @Arnei what is the issue with running republish-workflow (with "merge" strategy) that this pull resolves?

Pretty much what Lukas, getting changes published ASAP without the overhead of a workflow.

> For example, at our site, when we update event metadata and click submit for the changes, we automatically run a merge workflow. How will this pull impact that?

Could maybe lead to race conditions? However, if the auto-publish breaks anything you can just disable it (it will be disabled by default if/when this gets merged).

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
KatrinIhler: Yeah, that should definitely be checked.
Removal of old elements is a problem with our current merge workflow, too, though.

KatrinIhler (changes requested):
Alright, I didn't look at the code in detail, but I have some concerns:

1. If this is turned on, the old mechanism of updating all events of a series in search on a _series change_  is definitely unnecessary and should be removed/deactivated. Since that handler also updates each event in the archive, triggering a snapshot, that should be enough for this feature to work, no?

2. We should also take another look at the live schedule service which updates events in the search service _if they are live events_. Does the new mechanism do this as well? Or is that only triggered for events that are already processed?

3. The actual update of the publication doesn't seem kosher to me, especially in regards to the search service. We can't just take the media package from the archive, update some elements and put that into search, that's not the way that works. A media package in search is structured differently. (Also we don't remove old elements from the new publication.)
I think you should take a look at SearchUpdatedEventHandler, which is doing pretty much the same thing for series ACLs and metadata of an event.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
Arnei: I'd be interested in a nice way to make the download distribution services configurable :)

Also I'm wondering, is there a world where people would want to auto-republish to multiple distribution services? E.g. the default download distribution as well as S3?

karendolan:
@Arnei so if we wanted to keep the 'merge' strategy of publish-engage, we could by just disabling this new feature? IMHO, the overhead of using the workflow gives us more control over the process than hard coding it.

KatrinIhler:
Honestly, Karen raises a very valid point here - should this maybe happen in a workflow instead? We have similar situations where we start a workflow, e.g. on deletion or when uploading an asset. This allows institutions who have custom publication channels to handle these in an appropriate manner, something that this patch doesn't really allow - even if we update more than just search, we can't really know what should be re-published to these other channels.

JulianKniephoff:
I haven't yet read every comment since I was gone, but just to clarify the workflow vs. no worklfow stuff really quick:

The main goal of this PR is to make metadata/ACL updates automatic **and fast**, specifically fast enough so that they can happen synchronously. As an external application you don't want to manage this weird "we have to wait for your one character change to the title to be processed" state. Stuff like that should really be "set it and forget it," i.e. a Tobira for example sends a metadata update to Opencast and immediately gets a response that it either failed or went through.

If we can achieve that while starting a workflow in the background (and then waiting for its completion to make it synchronous) then sure, why not. But I doubt we would get reasonable response times from that.

Arnei:
> If this is turned on, the old mechanism of updating all events of a series in search on a series change is definitely unnecessary and should be removed/deactivated. Since that handler also updates each event in the archive, triggering a snapshot, that should be enough for this feature to work, no?

Probably would not work right now. The old mechanism also does series specific things like "Override Episode ACL", which we don't do here.

> We should also take another look at the live schedule service which updates events in the search service if they are live events. Does the new mechanism do this as well? Or is that only triggered for events that are already processed?

Define "already processed". If I understand the live schedule service correctly, it triggers a snapshot when it creates or updates a live event, so our PublicationEventUpdateHandler would run.

KatrinIhler:
> Probably would not work right now. The old mechanism also does series specific things like "Override Episode ACL", which we don't do here.

Okay, we could keep that part. But it would be enough for the handler to create a snapshot and leave the updating of search to your code instead of doing that as well.

> Define "already processed". If I understand the live schedule service correctly, it triggers a snapshot when it creates or updates a live event, so our PublicationEventUpdateHandler would run.

The live schedule service currently reacts to both AssetManager and SchedulerService triggers for updating live events. At least the first one would also trigger your code. And yes, the LiveScheduleService itself then creates another snapshot after updating he live publication in search, which would trigger your code...again. One would need to at least test this to see what consequences this has. 

If you need details for this, come talk to me, unfortunately I know the code quite well...

gregorydlogan:
`r/13.x` is reaching EOL here.  Are we going to move forward with this?

JulianKniephoff:
Not in it's current form apparently, and maybe not in 13.x, but we definitely still want this. Let me look into how to best proceed here. I'll assign myself for now, and maybe make this into a draft. (Although I want to reiterate that I don't want to be the person who hits merge on this in the end. 😅)

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -0,0 +1,229 @@
...
+  @Reference
+  protected void setSearchService(SearchService searchService) {
+    this.searchService = searchService;
+  }
+
+  @Activate
+  public void activate(ComponentContext cc) {
+    logger.info("Activating {}", PublicationEventUpdateHandler.class.getName());
+    Object pubChannelIdsValue = cc.getProperties().get(CONFIGURATION_PUBLICATION_CHANNEL_IDS);
+    configurationPublicationChannelIds.clear();
```
lkiesow, line 105: Why clear an empty list?

modules/index-service/src/main/java/org/opencastproject/index/service/impl/util/PublicationEventUpdateHandler.java
```diff
@@ -0,0 +1,229 @@
...
+    if (pubChannelIdsValue != null) {
```
lkiesow, line 106: Since you are already using StringUtils, consider using [`StringUtils::split`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#split-java.lang.String-). It basically does all the numm checks and trimming you are doing here manually.

mtneug:
13.13 was the last 13.x release. @Arnei, please redirect this to `r/14.x` or newer.

karendolan:
> @karendolan I don't know anything about this merge strategy you mentioned, but I can describe the advantages of this PR from my perspective, a Tobira dev.
> 
> There are several apps that want to allow users to update video metadata (e.g. Tobira, LMS plugins, ...). Doing that right now is not very robust and does not make for a good user experience. The UX should be: user presses "save", they see a spinner for a few hundred milliseconds, and then the metadata is updated. Doing that right now is impossible, not only due to the long time the workflow takes to finish. The workflow can randomly fail and it can randomly take a lot longer. To the best o my knowledge, external applications have to busy poll Opencast to know when the workflow is done. During the workflow, you can't use the editor AFAIK.

Are you saying that this pull allows multiple concurrent mediapackage updates on the same mediapackge from different systems at the same time? 

You also stated that a `workflow can randomly fail`. Is your working failing because it has difficulty communicating with an Opencast service? How will this pull solve that problem?

Our site has nearly 10 years of content and our metadata update merge workflow takes seconds. I don't understand why your update workflows are taking so long.

I get that it's frustrating to update the metadata and then launch the workflow. But a new endpoint could do that service for the 3rd party and figure out the callback. 20 seconds is way to long for a merge workflow and that implies that there are other issues with the system.

lkiesow (changes requested):
Okay, since we've contracted ELAN to implement this, let me tell you a bit more about the motivation. I also did another review and went through all the comments to see what is left to do here.

#### Motivation

> **tl;dr** Based on what we saw both in the community and in our own production system we think that makes integrations a lot easier and causes less confusion for users.

We want this to address two main issues we have seen over and over again over the last few years.

First, less important to us, is that users do not understand the process. If you look in our mailing lists or on Matrix you will find a lot of questions about assumed errors were a metadata update did not work. The reason is always that they did not know or understand that they need to run a workflow make the update actually happen in a way that users see them. Except, of course, when you use an integration and use the API to get the metadata which will be updated automatically but then, the player still has the old data. Worse, the behavior is different for series where we already do exactly what this pull request wants to do: we automatically update the metadata 😅

Given that situation, I can really understand why people are asking. Due to the different behavior, it's confusing and arguably quite unintuitive. Especially for new users.


That said, integrations can make things better and if you look at integrations they already try to imitate via API what this pull request is doing: When updating e.g. a title, they update the asset manager via API and then run a workflow. This is used quite extensively as we can see from the thousands of republish metadata workflows run in our system. But if you think about it: Why have every system imitate this, when you can do this in Opencast's core way more effective.

Unfortunately, we have also learned over the years in production that while imitating this in our LMS integration works. We still saw errors on a regular basis since users ran updates very quickly. The reasons for that vary from having multiple people work on data to just mistyping hitting save, noticing and fixing the typo. This did not work because the first update triggered a workflow which would then block the system until it is finished.

While we are not seeing Lukas 20s, depending on system load, dispatch interval, queued updates, … the process can still take a few seconds to complete. Enough time at least that people would actually run into these problems on a regular basis. Remember that this is something you will seldom see with a few people working on this, but you will see with hundreds of users editing stuff.

By now, we also worked around that problem in our integration by caching the changes in our LMS and then applying the bulk updates only every few minutes after the last batch of updates is hopefully done. This state is okayinsh but not great and really pretty complicated.


The final straw was that in our evaluation this year our QA team was looking at the editor and unanimously stated that no one gets what the hell users should expect from the finish section. Again, all the issues described above popped up. They told us that clicking save, users expected the title to actually change. No one understood why they would need to manually start workflows, … Again we could start working around this, but we really wanted to address the root of the problem.


So, that's the motivation of why we really want the situation to change. Thinking about it, we remembered the already existing discrepancy of how episode updates work and how series updates work. While acknowledging that the series implementation is far from being perfect. We thought that we could still do the same without the main issue of the series update (it's very slow because it does a lot of updates).

Updating the publication directly is fast enough to be a synchronous call and implementing that would mean that we can remove a huge amount of workarounds, making the updates faster and the usability more consistent and intuitive.

The downside to this is that you lose the “staging” capability and are no longer able to perform complex operations. The former I feel confident in just ignoring since integrations commonly ignore that as well and most users will never notice and if they notice, they are usually confused :stuck_out_tongue:. The latter could be something that people could add later on an optionally still launch a workflow if they wanted to. With all the advantages and problems that come with.

Given all that, please also remember two things:
- You can disable this behavior already and that switch will stay
- This already happens when updating series

#### Todo

Now, that my rambling is done (and you skipped to this section anyway ;-) let me try to address what I found still needs to be done with this pull request based on what I found and based on the comments above:


1. This should go into `r/15.x` or `develop`. Not in 13. Not in 14.
2. This should also handle S3. It may suffice to just get a reference to `@Reference(target = "(distribution.channel=aws-s3)")` as well and detecting automatically where publications happen. If that doesn't work, adding a manual configuration will probably do as well. If you need an S3 server for testing, @Arnei, please let me know.
3. Updating series will update asset manager and that updates the episodes on search. The series update code probably just does an additional series catalog update which would really just be configuration I think. This works without optimizing that. But we should at least take a look if we can optimize that.


Sorry for the very long text. But I really hope we can get this pushed forward.

- If I missed anything important please add that to my todo list
- The change requests I made are just nitpicks and none of them would block this

mtneug:
I will close this as 13 is EOL now. Feel free to refile for newer versions or redirect and reopen.

lkiesow:
Reopening this since it is still relevant, although it needs to be redirected to another branch.

JulianKniephoff:
Superseded by #6033.

PR #5183, opened 2023-08-24
Change default karaf shutdown command

mliradelc:
The Karaf shutdown command string was set in a string that is not random, but enough random to any who doesn't read carefully can forget easily.

I just changed the string for something more obvious.

* [x] have a concise title
* [x] pass automated tests
* [x] have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

### DIFF
--- a/etc/custom.properties
+++ b/etc/custom.properties
@@ -326,7 +326,7 @@ karaf.shutdown.port.file=${karaf.data}/port
 # Command for shutting down Opencast. If the shutdown port is enabled, Opencast will listen for this command to initiate
 # the shut down procedure.
 # Change this to something secret
-karaf.shutdown.command=3500d4e3-ce93-4ae3-abb4-5e90cef4deb
+karaf.shutdown.command=CHANGE_ME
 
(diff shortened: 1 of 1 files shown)


lkiesow:
I don't mind changing this, but I don't see any good reason to do that for the legacy release, or any release. Can we bring this to `develop`?

mliradelc:
I don't have a strong opinion about change it to develop, but could be considered a security risk? While it is clear in the docs, the experimented admins with opencast normally only read the release notes and the changelog. Thus, it is easy to forget to update something that seems "good as is" when is a critical security configuration.

Source: experience.

lkiesow:
Even changing this doesn't improve security much. That's why I originally set this to something which you don't send accidentally. The command is sent unencrypted via a network port and iirc must be readable from several scripts. It's probably not really hard to get when you have access to the machine.

On the other hand, this is inaccessible from the outside. That means that not just anyone can send a shutdown command.

mliradelc:
Fair enough, I just changed to develop 😄

PR #5191, opened 2023-08-30
Fix NPE when workflow user no longer exists

mtneug:
The Admin UI will not load the list of workflows if the workflow user no longer exists. This adds a check and only extracts additional user infos (name, email) if the user was found.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow (approved):
Confirmed the issue.
Confirmed that this fixes the main issue.

The presentation in the admin interface could be improved (`<> (username)`), but given that the admin interface is deprecated, I don't think it's that important. We should check the new interface, but that's for a different version and thus apart from this pull request.

PR #5226, opened 2023-09-01
Fix Workflow Index Rebuild

KatrinIhler:
Similar fix as in #5115, except for the Workflow Service this time. Setting the batch size to match the page size isn't enough for this service since only a subset of the items on a page are actually indexed. So let's just carry over the last batch to the next page if it didn't reach the batch size instead of skipping it. (If it's the last page, the last batch is always indexed.)


geichelberger:
You're right. I missed the `continue` statement. The bug was not triggered on our system because each Mediapackage had exactly one Workflow associated.

geichelberger (approved):

KatrinIhler:
Since this is approved and was discussed in the Technical Meeting, I'm gonna go ahead and merge this.

PR #5243, opened 2023-09-11
Extend analyze-mp for tag variables

KatrinIhler:
This extends the analyze-mediapackage WOH so it can also create variables containing whether elements with a specific flavor and tag exist, for example `captions_source_hastag_archive`. This allows this information to be used in workflow execution conditions.

These variables aren't created by default (because there can be a lot of them), but only when `set-tag-variables` is true.


KatrinIhler:
Now pointed at 14.x on popular demand.

Arnei (approved):

KatrinIhler:

PR #5247, opened 2023-09-12
Don't take snapshot on asset upload before workflow

KatrinIhler:
The workflow already takes a snapshot, and otherwise we can't figure out which assets are new in the workflow (which I need later).

This also actually makes the code a bit simpler.


Arnei (approved):
Looks good and works as expected.

Since taking a snapshot at the beginning of the asset upload workflow still seems to be a good idea, would it make sense to add a comment on that to `etc/event.upload.asset.options.propertires` (or maybe in the docs, wherever it would make more sense)?

Also, the comment for `startAddAssetWorkflow(JSONObject metadataJson, MediaPackage mediaPackage)` talks at length about a race condition due to taking a snapshot prior to starting a workflow. Since that's not happening anymore, I suppose the potential race condition can't happen anymore either?

KatrinIhler:
@Arnei As we talked about, I will not document the need for a snapshot, as that's fairly obvious, but I did remove the comment about the race condition since it no longer applies.

KatrinIhler:

PR #5248, opened 2023-09-12
Extend Asset Upload Workflow for Captions

KatrinIhler:
In accordance with the recent changes to subtitles in Opencast, captions need special handling when they're uploaded to an existing event, e.g. they must be cut. This extends the existing workflow to handle this use case.

**Attention:** This change depends on #5243 and #5247 to work and should not be merged separately!

Also contains a minor fix in another workflow.

### DIFF
--- a/etc/workflows/partial-error.xml
+++ b/etc/workflows/partial-error.xml
@@ -12,7 +12,7 @@
       fail-on-error="false"
       description="Preserve the current recording state">
       <configurations>
-        <configuration key="source-flavors">*/source,dublincore/*,security/*,captions/*</configuration>
+        <configuration key="source-flavors">*/source,dublincore/*,security/*</configuration>
         <configuration key="source-tags">archive</configuration>
       </configurations>
     </operation>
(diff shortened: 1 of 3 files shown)


gregorydlogan:
I *really* don't like that script.  The relevant bits are in the log, but it's damned near impossible to find:
```
Checking configuration files for tabs or trailing spaces…                                                                                                     
+ grep -rnI '   ' etc                                                                                                                                         
etc/workflows/partial-process-uploaded-captions.xml:66: <configuration key="target-tags">-archive,+generator:unknown,+engage-download</configuration>         
etc/workflows/publish-uploaded-assets.xml:30:   <configuration key="source-flavors">*/*</configuration>                                                       
etc/workflows/publish-uploaded-assets.xml:62:   <configuration key="download-source-tags">engage-download,-exclude-publish</configuration>  <!-- exclude neces
sary bc of download_source_flavors -->                                                                                                                        
+ echo 'Tabs found in config files!'                                                                                                                          
Tabs found in config files!                                                                                                                                   
+ ret=1
```

KatrinIhler:
@gregorydlogan Thanks, fixed! And I also changed my vimrc to highlight tabs.

Arnei (approved):
Makes sense to me, works for me and I couldn't break it.

KatrinIhler:
As this was approved and discussed in the Technical, I'm gonna go ahead and merge this.

PR #5260, opened 2023-09-21
Update ES index before triggering event handlers from AssetManager

KatrinIhler:
The event handler for live events will trigger another snapshot if something in the live publication changed. If we do the Elasticsearch update afterwards, this means the update happens for the new snapshot first, then for the old one, resulting in a lost update. This was broken when this part was changed to be synchronous instead of asynchronous in #3100.


Arnei (approved):

KatrinIhler:
Was approved and talked about in the Technical Meeting, so I'm gonna go ahead and merge this.

PR #5161, opened 2023-08-08
Revert "Enable Tobira adopter stats tracking"

geichelberger:
Reverts opencast/opencast#5040

The admin node is broken because the adopter registration code requires the tobira bundle, which is, for a distributed setup, only available on the presentation node.
```
2023-07-31T14:53:35,919 | ERROR | (BootFeaturesInstaller:113) - Error installing boot features
org.apache.felix.resolver.reason.ReasonException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=opencast-admin; type=karaf.feature; version="[13.8.0,13.8.0]"; filter:="(&(osgi.identity=opencast-admin)(type=karaf.feature)(version>=13.8.0)(version<=13.8.0))" [caused by: Unable to resolve opencast-admin/13.8.0: missing requirement [opencast-admin/13.8.0] osgi.identity; osgi.identity=opencast-adopter-registration-impl; type=osgi.bundle; version="[13.8.0,13.8.0]"; resolution:=mandatory [caused by: Unable to resolve opencast-adopter-registration-impl/13.8.0: missing requirement [opencast-adopter-registration-impl/13.8.0] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.opencastproject.tobira.impl)(version>=13.8.0)(!(version>=14.0.0)))"]]
        at org.apache.felix.resolver.Candidates$MissingRequirementError.toException(Candidates.java:1343) ~[?:?]
        at org.apache.felix.resolver.ResolverImpl.doResolve(ResolverImpl.java:392) ~[?:?]
        at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:378) ~[?:?]
        at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:332) ~[?:?]
        at org.apache.karaf.features.internal.region.SubsystemResolver.resolve(SubsystemResolver.java:257) ~[?:?]
        at org.apache.karaf.features.internal.service.Deployer.deploy(Deployer.java:401) ~[?:?]
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.doProvision(FeaturesServiceImpl.java:1063) ~[?:?]
        at org.apache.karaf.features.internal.service.FeaturesServiceImpl.lambda$doProvisionInThread$13(FeaturesServiceImpl.java:998) ~[?:?]
        at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
[...]

PR #5262, opened 2023-09-22
Re-enable Tobira Adopter Tracking

gregorydlogan:
This PR reenables Tobira adopter tracking, reverting #5161.  To accomplish this we needed to do a few things:

- Make `TobiraEndpoint` findable from the service registry
- Expose the relevant stats via REST endpoint
- Find `TobiraEndpoint` from the adopter registration code
- Pull the data between the nodes.

In the code then, we:

- Revert the previous PR
- Refactor `RemoteBase` to pull the inter-node service lookup code into its own utility class
- Make `TobiraEndpoint` a `JobProducer` to get it registered in the service registry (and thus findable from another node)
- Use the new `RemoteHostConnector` class to find and query the service, regardless of which node it lives on.

Notably, `TobiraEndpoint` does not produce jobs, and refuses to process any that are created (manually).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
@@ -38,6 +38,7 @@ import org.opencastproject.security.api.DefaultOrganization;
 import org.opencastproject.security.api.Organization;
 import org.opencastproject.security.api.OrganizationDirectoryService;
 import org.opencastproject.security.api.SecurityService;
+import org.opencastproject.security.api.TrustedHttpClient;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.security.api.User;
 import org.opencastproject.security.api.UserProvider;
@@ -134,6 +135,8 @@ public class ScheduledDataCollector extends TimerTask {
 
   protected TobiraRemoteRequester tobiraRemoteRequester;
 
+  protected TrustedHttpClient trustedHttpClient;
+
 
   //================================================================================
   // Properties
@@ -181,8 +184,6 @@ public class ScheduledDataCollector extends TimerTask {
     }
     this.sender = new Sender(Objects.toString(serverBaseUrl, DEFAULT_STATISTIC_SERVER_ADDRESS));
 
-    this.tobiraRemoteRequester = new TobiraRemoteRequester();
-
     // Send data now. Repeat every 24h.
     timer = new Timer();
     timer.schedule(this, 0, ONE_DAY_IN_MILLISECONDS);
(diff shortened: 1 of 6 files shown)


modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
```diff
@@ -408,4 +442,29 @@ public void setOrganizationDirectoryService(OrganizationDirectoryService orgDirS
     this.organizationDirectoryService = orgDirServ;
   }
 
+  @Reference
+  public void setTrustedHttpClient(TrustedHttpClient trustedHttpClient) {
+    this.httpClient = trustedHttpClient;
+  }
+
+  private class TobiraRemoteRequester extends RemoteBase {
+    private static final String SERVICE_TYPE = "org.opencastproject.tobira";
```
gregorydlogan, line 450: Magic constant is magical, but I didn't think adding a compile time dependency was worth it here?

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -204,4 +245,88 @@ private static Response badRequest(String msg) {
...
+  }
+
+  @Override
+  protected OrganizationDirectoryService getOrganizationDirectoryService() {
+    return organizationDirectoryService;
+  }
+
+  @Override
+  protected String process(Job job) throws Exception {
+    throw new RuntimeException("This should never be called for TobiraEndpoint!");
```
gregorydlogan, line 339: Deliberately raising a `RuntimeException` to crash the SR in dev if anyone is silly enough to start building Tobira jobs.

gregorydlogan:
Just thought of a cleaner way to do this, converting to draft until I can reimplement.

gregorydlogan:
Removed the previously existing utility class.  Ready for review.

KatrinIhler:
Okay, I spent way too much time thinking about this and you won't like the result:
Why don't we put the Tobira stats into the database? Maybe with a timestamp to see if they are still current? And overwriting any old statistics once the new one comes in? That would also mean they survive a reboot.
This solution isn't too bad, but I really don't like the TobiraEndpoint becoming a JobProducer. :sweat_smile:

gregorydlogan:
Conclusion here is that Greg will rework this into using the db. We don't really *need* the db but we probably will for other things later.

gregorydlogan:
Moving this to 15.x since it's not going to happen in time for the last 14.x release.

mtneug:
Note: This PR is targeting the legacy branch of Opencast (`r/15.x`). In [a recently passed proposal](https://groups.google.com/a/opencast.org/g/dev/c/VwYqk3vXdd0), we changed how we deal with PRs to legacy branches.

Our goal is to provide stability and minimize the risk of breaking changes. Ultimately, users should trust that patch releases of legacy versions can be installed easily.

To this end, developers targeting legacy branches now need to explain why those changes are necessary and can't just go into stable or develop. We'd appreciate it if you could provide such an explanation or, alternatively, retarget this PR to stable or develop.

Arnei:
The recent release of Opencast 15.12 is intended to be the final minor release for 15. Please retarget your PR or explain why it warrants another Opencast 15 release.

gregorydlogan:
This will be refiled someday, but updating this as is isn't going to happen anytime soon.

PR #5267, opened 2023-09-26
Paella 7: Add Focus Layout for any Flavors

### LINKED ISSUE #5258: How selecting streams should work in the player
> This is related to #5107, but that one is more about the specific issue that Paella would potentially fail to properly display multiple streams at all.

The layout manager of Paella player lets you configure specific layouts based on which you can show videos and which you can select in the layout selector of the player. For example, you can configure Paella to display `presenter/delivery` on the left and `presentation/delivery` on the right. Unfortunately, you have to hard-code the assumptions of how videos should be displayed and which flavors should be handled in the configuration, essentially providing all possible combinations that could occur.

I've seen this to cause problems on several occasions. For example, if you publish only one flavor that Paella does not know of, you just get an error (#2046), if you publish multiple streams, it is hard to or nearly impossible to configure all combinations of videos (#5107) or if you change your flavor naming at some point, you might accidentally drop half your streams with no visible notification that something is wrong (seen at UOS). This is even more problematic since older Opencast players (Engage, Theodul) would support all these cases just fine.

That is why I suggest adding just two simple, yet more flexible default layouts to Opencast's Paella player which allow you to either focus on one stream, or to show all available streams in a grid (row?).

As a default I suggest using a layout showing all available (published) video streams. This is basically #5107. We could turn this into a grid instead of a row if we have a lot of videos. But that's not that important. Showing all by default is what is essential. This would look something like this:

Screenshot from 2023-09-19 16-59-42

The second default layout should be a view focussing on one video with the other ones being displayed as small stream at the side like this (maybe make the non-focused even smaller):

Screenshot from 2023-09-19 16-59-42-focus

We can always switch between the layouts using the layout selector, but it would also be great if I could just click on one video in the grid layout to focus on that video and to click on a small video in the focus layout to focus that one.

In addition, we might also want to be able to filter down the flavors. But I think that's optional for the initial implementation since I don't necessarily know if anyone needs that, and since configuring these as the default layouts doesn't mean that the old layouts cannot be configured any longer. You still can if you want the old behavior.

dennis531:
Adds a new focus layout plugin to paella. It allows any flavor constellations and to maximize a video via button controls. 

Additional changes:
- The video layout background has the same color as the video container

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

Close #5258

### DIFF
--- a/etc/ui-config/mh_default_org/paella7/config.json
+++ b/etc/ui-config/mh_default_org/paella7/config.json
@@ -234,6 +234,18 @@
             ],
             "tabIndexStart": 40
         },
+        "org.opencast.paella.focusVideoStaticLayout": {
+            "enabled": true,
+            "validContent": [
+                {
+                    "id": "focus-video",
+                    "content": [],
+                    "icon": "present-mode-4.svg",
+                    "title": "Focus"
+                }
+            ],
+            "tabIndexStart": 40
+        },
 
         "/* video formats plugins */": "*********************************************************",
         "es.upv.paella.hlsVideoFormat": {
--- a/modules/engage-paella-player-7/public/default_theme/theme.css
+++ b/modules/engage-paella-player-7/public/default_theme/theme.css
@@ -32,6 +32,14 @@
   /* --video-container-padding: 10px; */
 }
 
+/* dark mode */
+@media (prefers-color-scheme: dark) {
+  :root {
+    --video-container-background-color: #000000;
+    --base-video-rect-background-color: #000000;
+  }
+}
+
 
 
 
--- a/modules/engage-paella-player-7/public/default_theme/theme.css
+++ b/modules/engage-paella-player-7/public/default_theme/theme.css
@@ -22,7 +22,7 @@
   --main-bg-gradient: #1F2937;
   /* --main-border-color: rgba(125,125,125,0.4); */
   /* --video-container-background-color: #e4e4e4; */
-  --base-video-rect-background-color: #9CA3AF;
+  --base-video-rect-background-color: #e4e4e4;
 
   /* video-container-message */
   /* --video-container-message-bkg: rgba(0, 0, 0, 0.4); */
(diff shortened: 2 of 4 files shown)


dennis531:
Should the old layouts, especially the dual layout (presenter and presentation), still be activated?

Arnei:
Played around with this a little bit and I'm not quite sure what the focus layout is trying to accomplish. Looking at a dual stream video on develop.opencast.org, there already is a "Maximize video" button that results in a layout similar to the focus layout. So what is the benefit of the focus layout?

Furthermore, the focus layout wastes spaces by introducing a "border" around the videos. And the layout does not adapt if there is only one video present, resulting in more wasted space.

Sorry if this sounds mean, I'm probably missing some context. I do agree that the UX of focusing a video in the dual stream layout could be improved. But I don't see why the focus layout is a good solution. Wouldn't it be better to improve on what's already there?

dennis531:
> Played around with this a little bit and I'm not quite sure what the focus layout is trying to accomplish. Looking at a dual stream video on develop.opencast.org, there already is a "Maximize video" button that results in a layout similar to the focus layout. So what is the benefit of the focus layout?

In comparison to the existing layout, the focus layout allows to display any video flavours and more than two videos. The existing layout only supports configured flavours.

> Furthermore, the focus layout wastes spaces by introducing a "border" around the videos. And the layout does not adapt if there is only one video present, resulting in more wasted space.

I hadn't considered the case that only one video can be present. The border problem occurs as static layouts define a fixed area (1280x720) on which the videos can be placed. The existing layout uses a dynamic layout where the videos are simply placed in a row and the focused video is set to a higher width.

Arnei:
> In comparison to the existing layout, the focus layout allows to display any video flavours and more than two videos. The existing layout only supports configured flavours.

So the goal of this PR is not actually to add a layout that allows the focus of one video, but add a layout that shows all published videos. Never would've guessed that from "Focus", probably better to call it something else. Maybe "All videos"? Should probably also be documented somewhere.

gregorydlogan:
Is this PR going to be moving forward?  If not, please close it.

dennis531:
We close this pull request and integrate the feature into the auto layout so that a single video can be enlarged using buttons. A new pull request will be created for this function.

PR #5269, opened 2023-09-29
Fix custom roles pattern now working in access policy tab

### LINKED ISSUE #5016: ACL are missing since the UI change
**Describe the bug**

Hi, we are experimenting with two issues since the new UI change in the ACL.

1.- You can't search for External Roles
2.- While the ACL from the existing events is selected, the ACL doesn't show the users, only the local roles.

acl error

**To Reproduce**

Enable ACL with LDAP

**Expected behavior**

It should show the users. and the role of external search should work.

Important info: We are using LDAP. The logging and roles work as intended, only the ACL panel in the admin UI.

I'm not very well versed with the front end. If someone can help me debug this issue, I can find and fix the problem's root. The dev tools from the browser show that roles exist and are retrieved.

Arnei:
Intends to fix #5016.

The issue was that the user part of the access policy tab in the details view in the admin ui would not allow for custom role patterns. With this PR, you should now able to add custom roles in the user part.

Note that you still have to respect your custom role pattern. So if your pattern is something like `ROLE_USER_[A-Z][A-Z0-9]*`, you still have to type the `ROLE_USER_` part. I suppose that could be improved upon, but I decided it's not worth the effort.

Entering a custom role for a user:
Bildschirmfoto vom 2023-09-29 10-51-40
After selecting the custom role. The custom role will show as ROLE, not as a user, until the view is refreshed:
Bildschirmfoto vom 2023-09-29 10-52-39
After saving and refreshing. The view now attempts to get user information for the custom role and display it.
Bildschirmfoto vom 2023-09-29 10-53-03

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -220,11 +220,21 @@ angular.module('adminNg.controllers')
 
               UserResource.get({ username: id }).$promise.then(function (data) {
                 policy.user = data;
+                $scope.policiesUser.push(policy);
+                // Did we get a user not present in Opencast (e.g. LDAP)? Add it to the list!
+                if ($scope.users.map(user => user.id).indexOf(id) == -1) {
+                  if ($scope.aclCreateDefaults['sanitize']) {
+                    // FixMe: See the FixMe above pertaining to sanitize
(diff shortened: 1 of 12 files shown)


mliradelc (approved):
Tried locally and our dev cluster, outstanding work! Even can retrieve the email from the user in LDAP, like in your original PR. 

Codewise is all ok from my part.

PR #5307, opened 2023-10-03
Modified dispatch.interval in Job Dispatcher to allow float values

### LINKED ISSUE #5300: Configure dispatch interval in milliseconds
It would be nice to be more precise when configuring the main dispatch interval:

https://github.com/opencast/opencast/blob/4970c798a3f25783f40d9b7b5196a7673bdf5ec2/etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg#L8

But right now, this value is read as an integer and interpreted as seconds:

https://github.com/opencast/opencast/blob/4970c798a3f25783f40d9b7b5196a7673bdf5ec2/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java#L206-L236

It would be great if we could interpret the configuration as float and then configure the scheduler with millisecond precision.

debanjanc01:
This closes https://github.com/opencast/opencast/issues/5300


lkiesow (approved):
Thanks for taking this on.
Reading through the patch, it looks good.

PR #5315, opened 2023-10-06
Improve Performance of Permission Check in AssetManager

KatrinIhler:
During the permission check in the AssetManager, each security property of a media package (containing a role and an action) is compared to each role of the user simultaneously. This can create a large number of threads. 

We have seen this actually cause exceptions in a production system, however I couldn't reproduce these locally, maybe not enough other things going on in an idle system. Still I felt like it wouldn't do any harm to reduce this a bit, so this filters the security properties by the action we're looking for before doing the comparing.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -1081,7 +1081,7 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
         return getDatabase().selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
                 .map(p -> p.getId().getName())
                 .filter(p -> p.endsWith(action))
-                .anyMatch(p -> roles.parallelStream().anyMatch(r -> r.equals(p)));
+                .anyMatch(p -> roles.stream().anyMatch(r -> r.equals(p)));
     }
   }
 
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -1080,6 +1080,7 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
                 .collect(Collectors.toList());
         return getDatabase().selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
                 .map(p -> p.getId().getName())
+                .filter(p -> p.endsWith(action))
                 .anyMatch(p -> roles.parallelStream().anyMatch(r -> r.equals(p)));
     }
   }


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1080,6 +1083,7 @@ private boolean isAuthorized(final String mediaPackageId, final String action) {
                 .collect(Collectors.toList());
         return getDatabase().selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
                 .map(p -> p.getId().getName())
+                .filter(p -> action.equals(getActionFromPropertyName(p)))
```
lkiesow: Simpler and more efficient still would probably be to just use [`String:endsWith`](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#endsWith-java.lang.String-):
```suggestion
                .filter(p -> p.endsWith(action))
```

KatrinIhler: You're totally right, and this simplifies the code a lot.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1080,6 +1080,7 @@ private boolean isAuthorized(final String mediaPackageId, final String action) {
                 .collect(Collectors.toList());
         return getDatabase().selectProperties(mediaPackageId, SECURITY_NAMESPACE).parallelStream()
                 .map(p -> p.getId().getName())
+                .filter(p -> p.endsWith(action))
                 .anyMatch(p -> roles.parallelStream().anyMatch(r -> r.equals(p)));
```
geichelberger: Wouldn't it make sense to remove the second parallel stream additionally? Most of the time, it won't even run in parallel because the default ForkJoinPool does not have that many threads.

```suggestion
                .anyMatch(p -> roles.stream().anyMatch(r -> r.equals(p)));
```

KatrinIhler: No idea tbh. Maybe somebody else with more knowledge can chime in here?

gregorydlogan: IMO removing it probably makes sense, but won't really matter perf wise.  How many roles is any one MP likely to have?  Running a regex against the full list sequentially is going to be as fast, if not faster than splitting it up into N substreams and doing the same thing against those in parallel.  But in the odd case where there are a truly astounding number of roles, is it *really* going to be that much faster to split them up?  I somehow doubt it.

In short, time(splitting substreams + attendant baggage + N * subsearch) > time(search).

KatrinIhler: >  How many roles is any one MP likely to have?

You haven't seen what I've seen. ;)

But then I guess let's do this as well.

PR #5317, opened 2023-10-06
Add config-keys to set minimum rules for cuts and segments duration

mliradelc:
The editor WoH that applies the Opencast Built-in editor has some hidden configurations that are not user reachable.

These configurations are:

- Minimum segment duration: If a segment has a lower duration as this parameter, it will be discarded.
- Minimum Cut segment duration: If the cut between two segments is less than the threshold, it will be discarded.

The PR adds these two parameters to the `etc/org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg` file.

Ready for Review

~~It will keep as a draft because I need to fix this undesirable log message:~~

```
2023-10-06 15:07:15,981 | WARN  | (LoadUtil:95) - No hosts found that can process jobs of type segments.min.duration with load 100.0
2023-10-06 15:07:15,982 | WARN  | (LoadUtil:95) - No hosts found that can process jobs of type segments.min.cut.duration with load 100.0
```

~~That error is because I used the same method to read the properties as `jobLoad` variable. I will fix it soon.~~

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
+++ b/modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
@@ -583,10 +583,14 @@ public class VideoEditorServiceImpl extends AbstractJobProducer implements Video
     logger.debug("Properties updated!");
 
     jobload = LoadUtil.getConfiguredLoadValue(properties, JOB_LOAD_KEY, DEFAULT_JOB_LOAD, serviceRegistry);
-    segmentsMinDuration = (int) LoadUtil.getConfiguredLoadValue(properties, SEGMENTS_MIN_DURATION_KEY,
-        Float.valueOf(DEFAULT_SEGMENTS_MIN_DURATION), serviceRegistry);
-    segmentsMinCutDuration = (int) LoadUtil.getConfiguredLoadValue(properties, SEGMENTS_MIN_CUT_DURATION_KEY,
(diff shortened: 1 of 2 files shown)


modules/videoeditor-ffmpeg-impl/src/main/java/org/opencastproject/videoeditor/impl/VideoEditorServiceImpl.java
```diff
@@ -312,7 +324,8 @@ protected Track processSmil(Job job, Smil smil, String trackParamGroupId) throws
 
       // If we are cutting video/audio, use ffmpeg
       if (videoclips.size() > 0) {
-        List<VideoClip> cleanclips = sortSegments(videoclips);    // remove very short cuts that will look bad
+        // remove very shortcuts that will look bad
```
Arnei:
```suggestion
        // remove very short cuts that will look bad
```

Arnei (approved):
Did not run this, but the code looks reasonable to me.

Maybe add an example to the explanation for `segments.min.cut.duration`, it's not the most intuitive parameter.

mliradelc:
> Did not run this, but the code looks reasonable to me.
> 
> Maybe add an example to the explanation for `segments.min.cut.duration`, it's not the most intuitive parameter.

I just expanded the explanation, what do you think?

Arnei:
Fine to me. Just want to get rid of the commented-out lines :)

mliradelc:
@Arnei repeated commented code deleted, please merge as you wish 😄

PR #5342, opened 2023-10-18
Add option to pass additional args to Whisper

KatrinIhler:
This allows us to pass additional arguments to Whisper without having to deal with every single option separately. 

For example, I'm using this to pass the location for the model to Whisper with `--model_dir`, since at least whisper-ctranslate2 defaults to the home directory of the user... but if your user doesn't have a home directory, this doesn't work. This is also the reason I'm aiming this at 13.x, though this might be controversial.


Arnei (approved):
Makes sense to me.

I think this can go into r/13.x, since it's completely opt-in.

KatrinIhler:

PR #5349, opened 2023-10-19
Youtube-Publication: Throw explicit error if client_secret is malformed

Arnei:
Parsing of the client_secret.json may go wrong, but currently the stack trace is not very helpful in making clear when that happens. This should help.

### DIFF
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeV3PublicationServiceImpl.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeV3PublicationServiceImpl.java
@@ -179,6 +179,8 @@ public class YouTubeV3PublicationServiceImpl
   public synchronized void activate(final ComponentContext cc) {
     super.activate(cc);
     properties.setBundleContext(cc.getBundleContext());
+
+    logger.debug("Activated Youtube Publication Service");
   }
 
   @Override
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeV3PublicationServiceImpl.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeV3PublicationServiceImpl.java
@@ -180,7 +180,7 @@ public class YouTubeV3PublicationServiceImpl
     super.activate(cc);
     properties.setBundleContext(cc.getBundleContext());
 
-    logger.debug("Activated Youtube Publication Service");
+    logger.debug("Activated YouTube Publication Service");
   }
 
   @Override
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
@@ -95,19 +95,19 @@ public final class ClientCredentials {
    * @throws java.io.FileNotFoundException
    * @throws java.io.IOException
    */
-  private String getValueFromArray(final File file) throws IOException, ParseException {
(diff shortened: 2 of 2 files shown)


modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeV3PublicationServiceImpl.java
```diff
@@ -179,6 +179,8 @@ public YouTubeV3PublicationServiceImpl() throws Exception {
   public synchronized void activate(final ComponentContext cc) {
     super.activate(cc);
     properties.setBundleContext(cc.getBundleContext());
+
+    logger.debug("Activated Youtube Publication Service");
```
JulianKniephoff:
```suggestion
    logger.debug("Activated YouTube Publication Service");
```

modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
```diff
@@ -99,7 +99,16 @@ private String getValueFromArray(final File file) throws IOException, ParseExcep
     final JSONParser parser = new JSONParser();
     final FileReader reader = new FileReader(file);
     final JSONObject jsonObject = (JSONObject) parser.parse(reader);
+
+    if (!jsonObject.containsKey("installed")) {
+      throw new IOException("client_secret file does not contain \"installed\" as a top level key. Did you set the "
```
JulianKniephoff: Not sure if `IOException` is the right thing to throw here. I would say in this context `IllegalArgumentException` would be more appropriate (since the given `File` argument doesn't have the necessary structure) or **maybe** `java.text.ParseException`. (**Note**, this is not the `ParseException` that's already imported by this file.) Because the `File` can't be parsed into the format needed to extract the thing. (From this POV it's not just JSON, but JSON with a "schema" if you will.)

modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
```diff
@@ -99,7 +99,16 @@ private String getValueFromArray(final File file) throws IOException, ParseExcep
...
+          + "type for your 'OAuth 2.0-Client-ID' correctly?");
+    }
     final JSONObject array = (JSONObject) jsonObject.get("installed");
+
+    if (!array.containsKey("client_id")) {
+      throw new IOException("client_secret file does not contain \"client_id\".");
```
JulianKniephoff: See above.

modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
```diff
@@ -99,7 +99,16 @@ private String getValueFromArray(final File file) throws IOException, ParseExcep
...
+    }
     return (String) array.get("client_id");
```
JulianKniephoff, line 113: Note that the JavaDoc is kinda sorta lying now.

JulianKniephoff (dismissed):

JulianKniephoff:
Was able to test this by now as well and it seemed to work. 👌

modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/auth/ClientCredentials.java
Arnei: Fair point, caught me being lazy about exceptions types there.

KatrinIhler (approved):
I'm not 100% happy with the IllegalArgumentException, but having two different ParseExceptions is also weird and this does the job. So LGTM.

PR #5350, opened 2023-10-19
YT-Publish WOH: Fail more aggressively

Arnei:
Currently the Youtube Publish WOH is a bit meek, often only logging little warnings when it fails to publish. This makes the WOH more assertive by having it throw exceptions instead, properly failing the workflow.

### DIFF
--- a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishYouTubeWorkflowOperationHandler.java
+++ b/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishYouTubeWorkflowOperationHandler.java
@@ -122,13 +122,12 @@ public class PublishYouTubeWorkflowOperationHandler extends AbstractWorkflowOper
       // Look for elements matching the tag
       final Collection<MediaPackageElement> elements = elementSelector.select(mediaPackage, true);
       if (elements.size() > 1) {
-        logger.warn("More than one element has been found for publishing to youtube: {}", elements);
-        return createResult(mediaPackage, Action.SKIP);
+        throw new WorkflowOperationException("More than one element has been found for publishing to youtube: "
+            + elements);
       }
 
       if (elements.size() < 1) {
-        logger.info("No mediapackage element was found for publishing");
-        return createResult(mediaPackage, Action.CONTINUE);
+        throw new WorkflowOperationException("No mediapackage element was found for publishing");
       }
 
       Job youtubeJob;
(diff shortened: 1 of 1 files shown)


JulianKniephoff (approved):
Code lgtm. Will try to get a test in later as well.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishYouTubeWorkflowOperationHandler.java
```diff
@@ -122,13 +122,12 @@ public WorkflowOperationResult start(final WorkflowInstance workflowInstance, Jo
...
-        logger.warn("More than one element has been found for publishing to youtube: {}", elements);
-        return createResult(mediaPackage, Action.SKIP);
+        throw new WorkflowOperationException("More than one element has been found for publishing to youtube: "
+            + elements);
       }
 
       if (elements.size() < 1) {
-        logger.info("No mediapackage element was found for publishing");
-        return createResult(mediaPackage, Action.CONTINUE);
+        throw new WorkflowOperationException("No mediapackage element was found for publishing");
```
JulianKniephoff: I tested at least these and they indeed work and fail the workflow. 👌

JulianKniephoff (approved):

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishYouTubeWorkflowOperationHandler.java
KatrinIhler: Should this really fail? I think "Skipped" was the right result in this case. We mark other operations as skipped when the selector doesn't match any elements as well.

KatrinIhler (changes requested):
I don't agree with one of the cases (see ~~below~~ above), the rest looks fine.

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishYouTubeWorkflowOperationHandler.java
Arnei: Fair enough, might've been a little too eager there.

KatrinIhler: Actually I just realized that it returns Action.CONTINUE there. I think it should be Action.SKIP to make it obvious nothing happened. Could you do that? :sweat_smile: I can't make a suggestion on unchanged code unfortunately.

Arnei: Can do that. Not the scope of this PR, but should it be SKIP in other WOHs too then? Like https://github.com/opencast/opencast/blob/cfc7bf1b6d7b713d71e107041844338075c0b38a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/PublishEngageWorkflowOperationHandler.java#L379

KatrinIhler: In my opinion yes. I already did this for the tagging operation a while ago, see #5108. But unfortunately this is wildly inconsistent right now.

KatrinIhler (approved):
Great!

PR #5351, opened 2023-10-19
Return ISO 639 language code for language detected by Whisper

KatrinIhler:
When Whisper uses language detection, it returns the language name in English instead of the ISO 639 code for some reason. But we need the latter for the language tag of the track. This makes an attempt to translate the language name in English to a matching two-letter language code using locales.

### DIFF
--- a/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
+++ b/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
@@ -45,6 +45,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


mliradelc:
Thanks for the PR! Very curious, it seems that was done in some update for whisper. Have you check if the other engine `ctranslate-2` carries the same issue?

mliradelc:
@KatrinIhler What version of whisper are you using actually?

KatrinIhler:
> Thanks for the PR! Very curious, it seems that was done in some update for whisper. Have you check if the other engine `ctranslate-2` carries the same issue?
What version of whisper are you using actually?

I actually should have been more precise: I verified this for the current version of `whisper-ctranslate2`. I looked at the code of OpenAI's whisper and it _seems_ that it does the same thing, but if you could confirm this, that would be great.

I actually assumed that this had always been the case and we just didn't notice - there's no error, the tag is just lang:English insteda of lang:en which actually looks almost fine in the player selection. But either way, if this code doesn't find a match it falls back to the original value, so that should be fine.

Arnei (approved):
Seems reasonable and works.

PR #5352, opened 2023-10-19
Set generator tags in speech-to-text WOH

KatrinIhler:
Since we know the subtitles were automatically generated and by which engine, might as well set the respective tags so we don't have to do that in the workflow.


Arnei (approved):
Does what it says which I think is a good thing.

KatrinIhler:

PR #5353, opened 2023-10-20
Remove metadata layer from preview image (presenter)

Piriya-VM:
This PR removes the WOH `cover-image` from the default `partial-publish` workflow. The presenter preview image will no longer contain any metadata. Here is the [discussion](https://groups.google.com/a/opencast.org/g/users/c/42vfpsuFjXY) to this topic.


KatrinIhler (approved):
Also changes indentation, but that's already inconsistent, so it's whatever. Looks fine otherwise.

oas777:
Puh! Thanks everyone.

PR #5354, opened 2023-10-20
Get thumbnail for editor frontend from archive

Arnei:
Intends to fix https://github.com/opencast/opencast-editor/issues/1154

When generating a thumbnail in the editor frontend, saving it and then reloading the editor, the generated thumbnail would not show up. This is because thumbnails are loaded from the internal publication, but the generated thumbnails are only saved to the archive.

This changes it so that thumbnails are loaded from the archive. If they cannot be found there, they are loaded from the internal publication instead.


KatrinIhler (approved):
Looks fine!

PR #5359, opened 2023-10-24
Docs: Hide bot PRs in changelog

Arnei:
When looking through the changelog in the docs to see what has changed, most people will not care about the semi-automatic dependency updates. Instead, they pose a hindrance as they need to be mentally filtered out.

This PR moves all dependabot PRs to the bottom of the version they appear in, and hides them away in a html dropdown.

Related helper scripts PR: https://github.com/opencast/helper-scripts/pull/71

### DIFF
--- a/docs/guides/admin/docs/css/extra.css
+++ b/docs/guides/admin/docs/css/extra.css
@@ -160,13 +160,6 @@ a.header-link:hover {
     opacity: 1;
 }
 
-details {
-  border-left: 2px solid silver;
-  padding: 5px;
-  margin-bottom: 10px;
-  margin-left: 10px;
-}
-
 details summary {
   color: darkblue;
   cursor: pointer;
--- a/docs/guides/admin/docs/changelog.md
+++ b/docs/guides/admin/docs/changelog.md
@@ -103,18 +103,6 @@ Opencast 14
   Drop orphan statistics database index
 - [[#4943](https://github.com/opencast/opencast/pull/4943)] -
   Update Paella 7 default theme
-- [[#4942](https://github.com/opencast/opencast/pull/4942)] -
-  Bump engine.io from 6.4.1 to 6.4.2 in /modules/admin-ui-frontend
-- [[#4936](https://github.com/opencast/opencast/pull/4936)] -
-  Bump grunt-contrib-cssmin from 4.0.0 to 5.0.0 in /modules/admin-ui-frontend
-- [[#4935](https://github.com/opencast/opencast/pull/4935)] -
-  Bump @types/react-dom from 18.2.0 to 18.2.1 in /modules/lti
-- [[#4934](https://github.com/opencast/opencast/pull/4934)] -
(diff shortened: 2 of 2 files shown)


docs/guides/admin/docs/css/extra.css
```diff
@@ -160,13 +160,6 @@ a.header-link:hover {
     opacity: 1;
 }
 
-details {
-  border-left: 2px solid silver;
-  padding: 5px;
-  margin-bottom: 10px;
-  margin-left: 10px;
-}
-
```
ziegenberg, line 169: Why remove the styling for `details`? That's used in other places as well.

Arnei: I figured this was a quick and easy way to make the changelog look better. I also looked at the other places `details` is used in, and found this acceptable. But I don't have a strong opinion on this, if you want this reverted I'm not against it.

Arnei: Pictures for comparison (the second ones are with the removed style):
Bildschirmfoto vom 2023-10-25 09-34-48
Bildschirmfoto vom 2023-10-25 09-36-06
Bildschirmfoto vom 2023-10-25 09-37-23
Bildschirmfoto vom 2023-10-25 09-37-08

KatrinIhler:
Maybe the description could be a bit easier to understand, something like "Dependency updates"? Because not everyone knows what bot pull requests are, but that's what we're using them for right now.

No hard opinion on the styling, both look fine.

Also maybe a stupid question, but why do the bot PRs have to be formatted in HTML instead of markdown?

Arnei:
> Maybe the description could be a bit easier to understand, something like "Dependency updates"? Because not everyone knows what bot pull requests are, but that's what we're using them for right now.

Makes sense, will change to "Dependency updates". If we get bots to do anything else in the future, we can always change it again then.

> Also maybe a stupid question, but why do the bot PRs have to be formatted in HTML instead of markdown?

Markdown does not do dropdowns (or at least I can't figure out how), but you can do them in HTML. But as soon as you do something in HTML, mkdocs will not treat anything inside the HTML as markdown anymore (which is in accordance with markdown spec afaik), so you then have to fully commit to HTML.

KatrinIhler (approved):
I think this looks fine. If we want to revert the styling change later, it'll be easy to do so.

ziegenberg:
Removing the styling makes it impossible, in some cases, to see what exactly was un-collapsed.

Screenshot from 2023-10-27 11-37-14

ziegenberg:
Here you can see it very easily.

Screenshot from 2023-10-27 11-24-48

ziegenberg:
And it got merged. Well, why bother waiting for a response from my side... 🤷

KatrinIhler:
Ah, I see, I got this turned around. In Arne's comment, the _first_ ones are actually those with the removed style, not the second ones. That makes more sense.

@ziegenberg Could you take a quick look at #5367?

Issue #5364, opened 2023-10-26
OutOfMemory: Java heap space Error when populating index


NgoniChoga:
Java heap space Error when populating index in `modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java:602`

This appears to happening as a result of holding too many media packages in memory during a reindex. Code causing the OutOfMemory error is located at, `modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java:227`:

```
List<Tuple<MediaPackage, String>> mediaPackageList = new LinkedList<>();
try {
  for (SearchEntity entity : searchEntities) {
    MediaPackage mediaPackage = MediaPackageParser.getFromXml(entity.getMediaPackageXML());
    mediaPackageList.add(Tuple.tuple(mediaPackage, entity.getOrganization().getId()));
  }
} catch (Exception e) {
  logger.error("Could not parse series entity: {}", e.getMessage());
  throw new SearchServiceDatabaseException(e);
} 
```

**To Reproduce**

1. Attempt to reindex an opencast instance with a large set of MediaPackages.

**Expected behavior**

Reindex should complete successfully.

**Stack trace**

```
ERROR o.o.s.i.SearchServiceImpl [FelixStartLevel] bundle opencast-search-service-impl:15.0.0.SNAPSHOT (160)[org.opencastproject.search.impl.SearchServiceImpl(322)] : The activate method has thrown an exception
java.lang.OutOfMemoryError: Java heap space
        at java.lang.StringUTF16.compress(StringUTF16.java:160) ~[?:?]
        at java.lang.String.<init>(String.java:3214) ~[?:?]
        at java.lang.String.<init>(String.java:276) ~[?:?]
        at org.apache.xerces.xni.XMLString.toString(Unknown Source) ~[org.apache.servicemix.bundles.xerces-2.12.0_1.jar:?]
        at org.apache.xerces.parsers.AbstractDOMParser.characters(Unknown Source) ~[org.apache.servicemix.bundles.xerces-2.12.0_1.jar:?]
        at org.apache.xerces.impl.XMLDocumentFragmentScannerImpl.scanContent(Unknown Source) ~[org.apache.servicemix.bundles.xerces-2.12.0_1.jar:?]
[...]

PR #5365, opened 2023-10-26
Stream media packages to avoid memory issues

NgoniChoga:
### Your pull request should…

This PR addressed https://github.com/opencast/opencast/issues/5364.

We are experiencing a java heap space Error when populating the index. As a solution we have implemented streaming to that section of code.

This appears to happening as a result of holding too many media packages in memory during a reindex. Code causing the OutOfMemory error is located at, modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java:227:

List<Tuple<MediaPackage, String>> mediaPackageList = new LinkedList<>();
try {
  for (SearchEntity entity : searchEntities) {
    MediaPackage mediaPackage = MediaPackageParser.getFromXml(entity.getMediaPackageXML());
    mediaPackageList.add(Tuple.tuple(mediaPackage, entity.getOrganization().getId()));
  }
} catch (Exception e) {
  logger.error("Could not parse series entity: {}", e.getMessage());
  throw new SearchServiceDatabaseException(e);


gregorydlogan:
Funny that you should find this @NgoniChoga, I was just going to fix this inside of #3921 :)

I'm assuming you've done this in prod?  It looks right to me, but I don't have a sufficiently large set of mediapackages to validate that this works as expected/

gregorydlogan:
Funny that you should find this @NgoniChoga, I was just going to fix this inside of #3921 :)

I'm assuming you've done this in prod?  It looks right to me, but I don't have a sufficiently large set of mediapackages to validate that this works as expected/

NgoniChoga:
> Funny that you should find this @NgoniChoga, I was just going to fix this inside of #3921 :)
> 
> I'm assuming you've done this in prod? It looks right to me, but I don't have a sufficiently large set of mediapackages to validate that this works as expected/

@gregorydlogan we actually caught this in our staging environment. We have run it successfully now using this update.

PR #5404, opened 2023-11-03
Add role for listprovider external api

### LINKED ISSUE #5402: API Listproviders endpoint needs a ROLE!
**Describe the bug**

The API list providers now work fine also with API version attached to the request. However, by passing an Authorization header to that call the response ends up in 401 Forbidden, which means there need to be a `listprovider` specific ROLE (like `ROLE_API_LISTPROVIDERS_VIEW`) for the API user to add. 

**To Reproduce**

1. In a clear—not-logged-in session, try to make the HTTP request to the `/api/listproviders/*`
2. Add the Authorization header, just like with default API username and password
3. Make a GET call to one of the `listproviders` endpoint (i.e. `/api/listproviders/providers.json` or `/api/listproviders/LANGUAGES.json`)
4. See the response, which is always 401 Forbidden!

**Expected behavior**

Either:

1. Having a separated ROLE like `ROLE_API_LISTPROVIDERS_VIEW` to set for the API user.
2. (Recommended) Make it accessible when the API user has `ROLE_API`

The first approach, although it seems to be more practical, is not of the favour from LMS side. Because it forces the admins to edit their API user in opencast and if they don't pay attention to the manual or helper docs (Who reads them 😅). They always end up with forbidden error!

***Previous Work of this topic***
Issue (https://github.com/opencast/opencast/issues/5308) => PR (https://github.com/opencast/opencast/pull/5344)

Arnei:
Fixes #5402.

Add a rule for the listprovider endpoints of the external api to the security xml, so that the endpoints are actually usable. Also creates a new role to go with the rule.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


lkiesow:
Can we change this to target `develop`?

JulianKniephoff:
Just for the record, and @lkiesow please correct me if you feel I misrepresent anything: The consensus in the technical meeting seemed to me to be that this is fine in `13.x`.

If there is no further objection, I will probably merge this in the coming days if no one beats me to it? 🤷‍♀️

PR #5407, opened 2023-11-06
Fix manually entering time in `datetimepicker`-s in Safari

JulianKniephoff:
When you try to manually edit the time of a `datetimepicker` (for example trying to change the start date/time metadata field of an event), instead of using the combo boxes from the picker, Safari does weird things to your focus, which makes the task really difficult.

This is a [known issue](https://github.com/trentrichardson/jQuery-Timepicker-Addon/issues/903) in the (seemingly unmaintained) `jqueryui-timepicker-addon` library. Luckily this patch seems to work around it.

It's a bit of a hack, but the thing we are tampering with here
is related to a feature we are not using, so I don't see the harm. 🤷‍♀️

See also https://trentrichardson.com/examples/timepicker/#alt_examples
in the tab "Options" and "Examples" to (try to) understand
what the option means (you might also need
https://api.jqueryui.com/datepicker/#option-altField)
and so verify that the bug is not Opencast, but in the library already.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/directives/datetimepickerDirective.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/directives/datetimepickerDirective.js
@@ -29,6 +29,38 @@ angular.module('adminNg.directives')
     return lc;
   }
 
+  // When you edit the time in the input field directly,
+  // instead of using any of the picker's means,
+  // the `timepicker` addon syncs that change with it's internal state,
+  // to reflect it in its controls.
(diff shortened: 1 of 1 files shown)


JulianKniephoff:
Actually never mind, this **doesn't** work around the problem. It was working for a minute, though! 🤔

JulianKniephoff:
Quite a bit uglier, but this should do it. 👀

I don't think there is a nicer way without going upstream, which I don't have high hopes about.

PR #5413, opened 2023-11-06
Add search index sort multi field

geichelberger:
The default sorting of the search index could be more intuitive. For example, lowercase letters follow uppercase letters, which leads to funny results that by sorting ascending, the letter `a` comes after `Z`.

By introducing a multi-field `sort` for selected fields, it is possible to ~normalize the string to lowercase, and with asciifolding~ use icu_collation_keyword field type which defaults to DUCET Collation a best effort attempt to better sort results.

https://www.elastic.co/guide/en/elasticsearch/plugins/current/analysis-icu-collation-keyword-field.html

An even better approach would be to use the `icu_collation` filter, which can be configured to use a collation table for a specific language.

Expected:
grafik

Current:
grafik

* [ ] have a concise title
* [ ] pass automated tests


ziegenberg:
I would suggest going for the `icu_collation` filter which can be configured.

gregorydlogan:
Considering that reindexing is already on the table for 16.0, is this going to be moved forward?

geichelberger:
It's on my todo list for OC 16

ziegenberg:
Installing the ICU plugin is not particularly difficult: `bin/opensearch-plugin install analysis-icu` :)

geichelberger:
### Current vs ICU DUCET collation
grafik

docs/guides/admin/docs/installation/debs.md
```diff
@@ -77,6 +77,10 @@ accomplished by running the following:
 If you wish to use the upstream OpenSearch repository directly be aware that Opencast only supported with OpenSearch 1.x
 and will not work with OpenSearch 2.x yet.  Future support for this is forthcoming.
 
+To install the ICU plugin for OpenSearch, run the following:
```
ziegenberg: We should tell the people why we need the ICU Plugin. Currently, this dependency is not justified in the docs.

ziegenberg (changes requested):

KatrinIhler:
The general tendency in the dev meeting was that making people re-index the whole thing instead of just the search part would be acceptable, especially since OC 16 is already a release with many breaking changes. So we probably don't want this to be configurable or only do the search index, but go all in.

docs/guides/admin/docs/installation/debs.md
gregorydlogan: This should also clarify that things *will not work* without this call. I would absolutely skip this as a naive just-testing adopter, and then be confused when it all looks like it should work, but doesn't.

docs/guides/admin/docs/installation/rpm-el.md
```diff
@@ -70,6 +70,8 @@ Make sure to install it on the node which also serves the admin interface.
 dnf install elasticsearch-oss
 ```
 
+
```
gregorydlogan, line 75: Presumably the rpms need the same call made as the debs?

gregorydlogan (commented):
Reminder that we need to incorporate this into #5597, or to incorporate that mapping file into this one, depending on who gets merged first.

geichelberger:
It would be easy to update the index without a complete rebuild.

**Series Index Example:**

`PUT /opencast_series/_mapping`
``` json
{
  "properties": {
    "title": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "description": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "subject": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "creator": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "organizers": {
      "copy_to": "organizer",
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "contributors": {
      "copy_to": "contributor",
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "publisher": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    },
    "rights_holder": {
      "type": "keyword",
      "fields": {
        "sort": {
          "type": "icu_collation_keyword",
          "index": false
        }
      }
    }
  }
}
```

Followed by:

` POST /opencast_series/_update_by_query?refresh&conflicts=proceed`

Update by query has many possible options. The version above is the simplest, and it may be enough, even if you have 200k entries, which are peanuts for elasticsearch.

gregorydlogan:
This seems like something we should add to the docs as an option for those leery of doing a full rebuild!

gregorydlogan:
@geichelberger this needs a note about the reindex, but then it should go into `develop` so it can be in for 16.0

mtneug:
Is this still targeting 16?

docs/upgrade/15_to_16/update_event_mapping.json
```diff
@@ -0,0 +1,83 @@
+{
```
gregorydlogan: These should be in 16_to_17 right?

gregorydlogan (approved):
Works as expected.  Should probably move the upgrade scripts, right?

docs/upgrade/15_to_16/update_event_mapping.json
geichelberger: Yes, I moved the files.

KatrinIhler:

PR #5419, opened 2023-11-08
Set publication variables in analyze-MP

KatrinIhler:
Similar to #5243 except this creates variables that contain which publication channels are part of the media package. These new variables are not created by default.

The use case for this was that I wanted to publish everything if the event wasn't published before, including encoding and all, but only merge specific assets into the publication if it existed. Here's a snippet if you want to test this for yourself:

```
    <operation
      id="include"
      description="Publish to Engage"
      if="NOT ${publication_engage_player_exists}">
      <configurations>
        <configuration key="workflow-id">partial-publish</configuration>
      </configurations>
    </operation>

    <operation
      id="include"
      description="Update Engage publication"
      if="${publication_engage_player_exists}">
      <configurations>
        <configuration key="workflow-id">partial-publish-generated-captions</configuration>
      </configurations>
    </operation>

```

Since the other one went into 14.x I'm aiming this at that branch as well, but I can aim at develop instead if somebody disagrees.

This also contains a small change to the docs that explains which WOHs can generate workflow variables, because I keep forgetting this.

### DIFF
--- a/docs/guides/admin/docs/configuration/workflow.md
+++ b/docs/guides/admin/docs/configuration/workflow.md
@@ -329,7 +329,7 @@ As the formal description above explains, such boolean expressions may contain�
 - …strings, which must be surrounded by single-quotes. Escaping of single quotes is supported, just use two single
   quotes next to each other: `'foo''bar'`
 - …as well as references to the variables of the workflow instance that contain these data types. Variables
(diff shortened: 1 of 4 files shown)


Arnei (approved):
Seems reasonable to me and works.

PR #5420, opened 2023-11-09
Paella 7: Hide/Show timeline bar on live events

miesgre:
Paella player used to start playing live streams from the beginning of the buffer, so in case the stream had a time buffer, paella started playing from the beginning of the buffer and not from the current moment.

This PR updates paella to version 1.46.1 that fixes that bug and also adds a property to the configuration file that allows to show or hide the time bar in live events.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/engage-paella-player-7/src/js/EpisodeConversor.js
+++ b/modules/engage-paella-player-7/src/js/EpisodeConversor.js
@@ -106,7 +106,7 @@ function getSourceData(track, config) {
   return data;
 }
 
-function getMetadata(episode) {
+function getMetadata(episode, config) {
   const { duration, title, language, series, seriestitle, subjects, license, type } = episode.mediapackage;
   const startDate = new Date(episode.dcCreated);
   const presenters = episode?.mediapackage?.creators?.creator
@@ -119,6 +119,9 @@ function getMetadata(episode) {
       ? episode.mediapackage.contributors.contributor
       : [episode.mediapackage.contributors.contributor])
     : [];
+
+  const isLive = episode?.mediapackage?.media?.track?.some((track) => track.live === true);
+  const visibleTimeLine = !(isLive && config?.hideTimeLineOnLive);
 
   const result = {
     title,
@@ -136,7 +139,8 @@ function getMetadata(episode) {
     location: episode?.dcSpatial,
     UID: episode?.id,
     type,
-    opencast: {episode}
+    opencast: {episode},
+    visibleTimeLine
   };
 
   return result;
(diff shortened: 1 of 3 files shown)


KatrinIhler (dismissed):
I'm gonna roll this out on our testing system in the next few days, but here are my first thoughts:

- As far as I can tell this doesn't actually update the player version (maybe because that already happened), just adds a new config option, right?

- I think we want to aim this at least at OC 15 (which isn't out yet, but will be soon), maybe even OC 14, since this is basically a fix for people who have live streams where you can go back in time.

- But in that case (and maybe in general), we shouldn't change the default behavior (which was hiding the timeline for live events if I understood correctly), just allow people to change it if they want to. So set the new config to true and in case the config key is missing, retain the previous behavior (this is probably easier if we flip the config from ` hide...` to ` show...` because then the default is `false`).

Sorry for the long wait!

Edit: I might have to retract some of that - if I understand correctly, the default behavior of Paella was changed a while ago and the default is now to always show the timeline. This adds a config to turn that timeline _off_ for live events, right? In that case, disregard everything I said, except I think this can still go into OC 15.

miesgre:
> I'm gonna roll this out on our testing system in the next few days, but here are my first thoughts:
> 
> * As far as I can tell this doesn't actually update the player version (maybe because that already happened), just adds a new config option, right?

paella-core needs tu be updated to 1.46.1, but `dependabot` updated it in commit https://github.com/opencast/opencast/commit/d48978491b1b0d2973fa5baedeb6bab46c5a16fa

> * I think we want to aim this at least at OC 15 (which isn't out yet, but will be soon), maybe even OC 14, since this is basically a fix for people who have live streams where you can go back in time.

Should I rebase to r/15.x? r/14.x?

KatrinIhler:
Let's aim this at r/15.x for now. If somebody complains, we can still backport this later.

mliradelc (approved):
I agree with the changes

mliradelc:
Agree with @KatrinIhler to be in `r/15.x` if someone needs for `r/14.x` they can back port this PR.

modules/engage-paella-player-7/src/js/EpisodeConversor.js
```diff
@@ -120,6 +120,9 @@ function getMetadata(episode) {
       : [episode.mediapackage.contributors.contributor])
     : [];
 
+  const isLive = episode?.mediapackage?.media?.track?.some((track) => track.live === true);
```
KatrinIhler, line 123: This breaks if there's only one track with "x.some is not a function" u_u

KatrinIhler: Fix is out, see below!

PR #3173, opened 2021-11-15
Ensure JWT Feature is Installed

lkiesow:
This patch updates the documentation to reflect that the JWT feature is
currently disabled by default (pull request #3065).

To allow the feature to be easily activated, this also adds the feature
to the installed features, essentially ensuring that it is packed in all
distributions.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #3065, opened 2021-10-12
Deactivate security-jwt by default

lkiesow:
As discussed at the technical meeting, due to potential problems causing
Opencast to not start up properly, this patch disables the module
`security-jwt` by default, allowing users to activate them using the
`opencast-security-jwt` Karaf feature.

We hope to have this enabled by default again in Opencast 11.

@gregorydlogan will include a note about the deactivation in the 10.4 release notes.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #5422, opened 2023-11-09
(Re-)Enable the JWT feature by default

JulianKniephoff:
The feature was initially enabled by default by virtue of it not actually being a (Karaf) feature, but having its bundles be part of `opencast-core` directly. It was later disabled by default in #3065 because of some flaky boot issues observed by Innsbruck. That's where the Karaf feature comes from, which you can install and enable in your site configuration if you are not affected by that issue, as most adopters weren't. In #3173 we added some documentation to that effect and re-included the feature in the distributions, so that you only need to enable it.

Since we haven't heard anyone else complain so far and JWT is become more important, we decided to give it another go and just enable it again. In that spirit, this PR reverts the documentation parts of #3173, and adds the feature created by #3065 to the default boot features of all the distribtuions.

I opted not to just revert #3065 because I felt there is a shift towards a more modular Opencast in the community which I'm generally supportive of. There was some talk in Matrix hinting at the preference for this feature to be a plugin, the plugin manager itself is an independent Karaf feature, there is also the CAS feature, etc.

Unfortunately, the JWT feature currently can't be a plugin because it contains a fragment bundle hosted by `kernel`, so if it were to be activated by the plugin manager after boot, `kernel` would have to restart, which Opencast doesn't survive atm. But we can honour the idea of more modularity by keeping the feature instead of just reintegrating the bundles into `opencast-core`.

Plus this has the added benefit of being rather easily deactivated if need be: Just do the reverse of what the docs that this PR throws out say and remove the feature from the boot features of your Karaf config.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/assemblies/dist-admin/pom.xml
+++ b/assemblies/dist-admin/pom.xml
@@ -111,6 +111,7 @@
             <!-- Opencast features -->
             <feature>opencast-admin</feature>
             <feature>opencast-pluginmanager</feature>
+            <feature>opencast-security-jwt</feature>
           </bootFeatures>
         </configuration>
       </plugin>
--- a/assemblies/dist-adminpresentation/pom.xml
+++ b/assemblies/dist-adminpresentation/pom.xml
@@ -111,6 +111,7 @@
             <!-- Opencast features -->
             <feature>opencast-adminpresentation</feature>
             <feature>opencast-pluginmanager</feature>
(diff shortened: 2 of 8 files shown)


geichelberger (approved):
LGTM

Opencast starts with and without configuring JWT.

PR #5433, opened 2023-11-23
Prepare opencast 15 release notes

Arnei:
A little late, but this adds the release notes and related stuff for Opencast 15.0 to the documentation.


KatrinIhler:
Could you take this one into account as well? #5434 I merged the mentioned PR a while ago and then realized it might be useful to mention that change in the release notes, finally did this just now...

PR #4617, opened 2023-01-23
Adds the 'add-force-flavors' config to the publish-engage workflow

marwyg:
You can use this new configuration if the publish strategy 'merge' is activated. Elements with these flavors will be added to an existing publication without deleting or overwriting existing elements.

For example, you can add a track with another resolution, without running the whole publishing workflow again.

PR #5434, opened 2023-11-24
Add release notes for PR 4617

KatrinIhler:
See #4617.


lkiesow:
Since #4617 has been merged in October, it will be part of the Opencast 15 release and this should probably go into `r/15.x`, not `develop`. Other than that, this is good to be merged.

KatrinIhler:
> Since #4617 has been merged in October, it will be part of the Opencast 15 release and this should probably go into `r/15.x`, not `develop`. Other than that, this is good to be merged.

You're right, but since Arne already picked this up in #5433, I can just close this. :D

PR #5436, opened 2023-11-30
Whisper transcription engine bug fixing

lkiesow:
This patch fixes a number of minor bugs and inconveniences in the whisper transcription engine. This includes loading of default values when the configuration has been modified which incorrectly kept their old values in some cases. It also does proper debug logging, not filtering out the actual output, which is what you want for debugging.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
+++ b/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
@@ -46,7 +46,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
-import java.util.regex.Matcher;
+import java.util.Objects;
 import java.util.regex.Pattern;
 
 /** Whisper implementation of the Speech-to-text engine interface. */
@@ -85,11 +85,7 @@ public class WhisperEngine implements SpeechToTextEngine {
   /** Config key for quantization */
   private static final String WHISPER_QUANTIZATION = "whisper.quantization";
 
-  /** Quantization for whisper-ctranslate2 */
-  private enum Quantizations {
-    auto, int8, int8_float16, int16, float16, float32
-  }
-  private Option<Quantizations> quantization = Option.none();
+  private String quantization;
 
   /** Config key for Voice Activity Detection */
   private static final String WHISPER_VAD = "whisper.vad_enabled";
@@ -104,7 +100,7 @@ public class WhisperEngine implements SpeechToTextEngine {
   private static final String WHISPER_ARGS_CONFIG_KEY = "whisper.args";
 
   /** Currently used Whisper args */
-  private String[] whisperArgs = {};
+  private String[] whisperArgs;
 
   /** Map to get ISO 639 language code for language name in English */
   private Map<String, String> languageMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -187,10 +174,10 @@ public Map<String, Object> generateSubtitlesFile(File mediaFile,
       transcriptionCommand.add(language);
     }
 
-    if (quantization.isSome()) {
-      logger.debug("Using quantization {}", quantization.get());
+    if (quantization != null) {
```
marwyg, line 177: I don't know if this is important. If you comment the configuration out, the variable will be null at this point. But if you comment it in and no value is configured (like this `whisper.quantization=`), it will become an empty string here and an error will be thrown later. Would an additional check for the empty string be beneficial at this point?

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
```diff
@@ -117,31 +113,22 @@ public String getEngineName() {
...
-    if (!StringUtils.isBlank(t)) {
-      quantization = Option.some(Quantizations.valueOf(t));
-    }
+    quantization = Objects.toString(prop.get(WHISPER_QUANTIZATION), null);
     logger.debug("Whisper quantization set to {}", quantization);
 
-    isVADEnabled = OsgiUtil.getOptCfgAsBoolean(cc.getProperties(), WHISPER_VAD);
-    logger.debug("Whisper Voice Activity Detection  set to {}", isVADEnabled);
+    isVADEnabled = OsgiUtil.getOptCfgAsBoolean(prop, WHISPER_VAD);
+    logger.debug("Whisper Voice Activity Detection set to {}", isVADEnabled);
```
marwyg: The log outputs something like that: "Whisper Voice Activity Detection set to Some(true)" 😂 But it works, so I think it doesn't matter.

marwyg (changes requested):
I tested several cases. It seems to work. I have just 2 little things.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
lkiesow: Not that this was different before… but it's fixed now. Instead of printing the `Opt` value, it is now printing the boolean value. So, now you should get something like:

> Whisper Voice Activity Detection set to true

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
lkiesow: I can see the argument for both sides:
- Setting a key to an empty string should be a valid input. The admin set this to `""` after all. Why should we discard their input. What if a key is e.g. for displaying additional text? Shouldn't you be able to set an empty text?
- Setting a key to an empty string is basically setting this to nothing. If nothing is configured, the default value should apply.

Unfortunately, Opencast is not very consistent when it comes to evaluating empty strings :-(
That also means, to be sure any given setting uses the default, you must comment it out right now. Otherwise, it might be the default… or not.

That's why I think that evaluating it like this isn't a problem unless we change all of Opencast's settings to consistently evaluate `""` as `null` and fall back to the default value (or throw an exception if the setting is required).

I also actually prefer this for two reasons:

- Evaluating this is very easy since you can usually just use Java's built-in `Objects::toString(prop_value, default)` to parse the configuration value, handle `null` values and set defaults. No need to build more complex code or load any additional libraries.
- You actually handle the case of an empty string being valid quite easily and don't have to worry about that. If an admin wants to set that as a value, they can.

That said, I don't have a strong opinion and while I don't think it will really help, I can make sure an empty string is turned into `null` as well if you want me to.

lkiesow:
@marwyg, are you happy with the change and explanation? Can we merge this into `r/15.x` before the 15.0 release tomorrow?

marwyg:
Sorry for the late reply. Everything looks fine to me.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperEngine.java
marwyg: Okay, let us keep it as it is then :)

lkiesow:
Thanks for the review → merging

PR #5472, opened 2023-12-05
Fix concurrency problem in Whisper integration

### LINKED ISSUE #5437: Concurrent whisper jobs may fail
Concurrent whisper jobs may overwrite each other's output causing the transcription of one event to end up attached to another event.

lkiesow:
The SpeechToTextService interface ensures that a unique file is passed to transcription engines which these can safely use to store output files. Unfortunately, most Whisper implementations work slightly different and you cannot specify a single output file. That is probably why the Whisper engine throes the safe file name overboard and just uses the parent directly to store files.

This has the side-effect that if `mediapackage-a/presenter.mp4` and `mediapackage-b/presenter.mp4` are processed concurrently, both write to the same file and an incorrect output file will end up being attached to one of the media packages.

This patch fixes that problem by creating an unique directory instead and passing that to the engines to operate in.

This also fixes the problem that most Whisper implementations create many more output files which whould have never been deleted before, slowly filling up the workspace.

This fixes  #5437

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/speech-to-text-api/src/main/java/org/opencastproject/speechtotext/api/SpeechToTextEngine.java
+++ b/modules/speech-to-text-api/src/main/java/org/opencastproject/speechtotext/api/SpeechToTextEngine.java
@@ -22,10 +22,27 @@
 package org.opencastproject.speechtotext.api;
 
 import java.io.File;
-import java.util.Map;
 
 /** Interface for speech-to-text implementations. */
 public interface SpeechToTextEngine {
+
+  class Result {
+    private final String language;
+    private final File subtitleFile;
+
+    public Result(String language, File subtitleFile) {
+      this.language = language;
+      this.subtitleFile = subtitleFile;
+    }
+
+    public String getLanguage() {
+      return language;
+    }
+
+    public File getSubtitleFile() {
+      return subtitleFile;
+    }
+  }
 
   /**
    * Returns the name of the implemented engine.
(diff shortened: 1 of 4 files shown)


mliradelc:
Great PR, I just checked it, and it's a good idea to have a new class to handle the result from the engine rather a Map.

I have a question, I though, that each media package has its folder within the workspace, is it not the case?

Should this target to r/14.x rather than r/15.x?

lkiesow:
> I have a question, I though, that each media package has its folder within the workspace, is it not the case?

Services can use the workspace however they want. Usually, if you load a file from a workspace into the workspace, it will end up in the `workspace/mediapackage/<id>` structure. But services usually operate in their own collections. That means that they need to ensure that files  they write are safe for concurrent operations.

lkiesow:
> Should this target to r/14.x rather than r/15.x?

I don't really care about 14 with the release of 15 next week. But I can port this to 14 if you want me to.

mliradelc:
> But services usually operate in their own collections. That means that they need to ensure that files they write are safe for concurrent operations.

That sounds like a feature to develop in the future, or a check when services are developed.

mliradelc (approved):
All good from my side, the new class makes the engine code clear and avoids code duplication.

mliradelc:
> I don't really care about 14 with the release of 15 next week. But I can port this to 14 if you want me to.

Yes, please!   🙏

lkiesow:
@mliradelc, rebased to `r/14.x`.

wsmirnow (approved):
I haven't run it due to missing whisper setup. But the code looks good for me.

wsmirnow:
@mliradelc Can you give it a try on OC 14?

snoesberger:
Thanks for this fix!
I have successfully tested this PR on OC 14. I uploaded two different video files with the same filename and started the workflows to generate captions for both events at the same time. The captions were generated without error and assigned to the correct event.

wsmirnow:
Thank you all. I will merge this.

PR #5481, opened 2023-12-13
Fix Paella Player 7 login redirect

KatrinIhler:
Our code for the Paella Player 7 checks the user role to decide whether to redirect to the login page. However, the user role can be different if you have configured a different user role prefix or simply turned sanitize off. A safer assumption seems to be to expect any logged-in user to have ROLE_USER.


Arnei (approved):

KatrinIhler:
Merging since it got approved & discussed. :)

PR #5487, opened 2023-12-18
Add maven wrapper

geichelberger:
The Maven wrapper is handy if you need to install the necessary Maven version but cannot upgrade/downgrade or don't want to install Maven at all.

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/guides/developer/docs/participate/infrastructure/maven-repository.md
+++ b/docs/guides/developer/docs/participate/infrastructure/maven-repository.md
@@ -17,7 +17,7 @@ There is a preconfigured Docker image for a Nexus server set-up for Opencast. To
 follow these steps:
 
     docker run \
-        --name ./mvnwcache \
+        --name mvncache \
         -p 8000:8000 \
         docker.io/lkiesow/opencast-maven-repository
 
--- a/docs/guides/developer/docs/participate/infrastructure/maven-repository.md
+++ b/docs/guides/developer/docs/participate/infrastructure/maven-repository.md
@@ -17,7 +17,7 @@ There is a preconfigured Docker image for a Nexus server set-up for Opencast. To
 follow these steps:
 
     docker run \
-        --name mvncache \
+        --name ./mvnwcache \
         -p 8000:8000 \
         docker.io/lkiesow/opencast-maven-repository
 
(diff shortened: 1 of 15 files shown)


.mvn/wrapper/maven-wrapper.jar
lkiesow, line 1: Looking at the script, this file will be downloaded automatically by the wrapper script if you use it, isn't it? So, we probably don't need to include this?

mvnw.cmd
```diff
@@ -0,0 +1,205 @@
+@REM ----------------------------------------------------------------------------
```
lkiesow: We don't support Windows and we know that Opencast won't build on Windows. We probably don't need this file, do we?

lkiesow (changes requested):
Should we also include additional patterns in `.gitignore`?

.mvn/wrapper/maven-wrapper.jar
geichelberger: If the jar file is not found, the script tries to download it. Maven wrapper supports different types of integration.
The default one is with the jar file, and then there is a source, script, and only script option.
https://maven.apache.org/wrapper/#usage-without-binary-jar

mvnw.cmd
geichelberger: I will remove the file, I almost removed it from the staging area.

mtneug:
Is it out of scope to "grep-and-replace" `mvn` with `./mvnw`, especially in the documentation?

gregorydlogan:
@mtneug agree with you, and easy enough to do in a spare minute.  Since I made the changes someone else (you?) should probably review.  Basic ./mvnw work seems to work just fine in my testing.

docs/guides/developer/docs/participate/infrastructure/maven-repository.md
```diff
@@ -17,7 +17,7 @@ There is a preconfigured Docker image for a Nexus server set-up for Opencast. To
 follow these steps:
 
     docker run \
-        --name mvncache \
+        --name ./mvnwcache \
```
mtneug: @gregorydlogan too much "grep-and-replace" 😉 

```suggestion
        --name mvncache \
```

mtneug (changes requested):

mtneug (approved):

.mvn/wrapper/maven-wrapper.jar
mtneug: If including the JAR is the default, then this is fine by me. I have only used it in that way as well.

PR #5492, opened 2023-12-20
Add support for custom actions in ACL to Tobira harvest API

LukasKalbertodt:
Previously, only read and write roles were transmitted.

This can go into 14 as it is not a breaking change. Well, one thing changes: `ROLE_ADMIN` is not explicitly included in the event ACL anymore. But since this is not a public free-for-all API, but only used by Tobira, and since Tobira has implicit `ROLE_ADMIN` checks everywhere, this is no problem. In fact, Tobira even removes `ROLE_ADMIN` in the incoming ACL before storing it in the DB.

Related issue: https://github.com/elan-ev/tobira/issues/1004

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Harvest.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Harvest.java
@@ -23,6 +23,7 @@ package org.opencastproject.tobira.impl;
 
 import org.opencastproject.search.api.SearchQuery;
 import org.opencastproject.search.api.SearchService;
+import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.series.api.SeriesException;
 import org.opencastproject.series.api.SeriesService;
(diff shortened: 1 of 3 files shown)


KatrinIhler:
I assume this can't really be tested yet bc the Tobira side of it isn't done yet, right? But I will still deploy this on our testing system to see if it breaks anything (and so we don't have to do this later, bc just updating Tobira is easier. :D)

Can already say that the code looks reasonable.

LukasKalbertodt:
> the Tobira side of it isn't done yet, right?

Correct. 

In addition to making sure nothing breaks, you could check `/tobira/harvest?since=0&preferredAmount=10` and make sure that the custom actions are actually mentioned in the response. Might need some fiddling with the parameters.

KatrinIhler:
> In addition to making sure nothing breaks, you could check `/tobira/harvest?since=0&preferredAmount=10` and make sure that the custom actions are actually mentioned in the response. Might need some fiddling with the parameters.

Looks exactly like I expected :)

KatrinIhler (approved):

wsmirnow:
@KatrinIhler do you want to merge this before 14.8 cut today?

PR #5526, opened 2024-01-03
Document how to turn static file auth off for Tobira usage.

gregorydlogan:
This PR documents how to turn off static file auth since Tobira currently requires that.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/configuration/serving-static-files.md
+++ b/docs/guides/admin/docs/configuration/serving-static-files.md
@@ -29,17 +29,10 @@ access control lists published to Engage before allowing access.
 The configuration options for requiring authentication can be found in the service's configuration file at
 `etc/org.opencastproject.fsresources.StaticResourceServlet.cfg`.
 
-For some external software, like the ILIAS plugin, it is required to set `authentication.required` to `false` as static files authorization is active by default.
+For some external software, like the ILIAS plugin and Tobira, it is required to set `authentication.required` to `false` as static files authorization is active by default.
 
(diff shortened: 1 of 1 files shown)


lkiesow:
Doesn't it make way more sense to document this in Tobira? Because no one will ever remember to update this if something changes over there.

gregorydlogan:
It's already documented in the second point of [Tobira's docs](https://elan-ev.github.io/tobira/setup/requirements#what-tobira-requires-of-opencast), but those docs don't include the *how* of how to disable SFA.  I'm not sure if makes sense to document the how part in the Tobira docs, hence putting it here.

What about genericising the section into a plain 'how to turn SFA off', not referencing Tobira.  I'm not sure how commonly that would be needed by anyone else, but it's also not a bad thing to have documented.

LukasKalbertodt:
> What about genericising the section into a plain 'how to turn SFA off', not referencing Tobira

That sounds useful to me. I mean, ideally Tobira should work with SFA (and we are kind of low-key working on it), but having a short section in the doc very simply describing how to disable SFA seems useful either way.

... though, I just checked the docs and there is already this sentence:

 > For some external software, like the ILIAS plugin, it is required to set authentication.required to false as static files authorization is active by default.

Third paragraph under "Securing Static Files". So that should be good enough? Maybe add Tobira as second example? :shrug:

gregorydlogan:
Changes made -> merging.

PR #5537, opened 2024-01-09
Fix ACL template selection breaking after first selection

Arnei:
The template selection dropdown for access policies in e.g. the event details is broken. After selecting an access policy, when trying to select another access policy, the rules from the access policy above the selected access policy will be selected.
For example, if you have `public, authenticated, private` and select `public`, you get the aces for `public`. If you then select `authenticated` you get the aces for `public`. Then selecting `private` will lead to an error.

This fixes that. Apparently the angularjs select thingy breaks if you set ng-model to a value that is not in ng-options (e.g. setting it ng-model to `0`, but ng-options only contains `951, 952, 953`). This commit sets it to `undefined` instead, which angularjs seems to be able to handle.

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
+++ b/modules/admin-ui-frontend/app/scripts/modules/events/controllers/eventController.js
@@ -484,6 +484,7 @@ angular.module('adminNg.controllers')
           $scope.access = EventAccessResource.get({ id: id }, function (data) {
             if (angular.isDefined(data.episode_access)) {
               $scope.baseAclId = data.episode_access.current_acl.toString();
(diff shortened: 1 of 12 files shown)


KatrinIhler:
Ha, hadn't even noticed this yet, but could reproduce the bug on our testing system. Will deploy this patch there and report back whether this fixes it. :)

KatrinIhler:
Fixes template selection like expected. However, you still can't deselect a template, which, glancing at the code, I think you were also trying to achieve. But that doesn't necessarily have to be fixed prior to merging @wsmirnow ;)

wsmirnow:
I can't remember the deselection was implemented in the past. The patch works for me. Good job, thank you @Arnei

wsmirnow (approved):

PR #5539, opened 2024-01-10
Fix Paella Player 7 for single stream videos

KatrinIhler:
Currently if we open the Paella Player 7 for an event that has only one track (careful: captions count as tracks!), we get an error "s.some is not a track". This was broken by the isLive check from #5420 and happens because a single track is not wrapped in an array when the episode is returned by the search service. This fixes that by doing that first. It won't win any awards for being pretty, but it works. ;)


Arnei (approved):
Does what it says

berthob98:
Tested the patch and it fixed the error for me

miesgre:
Tested. It works. Thanks!

PR #5543, opened 2024-01-16
Use edited thumbnails in partial-publish.xml

### LINKED ISSUE #5541: Make uploading thumbnails work out of the box
I think Opencast should support uploading thumbnails out of the box, as it's a pretty important feature for a video management system in my opinion. The new editor has a UI to upload thumbnails, which is disabled by default currently. But enabling that alone is not enough: the community workflows do not support this operation yet. The attached image is ignored and never published to the search service, for example.

So I think the workflows should be adjusted to work with uploaded thumbnails. And then the editor feature can be enabled by default.

When I briefly mentioned this in community chat, one concern was: "but the community workflows can't have support for everything, otherwise they get too huge". Two thoughts about this:
- I have never worked with workflows, but I would imagine this is a fairly small change which does not increase the execution time in a meaningful way.
- All other parts of Opencast already have support for uploading thumbnails. So we already pay with complexity here. So by having workflows not support it, it seems like 90% price paid for 0% feature to me.

Arnei:
Intended to resolve #5541.

While thumbnails created in the (new) editor are stored in Opencast, they currently go unused. This changes partial-publish.xml (aka the default editor workflow) to use thumbnails created in the editor for publication.

Discussion points:
- Currently this uses the thumbnails from the editor as is. Maybe they should be encoded in some way?
- This only touches upon the thumbnails for the engage player. Other thumbnails (for search, feed cover) are ignored. Should they also be changed?

### DIFF
--- a/etc/workflows/partial-publish.xml
+++ b/etc/workflows/partial-publish.xml
@@ -136,6 +136,7 @@
 
     <operation
       id="image"
+      if="NOT ${presentation/thumbnail_edited}"
       exception-handler-workflow="partial-error"
       description="Creating player preview image for presentation video">
       <configurations>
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
Not sure if this is the right place to ask this, but I'm running an 15.3 and this doesn't seem to work? The workflows in my installation are almost the community version, with only two unrelated (very sure about that) changes. When I upload a new thumbnail in the editor, click "save" and then "process changes", the workflow runs successfully, but the two steps added in this PR are skipped. And I end up with a completely black 16px * 9px thumbnail. 

Any ideas?

PR #5545, opened 2024-01-16
Improve encoding profiles by relaxing the GOP range requirements and using CRF 22

LukasKalbertodt:
**EDIT**: this description is not quite up to date anymore. Read the next two comments as well to understand what changed. However, the file size numbers in the table under "this PR" are basically still accurate -> the huge file size wins remain.

---

This PR changes the default encoding profiles, to be less restrictive about GOP size, giving the codec more opportunity to compress better. It also changes `--crf23` to `--crf22` for a slight bump in quality. This was possible due to the large file size savings achieved by the GOP size change. This results in slightly higher quality videos with reduced file size (for slide desktop recordings reduced by more than half!).

There are some arguments for why keeping the smaller GOP size might be beneficial (I discuss those below). And of course, encoding profiles are always a balance act between many factors. But I think this particular change is a good one. I also think that increasing the quality a bit, instead of "only" taking all the file size improvements, makes sense: users' expectations regarding video quality rises over time, and the video quality will have an effect of how happy users are with Opencast.

## Encoding comparisons

I tested five very different videos:
- *Presentation*: Desktop recording of slides. Very static. Ingested in 4k lossless compression.
- *Presenter*: The presenter video of my OC Summit 2023 talk "how to install Tobira". Ingested in 1080p already compressed.
- *NASA Moon*: CGI with camera constantly moving slowly. [Link](https://tobira.opencast.org/!v/HThra7F6aZc).
- *Garden Cat*: Short video of cat in garden plants, handheld camera, lots of small moving parts. [Link](https://tobira.opencast.org/!v/ECyD9lYw4-B)
- *Spring*: Animated Blender movie, short movie, mix of slow/static and fast scenes. [Link](https://www.youtube.com/watch?v=WhWc3b3KhnY)

I would think that vast majority of videos ingested into Opencast are similar to "Presentation" or "Presenter", so the results for those videos should be weighted higher I think.

Everything was encoded in three settings:
- Status Quo: CRF 23, GOP 13-25 (*), `no-scenecut`
- This PR: CRF 22, GOP 13-100
- Another test: CRF 23, GOP 20-100

(*): Yes, the current encoding profiles say `keyint=25:min-keyint=25`, but x264 actually sets `min-keyint` to half the `keyint` in this case. This can easily be verified by `mediainfo` the resulting files which say `keyint=25 / keyint_min=13`.

### File size

| Video        | Resolution | Status Quo  | This PR  | Another test | Status Quo vs. PR |
| ------------ | ---------- | ----------- | -------- | ------------ | ----------------- |
| Presentation | 480p       | 102 MiB     | 50 MiB   | 49 MiB       | 49%               |
|              | 720p       | 159 MiB     | 76 MiB   | 74 MiB       | 48%               |
|              | 1080p      | 232 MiB     | 101 MiB  | 97 MiB       | 44%               |
|              | 1440p      | 310 MiB     | 128 MiB  | 123 MiB      | 41%               |
|              | 2160p      | 480 MiB     | 190 MiB  | 180 MiB      | 39%               |
| ---------    | ---------- | ----------- | -------- | ----------   |                   |
| Presenter    | 480p       | 84 MiB      | 71 MiB   | 58 MiB       | 86%               |
|              | 720p       | 179 MiB     | 164 MiB  | 139 MiB      | 92%               |
|              | 1080p      | 427 MiB     | 397 MiB  | 348 MiB      | 93%               |
| ---------    | ---------- | ----------- | -------- | ----------   |                   |
| NASA Moon    | 720p       | 43 MiB      | 40 MiB   | 39 MiB       | 93%               |
|              | 1080p      | 123 MiB     | 109 MiB  | 103 MiB      | 88%               |
|              | 1440p      |             | 204 MiB  | 184 MiB      |                   |
|              | 2160p      |             | 360 MiB  | 332 MiB      |                   |
| ---------    | ---------- | ----------- | -------- | ----------   |                   |
| Garden Cat   | 720p       | 3.3 MiB     | 2.9 MiB  | 2.8 MiB      | 88%               |
|              | 1080p      | 10.5 MiB    | 8.4 MiB  | 7.6 MiB      | 80%               |
|              | 1440p      |             | 16.6 MiB | 14.6 MiB     |                   |
|              | 2160p      |             | 33.7 MiB | 30.3 MiB     |                   |
| ---------    | ---------- | ----------- | -------- | ----------   |                   |
| Spring       | 720p       | 62 MiB      | 58 MiB   | 55 MiB       | 93%               |
|              | 1080p      | 134 MiB     | 123 MiB  | 111 MiB      | 92%               |

### Quality

Given that the quality is controlled by `--crf` (should be, anyway), it's no surprise that the quality with this PR's encoding profile is a touch better than with the old profile. The difference between CRF 23 and CRF 22 is certainly not large, but comparing directly, a small improvement can be seen. 

### Encoding time (very approximate!)

I did all the encodings while also normally using my computer. So these numbers here are veeeeery noisy, so keep that in mind. Looking at the encoding times, I would clearly say that the noise is clearly way larger than any potential difference. And I wouldn't worry about encoding times for this change.

| Video        | Encoding profile | Time total | Time encoding delivery |
| ------------ | ---------------- | ---------- | ---------------------- |
| Presentation | Status Quo       | 3:10:00    | 1:24:00                |
|              | This PR          | 2:34:00    | 1:26:00                |
|              | Another test     | 2:27:00    | 1:35:00                |
| Presenter    | Status Quo       | 9:14       | 5:10                   |
|              | This PR          | 10:20      | 5:55                   |
|              | Another test     | 9:19       | 5:13                   |
| Spring       | Status Quo       | 7:48       | 3:25                   |
|              | This PR          | 5:44       | 2:05                   |
|              | Another test     | 6:43       | 3:00                   |

## Arguments for a smaller GOP size

Very brief background: There are different kinds of frames in a video. For the purpose of this PR, we simplify a bit and treat them as two kinds: full images that can be decoded independently of the rest of the video (I-frame), and frames that refer to parts of a previous frame (P-frame). A GOP (group of pictures) is a group of frames that starts with an i-frame and includes all succeeding P-frames until the next I-frame. 

Having a smaller (or more regular) GOP size has some advantages:

- Faster seeking: If you want to jump to any frame in a video, it's most likely a P-frame. To decode (and show) that frame, all previous frames until the closest I-frame have to be decoded (and thus downloaded). Larger GOP sizes means that on average, more frames need to be downloaded & decoded.
- Switching quality: Very similar to seeking, when switching quality (manually or automatically), the whole GOP of the current frame needs to be download and decoded. If a player that can automatically switch quality based on network conditions, restricts itself to only switching the quality at the beginning of a GOP, there are fewer switching opportunities with larger GOP sizes.
- Low latency live streaming: usually live streaming delay is at least as high as the GOP size, as the encoder won't release frames until a GOP is finished.
- For lossy networks, loosing one frame means that all remaining frames in the GOP are corrupted. Larger GOP sizes means a longer time until the image is fixed again.

The last arguments don't apply I think: Opencast is almost never the service actually encoding live streams (as far as I know). So we can mostly think about on-demand videos. Also, Opencast videos are almost never viewed over lossy networks.

The first two arguments apply, but I think the negative effect of my PR on this will be minimal. The maximum GOP size allowed by this PR is 4s, which is not even a lot. Also note that the I-frame in a GOP takes usually the majority of space required by that GOP. So having a longer GOP does not increase the download size by the same amount.

For comparison, I analyzed the YouTube 1080p video of the Spring blender movie. GOP sizes ranged from 4 to 133 frames (for a 24fps video). The average is 77. Here is a histogram of how often certain GOP sizes appear:

image

YouTube faces the same issues regarding seeking and quality switching, and yet they have even higher GOP sizes. So I think it's certainly not unreasonable to increase Opencast's GOP size to 100.

## GOP size independent of frame rate?

The GOP size is specified in frames, which might not be optimal. Specifying as seconds might be more appropriate, to work better for videos with different frame rates. I tried to do that but I couldn't find a way to do it. The only option I found was `-force_key_frames "expr:gte(t,n_forced*N)"`, but that forces exact GOP sizes, which is the exact thing I want to move away from. 

So I decided to just move along with my change anyway. Right now we are reencoding all videos with 25fps anyway (which bothers me greatly!), so it still works. This PR doesn't make anything worse in that regard.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/encoding/opencast.properties
+++ b/etc/encoding/opencast.properties
@@ -39,35 +39,35 @@ profile.adaptive-parallel.http.suffix.2160p-quality = -2160p.mp4
 profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-1600-900 = \
   -filter:v scale=1920:1920:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
-    -c:v libx264 -profile:v high -level 4.0 -crf 23 \
-    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 4000k -bufsize 8000k \
+    -c:v libx264 -profile:v high -level 4.0 -crf 22 \
+    -x264opts keyint=100:min-keyint=13 -maxrate 4000k -bufsize 8000k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.1080p-quality}
 profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-2440-1260 = \
   -filter:v scale=2560:2560:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
-    -c:v libx264 -profile:v high -level 5.0 -crf 23 \
-    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 9800k -bufsize 19600k \
+    -c:v libx264 -profile:v high -level 5.0 -crf 22 \
+    -x264opts keyint=100:min-keyint=13 -maxrate 9800k -bufsize 19600k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.1440p-quality}
 profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-3200-1800 = \
   -filter:v scale=3840:3840:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
-    -c:v libx264 -profile:v high -level 5.1 -crf 23 \
-    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 14800k -bufsize 29600k \
+    -c:v libx264 -profile:v high -level 5.1 -crf 22 \
+    -x264opts keyint=100:min-keyint=13 -maxrate 14800k -bufsize 29600k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.2160p-quality}
 profile.adaptive-parallel.http.ffmpeg.command = -i #{in.video.path} \
   -filter:v scale=854:854:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
-    -c:v libx264 -profile:v high -level 4.0  -crf 23 \
-    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 800k -bufsize 1600k \
+    -c:v libx264 -profile:v high -level 4.0  -crf 22 \
+    -x264opts keyint=100:min-keyint=13 -maxrate 800k -bufsize 1600k \
     -c:a aac -b:a 64k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.480p-quality} \
   -filter:v scale=1280:1280:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
-    -c:v libx264 -profile:v high -level 4.0  -crf 23 \
-    -x264opts keyint=25:min-keyint=25:no-scenecut -maxrate 1200k -bufsize 2400k \
+    -c:v libx264 -profile:v high -level 4.0  -crf 22 \
+    -x264opts keyint=100:min-keyint=13 -maxrate 1200k -bufsize 2400k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.720p-quality} \
   #{if-width-or-height-geq-1600-900} \
--- a/etc/encoding/opencast.properties
+++ b/etc/encoding/opencast.properties
@@ -40,34 +40,34 @@ profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-1600-900 = 
   -filter:v scale=1920:1920:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
     -c:v libx264 -profile:v high -level 4.0 -crf 22 \
-    -x264opts keyint=100:min-keyint=13 -maxrate 4000k -bufsize 8000k \
+    -force_key_frames "expr:gte(t,n_forced*4)" -maxrate 4000k -bufsize 8000k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.1080p-quality}
 profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-2440-1260 = \
   -filter:v scale=2560:2560:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
     -c:v libx264 -profile:v high -level 5.0 -crf 22 \
-    -x264opts keyint=100:min-keyint=13 -maxrate 9800k -bufsize 19600k \
+    -force_key_frames "expr:gte(t,n_forced*4)" -maxrate 9800k -bufsize 19600k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.1440p-quality}
 profile.adaptive-parallel.http.ffmpeg.command.if-width-or-height-geq-3200-1800 = \
   -filter:v scale=3840:3840:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
     -c:v libx264 -profile:v high -level 5.1 -crf 22 \
-    -x264opts keyint=100:min-keyint=13 -maxrate 14800k -bufsize 29600k \
+    -force_key_frames "expr:gte(t,n_forced*4)" -maxrate 14800k -bufsize 29600k \
     -c:a aac -b:a 96k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.2160p-quality}
 profile.adaptive-parallel.http.ffmpeg.command = -i #{in.video.path} \
   -filter:v scale=854:854:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
     -c:v libx264 -profile:v high -level 4.0  -crf 22 \
-    -x264opts keyint=100:min-keyint=13 -maxrate 800k -bufsize 1600k \
+    -force_key_frames "expr:gte(t,n_forced*4)" -maxrate 800k -bufsize 1600k \
     -c:a aac -b:a 64k -ac 2 \
     -movflags faststart #{out.dir}/#{out.name}#{out.suffix.480p-quality} \
   -filter:v scale=1280:1280:force_original_aspect_ratio=decrease:force_divisible_by=2,fps=25 \
     -pix_fmt yuv420p \
     -c:v libx264 -profile:v high -level 4.0  -crf 22 \
-    -x264opts keyint=100:min-keyint=13 -maxrate 1200k -bufsize 2400k \
+    -force_key_frames "expr:gte(t,n_forced*4)" -maxrate 1200k -bufsize 2400k \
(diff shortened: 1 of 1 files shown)


mtneug:
First of all thanks for bringing this up and for your analysis. I agree that the current encoding parameters are not ideal.

Let me first address the YouTube example. The following at least applies to AVC/h.264. I don't know enough how it works in VP9 and AV1.

For the histogram, I assume you simply counted every I-frame. This is not quite the right approach for a streaming scenario target. Instead, you need to differentiate between IDR (sometimes also called sync or keyframes) and non-IDR I-frames. When the decoder encounters an IDR frame, it can reset the reference buffer, i.e. all following frames will not depend on frames before the IDR. This is what allows quick seeking. In an HTTP Adaptive Streaming (HAS) scenario, you want every variation to have aligned IDRs, and every segment should start with one. The simplest way to achieve this is to output IDR frames every n frames or seconds. You may also let the encoder decide the positions for the first variant and base the IDR positions of this reference (there should be an FFmpeg/x264 flag for that). However, for HAS it makes sense to have similar distances between IDRs. This can also be seen in your YouTube example. I like to use [this tool](https://lulebo.github.io/) to get a visual representation. It will also inform you about constant keyframe distances, however, YouTube has slight variations (96 - 132, i.e. 4s - 5,5s). I plotted the distribution of IDR distances below. (BTW `ffprobe` seems to count somewhat differently and reports more frames having `key_frame=1`).

IDR

The encoder can decide to output additional I-frames, e.g. if coding the entire image is more efficient than using references (as may be the case for scene changes). x264 also does this, but many disable this for HAS using `no-scenecut`. I have never understood why people do this. If bitrate is a concern, why not set rate control parameters? Anyways, you have seen that YouTube is allowing additional non-IDR I-frames.

So, what should we use in Opencast? IMO, we should finally move to HAS, which would require aligned IDRs. Further, I'm not a fan of using fixed frame distances without considering the framerate. E.g. setting `keyint` to 100 for a 60 fps video is too short. I'm not sure if the default is that _everything_ is constrained to 25 fps (which I completely agree is horrible), but that doesn't mean we can't prepare for a future where this is no longer the case. I would therefore suggest to use `-force_key_frames "expr:gte(t,n_forced*N)"` where N is maybe something between 2 and 6 (we currently use 2 in Münster).

I'm unsure if `keyint` and `min-keyint` can be left empty. I probably tested this at some point in the past, but I can't remember. These values have defaults that may mess with equal IDR placement. Currently, we set them to something high, which will never be relevant in practice, and only rely on `-force_key_frames` for IDR placement. But this should ideally be evaluated.

Then I would remove `no-scenecut` as you did. If x264 thinks an I-frame makes sense, let it place one.

Finally, to make it explicit, I would set the x264 option `no-open-gop=1` (although I think this is already the default).

LukasKalbertodt:
> For the histogram, I assume you simply counted every I-frame.

I don't think I did. I used `ffprobe -show_frames` and then looked for `key_frame=1`, which only includes IDR frames according to [this](https://video.stackexchange.com/a/19252).
Thanks for the link to this tool, that was very helpful!

---

I tested all my videos with an encoding that would fit your recommendation, namely `-force_key_frames "expr:gte(t,n_forced*4)"` and no `keyint` or `no_scenecut` at all. Also in crf22. And initially to my surprise, the resulting videos are only marginally larger than the ones produced by my suggested encoding! 

| Video        | Resolution | Status Quo  | This PR  | Matthias1 |
| ------------ | ---------- | ----------- | -------- | --------- |
| Presentation | 480p       | 102 MiB     | 50 MiB   | 50 MiB    |
|              | 720p       | 159 MiB     | 76 MiB   | 77 MiB    |
|              | 1080p      | 232 MiB     | 101 MiB  | 101 MiB   |
|              | 1440p      | 310 MiB     | 128 MiB  | 129 MiB   |
|              | 2160p      | 480 MiB     | 190 MiB  | 192 MiB   |
| ---------    | ---------- | ----------- | -------- | --------- |
| Presenter    | 480p       | 84 MiB      | 71 MiB   | 72 MiB    |
|              | 720p       | 179 MiB     | 164 MiB  | 162 MiB   |
|              | 1080p      | 427 MiB     | 397 MiB  | 398 MiB   |
| ---------    | ---------- | ----------- | -------- | --------- |
| NASA Moon    | 720p       | 43 MiB      | 40 MiB   | 40 MiB    |
|              | 1080p      | 123 MiB     | 109 MiB  | 110 MiB   |
|              | 1440p      |             | 204 MiB  | 204 MiB   |
|              | 2160p      |             | 360 MiB  | 361 MiB   |
| ---------    | ---------- | ----------- | -------- | --------- |
| Garden Cat   | 720p       | 3.3 MiB     | 2.9 MiB  | 2.9 MiB   |
|              | 1080p      | 10.5 MiB    | 8.4 MiB  | 8.5 MiB   |
|              | 1440p      |             | 16.6 MiB | 16.6 MiB  |
|              | 2160p      |             | 33.7 MiB | 33.8 MiB  |
| ---------    | ---------- | ----------- | -------- | --------- |
| Spring       | 720p       | 62 MiB      | 58 MiB   | 60 MiB    |
|              | 1080p      | 134 MiB     | 123 MiB  | 128 MiB   |

But thinking about this more, it makes sense. I tried to visualize this:

```
               0       4       8       12      16      20      24      
mine:          I       I  I       I       I  I      I       I  I        
force:         I       I  i    I       I     i I    i  I       I         
scene cuts:               |                  |      |          | 
```

The first line is the time in seconds. The last line specifies where a scene cut in the video occurs (a completely new frame appears). The middle two lines show where IDR-frames (`I`) and I-frames (`i`) are set by the two different encoding profiles. 

- The `force` profile is what I tried last (what you suggested) and is easiest to understand: it emits IDR frames exactly every 4s, starting at 0. It also emits I-frames whenever a scene cut happens.
- The `mine` profile (what this PR suggests right now) is different: it also emits IDR frames at most 4s after the last one. But it restarts counting whenever one is emitted. So after a scene cut happens around second 5, the next IDR frame is only emitted at second 9 (5+4), and not 8. 
  - (This encoding can of course also decide to emit I instead of IDR frames at scene cuts, but that's irrelevant for this visualization; also, having checked a few videos, this profile rarely emits i frames.)

So it makes sense that the `force` encoding will emit at least as many IDRs frames as the `mine` encoding. BUT and this is what I now intuitively understood: if there are few scene changes (like with a presentation or a static presenter camera), then the difference is very small! So it also makes sense that for "Spring", we can see the largest difference as is has the most scene changes.

But I noticed that all of my videos had rather few scene changes overall, so I tried these two encodings again with [this video](https://www.youtube.com/watch?v=IvbhVPLS2fM) which is has rather lot scene changes I'd say. But to my surprise the difference was still fairly low:

| Resolution | This PR  | Matthias1 |
| ---------- | -------- | --------- |
| 480p       | 6.1 MiB  | 6.3 MiB   |
| 720p       | 9.7 MiB  | 10 MiB    |
| 1080p      | 21.8 MiB | 22.4 MiB  |

I visually compared both videos with [this tool](https://vivictorg.github.io/vivict/). In some scenes I felt like my encoding preserved more details, but overall the difference was very low and mostly inconclusive. 

---

So while I'm still not 100% convinced this "guaranteed idr frames at the same place" is that important, I will update this PR to this latest encoding profile (your suggestion). 
- The increase in file size over my profile is only very small (and it's still a huge win over the status quo) *and* 
- with this we now actually guarantee synced IDR frames between resolutions *and* 
- we can be FPS independent. 

So lots of wins.

LukasKalbertodt:
Oh and for the protocol: the 5 initial test videos had sound included, while the last one (ETH trailer) did not.

Ignoring audio in the comparison means the differences are even higher. For the "presentation 2160p" case, the video track with this PR's encoding is only 35% as large as the status quo!

snoesberger:
Thanks for the detailed analysis! It also makes perfect sense to me to increase the GOP length to save storage and bandwidth. To enable adaptive HLS streaming, it is definitely important to have regular IDR frames.

One more comment not directly related to this PR: With the standard encoding profile we create different resolutions. With the Paella 7 player, however, it is no longer possible to switch between these different resolutions during progressive download video streaming. Only one of the video resolutions is loaded and played by the player. See also https://github.com/polimediaupv/paella-core/issues/342#issuecomment-1863982955 why this feature was dropped in Paella 7. 
In my opinion, it would make sense to create and offer adaptive HLS directly in the standard Opencast installation. This way, the video resolution in the Paella 7 would automatically adapt to the available network bandwidth, and the user would again be able to manually switch between the different resolutions. Otherwise, there is no point in encoding the video in different resolutions. (I did open a [new issue](https://github.com/opencast/opencast/issues/5550) to cover this)

mtneug (approved):

mtneug:
Thx @LukasKalbertodt!

PR #5549, opened 2024-01-17
Add `skip_frame nokey` to timelinepreview operation for longer videos

LukasKalbertodt:
This significantly speeds up the operation. Some examples:
- 5min video h264 1080p 190MB: 5.0s -> 1.5s
- 5min video h265 2160p 540MB: 32.5s -> 3.3s
- 50min video h264 2160p 200MB: 75s -> 14s

For short videos, the timelinepreview will be less and less useful with this: if a very short video only has 5 key frames, the generated image will only contain 5 different frames, most repeated multiple times. For that reason, I only enabled this for videos longer than 120s. But see the discussion about the threshold below.

Ideally, in the future, this operation should use the smallest encoded output video as input (e.g. the 480p stream). The preview downscales the frame a lot anyway. Doing that would
- Speed things up further, given that ffmpeg has to decode smaller frames.
- Make sure that we have somewhat regular keyframes to completely rule out fringe cases.

But that's out of scope for this PR.

---

## Higher threshold?

I tested this with three videos (comparing before and after) and maybe I need to increase the threshold. 

- Moon: 5min, 37 keyframes in input file (8s average GOP size)
- Spring: 7:44, 106 keyframes in input file (4.3s average GOP size)
- Cat: 22s, 3 keyframes in input file (7.3s average GOP size)

For "cat", the preview did not change at all (as expected, due to the threshold). But for the other videos, the preview did get a bit worse. I don't think it's terribly worse, but I would undersatnd if not everyone would be ok with this. So I'm totally fine with increasing the threshold to 10min for example. I would guess that by then it's really no problem. And we still save time for the videos where it really matters. So yeah, let me know what you think!

**Moon (with this PR)**
moon-after

**Moon (before)**
moon-before

**Spring (with this PR)**
spring-after

**Spring (before)**
spring-before

Here is a comparison for a 50min video (I did this manually with ffmpeg, not through OC. So the results could be a tiny bit different):

**Presentation (with this PR)**
out2-key

**Presentation (before)**
out2

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/timelinepreviews-ffmpeg/src/main/java/org/opencastproject/timelinepreviews/ffmpeg/TimelinePreviewsServiceImpl.java
+++ b/modules/timelinepreviews-ffmpeg/src/main/java/org/opencastproject/timelinepreviews/ffmpeg/TimelinePreviewsServiceImpl.java
@@ -397,7 +397,7 @@ public class TimelinePreviewsServiceImpl extends AbstractJobProducer implements
         // minimal and not relevant for the user. Nothing would crash without
         // this duration check: short videos would just repeat keyframes in the
         // output image, making the preview less useful.
-        "-skip_frame", duration > 120.0 ? "nokey" : "default",
+        "-skip_frame", duration > 15 * 60.0 ? "nokey" : "default",
         "-i", mediaFile.getAbsolutePath(),
         "-vf", "fps=1/" + seconds + ",scale=" + width + ":" + height + ",tile=" + tileX + "x" + tileY,
         imageFilePath
--- a/modules/timelinepreviews-ffmpeg/src/main/java/org/opencastproject/timelinepreviews/ffmpeg/TimelinePreviewsServiceImpl.java
+++ b/modules/timelinepreviews-ffmpeg/src/main/java/org/opencastproject/timelinepreviews/ffmpeg/TimelinePreviewsServiceImpl.java
@@ -392,6 +392,12 @@ public class TimelinePreviewsServiceImpl extends AbstractJobProducer implements
         binary,
         "-loglevel", "error",
         "-t", String.valueOf(duration - seconds / 2.0),
+        // For longer videos, this operation only considers keyframes. This
+        // significantly speeds up this command. The difference in output is
+        // minimal and not relevant for the user. Nothing would crash without
+        // this duration check: short videos would just repeat keyframes in the
+        // output image, making the preview less useful.
+        "-skip_frame", duration > 120.0 ? "nokey" : "default",
         "-i", mediaFile.getAbsolutePath(),
         "-vf", "fps=1/" + seconds + ",scale=" + width + ":" + height + ",tile=" + tileX + "x" + tileY,
         imageFilePath


mtneug:
I would also question the threshold. I did some tests with, admittedly, artificially created scenarios. First of all, with this change, the video needs at least two IDR frames. The following inputs fail the timelinepreview operation for me:

```sh
❯ ffmpeg \
  -t 130 -i https://public.educast.cloud/media/master/caminandes-3-llamigos.mp4 \
  -force_key_frames '0:00:00' \
  -g 10000 -keyint_min 5000 -x264opts "no-scenecut" \
  out-1-idr.mp4

❯ ffmpeg \
  -t 130 -i https://public.educast.cloud/media/master/caminandes-3-llamigos.mp4 \
  -force_key_frames '0:00:00' \
  -g 10000 -keyint_min 5000 \
  out-iframes-and-1-idr.mp4
```

While the following is fine:

```sh
❯ ffmpeg \
  -t 130 -i https://public.educast.cloud/media/master/caminandes-3-llamigos.mp4 \
  -force_key_frames '0:00:00,0:01:00' \
  -g 10000 -keyint_min 5000 -x264opts "no-scenecut" \
  out-2-idr.mp4
```

The preview images for the last example are as follows (the first frame is black):

(before)

out-2-idr-before

(after)

out-2-idr-after

Now, is this relevant? I would say sort of, as I have seen Studio screen recordings with no changes on the screen produce very very long GOPs. IIRC, the default workflows have a normalization step for Studio recordings, so by the time we arrive at the timelinepreview operation, the video probably can be processed normally. On the other hand, we have to deal with user inputs for the normal workflow and they will always find a way to upload a messy video :stuck_out_tongue_winking_eye:

The other question is the quality of the previews. The main purpose is to give the user a preview when scrubbing in the player (btw Paella currently does not use this image; it still uses the section previews). For shorter videos, it is more relevant to get a precise correlation between the timestamp of the preview and the time on the player's timeline. I would therefore suggest increasing the threshold significantly. How much is debatable, but personally 10 min feels still too short for 100 images (default). Maybe 20 min? I would bring that up when discussing this PR during the next dev meeting.

LukasKalbertodt:
Thanks for the review and your tests. I did a few tests with Studio-recorded videos and have a few observations/notes:

- I recorded a 30min video of a completely unmoving tab in Chrome. In these cases, there is a 1s distance between each P-frame. (That's new I think, I am pretty sure, at one point individual P-frames could last for many seconds). But more importantly: every roughly 100s, there is an iframe! Sure, this is just one browser in one version for one video, but it gives me hope that a high enough threshold should catch even these fringe cases.

- As you said: the Studio workflow first encodes those weird inputs into something reasonable, so no problem there. What about someone downloading from Studio and then uploading with default workflow? Well, that already fails! The waveform and timelinepreview operations fail due to "has no duration", when uploading a file from Studio. So this PR doesn't make anything worse.

- I found [`MediaRecorderOptions.videoKeyFrameIntervalDuration`](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/MediaRecorder#videokeyframeintervalduration). Which unfortunately does not seem to be implemented in any browser yet. It was added to the spec roughly a year ago. But we might want to use that in the future.

So: I just pushed a commit increasing the threshold to 15min. I think that should be fine (a) to address the concerns about the quality of the generated images and (b) to make it super unlikely that 1 IDR-frame videos break the workflow. But sure, lets talk about this tomorrow in dev meeting.

mtneug (approved):

PR #5553, opened 2024-01-24
Don't duplicate user roles when switching ACL templates in old Admin UI

KatrinIhler:
When switching templates in the ACL view of the old Admin UI of _new_ events, series or ACL templates, the user roles are accidentally pushed twice. This is not visible in the UI, except if you have custom actions, which will result in them seeming to mysteriously multiply each time you select a different template:

grafik

This was initially broken by #5269.

### DIFF
--- a/modules/admin-ui-frontend/app/scripts/shared/services/wizards/new-acl/access.js
+++ b/modules/admin-ui-frontend/app/scripts/shared/services/wizards/new-acl/access.js
@@ -102,7 +102,6 @@ angular.module('adminNg.services')
 
               UserResource.get({ username: id }).$promise.then(function (data) {
                 policy.user = data;
-                me.ud.policiesUser.push(policy);
                 // Did we get a user not present in Opencast (e.g. LDAP)? Add it to the list!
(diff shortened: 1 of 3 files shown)


Arnei:
Unable to reproduce this. UI looks fine, POST request too.

KatrinIhler:
> Unable to reproduce this. UI looks fine, POST request too.

Needs some more config to reproduce, details in private message. Sorry!

Arnei:
Managed to reproduce the issue, unfortunately this PR did not fix it for me.

Arnei:
(For anyone curious, the additional config Katrin mentioned is:
- Does not happen with Admin or System user, gotta create your own.
- In `etc/listproviders/acl.default.create.properties`, comment in `keep_on_template_switch_role_prefixes=ROLE_USER_`
- The user needs a custom action
)

Arnei (approved):
After having explained to me that I tested the wrong setup, I can confirm that this indeed fixes the issue.

KatrinIhler:
Approved and discussed -> merging

PR #5554, opened 2024-01-24
Fix ACL Template Display for Series in old Admin UI

KatrinIhler:
PR #5537 fixed the template selection, but accidentally broke the display of already selected templates because `baseAclId` was always set to `undefined`. This fixes that (at least for series, events might be broken in a different way as well).


Arnei (changes requested):
Changes make sense to me and fix the issue for series. Events/Acls still dont display the template their based on, but afaik they also did not do that beforehand.

My only request is to aim this at r/14.x

KatrinIhler:
> My only request is to aim this at r/14.x

Done!

Arnei (approved):

KatrinIhler:
Approved and discussed -> merging

PR #5556, opened 2024-01-30
Don't create composites for new editor

KatrinIhler:
The new editor can display more than one track and it's default in OC 14, so there's no need to create composites for the internal publication anymore. But since lots of people still use the old one, I simply commented out the other operations. This might be debatable. I'm also unsure if this can go into a minor release - but config isn't updated automatically, right?


KatrinIhler:
Re-pointed at 15.x after discussion in the Dev meeting. This _will_ change the config for everyone who hasn't changed this workflow file themselves, so we should document this well.

Arnei (approved):

PR #5586, opened 2024-02-02
Update axios from 0 to 1 in ltitools

### LINKED ISSUE #5304: Update or remove axios from LTI tools
Opencast's LTI tools require the axios JavaScript library in an old version and simply updating does not seem to work:
See #5288

We should take a look into updating axios or, even better, removing it and replacing it with [JavaScript's native Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).

You can run the LTI tools separately from Opencast like this:
```
❯ cd modules/lti/
❯ npm ci
❯ ./copy-translations.sh
❯ npm run start
```

Arnei:
Resolves #5304.

Updates the dependency axios in the ltitools from major version 0 to 1.


gregorydlogan:
This seems to work for me, and the selenium test pass.

PR #5587, opened 2024-02-02
Fix ltitools Upload crashing if extendedMetadata exists

Arnei:
The ltitools Upload page can crash on loading, if the series specified for the upload is attached to an event which has at least one extendedMetadata dublincore catalog attached to it (or any other dublincore catalog beside the default one really).

This fixes that, although the fix relies on hardcoding the default flavor of the event dublincore catalogs, which is technically configurable. I have not found a way to easily access whether a dublincore catalog is the common one, and don't currently feel motivated to implement one :').

### DIFF
--- a/modules/lti/src/components/EditForm.tsx
+++ b/modules/lti/src/components/EditForm.tsx
@@ -4,7 +4,8 @@ import {
     EventMetadataField,
     EventMetadataContainer,
     EventMetadataCollection,
-    collectionToPairs
+    collectionToPairs,
+    findField
 } from "../OpencastRest";
 import Select, { OnChangeValue } from "react-select";
 import CreatableSelect from "react-select/creatable";
@@ -161,9 +162,9 @@ class TranslatedEditForm extends React.Component<EditFormProps> {
     }
 
     render() {
-        const languageField = this.props.data.fields.filter((field) => field.id === "language")[0];
+        const languageField = findField("language", this.props.data);
         let languageOptions: OptionType[] = [];
-        if(languageField.collection !== undefined){
+        if(languageField && languageField.collection !== undefined){
             languageOptions = collectionToOptions(languageField.collection, languageField.translatable, this.props.t);
         }
         return <form>
--- a/modules/lti/src/components/Upload.tsx
+++ b/modules/lti/src/components/Upload.tsx
@@ -86,7 +86,8 @@ class TranslatedUpload extends React.Component<UploadProps, UploadState> {
     componentDidMount() {
         getEventMetadata(this.state.eventId).then((metadataCollection) => {
             if (metadataCollection.length > 0) {
-                const metadata = metadataCollection[0];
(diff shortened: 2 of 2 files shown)


lkiesow:
This is open for a while now. I suspect it's mostly since no one wants to spend the time configuring a test system for this. Can you provide the necessary configuration files or `sed` commands to quickly get a system configured correctly to test this? Also, a few simple steps to follow so we can reproduce this?

Arnei:
How to reproduce:

0. You do NOT need to configure LTI for this. Any simple development instance is enough.
1. Configure an extended metadata catalog for events in your Opencast installation. You can find information on how to do so in the documentation https://docs.opencast.org/develop/admin/#configuration/metadata/#extended-metadata. Alternatively. use the information in the spoiler tag at the end of this post to create a file in `/etc`. It does not matter what is specified in the extended metadata catalog config, as long as there is a config.
2. In your Opencast, create a test series.
3. In your Opencast, create a test event with the test series and with an extended metadata catalog. If you configured your extended metadata catalog correctly in step 1, this can be achieved through the "Create Event" modal in the admin ui (there should be a step after "Metadata" called "Metadata II").
4. Go to the lti upload page http://{my-opencast-url}/ltitools/index.html?subtool=upload&series={my-test-series-id} (replace variables with your values).
5. See error. Or not. Iirc this error only occurs if the common metadata catalog is not the first in a list metadata catalogs, while the order of the metadata catalogs is not guaranteed.


Extended Metadata configuration example:
<details>
  <summary>org.opencastproject.ui.metadata.CatalogUIAdapterFactory-episode-fruits.cfg</summary>

type=events
organization=*
flavor=dublincore/fruits
title=FRUITS
common-metadata=false

# Name of the XML root element of the serialized catalog
xml.rootElement.name=dublincore
# Namespace binding for the XML root element
xml.rootElement.namespace.URI=http://www.opencastproject.org/xsd/1.0/dublincore/

# XML namespace bindings
xml.namespaceBinding.root.URI=http://www.opencastproject.org/xsd/1.0/dublincore/
xml.namespaceBinding.root.prefix=
xml.namespaceBinding.dc.URI=http://purl.org/dc/elements/1.1/
xml.namespaceBinding.dc.prefix=dc
xml.namespaceBinding.dcterms.URI=http://purl.org/dc/terms/
xml.namespaceBinding.dcterms.prefix=dcterms

# Title
property.title.inputID=title
property.title.label=EVENTS.EVENTS.DETAILS.METADATA.TITLE
property.title.type=text
property.title.readOnly=false
property.title.required=false
property.title.order=0

# Unique ID for Event
property.uid.inputID=identifier
property.uid.label=EVENTS.EVENTS.DETAILS.METADATA.ID
property.uid.type=text
property.uid.readOnly=true
property.uid.required=false
property.uid.order=17

  
</details>

gregorydlogan:
Thank you for the detailed STR, that was what needed to happen here.  Works as expected, merging.

PR #5592, opened 2024-02-05
Patch for 360° Video

mgilvera:
This is a patch to enable 360°videos with paella player 7.


mgilvera:
i did the icla and i work for the Elan ev

gregorydlogan:
The tests are legitimately failing @mgilvera, will need a fix before this can move forward.

PR #5594, opened 2024-02-07
Include all metadata in Tobira harvest API for series

LukasKalbertodt:
Before, only very limited data about a series was sent to Tobira. This fixes that. For example to improve https://github.com/elan-ev/tobira/issues/1098

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
@@ -21,6 +21,7 @@
 
 package org.opencastproject.tobira.impl;
 
+import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_CREATED;
 import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_DESCRIPTION;
 import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_TITLE;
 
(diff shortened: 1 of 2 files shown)


modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
```diff
@@ -373,12 +377,28 @@ private static Jsons.Val findTimelinePreview(MediaPackage mp) {
         Jsons.p("updated", series.getModifiedDate().getTime())
       );
     } else {
+      // Created date
+      var createdDateString = series.getDublinCore().getFirst(PROPERTY_CREATED);
+      var created = Jsons.NULL;
+      try {
+        var ta = DateTimeFormatter.ISO_INSTANT.parse(createdDateString);
```
gregorydlogan:
```
2024-02-15 12:49:27,600 | WARN  | (Item:387) - Series ID-openmedia-opencast has bad created-date:                                             
java.time.format.DateTimeParseException: Text '2024-02-15T18:36Z' could not be parsed at index 1
```

I hate date/time classes :)

Edit: This event+series is generated via the ingest.py script [here](https://github.com/opencast/helper-scripts/tree/master/ingest)
~in https://github.com/opencast/opencast-project-infrastructure/tree/master/ansible-demo-machines/roles/opencast/templates, which I just realized should *really* be copied into the helper script repo...~

gregorydlogan (changes requested):

modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
geichelberger: I think `DateTimeFormatter.ISO_INSTANT.parse()` can be replaced with `Instant.parse()` and then it should work.

LukasKalbertodt: Uhg yeah, I hate them as well. I pushed a commit that should probably fix it. Using the same helper functions as other parts of Opencast.

PR #5597, opened 2024-02-08
Replace Solr Search with OpenSearch

gregorydlogan:
This PR removes Solr (and all of its issues) and replaces it with Opensearch/Elasticsearch (yay, new issues!), unifying our indexing tools.

To accomplish this:
- We had to [remove the Feeds](a594019f82916a55d203efe3e518224bc8b50dab)
- We [replaced Solr with Opensearch](a594019f82916a55d203efe3e518224bc8b50dab)
  - Notably, I did not include URL signing support with this.  If this is a deal breaker we can work on it later.
- We [updated Tobira's integration](dedb54c)
- We [updated out Paella glue](25e76ac)
- We [fixed other integrations with Search](c32aeb6)

I've hit a few questions as I've reviewed this internally a bunch, so I'm going to copy those questions up here as an initial code review.

If you have a real system with lots of workflows and series, and want to give this a spin, know that this does not need anything but the database, and the dublincore metadata files.  You should be able to pull those out of production deployments and then test this in entirely separate systems.

Todo list:
~- Release notes note indicating that a reindex of the search service is required.~

Known, unfixable issues:
- Paella tests rely on `develop.opencast.org` to serve the `search/*` endpoints in the format expected by this PR.  This obviously will not happen until *after* this PR is merged.  Paella tests can be verified locally by setting up Opencast to run on port 80, then `npm run dev` and `npx playwright test` after the npm script finishes booting.

Closes #3921

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -862,36 +336,9 @@ protected UserDirectoryService getUserDirectoryService() {
...
-      if (solrRequester != null) {
-        solrRequester.setMediaPackageSerializer(null);
-      }
-    }
-  }
-
   @Override
-  public void updated(@SuppressWarnings("rawtypes") Dictionary properties) throws ConfigurationException {
+  public void updated(Dictionary properties) {
+    // TODO: Move to modified()
```
gregorydlogan: Derp, missed this.  Will move it with the first round of review comments.

modules/engage-paella-player-7/src/plugins/org.opencast.paella.episodesFromSeries.js
```diff
@@ -76,7 +76,7 @@ export default class EpisodesFromSeriesPlugin extends PopUpButtonPlugin {
 
     const thisId = this.player.videoId;
     if (this._episodesData) {
-      const result = this._episodesData['search-results'].result;
+      const result = this._episodesData['result'];
       (Array.isArray(result) ? result : [result]).forEach(({id,dcTitle,mediapackage}) => {
```
gregorydlogan: The dcTitle field is no longer present, moved to dc.title[0], but I can't find where this code is called at all.  If this is dead, can we remove it?

modules/conductor/pom.xml
```diff
@@ -89,6 +89,10 @@
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
     </dependency>
+    <dependency>
```
gregorydlogan, line 88: Not sure why this is suddenly required, but Maven says it's an undeclared usage if we don't have it.

modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/OpencastDctermsDublinCore.java
```diff
@@ -574,13 +574,13 @@ protected Opt<DublinCoreValue> getFirstVal(EName property) {
     return Opt.nul(dc.getFirstVal(property));
   }
 
-  protected void set(EName property, String value) {
+  public void set(EName property, String value) {
```
gregorydlogan, line 577: These were exposed for use in `SearchResult`'s `rehydrateDC` method.

modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
```diff
@@ -626,7 +626,7 @@ protected String getSubIndexIdentifier(String type) {
     return this.indexIdentifier + "_" + type;
   }
 
-  protected RestHighLevelClient getClient() {
+  public RestHighLevelClient getClient() {
```
gregorydlogan, line 629: `SearchServiceIndex` extends `AbstractIndexProducer`, but not `AbstractElasticsearchIndex`, so this needs to be public. `AIP` doesn't extend anything (merely implements), but I didn't notice the two Index related classes were unrelated until final notes were being made. Could investigate having `AIP` extend `AEI`, and then this could remain protected, but that's future work imo.

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchService.java
```diff
@@ -49,13 +59,15 @@ public interface SearchService {
...
    * @throws UnauthorizedException
-   *           if the current user is not authorized to add this mediapackage to the search index
+   *           if the current user is not authorized to add this media package to the search index
    * @throws ServiceRegistryException
    *           if the job for adding the mediapackage can not be created
    */
   Job add(MediaPackage mediaPackage) throws SearchException, MediaPackageException, UnauthorizedException,
           ServiceRegistryException;
 
+  Collection<Pair<Organization, MediaPackage>> getSeries(String seriesId);
```
gregorydlogan, line 78: Three new need javadocs

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
```diff
@@ -79,18 +82,17 @@ public interface SearchServiceDatabase {
...
-   * @param seriesId
-   *          the series identifier
+   * @param seriesId the series identifier
    * @return collection of media packages
-   * @throws SearchServiceDatabaseException
-   *           if there is a problem communicating with the underlying data store
+   * @throws SearchServiceDatabaseException if there is a problem communicating with the underlying data store
    */
-  Collection<MediaPackage> getMediaPackages(String seriesId) throws SearchServiceDatabaseException;
+  Collection<Pair<Organization, MediaPackage>> getSeries(String seriesId) throws SearchServiceDatabaseException;
```
gregorydlogan, line 100: I *strongly* dislike this, but `SearchUpdatedEventHandler` needs to know from which `Organization` a series comes from. Figuring that out some other way (`SeriesService`? Nothing obvious, to me at least) would be better. This can be refactored after merging if needed since *nothing else* needs to know.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
```diff
@@ -111,18 +79,13 @@
         "service.pid=org.opencastproject.search.impl.SearchServiceImpl"
     }
 )
+//FIXME: Should this be a ManagedService?
```
gregorydlogan: Yeah, probably not :)

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
gregorydlogan, line 1: ODO: There's no way to view things marked as deleted, and they hang around forever until a reindex (where they would get recreated...). Should we just skip reindexing deleted things? Should we expose a way to (with appropriate roles) query *with* deleted things?

NB: Tobira depends on storing delete flags, presumably. We can't just immediately delete things.

gregorydlogan (commented):
Initial review with my internal comments

gregorydlogan:
Aware of the broken Paella 7 tests, didn't catch that there were additional tests not run by `mvn`. Will address (hopefully) tomorrow, but should not block other testing.

modules/engage-paella-player-7/src/plugins/org.opencast.paella.episodesFromSeries.js
miesgre: If you play a video that is in a series with more videos, the plugin is visible at the top-right, and shows the user a list of the videos in the same series.
Example: https://develop.opencast.org/paella7/ui/watch.html?id=ID-westerberg

Captura de pantalla 2024-02-12 a las 19 44 11

gregorydlogan: Ah, thank you @miesgre.  This should now be working, although I just can't seem to get the tests working.  When I test locally, I end up with results from the search endpoints with `localhost:8080` in them, but then the webpack test proxy fails to load the media due to CORS rules.  Any hints on setting up a proper test environment?

miesgre: @gregorydlogan Could you change the `testenv:dev` package.json script to

```
"testenv:dev": "webpack serve --mode development --host=0.0.0.0 --env server=http://localhost:8080",
```

I can't test it now, but I think it should work.
You will need a local opencast server running in 8080 and ingest the `ID-dual-stream-demo` mediapackage.

gregorydlogan: That didn't seem to work, but setting up Opencast to work on :80, then running `npm run dev` and `npx playwright test` after `npm` booted up seemed to work.  Tests pass at least :)

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
```diff
@@ -19,68 +19,263 @@
...
+      if (SearchService.IndexEntryType.Episode.equals(type)) {
+        mp = MediaPackageParser.getFromXml((String) data.get(MEDIAPACKAGE_XML));
+      }
+      return new SearchResult(type, dc, acl, org, mp, modified, deleted);
+    } catch (MediaPackageException e) {
+      throw new SearchException(e);
+    }
+  }
+
+  public static Map<String, List<String>> dehydrateDC(DublinCoreCatalog dublinCoreCatalog, Instant now) {
```
Arnei: The `Instant` is going unused.

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
```diff
@@ -19,68 +19,263 @@
...
+    return result;
+  }
+
+  public static List<Map<String, ?>> dehydrateAclForREST(AccessControlList acl) {
+    return acl.getEntries().stream()
+        .map(ace -> Map.of("action", ace.getAction(), "role", ace.getRole(), "allow", Boolean.TRUE))
+        .collect(Collectors.toList());
+  }
+
+  @SuppressWarnings("unchecked")
```
Arnei, line 176: Are these `SuppressWarnings` still required? Does not seem like it to me, but if they are maybe there should be a comment on why this is totally fine (probably because we have to expect data from ES to be that way?)?

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
```diff
@@ -0,0 +1,475 @@
...
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+
+/**
+ * A Solr-based {@link SearchService} implementation.
```
Arnei: There are couple comments here mentioning Solr :D

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -211,10 +211,11 @@ public int countMediaPackages() throws SearchServiceDatabaseException {
   /**
    * {@inheritDoc}
    *
-   * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#getAllMediaPackages()
+   * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#getAllMediaPackages(int, int)
    */
   @Override
-  public Stream<Tuple<MediaPackage, String>> getAllMediaPackages() throws SearchServiceDatabaseException {
+  public Stream<Tuple<MediaPackage, String>> getAllMediaPackages(int pagesize, int offset)
```
Arnei, line 217: `pagesize` and `offset` are going unused.

Arnei (changes requested):
Testing this with a local develop installation and it seems to work, yay! Couple comments and change requests below, but nothing major.

This breaks the annotation tool, but the annotation tool can be adapted fairly easily to work with this https://github.com/opencast/annotation-tool/pull/623.

When deleting an event or series, they remain in the search index which is afaik intended. However, the `GET /search/episode.json` does not return deleted events, while the `GET /search/series.json` does return deleted series. This may be related to the `deleted` attribute of a series in the search index never being set (not when the series is deleted, nor when the event it was attached to is deleted). Seems inconsistent to me, is this intended?

Another thing I noticed is that the new search index in ElasticSearch is created with a replica, unlike our other indices. This is not really a problem, but it indicates that the search index is created with a different set of defaults than the other indices, likely because `SearchServiceIndex.java` defines it's own `createIndex()` method, while the other indicies over in `ElasticSearchIndex.java`. Which, again, is not really a problem, but I also don't see a reason why the Search Service Index needs it's own special create method, so for the sake of consistency it would be nice to unify behaviour.

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
gregorydlogan: It doesn't have to be there, but the unchecked casts further down throw warnings otherwise.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
gregorydlogan: Check my logic here, I think I ~did~fixed this right?

gregorydlogan:
> When deleting an event or series, they remain in the search index which is afaik intended. However, the `GET /search/episode.json` does not return deleted events, while the `GET /search/series.json` does return deleted series. This may be related to the `deleted` attribute of a series in the search index never being set (not when the series is deleted, nor when the event it was attached to is deleted). Seems inconsistent to me, is this intended?

This should be fixed.  Or at least it looks like it's working for me?

Arnei:
> > When deleting an event or series, they remain in the search index which is afaik intended. However, the `GET /search/episode.json` does not return deleted events, while the `GET /search/series.json` does return deleted series. This may be related to the `deleted` attribute of a series in the search index never being set (not when the series is deleted, nor when the event it was attached to is deleted). Seems inconsistent to me, is this intended?
> 
> This should be fixed. Or at least it looks like it's working for me?

Not working for me. The `deleted` property of a deleted series is still `null`. Here's some Screenshots of my local Opencast after deleting a series and all attached events:
Bildschirmfoto vom 2024-03-20 10-10-42
Bildschirmfoto vom 2024-03-20 10-11-00
Bildschirmfoto vom 2024-03-20 10-11-23

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
Arnei: Makes sense to me. Kinda surprised we did not already have a function like that in `common`, but better late than never :D

gregorydlogan:
Gah.  It didn't occur to me to test deleting the series without removing the recordings - I always went the other way 'round.  I'm surprised this is allowed.  I can reproduce this one, thankfully?

This may take a day or two to chase down.  It *looks* like it should be working via a goat trail through at least 5 different classes, but clearly it's not.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -293,54 +128,29 @@ public Response removeSeries(@PathParam("seriesid") String seriesId) throws Sear
...
           @RestParameter(
               name = "sort",
               isRequired = false,
               type = RestParameter.Type.STRING,
-              description = "The sort order.  May include any of the following: "
-                  + "DATE_CREATED, DATE_MODIFIED, TITLE, SERIES_ID, MEDIA_PACKAGE_ID, CREATOR, "
-                  + "CONTRIBUTOR, LANGUAGE, LICENSE, SUBJECT, DESCRIPTION, PUBLISHER. "
-                  + "Add '_DESC' to reverse the sort order (e.g. TITLE_DESC)."
+              description = "The sort order.  May include any of the following dublin core metadata: "
+              + "title, contributor, creator, modified. "
```
rute-santos: Can we also add sort by series id?

lkiesow:
> Gah. It didn't occur to me to test deleting the series without removing the recordings - I always went the other way 'round. I'm surprised this is allowed. I can reproduce this one, thankfully?
> 
> This may take a day or two to chase down. It _looks_ like it should be working via a goat trail through at least 5 different classes, but clearly it's not.

If I remember correctly, the search service publishes series when an event with a series is being published and uses information from that event only. This means that deleting a series from the series service should have no effect unless the events are updated and republished.

That may explain the described behavior?

gregorydlogan:
That's exactly why it was happening.  Well, that and the event handlers weren't hooked up quite right so there was no way to pass the message through :)

Removing a series without the events is unfortunately expensive, but since the rest of the system supports it we have to support it here.

lkiesow:
`Search` is missing from the REST docs:

> `POST /rebuild/{service}`
> `service`: The service to recreate index from. The available services are: Themes, Series, Scheduler, AssetManager, Comments and Workflow. The service order (see above) is very important! Make sure, you do not run index rebuild for more than one service at a time!

It should be one of the options.

lkiesow:
The re-index doesn't work:

```
2024-04-23 22:23:11,145 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 44/2000 finished, 2% complete.
2024-04-23 22:23:11,412 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 45/2000 finished, 2% complete.
2024-04-23 22:23:11,670 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 46/2000 finished, 2% complete.
2024-04-23 22:23:11,928 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 47/2000 finished, 2% complete.
2024-04-23 22:23:12,152 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 48/2000 finished, 2% complete.
2024-04-23 22:23:12,386 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 49/2000 finished, 2% complete.
2024-04-23 22:23:12,511 | INFO  | (IndexRebuildService:214) - Finished to rebuild the Elasticsearch index from service 'Search'
```

It shouldn't stop at 2%

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
```diff
@@ -0,0 +1,481 @@
...
+            logIndexRebuildProgress(logger, esIndex.getIndexName(), total, current.getAndIncrement(), pageSize);
+            indexMediaPackage(mediaPackage, acl, modificationDate, deletionDate);
+          } catch (SearchServiceDatabaseException | UnauthorizedException | NotFoundException e) {
+            logIndexRebuildError(logger, "search", total, current.get(), e);
+            //NB: Runtime exception thrown to escape the functional interfacing
+            throw new RuntimeException("Internal Index Rebuild Failure", e);
+          }
+        });
+        offset += pageSize;
+      } while (page.size() == pageSize);
```
lkiesow: This should probably be:
```suggestion
      } while (offset <= total);
```

lkiesow:
Sorry, @gregorydlogan, but this is still broken. Luckily, I think that I know why.
But first, running the newly patched version will still result in:

```
2024-04-24 21:08:04,257 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 47/2000 finished, 2% complete.
2024-04-24 21:08:04,498 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 48/2000 finished, 2% complete.
2024-04-24 21:08:04,759 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 49/2000 finished, 2% complete.
2024-04-24 21:08:04,894 | INFO  | (IndexRebuildService:214) - Finished to rebuild the Elasticsearch index from service 'Search'
```

After adding a bit of logging, you can see why:

```diff
         offset += pageSize;
+        logger.info("offset: {}, total: {}", offset, total);
       } while (offset <= total);
```

```
2024-04-24 21:17:56,220 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 47/2000 finished, 2% complete.                                                                                        
2024-04-24 21:17:56,481 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 48/2000 finished, 2% complete.                                                                                        
2024-04-24 21:17:56,745 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 49/2000 finished, 2% complete.                                                                                        
2024-04-24 21:17:56,899 | INFO  | (SearchServiceIndex:432) - offset: 50, total: 2000                                                                                                                                                          
2024-04-24 21:17:56,901 | INFO  | (SearchServiceIndex:432) - offset: 100, total: 2000                                                                                                                                                         
2024-04-24 21:17:56,901 | INFO  | (SearchServiceIndex:432) - offset: 150, total: 2000  
```

So, we don't get any new data after the first round.
Looking at `getAllMediaPackages`, it becomes immediately clear why:

```java
  public Stream<Tuple<MediaPackage, String>> getAllMediaPackages(int pagesize, int offset)                               
          throws SearchServiceDatabaseException {                                                                        
    List<SearchEntity> searchEntities;                                                                                   
    try {                                                                                                                
      int firstResult = pagesize * offset;                                                                               
      searchEntities = db.exec(namedQuery.findSome("Search.findAll", firstResult, pagesize, SearchEntity.class));
```

The second parameter of this method is called `offset` and that's what we pass to this function. However, this method passes `pagesize * offset` to the database instead. That means that the parameter should be called `int page` and we should pass the page to this method starting with `0` and increasing by `1` every time we want the next page. Alternatively, removing the `int firstResult = pagesize * offset;` should probably also fix the issue.

---

This patch should be fully functional (tried the revuild using this=:

```diff
diff --git a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
index 8ed05ae2a4..987c6ce325 100644
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
@@ -400,13 +400,13 @@ public final class SearchServiceIndex extends AbstractIndexProducer implements I
     try {
       int total = persistence.countMediaPackages();
       int pageSize = 50;
-      int offset = 0;
+      int pageOffset = 0;
       AtomicInteger current = new AtomicInteger(0);
       logIndexRebuildBegin(logger, esIndex.getIndexName(), total, "search");
       List<Tuple<MediaPackage, String>> page = null;
 
       do {
-        page = persistence.getAllMediaPackages(pageSize, offset).collect(Collectors.toList());
+        page = persistence.getAllMediaPackages(pageSize, pageOffset).collect(Collectors.toList());
         page.forEach(tuple -> {
           try {
             MediaPackage mediaPackage = tuple.getA();
@@ -428,8 +428,8 @@ public final class SearchServiceIndex extends AbstractIndexProducer implements I
             throw new RuntimeException("Internal Index Rebuild Failure", e);
           }
         });
-        offset += pageSize;
-      } while (offset <= total);
+        pageOffset += 1;
+      } while (pageOffset * pageSize <= total);
       //NB: Catching RuntimeException since it can be thrown inside the functional forEach here
     } catch (SearchServiceDatabaseException | RuntimeException e) {
       logIndexRebuildError(logger, "search", e);
```

You should be able to  just `git apply` this.

lkiesow:
The log message for an index rebuild update you get after an event has been processed. But the first event says that 0 events have been processed and the whole thing ends at 99%:

```
142371 2024-04-24T21:48:01,944 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 0/2000 finished, 0% complete.
...
144409 2024-04-24T21:56:14,074 | INFO  | (AbstractIndexProducer:145) - Updating the Elasticsearch index from service 'Search': 1999/2000 finished, 99% complete.
```

This is easily fixed by setting the initial value of `current` to `1`:

```diff
-      AtomicInteger current = new AtomicInteger(0);
+      AtomicInteger current = new AtomicInteger(1);
```

lkiesow:
You can use OpenSearch or Elasticsearch, but the logging still says `Updating the Elasticsearch index from service 'Search'`. Since we now only have one type of search index, maybe we can just change that to `Updating the search index from service 'Search'`?

lkiesow:
Summary of what is still missing here:

- [x] [REST docs are still incorrect](https://github.com/opencast/opencast/pull/5597#issuecomment-2073414435)
- [x] [Re-indexing is still broken](https://github.com/opencast/opencast/pull/5597#issuecomment-2075790533)
- [x] [Counting the updated items is off by one](https://github.com/opencast/opencast/pull/5597#issuecomment-2075796164)
- [x] [Don't use the name Elasticsearch any longer](https://github.com/opencast/opencast/pull/5597#issuecomment-2075818985)

gregorydlogan:
Patches applied, still building the dataset to properly test a reindex at scale.  You'll probably get to this first, so let me know!

lkiesow:
I just realized that `POST /index/rebuild` rebuilds the search service index, but `POST /index/clear` does not clear the search service index. I think there needs to be a clear distinction between the two different indexes in all those endpoints. We may also need a `POST /index/clear/{indexType}`

modules/elasticsearch-index/src/main/java/org/opencastproject/elasticsearch/index/endpoint/IndexEndpoint.java
```diff
@@ -133,7 +133,7 @@ public Response clearIndex() {
       description = "Repopulates the Index from an specific service",
       returnDescription = "OK if repopulation has started", pathParameters = {
       @RestParameter(name = "service", isRequired = true, description = "The service to recreate index from. "
-        + "The available services are: Themes, Series, Scheduler, AssetManager, Comments and Workflow. "
+        + "The available services are: Themes, Series, Scheduler, AssetManager, Comments, Search and Workflow. "
         + "The service order (see above) is very important! Make sure, you do not run index rebuild for more than one "
```
lkiesow: The order is actually “Comments, Workflow and Search”. Also, Search is separate from the other index. AssetManager, Scheduler, Workflows, … are services which actually write to one index. That is why the order is important. Search is completely separate.

```suggestion
        + "The available services are: Themes, Series, Scheduler, AssetManager, Comments, Workflow and Search. "
        + "The service order (see above) is very important except for Search! Make sure, you do not run index rebuild for more than one "
```

lkiesow:
The docs for `POST /resume/{service}` is still missing “Search”:

> The service to start recreating the index from. The available services are: Themes, Series, Scheduler, AssetManager, Comments and Workflow. All services that come after the specified service in the order above will also run. 

“Search” should be after “Workflow” in the order.

lkiesow:
If an event has a `isPartOf` metadata, but the linked series does not exist, the rebuild process for this event fails. This is probably not the expected behavior. If the series does not exist, I would expect the series to not be linked, but I would still expect a successful rebuild since the publication worked after all.

```
2024-05-06 14:40:32,674 | INFO  | (AbstractIndexProducer:44) - Starting Search index rebuild with 301 search                                                                                                                                  
2024-05-06 14:40:33,117 | WARN  | (SearchServiceIndex:275) - Could not get series DublinCoreValue(e12d5290-0825-4868-a8e6-1d75d4af9d41,__,None) from series service. Skipping its publication
```

To test this, just ingest a new event with a non-existing series identifier in `isPartOf`:

```
   curl -u admin:opencast http://localhost:8080/ingest/addMediaPackage/fast \
      -F title="I 🖤 Opencast" \
      -F acl='{"acl": {"ace": [{"allow": true,"role": "ROLE_USER","action": "read"},{"allow": true,"role": "ROLE_USER","action": "dance"}]}}' \
      -F creator="Opencats" \
      -F isPartOf="e12d5290-0825-4868-a8e6-1d75d4af9d41"
```

lkiesow:
Building this now fails with:

```
[ERROR] /home/lars/dev/opencast/opencast/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Harvest.java:44:8: Unused import - java.util.Objects. [UnusedImports]
```

wsmirnow:
I've tested this PR with extended metadata for series and episode. It works fine. But…

The LTI series tool is broken. Loading the page stuck with the `loading…` message. The search query works as expected but the frontend needs to be updated for the new structure. Maybe this work can be done separately but needs to be fixed before OC 16.0 release.

lkiesow:
@wsmirnow, how did you build this? Just removing the unused import from above results for me in:
```
[ERROR] /home/lars/dev/opencast/opencast/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java:[160,91] incompatible types: org.opencastproject.security.api.AccessControlList cannot be converted to java.util.List<org.open
castproject.security.api.AccessControlEntry>
```

wsmirnow:
> @wsmirnow, how did you build this? Just removing the unused import from above results for me in:
> 
> ```
> [ERROR] /home/lars/dev/opencast/opencast/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java:[160,91] incompatible types: org.opencastproject.security.api.AccessControlList cannot be converted to java.util.List<org.open
> castproject.security.api.AccessControlEntry>
> ```

I've pulled Gregs branch into develop branch (plus rebased ETH patches) yesterday at 8:35PM (german time). The last patch from today is missing.

gregorydlogan:
> I just realized that `POST /index/rebuild` rebuilds the search service index, but `POST /index/clear` does not clear the search service index. I think there needs to be a clear distinction between the two different indexes in all those endpoints. We may also need a `POST /index/clear/{indexType}`

Ugh, missed this.  Looked into the code.  The issue here is that the `SearchServiceIndex`  class created in this PR creates its index in different way, which is why it (and it alone) does not get cleared.  Hooking that up is going to require a bunch of extra code, and would delay things further.  It also might be a silly thing to do, especially in light of several institutions showing interest in *removing* the Index Service entirely, breaking things up into smaller and less-duplicative pieces.  IMO this is something that should be tracked as an issue, and either fixed eventually or disappeared when we refactor the Index itself.  I'm also not entirely sure how often one would want to clear their index without immediately rebuilding it, but whatever :)

PR #5598, opened 2024-02-12
Change default captions flavor for Editor Service

KatrinIhler:
For more details, see this discussion: https://github.com/opencast/opencast-editor/issues/1189


Arnei:
Might make sense to point this at `r/15.x`, since this config fix is related to the subtitle rework we did with 15.0?

KatrinIhler:
Maybe, but then I might have to add release notes, and I was too lazy to do that :D

KatrinIhler:
Need to think about what this means for editing captions.

KatrinIhler:
Obsolete with #5862

PR #5605, opened 2024-02-15
Remove Paella 6 with OC 16

gregorydlogan:
This PR removes Paella 6 from Opencast, per the proposal on the users list

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


.github/workflows/auto-update-npm-libraries.yml
```diff
@@ -35,15 +35,6 @@ jobs:
           npm run eslint
           git add package.json package-lock.json
 
-      - name: update paella-player test libraries
```
gregorydlogan, line 38: There's no equivalent stanza here for paella 7.  Should there be?

Arnei (changes requested):
While this does remove Paella6, it does not remove Paella6 from the docs :D We should probably do that so as to avoid confusion among our adopters.

Arnei:
Also you did not catch all entries in the ´feature.xml` I think?

Arnei:
Then there is also the config option to set either paella6 or 7, which is now obsolete? Even if we want to keep it around for the feature, we should not mention paella6 as a valid option.

https://github.com/opencast/opencast/blob/5ddc5ad23d80d86bf9f6ca1764318e19f03b17b2/etc/org.opencastproject.organization-mh_default_org.cfg#L185

Arnei:
These rules can go as well, in theory. Maybe some adopters are doing redirect shenanigans that are relying on these rules.

https://github.com/opencast/opencast/blob/5ddc5ad23d80d86bf9f6ca1764318e19f03b17b2/etc/security/mh_default_org.xml#L290

Arnei:
Otherwise I did some minimal testing to confirm that Paella7 still works (so we don't suddenly have zero working players).

gregorydlogan:
Think I got it all.  The only thing I didn't see was more references in the feature.xml file - that's just paella 7 as far as I can tell.

.github/workflows/auto-update-npm-libraries.yml
lkiesow: There could be, but Dependabot will pick that up.
Would reduce the Dependabot pull requests though.

PR #5609, opened 2024-02-16
Remove Entwine from Asset Manager API module

Arnei:
~Attempt at finally getting rid of entwine dependencies in the asset manager. WIP, I'll see how far I can take this.~

Removes all entwine imports from the `asset-manager-api` module. (Seems like I'm not getting to `asset-manager-impl` anytime soon). Entwine functionality is replaced with standard java library functionality where applicable.

The goal of this PR is to get rid of our entwine dependencies in the asset manager. While we usually try to get rid of entwine in passing and try to avoid adding it anywhere if we can avoid it, entwine dependencies are nested deep in the asset manager and we won't get rid of them unless we target them directly (or just rewrite the whole thing). 

The goal of this PR is not to greatly improve upon code structure or readability. In fact, some parts are likely to look just as complicated as before. Still, I hope that overall this will make work with the asset manager easier in the future. And naturally, while the code is not necessarily improving, it should also not be getting worse than before, so please do complain if you think that is the case :)


modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/Value.java
```diff
@@ -118,16 +116,39 @@ public final ValueType<?> getType() {
 
   /**
    * Decompose (or pattern match) the value instance. Provide a function to handle each possible type.
-   * Use {@link #doNotMatch()} as a placeholder that yields an {@link Prelude#unexhaustiveMatch()} error.
+   * Use {@link #doNotMatch()} as a placeholder that yields an error.
    */
+//  public final <A> A decompose(
```
gregorydlogan: This should probably go, no?

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/Value.java
```diff
@@ -138,22 +159,30 @@ public final <A> A decompose(
...
   /**
-   * Use as a placeholder that yields an {@link Prelude#unexhaustiveMatch()} error in
+   * Use as a placeholder that yields an error in
    * value decomposition.
    *
    * @see #decompose(Fn, Fn, Fn, Fn, Fn)
    */
-  public static <B> Fn<Object, B> doNotMatch() {
-    return new Fn<Object, B>() {
+//  public static <B> Fn<Object, B> doNotMatch() {
```
gregorydlogan: Likewise here

gregorydlogan (changes requested):
Very minor leftover comment stuff, but otherwise looks good to me!

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/fn/Product.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.opencastproject.assetmanager.api.fn;
+
+public final class Product<A> {
```
lkiesow, line 23: What does this do and why do we need this? This is just a generic wrapper around one object. Why do we need to wrap something in here?

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/RichAResult.java
```diff
@@ -40,36 +42,36 @@ public RichAResult(AResult result) {
...
-  public Stream<Property> getProperties() {
-    return result.getRecords().bind(ARecords.getProperties);
+  public List<Property> getProperties() {
+    return ARecords.getProperties(result.getRecords());
   }
 
   /** Count all properties contained in the result. */
   public int countProperties() {
-    return sizeOf(getProperties());
+    return new HashSet(getProperties()).size();
```
lkiesow: Since `getProperties` now returns a `List`, can't you just do:
```suggestion
    return getProperties().size();
```

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/query/RichAResult.java
```diff
@@ -40,36 +42,36 @@ public RichAResult(AResult result) {
...
+        .map(record -> record.getSnapshot())
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toList());
   }
 
   /** Count all snapshots contained in the result. */
   public int countSnapshots() {
-    return sizeOf(getSnapshots());
+    return new HashSet(getSnapshots()).size();
```
lkiesow: Same here?
```suggestion
    return getSnapshots().size();
```

modules/asset-manager-api/src/main/java/org/opencastproject/assetmanager/api/fn/Product.java
Arnei: I don't really know what this does, which is exactly why we need this unfortunately :D.

This is a class from the entwine library, stripped down to the bare minimum we actually use. I could not figure out how to replace this with "proper" Java code. To see how it is used take a look at `assetmanager/api/Value.java`.

PR #5610, opened 2024-02-19
Update config file and docs for correct shibboleth logout

pselsa:
Since `<sec:logout logout-success-url="/Shibboleth.sso/Logout?return=www.opencast.org" />` has no effect anymore, the config file and the docs should be updated to the correct way of redirecting after shibboleth logout.


pselsa:
i have filed the ICLA now

PR #5623, opened 2024-02-27
Fix use of s3 distribution service in live scheduler

rute-santos:
Fix for: if the download distribution service comes up before the s3 distribution service, it will be used even if the 'aws-s3' distribution service was set in the 'live.distributionService' configuration.
This fix has been in our code base for a long time but it seems we forgot to create a PR...

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
+++ b/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
@@ -169,13 +169,17 @@ public class LiveScheduleServiceImpl implements LiveScheduleService {
   private String streamMimeType;
   private String[] streamResolution;
   private MediaPackageElementFlavor[] liveFlavors;
-  private String distributionServiceType = DEFAULT_LIVE_DISTRIBUTION_SERVICE;
+  private String distributionServiceType;
   private String serverUrl;
   private Cache<String, Version> snapshotVersionCache
       = CacheBuilder.newBuilder().expireAfterWrite(5, TimeUnit.MINUTES).build();
   /** Which streaming formats should be published automatically */
   private List<String> publishedStreamingFormats = null;
   private String systemUserName;
+
+  /** References to all distribution services (used while activate not called yet) */
+  private HashMap<String, DownloadDistributionService> tempDistributionServices =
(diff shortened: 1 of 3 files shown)


modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -257,6 +261,12 @@ protected void activate(ComponentContext context) {
 
     if (!StringUtils.isBlank((String) properties.get(LIVE_DISTRIBUTION_SERVICE))) {
       distributionServiceType = StringUtils.trimToEmpty((String) properties.get(LIVE_DISTRIBUTION_SERVICE));
+    } else {
+      distributionServiceType = DEFAULT_LIVE_DISTRIBUTION_SERVICE;
+    }
+    if (tempDistributionServices.get(distributionServiceType) != null) {
```
gregorydlogan: The issue with this approach is that in a misconfiguration scenario (ie, I say I want a dist service which does not exist on this node), you'll have a Karaf service which *looks* like it's working but will NPE when you go to use it - since nothing has set `downloadDistributionService`.

Take a look at `etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg` and `modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/SpeechToTextServiceImpl.java`, which do *exactly* what you're trying to do with via `SpeechToTextEngine.target`, which set the OSGi DS filter at boot ;)

gregorydlogan (changes requested):

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
rute-santos: @gregorydlogan , the idea is great, but I don't think that example is working. It's true that there's a configuration, but the code does not use it and the reference in that example is static so when the component is up the reference is already set. Do you have another example that works?

rute-santos: Hi @gregorydlogan , thanks for the suggestion! I think it's much better now!

PR #5673, opened 2024-03-14
Paella7: Avoid opening downloaded video

Arnei:
This should help with some cases where clicking on a video in the Paella7 download plugin would open the video instead of triggering a "Do you want to download this?" dialog from the browser.


miesgre (approved):
seems good

PR #5674, opened 2024-03-14
Feature request: privacy statement and imprint

### LINKED ISSUE #5614: Feature request: privacy statement and imprint
When using the admin frontend for more than a few admins (e.g. lecturers that upload videos directly from the admin ui), there has to be a way to supply a privacy statement and an imprint. Both should off course be configurable and being linked in the UI´s footer.

Frontend issue: https://github.com/opencast/opencast-admin-interface/issues/297

pselsa:
closes #5614

- Added imprint and privacy statement sample files
- Added docs
- also see frontend PR: https://github.com/opencast/opencast-admin-interface/pull/306

### DIFF
--- a/etc/org.opencastproject.organization-mh_default_org.cfg
+++ b/etc/org.opencastproject.organization-mh_default_org.cfg
@@ -132,6 +132,13 @@ prop.org.opencastproject.admin.mediamodule.url=${prop.org.opencastproject.engage
 #
 #prop.org.opencastproject.admin.feedback.url=
 
+# Flag to display imprint and privacy in admin-UIs footer
+#
+# Value: <boolean>
+# Default: false
+#
+prop.org.opencastproject.admin.display_about=false
+
 # Link to the Engage UI.
 #
 # Value: <a complete url with scheme and port>
--- a/etc/ui-config/mh_default_org/admin-ui/imprint.de.html
+++ b/etc/ui-config/mh_default_org/admin-ui/imprint.de.html
@@ -1,0 +1,9 @@
+<h1>Muster Impressum</h1>
+
+<p>
+<b>Test Universität</b><br>
+Haupstraße 1<br>
+123456 Musterstadt<br>
+Tel. +49 12345678<br>
+Fax +49 12345678
+</p>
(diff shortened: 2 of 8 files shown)


etc/ui-config/mh_default_org/admin-ui/imprint.de.html
```diff
@@ -0,0 +1,9 @@
+<h1>Muster Impressum</h1>
```
gregorydlogan, line 1: IMO these should be renamed to something like `sample-imprint.$LANG.html`, or the default contents should be blank when rendered.  We don't have sample data like this showing up anywhere else in the UI, so I can see adopters being confused when this pops up with their next upgrade.

Not having the files needs to be properly handled by the UI anyway, but if you want to keep the files then we should provide blank (with commented out sample maybe) rendering defaults.

gregorydlogan (changes requested):

etc/ui-config/mh_default_org/admin-ui/imprint.de.html
pselsa: Hi, I added a configuration to display the footer elements (which defaults to false), hope this resolves your concern.

pselsa: (also see https://github.com/opencast/opencast-admin-interface/pull/306)

etc/org.opencastproject.organization-mh_default_org.cfg
```diff
@@ -132,6 +132,13 @@ prop.org.opencastproject.admin.mediamodule.url=${prop.org.opencastproject.engage
 #
 #prop.org.opencastproject.admin.feedback.url=
 
+# Flag to display imprint and privacy in admin-UIs footer
+#
+# Value: <boolean>
+# Default: false
+#
+prop.org.opencastproject.admin.display_about=false
```
gregorydlogan, line 140: Defaults should be commented out by default, like this

```suggestion
#prop.org.opencastproject.admin.display_about=false
```

Otherwise looks good to me.

gregorydlogan (changes requested):
Pending this very minor fix this should be in for 16.0

PR #5675, opened 2024-03-14
Add synchronous deletion to Search API

JulianKniephoff:
We will need this in #5172.

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
@@ -224,6 +224,47 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
   }
 
   @DELETE
+  @Path("removeSynchronously/{id}")
+  @Produces(MediaType.APPLICATION_XML)
+  @RestQuery(
+      name = "remove",
+      description = "Removes a mediapackage from the search index.",
+      pathParameters = {
+          @RestParameter(
+              description = "The media package ID to remove from the search index.",
+              isRequired = true,
+              name = "id",
+              type = RestParameter.Type.STRING
+          )
+      },
+      responses = {
+          @RestResponse(description = "OK", responseCode = HttpServletResponse.SC_OK),
+          @RestResponse(
+              description = "The mediapackage could not be delete by you or does not exist",
+              responseCode = HttpServletResponse.SC_BAD_REQUEST
+          ),
+          @RestResponse(
+              description = "There has been an internal error and the mediapackage could not be deleted",
+              responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+          )
+      },
+      returnDescription = "Nothing"
+  )
+  public Response removeSynchronously(@PathParam("id") String mediaPackageId) throws SearchException {
+    try {
+      boolean removed = searchService.deleteSynchronously(mediaPackageId);
+      if (removed) {
+        return Response.ok().build();
+      } else {
+        return Response.status(Response.Status.BAD_REQUEST).build();
+      }
+    } catch (Exception e) {
+      logger.info("Unable to remove mediapackage {} from search index: {}", mediaPackageId, e.getMessage());
+      return Response.serverError().build();
+    }
+  }
+
+  @DELETE
   @Path("/seriesId/{seriesid}")
   @Produces(MediaType.APPLICATION_XML)
   @RestQuery(
--- a/modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchService.java
+++ b/modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchService.java
@@ -98,6 +98,17 @@ public interface SearchService {
   Job deleteSeries(String seriesId) throws SearchException, UnauthorizedException, NotFoundException;
 
   /**
+   * Immediately removes the given mediapackage from the search service.
+   *
+   * @param mediaPackageId
+   *          the mediapackage
+   * @return <code>true</code> if the mediapackage was deleted
+   * @throws SearchException
+   *           if deletion failed
+   */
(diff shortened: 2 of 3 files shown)


modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -223,6 +223,47 @@ public Response remove(@PathParam("id") String mediaPackageId) throws SearchExce
...
+  public Response removeSynchronously(@PathParam("id") String mediaPackageId) throws SearchException {
+    try {
+      boolean removed = searchService.deleteSynchronously(mediaPackageId);
+      if (removed) {
+        return Response.ok().build();
+      } else {
+        return Response.status(Response.Status.BAD_REQUEST).build();
+      }
+    } catch (Exception e) {
+      logger.info("Unable to remove mediapackage {} from search index: {}", mediaPackageId, e.getMessage());
```
lkiesow, line 262: Is this an actual error? In that case, do we want this to be logged as an error including a stack trace? Or is this something that appears if you request a non-existing event and this should be more like a debug message?

lkiesow (commented):
Haven't tested this yet, but generally looks good.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
JulianKniephoff: I'm not exactly sure what you are asking. This catches everything that can go wrong in `deleteSynchronously`. I can change the log call to include the stack trace, sure. This pattern is mostly copied from all the other "wrappers" in this file. 🤔🤷‍♀️

gregorydlogan:
Mentioning #5597 to remind me to take this into account when I next look at that PR.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
gregorydlogan: He (and I) would be temped to handle things that are Exceptions-but-not-errors like deleting a non-existent mediapackage differently. By all means have a catch all, but don't log a huge stack if the reason is user error :D

gregorydlogan (approved):

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
JulianKniephoff: Deleting a non-existent MP should make `deleteSynchronously` return `false`, so that's handled.

As far as I can see this should only happen in truly exceptional cases. In which case we might want to log the stacktrace, fair, but that's not done anywhere else in this file.

I can change it if you (as in any of you) insist, but I also find it a bit weird to only change it in this method (and maybe also just in this file? haven't looked at other remote impls, though), in which case it's beyond the scope of this PR IMO.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -223,6 +223,47 @@ public Response remove(@PathParam("id") String mediaPackageId) throws SearchExce
...
+          @RestResponse(
+              description = "There has been an internal error and the mediapackage could not be deleted",
+              responseCode = HttpServletResponse.SC_INTERNAL_SERVER_ERROR
+          )
+      },
+      returnDescription = "Nothing"
+  )
+  public Response removeSynchronously(@PathParam("id") String mediaPackageId) throws SearchException {
+    try {
+      boolean removed = searchService.deleteSynchronously(mediaPackageId);
```
gregorydlogan, line 255: Oh boy.  You know what this *doesn't* do?  Update the index.  So you've removed it from search, but the admin ui doesn't know that :D

gregorydlogan (changes requested):
I'm sorry.  I'm so, so sorry.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
karendolan: The e.getMessage() is not the stack trace. It might be helpful to add an accompanying TRACE log to log the actual stack trace, "e", in case that INFO log indicates something difficult to debug.

Is there an INFO log somewhere else that logs that this mediapackage was requested to be deleted (and the user, if known)?

JulianKniephoff: I don't think this remote impl should log anything that doesn't have to do with its job of forwarding requests to the actual impl. If we want a log about users deleting things from search, this is not the place IMO. And again, the existing methods in here don't do that, either. 🤷‍♀️

JulianKniephoff: Don't get me wrong, y'all are making good suggestions, but in the interest of time and keeping this PR clean, I'd be tempted to resovle this particular convo in the upcoming days, if no one brings up a good reason why this particular PR needs to improve upon the logging of this entire particular class (or even others) that it happens to need to touch. 😅

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
JulianKniephoff: Neither does `remove`, right? My understanding is that this is an internal API that would be used by some other part in the system to delete stuff from search, and then ALSO delete it from other places using APIs/services responsible for **those** places.

gregorydlogan: Yeah, fair point now that you mention it.

PR #5685, opened 2024-03-20
Paella 7 mp4 quality selector

### LINKED ISSUE #5550: Paella 7: Switch between video resolutions no longer possible for progressive download video streaming
The default encoding profile in Opencast creates different video resolutions. With the Paella 7 player, however, it is no longer possible to switch between these different resolutions during progressive download video streaming. Only one of the video resolutions is loaded and played by the player. (See also https://github.com/polimediaupv/paella-core/issues/342#issuecomment-1863982955 why this feature was dropped in Paella 7).

Since it makes no sense to create multiple video resolutions without the ability to switch between them in the player, the default behaviour should be adjusted. One option would be to switch from progressive download to adaptive HLS streaming. This way, the video resolution in the Paella 7 would automatically adapt to the available network bandwidth, and the user would again be able to manually switch between the different resolutions.

miesgre:
### Your pull request should…

* [x] have a concise title
* [x] close #5550
* [ ] pass automated tests

### DIFF
--- a/etc/ui-config/mh_default_org/paella7/config.json
+++ b/etc/ui-config/mh_default_org/paella7/config.json
@@ -266,8 +266,12 @@
                 }
             }
         },
-        "es.upv.paella.mp4VideoFormat": {
+        "es.upv.paella.mp4MultiQualityVideoFormat": {
             "enabled": true,
+            "order": 1
+        },
+        "es.upv.paella.mp4VideoFormat": {
+            "enabled": false,
             "order": 1
         },
         "es.upv.paella.htmlVideoFormat": {
--- a/modules/engage-paella-player-7/package-lock.json
+++ b/modules/engage-paella-player-7/package-lock.json
@@ -7,7 +7,8 @@
       "name": "paella-7-opencast",
       "dependencies": {
         "paella-basic-plugins": "^1.44.0",
-        "paella-core": "^1.46.1",
+        "paella-core": "^1.48.0",
+        "paella-mp4multiquality-plugin": "^1.47.1",
         "paella-skins": "^1.32.3",
         "paella-slide-plugins": "^1.41.4",
         "paella-user-tracking": "^1.42.0",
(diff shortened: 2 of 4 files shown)


LukasKalbertodt (approved):
Code change looks good to me!

As context: this is the result of work in the past few months. After some back and forth, some changes to `paella-core` were made, but the majority of this quality switching logic is part of a new external plugin `mp4MultiQualityVideoFormat`.

There is a [Tobira PR with the exact same change](https://github.com/elan-ev/tobira/pull/1143). That PR has a test deployment and a couple comments detailing how this was tested already. Sascha found one small problem with Firefox in some scenarios, but that doesn't need to block this in our opinion.

snoesberger:
I did the same tests for this PR as I did already for the [Tobira PR](https://github.com/elan-ev/tobira/pull/1143) on my MacBook Pro with Safari, Chrome, Firefox and Opera and the quality selection works as expected.
As Lukas already mentioned, the really small problem with Firefox doesn't need to block this PR.

Thanks for the implementation!

Arnei (approved):
Tested it briefly with a develop install of Opencast and it works. Code changes to Opencast are straightforward enough :)

It is arguably unintuitive that the quality selector does not pop up if there only is one quality, or at least it had me wondering if I did something wrong. But that's no reason to block this.

Arnei:
By the way, we probably want this to go into earlier branches than develop. I'd say `r/15.x` at least, but maybe `r/14.x` could also work?

miesgre:
> By the way, we probably want this to go into earlier branches than develop. I'd say `r/15.x` at least, but maybe `r/14.x` could also work?

paella-core needs to be updated to `1.48.1`

- r/14.x uses `paella-core@1.43.0`
- r/15.x uses `paella-core@1.46.1`

If there is no problem in updating the version of paella-core in r/14.x and r/15.x, I can rebase to r/14.x or r/15.x.

Arnei:
I don't know how Paella does it its versioning, but since there are no major version changes it should be fine? Probably better to just test. I guess you would now better if updating `paella-core` is gonna cause inter-dependency issues in `r/15.x` or `r/14.x` ^^

KatrinIhler:
I'd probably aim this at 15, but I don't know how we handled paella updates in the past tbh.

miesgre:
> I don't know how Paella does it its versioning, but since there are no major version changes it should be fine? Probably better to just test. I guess you would now better if updating `paella-core` is gonna cause inter-dependency issues in `r/15.x` or `r/14.x` ^^

I should be secure to update from `paella-core@1.46.1` to `paella-core@1.48.1`.
I made a rebase and to r/14.x and create a new branch to test. In my test it works as expected. Please test it.
You can find the branch here: https://github.com/miesgre/opencast/tree/paella-mp4-quality-selector-r14x

The config file needs to be updated to use the quality change.
The PR updated the config file, but if some institution is not using the default, will need to update it.

Disable the `es.upv.paella.mp4VideoFormat` and add the `es.upv.paella.mp4MultiQualityVideoFormat`.

```json
{
    "plugins": {
        ...
        "es.upv.paella.mp4VideoFormat": {
            "enabled": false,
            "order": 1
        },
        "es.upv.paella.mp4MultiQualityVideoFormat": {
            "enabled": true,
            "order": 1
        },
      }
        ... other plugin settings
    }
}
```

KatrinIhler:
Result of the Technical Meeting: Could probably go into 14, but since it's soon EOL and it would be a "big" jump, let's aim at 15 instead.

miesgre:
> Result of the Technical Meeting: Could probably go into 14, but since it's soon EOL and it would be a "big" jump, let's aim at 15 instead.

Rebased to r/15.x

PR #5686, opened 2024-03-20
Fix editor track selection for updating tags

geichelberger:
Set a media package element reference on the edited track to identify it later so that the tags can be updated based on the source track.

The check if `editedTrack.getReference()` is null should not be necessary as every track should have a reference, or is that assumption wrong?

* [ ] have a concise title
* [ ] pass automated tests


Arnei:
In this case you yourself are guaranteeing that each track has a reference by setting it yourself, right? As far as I can tell, in general there is no guarantee that an MPE has a reference, so checking for `null` is probably still advisable?

PR #5684, opened 2024-03-20
Make the paella 7 tests locally without depending on develop.opencast…

miesgre:
@lkiesow as discussed at the Opencast summit in Zaragoza
The paella 7 tests depended on develop.opencast.org
This PR modifies this and now the tests are done locally.

A mediapackage has been added for testing and an mp4 file, reducing the quality so that the size of the repository does not increase much.

### Your pull request should…

* [x] have a concise title
* [x] close #5672
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #5750, opened 2024-04-16
Paella: Fix error displayed before authenticating user.

miesgre:
This Fix a bug in the loading process when the player tries lo authenticate the user. 
Before redirecting to the login page, an error message is briefly displayed.

To test, try to open a video that does not exists without being authenticated. Ex: https://develop.opencast.org/paella7/ui/watch.html?id=no-valid-video

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/engage-paella-player-7/src/js/PaellaOpencast.js
+++ b/modules/engage-paella-player-7/src/js/PaellaOpencast.js
@@ -78,10 +78,6 @@ function myWebsiteCheckConsentFunction(type) {
   return consent_level[type] || false;
 }
 
-function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
-
 const initParams = {
   customPluginContext: [
     require.context('../plugins', true, /\.js/),
@@ -106,6 +102,9 @@ const initParams = {
   },
 
   loadVideoManifest: async function (url, config, player) {
+    const stopLoadVideoManifest = () => {
+      return new Promise(() => {});
+    };
     // check cookie consent (if enabled)
     const cookieConsent = config?.opencast?.cookieConsent?.enable ?? true;
     const cookieConsentConfig = config?.opencast?.cookieConsent?.config ?? {
@@ -147,9 +146,9 @@ const initParams = {
         player.log.info('Video not found and user is not authenticated. Try to log in.');
         location.href = getUrlFromOpencastPaella('auth.html?redirect=' + encodeURIComponent(window.location.href));
         // We need to interrupt the player loading to redirect the user to the auth page.
-        await sleep(100);
+        await stopLoadVideoManifest();
         // This Error should not happen, as the user is redirected to the auth page.
-        throw Error('Video not found and user is not authenticated. Try to log in.');
+        throw Error('Video not found and user is not authenticated. Log in and try again.');
       }
       else {
         // TODO: the video does not exist or the user can't see it
--- a/modules/engage-paella-player-7/src/js/PaellaOpencast.js
+++ b/modules/engage-paella-player-7/src/js/PaellaOpencast.js
@@ -102,8 +102,13 @@ const initParams = {
   },
 
   loadVideoManifest: async function (url, config, player) {
-    const stopLoadVideoManifest = () => {
-      return new Promise(() => {});
+    const redirectToAuthPage = async () => {
+      location.href = getUrlFromOpencastPaella('auth.html?redirect=' + encodeURIComponent(window.location.href));
+      // Create a loader spinner.
+      player._loader = new player.initParams.Loader(player);
+      player._loader.create();
+      // We need to interrupt the player loading to redirect the user to the auth page.
+      await new Promise(() => {});
     };
     // check cookie consent (if enabled)
     const cookieConsent = config?.opencast?.cookieConsent?.enable ?? true;
@@ -144,9 +149,7 @@ const initParams = {
 
       if (!me.roles.includes('ROLE_USER')) {
         player.log.info('Video not found and user is not authenticated. Try to log in.');
-        location.href = getUrlFromOpencastPaella('auth.html?redirect=' + encodeURIComponent(window.location.href));
-        // We need to interrupt the player loading to redirect the user to the auth page.
-        await stopLoadVideoManifest();
+        await redirectToAuthPage();
         // This Error should not happen, as the user is redirected to the auth page.
         throw Error('Video not found and user is not authenticated. Log in and try again.');
       }
--- a/modules/engage-paella-player-7/src/js/PaellaOpencast.js
+++ b/modules/engage-paella-player-7/src/js/PaellaOpencast.js
@@ -78,6 +78,10 @@ function myWebsiteCheckConsentFunction(type) {
   return consent_level[type] || false;
 }
 
+function sleep(ms) {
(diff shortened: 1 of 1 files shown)


miesgre:
@KatrinIhler This PR should be to r/15.x? Or it is ok in r/14.x?

KatrinIhler:
> @KatrinIhler This PR should be to r/15.x? Or it is ok in r/14.x?

Since this is a small bug fix, 14 is fine imo.

modules/engage-paella-player-7/src/js/PaellaOpencast.js
```diff
@@ -142,6 +146,10 @@ const initParams = {
       if (!me.roles.includes('ROLE_USER')) {
         player.log.info('Video not found and user is not authenticated. Try to log in.');
         location.href = getUrlFromOpencastPaella('auth.html?redirect=' + encodeURIComponent(window.location.href));
+        // We need to interrupt the player loading to redirect the user to the auth page.
+        await sleep(100);
+        // This Error should not happen, as the user is redirected to the auth page.
+        throw Error('Video not found and user is not authenticated. Try to log in.');
       }
```
karendolan, line 155: Hi @miesgre, our site has tried this same approach in the same part of the code. We still see the red error circle flashing prior to redirecting the page. It is as if the browser requires the current running Javascript function to complete prior to addressing the `location.href` at the point that it's requested. We have also wrapped promises and timeouts around everything to try break the execution chain, but still get the red error flash before the redirect is executed.

Is it possible to throw an exception, like a `throw Error`, to break the execution chain, but when caught, the red error UI artifact is not shown? Maybe a special keyword in the Error message like "not an error but a redirect request". It's a common approach in Java to throw an Exception to go down a different code path.

Another idea, which is more hacky, is to create a UI artifact that obscures the red error UI artifact, so that when the Error is thrown and breaks the execution chain, allowing the browser to address the `location.href` request, no red Error UI artifact shows prior to the `location.href` being executed.

karendolan: In summary, the `location.href()` request is not executed in-line, at least in browsers like Chrome. It is held in queue until the end of the Javascript execution chain. Until after the `throw Error()` at the end of this code sequence.

karendolan: [EDIT] Our site takes longer to for the `location.href` to resolve because we do an extra redirect. Opencast adopters that use the regular Opencast login might not have the same lag time response from `location.href` as we do.

Adding this `sleep(100)` might be sufficient to prevent the Error artifact from displaying for the default login page redirect. I don't want to imply that it doesn't help at all.

miesgre: @karendolan , please test ithis exact code in your installation. I think it should word. The time in the sleep does not matter. The important is to break the execution chain because the location.href is not executed immediately..

The throw after the sleep is **never** executed!. I put it only in case one day browsers change this behavior.

You can even change the sleep function to a non resolve promise function if you want.

```js
function nonResolvingPromise() {
  return new Promise(() => {});
}

...
await nonResolvingPromise();
```

karendolan: @miesgre What is different about our site is that the` location.href` does a redirect chain. It ` returns an HTTP 301, then HTTP 302, then HTTP 200 on the redirected page. This is different than the default Opencast login page which returns a HTTP 200 immediately. Our sites issue could be related to the back and forth of the HTTP 301.

I'll test adding a dummy location.href, that immediately returns a HTTP 200 instead of the HTTP 301, to see if I still see the following log after the sleep().

```
      await sleep(100);
      player.log.debug('After performing the auth redirect and sleep. With persisted logging on in the browser, this log should not show.');
      // This Error should not happen, as the user is redirected to the auth page.
      throw Error('Video not found and user is not authenticated. Try to log in.');
      ```

Comment (karendolan) [reply modules/engage-paella-player-7/src/js/PaellaOpencast.js:155]: > I'll test adding a dummy location.href, that immediately returns a HTTP 200 instead of the HTTP 301, to see if I still see the following log after the sleep().

In Safari, I verified that it's the HTTP 301 that causes the havoc. The HTTP 200 URL displays the log line before the sleep and not the log line after the sleep. The HTTP 301 URL displays the log line before the sleep, the one after the sleep, the error log, and the flash of red error on the page prior to the redirect.

Test by adding and not adding the redirect param .../watch.html?id=<mpId>&redirect=301

```
      if (!me.roles.includes('ROLE_USER')) {
        player.log.info('Video not found and user is not authenticated. Try to log in.');
+            if( utils.getUrlParameter('redirect') === '301' ) {
+               // Redirect to a URL that returns a HTTP 301
+               location.href ='https://www.iana.org/domains/example';
+               player.log.debug('Redirected to HTTP 301 before sleep');
+               await sleep(100);
+               player.log.debug('Redirected to HTTP 301 after sleep');
+            } else {
+               // Redirect to a URL that returns a HTTP 200
+               location.href = 'https://example.com';
+               player.log.debug('Redirected to HTTP 200 before sleep');
+               await sleep(100);
+               player.log.debug('Redirected to HTTP 200 after sleep');
+            }
         // This Error should not happen, as the user is redirected to the auth page.
        throw Error('Video not found and user is not authenticated. Try to log in.');
      }
     
```

[EDIT] When I throttle the network and disable cache, both messages, the before and after sleep(), show on Chrome, Safari and Firefox. The HTTP 200 is faster than the HTTP 301 when the network is not throttled, so that the second message (and error message) is usually missing on fast loads. This implies that the location.href is asynchronous, but it cannot be awaited. The `await nonResolvingPromise();` is working in my local tests.

Comment (miesgre) [reply modules/engage-paella-player-7/src/js/PaellaOpencast.js:155]: Thanks @karendolan for testing it! I updated the PR with a non resolving promise.
Please test it!

Comment (miesgre) [reply modules/engage-paella-player-7/src/js/PaellaOpencast.js:155]: @karendolan I added a spinner while redirecting. This should solve your problem.

Comment (karendolan) [reply modules/engage-paella-player-7/src/js/PaellaOpencast.js:155]: @miesgre this works great for me!

I can approve it, but I can't merge it because there is a failing github actions Playright test that isn't related to the code but to the target test video.

```
    Error: Timed out 5000ms waiting for expect(locator).toHaveTitle(expected)

    Locator: locator(':root')
    Expected string: "Dual-Stream Demo - No series | Opencast"
    Received string: "Paella Player 7"
    ```

karendolan (changes requested):
The pull works, fixes the red error flash prior to auth redirect and the spinner shows when the redirect takes a few moments.

The github action Playright test fail needs to be addressed.

miesgre:
> The pull works, fixes the red error flash prior to auth redirect and the spinner shows when the redirect takes a few moments.
> 
> The github action Playright test fail needs to be addressed.

@karendolan Don't worry about paella tests. Fails because in `r/14.x` tests are done using https://develop.opencast.org/paella7/ui/watch.html?id=ID-dual-stream-demo, but this video is not public now.

In https://github.com/opencast/opencast/pull/5684 It is solved, but it is only in `develop`

karendolan (approved):
Then all good! Approved.

PR #5752, opened 2024-04-18
Reformat new admin UI `pom.xml`

JulianKniephoff:
What it says on the tin. Literally just white space changes for consistency. https://github.com/opencast/opencast/pull/5752/files?w=1 should be empty. Targeting 14.x since this is the oldest branch this is going to annoy me in.


lkiesow:
Looks good but will need manual merge intervention in all branches. Can we not just fix this in `develop` and ignore the old branches?

PR #5755, opened 2024-04-22
Skip the OBR

JulianKniephoff:
The `maven-bundle-plugin` installs the resulting bundles to a/the so called OBR, but we don't use that, so this just wastes build time.

On my M1 MacBook Air this saves roughly **2 minutes**!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -569,6 +569,7 @@
           <version>5.1.9</version>
           <inherited>true</inherited>
           <configuration>
+            <obrRepository>NONE</obrRepository>
             <instructions>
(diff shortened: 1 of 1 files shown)


gregorydlogan (approved):
TIL I can turn that off. Holy crap.

ziegenberg:
For future reference ...

Apache Felix OSGi Bundle Repository (OBR): https://felix.apache.org/documentation/subprojects/apache-felix-osgi-bundle-repository.html

ziegenberg:
Even @geichelberger does not know if this is useful for us. So it can go for now.

PR #5757, opened 2024-04-23
Add slide text and segments to Tobira harvest API

owi92:
See commits. With this PR, the generated slide text is exposed so we can use that in Tobira's search index. I'm not super sure if the filter I built for this could let through false positives or negatives, though in testing this didn't seem to be the case. Please let me know if you think this might be an issue, and/or have any suggestions how this could be solidified.

Furthermore, the slide segments with their respective starting time is passed in order to be used for Paella's slide plugin on the Tobira side.

Related Tobira issues: https://github.com/elan-ev/tobira/issues/368, https://github.com/elan-ev/tobira/issues/1065

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
@@ -165,6 +165,7 @@ class Item {
           }))),
           Jsons.p("captions", Jsons.arr(captions)),
           Jsons.p("slideText", slideText.map(t -> t.toString()).orElse(null)),
+          Jsons.p("segments", Jsons.arr(findSegments(mp))),
           Jsons.p("updated", event.getModified().getTime())
       );
     }
@@ -326,6 +327,17 @@ class Item {
         .map(a -> a.getURI().toString())
         .findFirst()
         .orElse(null);
+  }
+
+  private static List<Jsons.Val> findSegments(MediaPackage mp) {
+    return Arrays.stream(mp.getAttachments())
+      .filter(a -> a.getFlavor().getSubtype().equals("segment+preview"))
+      .map(s -> Jsons.obj(
+          Jsons.p("uri", s.toString()),
+          Jsons.p("startTime", s.getReference().getProperty("time"))
+        )
+      )
+      .collect(Collectors.toCollection(ArrayList::new));
   }
 
   private static Jsons.Val findTimelinePreview(MediaPackage mp) {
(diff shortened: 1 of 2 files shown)


modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
```diff
@@ -125,6 +125,18 @@ class Item {
...
 
+      // Get the generated slide text.
+      final var slideText = Arrays.stream(mp.getElements())
+          .filter(mpe -> {
+            final var flavor = mpe.getFlavor();
+            final var isCatalog = mpe.getElementType() == MediaPackageElement.Type.Catalog;
+            final var isXml = mpe.getMimeType().eq(MimeType.mimeType("text", "xml"));
+            final var isText = flavor.getSubtype().equals("text");
+            return isCatalog && isXml && isText;
+          })
```
mtneug: You probably want to only search for element with flavor `mpeg-7/text`.

mtneug (changes requested):

modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
owi92: Oh right, that is much simpler and more accurate. Thank you!

modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
```diff
@@ -125,6 +125,12 @@ class Item {
 
       final var captions = findCaptions(mp);
 
+      // Get the generated slide text.
+      final var slideText = Arrays.stream(mp.getElements())
+          .filter(mpe -> mpe.getFlavor().eq("mpeg-7/text"))
+          .map(element -> element.getURI())
+          .findFirst();
```
LukasKalbertodt, line 132: I wonder... what if there are multiple of these elements? You just take the first here. Can anyone say whether in the real world, there might be multiple such elements?

But this isn't a blocker. I also just trust Matthias with the `mpeg-7/text` filter. I don't have enough experience to say whether that filters for less or more than what we want. :shrug:

LukasKalbertodt (approved):
Tested it locally and it works! I only have one note, but that's not too important. I think this can be merged.

gregorydlogan:
This seems to work to me.  Have not checked the Tobira/UI side of things, but the server side stuff shows up just fine.

PR #5759, opened 2024-04-24
Add SERIES.WRITE_ONLY list provider

geichelberger:
This adds a version of the series list provider that returns only series with `write` permission.

The alternative is to make this the default behavior.

* [ ] have a concise title
* [ ] pass automated tests


Arnei:
Not sure this is working as intended. I have a series with only read rights, but the series is still returned.
Bildschirmfoto vom 2024-07-10 16-43-04
Bildschirmfoto vom 2024-07-10 16-43-20

Is it because I tested as admin? (I also tried to test as non-admin user, but could not see series with only read-rights in the first place).

Furthermore, am I even correct in assuming that the listprovider should return all series where any kind of write permissions are set?

geichelberger:
It's because you are an admin; an admin gets all results regardless of whether he is part of the ACL.

gregorydlogan (approved):

PR #5763, opened 2024-04-30
Bump java version to 17

geichelberger:
This bumps the Java version to 17 and tests against 21

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan:
Both 17 and 21 seem like they work to me, at least with the things I can test.  IMO this should be in OC 16 since we're going to have a bunch of breaking changes anyway.  @geichelberger can you add a bit into the release notes for this?

lkiesow:
Which version of Java do we want to support? If we continue to support 11, shouldn't we ttest against 11.  So… maybe 11 and 21?

geichelberger:
The idea is to drop support for 11 and require Java 17.

lkiesow:
Actually, looking at https://endoflife.date/redhat-build-of-openjdk, maybe it makes sense to require 17 after all. But then I would suggest applying this patch after the OC 16 branch cut and making this a requirement for Opencast 17.

*(you were faster)*

geichelberger:
And the other reason is developer joy.

gregorydlogan:
We talked about this in the meeting last week. The sense was that we've already got a bunch of breaking changes with 16, moving the jdk requirements might make sense as well.

Tbh I don't particularly care, as long as it's clear :)

gregorydlogan:
@geichelberger This needs release notes before it can go in.

mtneug:
FYI: I changed to a Java 17 container base image since 15.0.

PR #5765, opened 2024-04-30
Handle missing tags in Paella Player 7

KatrinIhler:
Not all tracks have tags, for example live streams usually do not. #5592 assumes that, though, which will cause an error when opening the player. Also handle missing config gracefully.

This should be aimed at OC 16 after the branch cut.

### DIFF
--- a/modules/engage-paella-player-7/src/js/EpisodeConversor.js
+++ b/modules/engage-paella-player-7/src/js/EpisodeConversor.js
@@ -89,6 +89,17 @@ function getStreamType(track) {
   return result;
 }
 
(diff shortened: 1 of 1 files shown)


KatrinIhler:
Player tests are failing, but I don't think that's my fault.

miesgre:
> Player tests are failing, but I don't think that's my fault.

No. I created a PR to solve it. https://github.com/opencast/opencast/pull/5767

KatrinIhler:
Now the admin ui tests fail :upside_down_face:

PR #5810, opened 2024-05-06
Make internal publication optionally visible in external api

ciegler:
Makes the internal publication optionally visible in the external API. 
This makes it possible for the Ilias-plugin to decide when to show the link to the video editor (which is currently coupled to the  engage-publication).

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -1611,18 +1611,22 @@ public class EventsEndpoint implements ManagedService {
              },
              restParameters = {
                @RestParameter(name = "sign", description = "Whether public distribution urls should be signed.",
-                              isRequired = false, type = Type.BOOLEAN)
+                              isRequired = false, type = Type.BOOLEAN),
+               @RestParameter(name = "internal", description = "Whether internal publication should be listed.",
+                     isRequired = false, type = Type.BOOLEAN)
              },
              responses = {
                   @RestResponse(description = "The list of publications is returned.", responseCode = HttpServletResponse.SC_OK),
                   @RestResponse(description = "The specified event does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
-  public Response getEventPublications(@HeaderParam("Accept") String acceptHeader, @PathParam("eventId") String id,
-          @QueryParam("sign") boolean sign) throws Exception {
+
+  // in Response: add boolean for internal_publications like sign
+    public Response getEventPublications(@HeaderParam("Accept") String acceptHeader, @PathParam("eventId") String id,
+          @QueryParam("sign") boolean sign, @QueryParam("internal") boolean internal) throws Exception {
     try {
       final ApiVersion requestedVersion = ApiMediaType.parse(acceptHeader).getVersion();
       final Opt<Event> event = indexService.getEvent(id, elasticsearchIndex);
       if (event.isSome()) {
-        return ApiResponses.Json.ok(acceptHeader, arr(getPublications(event.get(), sign, requestedVersion)));
+        return ApiResponses.Json.ok(acceptHeader, arr(getPublications(event.get(), sign, internal, requestedVersion)));
       } else {
         return ApiResponses.notFound(String.format("Unable to find event with id '%s'", id));
       }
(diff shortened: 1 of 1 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1739,13 +1752,17 @@ private List<JValue> getPublicationCatalogsJson(Publication publication, Boolean
...
                @RestParameter(name = "sign", description = "Whether public distribution urls should be signed.",
+                              isRequired = false, type = Type.BOOLEAN),
+               @RestParameter(name = "includeInternalPublication", description = "Whether internal publication should be included.",
                               isRequired = false, type = Type.BOOLEAN)
              },
              responses = {
                   @RestResponse(description = "The track details are returned.", responseCode = HttpServletResponse.SC_OK),
                   @RestResponse(description = "The specified event or publication does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
+
+  // same as above
```
KatrinIhler, line 1763: This can probably go.

KatrinIhler: This can probably go.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1632,7 +1638,14 @@ public Response getEventPublications(@HeaderParam("Accept") String acceptHeader,
     }
   }
 
-  private List<JValue> getPublications(Event event, Boolean withSignedUrls, ApiVersion requestedVersion) {
+
+  private List<JValue> getPublications(Event event, Boolean withSignedUrls, Boolean includeInternalPublication, ApiVersion requestedVersion) {
+    if (includeInternalPublication) {
```
KatrinIhler: This is missing the API version check

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1739,13 +1752,17 @@ private List<JValue> getPublicationCatalogsJson(Publication publication, Boolean
...
                               isRequired = false, type = Type.BOOLEAN)
              },
              responses = {
                   @RestResponse(description = "The track details are returned.", responseCode = HttpServletResponse.SC_OK),
                   @RestResponse(description = "The specified event or publication does not exist.", responseCode = HttpServletResponse.SC_NOT_FOUND) })
+
+  // same as above
   public Response getEventPublication(@HeaderParam("Accept") String acceptHeader, @PathParam("eventId") String eventId,
-          @PathParam("publicationId") String publicationId, @QueryParam("sign") boolean sign) throws Exception {
+          @PathParam("publicationId") String publicationId, @QueryParam("sign") boolean sign, @QueryParam("includeInternalPublication") boolean includeInternalPublication) throws Exception {
```
KatrinIhler: This and the line in the annotation above are probably leftovers and can go, since we don't use the parameter in this case.

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
KatrinIhler: Also you can write something like 

```
.filter(publication -> (includeInternalPublication || EventUtils.internalChannelFilter.apply(publication))
```

(not including the version check) :)

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1757,10 +1774,16 @@ public Response getEventPublication(@HeaderParam("Accept") String acceptHeader,
...
   }
 
+
   private JObject getPublication(String eventId, String publicationId, Boolean withSignedUrls, ApiVersion requestedVersion)
           throws SearchIndexException, NotFoundException {
     for (final Event event : indexService.getEvent(eventId, elasticsearchIndex)) {
-      List<Publication> publications = $(event.getPublications()).filter(EventUtils.internalChannelFilter).toList();
+      List<Publication> publications;
+      if (requestedVersion.isSmallerThan(VERSION_1_11_0)) {
+        publications = $(event.getPublications()).filter(EventUtils.internalChannelFilter).toList();
```
KatrinIhler: See comment to filter above

KatrinIhler (changes requested):
Does what it says on the tin. :D Just a couple of nitpicks below.

KatrinIhler:
We decided to re-aim this at 16.x after the branch cut and hope to get this into 16.0.

KatrinIhler (approved):

KatrinIhler:
This has a Nullpointer bug if you don't set the new parameter that I overlooked during the review, fix will come soon.

PR #5811, opened 2024-05-06
Turn old admin interface into plugin

lkiesow:
This patch disables the old admin interface by default. Users can still activate this if they really want/need the old interface. This should force users to acknowledge and test the new interface while allowing a fallback in some edge cases.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests


gregorydlogan (commented):
This works, and looks good.  One minor nit-pick is the operations in `fast.xml` and `partial-preview.xml`, both of which contain a hardcoded `admin-ng` path component.  Not sure how to address that though.

lkiesow:
I've updated the workflows to point to the new editor.
Though the links aren't actually used anywhere in Opencast I think.

PR #5831, opened 2024-05-08
Update ESLint

JulianKniephoff:
ESLint >= 9 brings a few breaking changes, making the dependabot PRs that try to update it fail; cf. #5824, #5819, #5813, #5812.

The main thing was a change in the default configuration format. There is a compatibility flag, but I don't think we want to use it, since the old way is also deprecated now. See also https://eslint.org/docs/latest/use/migrate-to-9.0.0#-new-default-config-format-eslintconfigjs.

Without that, updating basically entailed the following:

* The shared config in `docs/checkstyle` was converted to the new format,making it a full `npm`-package in the process.
  * This is required because of the way the new format handles plugins and other dependencies.
  * Note that this also makes the `resolve-plugins-relative-to` option obsolete, which is good since it is no longer supported. This patch removes it everywhere we used it.
  * I also added a dependabot rule to keep that package's few dependencies up to date.
* The `--config` option can't load such shared configs, though, so all the packages that use this combination got a new config of their own, basically deferring to that package.
* Modules which already had their own config file had it updated as
well.
* `.eslintignore` is no longer supported and thus got moved to the new config file(s).
* I also removed some unnecessary `eslint-disable` comments.

The `eslint-plugin-header` is no longer compatible with ESLint 9. I replaced it with `eslint-plugin-headers`, but that required kind of an ugly hack. See also https://github.com/robmisasi/eslint-plugin-headers/issues/4.

Note that this udpate might slightly change the linting rules; I did not check every last change to default rules and options! The one thing I hade to change was to add a setting to the `no-unused-vars`-rule, because the corresponding default changed, and we depend on the old default. See also https://eslint.org/docs/latest/use/migrate-to-9.0.0#-no-unused-vars-now-defaults-caughterrors-to-all.

Note also that the `lti`-module also uses ESLint, but as part of it being a `create-react-app`, so we don't control the ESLint version in there. I didn't touch that module. (Well, in this PR; see #5830.) We might want to change that in the future, though, since CRA is deprecated in itself.

Another thing we might want to change in the future is the `engage-paella-player-7`-module using the deprecated `@babel/eslint-parser`.

We also use a few deprecated ESLint rules, but for now I wanted to chagne the configs as little as possible, since the necessary changes are already quite sweeping. I might make a follow-up PR in the coming days.


JulianKniephoff:
OK so, CI fails here, because I failed to remember that installing local packages only symlinks them, so the target folder needs to have its dependencies already installed. I think there is a fix and I will try and push it in a second here, but this kind of raises the question of whether we shouldn't properly publish that shared config somewhere, be it on npm proper, or here on GitHub somehow.

And that in turn raises the question of how this relates to [the ESLint-config we already published](https://www.npmjs.com/package/@opencast/eslint-config-ts-react)…

PR #5058, opened 2023-06-28
Allow deletion of reference users from the Admin UI

Arnei:
Resolves #4789

External user providers will store users in the opencast database table `oc_user_refs`. However, external user providers may fail to remove users from Opencast when they should not exist anymore, and there is currently no easy way to remove users from that table (besides directly modifying the database). This can make it difficult for adopter to comply with data privacy policies.

This commit adds deletion functionality to the `JpaUserReferenceProvider`, and allows deletion of users from other providers than "system" or "opencast", even if they are not "manageable" (which applies to all these users). It does not allow any changes beyond that, i.e. you may now delete a user created by an external provider, but you still can't change their name.

### Your pull request should…

* [ ] have a concise title
* [x] [close an accompanying issue](https://help.github.com/en/articles/closing-issues-using-keywords) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
~* [ ] include migration scripts and documentation, if appropriate~
* [ ] pass automated tests
* [x] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #5833, opened 2024-05-13
Change how we make user references deletable in the admin UI(-s)

JulianKniephoff:
This indirectly forward-ports #5058 to the new admin UI; instead of special-casing the frontend, we just mark user references `manageable` in the backend.

Note, this reverts the frontend change made in #5058 as well since this is no longer needed.

/cc opencast/opencast-admin-interface#327.


Arnei:
Not sure what the actual effects of this change are, but if this makes all user references manageable I think it's a bad idea. Iirc #5058 was making an explicit exception on manageability for deletion purposes, but otherwise keeps it intact to prevent changes to references through the admin ui.

JulianKniephoff:
Fair, I thought `manageable` was only used for the delete button. We will just port @Arnei's original implementation to the new Admin-UI then. 🤷‍♀️

PR #5855, opened 2024-05-15
Fixed bug: tag-engage workflow failes when event is not published

ciegler:
Fixed the bug that the tag-engage workflow failed when trying to add or remove the video365 tag for an unpublished event. Now the workflow is skipped when the event has not been published yet.

### DIFF
--- a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/TagEngageWorkflowOperationHandler.java
+++ b/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/TagEngageWorkflowOperationHandler.java
@@ -134,7 +134,8 @@ public class TagEngageWorkflowOperationHandler extends AbstractWorkflowOperation
     try {
       mediaPackageForSearch = searchService.get(mediaPackageId);
     } catch (NotFoundException | UnauthorizedException e) {
-      throw new WorkflowOperationException(e);
+      logger.info("No tag engage workflow found for {}", mediaPackageId);
+      return createResult(WorkflowOperationResult.Action.SKIP);
     }
 
     // update tags & flavors in published mp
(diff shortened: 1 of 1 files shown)


modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/TagEngageWorkflowOperationHandler.java
```diff
@@ -134,7 +134,8 @@ public WorkflowOperationResult start(WorkflowInstance instance, JobContext conte
     try {
       mediaPackageForSearch = searchService.get(mediaPackageId);
     } catch (NotFoundException | UnauthorizedException e) {
-      throw new WorkflowOperationException(e);
+      logger.info("No tag engage workflow found for {}", mediaPackageId);
```
KatrinIhler, line 137: That logging message isn't quite correct: I'd write something like "Media package {} isn't published to Engage, skipping." Because the workflow does exist, and so does the media package, but not in the SearchService/Engage.

KatrinIhler: That logging message isn't quite correct: I'd write something like "Media package {} isn't published to Engage, skipping." Because the workflow does exist, and so does the media package, but not in the SearchService/Engage.

KatrinIhler (changes requested):

modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/TagEngageWorkflowOperationHandler.java
```diff
@@ -134,7 +134,8 @@ public WorkflowOperationResult start(WorkflowInstance instance, JobContext conte
     try {
       mediaPackageForSearch = searchService.get(mediaPackageId);
     } catch (NotFoundException | UnauthorizedException e) {
-      throw new WorkflowOperationException(e);
+      logger.info("Media package {} is not published to Engage, skipping", mediaPackageId);
```
KatrinIhler, line 137: Just realized that we should only skip on NotFound, and fail on Unauthorized. Sorry... :sweat_smile:

KatrinIhler (dismissed):

snoesberger:
Tested, I was now able to add/remove tags for unpublished events.

KatrinIhler (approved):
I somehow can't restart the jobs for the checks, but this looks fine now (and the failure is just the port binding stuff again).

KatrinIhler:
@gregorydlogan Can you restart the jobs? :thinking:

gregorydlogan:
This has gone stale, so a rebase would get things going, but I can't otherwise start them.

KatrinIhler:
@ciegler Can you rebase this onto current develop and push the branch? That should hopefully fix the test failures.

PR #5856, opened 2024-05-15
Add support for processing and playing audio only files

jduehring:
This PR is a draft to add out-of-the-box audio-only support to Opencast. It adds conditions to the partial-publish WF, so that steps like timelinepreviews, theming etc. will be skipped when only an audio stream is detected. I also adds a basic encoding profile to encode the incoming audio stream to mp3. There are different ways to achieve actually playing the processed audiofile which I would like to propose / discuss:  

1. Unloading Paellaplayer and loading the native HTML Audioplayer after detecting audio-only. This approach is currently used in this PR as it was my first approach. 

image

2. After adding the "audio" streamtype to `EpisodeConversor.json` PaellaPlayer7 is actually capable of detecting/playing audio-only and by default displays the PaellaPlayer7 Logo: 

image

3. In the past I often used a custom encoding profile to encode a video with a default background image from the audio-source. But since this ends up in a mp4 video I guess it should not be the standard solution, but rather playing actual audio-only files.  

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


miesgre:
Keep in mind that option 1 completely disables paella and uses a standard player.
This has some implications. Do you want to show any of the paella plugins? For example, display event metadata? change the playbackrate?
Will subtitles of the event be generated? If subtitles are generated, with option 1 we would not see them and we should choose option 2.

Taking this into account, I would bet on option 2, or maybe configurable in config.json?
A poster frame could be generated and paella will show that image instead of the paella logo.
Example: https://paellaplayer.upv.es/#/demos/audio-and-images

snoesberger:
I agree with @miesgre. From a user's point of view, I would prefer option 2. That way, the user always has the same control options, regardless of whether the stream is a video or audio only. For example, they can adjust the playback speed, show subtitles, skip forward or backward, etc., for audio-only streams, just as they are used to doing with video.

Option 1 is also likely to be browser dependent and the behaviour may vary between browsers.

karendolan:
I also agree with @miesgre and @snoesberger about maintaining a similar experience for the student and maintaining the Paella player context for the published material. Perhaps, make the option to use the native player a configuration option.

jduehring:
Thanks for your feedback!

I implemented a config option `audioOnlyHTMLPlayer` that will let you decide if you want to use option 1 or option 2. 
If option  2 is chosen, a generated cover-image  will be used a a preview picture and background in the PaellaPlayer. The cover-image uses the default `coverimage.xsl` but can of course be modified.

snoesberger:
Thanks for the configuration options!

I have some additional remarks:
- I'd prefer to use the same audio encoding for audio-only as we use for videos (aac instead of mp3).
- I miss the scissors/editor for audio-only events. I think it should also be possible to cut, edit subtitles, etc. on audio-only events. Is the editor capable of handling audio-only events, or do we need to make changes to the editor as well?

wsmirnow:
I would expect this to work with all audio only formats like mp3, m4a (audio only mp4), ogg vorbis, opus.

etc/workflows/partial-publish.xml
```diff
@@ -12,21 +12,21 @@
...
-        if="NOT (${presentation_prepared_media} OR ${presenter_prepared_media})"
-        id="clone"
-        fail-on-error="true"
-        exception-handler-workflow="partial-error"
-        description="Preparing media">
+      if="NOT (${presentation_prepared_media} OR ${presenter_prepared_media})"
+      id="clone"
+      fail-on-error="true"
+      exception-handler-workflow="partial-error"
+      description="Preparing media">
```
lkiesow: These (and below) seem like unwanted changes.

etc/encoding/opencast.properties
```diff
@@ -255,3 +255,11 @@ profile.text-analysis.http.suffix = .#{time}.png
...
   -vf "fps=1,scale=1280:720:force_original_aspect_ratio=decrease,pad=1280:720:-1:-1 [in]; color=black:s=1280x720:r=1:d=1 [bg]; [bg][in] overlay=0:0 [mux]; [mux] curves=preset=increase_contrast,format=pix_fmts=gray [out]" \
   -frames:v 1 #{out.dir}/#{out.name}#{out.suffix}
+
+# Encode audio-only to mp3
+profile.audio-to-mp3.http.name = audio-to-mp3
+profile.audio-to-mp3.http.input = audio
+profile.audio-to-mp3.http.output = audio
+profile.audio-to-mp3.http.suffix = -audio.mp3
+profile.audio-to-mp3.http.mimetype = audio/mpeg
+profile.audio-to-mp3.http.ffmpeg.command = -i #{in.video.path} -vn -map_metadata -1 -sn -acodec libmp3lame -ab 128k -ar 44100 -ac 2  #{out.dir}/#{out.name}#{out.suffix}
```
lkiesow: Instead of still using mp3, why not use AAC in an `.m4a` instead? That should work everywhere today and is more efficient.

Also, for lecture capture I would usually suggest just going with mono (`-ac 1`) or not modify the channel count if you want to keep it.

Something like:
```suggestion
profile.audio-to-mp3.http.suffix = -audio.m4a
profile.audio-to-mp3.http.mimetype = audio/mp4
profile.audio-to-mp3.http.ffmpeg.command = -i #{in.video.path} -vn -map_metadata -1 -sn -c:a aac -b:a 64k -ar 44100  #{out.dir}/#{out.name}#{out.suffix}
```

lkiesow (commented):
Haven't actually tested this. Just some notes…

etc/encoding/opencast.properties
wsmirnow: @lkiesow limiting bitrate of an audio encoder without setting the channels is a bad idea because the bitrate per channel will be `<bitrate>/<#channels>`. Example for an stereo stream: `64k/2=32k per channel`. This might be ok but unexpected. I would suggest `-b:a 64k -ac1` or `-b:a 128k -ac 2` for high quality stream.

KatrinIhler:

jduehring:
I updated this PR so that the Editor can be used to cut audio-only files. This was achieved by encoding a still image (in this case I just put the OC Logo as default) for the duration of the audio-file. In the Paella Player the generated cover-image is beeing used as a background. 

One issue I haven't figured out is that the PaellaPlayer doesn't seem to recognize the duration of the file and just shows "NaN", even though the editor and metadata seem to be set correctly.

gregorydlogan:
Is the point of this PR to support ingesting *just* audio files (ie, a single mp3 file) or to support removing all video and just distributing the audio portion?  When ingesting just a single mp3 in the presenter/source flavour, I get these exceptions:

```
2025-06-10 13:42:30,913 | WARN  | (EncoderEngine:243) - Error while encoding {video=/home/greg/opencast/upstream/build/opencast-dist-develop-18-SNAPSHOT/data/opencast/workspace/mediapackage/9b5a22b9-4fb1-4961-ab2a-9012e21dc055/965ba8a0-bb49-42c9-9680-00f31e74c3bb/audio.mp3}  using profile 'mp4-audio-preview.http'
org.opencastproject.composer.api.EncoderException: Encoder exited abnormally with status 254
        at org.opencastproject.composer.impl.EncoderEngine.process(EncoderEngine.java:237) [!/:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.parallelEncode(ComposerServiceImpl.java:494) [!/:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1478) [!/:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:315) [!/:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:274) [!/:?]
2025-06-10 13:42:30,960 | INFO  | (ServiceRegistryJpaImpl:2201) - State set to WARNING for current service org.opencastproject.composer on host http://localhost
2025-06-10 13:42:30,996 | ERROR | (AbstractJobProducer$JobRunner:345) - Error handling operation 'ParallelEncode':
org.opencastproject.serviceregistry.api.ServiceRegistryException: Error handling operation 'ParallelEncode'
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1596) ~[?:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:315) [!/:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:274) [!/:?]
Caused by: org.opencastproject.composer.api.EncoderException: org.opencastproject.composer.api.EncoderException: Encoder exited abnormally with status 254
        at org.opencastproject.composer.impl.EncoderEngine.process(EncoderEngine.java:252) ~[?:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.parallelEncode(ComposerServiceImpl.java:494) ~[?:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1478) ~[?:?]
        ... 6 more
Caused by: org.opencastproject.composer.api.EncoderException: Encoder exited abnormally with status 254
        at org.opencastproject.composer.impl.EncoderEngine.process(EncoderEngine.java:237) ~[?:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.parallelEncode(ComposerServiceImpl.java:494) ~[?:?]
        at org.opencastproject.composer.impl.ComposerServiceImpl.process(ComposerServiceImpl.java:1478) ~[?:?]
        ... 6 more
2025-06-10 13:42:33,869 | ERROR | (WorkflowOperationWorker:140) - Workflow operation 'operation:'encode, state:'FAILED'' failed
org.opencastproject.workflow.api.WorkflowOperationException: org.opencastproject.workflow.api.WorkflowOperationException: One of the encoding jobs did not complete successfully
        at org.opencastproject.workflow.handler.composer.EncodeWorkflowOperationHandler.start(EncodeWorkflowOperationHandler.java:120) ~[?:?]
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.start(WorkflowOperationWorker.java:212) ~[!/:?]
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.execute(WorkflowOperationWorker.java:117) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.runWorkflowOperation(WorkflowServiceImpl.java:722) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.process(WorkflowServiceImpl.java:1755) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2116) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2082) [!/:?]
Caused by: org.opencastproject.workflow.api.WorkflowOperationException: One of the encoding jobs did not complete successfully
        at org.opencastproject.workflow.handler.composer.EncodeWorkflowOperationHandler.encode(EncodeWorkflowOperationHandler.java:229) ~[?:?]
        at org.opencastproject.workflow.handler.composer.EncodeWorkflowOperationHandler.start(EncodeWorkflowOperationHandler.java:118) ~[?:?]
        ... 10 more
```

jduehring:
> Is the point of this PR to support ingesting _just_ audio files (ie, a single mp3 file) or to support removing all video and just distributing the audio portion? When ingesting just a single mp3 in the presenter/source flavour, I get these exceptions:

Thanks for testing. Yes the point is to be able to upload and process audio-files like mp3/wav etc. . 
I believe the error you are seeing is the result of the absolute path to the `branding` folder in the ffmpeg command: 

`profile.mp4-audio-preview.http.ffmpeg.command = -loop 1 -i /etc/opencast/branding/audio-only.jpg -i #{in.video.path} \
`
I think an absolute path is neccessary here(?), so I used the most common path used in OC installations. That would need to be adjusted for local installations...

gregorydlogan:
> Thanks for testing. Yes the point is to be able to upload and process audio-files like mp3/wav etc. . I believe the error you are seeing is the result of the absolute path to the `branding` folder in the ffmpeg command:
> 
> `profile.mp4-audio-preview.http.ffmpeg.command = -loop 1 -i /etc/opencast/branding/audio-only.jpg -i #{in.video.path} \ ` I think an absolute path is neccessary here(?), so I used the most common path used in OC installations. That would need to be adjusted for local installations...

I guess that depends on how much work you're willing to put in :D  As you can see, there's already support for interpolation of variables, so we'd to add support for more variables.  Alternatively, make the jpg into a pseudo-video for the purposes of the operation, but I'm not sure what kinds of knock-on effects that would have elsewhere in the system.

Alternatively figure out where that file lives relative to the CWD of the ffmpeg process, though that would be just as problematic since we're going to have different configurations across the world.

jduehring:
> > Thanks for testing. Yes the point is to be able to upload and process audio-files like mp3/wav etc. . I believe the error you are seeing is the result of the absolute path to the `branding` folder in the ffmpeg command:
> > `profile.mp4-audio-preview.http.ffmpeg.command = -loop 1 -i /etc/opencast/branding/audio-only.jpg -i #{in.video.path} \ ` I think an absolute path is neccessary here(?), so I used the most common path used in OC installations. That would need to be adjusted for local installations...
> 
> I guess that depends on how much work you're willing to put in :D As you can see, there's already support for interpolation of variables, so we'd to add support for more variables. Alternatively, make the jpg into a pseudo-video for the purposes of the operation, but I'm not sure what kinds of knock-on effects that would have elsewhere in the system.
> 
> Alternatively figure out where that file lives relative to the CWD of the ffmpeg process, though that would be just as problematic since we're going to have different configurations across the world.

I spent some more time on this and might have found a better approach. Instead of requiring to use a custom image the `partial-preview` WF now uses the same generated cover-image that is used in the PaellaPlayer as background. 

For this to work I added the duration as an output to the `AnalyzeTracksWorkflowOperationHandler` which in turn is used as input by the `image-to-video` WOH. This also fixed my issue of the PaellaPlayer not displaying the duration correctly, although I am not exactly sure why x)  

In the long run we should probably update the editor so that it doesn't necessarily require a video preview to work. 

If you are happy with this approach (and it works for you) I'll add the respective docs.

Nepoh:
Just a short remark, as I played around with audio-only content in our Opencast + Wowza setup recently: I think the `WowzaStreamingDistributionService` may need some attention as well in order to support audio-only files.

The produced mediapackage references a SMIL file streaming URL, but it looks like that file is not generated in the streaming directory for audio-only media files.

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AnalyzeTracksWorkflowOperationHandler.java
```diff
@@ -103,6 +103,9 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
       properties.put(varName + "_video", Boolean.toString(track.hasVideo()));
       properties.put(varName + "_audio", Boolean.toString(track.hasAudio()));
       properties.put(varName + "_subtitle", Boolean.toString(track.hasSubtitle()));
+      if (track.getDuration() != null) {
+        properties.put(varName + "_duration", Float.toString(track.getDuration().floatValue() / 1000));
```
mtneug, line 107: This needs to be documented.

mtneug (changes requested):

modules/workflow-workflowoperation/src/main/java/org/opencastproject/workflow/handler/workflow/AnalyzeTracksWorkflowOperationHandler.java
jduehring: Yes, I added the respective documentation to the analyze tracks WOH docs.

gregorydlogan (approved):
This works to my satisfaction, though I don't have a Wowza install.  @jduehring do you want this in 18.x?  I *think* it can be pointed that way since it's not modifying  config files, only adding.

jduehring:
> This works to my satisfaction, though I don't have a Wowza install. @jduehring do you want this in 18.x? I _think_ it can be pointed that way since it's not modifying config files, only adding.

@Nepoh Thanks for the info! I also currrently don't have a Wowza to test, but I might be able to have a look at it a bit later. I think it might make sense to merge this first and add Wowza support later?

@gregorydlogan I changed the target branch to r/18.x , would be great to have it in.

Nepoh:
> Thanks for the info! I also currrently don't have a Wowza to test, but I might be able to have a look at it a bit later. I think it might make sense to merge this first and add Wowza support later?

This is ok for me, thanks. Unfortunately I won't be of much help when implementing this, but can test as soon as there is a patch available.

elladoger:
As we move forward with implementing audio-only file support, we suggest Audio-Test as a key integration test. This test will focus on verifying the core functionality of loading, processing, and playing back audio files. Implementing Audio-Test will give us confidence that the new audio functionality integrates smoothly with the existing application and meets our quality standards.

PR #5858, opened 2024-05-16
Change segment `startTime` from mpeg7 to milliseconds in Tobira API

LukasKalbertodt:
All durations in the API are transmitted as milliseconds, so this should be too. On the Tobira side we want to save it as milliseconds too, since Paella wants that anyway.

This is a breaking API change and the API was already released as OC 15.5, so why no API version bump? The Tobira API is declared internal and only to be used by Tobira. And Tobira did not use it yet! So nothing can break.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/pom.xml
+++ b/modules/tobira/pom.xml
@@ -55,6 +55,11 @@
     </dependency>
     <dependency>
       <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-mpeg7</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.opencastproject</groupId>
       <artifactId>opencast-dublincore</artifactId>
       <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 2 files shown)


JulianKniephoff (approved):
Works and looks good to me. The "missing" version bump seems reasonable to me as well.

LukasKalbertodt:
> This is a breaking API change and the API was already released as OC 15.5, so why no API version bump? The Tobira API is declared internal and only to be used by Tobira. And Tobira did not use it yet! So nothing can break.

Interesting: no one noticed the obvious flaw in this logic. If someone runs OC 15.5 (with the old string `startTime`) and then upgrades to a newer Tobira that supports harvesting segments (with milliseconds `startTime`), then that Tobira fails to harvest events with segments.

That's not great. But I suppose nothing that we can do about it now anymore. Just make people aware of not using OC 15.5 with Tobira 2.10 or later. Will add a note to the Tobira release notes.

PR #5862, opened 2024-05-21
Remove prepared flavor

KatrinIhler:
This removes the `prepared` flavor from the workflows and configuration as it currently serves no purpose and causes issues like making it hard to know which captions to update. It was initially introduced to archive the output of `prepare-av`, but that step was removed during the rework f the workflows. This would make #5598 obsolete.

Might still be able to go into OC 16?

### DIFF
--- a/docs/guides/admin/docs/releasenotes/remove-prepared-flavor.txt
+++ b/docs/guides/admin/docs/releasenotes/remove-prepared-flavor.txt
@@ -1,0 +1,2 @@
+This removes all remnants of the prepared flavor from the workflows and the EditorServiceImpl.cfg as it currently serves
+no purpose and only makes things unnecessarily complicated. Instead, the source flavor will be used directly.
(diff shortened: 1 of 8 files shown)


Arnei:
I'm in favor of this and it works for new mediapackages. But it might require migration for users with old mediapackages that were created with `prepared-av` in mind. So if someone with such mediapackages could test this, that would be great :)

gregorydlogan:
Is there a sysadmin meeting this could be brought up at? :)

Arnei (approved):
Adding "official" approval, as I am in favor of this. @KatrinIhler Did anyone speak up against this?

KatrinIhler:
> Adding "official" approval, as I am in favor of this. @KatrinIhler Did anyone speak up against this?

Not so far, no. Brought it up at the SysAdmin meeting and got no comments.

mtneug:
This is all configuration so anybody not liking this can change it back.

PR #5864, opened 2024-05-23
Update HTTPS migration guide to be more general & use standard SQL

LukasKalbertodt:
This commit does several things:
- The guide was targetted at migrating to HTTPS, but it's equally useful when changing the domain. I did the latter and only found this guide because Matthias linked it to me. The new title and position should help in discoverability.

- I added some information on how to speed up the `find` command. This is info I wished was there before I ran mine. In my case, the NFS had some kind of snapshots, which made the find command run for way longer than necessary.

- I updated the SQL query: the `INSTR` function used before is not standard SQL. But it's also not necessary: if `replace` does not find the needle, the original string is returned. Yes, this technically touches more rows than necessary, but that's unlikely to be a problem.

Other than that, I can confirm that this guide still works for 15.5.

Open question:
- Is the guide placed too prominently? It is currently a root doc which might be too much for a rather rarely used thing like that. But I haven't found a good place for it. "configuration" seems wrong, as it has hardly anything to do with that. Maybe we should have a top level "guides" folder?
- My comment about only `archive/` and `downloads/` being relevant for `find`: is that correct? And will that keep being correct for the foreseeable future?

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


mtneug (commented):
This guide still refers to Solr since this is targeting 15. Will you create a new PR for 16 that drops those instructions?

docs/guides/admin/docs/migration-domain-https.md
```diff
@@ -0,0 +1,44 @@
...
+4. Update the media packages on disk.
+   Run the following in the folder configured in `org.opencastproject.storage.dir` (usually your NFS).
+   Note: only the subfolders `archive/` and `downloads/` will be affected by this, so you can also run it in those two
+   individually.
+   This is especially useful to reduce the execution time of the command if your NFS contains other, Opencast-unrelated
+   files (e.g. `.snapshots/`).
+
+         # This changes the doman AND from http to https. Carefully adjust the command as needed!
+         find . -type f -name "*.xml" -exec \
+            sed -i 's/http\:\/\/old-domain\.example\.com\:80/https:\/\/new-domain.example.com/g' {} +
```
gregorydlogan: I would be tempted to add a bunch more explanation here in terms of what the old-domain value *could* look like.  Your example here has the port explicitly listed, but I'd bet a bunch won't.  But some will, and if you don't handle both case we'll have support cases showing up wondering what happened.  Perhaps we grep for the old value, if there's a way, and then have them use that?

Also, `sed -E` to remove the escaping would make end-user life easier IMO.

docs/guides/admin/docs/migration-domain-https.md
```diff
@@ -0,0 +1,44 @@
...
+
+5. Update database tables.
+   Note: there more than the following two tables containing the old domain name, but only these two are relevant.
+
+         UPDATE oc_assets_snapshot SET mediapackage_xml = REPLACE(
+            mediapackage_xml, 'http://old-domain.example.com', 'https://new-domain.example.com');
```
gregorydlogan: Similar to above, what if they have `:80` tacked onto their URLs? :)

docs/guides/admin/mkdocs.yml
```diff
@@ -174,6 +173,7 @@ nav:
    - Wowza Streaming Distribution Service: 'configuration/streaming-wowza.md'
    - YouTube Publication: 'configuration/youtubepublication.md'
 - Version Support: 'version-support.md'
+- Domain change/HTTPS migration: 'migration-domain-https.md'
```
gregorydlogan: This might be a bit too high level.  What about under Configuration, at the same level as `HTTPS` and `Database`?  Then it's findable (ie, not hidden in HTTPS itself), but not super-duper header level important?

gregorydlogan (changes requested):
While this is an edge case, it's something that *does* happen and it's good to have this documented somewhere.

docs/guides/admin/docs/migration-domain-https.md
LukasKalbertodt: Mhh I'm not sure about this. I don't have the time to develop a script to automatically extract the current domain name. I think with the comment above the command and the previous text mentioning `org.opencastproject.server.url`, it should be fairly clear? If you have any specific ideas how to improve this, let me know though.

The `sed -E` thing I changed... I hope this is what you meant.

docs/guides/admin/docs/migration-domain-https.md
LukasKalbertodt: Same as above, I only added `:80` now to make it consistent with the `find` command above.

LukasKalbertodt:
> This guide still refers to Solr since this is targeting 15. Will you create a new PR for 16 that drops those instructions?

Will do, after this is merged.

docs/guides/admin/docs/configuration/migration-domain-https.md
```diff
@@ -0,0 +1,44 @@
...
+4. Update the media packages on disk.
+   Run the following in the folder configured in `org.opencastproject.storage.dir` (usually your NFS).
+   Note: only the subfolders `archive/` and `downloads/` will be affected by this, so you can also run it in those two
+   individually.
+   This is especially useful to reduce the execution time of the command if your NFS contains other, Opencast-unrelated
+   files (e.g. `.snapshots/`).
+
+         # This changes the domain AND from http to https. Carefully adjust the command as needed!
+         find . -type f -name "*.xml" -exec \
+            sed -i -E 's/http:\/\/old-domain.example.com:80/https:\/\/new-domain.example.com/g' {} +
```
gregorydlogan: I was wrong, it's not `-E` I was thinking of, it was replacing the delimiters.  Like so:

```suggestion
            sed -i 's#http://old-domain.example.com:80#https://new-domain.example.com#g' {} +
```

I can't count the number of times I missed (or added extra) escapes and then burned a bunch of time trying to figure out what I did wrong.  Using a different delimiter helps a ton.

docs/guides/admin/docs/configuration/migration-domain-https.md
```diff
@@ -0,0 +1,44 @@
...
+2. Change `org.opencastproject.server.url` and test your set-up with a new publication (i.e. uploading a video).
+3. Put all your nodes into maintenance mode or, at least, do not process any videos.
+4. Update the media packages on disk.
+   Run the following in the folder configured in `org.opencastproject.storage.dir` (usually your NFS).
+   Note: only the subfolders `archive/` and `downloads/` will be affected by this, so you can also run it in those two
+   individually.
+   This is especially useful to reduce the execution time of the command if your NFS contains other, Opencast-unrelated
+   files (e.g. `.snapshots/`).
+
+         # This changes the domain AND from http to https. Carefully adjust the command as needed!
```
gregorydlogan, line 25: What about something like this?

```suggestion
         # This changes the domain AND from http to https. Carefully adjust the command as needed!
         # Your old URL may or may not have the port explicitly listed.  Check the previous value of org.opencastproject.server.url and match that.
```

docs/guides/admin/docs/configuration/migration-domain-https.md
```diff
@@ -0,0 +1,44 @@
...
+         find . -type f -name "*.xml" -exec \
+            sed -i -E 's/http:\/\/old-domain.example.com:80/https:\/\/new-domain.example.com/g' {} +
+
+5. Update database tables.
+   Note: there more than the following two tables containing the old domain name, but only these two are relevant.
+
+         UPDATE oc_assets_snapshot SET mediapackage_xml = REPLACE(
```
gregorydlogan, line 34:
```suggestion
         -- Reminder: Your old domain name may or may not have its port listed.  Use old value for org.opencastproject.server.url here.
         UPDATE oc_assets_snapshot SET mediapackage_xml = REPLACE(
         
```

PR #5869, opened 2024-05-24
Fix for event state when workflow is paused when rebuilding ES index (fixes #5868)

### LINKED ISSUE #5868: Event state is incorrectly set to finished during index rebuild when workflow is in paused state
**Describe the bug**

When the elastic search index is rebuilt, events that have the last workflow in paused state are shown as "Finished" instead of "Paused". The paused state is an active state because there's a workflow that has not finished but the event state should not be set to finished.

**To Reproduce**

1. Run a workflow that has an operation with retry-strategy="hold" and make that operation fail so that the workflow state is set to "p\Paused". The event state will also be set to "Paused".
2. Rebuild the index.
3. The event state in the ui shows "Finished".

**Expected behavior**

The event state should show "Paused"

rute-santos:
This fixes the following: when the elastic search index is rebuilt, events that have the last workflow in paused state are shown as "Finished" instead of "Paused". The paused state is an active state because there's a workflow that has not finished but the event cannot be set to finished.
Fixes #5868

### DIFF
--- a/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
+++ b/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
@@ -2213,7 +2213,9 @@ public class WorkflowServiceImpl extends AbstractIndexProducer implements Workfl
               }
               current++;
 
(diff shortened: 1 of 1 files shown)


Arnei (approved):
Works in my local Opencast and looks reasonable to me.

PR #5874, opened 2024-05-27
Fix OCR text extraction handler issue

### LINKED ISSUE #5867: Text extraction operation does not produce any output
When uploading a video as presentation, the `extract-text` operation isn't working anymore since upgrading to `OC 16`. It's marked as `succeeded` in the operation details, but as far as I can tell from the log and harvested data it doesn't actually run.

**To Reproduce**

Using the `create event` interface, upload a video (ideally one with slides) as `Slides(track "presentation/source")` and choose the `Process upon upload and schedule` workflow.
When processing has finished, inspect `opencast.log` where you should see that the text extraction didn't run. Alternatively, using the interface, you can check the operation details of the processed event which should have identical `started` and `finished` times, also indicating that the operation didn't run.

**Expected behavior**

Up to `OC 15` (on the `r/15.x` branch), the operation would in fact produce the extracted text without any prior configuration, so I would expect the same behaviour in `OC 16`.

**Server environment:**
I am running this locally as a dev.
 - OS: macOS 14.4.1
 - Distribution: allinone
 - Version OC 16

**Additional context**

- We need the ocr results in Tobira since they should be searchable as per ETH requirements.
- I have narrowed this down to recent changes in the `TextAnalysisWorkflowOperationHandler` and I'll open a PR with a small fix for this soon. This issue is just pro forma as context for that PR.

owi92:
This patch addresses a problem where the text extraction process would not run because the segment catalogs weren't loaded correctly.
The fix involves reverting some changes in the operation handler that were made a while back and are included in OC 16.

This can be tested by uploading a presentation video and inspecting the output of `TextAnalyzerService`.

Closes #5867 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/textanalyzer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/textanalyzer/TextAnalysisWorkflowOperationHandler.java
+++ b/modules/textanalyzer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/textanalyzer/TextAnalysisWorkflowOperationHandler.java
@@ -38,7 +38,6 @@ import org.opencastproject.mediapackage.MediaPackageException;
 import org.opencastproject.mediapackage.MediaPackageReference;
 import org.opencastproject.mediapackage.MediaPackageReferenceImpl;
 import org.opencastproject.mediapackage.Track;
-import org.opencastproject.mediapackage.selector.CatalogSelector;
 import org.opencastproject.metadata.mpeg7.MediaDuration;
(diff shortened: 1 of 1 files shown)


Arnei (approved):
Does fix the bug. Seems to have been caused by a difference in how `CatalogSelector` and `MediaPackage.getCatalogsByTags` get the catalogs (namely the latter will return all catalogs if handed an empty tags set as opposed to none). Personally I would still like this code to be nicer, but getting it back into working condition comes first.

PR #5913, opened 2024-06-07
Add new Workflow Operation: Subtitle Timeshift

marwyg:
This PR adds a new Workflow Operation named: 'Subtitle Timeshift'.

With this Workflow Operation we define the flavor of a subtitle track and the flavor of a video track. The workflow operation will then shift all timestamps of the subtitle file by the duration of the video.

Use Case: Currently if you put a bumper/intro video in front of an already subtitled presenter track, the captions will mismatch. You can use this workflow to shift the times of the subtitle file by the duration of the bumper/intro video and the captions will match again.

To test this, first generate subtitles with the `speech to text` operation. You can find a partial-publish workflow with the added operation in the [attachment](https://github.com/user-attachments/files/15775191/workflows.zip). Then, create a theme with a bumper video in the admin ui and create a series that uses the theme. Add the new `subtitle-timeshift` operation in your workflows that looks like this:

```
    <operation
      id="subtitle-timeshift"
      description="Create new captions with time offset">
      <configurations>
        <configuration key="subtitle-source-flavor">captions/source</configuration>
        <configuration key="video-source-flavor">branding/bumper</configuration>
        <configuration key="target-flavor">captions/shifted</configuration>
      </configurations>
    </operation>
```

But you can also use my `partial-theming`workflow, that I added to the [attachment](https://github.com/user-attachments/files/15775191/workflows.zip) and adjust it.
As the last step, create an event and don't forget to select the series. When the event is processed, the video should have the intro/bumper video in front of it and the captions should match the audio.

If you use my workflows, please always use new created events for your tests. The workflows don't consider every possible combinations of handling subtitles.

This new operation also works for trimmed captions. When you cut for example, use the schedule and upload workflow and uncheck the "straight to publish" option. When processed, cut your video and publish it. The captions should also match.


mtneug:
IMO this operation does make sense.

Probably out of scope, but I was wondering if this could be generalized for all stream types (audio, video, subtitles). I think it should be possible to shift times using FFmpeg instead of implementing this in Java. This could be helpful for cases where you separate out streams into individual tracks (e.g. when using CMAF).

marwyg:
First, I thought I take a look at the editor operation since that's where subtitles get trimmed/modified, and I figured FFmpeg might be involved. But no, turns out it’s not. Unless I missed something, the editor operation also uses Java, so I went with that approach too.

I also considered tweaking the concat workflow to automatically adjust the timing. The concat operation is what's causing the subtitles to go out of sync. But since concat is pretty generic, adding more workflow configs to it would probably just make things messy.

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
```diff
@@ -0,0 +1,53 @@
...
+| video-source-flavor    | yes      | branding/bumper  | Flavor of the video that will be used to determine the duration for the shifting of the subtitle timestamps. |
+| target-flavor          | yes      | captions/shifted | Flavor of the subtitle file that will be created including the shifted timestamps                            |
+
+
+Requirements
+------------
+
+Please select flavors that will select single/unique videos and subtitles. If multiple videos or 
+subtitles are selected, or (in other words) if it's not clear which video or subtitle file the operation has to use, 
+the operation will be skipped.
```
Arnei: Would it be possible to make this work for multiple subtitles at once? I can imagine a scenario were i have a variable number of subtitles which all need to be shifted due to the same bumper video.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
```diff
@@ -0,0 +1,291 @@
...
+      Track[] subtitleTracks = mediaPackage.getTracks(configuredSubtitleFlavor);
+      Track[] videoTracks = mediaPackage.getTracks(configuredVideoFlavor);
+      // Check if we found exactly 1 subtitle track and 1 video track
+      boolean foundTracksAreValid = validateTracks(configuredSubtitleFlavor.toString(), subtitleTracks.length,
+          configuredVideoFlavor.toString(), videoTracks.length);
+      // If we found the wrong amount, we skip the workflow operation
+      if (!foundTracksAreValid) {
+        logger.info("Skipping subtitle-timeshift workflow, because it couldn't find "
+            + "the right amount of subtitle and video tracks. Mediapackage: {}", mediaPackage.getIdentifier());
+        return createResult(mediaPackage, Action.SKIP);
```
Arnei: A wrong number of arguments indicates a problem with the workflow, therefore it might make more sense to have this operation fail to notify the user of the problem.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
```diff
@@ -0,0 +1,291 @@
...
+      } else {
+        logger.info("Valid tracks found. Start shifting subtitle track '{}' by duration from video '{}'",
+            subtitleTracks[0], videoTracks[0]);
+        originalSubtitleTrack = subtitleTracks[0];
+        videoTrack = videoTracks[0];
+      }
+    } catch (Exception e) {
+      logger.error("Unexpected error occurred in subtitle-timeshift workflow", e);
+      throw e;
```
Arnei: Should probably rethrow this as a `WorkflowOperationException`

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
```diff
@@ -0,0 +1,53 @@
...
+ID: `subtitle-timeshift`
+
+Description
+-----------
+
+The subtitle timeshift operation can be used to offset the timestamps of subtitle files. One use case regards 
+bumper/intro videos. For example: If a bumper/intro video gets added in front of an already subtitled presenter track,
+the subtitles would start too early. With this operation, you can select a video and a subtitle track and the timestamps
+of the captions file will be shifted backwards by the duration of the selected video. This ensures that the subtitles
+match the presenter track as intended.
```
Arnei, line 13: Although some documentation somewhere states that only subtitle file type truly supported by Opencast is VTT, I think it would be nice to mention that this operation only works for VTT files.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
```diff
@@ -0,0 +1,291 @@
...
+   */
+  private Track createNewTrackFromSubtitleUri(URI subtitleFile, MediaPackageElementFlavor targetFlavor, String[] tags) {
+    String id = UUID.randomUUID().toString();
+    TrackImpl track = new TrackImpl();
+    track.setIdentifier(id);
+    track.setFlavor(targetFlavor);
+    track.setURI(subtitleFile);
+    track.setTags(tags);
+    return track;
+  }
```
Arnei: Are we potentially losing information here and is that ok? What do you think about using something like `track.clone()` here?

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
```diff
@@ -0,0 +1,291 @@
...
+          videoFlavor);
+      return false;
+    }
+    return true;
+  }
+
+
+}
+
+
```
Arnei: While it's good to end a file with a newline, one is usually enough.

modules/subtitle-parser/src/main/java/org/opencastproject/subtitleparser/webvttparser/WebVTTSubtitle.java
```diff
@@ -63,4 +63,20 @@ public void addStyle(WebVTTSubtitleStyle style) {
...
   }
+
+  /**
+   * Shifts all timestamps in the Subtitle file by a given amount.
+   * For example: If you put a bumper video in front of an already subtitled presenter track,
+   * you simply can shift the timestamps of the subtitle file and use the duration of the bumper video as the time.
+   *
+   * @param timeInMilliseconds Time in milliseconds by which all timestamps shall be shifted.
+   */
+  public void shiftTimeBy(long timeInMilliseconds) {
```
Arnei: This very well might just be me, but I feel like this does not quite belong here? Like, this is not directly related to WebVTT structure or parsing and would make more sense as a function in the WOH?

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
```diff
@@ -0,0 +1,53 @@
...
+------------
+The operation will create a new subtitle file in the process and will not override the old one. 
+The tags of the original subtitle file will be copied to the new subtitle file.
+
+
+Operation Examples
+------------------
+
+```XML
+<operation
```
Arnei, line 44: As the community workflows contain both subtitles and theming, would it be reasonable/possible to add this to the community workflows?

Arnei (changes requested):

Arnei:
Did some local testing and it worked as advertised!

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
marwyg: Thank you for the review! I am currently working on the code.

This part you mentioning here is not for checking the amount of arguments, rather it counts how many tracks were found with the configured flavor. If someone configures `*/source` as the `video-source-flavor` for example, the operation will find 2 tracks (in the case of a dual stream) and the operation expects 1 video. It would be no Problem to throw an error at this point. What would you suggest for the following cases?:

1. The operation finds no video with the given `video-source-flavor`: throw an error?
2. The operation finds more than 1 video: throw an error?
Doing it like this would be okay for me. So what we saying then is: "if the operation doesn't find exactly 1 video with the configured flavor, the user did not configured the workflow operation correctly and should fix it".

I am also changing the behavior of what the operations does, when it finds a certain amount of subtitle files:

1. The operation finds no subtitle file: simply skip operation?
2. The operation finds >= 1 subtitle files: timeshift all of them?
Currently the operation will skip if it finds more than 1 subtitle file, but you suggested supporting the shifting of multiple files at once (this would be the case for example when there are more languages involved). I think you are right and I will implement it this way.

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
marwyg: The operation can now shift multiple transcriptions.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
marwyg: I pushed a new version. The Operation will now fail, when more than 1 video track was found.

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
marwyg: I added this.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
marwyg: I am using clone() now.

modules/subtitle-parser/src/main/java/org/opencastproject/subtitleparser/webvttparser/WebVTTSubtitle.java
marwyg: I moved the code to the woh.

docs/guides/admin/docs/workflowoperationhandlers/subtitle-timeshift-woh.md
marwyg: I made some changes to the partial-publish workflow and the partial-theming workflow so they work now with the new timeshift operation.

marwyg:
Thank you again for the review @Arnei. I made several changes and tested everything as much as I can. I commented on your change requests. One is still open if you want to add something, but overall I changed everything you suggested.

modules/subtitle-timeshift-workflowoperation/src/main/java/org/opencastproject/workflow/handler/subtitletimeshift/SubtitleTimeshiftWorkflowOperationHandler.java
Arnei: That makes sense to me.

Arnei (approved):
My concerns have been addressed and this still seems to work fine.

PR #5932, opened 2024-06-12
Fix link in editor config file

Arnei:
This change was part of an older editor release and was overlooked (likely by me) when merging editor releases into Opencast. So here it is.


lkiesow (approved):
I call this documentation. Merging…

PR #5937, opened 2024-06-14
Fix Non-Admin Access to New Admin UI

KatrinIhler:
Some static files were moved around for the new Admin UI and we need to reflect these changes in our security config. Without this, a non-admin user can't see the new Admin UI despite having all the UI roles, just a blank page. This is easily reproduced on stable.opencast.org (but doesn't occur on legacy.opencast.org).

I couldn't find the exact PR responsible for this in the new admin interface repo, so maybe somebody who knows more can check if a blanket permission for `admin-ui/assets` is reasonable.

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -276,9 +276,7 @@
     <sec:intercept-url pattern="/admin-ng/shared/**" access="ROLE_ANONYMOUS" />
     <sec:intercept-url pattern="/admin-ng/styles/**" access="ROLE_ANONYMOUS" />
 
-    <sec:intercept-url pattern="/admin-ui/static/media/**" access="ROLE_ANONYMOUS" />
-    <sec:intercept-url pattern="/admin-ui/static/js/**" access="ROLE_ANONYMOUS" />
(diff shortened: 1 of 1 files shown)


Arnei:
But can this go into r/15.x? Or is r/15.x fine as you said "doesn't occur on legacy.opencast.org"?

KatrinIhler:
This can't go into 15.x because there the files still reside in the old place.

Arnei (approved):
With this PR, my local user with only `ROLE_ADMIN_UI` and related `UI` roles can open the admin ui again.

As an aside, I am getting a lot of 
```
http://localhost:8080/services/health.json
[HTTP/1.1 403 Forbidden 3ms]
```
in the web console with my local user, not sure if this related though.

mtneug:
> As an aside, I am getting a lot of
> 
> ```
> http://localhost:8080/services/health.json
> [HTTP/1.1 403 Forbidden 3ms]
> ```
> 
> in the web console with my local user, not sure if this related though.

This is "normal" and was also the case for the old admin UI. This endpoint is for the health icon at the top. Since your user doesn't have access rights for this endpoint, you get a 403. The UI will then just not display this information. I'm not sure if it makes sense to complicate the logic to only try once etc.

PR #5863, opened 2024-05-23
Adjust `created` field of Tobira harvest response for events

owi92:
This changed in OC 16 with the Solr/Opensearch swap as far as I can see, but it needs to be in a specific format in order to be harvested by Tobira. (Edit: I am referring to this commit: https://github.com/opencast/opencast/commit/5880526444317ea9396c97f4fa876480b83cd09d)

No version bump since this only restores old behaviour and is not a breaking change.

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #5939, opened 2024-06-18
Add `SearchResult::getCreatedDate` and use it to fix Tobira harvest API

LukasKalbertodt:
Before 16, the search service already returned a `Date` for `created`. When we switched to ElasticSearch, this was turned into a formatted String instead, which first went unnoticed. When noticed, it was fixed by calling `Instant.parse` which works for some cases, but not all.

The date format returned by ES is not always the same. When adding documents to ES, it will parse fields declared as dates and store the date as "ms since epoch UTC" internally to be able to filter by it. But it also remembers the original format and does not format the ms in a well defined format. The formats that can be parsed by ES can be configured, but by default (our case) it's
"strict_date_optional_time||epoch_millis". Documents with invalid dates will lead to an exception, i.e. will not be added to the index. So when querying the index, we have to expect any date format that can be parsed by the above default format. Compare [1].

And well, that's what this PR does: use the correct method from ES to parse it. I added the method to `SearchResult` as I figured that might be useful in the future.

[1]: https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html

This is a fix for this fix: #5863 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
+++ b/modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
@@ -35,6 +35,8 @@ import org.opencastproject.security.api.AccessControlList;
 
 import com.google.gson.Gson;
 
+import org.elasticsearch.index.mapper.DateFieldMapper;
+
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


mtneug:
Is this an issue only for internal Java use or also for all clients of the search API including external ones? I.e. is the date in the JSON response consistent, and if not, should it be?

LukasKalbertodt:
The `created` field in the response of the search API is also *not* consistent. For example, it can return `"2018-04-21T13:14:31Z"` as well as `"2024-06-18T00:04Z"` (these two examples are returned by stable.opencast.org right now; it seems that ingesting via ingest API will result in a different date format than uploading via admin UI).

Should they be? Probably yes. Otherwise all downstream clients have to deal with that. As I mentioned before, I would strongly prefer if OC would use more "checked" fields for important data, instead of saying "could be anything" because all data comes from an ingested XML blob. 
So should I just fix that as well and make the field in the response (`dc.created`) consistent? Or do we want to open a bigger discussions of whether `created` should maybe be a top level field like `updated`?

PR #5944, opened 2024-06-20
Fix Opencast startup failure when index service boots later

### LINKED ISSUE #5943: Opencast breaks if booted before Elastic-/OpenSearch
**Describe the bug**

When starting Opencast before Elastic-/OpenSearch, multiple errors occur, and Opencast fails to start even if the index service starts afterwards. This issue was reported following a power failure at an institution.

**To Reproduce**

1. Start Opencast without Elastic-/OpenSearch running.
2. Observe that Opencast fails to start (check the logs or attempt to log in).
3. Start Elastic-/OpenSearch.
4. Notice that Opencast still doesn't start properly.

**Expected behavior**

1. Start Opencast (index service didn't boot yet).
2. Opencast detects the missing connection to the index service and waits.
3. Elastic-/OpenSearch starts.
4. Opencast connects to Elastic-/OpenSearch and completes its startup process.

marwyg:
Fixes #5943

This PR adds a check in the Elasticsearch initialization method. The check includes sending a request to the index service and evaluating the response. If the service is not reachable, a retry loop will continuously check the connection until it succeeds.

**Changes:**

- Added a connection check in the Elasticsearch init method.
- Implemented a retry loop with a configurable wait period.

**How to test:**

1. Start Opencast without Elastic-/OpenSearch running.
2. Check the logs. Opencast should log every 10 seconds that it is retrying to connect to the index service.
3. Start the index service.
4. Opencast should continue with the startup process once the index service is available.

If steps 2 and 4 are successful, the test is complete.

**New Configuration:**

A new configuration value has been added to `etc/org.opencastproject.elasticsearch.index.ElasticsearchIndex.cfg`:

- retry.delay.on.startup: This sets the interval (in milliseconds) between retry attempts.

**Testing the New Configuration:**

1. Change the value of retry.delay.on.startup, for example, to 1 second (1000 milliseconds).
2. Start Opencast without Elastic-/OpenSearch running.
3. Check the logs and verify that the time between retry log messages matches the configured value.

### DIFF
--- a/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
+++ b/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
@@ -34,6 +34,7 @@ import org.opencastproject.util.requests.SortCriterion;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 import org.apache.http.HttpHost;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
@@ -82,6 +83,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.ConnectException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -166,6 +168,11 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
   /** Password of an external Elasticsearch server to connect to. */
   private String password;
 
+  /** Defines how long to wait between retries, when the connection to OpenSearch failed on startup */
+  private int retryDelayOnStartup;
+  private static final String RETRY_DELAY_ON_STARTUP = "retry.delay.on.startup";
+  private static final int DEFAULT_RETRY_DELAY_ON_STARTUP = 10000;
+
   /**
    * Returns an array of document types for the index. For every one of these, the corresponding document type
    * definition will be loaded.
@@ -216,6 +223,14 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
     indexName = StringUtils.defaultIfBlank((String) properties.get(INDEX_NAME_PROPERTY),
             DEFAULT_INDEX_NAME);
     logger.info("Index name set to {}.", indexName);
+
+    retryDelayOnStartup = NumberUtils.toInt((String) properties.get(RETRY_DELAY_ON_STARTUP),
+        DEFAULT_RETRY_DELAY_ON_STARTUP);
+    if (retryDelayOnStartup <= 0) {
+      throw new IllegalArgumentException(RETRY_DELAY_ON_STARTUP
+          + " was wrongly configured. Value has to be greater than 0.");
+    }
+    logger.info("Retry delay on startup set to {} ms.", retryDelayOnStartup);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


mtneug:
As a general remark (I'm fine with this going in): IMO the start ordering problem should usually be solved by the systems that start Opencast (systemd, Kubernetes, etc.). I like the approach to simply fail and exit with an error code. Hopefully, the dependencies will be fulfilled after the application gets restarted. In this way, the order is not explicitly coded and the system should eventually start up. However, this is tricky with Opencast / OSGi (e.g. failing all modules; long startup times).

I, therefore, implemented the same try-and-sleep mechanism for the container images (see for the [database](https://github.com/opencast/opencast-docker/blob/master/rootfs/opencast/docker/scripts/TryToConnectToDb.java) and [Elasticsearch](https://github.com/opencast/opencast-docker/blob/master/rootfs/opencast/docker/scripts/elasticsearch.sh#L53-L83)).

etc/custom.properties
```diff
@@ -195,11 +195,11 @@ org.opencastproject.working.file.repository.cleanup.collections=failed.zips
 
 # The username of Elasticsearch for Opencast to use.
 # Default: None
-#org.opencastproject.elasticsearch.username=
+org.opencastproject.elasticsearch.username=admin
 
 # The password of Elasticsearch for Opencast to use.
 # Default: None
-#org.opencastproject.elasticsearch.password=
+org.opencastproject.elasticsearch.password=qwe123
```
lkiesow: I don't think you wanted to commit these changes, did you?

modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
```diff
@@ -414,6 +426,17 @@ protected void init(int version)
       client = new RestHighLevelClient(builder);
     }
 
+    String indexServiceUrl = getIndexServiceUrl();
+    while (!isIndexServiceReachable(indexServiceUrl)) {
+      logger.warn("Could not reach Elasticsearch at {}. Trying again after few seconds...", indexServiceUrl);
```
lkiesow: Since we want people to use OIpenSearch?
```suggestion
      logger.warn("Could not reach OpenSearch at {}. Trying again after few seconds...", indexServiceUrl);
```

modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
```diff
@@ -760,4 +792,24 @@ public List<String> getTermsForField(String field, String type) {
...
+   */
+  private boolean isIndexServiceReachable(String elasticSearchUrl) {
+    try {
+      URL url = new URL(elasticSearchUrl);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod("HEAD");
+      connection.setConnectTimeout(3000);
+      connection.setReadTimeout(3000);
+      int responseCode = connection.getResponseCode();
+      return (200 <= responseCode && responseCode <= 399);
```
lkiesow: - Why try doing an HTTP request and not use the library to try connecting to the search index?
- A 3xx response code doesn't really ensure that the connection would work, would it?

lkiesow: Also, does this work with required authentication?

marwyg:
Thank you for the review!

- I removed the custom.properties file.
- I changed my code so it says OpenSearch instead of ElasticSearch.
- I am using the `client` variable (instance of RestHighLevelClient) to send the request now instead of writing my own HTTP request by hand.

The code should work with authenticated requests too, but I didn't test it, because I wasn't able to set it up yet. If someone has an opensearch running with authentication and could test this, I really would appreciate it. Otherwise (and if you want me to) I take another try to set it up for testing. But in general, when someone is not authenticated correctly, an ElasticsearchException should be thrown regarding the documentation, so I added a catch block to handle this scenario: https://github.com/opencast/opencast/pull/5944/files#diff-cff431f7b119ce82720423f6e745e1fcf22c4005aad5dba8decdf2c2fc21513fR477.

Arnei (commented):
Works and code looks reasonable to me.

When spinning up Elasticsearch after Opencast (via docker), the Opencast logs get spammed with error messages while Elasticsearch is still starting. The error messages are the same ones you would get prior to this PR when starting Opencast without a running Elasticsearch, they eventually stop and everything runs fine. Still, do you think it is possible to avoid having the log spammed with error messages in this case?

marwyg:
I'll take a look. It would be nice if the log doesn't get spammed.

gregorydlogan:
Doing the same thing as @Arnei, Opencast's log stops scrolling, the admin UI looks like it works, but attempting to ingest anything gets me a 404 or a 405 on lots of endpoints (ingest, and search notably).

Looking in `bundle:list`, the `elasticsearch-index` bundle is marked as `Waiting`, which explains why nothing's working :D

KatrinIhler:

KatrinIhler:
Since we have a new version of this one, this PR can be closed, no?

PR #5946, opened 2024-06-20
LTI translation strings

lkiesow:
The LTI tool was always a bit broken in the sense that it was copying internal translations from the admin interface with no check if they are actually valid or if they changed. This was always dangerous and could easily break but will now definitely break when we remove the old admin interface.

That is why this patch is finally transferring the relevant translation keys to the LTI module itself.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
As discussed in the technical meeting, I moved this to `develop` to avoid broken translations in Opencast 16

PR #5955, opened 2024-06-25
Refactor Activator.java to improve database connection handling

### LINKED ISSUE #5952: Make Opencast Wait on Database during Startup
### Description:

When starting Opencast, the application crashes if the database service is not running. Ideally, Opencast should handle the absence of the database gracefully, possibly by retrying the connection or providing a meaningful error message without crashing. See similar issue with elastic search here: https://github.com/opencast/opencast/issues/5943 This issue was reported following a power failure at an institution.

### Steps to Reproduce:

1. Ensure the database service required by Opencast is stopped or not running.
2. Attempt to start the Opencast application.
3. Observe that Opencast crashes immediately after startup.

### Expected Behavior:
Opencast should not crash if the database service is not available. Instead, it should:

1. Log a meaningful error message indicating that the database connection could not be established.
2. Retry the connection to the database after a certain interval.
3. (Optional?) Allow for a configuration option to set the retry interval and the number of retry attempts.

### Additional Information:
This issue affects the robustness and reliability of the Opencast application. Implementing a more fault-tolerant startup process would improve the user experience and operational stability 🚀

marwyg:
Fixes #5952

### Description

Implemented a wait mechanism to ensure database availability before starting the application. This resolves issues where the application fails to start if the database is not yet available.

### To test this:

1. Make sure the database is not running
2. Start opencast
3. Observe the log, you should see, that Opencast retries to connect to the database
4. Start the Database
5. Observe the log, Opencast should continue the startup process without errors

### Test Scenario 2:

I also included a check for credentials, to test this, do the following:

1. Configure a wrong username or password in the `custom.properties` file
2. Make sure the database is not running
3. Start opencast
4. Observe the log: The startup of Opencast should fail with clear logging

### DIFF
--- a/modules/db/src/main/java/org/opencastproject/db/Activator.java
+++ b/modules/db/src/main/java/org/opencastproject/db/Activator.java
@@ -217,9 +217,8 @@ public class Activator implements BundleActivator {
           connectionEstablished = connection != null && !connection.isClosed();
         } catch (SQLInvalidAuthorizationSpecException e) {
           // Invalid credentials
-          String errorMsg = "Couldn't connect to database. Probably invalid database credentials. Check you config.";
-          logger.error(errorMsg);
-          throw new RuntimeException(errorMsg, e);
+          throw new RuntimeException(
+              "Couldn't connect to database. Probably invalid database credentials. " + "Check you config.", e);
         } catch (SQLNonTransientConnectionException e) {
           // Connection failed, database is not reachable (yet? maybe its just starting up, so we wait)
           logger.error("Database connection attempt failed, message: {}", e.getMessage());
--- a/modules/db/src/main/java/org/opencastproject/db/Activator.java
+++ b/modules/db/src/main/java/org/opencastproject/db/Activator.java
@@ -32,8 +32,11 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.sql.Connection;
+import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLInvalidAuthorizationSpecException;
+import java.sql.SQLNonTransientConnectionException;
 import java.sql.Statement;
 import java.util.Hashtable;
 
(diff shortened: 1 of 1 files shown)


modules/db/src/main/java/org/opencastproject/db/Activator.java
```diff
@@ -198,6 +191,50 @@ public void start(BundleContext bundleContext) throws Exception {
...
+   */
+  private void waitUntilDatabaseIsAvailable(String jdbcDriver, String jdbcUrl, String jdbcUser, String jdbcPass) {
+    logger.info("Testing connectivity to database at {}", jdbcUrl);
+
+    // Load the JDBC driver for our basic connection test
+    try {
+      Class.forName(jdbcDriver);
+    } catch (ClassNotFoundException e) {
+      logger.error("Couldn't load JDBC Driver {} for initial connection test", jdbcDriver);
+      throw new RuntimeException(e);
```
lkiesow, line 210: This is different from the connection settings we use above, Wouldn't it be better to use the same? In fact, couldn't you just put a loop around the `try-catch` block you removed above and modify the ctach block?

modules/db/src/main/java/org/opencastproject/db/Activator.java
```diff
@@ -198,6 +191,50 @@ public void start(BundleContext bundleContext) throws Exception {
...
+    boolean connectionEstablished = false;
+    try {
+      while (!connectionEstablished) {
+        try (Connection connection = DriverManager.getConnection(jdbcUrl, jdbcUser, jdbcPass)) {
+          connectionEstablished = connection != null && !connection.isClosed();
+        } catch (SQLInvalidAuthorizationSpecException e) {
+          // Invalid credentials
+          String errorMsg = "Couldn't connect to database. Probably invalid database credentials. Check you config.";
+          logger.error(errorMsg);
+          throw new RuntimeException(errorMsg, e);
```
lkiesow: It may seem minor, but it's actually pretty annoying to log errors and then throw them. The thrown error will either be caught and handled properly, turning the error into a non-error, or it will stay an error and be logged again, causing the error to appear at two separate places in the logs.

So, just throw the error and do not log:
```suggestion
          throw new RuntimeException("Couldn't connect to database. Probably invalid database credentials. Check you config.";, e);
```

lkiesow (changes requested):
Overall, this looks good. I'm just not entirely shure about the connection method

modules/db/src/main/java/org/opencastproject/db/Activator.java
marwyg: I removed it.

modules/db/src/main/java/org/opencastproject/db/Activator.java
marwyg: Thank you for the review! Creating a loop around the original connection method was my first approach. However, I chose a different method for connecting to the database because the `pooledDataSource` logs exceptions upon failure. This alternative approach helps maintain a cleaner log. But we can use the `pooledDataSource` if the exceptions are acceptable.

Arnei (approved):
Works with my local setup and seems reasonable to me. Agree with Lars that we optimally would use the exact same connection settings, but since this only wants to be a simple a connection check, this is fine I think?

gregorydlogan (approved):
Works as expected, although I agree with Lars that it'd be better to log more and reuse the sensitive connection logic.

This retry code really shouldn't be needed most of the time, so if the exceptions are the issue then the adopter should probably fix their init system.

KatrinIhler:

@marwyg  @lkiesow Can you come to an agreement here on the last outstanding comment?

geichelberger (changes requested):
Wrapping an exception should add a value by throwing a more specific exception. The RuntimeException does not add this value, and because of the nested try-catch, the original exception is nested two times in a RuntimeException.

gregorydlogan:
Is this going to be moving forward in time for OC18?

PR #5959, opened 2024-06-26
Fix log format strings

JulianKniephoff:
This PR brought to you by IntelliJ inspections.
Most of these are due to misunderstanding the special handling of `Throwable` parameters
by SLF4J, but there are a few more subtle ones.
Also, a surprising number of instances that seem to thing SLF4J uses "`printf`-format."


lkiesow (approved):
Looks good

PR #5970, opened 2024-06-28
Display state mappings in the status column again

Arnei:
The events table in the admin ui has a status column that informs users about the current state of the latest workflow. A workflow can define "state mappings" to overwrite what is usually displayed. For example, a delete workflow may overwrite the usual "Running" with "Deleting". This has probably been broken since Opencast 12-ish, and this commit should fix it.

Example:
Bildschirmfoto vom 2024-06-28 14-59-10

Fixes https://github.com/opencast/opencast-admin-interface/issues/402.


lkiesow (approved):
Works as expected, although it has some weird side effects (see #5996)

PR #5971, opened 2024-06-28
No Playlists on Presentation

lkiesow:
The playlist service requires the index service which is not present on presentation nodes, causing the module to fail. This just removes playlist from the node.

That's not great and ideally, the service shouldn't require the index service but this is a quick fix to the immediate problem.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/assemblies/karaf-features/src/main/feature/feature.xml
+++ b/assemblies/karaf-features/src/main/feature/feature.xml
@@ -363,6 +363,7 @@
     <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-api/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-remote/${project.version}</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-timelinepreviews-workflowoperation/${project.version}</bundle>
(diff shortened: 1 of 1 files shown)


geichelberger:
~~The Tobira Integration needs the playlist stuff on the presentation node.~~

Between writing and posting, the tobira endpoint moved to the admin node, which I would not expect in a minor version.

KatrinIhler:
I think this might be nicer: https://github.com/opencast/opencast/pull/5972

lkiesow:
> Between writing and posting, the tobira endpoint moved to the admin node, which I would not expect in a minor version.

That shouldn't matter since the integration is broken in 16.0 anyway. That means no one will be surprised at this suddenly not working and people have to take a look at this anyway.

But we also have Katrin's suggestion which may be preferable.

KatrinIhler:
> That shouldn't matter since the integration is broken in 16.0 anyway.

That's true, but i think Lukas has already fixed that for 16.1. So it would be nice if Tobira was working.

lkiesow:
Tobira would be working. You just needed to adapt the configuration. The Tobira endpoint is just moving.
But we have another solution. So, I'm closing this in favor of #5972

PR #5972, opened 2024-06-28
Fix distributed OC 16

KatrinIhler:
...by using the index directly, which we already depend on, instead of the index service. Not so pretty, but works. 
Bonus: No functional library!

### DIFF
--- a/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
+++ b/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
@@ -408,26 +408,25 @@ public class PlaylistService {
    * itself.
    */
   public JaxbPlaylist enrich(Playlist playlist) {
-    JaxbPlaylist jaxbPlaylist = new JaxbPlaylist(playlist);
+    var jaxbPlaylist = new JaxbPlaylist(playlist);
+    var org = securityService.getOrganization().getId();
+    var user = securityService.getUser();
 
     // Add additional infos about events
     List<JaxbPlaylistEntry> jaxbPlaylistEntries = jaxbPlaylist.getEntries();
-    for (int i = 0; i < jaxbPlaylistEntries.size(); i++) {
+    for (JaxbPlaylistEntry entry : jaxbPlaylistEntries) {
       try {
-        JaxbPlaylistEntry entry = jaxbPlaylistEntries.get(i);
         if (entry.getType() == PlaylistEntryType.EVENT) {
 
           // We only get an event from the index if we have permission to do so (and if it exists ofc)
-          SearchResult<Event> result = elasticsearchIndex.getByQuery(new EventSearchQuery(
-                  securityService.getOrganization().getId(), securityService.getUser())
-                  .withIdentifier(jaxbPlaylistEntries.get(i).getContentId()));
+          SearchResult<Event> result = elasticsearchIndex.getByQuery(
+              new EventSearchQuery(org, user).withIdentifier(entry.getContentId()));
           if (result.getPageSize() != 0) {
             Event event = result.getItems()[0].getSource();
             entry.setPublications(event.getPublications());
           } else {
             entry.setType(PlaylistEntryType.INACCESSIBLE);
           }
-          jaxbPlaylistEntries.set(i, entry);
         }
       } catch (SearchIndexException e) {
         throw new RuntimeException(e);
--- a/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
+++ b/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
@@ -23,9 +23,10 @@ package org.opencastproject.playlists;
 import static org.opencastproject.security.api.SecurityConstants.GLOBAL_ADMIN_ROLE;
 
 import org.opencastproject.elasticsearch.api.SearchIndexException;
+import org.opencastproject.elasticsearch.api.SearchResult;
 import org.opencastproject.elasticsearch.index.ElasticsearchIndex;
 import org.opencastproject.elasticsearch.index.objects.event.Event;
-import org.opencastproject.index.service.api.IndexService;
+import org.opencastproject.elasticsearch.index.objects.event.EventSearchQuery;
 import org.opencastproject.playlists.persistence.PlaylistDatabaseException;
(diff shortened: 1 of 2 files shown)


modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
```diff
@@ -423,10 +416,13 @@ public JaxbPlaylist enrich(Playlist playlist) {
       try {
         JaxbPlaylistEntry entry = jaxbPlaylistEntries.get(i);
         if (entry.getType() == PlaylistEntryType.EVENT) {
-          Opt<Event> optEvent = indexService.getEvent(jaxbPlaylistEntries.get(i).getContentId(), elasticsearchIndex);
-          // We only get an event from the indexService if we have permission to do so (and if it exists ofc)
-          if (optEvent.isSome()) {
-            Event event = optEvent.get();
+
+          // We only get an event from the index if we have permission to do so (and if it exists ofc)
```
gregorydlogan, line 421: Is this actually true? We're talking directly to the index right? The index has the acl, but doesn't know what to do with it. It can't check permissions like this, you have to do that.

gregorydlogan (commented):
If I'm wrong about this feel free to disregard. Code looks good assuming it works like Katrin thinks it does 😅

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
KatrinIhler: As discussed in chat, the ES index does actually do permission checking, I think. Either way, the index service didn't do much more, just retrieve the event from the index.

lkiesow: At least, the permissions are stored in the index. I'll take a look later.

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
```diff
@@ -41,7 +42,6 @@
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.requests.SortCriterion;
 
-import com.entwinemedia.fn.data.Opt;
```
lkiesow, line 44: Extra credit for removing Entwine Fn code :tada:

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
```diff
@@ -423,10 +416,13 @@ public JaxbPlaylist enrich(Playlist playlist) {
       try {
         JaxbPlaylistEntry entry = jaxbPlaylistEntries.get(i);
         if (entry.getType() == PlaylistEntryType.EVENT) {
-          Opt<Event> optEvent = indexService.getEvent(jaxbPlaylistEntries.get(i).getContentId(), elasticsearchIndex);
-          // We only get an event from the indexService if we have permission to do so (and if it exists ofc)
-          if (optEvent.isSome()) {
-            Event event = optEvent.get();
+
+          // We only get an event from the index if we have permission to do so (and if it exists ofc)
+          SearchResult<Event> result = elasticsearchIndex.getByQuery(new EventSearchQuery(
```
lkiesow: Works great and I will merge it like this if you want me to. But while the following suggestion is purely cosmetic, I think this makes the code easier to read while avoiding a few unnecessary operations:

```diff
diff --git a/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java b/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
index 33e8e1c142..e927e12a17 100644
--- a/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
+++ b/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
@@ -408,26 +408,25 @@ public class PlaylistService {
    * itself.
    */
   public JaxbPlaylist enrich(Playlist playlist) {
-    JaxbPlaylist jaxbPlaylist = new JaxbPlaylist(playlist);
+    var jaxbPlaylist = new JaxbPlaylist(playlist);
+    var org = securityService.getOrganization().getId();
+    var user = securityService.getUser();
 
     // Add additional infos about events
     List<JaxbPlaylistEntry> jaxbPlaylistEntries = jaxbPlaylist.getEntries();
-    for (int i = 0; i < jaxbPlaylistEntries.size(); i++) {
+    for (JaxbPlaylistEntry entry : jaxbPlaylistEntries) {
       try {
-        JaxbPlaylistEntry entry = jaxbPlaylistEntries.get(i);
         if (entry.getType() == PlaylistEntryType.EVENT) {
 
           // We only get an event from the index if we have permission to do so (and if it exists ofc)
-          SearchResult<Event> result = elasticsearchIndex.getByQuery(new EventSearchQuery(
-                  securityService.getOrganization().getId(), securityService.getUser())
-                  .withIdentifier(jaxbPlaylistEntries.get(i).getContentId()));
+          SearchResult<Event> result = elasticsearchIndex.getByQuery(
+              new EventSearchQuery(org, user).withIdentifier(entry.getContentId()));
           if (result.getPageSize() != 0) {
             Event event = result.getItems()[0].getSource();
             entry.setPublications(event.getPublications());
           } else {
             entry.setType(PlaylistEntryType.INACCESSIBLE);
           }
-          jaxbPlaylistEntries.set(i, entry);
         }
       } catch (SearchIndexException e) {
         throw new RuntimeException(e);
```

lkiesow (approved):
Works. Also checks permission.
I only have a cosmetic suggestion.

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
lkiesow: Permissions are checked.

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
KatrinIhler: I know, right? :heart:

modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
KatrinIhler: Agreed, this looks nicer and is more efficient.

PR #6006, opened 2024-07-04
Update karaf version to 4.4.6

geichelberger:
This updates Karaf from version 4.4.4 to version 4.4.6 and updates the following dependencies:

- commons-compress 1.26.1
- commons-lang3 3.14.0
- commons-logging 1.3.1

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/db/pom.xml
+++ b/modules/db/pom.xml
@@ -115,6 +115,7 @@
               !software.amazon.awssdk.*,
               !waffle.*,
               !org.codehaus.groovy*,
+              !org.locationtech.jts.geom.*,
               org.w3c.dom;version=0,
               *
             </Import-Package>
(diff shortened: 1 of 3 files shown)


lkiesow (approved):
Seems to work as expected.
Any reason to update `r/16.x` and not just `develop`?

gregorydlogan:
Regardless of whether this makes 16, it should definitely go into develop sooner than later since this may affect the Spring upgrade.

geichelberger:
>  Any reason to update r/16.x and not just develop?
Developer Joy, because the next version, 4.4.7, will contain the EclisepLink upgrade, allowing developers to switch between branches without funny stack traces while using Java 21.

Karaf 4.4.5 fixes a CVE regarding jetty and http2; the feature is not enabled by default, and maybe no opencast adopter has enabled it.

lkiesow:
Fine. I'll merge it. In general, I always think that this is the type of patch that can easily break something we don't notice since far less testing is going to happen before this ends up in a release and this means that adopters which just want a minor update to fix some bugs are effected.

PR #6008, opened 2024-07-05
Add basic date validation check to ingest/addMediaPackage/{wfId}

### LINKED ISSUE #5954: Ingest API returns 500 on invalid date
Running this:

```
curl -f -i -u admin:opencast https://stable.opencast.org/ingest/addMediaPackage/fast \
    -F mediaUri='https://nextcloud.elan-ev.de/index.php/s/8ESyZAMkSoiFEwr/download' \
    -F title="Text" \
    -F created="banana" \
    -F acl='{"acl": {"ace": [{"allow": true,"role": "ROLE_ANONYMOUS","action": "read"},{"allow": true,"role": "ROLE_ADMIN","action": "write"}]}}'
```

Will return 500. Fixing the `created` field to a proper date fixes that.

**Server environment:**
 - Version 16.x

Arnei:
Attempts to fix #5954, by parsing date fields in the IngestRestService method and returning 400 if an exception occurs.

Arguably there should be a full on metadata validation function we could make use of here, but I could not find one and did not feel like writing one myself.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
+++ b/modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
@@ -23,6 +23,10 @@ package org.opencastproject.ingest.endpoint;
 
 import static org.apache.commons.lang3.StringUtils.trimToNull;
 import static org.opencastproject.mediapackage.MediaPackageElements.XACML_POLICY_EPISODE;
+import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_CREATED;
+import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_DATE;
+import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_IDENTIFIER;
+import static org.opencastproject.metadata.dublincore.DublinCore.PROPERTY_TEMPORAL;
 
 import org.opencastproject.authorization.xacml.XACMLUtils;
 import org.opencastproject.capture.CaptureParameters;
@@ -767,7 +771,9 @@ public class IngestRestService extends AbstractJobProducerEndpoint {
             String fieldName = item.getFieldName();
             String value = Streams.asString(item.openStream(), StandardCharsets.UTF_8.toString());
             if (dcterms.contains(fieldName)) {
-              if ("created".equals(fieldName) || "date".equals(fieldName) || "temporal".equals(fieldName)) {
+              if (PROPERTY_CREATED.getLocalName().equals(fieldName)
+                  || PROPERTY_DATE.getLocalName().equals(fieldName)
+                  || PROPERTY_TEMPORAL.getLocalName().equals(fieldName)) {
                 try {
                   OpencastMetadataCodec.decodeDate(value);
                 } catch (IllegalArgumentException e) {
@@ -802,7 +808,7 @@ public class IngestRestService extends AbstractJobProducerEndpoint {
               tags.add(value);
               /* Fields for DC catalog */
             } else if (dcterms.contains(fieldName)) {
-              if ("identifier".equals(fieldName)) {
+              if (PROPERTY_IDENTIFIER.getLocalName().equals(fieldName)) {
                 /* Use the identifier for the mediapackage */
                 mp.setIdentifier(new IdImpl(value));
               }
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
Fixes the issue at hand, i.e. that particular `curl` command results in 400 now. I lack the experience to say whether `OpencastMetadataCodec.decodeDate` is exactly the correct function to use here and whether that restricts the allowed date formats in some way (i.e. breaking something that previously worked). But it seems fine to me.

LukasKalbertodt (changes requested):
Oh woops, I just noticed something now: this validation should happen at the very start before doing anything else. The way it is done now, I get log output about "added track" and stuff like that.

Arnei:
> Oh woops, I just noticed something now: this validation should happen at the very start before doing anything else. The way it is done now, I get log output about "added track" and stuff like that.

I guess I could move the code into it's own loop that runs prior to the loop where stuff is added. On the other hand, this won't prevent the situation you described in other cases. For example, there could be a log message about successfully adding an ACL catalog, but adding an ill-configured media file after that results in a bad request. So we can't really completely prevent this "weird" logging.

LukasKalbertodt:
I'm actually less concerned about the logging and more about what these log statements mean in terms of actions that are not cleaned up in case of a bad request. While you are correct and I wouldn't expect you (in this PR) to fix the other things as well, I think it's a good idea to add a second loop now, so that over time, these checks can move there. Or do it in some other way that's a good step towards "not doing anything until the whole request is checked".

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -759,6 +760,24 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
       int episodeDCCatalogNumber = 0;
       boolean hasMedia = false;
       if (ServletFileUpload.isMultipartContent(request)) {
+        // Validate first
+        for (FileItemIterator iter = new ServletFileUpload().getItemIterator(request); iter.hasNext();) {
+          FileItemStream item = iter.next();
+          if (item.isFormField()) {
+            String fieldName = item.getFieldName();
+            String value = Streams.asString(item.openStream(), "UTF-8");
```
geichelberger:
```suggestion
            String value = Streams.asString(item.openStream(), StandardCharsets.UTF_8);
```

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -759,6 +760,24 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
...
+            if (dcterms.contains(fieldName)) {
+              if ("created".equals(fieldName) || "date".equals(fieldName) || "temporal".equals(fieldName)) {
```
geichelberger: The Dublin Core properties are defined in that interface:
https://github.com/opencast/opencast/blob/develop/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCore.java

geichelberger (changes requested):

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
```diff
@@ -759,11 +764,31 @@ public Response addMediaPackage(@Context HttpServletRequest request, @PathParam(
       int episodeDCCatalogNumber = 0;
       boolean hasMedia = false;
       if (ServletFileUpload.isMultipartContent(request)) {
+        // Validate first
         for (FileItemIterator iter = new ServletFileUpload().getItemIterator(request); iter.hasNext();) {
           FileItemStream item = iter.next();
           if (item.isFormField()) {
             String fieldName = item.getFieldName();
-            String value = Streams.asString(item.openStream(), "UTF-8");
+            String value = Streams.asString(item.openStream(), StandardCharsets.UTF_8.toString());
```
gregorydlogan, line 772: Should this logic not also be added to `addZippedMediaPackage`?

gregorydlogan, line 772: Should this logic not also be added to `addZippedMediaPackage`?

gregorydlogan:
Ping @Arnei

modules/ingest-service-impl/src/main/java/org/opencastproject/ingest/endpoint/IngestRestService.java
Arnei: It does not have form params of type date, so no? Not that I'm against better param validation.

gregorydlogan: Fair.  I misread 773 to mean that we were parsing the data inside the DC rather than validating straight endpoint inputs.  This is fine.

gregorydlogan (approved):

PR #6018, opened 2024-07-11
Use JSON instead of XACML for ACL templates

lkiesow:
This patch switches from XACML to JSON for the format of our ACL templates provided via `etc/acl/`. The format is identical to what the REST API uses. This makes it much easier to read, write and understand the templates.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


geichelberger:
What do you think about switching to Yaml?

```json
{
  "acl": {
    "ace": [
      {
        "role": "ROLE_USER",
        "action": "read"
      }
    ]
  }
}
```
vs 
```yaml
---
acl: 
  ace: 
  - role: "ROLE_USER"
    action: "read"
```

lkiesow:
We already have a JSON parser. We don't have a YAML parser. So, I think that would be a larger update and this already improves the situation a lot.

Arnei (changes requested):
Looks to generally work, and I'm also in favour of getting rid of XACML.

There are still a lot of comments and logging statements using "XACML". Those should probably say "JSON" now to avoid confusion? 

Could also remove the `throws JAXBException` from `removeAcl(File artifact)` (which is was never actually throwing anyway) to get rid of the JAXBException import statement.

lkiesow:
Fixed that.

Arnei (approved):

mtneug:
@Arnei, can this be merged?

Issue #5993, opened 2024-07-01
Filter for isPublished in Admin UI


KatrinIhler:
Sometimes it would be nice if we could filter for only published events in the Admin UI (where events with only the internal publication would not count), e.g. for trying to find scheduled events with live publications, or only published processed events. The Admin UI shows this information, but you can't filter for it currently.

mtneug:
I like that. Currently, I sort by the publication column to group published events.

PR #5965, opened 2024-06-27
Remove old admin interface 

lkiesow:
This patch removes the old admin interface frontend.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #6023, opened 2024-07-15
Adding a Filter to filter Events according to their is_published state

ciegler:
This creates an additional filter for events. With this filter events can be filtered according to their is_published state in the index which states whether an event is published or not. To apply the filter an index rebuild of the existing event index is required to add their is_published state to the index. Connected issue: #5993

### DIFF
--- a/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -1982,6 +1982,11 @@
        "HOSTNAME": {
          "LABEL": "Host name"
        },
+       "IS_PUBLISHED": {
+         "LABEL": "Event is published",
+         "YES": "Yes",
+         "NO": "No"
+       },
        "LANGUAGE": {
          "LABEL": "Language"
        },
--- a/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
+++ b/modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
@@ -1982,11 +1982,6 @@
        "HOSTNAME": {
          "LABEL": "Host name"
        },
-       "IS_PUBLISHED": {
-         "LABEL": "Event is published",
-         "YES": "Yes",
-         "NO": "No"
-       },
        "LANGUAGE": {
          "LABEL": "Language"
        },
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -85,6 +85,7 @@ import org.opencastproject.index.service.exception.IndexServiceException;
 import org.opencastproject.index.service.exception.UnsupportedAssetException;
 import org.opencastproject.index.service.impl.util.EventUtils;
 import org.opencastproject.index.service.resources.list.provider.EventsListProvider.Comments;
+import org.opencastproject.index.service.resources.list.provider.EventsListProvider.IsPublished;
 import org.opencastproject.index.service.resources.list.query.EventListQuery;
 import org.opencastproject.index.service.resources.list.query.SeriesListQuery;
 import org.opencastproject.index.service.util.JSONUtils;
@@ -2580,6 +2581,17 @@ public abstract class AbstractEventEndpoint {
           default:
             logger.info("Unknown comment {}", filters.get(name));
             return Response.status(SC_BAD_REQUEST).build();
+        }
+      }
+      if (EventListQuery.FILTER_IS_PUBLISHED_NAME.equals(name)) {
+        switch (IsPublished.valueOf(filters.get(name))){
+          case YES:
+            query.withIsPublished(true);
+            break;
+          case NO:
+            query.withIsPublished(false);
+            break;
+          /* #TODO write default case */
         }
       }
       if (EventListQuery.FILTER_STARTDATE_NAME.equals(name)) {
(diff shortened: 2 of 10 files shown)


modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/EventsListProvider.java
```diff
@@ -59,13 +59,18 @@ public class EventsListProvider implements ResourceListProvider {
...
   public static final String PUBLISHER = PROVIDER_PREFIX + ".PUBLISHER";
+  public static final String ISPUBLISHED = PROVIDER_PREFIX + ".ISPUBLISHED";
 
   public enum Comments {
     NONE, OPEN, RESOLVED;
   }
 
+  public enum IsPublished {
+    YES, NO;
+  }
```
Arnei, line 70: This is basically a boolean. So why not use a boolean instead?

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
```diff
@@ -1982,6 +1982,11 @@
        "HOSTNAME": {
          "LABEL": "Host name"
        },
+       "IS_PUBLISHED": {
```
Arnei: I don't think the old admin ui will automatically use this new filter anyway, so these translations strings will go unused. Furthermore, the old admin ui is deprecated and will likely be removed with the next version of Opencast (see #5965). If you would like to make this filter available in the admin UI, consider the new admin ui instead https://github.com/opencast/opencast-admin-interface.

(To be clear, the requested change here is only to not add translation strings that will not be used.)

Arnei (changes requested):
Works and code reads well. Only one issue/question below.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
ciegler: I deleted the translation stings from the old admin ui, do I have to insert them in the new one somehow?

modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/EventsListProvider.java
ciegler: To be honest I do not have any specific reason for choosing an enum except that I started to build the filter in accordance to the comment filter, therefore I just copy-pasted the existing structure for the comment filter

Arnei: In that case I suggest using a boolean. No need to reinvent primitives.

modules/admin-ui-frontend/resources/public/org/opencastproject/adminui/languages/lang-en_US.json
Arnei: If you know that the new admin ui will makes use of the translations strings, add them. If not, don't.

modules/index-service/src/main/java/org/opencastproject/index/service/resources/list/provider/EventsListProvider.java
ciegler: As we discussed, after looking into it again, it turned out that the frontend does require a string, therefore a possible boolean needs to be mapped onto a "yes"/"no" string in the `EventListProvider.java` anyway, which means that working with an enum directly saves the step of mapping the boolean to a string. Additionally the existing filter-architecture is expecting strings so that a boolean might require a change of the already existing architecture in several places.

Arnei (approved):

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -2582,6 +2583,19 @@ public Response getEvents(@QueryParam("id") String id, @QueryParam("commentReaso
...
+            query.withIsPublished(true);
+            break;
+          case NO:
+            query.withIsPublished(false);
+            break;
+          default:
+            logger.info("Unknown is published status {}", filters.get(name));
+            return Response.status(SC_BAD_REQUEST).build();
+        }
+      }
```
gregorydlogan: Logically what you've done here makes sense, but Java throws this if you pass something *other* than `true`/`false`:
```
java.lang.IllegalArgumentException: No enum constant org.opencastproject.index.service.resources.list.provider.EventsListProvider.IsPublished.
FALSE
        at java.lang.Enum.valueOf(Enum.java:273) ~[?:?]
        at org.opencastproject.index.service.resources.list.provider.EventsListProvider$IsPublished.valueOf(EventsListProvider.java:68) ~[?:?]

```

So let's check if the value of the filter is valid, throw a 400 if it's not, and then handle the case where `isPublished` is unspecified.

```suggestion
      if (EventListQuery.FILTER_IS_PUBLISHED_NAME.equals(name)) {
        if (filters.containsKey(name)) {
          switch (IsPublished.valueOf(filters.get(name))) {
            case YES:
              query.withIsPublished(true);
              break;
            case NO:
              query.withIsPublished(false);
              break;
            default:
              break;
          }
        } else {
          logger.info("Query for invalid published status: {}", filters.get(name));
          return Response.status(SC_BAD_REQUEST).build();
        }
      }
```

And yes, I'm morbidly aware that just a few lines above this we do *exactly the same wrong thing*

gregorydlogan (changes requested):

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
ciegler: I have now applied your suggested changes and thanks, I actually did not knew that this could happen :) 
Also sorry for taking so long to reply.

PR #6025, opened 2024-07-16
Add config changes to upgrade.md

pselsa:
The changes from https://github.com/opencast/opencast/pull/5937 should be announced in the upgrade manual so that the admin ui wont break for non-admin users.

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -32,9 +32,14 @@ to rebuild and you don't need to rebuild all indexes. To start this process:
 
 ## Configuration changes
 
-For the new admin-UI to work for non-admin users, access has to be granted for the assets directory in
-`etc/security/mh_default_org.xml`. Make sure the changes from [#5937](https://github.com/opencast/opencast/pull/5937/files)
-are applied to your local config file.
+Check for changes in configuration files and apply those to your local copies. You can use the following command
+to list all changes:
+```
+git diff origin/r/15.x origin/r/16.x -- etc/
+```
+
+Please make sure to apply at least the following changes since they are crucial for a stable setup:
+- Grant acces to the assets directory in `etc/security/mh_default_org.xml`. @see [#5937](https://github.com/opencast/opencast/pull/5937/files)
 
 ## API Changes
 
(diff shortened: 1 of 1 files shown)


lkiesow:
I think adding the section is fine, but it's not good to just have one change in the section since then users will assume it's the only change and not check the others. We could just tell them to compare the `r/15.x` and `r/16.x` configuration folders?

pselsa:
I mean yeah, there could be a general note on every release that you should check your config files

pselsa:
@lkiesow can you have a look now?

pselsa:
This part could be copy&pasted to all next releases

docs/guides/admin/docs/upgrade.md
```diff
@@ -29,6 +30,16 @@ to rebuild and you don't need to rebuild all indexes. To start this process:
 - Set `Search` as service and submit
 - The logs should show the status of the index rebuild
 
+## Configuration changes
+
+Check for changes in configuration files and apply those to your local copies. You can use the following command
+to list all changes:
+```
+git diff origin/r/15.x origin/r/16.x -- etc/
```
gregorydlogan: Since this is something that's probably a good idea to do *every* upgrade, why don't we make this more maintainable with something like this?

```suggestion
git diff origin/r/{{ opencast_major_version() | int - 1 }} origin/r/{{ opencast_major_version() }} -- etc/
```

pselsa: Is that some kind of script language for the markdown files? If it works, why not!

gregorydlogan: It's a jinja macro IIRC, but yes this works when I tested it.  It's also in use in a few places throughout the docs so it should remain stable.

lkiesow: The `opencast_major_version()` is a macro we define for our mkdocs builds:
[/docs/guides/admin/main.py](https://github.com/opencast/opencast/blob/3300def2ddcc27763d4ff551d61757c9ba498f73/docs/guides/admin/main.py)

Works and I'll include that, but you forgot about the `.r`, Greg:

```suggestion
git diff origin/r/{{ opencast_major_version() | int - 1 }}.x origin/r/{{ opencast_major_version() }}.x -- etc/
```

PR #6028, opened 2024-07-18
Add waveform option filter mode

geichelberger:
This adds a FFmpeg `showwavepics` filter option to change the filter mode from the default `average` to `peak`.

I didn't change the default value, but maybe it would make sense to make `peak` the default mode?

**Current default**
default

**Current default + filter mode peak**
default-draw-scale-filter-peak

**Audacity**
grafik


lkiesow:
Just looking at the picture I think `peak` as default is reasonable.

Arnei (approved):
Works, code looks reasonable and waveform looks better (and going back to the way it was before is also possible!).

mtneug:
@Arnei, can this be merged?

Arnei:
Sure, let's merge it.

Issue #5963, opened 2024-06-27
Retract Files on Engage Merge


KatrinIhler:
We remove the old files from search/the publication when updating the engage publication via merge, but don't actually retract the files.

mtneug:
For a detailed description see https://github.com/opencast/opencast/issues/6035#issue-2426743924

PR #6033, opened 2024-07-23
Automatic republication of metadata and ACLs to Engage

JulianKniephoff:
Supersedes #5172.

This PR refactors publication to Engage in multiple ways.

* It extracts the logic of the `publish-engage` WOH into its own publication service, roughly modeled after preexisting ones.
  * It (hopefully) goes the whole nine yards in doing so, remote-impl and everything.
* It extends said service so that publication can happen synchronously.
* This is then used in the `publish-engage` WOH, and to implement the feature this PR is actually about, but see below.
  * Note that the behavior of `publish-engage` might have changed slightly, mostly in terms of when and how it spawns jobs and when it returns `SKIPPED` vs. `CONTINUE`. I hope this is fine, but if this is reason enough for people to not want it in 16.x, I can retarget, of course.

All of that happens in the first commit, and is in service to the feature implemented in the second one: Automatically republishing metadata and ACLs to Engage. To do this, this PR uses the approach initially suggested in #5172, namely just hooking into the appropriate index service methods. Instead of duplicating the (re-)publication logic, though, we can now reuse it properly.

The approach of using the asset manager event mechanism was discarded, mostly due to difficulties in determining when, what and how to update the publication, while also avoiding infinite loops.

In contrast to #5172 this should work with S3 as well. However, this is still not perfect. A number of issues remain relevant from #5172 and/or arose during further development:

* There is no transaction mechanism involved here: If republication fails, the corresponding index service methods fail without any form of rollback, potentially leaving the system in an inconsistent state.
  * This sadly seems to be the norm in Opencast, and creating the necessary facilities to avoid this is outside the scope of this PR and outside the budget of the parties interested in this.
* This **only** works with Engage for now, no other publication channel. Extending this cleanly to others is another huge task, and would require harmonizing a lot of the ad-hoc structures that surround the concept of publication right now.
  * To wit: "publication service" is not a thing in Opencast; these are all disparate interfaces that can do different things for different publication channels (whatever was needed for the particular use case that necessitated them).
* This now only works for updates done to events directly through the appropriate index service methods.
  * Specifically it does not automatically republish changes to series metadata and/or ACLs. There is another mechanism for that, using the asset manager event mechanism. I would **love** to harmonize these mechanisms, but again, that's outside the scope of this PR.
    * Note that that mechanism is broken in a way, since it directly updates the publication, without letting the asset manager know about it.
  * The live schedule service has another version of this logic for live events, which this solution doesn't touch at all. In a perfect world, these would be unified as well. :(


modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
```diff
@@ -2079,4 +2144,13 @@ public boolean isWorkflowActive(String workflowState) {
             || WorkflowState.PAUSED.toString().equals(workflowState);
   }
 
+  private void updateEngagePublication(MediaPackage mediaPackage) throws PublicationException {
+    if (!autoRepublish) {
+      return;
+    }
+
+  engagePublicationService.publishSync(mediaPackage, null, PUBLISH_STRATEGY_MERGE,
```
mtneug, line 2152: Why is check availability and merge strategy not configurable?

mtneug (changes requested):
Still not a huge fan of this.

* Why not start a workflow automatically giving users the chance to adapt what is being done (e.g. change the thumbnail)?
* What will happen, if a synchronized publish is started multiple times (e.g. two quick button presses / API calls)?

LukasKalbertodt:
> Why not start a workflow automatically giving users the chance to adapt what is being done (e.g. change the thumbnail)?

Opencast is used by lots of external apps (LMSes, Tobira, ...) basically as a database. Opencast is supposed to hold data about videos, not the individual apps. We specifically don't want apps to duplicate data -> everything is centralized in Opencast.

These apps want to offer editing event metadata or ACLs. In the current system (which you suggest to keep, if I'm understanding you correctly), there is no good way to do that:
- Option A: the UI shows a spinner until the app has verified that Opencast did correctly persist the changes. Which is annoying to check for and can take 10s of seconds, or even longer. Not acceptable UX-wise.
- Option B: the app pretends like the changes are already persisted, optimistically updating its own database with the changes and shows them to the user, rolling back the changes in case Opencast did in fact not persist them (which again, is actually hard to find out). Now we have two versions of the data and a situation that easily leads to inconsistencies.
- Option C: The app shows a "might take a while to update" note after saving the changes and keeps showing the old metadata until Opencast is done. Unclear at what point the app should show a "the changes probably failed to apply" warning instead. Also requires additional logic in the app and from a UX standpoint clunky at best.

Opencast needs to function more like a database that apps can trust on. Much of the complexity and the bugs in the Opencast ecosystem stems from the fact that Opencast currently promises absolutely nothing about its data. That leads to increased complexity (due to defensive programming) in the external apps (and Opencast itself), as well as lots of incorrect assumptions (which lead to bugs). So I think Opencast needs to start promising stuff, and related to this PR: promising some basic form of atomic and reliable data updates. 

And yes, that means we disallow arbitrary logic to run *during* these updates. In my world view, things like "updating S3" or "regenerating thumbnails" or "updating youtube publications" are things that Opencast has to handle itself, *after* the actual data update. Yes, those might fail and Opencast should have some kind of retry mechanism or something. But these fallible operations should not affect the actual data-store operation of Opencast.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
JulianKniephoff: My thinking was that for the use case proposed here these are basically set, i.e. you don't want `checkAvailability` and `strategy` needs to be `"merge"`. It was further inspired/cemented by the `republish-metadata` workflow. That assumption might be wrong, though, and obviously adding config for those options isn't a problem if people want it.

JulianKniephoff:
@LukasKalbertodt pretty much answered the first point @mtneug raised. As for the second one:

That's a fair question which I don't know the answer to. But I would say that that's a preexisting problem. I mentioned transactions in my OP, with a focus on the missing atomicity, but the same applies to isolation (thinking ACID here): Apart from workflows Opencast has no safeguard against this I'm aware of, and the data race you mention already exists for snapshots as far as I can tell.

We could of course do something crude and `synchronize` these methods on the MP ID or something like that. 🤔

geichelberger:
> We could of course do something crude and synchronize these methods on the MP ID or something like that. 🤔

You would need a distributed lock, or else you would kill any chance of creating a HA Setup.

modules/index-service/src/main/java/org/opencastproject/index/service/impl/IndexServiceImpl.java
mtneug: Well `merge` is a problem because of #5963. Not for ACL and dublincore/episode, but next people want to republish preview images and then you have leaks.

mtneug:
> These apps want to offer editing event metadata or ACLs. In the current system (which you suggest to keep, if I'm
> understanding you correctly), there is no good way to do that:
>
> ...
>
> Opencast needs to function more like a database that apps can trust on. Much of the complexity and the bugs in the
> Opencast ecosystem stems from the fact that Opencast currently promises absolutely nothing about its data. That leads to
> increased complexity (due to defensive programming) in the external apps (and Opencast itself), as well as lots of
> incorrect assumptions (which lead to bugs). So I think Opencast needs to start promising stuff, and related to this PR:
> promising some basic form of atomic and reliable data updates.
>
> And yes, that means we disallow arbitrary logic to run during these updates. In my world view, things like "updating S3"
> or "regenerating thumbnails" or "updating youtube publications" are things that Opencast has to handle itself, after the
> actual data update. Yes, those might fail and Opencast should have some kind of retry mechanism or something. But these
> fallible operations should not affect the actual data-store operation of Opencast.

@LukasKalbertodt, @JulianKniephoff, let me be clear: I understand the bad UX. Also, I agree that Opencast ideally would make guarantees about (atomic) updates. However, you are mixing multiple concerns with each other. First, speed of update mostly related to UX. Second, transactional guarantees comparable to mentioned databases, e.g. ACID or BASE. Both concerns are _ not _ (fully) addressed by this PR.

**Speed**

Updating metadata and ACLs is (potentially) slow with and without this PR. Period. The reason is that each new change will create a new snapshot in the asset manager, i.e. the code before this PR. Creating snapshots is especially expensive if you use S3. Basically, these steps are performed:

```
1 gather all files from last snapshot to local file system
2 change metadata/acl/etc.
3 create snapshot in local asset store
4 (upload snapshot to S3 maybe after some time or with a workflow)
```

1 is fast if everything is living on the same file system and you can simply create hard links. 1 is no at all fast if the last snapshot lives in S3 and _ everything _ (inkl. videos etc.) is downloaded from S3 to the local file system. This PR is doing nothing regarding this slow operation and as such external apps still need to deal with slow updates!

(@LukasKalbertodt "updating S3" in terms of the asset manager is thus already happening later in the background. "updating S3" in terms of the publications is another thing and should not happen in the future with your PR)

**Consistency Guarantees**

This PR does not at all deal with consistency guarantees, but it rather just arbitrarily redefines the operational model for one specific use case while ignoring other totally valid use cases. In fact, if consistency guarantees are the concern, you should reject this PR as it introduces more potential race conditions.

As it stands right now, Opencast, at a very high level, tries to define the following transactional model:

```
1
http_transaction {
  create_new_snapshot_version {
    update elements
  }
}

2
workflow_transaction {
  create_new_snapshot_version {
    update elements as defined by the use cases I want to implement
    update publications as defined by the use cases I want to implement
  }
}
```

What you suggest is to redefine this model to the following:

```
3
http_transaction {
  create_new_snapshot_version {
    update elements
  }
  (arbitrary) update one specific publication channel with (arbitrarily) defined elements
}

4
workflow_transaction {
  create_new_snapshot_version {
    update elements as defined by the use cases I want to implement
    update publications as defined by the use cases I want to implement
  }
}
```

Now, I'm certain that both models don't hold up against BASE or even ACID guarantees in the current implementations. However, I'm reasonably certain that the first model is robust in practice in terms of consistency. First of all, media packages are versioned and hopefully, parallel `http_transactions` will create distinct versions. Further, Opencast will more or less guarantee that there is only one `workflow_transaction` at a time. Note that publications are not versioned and the only place publications are updated is during a workflow (please ignore the live stream stuff that doesn't use workflows at all and is, therefore, also consistent).

Now, regarding the second model. I already raised the question regarding two parallel `http_transactions`. BTW, this is a problem for a single admin instance; we run two in parallel further increasing the risk. Additionally, what will happen if multiple `http_transactions` and a `workflow_transaction` run at the same time? This PR thus introduces more potential races.

Ignoring the races, this PR is complicating the operational model and introducing a "magic" update step. It's hard to reason about the state of a media package (is it already published? was there an error? etc.). This complicates development and operations. Instead, automatically triggering a workflow is clear for all.

This PR also just arbitrarily caters for this specific use case. What about people who have other publication channels? What if I want to update my thumbnails because I use metadata fields to generate them? Yada yada? We already have a way to solve this and we call that a workflow.

**Other Issues**

**Bug in `merge` (solvable):** We currently have a bug in the merge strategy (see #5963). This is solvable and also does not affect ACLs/metadata (but potentially thumbnails and other elements).

**LTI code is apparently triggering a republish (solvable):** The LTI code seems to trigger [an automatic republish](https://github.com/opencast/opencast/blob/58b5beda61b1aaf052a351153a04e69884b7e028/modules/lti-service-impl/src/main/java/org/opencastproject/lti/service/impl/LtiServiceImpl.java#L403-L423). BTW, the workflow ID is hard-coded!

**Is `updateEventAcl` called by the Admin UI API? (solvable):** I could only find calls from the external API to this method leading to potential inconsistencies when updating stuff form the Admin UI.

mtneug:
Alternative suggestion from the committer meeting at DACH Bremen 2024:

**Metadata and ACLs only exist once for an event (and are stored in the database). I.e. there is no longer a difference between metadata/ACLs in the asset manager, search, custom publication channels, (anything else?).**

As a consequence:
* Changes don't need to be published anymore.
* Maybe we can even make changes during a workflow execution (still, there might be some races for operations that read/write metadata/ACLs).

Still open for discussion:
* Are metadata/ACLs versioned/snapshotted?
* How do you change elements (e.g. thumbnails) after a metadata change?
* Probably more consequences we need to think about.

gregorydlogan:
Piling on here, should this still go into `r/16.x` rather than `r/17.x` since the latter is now the legacy release?

JulianKniephoff:
The fate of this PR is pretty much unclear at this point anyway. The proposed meeting between @mtneug, @lkiesow and I still hasn't happened, and I don't know how much sense this even still makes in light of the work planned with the Rüdiger-money. I know that at least the Tobira project is currently “not interested” in this anymore because of this.

Still, I would hesitate to close it before it gets any kind of ... closure. 😅 We can turn it back into a draft for now, though.

oas777:
> I know that at least the Tobira project is currently “not interested” in this anymore because of this.

You spelled "the Tobira project is very much disappointed this prevents efficient metadata editing and updating" wrong.

JulianKniephoff:
Cleaning up my stuff a bit here, I'm finally ready to close this. The underlying problem will be solved in https://github.com/opencast/opencast/issues/6570, and I don't really see a reality where we get this feature without/before that. 🤷‍♀️

PR #6037, opened 2024-07-24
Make in-place text substitution in the docs more portable

JulianKniephoff:
Unfortunately there is no way to make calls to `sed -i` portable and at the same time have it not create a buckup copy.

Luckily, Perl is an easy replacement.

### DIFF
--- a/docs/guides/admin/docs/configuration/migration-domain-https.md
+++ b/docs/guides/admin/docs/configuration/migration-domain-https.md
@@ -25,7 +25,7 @@ In that case, the following steps might help.
          # This changes the domain AND from http to https. Carefully adjust the command as needed!
          # Your old URL may or may not have the port explicitly listed.  Check the previous value of org.opencastproject.server.url and match that.
          find . -type f -name "*.xml" -exec \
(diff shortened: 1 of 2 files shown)


mtneug:
Perl refuses to die 🫣

Is it safe to assume that Perl is installed on every system? Are backup files really so bad 😅?

JulianKniephoff:
One of the commands I changed is for migrating URLs in production data, I personally wouldn't want backup files lying around in there. 🤔 Of course one could do the `perl` solution only for that, but that's inconsistent. Or one adds a note about potentially deleting the backups.

Regarding the portability of `perl`: That's fair, however, `sed -i` is also non-standard, so you could also end up on a system that doesn't have it. Most Unices have a version of it, of course, but the same is true for `perl`.

The only real portable way in the sense of being at least backed by POSIX would be `sed .. > some_file` and then manually `mv some_file original_file`. Of course with these commands using wildcards as well that gets relatively complex pretty quickly. 🤔

mtneug:
Fair. I also thought about `sed < f > f`, but this doesn't work as you need a buffer (e.g. additional file). Well as I said, guess Perl still lives another day :D

gregorydlogan (approved):

PR #6045, opened 2024-07-26
Fix downloadsPlugin in PaellaPlayer7

jduehring:
In OC16 there have been some changes to the search-API, which in part were not correctly adressed for the downloadsPlugin to work properly. If you check any current OC build >= 16.x the download tab will not be shown in Paella,  even if it is correctly configured. 

This PR should fix this problem.

### DIFF
--- a/modules/engage-paella-player-7/src/js/OpencastAuth.js
+++ b/modules/engage-paella-player-7/src/js/OpencastAuth.js
@@ -87,16 +87,15 @@ export default class OpencastAuth {
       if (!(roles instanceof Array)) { roles = [roles]; }
 
       let canWrite = false;
-      if (acl?.ace) {
-        let aces = acl?.ace;
-        if (!(aces instanceof Array)) { aces = [aces]; }
+      if (acl) {
+        if (!(acl instanceof Array)) { acl = [acl]; }
 
         canWrite = roles.some(function(currentRole) {
           if (currentRole == userInfo.org.adminRole) {
             return true;
           }
           else {
-            return aces.some(function(currentAce) {
+            return acl.some(function(currentAce) {
               if (currentRole == currentAce.role) {
                 if (currentAce.action == 'write') {
                   return true;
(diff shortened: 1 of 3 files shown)


lkiesow:
Can you provide the necessary configuration change to the Paella configuration and a few steps to easily test this?

jduehring:
> Can you provide the necessary configuration change to the Paella configuration and a few steps to easily test this?

The DownloadsPlugin is currently activated by default and you should see the download buttons in Paella if you have write-access or if the licenses match. 

So basically just check that this is not the case on stable.opencast.org or develop.opencast.org , build a local OC with this patch and hopefully see the download buttons again, after uploading a video :D  No config changes neccessary.

Arnei (changes requested):
When clicking on the cog wheel in the Paella Player, nothing happens. An error is logged in the web console.

```
Uncaught (in promise) TypeError: o.filter is not a function
    t org.opencast.paella.downloadsPlugin.js:62
```

Looks related to the changes you made to `getEpisode({episodeId})`

jduehring:
> When clicking on the cog wheel in the Paella Player, nothing happens. An error is logged in the web console.
> 
> ```
> Uncaught (in promise) TypeError: o.filter is not a function
>     t org.opencast.paella.downloadsPlugin.js:62
> ```
> 
> Looks related to the changes you made to `getEpisode({episodeId})`

Did you do anything other than upload and play the video? For me it seems to work without any related errors in the console (tested with chrome and firefox): 

image

owi92:
I can't reproduce the error @Arnei mentioned either (checked out and installed this branch and uploaded some videos). Tested both single and dual streams.

gregorydlogan:
I can reproduce this, in my case `tracks` is an Object.  Screenshot from the Chrome dev tools:
image

jduehring:
Thanks, I could reproduce this now using the `fast` workflow instead of `schedule-and-upload`. I'll look into it.

jduehring:
I changed the code so that if `tracks` is an object it will be wrapped inside an array so the rest can be processed as before. 
With this change it should work in both cases now. 

I also looked into ensuring that `tracks` is always an array before the function call, but couldn't get it working properly. Some other parts in the code seem to also assume that `tracks`  can be an array **or** an object, so I came back to my first solution.

PR #5829, opened 2024-05-08
Ignore subtitle tracks in select-tracks WHO

snoesberger:
fixes https://github.com/opencast/opencast/issues/5817

Subtitle tracks interfere with the proper functioning of the select-track WHO. Therefore we have to exclude them from the selection process.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* ~[ ] include migration scripts and documentation, if appropriate~
* [x] pass automated tests
* [x] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)

PR #6046, opened 2024-07-26
Don't duplicate captions during publication

KatrinIhler:
#5829 requires a change to the publication workflow, otherwise we duplicate captions during publication.

### DIFF
--- a/docs/guides/admin/docs/releasenotes.md
+++ b/docs/guides/admin/docs/releasenotes.md
@@ -14,6 +14,12 @@ See [changelog](changelog.md#opencast-157) for a comprehensive list of changes.
 
 ## Opencast 15.6
 
+**Attention:**
+The fix for the handling of captions by the select-tracks operation
+([5829](https://github.com/opencast/opencast/pull/5829)) requires the removal of the operation directly afterwards from
+the `partial-publish` community workflow, otherwise captions will be duplicated during publication (see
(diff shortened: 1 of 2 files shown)


docs/guides/admin/docs/releasenotes.md
```diff
@@ -2,6 +2,12 @@
 
 ## Opencast 15.6
 
+**Attention:**
+The fix for the handling of captions by the select-tracks operation
+([5829](https://github.com/opencast/opencast/pull/5829)) requires the removal of the operation directly afterwards from
+the `partial-publish` community workflow, otherwise captions will be duplicated during publication (see
+[6046](https://github.com/opencast/opencast/pull/6046) which is _not_ included in this release).
```
lkiesow: I find this wording pretty confusing. Maybe something like this is clearer:
```suggestion
the `partial-publish` workflow, otherwise captions will be duplicated during publication.
The workflow has beein fixed in 15.7.
For details, see [#6046](https://github.com/opencast/opencast/pull/6046).
```

KatrinIhler: > The workflow has been fixed in 15.7.

That's not true though. It will hopefully be fixed in 15.8, but that depends on whether this PR gets merged in time. But I guess one could update the release notes afterwards similarly to how I did it for 16.

PR #6082, opened 2024-08-01
Sign urls in search rest service again when id is present

### LINKED ISSUE #5957: OC16 Seach Index does not support URL signing
From #5597 
"Notably, I did not include URL signing support with this. If this is a deal breaker we can work on it later."

This leads to Stream Security not working anymore.

In our case we have to disable stream security on our Wowza Server to play videos again with Opencast 16.

This is still documentented[1] but does not work anymore.

[1] https://docs.opencast.org/r/16.x/admin/#configuration/stream-security/stream-security-config/#configuration-of-wowza-url-signing

pselsa:
closes #5957

Since the search rest service doesnt sign urls anymore, the engage player wont work when signing is required for all urls. This PR adds url signing again to the search rest service but only when the id parameter is present so that it wont slow down the api when the result returns multiple events. With that the engage player will work again since it always presents an id.


lkiesow (changes requested):
Sorry for the late review, but along the way, I realized that my comment in https://github.com/opencast/opencast/issues/5957#issuecomment-2255424612 is incorrect and the implementation has significantly deviated from the original draft. This meant, I needed a closer look.

In particular the assumption that we cannot modify items on the fly and can only do this for single items (basically just when using `id` is just wrong). I don't quite know why we need to parse all the index results, but we do that, so there is no reason to not use what we have 🤷

That means we can just sign all results and not only if `id` is being used. This also means that there is less magic involved. Otherwise, I would have at least complained about the lack of documentation for the difference in behavior.

However, my recommendation would be to re-implement the old behavior of `GET /search/episode.json` which had a [`sign` URL parameter](https://legacy.opencast.org/docs.html?path=/search#episodes-1) and only sign if it is set to `true`.

Since the default was `true`, that should work just fine for you, I hope.

As for the changes, that should just be to
- re-introduce [the `sign` parameter](https://github.com/opencast/opencast/blob/6e59c86ca46cbe6f3b34dd45a3ff1238eff1871a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java#L379-L385)
- Ensure that an empty value is interpreted as `true` (default)
- Replace `StringUtils.isNotEmpty(id)` from you patch with `signParamValueAsBool` (←please choose a better name 😁)


Finally, so that I can quickly test this, can you post a minimal configuration for the URL sign provider to test this with?

lkiesow:
Forgot one thing, sorry. Since you want this to go into 16.3, can you target `r/16.x` instead of `develop` with this pull request?

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -433,6 +446,39 @@ public Response getEpisodes(
...
+          if (this.urlSigningService.accepts(urlToSign)) {
+            try {
+              String signedUrl = this.urlSigningService.sign(
+                  urlToSign,
+                  expireSeconds,
+                  null,
+                  null);
+              map.put(entry.getKey(), signedUrl);
+            } catch (UrlSigningException e) {
+              logger.debug("Unable to sign url '" + urlToSign + "'.");
```
lkiesow, line 468: I've overlooked this. You should usually avoid using string concatenation in log statements. Use the built-in placeholders instead:
```suggestion
              logger.debug("Unable to sign url '{}'.", urlToSign);
```

pselsa:
> Forgot one thing, sorry. Since you want this to go into 16.3, can you target `r/16.x` instead of `develop` with this pull request?

Made a new PR targeting r/16.x: https://github.com/opencast/opencast/pull/6094

pselsa:
> Finally, so that I can quickly test this, can you post a minimal configuration for the URL sign provider to test this with?

Easy! Just follow this minimal config example: https://docs.opencast.org/r/16.x/admin/#configuration/stream-security/stream-security-config/#minimal-configuration-example

pselsa:
Uh shit, just realized i could have changed the target of this PR.. whatever, gonna close this one now.. see you there: #6094

lkiesow:
> > Forgot one thing, sorry. Since you want this to go into 16.3, can you target `r/16.x` instead of `develop` with this pull request?
> 
> Made a new PR targeting r/16.x: #6094

You can change the target branch of pull requests later on in GitHub. You will find the option if you hit the „edit“ next to the pull requests title.

PR #6083, opened 2024-08-01
Always send adopter registration statistics in rest api

### LINKED ISSUE #858: Adopter registration shows incorrect data to the user
As a user, when you fill out the adopter registration for the first time, Opencast may show you incorrect information about what is being sent to register.opencast.org. Specifically, if “send usage statistics” was not selected before and you select it, Opencast would show that no statistics information are sent.

https://github.com/user-attachments/assets/d2515534-c857-4bb7-b8aa-bd70da0d0bbd

dennis531:
Currently, the `/admin-ng/adopter/summary` route only returns the statistics data if statistics are allowed. This causes the admin ui problem mentioned in opencast/opencast-admin-interface#858. These changes return this data always.

Close opencast/opencast-admin-interface#858

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
+++ b/modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/statistic/ScheduledDataCollector.java
@@ -242,12 +242,7 @@ public class ScheduledDataCollector extends TimerTask {
   public String getRegistrationDataAsString() throws Exception {
     Form adopter = (Form) adopterFormService.retrieveFormData();
     String generalJson = collectGeneralData(adopter);
-    String statsJson;
-    if (adopter.allowsStatistics()) {
-      statsJson = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
-    } else {
-      statsJson = "{}";
-    }
+    String statsJson = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
     //It's not stupid if it works!
     return "{ \"general\":" + generalJson + ", \"statistics\":" + statsJson + "}";
   }


Arnei:
I'm getting 500 from the `/admin-ng/adopter/summary` endpoint, but that might be related to my setup? Do I need to configure anything specific to test this?

Won't this cause the frontend to then always send statistics information no matter what the user selected?

Can we get this for r/15.x? :D

gregorydlogan:
@Arnei Is there a stack trace or steps to reproduce we can see?  I'm not seeing this, but it shouldn't be doing that regardless of setup.

@dennis531 I find this change confusing since that endpoint's output is never (as far as I can tell) used directly.  I feel leery about generating that summary data regardless of whether it actually gets sent.  What's the reasoning here in terms of generating that data?

dennis531:
> Won't this cause the frontend to then always send statistics information no matter what the user selected?

No, the frontend doesn't send the data directly to the registration server. This data is only used to display the registration and statistics data to the admin in the admin ui. The frontend posts the entered user values to Opencast.

> Can we get this for r/15.x? :D

I guess yes, and try to change the target branch to r/15.x.

dennis531:
> @dennis531 I find this change confusing since that endpoint's output is never (as far as I can tell) used directly. I feel leery about generating that summary data regardless of whether it actually gets sent. What's the reasoning here in terms of generating that data?

The summary data is used in the admin ui to get previous values and display them in the adopter registration form. In addition, a preview of the statistics is also shown in the ui. Depending on the previous values, the statistics preview could be empty, as mentioned in https://github.com/opencast/opencast-admin-interface/issues/858. These changes always provide the statistics to the admin ui regardless of the previously stored values.

image

gregorydlogan:
Ok, I can see the usecase.  I'd still rather see a way to temporarily generate the stats rather than this approach, but that would be a whole bunch of extra engineering effort rather than just getting it done.

PR #6094, opened 2024-08-12
Sign urls in search rest service again

### LINKED ISSUE #5957: OC16 Seach Index does not support URL signing

pselsa:
closes #5957

Since the search rest service doesnt sign urls anymore, the engage player wont work when signing is required for all urls. This PR adds url signing again to the search rest service and therefore reintroduces the `sign` parameter, having the default to true.

Previous PR that falsely targeted develop: https://github.com/opencast/opencast/pull/6082

### DIFF
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
@@ -116,8 +116,6 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
 
   private UrlSigningService urlSigningService;
 
-  private long expireSeconds = UrlSigningServiceOsgiUtil.DEFAULT_URL_SIGNING_EXPIRE_DURATION;
-
   @GET
   @Path("series.json")
   @Produces(MediaType.APPLICATION_JSON)
@@ -468,7 +466,7 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
             try {
               String signedUrl = this.urlSigningService.sign(
                   urlToSign,
-                  expireSeconds,
+                  UrlSigningServiceOsgiUtil.DEFAULT_URL_SIGNING_EXPIRE_DURATION,
                   null,
                   null);
               map.put(entry.getKey(), signedUrl);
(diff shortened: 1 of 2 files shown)


modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -111,6 +114,10 @@ public class SearchRestService extends AbstractJobProducerEndpoint {
 
   private final Gson gson = new Gson();
 
+  private UrlSigningService urlSigningService;
+
+  private long expireSeconds = UrlSigningServiceOsgiUtil.DEFAULT_URL_SIGNING_EXPIRE_DURATION;
```
lkiesow: Setting this to `DEFAULT_URL_SIGNING_EXPIRE_DURATION` implies that this is configurable. Otherwise, you could also have used the constant in `findURLsAndSign(…)` directly. But this is never updated. Is this supposed to be configurable?

lkiesow: I'm merging it like this to get it into 16.3. It works and shouldn't hurt after all. I've filed this issue as https://github.com/opencast/opencast/issues/6099. If you can, it would be nice if you would take a look at that issue.

lkiesow:
Ah, just when I wanted to hit merge :D
Will do a quick review and see that this makes it into 16.3 after CI is done.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
pselsa: Ok cool! You can assign the issue to me if you like ( i cannot do that myself)

PR #6111, opened 2024-08-27
Treat `filter` parameter elements in admin UI APIs as URL encoded

JulianKniephoff:
This allows us to use literal `,` and `:` in filters, cf. #1673, specifically https://github.com/opencast/opencast/issues/1673#issuecomment-2093704704.

Note that this targets `r/15.x` despite it changing behavior! Arguably a comma leading to filters being misinterpreted is a bug, but this "fixes" it by also changing the meaning of certain other filters. (Think literal `%2C` for example.)

I did this to not lock ourselves out of updating the admin UI in `legacy` and `stable` once the corresponding PR for the admin UI lands.

Note also that I had to duplicate the old version of the method used into the `GroupsEndpoint` of the external API as to not change its behavior. This makes most (hopefully all) admin UI endpoints' behavior regarding the `filter` parameter (more) inconsistent with the external API.

Fixing that would be disproportionally more effort in my opinion, and if I were to do that, I would probably completely revamp how filters are passed/processed to begin with, because frankly naively splitting a string at `,` was never a good idea.

This way at least clicking on a name with a comma in the admin UI does the right thing. 🤷‍♀️

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
@@ -22,6 +22,7 @@
 package org.opencastproject.adminui.endpoint;
 
 import org.opencastproject.adminui.exception.JsonCreationException;
+import org.opencastproject.index.service.util.RestUtils;
 import org.opencastproject.list.impl.ResourceListQueryImpl;
 import org.opencastproject.list.query.StringListFilter;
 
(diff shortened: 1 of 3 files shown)


lkiesow:
I would suggest putting this at least into `r/16.x`. We are already halve through the legacy cycle of Opencast 15, and we should really only put in what we really need to avoid any problems with releases, updates and incompatibilities.

JulianKniephoff:
Rebased!

gregorydlogan (changes requested):
> This way at least clicking on a name with a comma in the admin UI does the right thing. 🤷‍♀️

So, uh, when I test this using [this](https://github.com/opencast/helper-scripts/blob/2b14f2c4cde55fb685533a1db27ca12dc8cef81d/ingest/media.yml#L195) list of media, when I click on Sascha's name I get a blank list in the admin ui.  This is caused by the changes in `RestUtils` doing this:

image

JulianKniephoff:
Wait, I'm confused. How did you test this? Note that the admin UI does not URL-encode the filters, yet. This just changes the backend to interpret them as such. So yeah, the behavior you are describing is exactly what we are seeing and trying to fix here, and this patch alone doesn't fix it, but it's a requirement to do so.

JulianKniephoff:
I reused this logic in another method used by a few other admin UI endpoints. This should make it even more consistent within the `/admin-ng` namespace, and it should make the frontend PR much easier.

JulianKniephoff:
I just released the frontend PR: https://github.com/opencast/opencast-admin-interface/pull/958

PR #6156, opened 2024-09-11
Fix non-themed subtitle publications

### LINKED ISSUE #6138: Subtitles no longer published for events without branding
**Describe the bug**

Since the introduction of https://github.com/opencast/opencast/pull/5913 subtitles for videos without a branding are no longer published. If the video is part of a series with branding the subtitles are correctly published.

**To Reproduce**

1. Create a series without branding
2. Add a new event and upload a video with a subtitles file (or let the speech-to-text WOH generate subtitles)
3. Publish the video with the subtitles
4. Open the video in the paella player -> subtitles are not available

**Expected behavior**
Subtitles should be available in the paella player.

**Server environment:**
 - Version 15.x

snoesberger:
Fixes #6138 

At the end of the workflow partial-theming captions/trimmed is tagged as captions/shifted, this is necessary for the publication of the captions. If there is no branding/theme configured for an event, there is no captions/trimmed flavour available to tag. This is because captions/trimmed is replaced by captions/themed in partial-theming WF. To fix this, a copy needs to be created instead of replacing.

### Your pull request should…

* [x] have a concise title
* [x [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [x] pass automated tests

### DIFF
--- a/etc/workflows/partial-theming.xml
+++ b/etc/workflows/partial-theming.xml
@@ -46,7 +46,7 @@
       if="NOT(${theme_active})"
       description="Tag elements as themed">
       <configurations>
-        <configuration key="source-flavors">*/trimmed</configuration>
+        <configuration key="source-flavors">presenter/trimmed,presentation/trimmed</configuration>
         <configuration key="target-flavor">*/themed</configuration>
         <configuration key="copy">false</configuration>
       </configurations>


Arnei:
From my testing, this fixes the issue.

Although I worry about setting `copy` to true. This will clone everything in `*/trimmed`, including potentially very large video files which could be fairly expensive, no?

snoesberger:
> Although I worry about setting `copy` to true. This will clone everything in `*/trimmed`, including potentially very large video files which could be fairly expensive, no?

You're right @Arnei, setting `copy` to `true` is not a good idea. I overlooked that with this setting, each video resolution is published twice for videos without branding. I'll change the PR.

snoesberger:
I have now changed the tagging workflow step to simply change the flavour of `presenter/trimmed` and `presentation/trimmed` to `*/themed`. That way, `captions/trimmed` will still be available for publication afterwards.
This way no copy of the videos is necessary.

mtneug:
Note: This PR is targeting the legacy branch of Opencast (`r/15.x`). In [a recently passed proposal](https://groups.google.com/a/opencast.org/g/dev/c/VwYqk3vXdd0), we changed how we deal with PRs to legacy branches.

Our goal is to provide stability and minimize the risk of breaking changes. Ultimately, users should trust that patch releases of legacy versions can be installed easily.

To this end, developers targeting legacy branches now need to explain why those changes are necessary and can't just go into stable or develop. We'd appreciate it if you could provide such an explanation or, alternatively, retarget this PR to stable or develop.

snoesberger:
I think it is worth adding this PR to r/15x because it fixes a bug that makes it impossible to publish subtitles (and subtitles are officially supported in r/15.x), see also my explanations in https://github.com/opencast/opencast/issues/6138.

PR #6170, opened 2024-09-17
Skip elements during search-index rebuild that throw a NotFoundException

jduehring:
If during a rebuild of the search-index a NotFoundException is thrown, right now the rebuild will just stop and not process any more events. So in a cluster with a lot of events it is enough if there is one broken mediapackages in the system (for example caused by a faulty retraction, missing dublincore etc.) to not be able to continue the rebuild. 

With this patch instead of throwing a NotFoundException and stopping the process, the error gets logged and the element is beeing skipped, so that the rebuild can finish regardless.

### Your pull request should…

* [x] have a concise title
* [ ] pass automated tests


lkiesow:
For anyone else confused about this: This is only about the search service search index.
Looks reasonable. Will test this in a minute…

lkiesow (approved):

PR #6177, opened 2024-09-23
Proposal: standard JWT schema and path for better JWT support

LukasKalbertodt:
This pull request suggests a path forward for Opencast to improve JWT support and cover a wide range of use cases, while also already implementing the first step of that plan. So this PR consists of not only the git diff, but also of the description containing the full proposal. I intend to discuss this with the community, in particular also discussing this in person on the D/A/CH conference tomorrow.

---

## Why?

There are many parties interested in better JWT support. [This discussion](https://github.com/opencast/opencast/discussions/5334) has been open for nearly a year now, while need for this has been around for way longer. Tools like LMSs and Tobira require Opencast to disable static file authorization completely in order to work. The annotation tool might require similarly problematic workarounds. And external apps cannot properly link to the Opencast editor (granting access to edit a video). There is also interest in having a service that can serve static Opencast files without needing a database.

At elan, many clients asked us about these use cases and specifically about JWT in Opencast. The GitHub discussion thread was an attempt to get this started, but I think at this point, one simply has to start working on it.

## What?

Talking to various parties, thinking about different use cases, and examining what Opencast is already capablable of, I think we should take roughly the following steps to improve JWT support. Each of these is described further down.

1. Define a standard JWT schema and make it work out-of-the-box in Opencast (That's what this PR does!)
2. Add proper way to use JWTs in SFA (static file auth)
3. Add ability to merge login sessions
4. Create service for serving static files with authorization that keeps working when Opencast is down

EDIT: see [this comment](https://github.com/opencast/opencast/pull/6177#issuecomment-2377204640) for a change of priorities.

### (1) Standard JWT schema

After the cryptographic verification, a JWT is just a bunch of "claims" (JSON fields). How to interpret those fields is up to the communicating services. Since there are quite a few external projects in the Opencast ecosystem (Tobira, at least 3 LMS plugins, Annotation Tool), agreeing on a standard is reasonable to avoid diverging uses of JWTs. Having such a standard also means that we can make it work out of the box, without the need for admins to write complex role mappings.

This is what this PR does. Please check the git diff to read the actual schema proposal.

I think having one standard for all use cases makes sense. One could argue that one shouldn't mix static file auth with general user authentication/authorization. But this doesn't do that. This just defines a schema in which both, or a mix of the two, can be expressed. For static file auth, one would onle specify a single `"oc": ["read:e:<id>"]`, while for usage of Studio you would specify user info an `"roles": ["STUDIO"]`. For usage of the editor, you can mix both: give specific roles, but also write access to a specific event. Having use case specific JWT schemata could possibly make the JWTs for each use case smaller, but only by a a few bytes. And personally, I think this tiny overhead is acceptable in order to have only one schema.

This PR also implements this standard as `ocStandardRoleMappings`, so that it's not necessary anymore to write role mappings manually. 

With this done, external tools can start using this standard to attach JWTs to requests. And in Opencast, we can improve support for certain use cases.

### (2) JWTs in SFA

While step (1) already makes this possible, it is far from perfect since the JWT handler require a username, display name and email for sessions created by the JWT. For static file authorization, all of this shouldn't be necessary. So in step 2, we should adjust Opencast's SFA to correctly work with JWTs that do not contain user info, but just a single `oc` entry giving access to one event. This can also improve performance of static file serving, as in the JWT case, loading the event's ACL is often not necessary.

I tried implementing this already, but didn't come very far due to lacking familiarity with Spring security and Opencast's auth system. But I think implementation details can be done at a later stage anyway.

### (3) Merge login sessions

Currently, creating a login session via JWT while another login session exists either does nothing or, if you set `checkForPrincipalChanges` to true, it fully replaces the other session. This is not ideal. Consider someone opening the video editor via link in Tobira for two different videos in two different browser tabs. The first session that's created grants access to one video, the second one to the second video, rendering the first editor tab broken. There are a number of other scenarios where this can lead to problems.

Session merging should merge roles between the existing and new session, so that permissions stack. Details need to be figured out.

### (4) External static file serving application

The goal is to create a small program that, together with a webserver like nginx, can serve Opencast's static files, even while Opencast is down -- without disabling authorization! This external service would only support JWT authorization, as well as serving public files without requiring a JWT. But even this limited auth support can cover a wide range of use cases. If the request cannot be authorized via JWT, the service would forward it to Opencast to let it decide.

This has three main advantages:

- **Reducing downtime**: be it crashes or updates, Opencast downtime is just a fact of life and can become quite disruptive for institutions with large data sets during updates. With this file server, we can significantly reduce the *perceived* downtime for most users. The vast majority of time during "Opencast usage" is spent watching a video, i.e. downloading static files. And generally, most usage is read-only. Apps like Tobira or some LMSs keep a local copy of Opencast metadata and partially keep functioning during Opencast downtimes (this was an explicit requirement of Tobira). Keeping file serving working without Opencast, means that most users won't notice Opencast being down. 

- **Improving performance**: involving Opencast (and its whole framework/stack) in file serving, makes the operation a lot slower and use up a lot more computing resources. We are talking three-digit factors or more, compared to just nginx file serving. 

- **Stateless file serving**: this would make file serving less dependent on external factors, making it easier to scale it, deploy it to servers at different locations, use external storage solutions and the like. 

For this to work well, Opencast should adopt some changes to its static file URL pattern and its NFS file paths. Public files (i.e. from events that `ROLE_ANONYMOUS` can read) should be placed in the file system in a special way such that an external service can easily identify them. Ideally the URL pattern is also adjusted. And for protected files, it should be possible to omit the ID from the URL path, since the JWT will contain the event's ID anyway and we don't want to specify it twice. Both these techniques are commonly used by big CDNs. 

## Open questions

- Should we skip step 2 in favor of step 4? Once we have the external service, everyone should use it instead of letting Opencast serve. It just felt correct to have this work directly with Opencast itself. Or is that just wasted work?
- When and when not to load existing user data for the supplied username? Currently the JWT login handler looks up user info (including roles) in the user reference table. We should decide when that should happen, and whether only the user references should be checked.
- Similarly, decide when the user session created by a JWT is stored as user reference. For static file serving, it shouldn't. For Studio, Editor and other use cases, it should. Should it be saved whenever the user information is supplied in the JWT?
- Maybe use `kid` claim (key id) to avoid having to check multiple keys? That can happen in the far future though.

## Other TODOs

Here is an unordered list of other things that still need to happen.

- (in this PR) add example in security config for returning list of roles
- (in this PR) stop parsing SPeL expressions on every JWT request, but only once, using the parsed representation later.
- Frontends/players need to be adjusted to deal with expired JWTs. If a user stops watching a video and returns a day later (keeping the tab open), that should work. 

## Background info: JWTs in web servers

Webservers like nginx can be configured to verify JWTs. The important summary is that the business-logic claim validation is always deferred to some user specified, turing-complete "script". So we are not constrained by this when it comes to the design of our JWT schema. 

<details>

- Nginx
    - Built-in: https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-jwt-authentication
        - only in nginx commercial solution
        - Can use external JWKS, with configurable caching
        - Verifies `exp` and `nbf`
        - Support for custom claim validation. Logic is encoded in `map` of externally via auth sub requests.
    - https://github.com/kjdev/nginx-auth-jwt
        - Seems to be basically an open source version of the official stuff?
    - https://github.com/TeslaGov/ngx-http-auth-jwt-module
        - Seems to be intended for user sessions (JWT is bad for that)
        - Cannot use JWKS URLs
        - Can extract claims into variables, custom checking in nginx code. But only string claims! (no number, bool, array, object)
        - -> Doesn't seem that great
- Apache
    - Seems to have only very basic module built in... dunno

It seems like many people would use their own script/server to validate JWTs. Webservers would then call into that (via auth subrequest for example) to determine whether to serve a file. And then, super custom logic is possible of course.

</details>

## Next steps

I invite everyone to read this proposal and give feedback. This will also be discussed at D/A/CH starting tomorrow. My intention is to get this started soon though. I don't want this PR to bit rod for half a year. So if you want to be involved in this discussion, please do so fairly quickly. In fact, I hope that after D/A/CH we will have agreed on a general direction.

### DIFF
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -40,10 +40,10 @@ This schema is designed to be flexible and support all current and conceivable f
 
     // Authorization, i.e. privileges the request/user should have
     "roles": ["ROLE_STUDIO", "ROLE_API_EVENTS_VIEW"],
-    "oc": [
-        "read+customaction:e:d622b861-4264-4947-8db1-c754c5956433",
-        "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"
-    ]
+    "oc": {
+        "e:d622b861-4264-4947-8db1-c754c5956433": ["read, "annotate"],
+        "s:4ed02421-144c-42a1-b98a-22e84f3ac691": ["write"]
+    }
 }
 ```
 
@@ -62,15 +62,12 @@ A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to i
 
 - *Authorization*: at least one of these claims has to be set, as otherwise no privileges are granted (which is the same as not sending a JWT at all).
     - `roles`: array of strings, specifying roles to grant to the request/user directly.
-    - `oc`: array of strings, grants access to single items in Opencast. The custom syntax is terse in order to keep JWTs as small as possible. Each string consists of three parts, which are separated by a colon (`:`):
-        - *Action(s)*: the first part specifies what actions are granted, corresponding to actions in the OC ACLs, e.g. `read` and `write`. Multiple actions can be specified by joining them with `+`. To keep the syntax unambigious, actions cannot contain `:` or `+` (ideally they are ASCII alphabetic only!). To further preserve space, some actions imply others:
-            - `write` implies `read` access
-            - `annotate` implies `read` access
-        - *Item Type*: The second part specifies what type of item this applies to. Note: giving access to a series or playlist only gives access to that item directly, and *not* to its videos (e.g. give write access for a series to be able to edit its metadata or add videos to it).
+    - `oc`: JSON map that grants access to single items in Opencast, as identified by the map key.
+        - The map key consists of a one-letter prefix (the item type), a colon, and finally the actual ID of the Opencast item. The following item types are supported. Note: giving access to a series or playlist only gives access to that item directly, and *not* to its videos (e.g. give write access for a series to be able to edit its metadata or add videos to it).
             - `e`: event
             - `s`: series
             - `p`: playlist
-        - *ID*: the final part specifies the ID of the item that is granted access to.
+        - The map value is just an array of actions, corresponding to actions in the OC ACLs, e.g. `read` and `write`.
 
 
 ### Examples
@@ -82,7 +79,7 @@ An external application wants a user to load a protected static file from Openca
 ```json
 {
     "exp": 1499863217,
-    "oc": ["read:e:d622b861-4264-4947-8db1-c754c5956433"]
+    "oc": { "e:d622b861-4264-4947-8db1-c754c5956433": ["read"] }
 }
 ```
 
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -1,17 +1,106 @@
 Configuration for JWT-based Authentication and Authorization
 ============================================================
 
-This page describes how to configure Opencast to enable authentication and authorization based on
-[JSON Web Tokens (JWTs)](https://datatracker.ietf.org/doc/html/rfc7519). With this feature, a login-mechanism based
-on the [OpenID Connect (OIDC)](https://openid.net/connect/) protocol can be configured, since OIDC uses JWTs.
+This page describes how to configure Opencast to enable authentication and/or authorization based on
+[JSON Web Tokens (JWTs)](https://datatracker.ietf.org/doc/html/rfc7519). JWTs can be used in various ways:
+
+- With static file authorization (particularly useful with external tools).
+- External apps can send users to Opencast (e.g. for Studio or Editor) with a session that has certain roles.
+- A login-mechanism based on the [OpenID Connect (OIDC)](https://openid.net/connect/) protocol can be configured.
+
+*Note*: Some of these use cases might not be perfectly supported by Opencast yet, but are possible in principle.
 
 Prerequisites
 -------------
 
-This guide assumes that a JWT provider is already setup and Opencast receives the JWT within an HTTP request header,
-or in a request parameter (i.e. a query parameter for `GET`-requests, and a form parameter for `POST`-requests).
+Opencast does not generate JWTs, but reads them in incoming requests. This guide assumes that you already have a JWT provider, i.e. another services that generates JWTs for consumption in Opencast. This could be [Tobira](https://elan-ev.github.io/tobira/setup/auth/jwt), an LMS, an OIDC provider, or something else. These generally have their own setup guides. The JWT has to be passed to Opencast via HTTP header or via request parameter (i.e. a query parameter for `GET`-requests, and a form parameter for `POST`-requests).
+
 In order to integrate Opencast with an OIDC provider you could use the
 [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy).
+
+Standard OC Schema for JWTs
+---------------------------
+
+After the signature verification, a JWT is just a bunch of "claims" (JSON fields). How to interpret these claims is completely up to the communicating parties. Protocols like OIDC define exactly which claims are to be used and how. For other cases where external Opencast apps (like Tobira or an LMS) need to communicate with Opencast, this section defines a standard schema to follow. You can configure Opencast differently, but this aims at standardizing JWT usage across Opencast-related applications and offer an out-of-the-box solution.
+
+This schema is designed to be flexible and support all current and conceivable future use cases, while keeping the JWT size for each use case small.
+
+### Quick overview
+
+```json5
+{
+    // Standard claims
+    "exp": 1499863217, // required
+    "nbf": 1548632170, // optional
+
+    // User info
+    "username": "jose",
+    "name": "José Carreño Quiñones",
+    "email": "jose@example.com",
+
+    // Authorization, i.e. privileges the request/user should have
+    "roles": ["STUDIO", "API_EVENTS_VIEW"], // without `ROLE_` prefix
+    "oc": [
+        "read+customaction:e:d622b861-4264-4947-8db1-c754c5956433",
+        "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"
+    ]
+}
+```
+
+### Claims
+
+A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to include it. Opencast will reject all JWTs which lack required claims.
+
+- `exp` (expiration time): As [defined in the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4), a number timestamp in seconds since the unix epoch. JWTs with `exp < now()` are rejected by Opencast. This claim is *required*.
+
+- `nbf` (not before): As [defined in the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5), a number timestamp in seconds since the unix epoch. JWTs with `nbf > now()` are rejected by Opencast. This claim is optional.
+
+- *User Info*: information about the user that's possibly used by parts of Opencast. Some JWT use cases don't require user information at all (e.g. static file auth), while others do (e.g. Studio). Therefore, these claims are optional. (TODO: specify when this is stored as user reference)
+    - `username`: string. (TODO: explain how this is used to lookup existing users, if at all)
+    - `name`: display name, string.
+    - `email`: string.
+
+- *Authorization*: at least one of these claims has to be set, as otherwise no privileges are granted (which is the same as not sending a JWT at all).
+    - `roles`: array of strings, specifying roles to grant to the request/user directly. The `ROLE_` prefix must be omitted as Opencast adds it itself.
+    - `oc`: array of strings, grants access to single items in Opencast. The custom syntax is terse in order to keep JWTs as small as possible. Each string consists of three parts, which are separated by a colon (`:`):
+        - *Action(s)*: the first part specifies what actions are granted, corresponding to actions in the OC ACLs, e.g. `read` and `write`. Multiple actions can be specified by joining them with `+`. To keep the syntax unambigious, actions cannot contain `:` or `+` (ideally they are ASCII alphabetic only!). To further preserve space, some actions imply others:
+            - `write` implies `read` access
+            - `annotate` implies `read` access
+        - *Item Type*: The second part specifies what type of item this applies to:
+            - `e`: event
+            - `s`: series
+            - `p`: playlist
+        - *ID*: the final part specifies the ID of the item that is granted access to.
+
+
+### Examples
+
+#### Static file authorization
+
+An external application wants a user to load a protected static file from Opencast, e.g. a video file. The request needs read permission for the that event. The external app can achieve this with the following JWT (which can simply be attached to the URL as query parameter):
+
+```json
+{
+    "exp": 1499863217,
+    "oc": ["read:e:d622b861-4264-4947-8db1-c754c5956433"]
+}
+```
+
+#### Opencast Studio
+
+An external application wants to let a user use Opencast Studio to record and upload a video. Since this does not happen in a single request, a browser session needs to be created. This can be achieved with the [`/redirect` API](https://develop.opencast.org/docs.html?path=/redirect). The JWT sent to that API needs to grant permissions to use all required APIs (which can be done by granting `ROLE_STUDIO`) and also specify user info, since Studio uses some use information.
+
+```json
+{
+    "exp": 1499863217,
+    "username": "peter",
+    "name": "Peter Lustig",
+    "email": "peter@example.com",
+    "roles": ["STUDIO"]
+}
+```
+
+
 
 Spring Security Configuration
 -----------------------------
@@ -89,6 +178,10 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
   <property name="nameMapping" value="['name'].asString()" />
   <!-- Mapping used to extract the email from the JWT (default: null) -->
   <property name="emailMapping" value="['email'].asString()" />
+  <!-- Opencast standard role mapping as defined above in this chapter of the docs. -->
+  <!-- I.e. reads `roles` and `oc` claims to specify roles. -->
+  <property name="ocStandardRoleMappings" value="true" />
+  <!-- Custom role mappings, optional -->
   <property name="roleMappings" ref="jwtRoleMappings" />
   <!-- JWT cache holds already validated JWTs to not always re-validate in subsequent requests -->
   <!-- Maximum number of validated JWTs to cache (default: 500) -->
@@ -126,8 +219,8 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
   <value>['affiliation'].asList(T(String)).contains('faculty@example.org')</value>
 </util:list>
 ```
-* Configure the `jwtRoleMappings` list. This list contains expressions used to construct Opencast roles from JWT
-  claims.
+* Optionally configure the `jwtRoleMappings` list. This list contains expressions used to construct Opencast roles from JWT
+  claims. This can be used instead of or in addition to the OC standard role mapping. Each entry in this list contains a SPeL expression that needs to return `null`, or a string, or a `List<String>`, or `String[]`.
 ```xml
 <!-- The mapping from JWT claims to Opencast roles -->
 <util:list id="jwtRoleMappings" value-type="java.lang.String">
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -39,7 +39,7 @@ This schema is designed to be flexible and support all current and conceivable f
     "email": "jose@example.com",
 
     // Authorization, i.e. privileges the request/user should have
-    "roles": ["STUDIO", "API_EVENTS_VIEW"], // without `ROLE_` prefix
+    "roles": ["ROLE_STUDIO", "ROLE_API_EVENTS_VIEW"],
     "oc": [
         "read+customaction:e:d622b861-4264-4947-8db1-c754c5956433",
         "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"
@@ -61,7 +61,7 @@ A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to i
     - `email`: string.
 
 - *Authorization*: at least one of these claims has to be set, as otherwise no privileges are granted (which is the same as not sending a JWT at all).
-    - `roles`: array of strings, specifying roles to grant to the request/user directly. The `ROLE_` prefix must be omitted as Opencast adds it itself.
+    - `roles`: array of strings, specifying roles to grant to the request/user directly.
     - `oc`: array of strings, grants access to single items in Opencast. The custom syntax is terse in order to keep JWTs as small as possible. Each string consists of three parts, which are separated by a colon (`:`):
         - *Action(s)*: the first part specifies what actions are granted, corresponding to actions in the OC ACLs, e.g. `read` and `write`. Multiple actions can be specified by joining them with `+`. To keep the syntax unambigious, actions cannot contain `:` or `+` (ideally they are ASCII alphabetic only!). To further preserve space, some actions imply others:
             - `write` implies `read` access
@@ -96,7 +96,7 @@ An external application wants to let a user use Opencast Studio to record and up
     "username": "peter",
     "name": "Peter Lustig",
     "email": "peter@example.com",
-    "roles": ["STUDIO"]
+    "roles": ["ROLE_STUDIO"]
 }
 ```
 
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -34,7 +34,7 @@ This schema is designed to be flexible and support all current and conceivable f
     "nbf": 1548632170, // optional
 
     // User info
-    "username": "jose",
+    "sub": "jose", // username
     "name": "José Carreño Quiñones",
     "email": "jose@example.com",
 
@@ -56,7 +56,7 @@ A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to i
 - `nbf` (not before): As [defined in the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5), a number timestamp in seconds since the unix epoch. JWTs with `nbf > now()` are rejected by Opencast. This claim is optional.
 
 - *User Info*: information about the user that's possibly used by parts of Opencast. Some JWT use cases don't require user information at all (e.g. static file auth), while others do (e.g. Studio). Therefore, these claims are optional. (TODO: specify when this is stored as user reference)
-    - `username`: string. (TODO: explain how this is used to lookup existing users, if at all)
+    - `sub`: Opencast username, string. (TODO: explain how this is used to lookup existing users, if at all)
     - `name`: display name, string.
     - `email`: string.
 
@@ -93,7 +93,7 @@ An external application wants to let a user use Opencast Studio to record and up
 ```json
 {
     "exp": 1499863217,
-    "username": "peter",
+    "sub": "peter",
     "name": "Peter Lustig",
     "email": "peter@example.com",
     "roles": ["ROLE_STUDIO"]
@@ -173,7 +173,7 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
   <property name="expectedAlgorithms" ref="jwtExpectedAlgorithms" />
   <property name="claimConstraints" ref="jwtClaimConstraints" />
   <!-- Mapping used to extract the username from the JWT (default: null) -->
-  <property name="usernameMapping" value="['username'].asString()" />
+  <property name="usernameMapping" value="['sub'].asString()" />
   <!-- Mapping used to extract the name from the JWT (default: null) -->
   <property name="nameMapping" value="['name'].asString()" />
   <!-- Mapping used to extract the email from the JWT (default: null) -->
@@ -207,7 +207,7 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
 <util:list id="jwtClaimConstraints" value-type="java.lang.String">
   <value>containsKey('iss')</value>
   <value>containsKey('aud')</value>
-  <value>containsKey('username')</value>
+  <value>containsKey('sub')</value>
   <value>containsKey('name')</value>
   <value>containsKey('email')</value>
   <value>containsKey('domain')</value>
@@ -225,8 +225,8 @@ sections found in `etc/security/mh_default_org.xml`. Some of the options are con
 <!-- The mapping from JWT claims to Opencast roles -->
 <util:list id="jwtRoleMappings" value-type="java.lang.String">
   <value>'ROLE_JWT_USER'</value>
-  <value>'ROLE_JWT_USER_' + ['username'].asString()</value>
-  <value>('ROLE_JWT_OWNER_' + ['username'].asString()).replaceAll("[^a-zA-Z0-9]","_").toUpperCase()</value>
+  <value>'ROLE_JWT_USER_' + ['sub'].asString()</value>
+  <value>('ROLE_JWT_OWNER_' + ['sub'].asString()).replaceAll("[^a-zA-Z0-9]","_").toUpperCase()</value>
   <value>['domain'] != null ? 'ROLE_JWT_ORG_' + ['domain'].asString() + '_MEMBER' : null</value>
   <value>['username'].asString() eq ('j_doe01@example.org') ? 'ROLE_ADMIN' : null</value>
   <value>['affiliation'].asList(T(String)).contains('faculty@example.org') ? 'ROLE_GROUP_JWT_TRAINER' : null</value>
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -66,7 +66,7 @@ A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to i
         - *Action(s)*: the first part specifies what actions are granted, corresponding to actions in the OC ACLs, e.g. `read` and `write`. Multiple actions can be specified by joining them with `+`. To keep the syntax unambigious, actions cannot contain `:` or `+` (ideally they are ASCII alphabetic only!). To further preserve space, some actions imply others:
             - `write` implies `read` access
             - `annotate` implies `read` access
-        - *Item Type*: The second part specifies what type of item this applies to:
+        - *Item Type*: The second part specifies what type of item this applies to. Note: giving access to a series or playlist only gives access to that item directly, and *not* to its videos (e.g. give write access for a series to be able to edit its metadata or add videos to it).
             - `e`: event
             - `s`: series
             - `p`: playlist
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -21,7 +21,7 @@ In order to integrate Opencast with an OIDC provider you could use the
 Standard OC Schema for JWTs
 ---------------------------
 
-After the signature verification, a JWT is just a bunch of "claims" (JSON fields). How to interpret these claims is completely up to the communicating parties. Protocols like OIDC define exactly which claims are to be used and how. For other cases where external Opencast apps (like Tobira or an LMS) need to communicate with Opencast, this section defines a standard schema to follow. You can configure Opencast differently, but this aims at standardizing JWT usage across Opencast-related applications and offer an out-of-the-box solution.
+After the signature verification, a JWT is just a bunch of "claims" (JSON fields). The communicating parties can decide fairly freely on how to interpret these claims. Protocols like OIDC define exactly which claims are to be used and how. For other cases where external Opencast apps (like Tobira or an LMS) need to communicate with Opencast, this section defines a standard schema to follow. You can configure Opencast differently, but this aims at standardizing JWT usage across Opencast-related applications and offer an out-of-the-box solution.
 
 This schema is designed to be flexible and support all current and conceivable future use cases, while keeping the JWT size for each use case small.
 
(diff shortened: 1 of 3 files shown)


mtneug:
@LukasKalbertodt thank you very much for this proposal and the discussion at the DACH 2024 meeting!

Some thoughts:

## Comments

>  external service would only support JWT authorization, as well as serving public files without requiring a JWT. But even this limited auth support can cover a wide range of use cases. If the request cannot be authorized via JWT, the service would forward it to Opencast to let it decide.

In what scenario can the external service not decide based on JWTs and needs to forward requests to Opencast? I would suggest simplifying and only supporting JWT authorization for serving files. Instead of forwarding to Opencast, simply fail the request with 403.

> For this to work well, Opencast should adopt some changes to its static file URL pattern and its NFS file paths.

Please don't forget S3 as distribution option. It basically uses the same path as the `fs` distribution for the object prefix, so the same thoughts apply to it.

> Public files (i.e. from events that ROLE_ANONYMOUS can read) should be placed in the file system in a special way such that an external service can easily identify them. Ideally, the URL pattern is also adjusted. And for protected files, it should be possible to omit the ID from the URL path, since the JWT will contain the event's ID anyway and we don't want to specify it twice. Both these techniques are commonly used by big CDNs.

This is tricky:

* We decided at the committers meeting to store ACLs in the DB and not require a republish when changing them. However, with your suggestion adding `ROLE_ANONYMOUS` would require a republish. Moving files might be fast for NFS, but this is not the case for S3 as you would need to reupload files. Therefore, ideally, you would not need to move files around when changing ACLs.
* Changing the path (and thus the URL) is tricky for existing installations. Changing the element URLs would mean that we need to republish every event to update the URLs in the search service. Also, you need to change the files on disk / in S3 (i.e. reupload everything). Imagine doing this as Manchester (although they don't use search IIRC)!

I would, therefore, very much like to leave files at their current location and find a solution for public vs protected files. Maybe just require ACLs to be published and read them from disk/S3 + in-memory caching in the case where no JWT is given? The `fast` workflow doesn't seem to publish ACLs, but `schedule-and-upload` does.

> Should we skip step 2 in favor of step 4? Once we have the external service, everyone should use it instead of letting Opencast serve. It just felt correct to have this work directly with Opencast itself. Or is that just wasted work?

+1 and I think this was the consensus at DACH.

## How should the external service look like?

We could implement it in various ways (maybe even implement it multiple times if we cannot decide on one):

* Webserver plugin: JWT validation is implemented as a plugin for the webserver (e.g. nginx module, Caddy plugin, ...).
* Service running next to the webserver and using something like [`auth_request`](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html), i.e. the webserver would send a request to the service and it would answer if the original request should be allowed or not.
* Service running as proxyed server and using something like `X-Accel-Redirect` headers.

## User References & Sessions

JWT authn currently creates user references in Opencast and a new session tracked by a cookie. User references are necessary if this is the only way to load a user, e.g. during a workflow execution. Sessions are necessary for further requests of the user if those requests don't attach the JWT again. For the simple use case of serving files, both user references and sessions (as stored in cookies) are not necessary and I would like to avoid those. Usually, there are many potential users who watch videos, but comparatively few who are able to add new ones / start workflows. In the interest of keeping the DB tables small, there shouldn't be user references for watching-only users.

> Similarly, decide when the user session created by a JWT is stored as user reference. For static file serving, it shouldn't. For Studio, Editor and other use cases, it should. Should it be saved whenever the user information is supplied in the JWT?

I think this makes sense.

## JWT Schema

> Maybe use `kid` claim (key id) to avoid having to check multiple keys? That can happen in the far future though.

This shouldn't be hard to implement and I would also suggest to allow keys that the schema does not specify. The implementation should just ignore those claims.

I will add some more comments to the code lines.

## JWKS

Maybe we should implement a / use an existing JWKS service. Tobira is doing this internally already (I learned), but we need a central one that services public keys for all applications that will create JWTs (LMS, Tobira, etc.).

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+After the signature verification, a JWT is just a bunch of "claims" (JSON fields). How to interpret these claims is completely up to the communicating parties. Protocols like OIDC define exactly which claims are to be used and how. For other cases where external Opencast apps (like Tobira or an LMS) need to communicate with Opencast, this section defines a standard schema to follow. You can configure Opencast differently, but this aims at standardizing JWT usage across Opencast-related applications and offer an out-of-the-box solution.
+
+This schema is designed to be flexible and support all current and conceivable future use cases, while keeping the JWT size for each use case small.
+
+### Quick overview
+
+```json5
+{
+    // Standard claims
+    "exp": 1499863217, // required
```
mtneug, line 33: Required means that we will reject JWTs without `exp`, right?

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+    "nbf": 1548632170, // optional
+
+    // User info
+    "username": "jose",
```
mtneug: I would suggest to use `sub` instead. At least `username` is [not a registered claim](https://www.iana.org/assignments/jwt/jwt.xhtml#claims).

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+    "name": "José Carreño Quiñones",
+    "email": "jose@example.com",
+
+    // Authorization, i.e. privileges the request/user should have
+    "roles": ["STUDIO", "API_EVENTS_VIEW"], // without `ROLE_` prefix
```
mtneug: I know that you want to keep the JWT short, but I would like to avoid magic steps. As such, I'm in favor of always including full role names.

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+### Claims
+
+A claim being "required" means that the JWT provider (e.g. Tobira, LMS) has to include it. Opencast will reject all JWTs which lack required claims.
+
+- `exp` (expiration time): As [defined in the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4), a number timestamp in seconds since the unix epoch. JWTs with `exp < now()` are rejected by Opencast. This claim is *required*.
+
+- `nbf` (not before): As [defined in the JWT RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5), a number timestamp in seconds since the unix epoch. JWTs with `nbf > now()` are rejected by Opencast. This claim is optional.
+
+- *User Info*: information about the user that's possibly used by parts of Opencast. Some JWT use cases don't require user information at all (e.g. static file auth), while others do (e.g. Studio). Therefore, these claims are optional. (TODO: specify when this is stored as user reference)
+    - `username`: string. (TODO: explain how this is used to lookup existing users, if at all)
```
mtneug: This should happen with the usual user/role providers (if we use the JWT for authn).

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+    - `name`: display name, string.
+    - `email`: string.
+
+- *Authorization*: at least one of these claims has to be set, as otherwise no privileges are granted (which is the same as not sending a JWT at all).
+    - `roles`: array of strings, specifying roles to grant to the request/user directly. The `ROLE_` prefix must be omitted as Opencast adds it itself.
+    - `oc`: array of strings, grants access to single items in Opencast. The custom syntax is terse in order to keep JWTs as small as possible. Each string consists of three parts, which are separated by a colon (`:`):
+        - *Action(s)*: the first part specifies what actions are granted, corresponding to actions in the OC ACLs, e.g. `read` and `write`. Multiple actions can be specified by joining them with `+`. To keep the syntax unambigious, actions cannot contain `:` or `+` (ideally they are ASCII alphabetic only!). To further preserve space, some actions imply others:
```
mtneug: Know actions include `read`, `write`, `annotate` and `annotate-admin`. I think TU Vienna uses some custom actions and I'm defining one as well. I don't think we restrict the action name currently, but at the same time, it's probably unlikely that `+` and `:` is used.

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+            - `write` implies `read` access
+            - `annotate` implies `read` access
```
mtneug: I'm not a fan of this as this again feels like magic and I don't want auth to rely on implicit rules.

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+        - *Item Type*: The second part specifies what type of item this applies to:
+            - `e`: event
+            - `s`: series
+            - `p`: playlist
+        - *ID*: the final part specifies the ID of the item that is granted access to.
```
mtneug: Not sure, but IDs can probably also contain `:`. Shouldn't be a problem, if this string is parsed correctly.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
```diff
@@ -261,11 +268,78 @@ private String extractEmail(DecodedJWT jwt) {
...
+        var rolesClaim = jwt.getClaim("roles");
+        if (rolesClaim != null && !rolesClaim.isNull()) {
+          for (String r : rolesClaim.asArray(String.class)) {
+            if (!r.isBlank()) {
+              addRole.accept("ROLE_" + r);
+            }
+          }
+        }
+      } catch (JWTDecodeException e) {
+        logger.warn("claim 'roles' is not an array of strings, ignoring");
```
mtneug: I would suggest using `logger.debug` (also below). This shouldn't happen if the integration is using a correct schema, but if not this will fill the logs for adopters.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
```diff
@@ -261,11 +268,78 @@ private String extractEmail(DecodedJWT jwt) {
...
+            var id = parts[2];
+            for (var action : actions.split("\\+")) {
+              if (action.isBlank()) {
+                continue;
+              }
+
+              if (type.equals("e")) {
+                addRole.accept(SecurityUtil.getEpisodeRoleId(id, action));
+              } else {
+                logger.warn("in 'oc' claim: granting access to item type '{}' is not yet supported", type);
```
mtneug: Implementing series ACL for external services would mean reading the metadata from disk/S3 (if we still publish the dublincore XML) to determine if the fetched event is from this series. I don't know how to implement this in external services for playlists.

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+
 In order to integrate Opencast with an OIDC provider you could use the
 [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy).
 
+Standard OC Schema for JWTs
+---------------------------
+
+After the signature verification, a JWT is just a bunch of "claims" (JSON fields). How to interpret these claims is completely up to the communicating parties. Protocols like OIDC define exactly which claims are to be used and how. For other cases where external Opencast apps (like Tobira or an LMS) need to communicate with Opencast, this section defines a standard schema to follow. You can configure Opencast differently, but this aims at standardizing JWT usage across Opencast-related applications and offer an out-of-the-box solution.
```
mtneug: > How to interpret these claims is completely up to the communicating parties.

We do have RFCs that specify constraints and give definitions for certain claims. I would suggest removing or altering this sentence.

LukasKalbertodt:
We disused this two days ago on the DACH (un)conference. We didn't have much time to really dive deep, but the main points I took home were:

- Generally, the plan makes sense and there is no big blocker that I overlooked.
- The static file use cases should be prioritized over the tool use case for now. Especially since for some LMS, LTI already covers the tool use case.
- The tool use case requires additional thought (regarding session merging, username matching, ...). Also a good reason to first do static file auth.
- Regardless of priorities, people felt that using the suggested JWT schema, which is very general and can support far more than static file auth, is fine. 
- The question of whether omitting `ROLE_` is fine was brought up, but hasn't been discussed further since at the end of the day, this part of the schema doesn't affect anything else.

LukasKalbertodt:
Now replying to Matthias:

> In what scenario can the external service not decide based on JWTs and needs to forward requests to Opencast? I would suggest simplifying and only supporting JWT authorization for serving files. Instead of forwarding to Opencast, simply fail the request with 403.

In cases where there is an Opencast Cookie-based session (i.e. JSESSION). This would be the case if someone simply logged into the admin UI, watching videos from there, for example. Eventually, I wouldn't mind doing everything via JWT! I just didn't want to make such a bold move now already, since it would require adjusting a few apps (like the admin UI). 

> > Public files (i.e. from events that ROLE_ANONYMOUS can read) should be placed in the file system in a special way such that an external service can easily identify them. Ideally, the URL pattern is also adjusted. And for protected files, it should be possible to omit the ID from the URL path, since the JWT will contain the event's ID anyway and we don't want to specify it twice. Both these techniques are commonly used by big CDNs.
> 
> This is tricky:
> 
> * We decided at the committers meeting to store ACLs in the DB and not require a republish when changing them. However, with your suggestion adding `ROLE_ANONYMOUS` would require a republish. Moving files might be fast for NFS, but this is not the case for S3 as you would need to reupload files. Therefore, ideally, you would not need to move files around when changing ACLs.
> * Changing the path (and thus the URL) is tricky for existing installations. Changing the element URLs would mean that we need to republish every event to update the URLs in the search service. Also, you need to change the files on disk / in S3 (i.e. reupload everything). Imagine doing this as Manchester (although they don't use search IIRC)!
> 
> I would, therefore, very much like to leave files at their current location and find a solution for public vs protected files. Maybe just require ACLs to be published and read them from disk/S3 + in-memory caching in the case where no JWT is given? The `fast` workflow doesn't seem to publish ACLs, but `schedule-and-upload` does.

Mhhhh yeah, I suppose I forgot about S3 and indeed assumed doing these "moving" operations are super fast. Mh this is tricky. I suppose it would be nice if Opencast would write the ACLs to the FS somehow so that this external service can read them. But I'm afraid this adds lot of complexity and depenencies between the service and OC. I would probably suggest that for now, we treat public files just like protected files and always require a JWT. Then we can see how much overhead the JWT generation and validation is and how strongly we should add the "public file optimization". That's after all all it is: an optimization to avoid some JWT operations. We can always add it later.


> > Should we skip step 2 in favor of step 4? Once we have the external service, everyone should use it instead of letting Opencast serve. It just felt correct to have this work directly with Opencast itself. Or is that just wasted work?
> 
> +1 and I think this was the consensus at DACH.

I'm not sure I took that as a consensus from DACH. Not that I mind, I just don't remember having such a clear answer on it.

 
> ## How should the external service look like?
> We could implement it in various ways (maybe even implement it multiple times if we cannot decide on one):
> 
> * Webserver plugin: JWT validation is implemented as a plugin for the webserver (e.g. nginx module, Caddy plugin, ...).
> * Service running next to the webserver and using something like [`auth_request`](https://nginx.org/en/docs/http/ngx_http_auth_request_module.html), i.e. the webserver would send a request to the service and it would answer if the original request should be allowed or not.
> * Service running as proxyed server and using something like `X-Accel-Redirect` headers.

I have no opinion on this. I suppose the last two are close enough such that converting one to the other should be very easy. Therefore, we could simply try both and check which one performs better. I'd be also interested to see if an nginx plugin would be faster, and I even think it should be quite doable, but I feel like this would make setup for admins quite a bit more annoying. Not sure though.

> ## JWKS
> Maybe we should implement a / use an existing JWKS service. Tobira is doing this internally already (I learned), but we need a central one that services public keys for all applications that will create JWTs (LMS, Tobira, etc.).

I'm not sure I understand why we need a service for this? Every JWT source (Tobira, LMS, ...) would have its own JWKS publicly available. Opencast knows all these URLs and fetches them whenever necessary. Opencast can cache them internally however it sees fit. 

---

👍 to everything else in your first comment

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Correct

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Mh I swear I did find `username` being used as part of some existing standard using JWT. Oh well.

I don't care. I feel `username` is a better description of the field, but `sub` is shorter and if you say that it's well understood to be the username most of the time, I'm totally fine with that.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Ok lets use this thread to continue the discussion from DACH: people felt uncomfortable omitting the prefix. Lars even seemed to say that we can't? Can someone explain? 

Because I approached this from the other side multiple sides: I asked whether I could use roles in ACLs that didn't start with `ROLE_` (e.g. for the magic roles in [this PR](https://github.com/opencast/opencast/pull/5056)). And there, I got the answer "hell no, please always use the `ROLE_` prefix!". So what is it now? Either we are prohibiting us from using anything not starting with `ROLE_`, but then we should be able to rely on this in scenarios like this JWT schema. Or we cannot rely on it, but then we should be able to use roles without that prefix in scenarios where it makes sense. Restricting ourselves in both situations makes no sense to me. 
Also note that this is just about JWTs, so worst case is that you cannot grant roles not starting with `ROLE_` via JWT. It doesn't affect any other part of Opencast.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Yeah, that was my thinking: I don't know any actions that use these two characters. We can always define some escape-mechanism later, if people really need to.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: But as long as these rules are simple and well defined? 
I mean, I don't care too much, I won't fight for this. I think this is nice to reduce JWT size and simply makes sense (in fact, it is an implicit assumption in much of Opencast! So maybe making it explicit *is* good.)

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Exactly, that's why the ID is the last part of the string. Easy to just use `entry.split(":", 3)` in code, i.e. only consider the first 2 `:` as separators.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
LukasKalbertodt: Sorry, I don't understand this comment. I don't think we ever need to determine series/playlist-membership of an event?

docs/guides/admin/docs/configuration/security.jwt.md
mtneug: What exactly is the argument for removing `ROLE_`? I only see shorter tokens, which I hope is not really a problem.

What about reducing "magic" steps?

Regarding the prefix: [You can configure the prefix](https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.userdirectory.UserIdRoleProvider.cfg#L6-L8) used for user roles, i.e. `ROLE_USER_`. Not sure if people do that and also not sure if they then don't use `ROLE_`, but you could. I don't think you can configure other role prefixes (for groups, normal roles, etc.), but you can set roles without `ROLE_` for events. Roles typically set by LTI have the form `{number}_(LEARNER|INSTRUCUTOR)` with no prefix (`number` is then usually the LMS course ID).

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
mtneug: ok maybe I don't understand the schema then.

What will happen in this scenario:

```
Event:
  id:       ad84d9eb-3a53-4428-bf92-d60f99438dfa
  isPartOf: 14702b3b-4636-4378-82c7-5f793afa4623

Series:
  id:       14702b3b-4636-4378-82c7-5f793afa4623

JWT:
{
    "exp": 1499863217,
    "oc": ["read:s:14702b3b-4636-4378-82c7-5f793afa4623"]
}
```

A request is then made using this JWT to fetch a static element, e.g. https://cdn.example.com/static/mh_default_org/engage-player/ad84d9eb-3a53-4428-bf92-d60f99438dfa/c7d137c4-ba00-48ac-b15d-df1ef925374b/video.mp4.

Is this not supported? What should happen?

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -1,18 +1,107 @@
...
+
+- *User Info*: information about the user that's possibly used by parts of Opencast. Some JWT use cases don't require user information at all (e.g. static file auth), while others do (e.g. Studio). Therefore, these claims are optional. (TODO: specify when this is stored as user reference)
+    - `username`: string. (TODO: explain how this is used to lookup existing users, if at all)
+    - `name`: display name, string.
+    - `email`: string.
+
+- *Authorization*: at least one of these claims has to be set, as otherwise no privileges are granted (which is the same as not sending a JWT at all).
+    - `roles`: array of strings, specifying roles to grant to the request/user directly. The `ROLE_` prefix must be omitted as Opencast adds it itself.
+    - `oc`: array of strings, grants access to single items in Opencast. The custom syntax is terse in order to keep JWTs as small as possible. Each string consists of three parts, which are separated by a colon (`:`):
```
mtneug: I would argue that the Opencast organization should be part of the JWT. It should either be another claim (should be enough I think) or another part of the `oc` string.

mtneug:
> I'm not sure I understand why we need a service for this? Every JWT source (Tobira, LMS, ...) would have its own JWKS publicly available. Opencast knows all these URLs and fetches them whenever necessary. Opencast can cache them internally however it sees fit.

We currently can only configure one JWKS URL in a JWT login handler. I don't care if we want to add the ability to add a list, but I personally would still configure Opencast to only fetch from one source which lists all keys.

docs/guides/admin/docs/configuration/security.jwt.md
reiferschris: We also have roles in ACLs without `ROLE_` Prefix like:
- `LDAP_XYZ_ADMIN` which are derived from a ldap attribute, which can be configured here: https://github.com/opencast/opencast/blob/develop/etc/org.opencastproject.userdirectory.ldap.cfg.template#L85-L89
- `XYZ_PUBLIC_VIEW` which we also use to determine if a video is public

I agree with @mtneug on this. "Magic" could lead to unexpected issues here.

LukasKalbertodt: Alright, i was not aware of people using roles without that prefix. It seems like there is plenty of precedence. As to my previous point: then we should have not restricted us previously to only use roles starting with `ROLE_`. But oh well, that happened now. 

Will remove the "prefix is added" part of the schema definition soon.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
LukasKalbertodt: Ohh I see. You are right, my text should explain this. What I intended this to do is only giving read/write access to the item itself, i.e. the series metadata or the playlist metadata+entries. NOT to the events in the series/playlist. With this, giving read access to a series is indeed fairly useless, as series metadata is always public in Opencast. But giving write access to a series makes sense to be able to edit its metadata or add events to the series. Write access to a playlist makes sense for the same reason. Read access to a playlist also makes sense, as playlists can actually be protected, i.e. you have to have certain roles to even see its metadata and entries.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Can you elaborate? Do you want the string "opencast" somewhere in the JWT?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
mtneug: > With this, giving read access to a series is indeed fairly useless, as series metadata is always public in Opencast.

Not sure if this is true for the external API.

LukasKalbertodt: Yeah... I always wanted to get to the bottom of this, but suffice to say: how the read ACLs of series are used is... unclear and different code treat them differently.
Anyway, useless or not, this particular thing (`read:s:*`) just falls out of the schema and doesn't hurt. So I guess this discussion thread can be resolved?

docs/guides/admin/docs/configuration/security.jwt.md
oellers: Wouldn't it be easier to deal with a claim structured in the following way later on?

```
"entitlements" : {"uuid-1":["read","write"],"uuid-2":["read","write", "annotate"]}
"roles" : ["studio","editor"]
"groups" : ["admins"]
```
Do we need to differentiate between series, playlists and episodes? Or is the uuid sufficiant?

If we stick to the custom claims, we might want to register it at the [IANA JSON Web Token Claims Registry](https://www.iana.org/assignments/jwt/jwt.xhtml#claims) or prefix it with a namespace?

[entitlements, roles and groups claims -> see RFC 7643](https://www.rfc-editor.org/rfc/rfc7643.html)

docs/guides/admin/docs/configuration/security.jwt.md
oellers: I think @mtneug meant multi tenancy like "mh_default_org" or others

oellers:
> > I'm not sure I understand why we need a service for this? Every JWT source (Tobira, LMS, ...) would have its own JWKS publicly available. Opencast knows all these URLs and fetches them whenever necessary. Opencast can cache them internally however it sees fit.
> 
> We currently can only configure one JWKS URL in a JWT login handler. I don't care if we want to add the ability to add a list, but I personally would still configure Opencast to only fetch from one source which lists all keys.

Such a federated aggregation of metadata would require frequent updates in case of key rollovers and revocations;
A list of JWKS URLs could be a simpler solution, an aggregation service would still be possible then.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: <sup>Sorry for late reply, I was sick</sup>

`entitlements` is very interesting and I didn't know about that claim. It does indeed sound like pretty much what I described with my `oc` claim. Of course, `oc` is shorter than `entitlements`, I'm not sure whether it's worth ignoring standards for some JWT size improvements (aside: I hate that this is even something we have to think about, a factor to weight against, exchange formats shouldn't be designed to be human readable, ramble ramble ramble...).

The structure of `entitlements` you suggest is also interesting. Again, slightly larger than my suggestion but probably insignificantly so. 

> Do we need to differentiate between series, playlists and episodes? Or is the uuid sufficiant?

I would argue the "type" is important. Sure, UUIDs rarely clash, but remember that IDs can be given manually and you don't want accidentally also give access to some totally unrelated item. Your "JSON object" structure could just be extended by using `e:uuid1` as key names.

Finally: `groups`. Not sure if we want to add this as well. Yes Opencast has groups, but these don't have a particularly nice identifier. You can always give the `ROLE_GROUP_` role instead to have the same effect. And we can always add this feature later. So I'd suggest ignoring it for now, to be able to merge this faster.

---

So, for me, `entitlements` and `oc` are both fine. Same for the structure: my array or your object (with the type of item added to the key) are both fine. Your suggestions are probably slightly nicer in both cases. However, they make the JWT larger and I don't have any context for RFC 7643, so not sure how useful that is.

I will bring this up in the technical meeting tomorrow.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Ah yeah, I see. But why would you think it should be part of the JWT? Would it influence auth in any way? Event IDs are still unique, right? I.e. there cannot be the same event ID in two different organizations, right? And same with roles: a role is a role and its meaning does not depend on organization context, right?

docs/guides/admin/docs/configuration/security.jwt.md
mtneug: If we want to follow RFC 7643, the suggested structure is as follows:

```json
"entitlements": [
  {
    "value": "entitlement-1",
    "display": "Entitlement 1 gives some rights",
    "type": "entitlement-1-function",
    "primary": false
  }
]
```

`value`, `display`, `type`, and `primary` are not required. I'm not sure if this is really what we want.

oellers: > If we want to follow RFC 7643, the suggested structure is as follows:

This seems like an implementation suggestion. As all properties are listed as _required: false_ within the scheme, you are basically allowed to extend the scheme of entitlements. The definition for entitlements itself says:
 - "No vocabulary or syntax is specified; [...] service providers and clients are expected to encode sufficient information in the value [...] This value has no canonical types, although a type may be useful as a means to scope entitlements." ([Source](https://datatracker.ietf.org/doc/html/rfc7643))

I think a type could be a custom "OC" Resource Type Schema, comparable to e.g. the [group resource schema.](https://www.rfc-editor.org/rfc/rfc7643#section-4.2), defining kind of an oc class as the rfc states that it provides a  schema and extension model for representing users and groups and **_other resource types_** in JSON format, also gives hints to "defining new resource types" (chapter 3.2, chapter 6 and chapter 8.7).

So with the scheme proposed here originally:
```
"oc": [
        "read+customaction:e:d622b861-4264-4947-8db1-c754c5956433",
        "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"
    ]
```
following the rfc implementation suggestion:
```
  "entitlements": [
        {"value": "read+customaction:e:d622b861-4264-4947-8db1-c754c5956433"},
        {"value": "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"}
    ]
```
That might not be too far off, although I'd recommend that the claim should be as easily processable as possible, e.g. I'd think of json properties instead of needing to split strings by colons, plus or similar, so with an "OC" resource type schema, e.g.

```
"entitlements" : [
        {"uuid-1":["read","write"]},
        {"uuid-2":["read","write", "annotate"]}
]
```
or some more complex, e.g.
```
  "entitlements": [
        {"value": "d622b861-4264-4947-8db1-c754c5956433","acl": ["read", "customaction"],"type":["e"]},
       {"value": "4ed02421-144c-42a1-b98a-22e84f3ac691","acl": ["write"],"type":["s"]}
 ]
```
Json objects instead of strings should also allow for easier future modifications, since properties could be added more easily. Also operations (e.g. granting access / searching) on these objects should be faster as there are standardized ways for dealing with json objects. 

With json properties we also do not need to rely on the exact index positions that could change in the future and break implementations, e.g. if we split by colon:
read+customaction:e:d622b861-4264-4947-8db1-c754c5956433
Index [0] = ACL
Index [1] = Type
Index [2] = UUID

Presumably we will be looking up grants mainly by UUIDs later? Or do we construct the required ACL and check if it is e.g. present in an array? Wouldn't it then make sense to split up those ACLs instead of combining with e.g. "plugs"? That would probably bloat the jwt with redundant information:
```
"oc": [
        "read:e:d622b861-4264-4947-8db1-c754c5956433",
        "customaction:e:d622b861-4264-4947-8db1-c754c5956433",
        "write:s:4ed02421-144c-42a1-b98a-22e84f3ac691"
    ]
```
But maybe these thoughts are too far ahead.

> I would argue the "type" is important. Sure, UUIDs rarely clash, but remember that IDs can be given manually and you don't want accidentally also give access to some totally unrelated item. Your "JSON object" structure could just be extended by using e:uuid1 as key names.

Is this a real use case? Aren't UUIDs of an OC instance always unique? Or is the uniqueness only assured by types? There is probably also a uniqueness check for manual IDs?

LukasKalbertodt: > Is this a real use case? Aren't UUIDs of an OC instance always unique? Or is the uniqueness only assured by types? There is probably also a uniqueness check for manual IDs?

No they are not. I justed created a series and an event in Opencast, both with the id `foo`. So yes, uniqueness is only ensured if we add the type to the ID.

---

Regarding the data structure discussion: I would absolutely not go for a verbose structure like quote by Matthias. Even if most fields are optional, just having the same field name in every JWT is just such a waste. I like the argument of using standard JSON to avoid custom parsing of strings. Again, I'm in favor of either my initial suggestion or, slightly preferred, the first one by @oellers (the json object with uuids+type as keys). 

> Wouldn't it then make sense to split up those ACLs instead of combining with e.g. "plugs"?

The reason I suggested this is that granting multiple "actions" to the same item is common and we want to avoid repeating the ID multiple times.

KatrinIhler:
@LukasKalbertodt Shouldn't this be a draft still? :thinking:

LukasKalbertodt:
@KatrinIhler that depends on how Opencast uses "draft PRs". There are open discussions, but it's not like this PR is "not yet done". Are there any guidelines when PRs should be drafts in this repository? (Maybe lets move this discussion elsewhere tho)

docs/guides/admin/docs/configuration/security.jwt.md
mtneug: Summary of the dev meeting discussion regarding this issue: do not include it. It's easier to include it later that to remove it if we don't need it.

KatrinIhler:
Draft is basically when the author wants to prevent people from merging bc the PR is not yet ready. I thought this might be the case here but now I think I might've been wrong.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: We discussed this in the technical meeting just now and decided:

- To use `oc` instead of `entitlements`: using the latter is only useful if we plan on interfacing with any systems using the linked RFC. This is very unlikely and we cannot think of any reasonable situation where we might. Further, the RFC suggests (yes, only suggests, but still) a schema for that claim, which we wouldn't use anyway. So only half following the RFC seems not useful.
- To use the `{ "e:uuid": ["read"] }` syntax for the `oc` claim instead of the array one I proposed. Using proper JSON syntax instead of custom string syntax avoids writing custom parsers (even if they are super simple), side steps the problem of having actions (e.g. `read`) containing colons or plus signs (which would require escaping), and also makes this a bit more flexible as we could add other fields to the object in the future.

I will update this PR soon.

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: I removed this in the latest commit. For static file auth, only `read` is required/used so for that use case, these extra rules don't improve anything. And for other use cases, JWT size is not as much of an issue.

gregorydlogan:
Considering the goal here, I'm going to merge this.  While this implementation likely isn't complete or perfect, it's the basis of future (funded) work and not having this done is/will be blocking that work.

PR #6233, opened 2024-10-08
Handle content id being null for playlist entries

### LINKED ISSUE #5744: Setting `contentId` to `null` of a playlist entry returns a 500 error code
**Describe the bug**

The playlist API route `POST /{id}/entries` allows to set `null` values for `contentID`. It seems that there are no checks, which breaks the GET routes of the playlists api.

**Stack trace**
```
java.lang.IllegalArgumentException: Value of field 'contentId' must not be null
 	at com.entwinemedia.fn.data.json.Field.<init>(Field.java:31) ~[?:?]
 	at com.entwinemedia.fn.data.json.Jsons.f(Jsons.java:89) ~[?:?]
	at org.opencastproject.external.endpoint.PlaylistsEndpoint.playlistEntryToJson(PlaylistsEndpoint.java:401) ~[?:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
	at java.util.Vector$VectorSpliterator.forEachRemaining(Vector.java:1470) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682) ~[?:?]
	at org.opencastproject.external.endpoint.PlaylistsEndpoint.playlistToJson(PlaylistsEndpoint.java:384) ~[?:?]
	at org.opencastproject.external.endpoint.PlaylistsEndpoint.lambda$getPlaylistsAsJson$0(PlaylistsEndpoint.java:227) ~[?:?]
        ...
```

**Server environment:**
 - https://develop.opencast.org

KatrinIhler:
Currently it's possible to create playlist entries with empty or null content ids. This causes problems when trying to get the playlist both from the External API (during the transformation to json) and the endpoint of the service itself (though only when `withPublications` is set).

In my opinion the real solution would be not allowing empty content ids in the first place because I don't see how that could be useful, but I didn't manage that in the limited time I had. For now this should at least fix #5744.

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/PlaylistsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/PlaylistsEndpoint.java
@@ -21,6 +21,7 @@
 package org.opencastproject.external.endpoint;
 
 import static com.entwinemedia.fn.data.json.Jsons.BLANK;
+import static com.entwinemedia.fn.data.json.Jsons.NULL;
 import static com.entwinemedia.fn.data.json.Jsons.arr;
 import static com.entwinemedia.fn.data.json.Jsons.f;
 import static com.entwinemedia.fn.data.json.Jsons.obj;
@@ -350,7 +351,7 @@ public class PlaylistsEndpoint {
     List<Field> fields = new ArrayList<>();
 
     fields.add(f("id", v(playlistEntry.getId())));
-    fields.add(f("contentId", v(playlistEntry.getContentId())));
+    fields.add(f("contentId", v(playlistEntry.getContentId(), NULL)));
     fields.add(f("type", enumToJSON(playlistEntry.getType())));
     return obj(fields);
   }
--- a/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
+++ b/modules/playlists/src/main/java/org/opencastproject/playlists/PlaylistService.java
@@ -417,12 +417,20 @@ public class PlaylistService {
     // Add additional infos about events
     List<JaxbPlaylistEntry> jaxbPlaylistEntries = jaxbPlaylist.getEntries();
     for (JaxbPlaylistEntry entry : jaxbPlaylistEntries) {
+      String contentId = entry.getContentId();
+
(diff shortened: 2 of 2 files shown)


lkiesow:
I think playlists were added in 16.0. Does it make sense to target `r/16.x` with this patch?

KatrinIhler:
I actually wrote this for 16, just forgot to change the target. Thank you for reminding me :D

Arnei (approved):
Did not run this, but the changes read reasonable to me.

geichelberger:
Does it make sense that a playlist entry has no content? Maybe the ID is invalid, but it should have pointed to some content. Should we enforce the rule that the field cannot be empty?

Arnei:
Talked to Lukas about this and we agree that there is no good reason for the content id to be `null`. So we could "just" enforce that in the database via JPA.

gregorydlogan:
Enforcing it via JPA makes a lot of sense to me.  Do we want to merge this and then do a follow on patch, or should someone push the change to this branch and then we can take a look?

lkiesow:
@dennis531, @tgloeggl, I vaguely remember you discussing this. Does this make sense for you as well. After all the Stud.IP plugin is using this already.

dennis531:
> @dennis531, @tgloeggl, I vaguely remember you discussing this. Does this make sense for you as well. After all the Stud.IP plugin is using this already.

I agree that we should not allow null entries, as I currently see no use for them. The plugin should not set null content ids. The problem was caused by missing consideration of edge cases resulting in currupt data.

gregorydlogan:
While I think making the column itself enforce not-nullability, that doesn't fix existing tables or their data.  That's only going to happen in a major release I suspect, so I'm going to merge this now since it fixes the current issue.

PR #6237, opened 2024-10-09
Allow rolling back to a specified snapshot in asset manager

lkiesow:
This patch improves the `asset-delete` workflow operation handlöer to not only allow for deleting all or all except the latest snapshot of an event, but also allows deleting all version newer than a given version. This essentially means that you can roll back to a given event in the asset manager.

Along the way, this also fixes the problem of this operation creating errors since it may delete files the current workflow requires to work on. This could previously be worked around using `ingest-download`, but is actually quite unnecessary.

*This keeps the current behavior (e.g. if you were already working around the problem using `ingest-download`) unchanged*

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/asset-delete-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/asset-delete-woh.md
@@ -3,34 +3,51 @@ Asset Manager Delete Workflow Operation
 
 ID: `asset-delete`
 
+<div class=warn>
+This operation will permanently delete internal Opencast data.
+This can result in data loss.
+Make sure to carefully read the documentation and understand the side-effects.
+This especially means the publication storage mentioned below.
+</div>
+
 Description
 -----------
 
-The delete handler is responsible for deleting an episode, identified by the workflow’s current media package, from the
-asset manager.
+The delete handler can be used for deleting snapshotted versions of an episode, identified by the workflow’s current
+media package, from the asset manager.
 
-If no parameter is given, the whole episode and all of its snapshots are deleted.
-If the keep-last-snapshot parameter is used, it is advised to use the *ingest-download* workflow before *asset-delete*.
-Otherwise there will be logged a lot of errors for unreferenced snapshots and ACLs may vanish.
-
+The media package resulting from this operation is the one from the latest snapshot after the deletion, but with the
+publication section from the current workflow from when the operation started.
 
 Parameter Table
 ---------------
 
-|Configuration Key         |Example           |Description                                       |
-|--------------------------|------------------|--------------------------------------------------|
-|keep-last-snapshot        |true              |Deletes every snapshot except the last one.        |
+|Configuration Key  |Example |Description                                                           |
+|-------------------|--------|----------------------------------------------------------------------|
+|keep-last-snapshot |true    |Deletes every snapshot except the last one.                           |
+|roll-back-to       |5       |Rolls back to the specified snapshot version, deleting all newer ones.|
+
+
+Publications
+------------
+
+Opencast stores the current version of a publication in the latest snapshot.
+This means that rolling back and thus deleting newer snapshots may result in the loss of the current publication data.
+If that is what you need or if the publications were not modified in the snapshots you deleted, that's fine.
+
+In case you do need to keep the current list of publications, however, the operation will return the media package of
+the version you rolled back to, but with the publication section from when the operation started.
+This allows you to easily create a new snapshot with the media from the version you rolled back to,
+but the current list of publications.
+You just need to add a `snapshot` operation after this operation.
 
 
 Operation Example
 -----------------
 
-```xml
-<operation
-    id="asset-delete"
-    description="Delete from AssetManager">
-  <configurations>
-    <configuration key="keep-last-snapshot">true</configuration>
-  </configurations>
-</operation>
+```yaml
+- id: asset-delete
+  description: Delete all but latest snapshot
+  configurations:
+    - keep-last-snapshot: true
 ```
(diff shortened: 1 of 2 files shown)


docs/guides/admin/docs/workflowoperationhandlers/asset-delete-woh.md
```diff
@@ -6,31 +6,27 @@ ID: `asset-delete`
...
 
-The delete handler is responsible for deleting an episode, identified by the workflow’s current media package, from the
-asset manager.
+The delete handler can be used for deleting snapshotted versions of an episode, identified by the workflow’s current
+media package, from the asset manager.
 
-If no parameter is given, the whole episode and all of its snapshots are deleted.
-If the keep-last-snapshot parameter is used, it is advised to use the *ingest-download* workflow before *asset-delete*.
-Otherwise there will be logged a lot of errors for unreferenced snapshots and ACLs may vanish.
+If no parameter is given, the all snapshots are deleted.
```
KatrinIhler:
```suggestion
If no parameter is given, all snapshots are deleted.
```

KatrinIhler (commented):
Code looks reasonable. I didn't try it out yet. There could be one big problem though: If we remove the latest x snapshots, we need to update that data in Elasticsearch, and from my quick glance at the code, that doesn't happen when we delete snapshots, only when we _take_ snapshots or delete a whole episode. So far that wasn't a big deal because we either kept the latest snapshot, meaning ES was up to date, or we threw the whole event away anyway.

What would need to happen is we'd need treat the latest remaining snapshot like a new one for an ES update.

lkiesow:
> So far that wasn't a big deal because we either kept the latest snapshot […]or we threw the whole event away anyway.

Actually, just deleting all snapshots but not the event was the default behavior of this operation before and isn't new. So you already had the problem in the old code. But I see your point.

KatrinIhler:
In the Dev meeting we discussed that simply going back to an old snapshot means that the publications referenced by the AssetManager and the actually existing publication data will potentially no longer be consistent, which can be a problem e.g. for retraction later or use of the External API. We couldn't reach an agreement what exactly we should do in this case, but we should be consistent in some way, e.g. keeping the publication references when rolling back, retracting the publications by default... but we can't go back to an older publication state since that data isn't versioned.

lkiesow:
> e.g. keeping the publication references when rolling back

That's not impossible since the publication information are versioned, and you cannot roll back to version `n` and modify it. You would need to roll back, modify the publication and create a new version `n'` in the asset manager which would contain the modifications.

To address the problems, I made a few changes which we can hopefully agree on:

- Add a warning to the documentation, that this is a dangerous operation
- Target `develop` due to the additional changes
- Don't allow removing *all* versions. We agreed that this never really made sense
- Have the operation return the media package from the snapshot you rolled back to, but with the current publications transferred over. This has two advantages over how things worked before:
  1. No more need for keeping old files in the workspace. The remainder of the workflow will continue to run on valid data.
  2. If you want to keep your current publications, you can easily do that with a `snapshot` operation which will then create a new version with the updated publication data.
- I've added the publication problems to the docummentation

KatrinIhler (changes requested):
Just looked at this again and I'm happy with most of the changes but still have a few complaints:

- Nitpick: Document that one of the two parameters (and only one) needs to be set for the operation, _or_ change the default behavior to keepLatestSnapshot and have only one parameter to set a different snapshot to keep. The latter is nicer imo, but also more work.
- The index still isn't being updated as far as I can tell.
- And now to the big one :grimacing: 

> This means that rolling back and thus deleting newer snapshots may result in the loss of the current publication data.
If that is what you need or if the publications were not modified in the snapshots you deleted, that's fine.

I disagree with this.

The 'not modified'  case is fine, but can be difficult/annoying for a user to ascertain (e.g. if a mp was republished in between with slight modifications e.g. a new attachment or different tags). 

I disagree that the user could ever want to go back to an earlier publication state like this though because we're telling neither the file system nor other services like search that this happened, meaning that currently distributed files might no longer be referenced/the old URLs potentially point at distributed files that no longer exist, and search could contain different information about the engage publication than the AssetManager. Why would we ever want that?

The nicest solution imo is to, as you said,

> roll back, modify the publication and create a new version n in the asset manager which would contain the modifications

either always (which also solves the index update problem, but might create a snapshot unnecessarily) or only if the publication state differs in any way (is annoying to test for and leaves the index problem to be solved in a different way)

Or, if we want to get really nasty, we could enable the AssetManager to change the publications of an existing version _without_ a snapshot in a sort of merge, basically creating an updated version of the same snapshot. Of course you could argue that this goes against everything snapshots stand for, but I would reason that the publication information isn't as relevant to what a snapshot is than its assets (and in fact I could argue that it shouldn't be part of the snapshot in the first place, but that's what we've got right now.) It's something that one would have to consider more carefully though.

TLDR: At the very least i want a big fat warning à la

> It is strongly advised to do a snapshot after this operation if there is any chance that the publication information between the two snapshots is different in any way, otherwise you could end up with inconsistent data. (Retracting/republishing might not be enough because that doesn't usually affect the internal publication.)

in the documentation.

lkiesow:
Fair enough. Since we disagree. Let's just close this. Let's leave it in the broken state it is right now :shrug: 

There is a big warning block already.
I strongly disagree with modifying old snapshots. I also wouldn't automatically create new snapshots. Again, that's not the idea of a rollback.

Sure, the publications in the snapshots are not a great idea in the first place. They shouldn't be stored in a versioned way since they are not versioned. But we shouldn't do weird stuff just because of that.

KatrinIhler:
>  Let's leave it in the broken state it is right now. 🤷

This feels a bit passive aggressive to me tbh, and I don't really understand why. I agree that the state this WOH is in is not great right now. But I don't think that means we should not be critical of a potential fix (and this PR is a bit more than that), and I don't think you do either.

It's unfortunate that we can't come to an agreement here, not even in regards to a rewording of the warning. While we can (and should) argue about the solutions (and I do see their drawbacks), I stand by my opinion that the current documentation downplays the potential data inconsistencies after this operation.

I'm open to discuss this some more, but if that's not something you have time for right now, I guess we can leave it closed and maybe somebody else can get back to this later.

PR #6242, opened 2024-10-15
Backport "Attach Amberscript transcriptions as tracks"

### LINKED ISSUE #6179: Backport of changes for amberscript from Opencast Version 16 to Opencast Version 15 possible?
Is it possible to make the [changes](https://github.com/opencast/opencast/pull/6142/files) made for amberscript in version 16.5 (see also [#6143](https://github.com/opencast/opencast/pull/6143) & [#6141](https://github.com/opencast/opencast/pull/6141)) also available for version 15?

Arnei:
Fixes #6179

Combines abfecd9b4cc8efab496ca961e7f195b6fb78f747, e4653e927e2d275401aa94cad2e67e4e51199c94 and
929bbc4273eacadb7a3231c129c1dacf8b2e90a4 into a single commit and points the changes at r/15.x instead of r/16.x.

Due to lack of access to Amberscript I did not test this., but at least there is a PR now which people with Amberscript can test.

#### Why is this aimed at legacy?

This intends to fix a bug that breaks Amberscript transcription.


mtneug:
I'm not per se against this PR, but in light of the recently passed proposal, changes to legacy branches have to be explained in order to prevent breaking changes (see https://groups.google.com/a/opencast.org/g/dev/c/VwYqk3vXdd0 and https://github.com/opencast/opencast/pull/6239).

mwuttke:
We will use version 15.x in our institution for this semester and will not switch to version 16 during operation.
Our users use both amberscript or wishper for subtitling their videos.
And in our current 15 productive instance - especially for the use of wishper - we have switched from attachments to tracks for subtitles and thus want to ensure that the use of amberscript is also possible.

@mtneug: Is that enough justification for you?

mtneug:
> @mtneug: Is that enough justification for you?

Not only my decision! Each PR is discussed shortly in the weekly Opencast dev meeting every Tuesday. I don't think we really reached a decision for this PR today. So further discussion will take place here.

Note that this change in our approach for legacy branches is new and we don't have a history of what "good" reasons for changes to legacy branches are. However, I tend to classify "Institution A wants feature X in legacy in order to use it now" as not a valid reason. Ultimately, we want to minimize changes such that adopters can trust releases of the legacy version to install without any problems.

mwuttke:
Regardless of what we as an institution imagine to be a desirable feature for a still supported version, we would have expected that the switch from attachments to tracks for subtitles with the Opencast release of version [15.0 in December 2023](https://docs.opencast.org/r/15.x/admin/#releasenotes/#opencast-150) (sic!) would also guarantee the functionality of the translation services supported by Opencast, such as amberscript. Hence our wish to approve this PR. Or is this expectation exaggerated?

lkiesow:
**tl;dr**: I think it's fine for this to go into `r/15.x` if, and only if, this gets tested properly.

---

Since I caused this change in policy as a consequence of bringing the results of the “Make updates easier/more stable” to a vote on list, let me chime in here.

I fully agree with Matthias that “I need it now while we don't want to update” isn't really a good argument in general. Also the “fix a bug that breaks Amberscript transcription” is just a half-truth since it still works exactly as it did before and Opencast can still use subtitles as attachments. The idea behind the policy change is to force someone who wants to push things into legacy to really think about if this is necessary and what possible complications are. We have seen problems in the recent past because we carelessly added/back-ported patches to legacy without much testing. If you think about it, the bug has now existed in a release for over 10 months without anyone complaining and we are about 1.5 months before EOL of Opencast 15.

That being said, if I would want to argue for this, I think that Opencast encourages using tracks, that the operation's documentation advertises it will do that, and then it silently doing something else would be a good argument. Combine this with the observation that this is a low risk patch which would likely just break the `amberscript-attach-transcription` operation if this doesn't work properly, I think it is fine for this to go into legacy.

But then looking at f3105b1, I see a good argument to not merge this as is:

> Due to lack of access to Amberscript I did not test this.

We shouldn't push things into legacy hoping it's fine and occasionally breaking a release. Legacy should be really stable. That means that patches need to be properly tested before we merge them. This should be the case with `develop`, but becomes much more important the older the release branches become. Michael, since you are interested, maybe you can help with testing this.

mwuttke:
I understand your arguments and will bow to the policy, especially as I don't have time to test the patch at the moment.

Arnei:
100% agree that this should not be merged without proper testing, which is exactly why I emphasized that I did not do that.

> Also the “fix a bug that breaks Amberscript transcription” is just a half-truth since it still works exactly as it did before and Opencast can still use subtitles as attachments.

Sure, it still works like before, but the context around it changed. Opencast not only encourages subtitles as tracks since 15.0, but some parts of Opencast will straight up only work with subtitles if they are tracks, i.e. [the editor](https://github.com/opencast/opencast-editor/). Therefore, I would consider this a bug that should have been solved as part of #4407.

Arnei:
The recent release of Opencast 15.12 is intended to be the final minor release for 15. Please retarget your PR or explain why it warrants another Opencast 15 release.

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
```diff
@@ -38,155 +38,81 @@ dropdowns below.
...
+    - source-tag: transcript
+    - language: de
+    - jobtype: direct
+    - skip-if-flavor-exists: captions/vtt
 
 </details>
 
 <details>
 
-<summary>amberscript-start-transcription.xml</summary>
```
gregorydlogan, line 128: This appears to be some kind of merge/squash issue, renders like this:

image

gregorydlogan (changes requested):
Docs be busted.

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
lkiesow: Yep, looks like a merge conflict.
It's fine in the 16.x docs: https://docs.opencast.org/r/16.x/admin/#configuration/transcription.configuration/amberscripttranscripts/

lkiesow:
If you are testing this, @gregorydlogan, I posted a complete Amberscript configuration (except for the API key) in pull request #6299 (obviously, I also wouldn't complain if you also test those patches :grin: )

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
```diff
@@ -38,155 +38,81 @@ dropdowns below.
...
+operations:
+- id: encode
+  fail-on-error: false
+  exception-handler-workflow: partial-error
+  description: Encoding audio for transcription
+  configurations:
+    - source-flavor: "*/source"
+    - target-flavor: audio/mp3
+    - target-tags: transcript
+    - encoding-profile: audio-mp3
```
gregorydlogan: `audio-mp3` doesn't exist, and `audio-only.copy` doesn't seem to work either.

lkiesow: Forgot the encoding profile file on my patch. Updated the patch file there as well, but here is the updated config patch:

[amberscript-config-2.patch.txt](https://github.com/user-attachments/files/17822295/amberscript-config-2.patch.txt)

gregorydlogan: Works correctly with the attached patch!

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
Arnei: Should be fixed now.

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
Arnei: Thank you two for the review :)

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
```diff
@@ -38,155 +38,81 @@ dropdowns below.
...
-        <configuration key="delete-external">true</configuration>
-        <!-- FixMe Don't clean up ACLs until workflow service no longer looks for them in the WFR. -->
-        <configuration key="preserve-flavors">security/*</configuration>
-      </configurations>
-    </operation>
-
-  </operations>
-
-</definition>
+<summary>amberscript-start-transcription.yaml</summary>
```
gregorydlogan, line 41: @Arnei this is `amberscript-start-transcription`, as is the one below.  This one was originally `amberscript-attach-transcription` - so our docs are missing the attach part of the workflows

gregorydlogan (changes requested):

docs/guides/admin/docs/configuration/transcription.configuration/amberscripttranscripts.md
Arnei: 🤦. Thanks for catching that.

lkiesow:
This caused problems when merging `r/15.x` into `r/16.x`. Given that this was a backport of something that's already in `r/16.x` my solution was to just use the current version that's in `r/16.x`.

PR #6247, opened 2024-10-20
Allow generating subtitles in the background

lkiesow:
This patch allows generating subtitles in the backgrould, starting the jobs first and attaching the result later in the workflow.

Unfortunately, looking into the code, I also found a lot of minor issues which I fixed as well, but which were hard to separate from this patch:

- `language-fallback` didn't work
- Defaults for generator enging and Whisper.cpp binary fixed
- Defaults for several operations
- Simplified code

### How to test this patch

- Make sure to install Whisper.cpp. You may need to update the configuration `etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg`. Likely you just have to configure something like:
  - `whispercpp.root.path=/home/lars/dev/whisper.cpp/main`
  - `whispercpp.model=/home/lars/dev/whisper.cpp/models/ggml-tiny.bin`
- Update the `fast.yaml` workflow:
```diff
diff --git a/etc/workflows/fast.yaml b/etc/workflows/fast.yaml
index 7a0494ed46..f526ed0a60 100644
--- a/etc/workflows/fast.yaml
+++ b/etc/workflows/fast.yaml
@@ -25,6 +25,7 @@ configuration_panel_json: |-
     ]
   }]
 operations:
+
   - id: defaults
     description: "Applying default configuration values"
     configurations:
@@ -45,6 +46,14 @@ operations:
       - overwrite: false
       - accept-no-media: false
 
+  - id: speechtotext
+    description: Generates subtitles for video and audio files
+    configurations:
+      - source-flavor: '*/prepared'
+      - target-flavor: captions/source
+      - limit-to-one: true
+      - async: true
+
   - id: encode
     fail-on-error: true
     exception-handler-workflow: "partial-error"
@@ -61,9 +70,7 @@ operations:
     description: "Tag captions for publication"
     configurations:
       - source-flavor: "captions/source"
-      - target-flavor: "captions/preview"
       - target-tags: "+engage-download"
-      - copy: true
 
   - id: image
     if: "${straightToPublishing}"
@@ -110,6 +117,12 @@ operations:
       - target-tags: "engage-download"
       - encoding-profile: "player-slides.http"
 
+  - id: speechtotext-attach
+    description: Generates subtitles for video and audio files
+    configurations:
+      - target-flavor: captions/source
+      - target-tags: engage-download
+
   - id: publish-configure
     exception-handler-workflow: "partial-error"
     description: "Publish to preview publication channel"
```

- Download a video with subtitles like:
  https://github.com/user-attachments/assets/932ef91b-580a-4d11-9717-011b7bbdb142
- Convert this to a wav file:
    ```
    ffmpeg -i github-multiline-suggestion.mp4 github-multiline-suggestion.wav
    ```
- Ingest both files:
    ```
    curl -i -u admin:opencast http://localhost:8080/ingest/addMediaPackage/fast -F 'flavor=presentation/source' -F BODY=@github-multiline-suggestion.mp4 -F flavor=presentation/prepared -F BODY2=@github-multiline-suggestion.wav -F title="I 🖤 Opencast"
    ```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-attach-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/speechtotext-attach-woh.md
@@ -1,0 +1,41 @@
+Speech to Text Attach Workflow Operation
+========================================
+
+ID: `speechtotext-attach`
+
+Description
+-----------
+
+The speech to text attach operation can be used to attach the generated subtitles from a previously run
+[`speechtotext`](speechtotext-woh.md) operation if the process was started asynchronously.
+
+Parameter Table
+---------------
+
+| Configuration Keys       | Required | Example          | Description
+|--------------------------|----------|------------------|-------------
+| target-flavor            | yes      | archive          | Flavor of the produced subtitle
+| target-tags              | no       | captions/source  | Tags for the subtitle file².
+| target-element           | no       | track            | Define where to append the subtitles file. Possibilities are: as a 'track' or as an 'attachment' (default: `track`).
+
+
+Operation Examples
+------------------
+
+```yaml
+- id: speechtotext
+  description: Generates subtitles for video and audio files
+  configurations:
+    - source-flavor: '*/source'
+    - target-flavor: captions/source
+    - limit-to-one: true
+    - async: true
+
+- … other operations
+
+- id: speechtotext-attach
+  description: Attach generated subtitles
+  configurations:
+    - target-flavor: captions/source
+    - target-tags: engage-download
+```
--- a/etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
@@ -4,5 +4,5 @@
 
 # Select STT engine.
 # Available engines: vosk, whisper, whispercpp
-# Default: (enginetype=whisper)
-#SpeechToTextEngine.target=(enginetype=whisper)
+# Default: (enginetype=whispercpp)
+#SpeechToTextEngine.target=(enginetype=whispercpp)
--- a/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
@@ -1,6 +1,6 @@
-# Configuration for setting a custom path to the Whisper command line tool
-# Default: whisper
-#whispercpp.root.path=whispercpp
+# Configuration for setting a custom path to the Whisper.cpp command line tool
+# Default: whisper.cpp
+#whispercpp.root.path=whisper.cpp
 
 # Absolute path to language model
 # RPMs install models to /usr/share/ggml/ggml-<model>.bin
--- a/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
+++ b/modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
@@ -103,6 +103,12 @@ public class
   /** Configuration: Limit to One (If true, max 1 subtitle file will be generated) */
   private static final String LIMIT_TO_ONE = "limit-to-one";
 
+  /** Configuration: Synchronous or asynchronous mode */
+  private static final String ASYNCHRONOUS = "async";
+
+  /** Workflow configuration name to store jobs in */
+  private static final String JOBS_WORKFLOW_CONFIGURATION = "speech-to-text-jobs";
+
   private enum TrackSelectionStrategy {
     PRESENTER_OR_NOTHING,
     PRESENTATION_OR_NOTHING,
@@ -158,6 +164,9 @@ public class
     MediaPackage mediaPackage = workflowInstance.getMediaPackage();
     logger.info("Start speech-to-text workflow operation for media package {}", mediaPackage);
 
+    // Defaults to `false` if `null`
+    var async = BooleanUtils.toBoolean(workflowInstance.getCurrentOperation().getConfiguration(ASYNCHRONOUS));
+
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(workflowInstance,
             Configuration.none, Configuration.one,
             Configuration.many, Configuration.one);
@@ -178,10 +187,10 @@ public class
     String languageCode = getMediaPackageLanguage(mediaPackage, workflowInstance);
 
     // How to save the subtitle file? (as attachment, as track...)
-    AppendSubtitleAs appendSubtitleAs = howToAppendTheSubtitles(mediaPackage, workflowInstance);
+    AppendSubtitleAs appendSubtitleAs = howToAppendTheSubtitles(workflowInstance);
 
     // Translate to english
-    Boolean translate = getTranslationMode(mediaPackage, workflowInstance);
+    Boolean translate = getTranslationMode(workflowInstance);
 
     // Create sublist that includes only the tracks that has audio
     List<Track> tracksWithAudio = tracks.stream().filter(Track::hasAudio).collect(Collectors.toList());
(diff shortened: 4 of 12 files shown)


etc/org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.cfg
```diff
@@ -4,5 +4,5 @@
 
 # Select STT engine.
 # Available engines: vosk, whisper, whispercpp
-# Default: (enginetype=whisper)
-#SpeechToTextEngine.target=(enginetype=whisper)
+# Default: (enginetype=whispercpp)
```
lkiesow, line 7: Since we provide whisper.cpp in the Opencast repositories while not providing the other tools, I think it makes sense to make it the default.

etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
```diff
@@ -1,6 +1,6 @@
-# Configuration for setting a custom path to the Whisper command line tool
-# Default: whisper
-#whispercpp.root.path=whispercpp
+# Configuration for setting a custom path to the Whisper.cpp command line tool
+# Default: whisper.cpp
```
lkiesow, line 2: If I install Whisper.cpp from any of the Opencast repositories, the binary is actually named `whisper.cpp`. So, let's make that the actual default.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -192,33 +201,29 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
 
     // Use the selection strategy from the workflow config to get the tracks we want to transcribe
     List<Track> tracksToTranscribe = filterTracksByStrategy(tracksWithAudio, trackSelectionStrategy);
+    if (tracksToTranscribe.isEmpty()) {
```
lkiesow, line 204: If there are no tracks we can skip right here. No need to run through everything else and then still skip at the end.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -192,33 +201,29 @@ public WorkflowOperationResult start(WorkflowInstance workflowInstance, JobConte
...
+      logger.info("No subtitles were created for media package {}. "
+          + "Workflow Configuration 'track-selection-strategy' is set to {}", mediaPackage, trackSelectionStrategy);
+      return createResult(mediaPackage, WorkflowOperationResult.Action.SKIP);
+    }
 
     // Load the 'limit-to-one' configuration from the workflow operation.
     // This configuration sets the limit of generated subtitle files to one
     boolean limitToOne = BooleanUtils.toBoolean(workflowInstance.getCurrentOperation().getConfiguration(LIMIT_TO_ONE));
+    if (limitToOne) {
+      tracksToTranscribe = List.of(tracksToTranscribe.get(0));
```
lkiesow, line 214: This replaces the very complex block below. This actually does the same since `createSubtitle(…)` always returned `true`, meaning it would always stop after the first track had been processed. That means we can just reduce the number of tracks to one right here.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -284,22 +289,15 @@ private List<Track> filterTracksByStrategy(List<Track> tracksWithAudio,
...
    */
-  private boolean createSubtitle(Track track, String languageCode, MediaPackage parentMediaPackage,
+  private void createSubtitle(Track track, String languageCode, MediaPackage parentMediaPackage,
           ConfiguredTagsAndFlavors tagsAndFlavors, AppendSubtitleAs appendSubtitleAs, Boolean translate)
           throws WorkflowOperationException {
 
     // Start the transcription job, create subtitles file
     URI trackURI = track.getURI();
 
-    if (!track.hasAudio()) {
```
lkiesow, line 297: We actually checked that before. It's one of the track filters.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -390,79 +416,51 @@ private TrackSelectionStrategy getTrackSelectionStrategy(MediaPackage mediaPacka
...
-      trackSelectionStrategy = TrackSelectionStrategy.EVERYTHING; // "transcribe everything" is the default/fallback
-    } else {
-      try {
-        trackSelectionStrategy = TrackSelectionStrategy.fromString(strategyCfg);
-      } catch (IllegalArgumentException e) {
-        throw new WorkflowOperationException(String.format(
-            "Speech-to-Text job for media package '%s' failed, because of wrong workflow configuration. "
-                + "track-selection-strategy of type '%s' does not exist.", mediaPackage, strategyCfg));
-      }
+      return TrackSelectionStrategy.EVERYTHING; // "transcribe everything" is the default/fallback
```
lkiesow, line 420: We can just return if we already know the result. No need to continue. This makes the structure simpler.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -390,79 +416,51 @@ private TrackSelectionStrategy getTrackSelectionStrategy(MediaPackage mediaPacka
...
-        translateMode = false;
-      } else {
-        throw new IllegalArgumentException(String.format(
-            "Speech-to-Text job for media package '%s' failed, Because invalid \"translate\" value ('%s')."
-                + "Valid values types: \"true\", \"1\", \"yes\", \"false\", \"0\", or \"no\".", mediaPackage,
-            stringTranslateMode));
-      }
-    }
-    return translateMode;
+    return BooleanUtils.toBoolean(StringUtils.trimToEmpty(operation.getConfiguration(TRANSLATE_MODE)));
```
lkiesow, line 463: This one line basically replaces the whole function :D
We don't have to re-invent parsing booleans.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -489,6 +487,11 @@ private String getMediaPackageLanguage(MediaPackage mediaPackage, WorkflowInstan
       language = StringUtils.trimToEmpty(mediaPackage.getLanguage());
     }
 
+    if (language.isEmpty()) {
+      // If nothing helps, use the fallback language
+      language = Objects.toString(operation.getConfiguration(LANGUAGE_FALLBACK), "en");
```
lkiesow: We were never using language-fallback. Now we do.

KatrinIhler:

lkiesow:
Fixed merge conflict

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
marwyg: Whisper has this language detection feature. The language detection will be triggered at a later point, when the language variable is empty. When we set the language to the fallback here at this point, we have no chance to tell whisper, it has to use the language detection.

I am not sure what to do here, but I think we should either remove this if block or introduce a new workflow operation variable named "use-language-detection" or something. Or maybe allow the speechtotext operation to take "auto" as a valid "language-fallback" value?

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-attach-woh.md
```diff
@@ -0,0 +1,41 @@
...
+The speech to text attach operation can be used to attach the generated subtitles from a previously run
+[`speechtotext`](speechtotext-woh.md) operation if the process was started asynchronously.
+
+Parameter Table
+---------------
+
+| Configuration Keys       | Required | Example          | Description
+|--------------------------|----------|------------------|-------------
+| target-flavor            | yes      | archive          | Flavor of the produced subtitle
+| target-tags              | no       | captions/source  | Tags for the subtitle file².
```
marwyg, line 18: There is a little typo at the end `²`.

As mentioned in my review comment, there could be maybe a note, saying, that the defined tags in the speechtotext operation will be overridden by this target-tags.

marwyg (changes requested):
The code looks way better as before. I couldn't find anything wrong with it.

### I tested the following things (which worked):

- async on
	- [x] case 1: the transcription is faster then the attach operation 
	- [x] case 2: the transcription is slower then the workflow reaching the attach operation (in this case, the process waits for the transcription job to finish)
- async off
	- [x] everything works as before
- limit to one
	- [x] set to false and async to true -> multiple speech-to-text-job IDs stored in workflow variables, everything works as expected
	- [x] set to false and async to false -> works
	- [x] set to true with async to true -> works
	- [x] set to true with async to false -> works
- attach operation: 
- tags and flavors

### The following things didn't work:

- language fallback
  - removes the possibility to let whisper auto detect the language
- attach operation
  - I don't know if this should be labeled as "not working", but the defined target-tags of the attach-operation are overriding the tags from the speechtotext operation. Don't know what a good solution would look here: Maybe we should add the attach-operation tags to the speechtotext-operation ones? Maybe we let it that way but add a note to the Docs?
- metadata: Wrong language codes. Metadata provides 3 digits and whisper expects 2 digit codes i guess
	- metadata provides "deu" for example.. (but whisper works with language code "de"). Thats not like a Bug that was introduced. We could create another Issue for this or maybe this can be fixed easily?

lkiesow:
> - language fallback
>   - removes the possibility to let whisper auto detect the language

That may be why someone broke the fallback before. But I'll change it so that this value has no default. That way, we can skip it if not present.

> I don't know if this should be labeled as "not working", but the defined target-tags of the attach-operation are overriding the tags from the speechtotext operation. Don't know what a good solution would look here: Maybe we should add the attach-operation tags to the speechtotext-operation ones? Maybe we let it that way but add a note to the Docs?

I think this is a misconception. Target tags apply to media package elements. Since the `speechtotext` operation does not generate any media package element, there are no tags. In other words, setting `target-tags` at that operation just has no effect at all. So, this is expected behavior. I'll add a notice to the docs.

> metadata: Wrong language codes. Metadata provides 3 digits and whisper expects 2 digit codes i guess 

That I didn't touch and won't fix in here. If I remember correctly, the Vosk code does some mapping internally. We could do something similary if we want to? But that's a separate issue.

mtneug:
The language issue was fixed for `WhisperEngine.java` in #5950, but not for `WhisperCppEngine.java`.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
lkiesow: Commit 12bc12857d750851f65ab9e202f5cbd248306341 fixes the problem by removing the default, truly making this an optional argument.

docs/guides/admin/docs/workflowoperationhandlers/speechtotext-attach-woh.md
lkiesow: Fixed with 37642f463882b803fd35cdff1ab280c0e6a9e69b which now clearly states that this option has no effect in asynchronous mode.

lkiesow:
Pull request has been updated.
- The problem with the language fallback should be fixed.
- The `target-tags` documentation got updated to explain what's happening
- I'm not touching the 2 vs 3 letter language code in here. I didn't touch that in the first place and it seems like we already have different approaches. It may therefor make sense to think avout this for a bit and not rush this.

marwyg (approved):

marwyg:
The changes are looking good. I tested the fallback again -> It works as expected.

The PR can be merged if you like.

marwyg:
I created an Issue for the language Code problem: https://github.com/opencast/opencast/issues/6293

PR #6248, opened 2024-10-20
Auto-encode audio track for Whisper

lkiesow:
Whisper.cpp requires a very specific input audio format. Instead of relying on adopters working around this in the workflow and encoding tracks in advance, this patch includes the encoding in the operation itself.

This not only has the advantage that the configuration is much easier, but it also means that only the tracks selected to actually be used for generating the subtitles are encoded.

### How to test this patch

Configure Whisper.ccp and add the `speechtotext` operation to the fast testing workflow:

```yaml
  - id: speechtotext                                                                                                     
    description: Generates subtitles for video and audio files                                                           
    configurations:                                                                                                      
      - source-flavor: '*/source'                                                                                        
      - target-flavor: captions/source                                                                                   
      - target-tags: engage-download                                                                                     
      - target-flavor: captions/source                                                                                   
      - limit-to-one: true 
```

If you now ingest an `.mp4` file, the operation should be able to work on it.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
@@ -49,3 +49,7 @@
 
 # Do not use temperature fallback while decoding
 #whispercpp.no-fallback=false
+
+# Automatically convert media files to wav
+# Default: true
+#whispercpp.auto-encode=true
--- a/docs/guides/admin/docs/releasenotes/whisper.cpp-autoencode.txt
+++ b/docs/guides/admin/docs/releasenotes/whisper.cpp-autoencode.txt
@@ -1,0 +1,2 @@
+The Whisper.cpp integration is now automatically re-encode audio to a suitable format by default. This means that it is
+no longer necessary to include a separate encode operation for this in your workflow.
--- a/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
+++ b/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
@@ -31,6 +31,7 @@ import org.opencastproject.util.data.functions.Strings;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.BooleanUtils;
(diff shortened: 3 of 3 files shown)


etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
```diff
@@ -1,12 +1,18 @@
+<<<<<<< HEAD
 # Configuration for setting a custom path to the Whisper command line tool
 # Default: whisper
 #whispercpp.root.path=whispercpp
+=======
+# Configuration for setting a custom path to the Whisper.cpp command line tool
+# Default: whisper.cpp
+whispercpp.root.path=/home/lars/dev/whisper.cpp/main
+>>>>>>> f7b28fafd2 (Auto-encode audio track for Whisper)
```
geichelberger: Unresolved merge conflict residues.

geichelberger (dismissed):

etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
lkiesow: Ups, that's a leftover from some local tests.
Fixed.

marwyg (approved):
I tested following scenarios with the fast Workflow:
- auto encode: true -> works as expected
- auto encode: false + .mp4 file -> fails as expected saying "wrong media extension"
- auto encode: false + .wav file -> works as expected

The last 2 test did ensure, that the behavior stays like before, when setting the config to false. But the default is true, so I guess the target branch (develop) is okay then? I like the idea, that auto-encode should be the default behavior. So for my part it can stay like this.

I guess it would be nice if this feature would be mentioned somewhere in the release notes? If adopter use whispercpp and doesn't change this, it would not break something I think. There would be only an unnecessary encoding operation in front of their speechtotext operation.

Regarding the code, I have nothing to complain about.

mtneug:
> I guess it would be nice if this feature would be mentioned somewhere in the release notes?

+0

lkiesow:
> I guess it would be nice if this feature would be mentioned somewhere in the release notes? 

I added a note to `docs/guides/admin/docs/releasenotes`.
I hope this can be merged like this.

lkiesow:
@marwyg, are you okay with the current state and can this be merged?

marwyg:
Yes, everything is fine. I don't see anything that would speak against a merge.

KatrinIhler:
In that case...lgtm

PR #6251, opened 2024-10-23
Improve event catalog handling in External API

mtneug:
This is an updated version of closed #4417

This adds four improvements to event catalog handling in the External API

1. Avoid DB query is not necessary.
2. Cache map of organization to list of catalogs instead of always calculating it.
3. Avoid calling list providers to fetch collection data if this is later on not used (e.g. all series).
4. Remove collection data from responses for extended metadata.

The last point can be seen as a change to the External API. However, one can also see this as a bug, as other methods in the class hide the collection data from extended metadata.

All changes can add up when requesting multiple events with `/events` especially with many events and series.

### How to test this patch

Performance degradation is most noticeable when there are a lot of series in Opencast.

1. Create 1000 series (e.g.  using https://github.com/opencast/helper-scripts/tree/master/create-series)
2. Use External API to retrieve event infos and include `withmetadata=true`
3. Apply patch and repeat 2.
4. Compare the performance of 2. and 3.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


geichelberger (approved):
LGTM

PR #6253, opened 2024-10-24
Add per-organization bucket name configuration

mtneug:
Add the ability to configure S3 buckets for each organization (asset manager and distribution). The old configuration is still compatible and will result in a default bucket configuration, i.e. if no org-specific buckets are configured, the default bucket will be used.

This likely depends on #6252 (I have not tested without this additional patch).

### How to test this patch

1. Configure two tenants (e.g. using https://github.com/opencast/community-integrations/tree/main/multi-tenant)
2. Configure different S3 buckets for each tenant
3. Upload videos to each tenant
4. Observe which S3 buckets are used

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
+++ b/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
@@ -79,9 +79,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.Dictionary;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 @Component(
@@ -108,6 +111,7 @@ public class AwsS3AssetStore extends AwsAbstractArchive implements RemoteAssetSt
   public static final String AWS_S3_SECRET_ACCESS_KEY_CONFIG = "org.opencastproject.assetmanager.aws.s3.secret.key";
   public static final String AWS_S3_REGION_CONFIG = "org.opencastproject.assetmanager.aws.s3.region";
   public static final String AWS_S3_BUCKET_CONFIG = "org.opencastproject.assetmanager.aws.s3.bucket";
+  public static final String AWS_S3_BUCKET_CONFIG_PREFIX = "org.opencastproject.assetmanager.aws.s3.bucket.";
   public static final String AWS_S3_ENDPOINT_CONFIG = "org.opencastproject.assetmanager.aws.s3.endpoint";
   public static final String AWS_S3_PATH_STYLE_CONFIG = "org.opencastproject.assetmanager.aws.s3.path.style";
   public static final String AWS_S3_MAX_CONNECTIONS = "org.opencastproject.assetmanager.aws.s3.max.connections";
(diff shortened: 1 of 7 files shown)


KatrinIhler:
ICLA check still fails, wtf?

KatrinIhler:

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -180,9 +186,21 @@ public void activate(final ComponentContext cc) throws IllegalStateException, Co
...
+      // AWS S3 org bucket name mapping
+      Collections.list(cc.getProperties().keys()).stream()
+          .filter(s -> s.startsWith(AWS_S3_BUCKET_CONFIG_PREFIX))
+          .forEach(s -> {
+            String orgId = s.substring(AWS_S3_BUCKET_CONFIG_PREFIX.length());
+            String bucketName = OsgiUtil.getComponentContextProperty(cc, s);
+            orgBucketNameMap.put(orgId, bucketName);
+          });
 
       // AWS region
```
gregorydlogan, line 209: So what if I don't define *any* buckets by accident?  Can we check at this point to make sure at least one bucket definition exists?  The previous behaviour of the `createAWSBucket` method (poorly) enforced this by throwing a NPE when `s3.listObjects(null)` was called :)

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -257,9 +262,21 @@ public void activate(ComponentContext cc) {
...
+      // AWS S3 org bucket name mapping
+      Collections.list(cc.getProperties().keys()).stream()
+          .filter(s -> s.startsWith(AWS_S3_BUCKET_CONFIG_PREFIX))
+          .forEach(s -> {
+            String orgId = s.substring(AWS_S3_BUCKET_CONFIG_PREFIX.length());
+            String bucketName = OsgiUtil.getComponentContextProperty(cc, s);
+            orgBucketNameMap.put(orgId, bucketName);
+          });
 
       // AWS region
```
gregorydlogan, line 285: Similar complaint to above, lets verify that at least one is set.

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -544,12 +572,24 @@ private void restoreGlacierObject(String objectName, Integer objectRestorePeriod
     }
   }
 
+  private String getBucketName(String orgId) {
```
gregorydlogan, line 579: This needs release notes.  Users with existing S3 configs *must* ensure that the default bucket is still available (either as the default, or as an org bucket) otherwise the data won't be accessible since this method won't return the right bucket name

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
```diff
@@ -994,6 +1021,18 @@ protected void setStorageTmp(String path) {
     }
   }
 
+  private String getBucketName(String orgId) {
```
gregorydlogan, line 1028: Same comment about release notes.

gregorydlogan (changes requested):

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
mtneug: Users with unchanged config use `org.opencastproject.assetmanager.aws.s3.bucket`. This is now stored in the map as the default bucket. I don't think users need to change anything if the current behavior is good enough for them.

modules/distribution-service-aws-s3/src/main/java/org/opencastproject/distribution/aws/s3/AwsS3DistributionServiceImpl.java
mtneug: Same implementation with the `org.opencastproject.distribution.aws.s3.bucket` config key.

mtneug:
@gregorydlogan I rebased and added the checks for bucket configuration. As for the release notes, I can add a line, but this change should not be breaking with existing configuration.

gregorydlogan:
If we don't need to alter configs to keep things working then IMO this is just a new feature and doens't need the notes.

PR #6255, opened 2024-10-24
Fix NPE when MP element can't be found in local asset manager store

mtneug:
Opencast currently assumes that assets can be found locally when checking for the correct local path. This is not true when using remote asset stores (e.g. S3), resulting in an NPE. This patch adds a null check.

### How to test this patch

1. Configure S3 asset store
2. Force all files to move to S3
3. Try to retrieve assets from asset manager endpoint

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/shared-filesystem-utils/src/main/java/org/opencastproject/assetmanager/util/AssetPathUtils.java
+++ b/modules/shared-filesystem-utils/src/main/java/org/opencastproject/assetmanager/util/AssetPathUtils.java
@@ -137,6 +137,11 @@ public final class AssetPathUtils {
         localPath = path;
       }
     }
+    // if remote asset stores are used, localPath could be null (i.e. all snapshots live remotely)
+    if (localPath == null) {
+      logger.debug("Local snapshot for {} not available.", uri);
+      return null;
+    }
     final File file = Paths.get(localPath, organizationId, mediaPackageID, version, filename).toFile();
     if (file.isFile()) {
       logger.debug("Converted {} to local file at {}", uri, file);


modules/shared-filesystem-utils/src/main/java/org/opencastproject/assetmanager/util/AssetPathUtils.java
```diff
@@ -130,7 +130,7 @@ public static File getLocalFile(final List<String> localPaths, final String orga
     final String version = assetPath[5];
     final String filename = mediaPackageElementID + '.' + FilenameUtils.getExtension(assetPath[6]);
     // Check under which path the mediapackage exists
-    String localPath = null;
+    String localPath = localPaths.get(0);
```
lkiesow: This seems wrong and feels like this will surely have side effects. The function accepts a list of potential local paths, iterating over the list and filters them using some criteria. With your patch, the first entry just skips all checks and requirements, so that the results does no longer guarantee these properties.

I don't know this code well, but I would expect `null` to be a valid return value if there is no valid local path for the (remote) file.

mtneug: `null` is not valid. Line 133 initializes the value to `null`. Lines 134-139, check each path in `localPaths` to see if the mediapackage folder exists in that path. The Line 140 assumes that the loop actually found a path, which is not a valid assumption in the case of remote asset managers. This line will throw an exception if `localPath` is `null`. Line 141 then again checks if the actual file exists. For this reason, it is ok to set the default `localPath` to the first one.

mtneug: We discussed this in the dev meeting and agreed that this approach is confusing. I changed the path to a simple null check.

gregorydlogan (approved):

lkiesow:
If you are waiting for my comment to merge this, I'm fine if you discussed this at the technical meeting. I don't need another look. I trust Greg.

PR #6262, opened 2024-11-01
Fix Amberscript Transcription Jobs Hanging

lkiesow:
This patch fixes the problem that AmberScript transcription jobs will never finish if the event has been deleted between starting and finishing the transcription.

The code checked if a snapshot exists, skipping the operation and repeating this on the next run again. An existing event will always have at least one snapshot, since one is created at ingest or even earlier if the event was scheduled. If you delete the event, however, there will be no snapshot, but there will obviously also never be any snapshot. That is why the jobs will just wait forever.

### How to test this patch

- Configure AmberScript
- Add a new media (triggering the transcription service)
- Delete the event as soon as it finished processing

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


lkiesow:
If you test this, you may also want to test #6299 at the same time

jduehring (approved):
Code changes make sense and are looking good to me. 

I tested this PR for a bunch of videos with your provided configuration in https://github.com/opencast/opencast/pull/6299 . 
Everything works as expected: No hanging jobs and the warning gets logged.

lkiesow:
Thanks for testing this. Merging.

Issue #6260, opened 2024-10-29
Session timeout value has no effect on OC 15 and OC 16 and workaround


wsmirnow:
Opencast configuration [etc/org.ops4j.pax.web.cfg](https://github.com/opencast/opencast/blob/r/15.x/etc/org.ops4j.pax.web.cfg) has an option to define the user session timeout value in https://github.com/opencast/opencast/blob/4bd7a76f43c217c84fa082f8c3bef482bdda61c5/etc/org.ops4j.pax.web.cfg#L54-L55

Setting this value has no effect on Opencast 15 and 16. The root cause are two bugs in underlying framework pax-web. The issues are documented and already fixed in the newer versions but it has to find a way back to Opencast.

* https://github.com/ops4j/org.ops4j.pax.web/issues/1912
  * The value of `org.ops4j.pax.web.session.timeout` will be overwritten by the default value of 30 minutes
  * fixed in pax-web 8.0.25, is part of Karaf 4.4.6 -> available in Opencast 16
* https://github.com/ops4j/org.ops4j.pax.web/issues/1952
  * The value of `org.ops4j.pax.web.session.timeout` is interpreted as seconds and not minutes as it should be
  * fixed in pax-web 8.0.28, not available in latest Karaf release, probably is part of Karaf 4.4.7

## How to reproduce

1. Edit `/etc/opencast/org.ops4j.pax.web.cfg` and set  `org.ops4j.pax.web.session.timeout=1` (1 minute session timeout).
2. Restart Opencast
3. Login to Opencast **by unchecking the remember-me checkbox!**
4. Wait a bit longer then 1 minute. 
5. Your session should be invalid and you should be redirected to the login page.

But this is not the case. There are workarounds available.

### Opencast 15 workaround

Run this commands and restart Opencast (updated)
```
sudo sed -i 's/8.0.22/8.0.29/g' /usr/share/opencast/system/org/apache/karaf/features/standard/4.4.4/standard-4.4.4-features.xml
sudo sed -i 's/8.0.22/8.0.29/g' /etc/opencast/org.apache.karaf.features.cfg
```

### Opencast 16 workaround

Set the value of `org.ops4j.pax.web.session.timeout` in seconds in `/etc/opencast/org.ops4j.pax.web.cfg`.

marwyg:
[...]

PR #6292, opened 2024-11-05
Fix default session timeout

lkiesow:
Due to a bug in Pax Web, the default session timeout for Opencast was set to 4 minutes instead of 4 hours. This caused users to be logged out automatically if they do not have an additional remember-me cookie (e.g. LTI users don't have that). This basically means that users are unable to watch videos longer than four minutes for which they need authentication.

This is a temporary fix for #6260

### How to test this patch

To test this (check that the value is indeed seconds):

- Set `org.ops4j.pax.web.session.timeout=20`
- Go to http://localhost:8080, make sure to **remove the check mark for remember me** and log-in
- Wait for 21 seconds
- Go to http://localhost:8080 again
  - You will be logged out and asked to log in again

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/org.ops4j.pax.web.cfg
+++ b/etc/org.ops4j.pax.web.cfg
@@ -51,5 +51,8 @@ org.ops4j.pax.web.ssl.key.password=password
 # Prevent cookie from being accessed by client side scripts, default false
 org.ops4j.pax.web.session.cookie.httpOnly=true
 
-# Timeout for user sessions in minutes.
-org.ops4j.pax.web.session.timeout=240
+# Timeout for user sessions in seconds
+# Note: This is seconds, not minutes due to a bug in Pax Web:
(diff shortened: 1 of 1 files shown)


gregorydlogan (approved):
Works as expected.  So when/if Pax fixes this our timeouts are going to be *enormous* unless we change it back.  But that's not necessarily the end of the world.

wsmirnow (approved):
Thank you @lkiesow

lkiesow:
Even if they fix it, we have to upgrade Pax Web (which for us essentially means updating Karaf) to get this fix. In that case, yes, we should update this configuration file again. But that's not happening in 16 or 17.

mtneug:
I opened #6295 to track the revert for probably 18.

PR #6296, opened 2024-11-07
Upgrade upload-artifacts from v3 to v4

gregorydlogan:
This PR updates the `upload-artifacts` workflow action from v3 to v4.  v3 will be deprecated Dec 5, with brownouts occurring sooner.  This PR should not change behaviour in terms of uploaded builds.

This PR should go into `r/15.x` because the v3 will likely stop working around Dec 5, and this will also prevent hte brownouts (Nov 14 and 21) from causing unneccessary workflow failures.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -173,7 +173,8 @@ jobs:
         if: >
           matrix.java == 11
           && github.repository == 'opencast/opencast'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: opencast-allinone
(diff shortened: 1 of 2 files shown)


LukasKalbertodt:
Quick note in case this affects us: hidden files are now excluded by default: https://github.com/actions/upload-artifact/issues/602

This caused me trouble before. No idea if it causes trouble here, just wanted to leave this information here.

gregorydlogan:
It shouldn't, since we shouldn't have any hidden files in our builds.  IMO, if we're relying on hidden files then that's a bug that ought to be addressed.

PR #6317, opened 2024-11-19
Add tobira series endpoint to remove paths

owi92:
This will allow removing ("unmounting") a series in Tobira.
The necessary endpoints in Tobira itself were already existing and also called by this endpoint in Opencast.
This commit just makes some form params optional.

### How to test this patch
<sub>The following is a bunch of information that is useful to test this, but in the interest of time I'd suggest that this should be tested by @Arnei, since he has done so before and won't need the extra preparation.</sub>
If you want to test this and haven't done any Tobira testing before (though that is probably too much work for this small PR for anyone that hasn't tested these Tobira endpoints before):
- You'll need at least Tobira 2.11 and configure it to talk to your Opencast with JWT setup (see our docs for reference: https://elan-ev.github.io/tobira/setup/auth/jwt)
- The relevant endpoint is `updateSeriesTobiraPath` and can be tested via `{your-opencast}/docs.html?path=/admin-ng/series#updateSeriesTobiraPath-10`
- In order to use that endpoint for series unmounting:
  -  you'll need the `seriesId` (Opencast ID) of a mounted series in Tobira
  - The page must not have any other content that the series
  - `pathComponents` must be empty
  - `currentPath` needs to be the path of that series in Tobira (i.e. "/page/my_series")
  - `targetPath` must be empty

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
@@ -1145,7 +1145,7 @@ public Response getSeriesHostPages(@PathParam("seriesId") String seriesId) {
           restParameters = {
                   @RestParameter(
                           name = "pathComponents",
-                          isRequired = false,
+                          isRequired = true,
                           description = "List of realms with name and path segment on path to series.",
                           type = TEXT),
                   @RestParameter(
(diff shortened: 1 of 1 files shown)


gregorydlogan:
As requested in the body ;)

Arnei (changes requested):
Works. I would prefer to have separate endpoints for updating and deleting, to make the whole thing a little more CRUD (in line with the other endpoints). But if that does not make sense because of how Tobira handles things, then so be it.

Arnei (approved):

PR #6331, opened 2024-11-28
Adding LTI-tools configuration to exclude Live Streams in the LTI-Series-Tool

ciegler:
This PR adds a configuration file for the LTI-tools `etc/ui-config/mh_default_org/ltitools/config.json` and the required changes to exclude Live Streams in the LTI-series-tool when this is set in the configuration file. The configuration file includes the Boolean value `excludeLiveStreams` , which is by default false. When set to true Live Streams are filtered out and not shown in the LTI-series-Tools.

### DIFF
--- a/modules/lti/src/OpencastRest.ts
+++ b/modules/lti/src/OpencastRest.ts
@@ -30,7 +30,7 @@ export interface SearchEpisodeResult {
     readonly mediapackage: MediaPackage;
     readonly languageShortCode: string;
     readonly licenseKey: string;
-    readonly isLive: boolean;
+    readonly live: boolean;
 }
 
 export interface SearchEpisodeResults {
@@ -149,7 +149,7 @@ export async function getConfig(): Promise<Config> {
 }
 
 export function filterLiveEvents(sr: SearchEpisodeResults) {
-    const filtered_results = sr.results.filter(result => result.isLive === false);
+    const filtered_results = sr.results.filter(result => !result.live);
     const filtered_sr : SearchEpisodeResults = {results: filtered_results, total : filtered_results.length, limit : sr.limit, offset : sr.offset};
     return filtered_sr
 }
@@ -203,6 +203,22 @@ const parseTracksFromResult = (result: any) => {
   return undefined;
 }
 
+const parseLiveFromResults = (result :any) : boolean => {
+    if (Array.isArray(result.mediapackage.media.track)) {
+        let live : boolean = false;
+        result.mediapackage.media.track.forEach( (track: any) => {
+            if ('live' in track && track.live === true) {
+                live = true;
+            }
+        } )
+        return live;
+    } else if (result.mediapackage.media.track !== null) {
+        if (!('live' in result.mediapackage.media.track)) { return false }
+        return result.mediapackage.media.track.live === true;
+    }
+    return false;
+}
+
 export async function searchEpisode(
     limit: number,
     offset: number,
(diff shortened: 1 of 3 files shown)


KatrinIhler:
Test keeps failing, maybe this indeed breaks something or the test needs to be updated. @ciegler Can you take another look at this when you have time?

modules/lti/src/OpencastRest.ts
```diff
@@ -224,7 +243,8 @@ export async function searchEpisode(
                         url: attachment.url
                     })),
                 tracks: parseTracksFromResult(result)
-            }
+            },
+            isLive: result.mediapackage?.media?.track?.some((track: any) => track.live)
```
lkiesow: I think this line is breaking the tests.
Launching `selenium-tests gui`, you see the error:
```
Error: _result$mediapackage$2.some is not a function
```
This will also break in production depending on the data source, I think.

lkiesow (changes requested):

modules/lti/src/OpencastRest.ts
ciegler: Thank you for pointing that out, I was not sure where the error was coming from. By that I also realized that I could run into cases where the `SearchEpisodeResult.isLive` is undefined, which I do not want, so I also changed that.

ciegler:
Sorry for the commit chaos, I just found out that I can actually squash several commits into one. In future I will do this before creating a PR, so that the commit is then laid out clearer.

modules/lti/src/OpencastRest.ts
```diff
@@ -224,7 +243,8 @@ export async function searchEpisode(
                         url: attachment.url
                     })),
                 tracks: parseTracksFromResult(result)
-            }
+            },
+            isLive: result.mediapackage.media.track !== undefined ? result.mediapackage.media.track.isLive : false
```
lkiesow: I haven't actually tested this, but I'm not sure if this will always work. It's quite annoying, but `track` may be a JS object *or* an array. That's why there are all the weird checks above. I think code will not work if there are multiple tracks (i.E. `track` is an array).

KatrinIhler:
Apparently Martin has also found a problem while testing this, he will comment the details later this week.

majosch:
@ciegler Thanks for the update! Just tested this on a fresh `r/17.x` and works as expected :tada: 

[Bildschirmaufzeichnung vom 2024-12-13 11-21-10.webm](https://github.com/user-attachments/assets/f2241c1b-9102-4381-9a98-87abcc85b12b)

KatrinIhler:
@lkiesow Have your concerns been sufficiently addressed? :)

KatrinIhler:
@lkiesow Ping :D

majosch:
I just noticed, that series with more than 15 events are broken, because the navigation displays only one page.

When setting `excludeLiveStreams` to false, the navigation behaves as expected.

[Bildschirmaufzeichnung vom 2025-02-21 11-33-55.webm](https://github.com/user-attachments/assets/f970670a-4b72-4153-884e-f103d669dbde)

ciegler:
Thanks for caching the bug. I changed the `total` value for the `SearchEpisodeResults` in the original commit and reverted these changes now. This should fix the bug in the navigation.

ciegler:
Since filtering out the live episodes in the LTI frontend after the request to the `search index` lead to a wrong total page and total Episodes count, it was requested to implement the possibility to filter out live events already in the search index. Therefore I also had to integrate the configuration file differently. I opened a new PR with the newest changes (https://github.com/opencast/opencast/pull/7032) and will close the PR now.

PR #6362, opened 2024-12-03
Add "Show for new or existing" attribute to asset upload options

Arnei:
Adds the attributes `showForNewEvents` and `showForExistingEvents` to the event upload options, which influence what is displayed in the "Create Event" and "Event Details" modals in the admin ui. The main goal here is to let  users specify an asset upload option for existing events but not new events and vice versa.

**This accompanies the admin ui PR https://github.com/opencast/opencast-admin-interface/pull/698. This will not work without it.**

### Why is this targeting legacy?
- So it can go along with its related admin ui PR. Really this should point at whatever version its related admin ui PR gets merged into.
- This is a configuration only change that does not impact existing configuration at all.

### How to test this patch
The new attributes will only have any effect with the related admin ui PR. Please see the admin ui PR for more details on how to test it.

### DIFF
--- a/etc/listproviders/event.upload.asset.options.properties
+++ b/etc/listproviders/event.upload.asset.options.properties
@@ -21,6 +21,7 @@
 # - Attribute boolean "multiple" is for uploading multiple source track files to the same flavor
 # - Attribute "tags" adds one or multiple tags to the uploaded asset
 # - - Multiple tags are delimited by "," (no whitespace)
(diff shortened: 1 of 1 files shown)


gregorydlogan:
@lkiesow are we likely seeing another admin ui updated in 16.x or should this be retargetted?

gregorydlogan:
As discussed in the technical meeting, we're *probably* not doing any further admin ui releases, and since this breaks things without such a release we should move this to `r/17.x`

Arnei:
Rebased onto r/17.x.

Issue #5665, opened 2024-03-05
Partial-Retract WOH does not work in combination with publication to wowza 


thet0ast3r:
**Describe the bug**

Attempting to use 'partial-retract' in a workflow, even if no flavors or tags are selected to partially retract (expected behavior: no change) results in a broken , non-playable, wowza stream. 

**To Reproduce**

1. run opencast in conjunction with wowza to serve videos.
2. Create an event with a working, playable publication. 
3. add a workflow definition that attempts to partially retract any flavor/tag/file or even nothing. 
4. Run the workflow.
5. Attempt to play the video - broken.

**Expected behavior**

Running the partial-retract wf operation should do what it states: just remove parts of the publication according to the selected flavors/tags in the wf-operation configuration. 

**Stack trace**

OC log from running a minimal workflow that does only partial-retract on a fresh event:

```
2024-03-05T12:24:34,237 | INFO  | (WorkflowServiceImpl:682) - Scheduling workflow 823133 for execution
2024-03-05T12:24:36,272 | INFO  | (PartialRetractEngageWorkflowOperationHandler:118) - Partially retracting 46639333-866e-4d9d-9dbe-d75f19e2ff21
2024-03-05T12:24:38,378 | WARN  | (WowzaStreamingDistributionService:931) - /datadisk/exports/matterhorn/content/streams/mh_default_org/engage-player/46639333-866e-4d9d-9dbe-d75f19e2ff21/cb17c298-fd1c-49ad-a236-077de2789f91/cam.mp4 does not exist but was to be deleted
2024-03-05T12:24:38,452 | WARN  | (WowzaStreamingDistributionService:931) - /datadisk/exports/matterhorn/content/streams/mh_default_org/engage-player/46639333-866e-4d9d-9dbe-d75f19e2ff21/0b1e7f18-d8c8-4bd4-9826-3e8e42eafea3/cam.mp4 does not exist but was to be deleted
2024-03-05T12:24:38,539 | WARN  | (WowzaStreamingDistributionService:931) - /datadisk/exports/matterhorn/content/streams/mh_default_org/engage-player/46639333-866e-4d9d-9dbe-d75f19e2ff21/636a6429-c49e-40c2-b1bb-89b7d0f8af9a/cam.mp4 does not exist but was to be deleted
[...]

PR #6363, opened 2024-12-03
Fix partial retract streaming elements

geichelberger:
The retract-partial operation removes all streaming distribution files instead of only the media package elements,
 which are requested for retracting.

The strange part of the retract operation is that we lose the information on which element belongs to which distribution, and we have to try all IDs for each distribution.

### How to test this patch

1. Setup Opencast with a streaming distribution
2. Add a video with subtitles (or create one via the editor)
3. Publish the video
4. Check if the subtitle is visible in the player
5. Run the workflow below and remove the caption
6. The publication should be broken, and the video is not playable
7. Apply the patch and repeat the steps 2 - 5,
8. Check if the video is playable and the subtitle is removed 

```yaml
---
id: remove-assets
title: Remove-assets
operations:
  - id: tag
    description: "Changing flavor of source elements: captions/delivery"
    configurations:
      - source-flavors: captions/delivery
      - target-tags: -archive

  - id: retract-partial
    description: "Retracting elements from Engage: captions/delivery" 
    configurations:
      - retract-flavors: captions/delivery
      - retract-tags: type:subtitle
      
 - id: snapshot
    description: Archive publishing information
    configurations:
      - source-tags: archive      
```

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/RetractEngageWorkflowOperationHandler.java
+++ b/modules/distribution-workflowoperation/src/main/java/org/opencastproject/workflow/handler/distribution/RetractEngageWorkflowOperationHandler.java
@@ -50,6 +50,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


geichelberger:
Maybe fixes https://github.com/opencast/opencast/issues/5665 too.

majosch:
I have tested this patch on our staging instance with a fresh r/16.x build.

I can confirm that - without this patch - `partial-retract` is removing all streaming distribution files. Like in #5665 described, this also happens if running an "empty" `partial-retract`-WOH, although I have not seen any errors and warning in the logs as reported there.

After applying this patch `partial-retract` works as expected and removes only assets as requested.

PR #6379, opened 2024-12-09
Fix security config to actually make `/tobira/version` public

LukasKalbertodt:
This was the intention all along, but apparently no one tested it and the two entries were in the config in the wrong order.

### How to test this patch

```
curl https://develop.opencast.org/tobira/version -v
```

-> 302 redirect to login page

With this page:

```
curl localhost:8080/tobira/version
```

-> `{"version":"1.6"}`

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -330,8 +330,8 @@
     <sec:intercept-url pattern="/editor-ui/**" access="ROLE_ADMIN, ROLE_USER" />
 
     <!-- Enable access to the Tobira module -->
-    <sec:intercept-url pattern="/tobira/**" access="ROLE_ADMIN" />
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Per discussion in the technical meeting, this may become important within the next year for Tobira work.  Since we're modifying the config files this would normally be barred from going into not-`develop`, *but*...

- Opencast's functionality is unaffected regardless of patch application or not
- It's a easily applied change to the config file

@KatrinIhler @lkiesow thoughts?  Objections to moving this into something earlier?

KatrinIhler:
No objections from me.

LukasKalbertodt:
Soooo should I change the branch? I don't really care.

gregorydlogan:
I think so. This one won't hurt regardless of whether it's applied or not.

PR #6382, opened 2024-12-11
Add user info to roles for admin ui

Arnei:
Adds user information for user roles to admin ui GET endpoints. The frontend wants to render user roles like ROLE_USER_ARNE with user information like username and email, so instead of having to do perform multiple GET requests and then mix-and-matching roles and users in the frontend, it seemed prudent to change the backend instead.

To that end, ACLs are not sent as is anymore, but transformed into the format used by the frontend so that user information can be more easily attached.

**Requires frontend changes, otherwise this PR will break the frontend: https://github.com/opencast/opencast-admin-interface/pull/1006**

### How to test this patch

Check the frontend PR for more information on how to test this patch.

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -2274,7 +2274,6 @@ public abstract class AbstractEventEndpoint {
     JSONObject episodeAccessJson = new JSONObject();
     episodeAccessJson.put("current_acl", currentAcl.isSome() ? currentAcl.get().getId() : 0L);
     episodeAccessJson.put("acl", transformAccessControList(activeAcl, getUserDirectoryService()));
-//    episodeAccessJson.put("acl", AccessControlParser.toJsonSilent(activeAcl));
     episodeAccessJson.put("privileges", AccessInformationUtil.serializePrivilegesByRole(activeAcl));
     if (StringUtils.isNotBlank(optEvent.get().getWorkflowState())
             && WorkflowUtil.isActive(WorkflowInstance.WorkflowState.valueOf(optEvent.get().getWorkflowState())))
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -40,6 +40,7 @@ import static javax.servlet.http.HttpServletResponse.SC_SERVICE_UNAVAILABLE;
 import static javax.servlet.http.HttpServletResponse.SC_UNAUTHORIZED;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
+import static org.opencastproject.adminui.endpoint.EndpointUtil.transformAccessControList;
 import static org.opencastproject.index.service.util.RestUtils.conflictJson;
 import static org.opencastproject.index.service.util.RestUtils.notFound;
 import static org.opencastproject.index.service.util.RestUtils.notFoundJson;
@@ -2272,7 +2273,8 @@ public abstract class AbstractEventEndpoint {
 
     JSONObject episodeAccessJson = new JSONObject();
     episodeAccessJson.put("current_acl", currentAcl.isSome() ? currentAcl.get().getId() : 0L);
-    episodeAccessJson.put("acl", AccessControlParser.toJsonSilent(activeAcl));
+    episodeAccessJson.put("acl", transformAccessControList(activeAcl, getUserDirectoryService()));
+//    episodeAccessJson.put("acl", AccessControlParser.toJsonSilent(activeAcl));
     episodeAccessJson.put("privileges", AccessInformationUtil.serializePrivilegesByRole(activeAcl));
     if (StringUtils.isNotBlank(optEvent.get().getWorkflowState())
             && WorkflowUtil.isActive(WorkflowInstance.WorkflowState.valueOf(optEvent.get().getWorkflowState())))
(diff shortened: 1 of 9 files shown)


gregorydlogan:
This works correctly, however it needs to be held until the next admin ui release, and should be merged concurrently with the (future) PR where that release is brought into 17.x!

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
```diff
@@ -2272,7 +2273,8 @@ public Response getEventAccessInformation(@PathParam("eventId") String eventId)
 
     JSONObject episodeAccessJson = new JSONObject();
     episodeAccessJson.put("current_acl", currentAcl.isSome() ? currentAcl.get().getId() : 0L);
-    episodeAccessJson.put("acl", AccessControlParser.toJsonSilent(activeAcl));
+    episodeAccessJson.put("acl", transformAccessControList(activeAcl, getUserDirectoryService()));
+//    episodeAccessJson.put("acl", AccessControlParser.toJsonSilent(activeAcl));
```
marwyg: Should this comment stay like this?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
```diff
@@ -91,4 +100,56 @@ public static void addRequestFiltersToQuery(final String filterString, ResourceL
...
+   * Transform ACL into the format the admin ui frontend uses.
+   * We do this in the backend so we can attach information about users to user roles.
+   */
+  public static JSONArray transformAccessControList(AccessControlList acl, UserDirectoryService userDirectoryService) {
+    class TransformedAcl {
+      protected String role;
+      protected boolean read = false;
+      protected boolean write = false;
+      protected List<String> actions = new ArrayList();
+    }
```
marwyg, line 114: You could remove the "protected". Its easier to read (and doesn't make any difference I think)

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/SeriesEndpoint.java
```diff
@@ -331,7 +340,8 @@ public Response getSeriesAccessInformation(@PathParam("seriesId") String seriesI
               adminUIConfiguration.getMatchManagedAclRolePrefixes());
       seriesAccessJson.put("current_acl", currentAcl.isSome() ? currentAcl.get().getId() : 0);
       seriesAccessJson.put("privileges", AccessInformationUtil.serializePrivilegesByRole(seriesAccessControl));
-      seriesAccessJson.put("acl", AccessControlParser.toJsonSilent(seriesAccessControl));
+      seriesAccessJson.put("acl", transformAccessControList(seriesAccessControl, userDirectoryService));
+//      seriesAccessJson.put("acl", AccessControlParser.toJsonSilent(seriesAccessControl));
```
marwyg: Should this comment also stay in?

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
```diff
@@ -313,6 +315,38 @@ public Response getUsers(@QueryParam("filter") String filter, @QueryParam("sort"
...
+        if (user != null) {
+          roleUserMap.put(role, generateJsonUser(user));
+        } else {
+          roleUserMap.put(role, null);
+        }
+      }
+    }
+
+    return Response.ok(gson.toJson(roleUserMap)).build();
+  }
```
marwyg, line 348: Could this be a security issue? Because I see no security check, but there is a check in "loadUser()" I think. I will test this. Other endpoints in this class are also checking the input and returning a "bad request" in case the input is empty for example.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
```diff
@@ -511,7 +545,7 @@ private Set<JpaRole> parseJsonRoles(final String roles) throws IllegalArgumentEx
     return rolesSet;
   }
 
-  private Map<String, Object> generateJsonUser(User user) {
+  public Map<String, Object> generateJsonUser(User user) {
```
marwyg: Is there a reason changing this to public? I didn't find a usage outside this class.

marwyg (changes requested):
I looked through the code and have some questions and suggestions. There is one case I would also like to test, but not today anymore.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
marwyg: Okay never mind, I did take a second look at this and especially at the other endpoints in this class.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/EndpointUtil.java
Arnei: Checkstyle requires us to specify access level modifiers.

modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
Arnei: The original idea was to use this function in the event and series endpoints as well, as they now send user data. That did not end up working out. Changed it per your suggestion.

snoesberger:
I did test the functionality of this PR together with the admin UI changes (https://github.com/opencast/opencast-admin-interface/pull/1006) successfully. The changes worked as expected in the new admin UI.
But first I had to do a rebase to r/17.x (locally on my side), some of the files, like `modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java`, are outdated.

marwyg (approved):
I took a second look. My points were addressed and things were changed where possible. It can be merged.

PR #6385, opened 2024-12-17
Allow live events to be published by the CA user

### LINKED ISSUE #6386: Live Events cannot be retracted and directly published
For the sake of completeness, I added an Issue for the PR I wrote: #6385 

**Describe the bug**
When scheduling a live event, the log outputs some errors at the retracting step. When a workflows was chosen, where the live events will be published directly after the recording, it fails at the `publish-engage` step later.

**To Reproduce**

0. Configure a capture agent and a workflow to allow live events
1. Schedule a live event
2. Choose a workflow with the `straight-to-publish` feature
3. Observe the errors in the retracting step
4. Observe the workflow fail at the `publish-engage` step

**Expected behavior**
The live events should be recorded and published normally.

**Stack trace**
tbd

**Server environment:**
- OC Version 16
- Tested on different machines, the error also occurs locally with pyca

marwyg:
Fixes #6386

Some recently implemented permission checks prevent the Capture Agent (CA) user from retracting and publishing live events. This commit introduces the following changes:

- LiveScheduleServiceImpl: The existing behavior of promoting the CA user to the System User when retrieving live media packages has been extended to also apply when retracting live media packages. This ensures that necessary actions can be performed during both retrieval and retraction, despite stricter permission checks.

- SearchServiceIndex: Previously, the write permission check for the CA user would fail when attempting to publish a new media package with the workflow operation `publish-engage`. This was because the index service checked the ACL of the previously deleted live media package.

  - A new `exists` method has been implemented, which returns `false` both for media packages that never existed and those that have been deleted.
  - This change ensures the CA user always has write permission for publishing new media packages.

### How to Reproduce
Schedule a live event and use a workflow, that straight away publishes the live event after recording. You should see errors while retracting the live event after the recording finishes. Later, you will see the workflow fail at the publish-engage step.

### Why it's important to go into this branch
We have the strategy, to always use the 2nd latest Opencast. In Version 16, the live events are not working without this commit.

### How to test this patch
You need to configure a capture agent and enable live events via workflows. With this commit, everything should work correctly.

### DIFF
--- a/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
+++ b/modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
@@ -450,29 +450,43 @@ public class LiveScheduleServiceImpl implements LiveScheduleService {
   }
 
   void retract(MediaPackage mp) throws LiveScheduleException {
+    Organization org = securityService.getOrganization();
+    User prevUser = org != null ? securityService.getUser() : null;
     try {
-      List<Job> jobs = new ArrayList<Job>();
+      securityService.setUser(SecurityUtil.createSystemUser(systemUserName, org));
       Set<String> elementIds = new HashSet<String>();
-      // Remove media package from the search index
       String mpId = mp.getIdentifier().toString();
       logger.info("Removing LIVE media package {} from the search index", mpId);
 
-      jobs.add(searchService.delete(mpId));
-      // Retract elements
       for (MediaPackageElement mpe : mp.getElements()) {
         if (!MediaPackageElement.Type.Publication.equals(mpe.getElementType())) {
           elementIds.add(mpe.getIdentifier());
         }
       }
-      jobs.add(downloadDistributionService.retract(CHANNEL_ID, mp, elementIds));
 
-      if (!waitForStatus(jobs.toArray(new Job[jobs.size()])).isSuccess()) {
-        throw new LiveScheduleException("Removing live media package from search did not complete successfully");
+      List<String> failedJobs = new ArrayList<>();
+      // Remove media package from the search index
+      Job searchDeleteJob = searchService.delete(mpId);
+      if (!waitForStatus(searchDeleteJob).isSuccess()) {
+        failedJobs.add("Search Index");
+      }
+
+      // Removing media from the download distribution service
+      Job distributionRetractJob =  downloadDistributionService.retract(CHANNEL_ID, mp, elementIds);
+      if (!waitForStatus(distributionRetractJob).isSuccess()) {
+        failedJobs.add("Distribution");
+      }
+
+      if (!failedJobs.isEmpty()) {
+        throw new LiveScheduleException(
+            String.format("Removing live media package %s from %s failed", mpId, String.join(" and ", failedJobs)));
       }
     } catch (LiveScheduleException e) {
       throw e;
     } catch (Exception e) {
       throw new LiveScheduleException(e);
+    } finally {
+      securityService.setUser(prevUser);
     }
   }
 
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
@@ -164,9 +164,11 @@ public interface SearchServiceDatabase {
    *           if exception occurs
    * @throws NotFoundException
    *           if media package with specified id is not found
+   * @throws UnauthorizedException
+   *           if the current user is not authorized to perform this action
    */
   void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws SearchServiceDatabaseException,
-          NotFoundException;
+          NotFoundException, UnauthorizedException;
 
   /**
    * Store (or update) media package.
@@ -185,4 +187,5 @@ public interface SearchServiceDatabase {
   void storeMediaPackage(MediaPackage mediaPackage, AccessControlList acl, Date now)
           throws SearchServiceDatabaseException, UnauthorizedException;
 
+  boolean exist(String mediaPackageId) throws SearchServiceDatabaseException;
 }
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
@@ -187,5 +187,12 @@ public interface SearchServiceDatabase {
   void storeMediaPackage(MediaPackage mediaPackage, AccessControlList acl, Date now)
           throws SearchServiceDatabaseException, UnauthorizedException;
 
-  boolean exist(String mediaPackageId) throws SearchServiceDatabaseException;
+  /**
+   * Checks if a mediapackage is available.
+   *
+   * @param mediaPackageId The ID of the {@link MediaPackage} for which the availability is to be checked
+   * @return True if two conditions are met: The mediapackage exists in the database and the deletion date is not set
+   * @throws SearchServiceDatabaseException if an exception occurs
+   */
+  boolean isAvailable(String mediaPackageId) throws SearchServiceDatabaseException;
 }
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
@@ -479,7 +479,12 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
     }
   }
 
-  public boolean exist(String mediaPackageId) throws SearchServiceDatabaseException {
+  /**
+   * {@inheritDoc}
+   *
+   * @see org.opencastproject.search.impl.persistence.SearchServiceDatabase#isAvailable(String)
+   */
+  public boolean isAvailable(String mediaPackageId) throws SearchServiceDatabaseException {
     try {
       return db.execTxChecked(em -> {
         Optional<SearchEntity> searchEntity = getSearchEntityQuery(mediaPackageId).apply(em);
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
@@ -157,7 +157,7 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
    */
   @Override
   public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws SearchServiceDatabaseException,
-          NotFoundException {
+          NotFoundException, UnauthorizedException {
     try {
       db.execTxChecked(em -> {
         Optional<SearchEntity> searchEntity = getSearchEntityQuery(mediaPackageId).apply(em);
@@ -185,7 +185,7 @@ public class SearchServiceDatabaseImpl implements SearchServiceDatabase {
         searchEntity.get().setModificationDate(deletionDate);
         em.merge(searchEntity.get());
       });
-    } catch (NotFoundException e) {
+    } catch (NotFoundException | UnauthorizedException e) {
       throw e;
     } catch (Exception e) {
       logger.error("Could not delete episode {}: {}", mediaPackageId, e.getMessage());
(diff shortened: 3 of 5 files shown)


modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -450,29 +450,43 @@ void publishToSearch(MediaPackage mp) throws LiveScheduleException {
   }
 
   void retract(MediaPackage mp) throws LiveScheduleException {
+    Organization org = securityService.getOrganization();
+    User prevUser = org != null ? securityService.getUser() : null;
     try {
-      List<Job> jobs = new ArrayList<Job>();
+      securityService.setUser(SecurityUtil.createSystemUser(systemUserName, org));
```
marwyg, line 456: There is another place in the liveSchedulerServiceImpl where this approach is used

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -450,29 +450,43 @@ void publishToSearch(MediaPackage mp) throws LiveScheduleException {
...
+        failedJobs.add("Search Index");
+      }
+
+      // Removing media from the download distribution service
+      Job distributionRetractJob =  downloadDistributionService.retract(CHANNEL_ID, mp, elementIds);
+      if (!waitForStatus(distributionRetractJob).isSuccess()) {
+        failedJobs.add("Distribution");
+      }
+
+      if (!failedJobs.isEmpty()) {
```
marwyg, line 480: While debugging, I had some strange behavior. I am not 100% certain, but I think it could be some race condition when executing both jobs in parallel. So I made the jobs executing in series (they are not taking long anyway).

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabase.java
```diff
@@ -164,9 +164,11 @@ Collection<AccessControlList> getAccessControlLists(String seriesId, String ...
    *           if exception occurs
    * @throws NotFoundException
    *           if media package with specified id is not found
+   * @throws UnauthorizedException
+   *           if the current user is not authorized to perform this action
    */
   void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws SearchServiceDatabaseException,
-          NotFoundException;
+          NotFoundException, UnauthorizedException;
```
marwyg, line 171: I added some (potentially) missing exceptions. Other methods in this class are using these exceptions, and it seems they were simply overlooked.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -477,6 +479,18 @@ public Date getDeletionDate(String mediaPackageId) throws NotFoundException, Sea
     }
   }
 
+  public boolean exist(String mediaPackageId) throws SearchServiceDatabaseException {
```
marwyg, line 482: I added this method to be able to check if a mediapackage exists. The "getDeletionDate()" method does another permission check and cannot be used for this purpose.

modules/search-service-impl/src/test/java/org/opencastproject/search/impl/persistence/SearchServicePersistenceTest.java
```diff
@@ -162,7 +162,15 @@ public void testRetrieving() throws Exception {
 
     Date deletionDate = new Date();
     searchDatabase.deleteMediaPackage(mediaPackage.getIdentifier().toString(), deletionDate);
-    episode = searchDatabase.getMediaPackage(mediaPackage.getIdentifier().toString());
```
marwyg, line 165: This now throws a NotFoundException. Interestingly, the episode variable wasn't used here.

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
```diff
@@ -450,29 +450,43 @@ void publishToSearch(MediaPackage mp) throws LiveScheduleException {
...
+        throw new LiveScheduleException(
+            String.format("Removing live media package %s from %s failed", mpId, String.join(" and ", failedJobs)));
```
marwyg, line 482: While debugging, I had some strange behavior. I am not 100% certain, but I think it could be some race condition when executing both jobs in parallel. So I made the jobs executing in series (they are not taking long anyway).

modules/live-schedule-impl/src/main/java/org/opencastproject/liveschedule/impl/LiveScheduleServiceImpl.java
marwyg, line 456: There is another place in the liveSchedulerServiceImpl where this approach is used. So I used it, too.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -477,6 +479,18 @@ public Date getDeletionDate(String mediaPackageId) throws NotFoundException, Sea
...
+    try {
+      return db.execTxChecked(em -> {
+        Optional<SearchEntity> searchEntity = getSearchEntityQuery(mediaPackageId).apply(em);
+        return searchEntity.stream().anyMatch(entity -> entity.getDeletionDate() == null);
+      });
+    } catch (Exception e) {
+      logger.error("Error while checking if mediapackage {} exists in database: {}", mediaPackageId, e.getMessage());
+      throw new SearchServiceDatabaseException(e);
+    }
+  }
```
marwyg: I added this method to be able to check if a mediapackage exists. The "getDeletionDate()" method does another permission check and cannot be used for this purpose.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
gregorydlogan: I'm being pedantic here, but this should be `exists`.  I would probably also rename it completely, although I'm not sure to what.  The mediapackage *does* exist, it's just deleted.  But naming this `notDeleted` isn't very usable either.  Could we flip the logic internally, and call it `deleted` or something?

gregorydlogan (commented):
Minor nits, but otherwise this makes sense to me.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
marwyg: @gregorydlogan Thank you for the review! :)
Would `isAvailable()` be better? I will put Javadoc above it in any case (forgot that).

gregorydlogan: Yeah actually, that's way better!

marwyg: Done

marwyg:
I changed the method name and added javadoc. Is this PR good to go now?

gregorydlogan:
@marwyg out of curiosity, were you testing this with SFA turned on?  I couldn't get events to schedule, with errors like this

```
2025-01-15T14:58:40,251 | INFO  | (StaticResourceServlet:116) - Added static file authorization for [^/([^/]+)/engage-player/([^/]+)/.*$]
2025-01-15T14:58:40,251 | INFO  | (StaticResourceServlet:169) - Serving static files from '/home/greg/opencast/upstream/build/opencast-dist-develop-16-SNAPSHOT/data/opencast/downloads'
2025-01-15T14:58:40,259 | WARN  | (WorkspaceImpl:484) - Received unexpected response status 403 while trying to download from http://localhost/static/mh_default_org/engage-live/590cd696-497a-4174-9131-43c9ffe3d9ce/security-policy-episode/xacml.xml
2025-01-15T14:58:40,273 | WARN  | (WorkspaceImpl:484) - Received unexpected response status 403 while trying to download from http://localhost/static/mh_default_org/engage-live/590cd696-497a-4174-9131-43c9ffe3d9ce/44874e87-3cc3-4f97-9e22-b3f59e8eac51/dublincore.xml
2025-01-15T14:58:40,274 | ERROR | (DublinCoreUtil:79) - Unable to load metadata from catalog 'http://localhost/static/mh_default_org/engage-live/590cd696-497a-4174-9131-43c9ffe3d9ce/44874e87-3cc3-4f97-9e22-b3f59e8eac51/dublincore.xml'
org.opencastproject.util.NotFoundException: null
        at org.opencastproject.workspace.impl.WorkspaceImpl.downloadIfNecessary(WorkspaceImpl.java:535) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl$2.xapply(WorkspaceImpl.java:557) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl$2.xapply(WorkspaceImpl.java:554) ~[?:?]
        at org.opencastproject.util.data.Function$X.apply(Function.java:104) ~[!/:?]
        at org.opencastproject.util.IoSupport.locked(IoSupport.java:478) ~[!/:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.get(WorkspaceImpl.java:402) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.read(WorkspaceImpl.java:437) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.loadDublinCore(DublinCoreUtil.java:76) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.lambda$loadEpisodeDublinCore$0(DublinCoreUtil.java:64) ~[?:?]
        at java.util.Optional.map(Optional.java:265) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.loadEpisodeDublinCore(DublinCoreUtil.java:64) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.indexMediaPackage(SearchServiceIndex.java:256) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.indexMediaPackage(SearchServiceIndex.java:248) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.addSynchronously(SearchServiceIndex.java:243) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceImpl.lambda$process$0(SearchServiceImpl.java:263) ~[?:?]
        at org.opencastproject.security.util.SecurityUtil.runAs(SecurityUtil.java:79) [!/:?]
        at org.opencastproject.search.impl.SearchServiceImpl.process(SearchServiceImpl.java:261) [!/:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314) [!/:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:273) [!/:?]
```

which was easily enough solved by adding the missing `engage-live` slug to `SearchServiceImpl`. cf

```
diff --git a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
index d423262554..0545626d56 100644
--- a/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
+++ b/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceImpl.java
@@ -131,7 +131,8 @@ public final class SearchServiceImpl extends AbstractJobProducer implements Sear
 
   private final LoadingCache<Tuple<User, String>, Boolean> cache;
 
-  private static final Pattern staticFilePattern = Pattern.compile("^/([^/]+)/engage-player/([^/]+)/.*$");
+  private static final Pattern staticFilePattern =
+      Pattern.compile("^/([^/]+)/(?:engage-player|engage-live)/([^/]+)/.*$");
 
   /**
    * Creates a new instance of the search service.
```

marwyg:
I think I did not use SFA, but I am not sure, I build and tested Opencast locally (and on a testing cluster). I found this config in the `StaticResourceServlet.cfg`:

```
# Shall the static file server check authorization?
# You might want to disable this if you use other means of checking authentication like authentication tokens.
# Default: true
authentication.required = false
```

That's the config for turning it on, I guess? It's on `false` in my config.

gregorydlogan:
That's a non-standard default, and explains why it worked for you and not for me.  Your config turns it off, the default (as listed in the config) is commented out and thus `true`.  Will file my change as a separate PR.  This one's good to go.

lkiesow:
@marwyg, @gregorydlogan, this seems to cause merge conflicts between `r/16.x` and `r/17.x`. Since you already worked with this, can one of you take a look and to the 16 to 17 branch merge?

marwyg:
I am on it :)

PR #6387, opened 2024-12-17
Add configuration option for jetty cache eviction

ciegler:
Adding the option to set the session cache `evictionPolicy` to "evict on session exit" in the existing jetty configuration file: etc/jetty-opencast.xml 
This fixes the jetty exceptions "opencast already in cache" and "Create session failed" that show up when locally stopping and restarting Opencast and then interacting with the Admin UI. Apparently these exceptions are caused by the default setting of the jetty cache `evictionPolicy`, which is `NEVER_EVICT` (-1). By setting the `evictionPolicy` in the `AbstractSessionCacheFactory` to `EVICT_ON_SESSION_EXIT` (0) the old session will be removed from the cache when exiting the session.
When stopping and starting Opencast with this configuration a new session is started and the Admin UI will automatically load the login page when Opencast started. 

To load the opencast jetty configuration uncomment line 45 in etc/org.ops4j.pax.web.cfg (org.ops4j.pax.web.config.file=${karaf.etc}/jetty-opencast.xml).

### DIFF
--- a/etc/jetty-opencast.xml
+++ b/etc/jetty-opencast.xml
@@ -20,7 +20,6 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_0.dtd">
 
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
-
   <!-- =========================================================== -->
   <!-- Set connectors -->
   <!-- =========================================================== -->
@@ -98,6 +97,19 @@
             </Item>
           </Array>
         </Set>
+      </New>
+    </Arg>
+  </Call>
+
+  <!-- =========================================================== -->
+  <!-- Configure Cache Eviction -->
+  <!-- by default the jetty session cache is never evicted (default value: -1) -->
+  <!-- setting the evictionPolicy to 0 results in cache eviction after session exit -->
+  <!-- =========================================================== -->
+  <Call name="addBean">
+    <Arg>
+      <New class="org.eclipse.jetty.server.session.AbstractSessionCacheFactory">
+        <Set name="evictionPolicy">0</Set>
       </New>
     </Arg>
   </Call>


lkiesow:
Tried running Opencast with this patched:
```diff
diff --git a/etc/jetty-opencast.xml b/etc/jetty-opencast.xml
index f6d7075111..bf3405344f 100644
--- a/etc/jetty-opencast.xml
+++ b/etc/jetty-opencast.xml
@@ -20,7 +20,6 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_0.dtd">
 
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
-
   <!-- =========================================================== -->
   <!-- Set connectors -->
   <!-- =========================================================== -->
@@ -102,4 +101,17 @@
     </Arg>
   </Call>
 
+  <!-- =========================================================== -->
+  <!-- Configure Cache Eviction -->
+  <!-- by default the jetty session cache is never evicted (default value: -1) -->
+  <!-- setting the evictionPolicy to 0 results in cache eviction after session exit -->
+  <!-- =========================================================== -->
+  <Call name="addBean">
+    <Arg>
+      <New class="org.eclipse.jetty.server.session.AbstractSessionCacheFactory">
+        <Set name="evictionPolicy">0</Set>
+      </New>
+    </Arg>
+  </Call>
+
 </Configure>
diff --git a/etc/org.ops4j.pax.web.cfg b/etc/org.ops4j.pax.web.cfg
index 6fad50a3ea..7d529ea462 100644
--- a/etc/org.ops4j.pax.web.cfg
+++ b/etc/org.ops4j.pax.web.cfg
@@ -42,7 +42,7 @@ org.ops4j.pax.web.ssl.key.password=password
 
 # Custom jetty config file. Uncomment this if necessary to configure the jetty http connector idleTimeout value (MH-12329)
 # Adjust the settings in the file below to match the listening address and port settings above.
-#org.ops4j.pax.web.config.file=${karaf.etc}/jetty-opencast.xml
+org.ops4j.pax.web.config.file=${karaf.etc}/jetty-opencast.xml
 
 # Session cookie HTTPS flag
 # Set to true if Opencast handling HTTPS or behind a proxy handling HTTPS, default false
```

I still see the same error. But what's new is that I get an additional error that the configuration is invalid:

```
❯ head build/opencast-dist-develop-18-SNAPSHOT/data/log/opencast.log
2024-12-17T15:40:22,676 | WARN  | (JettyAwareLogger:625) - Config error at <Call name="addBean"><Arg>
      <New class="org.eclipse.jetty.server.session.AbstractSessionCacheFactory"><Set name="evictionPolicy">0</Set></New>
    </Arg></Call>
2024-12-17T15:40:22,685 | WARN  | (JettyServerWrapper:2648) - Problem parsing file:///home/lars/dev/opencast/build/opencast-dist-develop-18-SNAPSHOT/etc/jetty-opencast.xml: No suitable constructor: <New class="org.eclipse.jetty.server.session.AbstractSessionCacheFactory"><Set name="evictionPolicy">0</Set></New> on Server@7221b447{STOPPED}[9.4.54.v20240208]
java.lang.IllegalStateException: No suitable constructor: <New class="org.eclipse.jetty.server.session.AbstractSessionCacheFactory"><Set name="evictionPolicy">0</Set></New> on Server@7221b447{STOPPED}[9.4.54.v20240208]
        at org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration.newObj(XmlConfiguration.java:1039) ~[!/:9.4.54.v20240208]
        at org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration.itemValue(XmlConfiguration.java:1564) ~[!/:9.4.54.v20240208]
        at org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration.value(XmlConfiguration.java:1465) ~[!/:9.4.54.v20240208]
        at org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration.access$700(XmlConfiguration.java:419) ~[!/:9.4.54.v20240208]
        at org.eclipse.jetty.xml.XmlConfiguration$JettyXmlConfiguration$Args.<init>(XmlConfiguration.java:1723) ~[!/:9.4.54.v20240208]
```

lkiesow:
Also, looking [at the Jetty docs for evictionPolicy](https://jetty.org/docs/jetty/12/operations-guide/session/index.html):

> *jetty.session.evictionPolicy*
>    Integer, default -1. This controls whether session objects that are held in memory are subject to eviction from the cache. Eviction means that the session is removed from the cache. This can reduce the memory footprint of the cache and can be useful if you have a lot of sessions. Eviction is usually used in conjunction with a SessionDataStore that persists sessions. The eviction strategies and their corresponding values are:
> […]
> - 0 (EVICT AFTER USE)
>        sessions are evicted from the cache as soon as the last active request for it finishes. The session will be passed to the SessionDataStore to be written out before eviction.

This would mean that we don't cache the session (for most requests) any longer, woudn't it. It would need to be re-evaluuated every time except for parallel reqeuests (that would be my interpretation at least). That certainly fixes the errors, but also sounds like there might be performance implications? Not sure if that's relevant for us. Did you look into that?

ciegler:
My original assumption that Jetty Error Logs occur due to the configuration of the SessionCache is wrong. You are right that a change in the Eviction Policy could lead to to problems in other, non-local contexts . Therefore I will just close this PR.

mtneug:
@ciegler did you find a fix for the exceptions?

ciegler:
Hey, sorry for the late answer. I still need to test my idea for fixing the exceptions but at least I now know why they occur.
Apparently we manage the session maintenance with cookies (the JSESSIONID). In `modules/kernel/src/main/java/org/opencastproject/kernel/security/AuthenticationSuccessHandler.java` we call `HttpServletRequest.getSession()` which does not only tries to get the a session with a specific session id but also attempts to create a new session with this session id if it has not found a session in the (i think) SessionCache. When using cookies to maintain session integrity this leads to an IllegalStateException ([see documentation](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html#getSession(boolean))). We are (as far as I can see) still using Jetty 9.4 in which the session maintenance strategies changed. There is actually a Jetty function to get the session cookie, but I still need to figure out how to call it via the HttpServletRequest since we do not call any Jetty functions directly.

PR #1354, opened 2020-01-13
Addition of trim segment configuration and new documentation

NgoniChoga:
This adds [the feature defined in this issue](https://github.com/opencast/opencast/issues/1351) of no trim segments at the start and at the end of the video

Issue #1351, opened 2020-01-13
Buffers at the start and at the and of videos


NgoniChoga:
**The Bug**
When trimming a video, there are no buffers/trim segments at the start and at the end of the video. Adding buffers would help resolve the issue of out of sync audio and video.

The video from a network camera is encoded e.g. H264, which means when data is captured, it could be between key frames and therefore video and audio could be out of sync when it gets processed by ffmpeg.

**Environment Information**
Opencast 8.x

PR #6421, opened 2025-01-10
Added default padding and default duration to editor service with cfg

npmali:
### Describe the bug
In version 8.x, a video and audio synchronization issue https://github.com/opencast/opencast/issues/1351 was identified when playing videos after trimming. The issue stemmed from the lack of default buffers at the beginning and end of the video. This caused desynchronization because network camera videos, often encoded in formats like H264, can capture data between keyframes. When processed by FFmpeg, this lack of buffers resulted in audio and video being out of sync.

This issue was addressed by introducing a default buffer of 3 seconds at both the beginning and end of a video. However, this fix was not carried forward into the newer versions of Opencast which introduced a new editor, resulting in a regression of the sync issue.

This fix reinstates the previously implemented solution for issue https://github.com/opencast/opencast/issues/1351, and resolves the regression in newer versions.

**Related references:**

- Original issue: [#1351](https://github.com/opencast/opencast/issues/1351)
- Fix in 8.x : [PR #1354](https://github.com/opencast/opencast/pull/1354)

**Environment Information**
Identified and tested in Opencast 16.x

### To reproduce
- Open the editor in a version later than 13.x.
- Observe that there is no default buffer of 3 seconds at the beginning and end of the video.
- Delete segments from the video.
- Save the edited video and play it back.
- Observe the video and audio are not synchronized.

### How to test this patch
- Open the editor with the fix applied in a version later than 13.x.
- Verify that a default 3-second buffer is automatically added at the beginning and end of the video.
- Delete segments from the video.
- Save the edited video.
- Play back the video to confirm that the video and audio tracks are synchronized correctly.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/etc/org.opencastproject.editor.EditorServiceImpl.cfg
+++ b/etc/org.opencastproject.editor.EditorServiceImpl.cfg
@@ -94,3 +94,15 @@
 # This Property sets the time (in seconds) after which the editor should refresh the lock on a mediapackage.
 # Default: 60
 #lock.refresh.after.seconds=60
+
+# Default values for  trim segments in the editor.
+# Default: In milliseconds
+#
+# This property sets the length of the default trim segment at the beginning and end of each video
+# Default: 3000
+default.padding.length = 3000
+
+# This property sets the minimum duration of the video to add default trim segments to. If the video is shorter than 
+# this duration, no default trimm segment will be added to it.
+# Default: 6000
+min.video.duration = 6000
--- a/etc/org.opencastproject.editor.EditorServiceImpl.cfg
+++ b/etc/org.opencastproject.editor.EditorServiceImpl.cfg
@@ -102,7 +102,7 @@
 # Default: 3000
 default.padding.length = 3000
 
-# This property sets the minimum duration of the video to add default trim segments to. If the video is shorter than 
+# This property sets the minimum duration of the video to add default trim segments to. If the video is shorter than
 # this duration, no default trimm segment will be added to it.
 # Default: 6000
 min.video.duration = 6000
--- a/etc/org.opencastproject.editor.EditorServiceImpl.cfg
+++ b/etc/org.opencastproject.editor.EditorServiceImpl.cfg
@@ -100,9 +100,9 @@
 #
 # This property sets the length of the default trim segment at the beginning and end of each video
 # Default: 3000
-default.padding.length = 3000
+#default.padding.length = 3000
 
 # This property sets the minimum duration of the video to add default trim segments to. If the video is shorter than
 # this duration, no default trimm segment will be added to it.
 # Default: 6000
-min.video.duration = 6000
+#min.video.duration = 6000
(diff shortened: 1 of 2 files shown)


etc/org.opencastproject.editor.EditorServiceImpl.cfg
```diff
@@ -94,3 +94,15 @@
...
+# Default: In milliseconds
+#
+# This property sets the length of the default trim segment at the beginning and end of each video
+# Default: 3000
+default.padding.length = 3000
+
+# This property sets the minimum duration of the video to add default trim segments to. If the video is shorter than
+# this duration, no default trimm segment will be added to it.
+# Default: 6000
+min.video.duration = 6000
```
mtneug: The default configuration should not add trim segments. Also please comment out default configuration keys.

mtneug (changes requested):

mtneug:
Can I ask how you capture the network stream (e.g. what exact `ffmpeg` command and streaming protocol)? I don't really understand why trimming would solve A/V sync issues.

smarquard:
@mtneug The tracks used here are captured by Galicaster with separate audio and video tracks. When tracks are edited with a start time before the first full key frame of the video sources, the A/V sync can be substantially out (a second or more). This PR ports forward changes that were previously in Opencast in the old editor (but more client-side in the old editor, whereas this is more server-side).

modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
```diff
@@ -192,6 +194,8 @@ public class EditorServiceImpl implements EditorService {
   private static final List<MediaPackageElementFlavor> DEFAULT_THUMBNAIL_PRIORITY_FLAVOR = new ArrayList<>();
   private static final int DEFAULT_LOCK_TIMEOUT_SECONDS = 300; // ( 5 mins )
   private static final int DEFAULT_LOCK_REFRESH_SECONDS = 60;  // ( 1 min )
+  private static final String DEFAULT_PADDING_LENGTH = "0"; // (0 secs)
```
gregorydlogan: This value does match the default in the config file

gregorydlogan (changes requested):
Aside from the identified issues, this works as expected.

gregorydlogan:
@mtneug I've pushed the relevant changes here.  Since you've already been over this can you take a quick peek to make sure I haven't fireballed things and then merge this?

lkiesow:
> @mtneug I've pushed the relevant changes here. Since you've already been over this can you take a quick peek to make sure I haven't fireballed things and then merge this?

I don't think you did. I fully agree with @mtneug here:

> The default configuration should not add trim segments. 

I don't mind having the option to automatically cut away some vide material, but this should not be the default. This is a workaround for a bug or a broken video file and more a quick-fix than a good solution since it is essentially destroying video material uploaded by users without telling them. Most of the time it will admittedly not matter, but sometimes it will. And we really shouldn't make that the default.

smarquard:
@npmali Please update to address comments - specifically the default behaviour should not change, so the default for initial trim length should be 0.

PR #6434, opened 2025-01-21
Handle Opencast starting up before index is available

### LINKED ISSUE #5943: Opencast breaks if booted before Elastic-/OpenSearch

gregorydlogan:
This PR supercedes #5944, handling additional cases which caused the initial PR to hang or otherwise misbehave.

Fixes #5943

### How to test this patch

Make sure Opensearch is not running, then start Opencast an observe the waiting loop.  Play with the timing.  Then start Opensearch.

You may see exceptions related to circular dependencies.  These happen without that patch, and are their own separate issue.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests
* [x] ~explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch~

### DIFF
--- a/docs/guides/admin/docs/configuration/searchindex/elasticsearch.md
+++ b/docs/guides/admin/docs/configuration/searchindex/elasticsearch.md
@@ -33,6 +33,7 @@ Additionally, the following settings can be configured in
 * `max.retry.attempts.update`
 * `retry.waiting.period.get`
 * `retry.waiting.period.update`
+* `retry.delay.on.startup`
 
 The identifier defines which index opencast is looking for. This might be interesting if you run an
 Elasticsearch cluster and want to follow a naming scheme. But you should be aware that the index actually consists of
@@ -44,6 +45,10 @@ are too many concurrent requests. The retry behavior can be configured different
 This way you could set more retry attempts for index updates because of the more serious consequences if those requests
 fail. The waiting period is used to not overwhelm the Elasticsearch with retry requests, making the problem worse. By
 default, no retry will be attempted.
+
+The `retry.delay.on.startup` defines how long Opencast will wait between retry attempts 
+when the connection to the index service fails on startup. The default is 10 seconds.
+
 
 Version
 -------
--- a/docs/guides/admin/docs/configuration/searchindex/elasticsearch.md
+++ b/docs/guides/admin/docs/configuration/searchindex/elasticsearch.md
@@ -47,7 +47,7 @@ fail. The waiting period is used to not overwhelm the Elasticsearch with retry r
 default, no retry will be attempted.
 
 The `retry.delay.on.startup` defines how long Opencast will wait between retry attempts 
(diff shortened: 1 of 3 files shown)


KatrinIhler:
Soo, technically this is a bug fix, so I'd say this can go into 17. Of course I have a vested interest in seeing this go into 16 as well because then we don't have to work around this issue for those systems. @lkiesow Do you have an opinion on this?

lkiesow:
I don't care

marwyg (approved):
The changes in this PR were already approved in PR #5944, with one change request: to use the original connection method instead of the one I had used.

This PR addresses that change request by switching out the connection method. I tested it in my local setup with OpenSearch, and everything worked as expected. Additionally, I tested changing the "retry delay" to a small value (10ms), and it worked fine.

The code changes also look good to me.

Thank you @gregorydlogan! This PR looks good to me and can be merged, as far as I am concerned.

marwyg:
I also tested this with Branch 16.x right now. It works as expected.

marwyg:
> The changes in this PR were already approved in PR #5944, with one change request: to use the original connection method instead of the one I had used.
> 
> This PR addresses that change request by switching out the connection method. I tested it in my local setup with OpenSearch, and everything worked as expected. Additionally, I tested changing the "retry delay" to a small value (10ms), and it worked fine.
> 
> The code changes also look good to me.
> 
> Thank you @gregorydlogan! This PR looks good to me and can be merged, as far as I am concerned.

Ah okay.. I mixed the PRs up. The problem was not the connection method. It was that if starting ES/OS via container, the module wouldn't start properly. I tested this with a container and got the same error, I will investigate this.

marwyg:
I created a PR against your branch @gregorydlogan. I added another catch that deals with the case, when opensearch is started in a container.

gregorydlogan:
As discussed in the dev meeting, this would be nice to have in 16.x.  I will rebase this to 16.x and then we can merge it.

gregorydlogan:
@KatrinIhler Done, you wanna merge this since both Martin and I have touched it?

KatrinIhler:
Changed "index service" to "index" to avoid confusion with Opencast's internal index service. Won't nitpick at the code bc of time, overall execution looks good.

PR #6461, opened 2025-02-03
Allow sorting events by UID

Arnei:
Adds the option to sort results for an event search query by unique identifier. Elasticsearch does not guarantee the ordering of results if no sorting is specified, thus this option may be useful if guaranteed ordering is required, but the schema by which is sorted does not matter.

My motivation here is to use this in the admin ui frontend. This will not break the admin ui frontend.

### How to test this patch

Install this patch in your Opencast, then test this by making requests against the `admin-ng/events.json` endpoint using `uid:asc` or `uid:desc`. You can use the rest docs for this https://stable.opencast.org/docs.html?path=/admin-ng/event#getevents-18.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/AbstractEventEndpoint.java
@@ -2668,6 +2668,9 @@ public abstract class AbstractEventEndpoint {
       Set<SortCriterion> sortCriteria = RestUtils.parseSortQueryParameter(optSort.get());
       for (SortCriterion criterion : sortCriteria) {
         switch (criterion.getFieldName()) {
+          case EventIndexSchema.UID:
+            query.sortByUID(criterion.getOrder());
+            break;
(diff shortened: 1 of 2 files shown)


gregorydlogan:
This doesn't seem to work:
`ocreq 'http://localhost/admin-ng/event/events.json?filter=uid:asc' | jq '.results[].id'` gets me the same results as `ocreq 'http://localhost/admin-ng/event/events.json?filter=uid:desc' | jq '.results[].id'`.  I would expect the order here to flip?

gregorydlogan (changes requested):
This doesn't seem to work:
ocreq 'http://localhost/admin-ng/event/events.json?filter=uid:asc' | jq '.results[].id' gets me the same results as ocreq 'http://localhost/admin-ng/event/events.json?filter=uid:desc' | jq '.results[].id'. I would expect the order here to flip?

Arnei:
You want `sort` instead of `filter` as the query param.

gregorydlogan:
Whoops :sweat_smile: 

Works fine when you use it right!

PR #6475, opened 2025-02-06
Fix the multiple creation of subtitle files

marwyg:
Fixes https://github.com/opencast/opencast-editor/issues/1550

Fixes a Bug, where multiple subtitle files get created.

### How to reproduce
- Activate the subitle editor in the editor-settings.toml
- Create a subtitle file by clicking on `+`
- Click on the "Finish" tab and click "Save"
- Click on the "Finish" tab and click "Save"
- Click on the "Finish" tab and click "Save"
- Click on the "Finish" tab and click "Save"
- Click on the "Finish" tab and click "Save"
- ...
- Reload with F5
- You have created multiple Subtitle Files

### How to test this patch
- Do the steps described above
- There should be only 1 subtitle File at the end
- To be sure, test also the following things:
  - Test the editing of subtitle files
  - Test the deletion of subtitle files
  - ... Upload
  - ... Download
  - ... the creation of multiple files (different Languages)

### DIFF
--- a/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
+++ b/modules/editor-service/src/main/java/org/opencastproject/editor/EditorServiceImpl.java
@@ -547,7 +547,7 @@ public class EditorServiceImpl implements EditorService {
           throws IOException, IllegalArgumentException {
     for (EditingData.Subtitle subtitle : subtitles) {
       // Generate ID for new tracks
-      String subtitleId = UUID.randomUUID().toString();
+      String subtitleId = subtitle.getId();
       String trackId = null;
 
       // Check if subtitle already exists


geichelberger:
This is a frontend issue that needs to be addressed: the subtitles should have their state reset to prevent resubmission. Additionally, for new subtitles, the editor should refrain from assigning an ID, as we shouldn't depend on user input to determine our entity IDs.

geichelberger:
We discussed this in the dev meeting and concluded that the IDs should be created from the backend.

gregorydlogan:
So perhaps a dumb question here, but is that what this does? :)

KatrinIhler:
What's the status here? Does anybody remember?

gregorydlogan:
Testing, per the PR description says this works for me.  Given that 17 is about to be EOL do we want to fix this?  Will be doing a 17.x pass at the next meeting.

geichelberger:
Perhaps I misunderstood the code, but does that mean that when creating a new subtitle track, user input can define our internal unique IDs?

KatrinIhler:
Talked to Martin, he will repoint to 18 when he has the time.

gregorydlogan:
Closing this since 17 is now done.  Please refile when ready.

PR #6491, opened 2025-02-17
Wait for OpenSearch cluster state yellow or green

geichelberger:
This update changes the health check to accept a yellow status. A yellow state can occur during normal operations or due to issues like node failure, and it shouldn't prevent Opencast from starting.

### How to test this patch

Without the patch Opencast will remain on hold in step 6, patiently waiting for the cluster state to transition to green before proceeding, which will not happen.

1. Start the Elasticsearch Developer Container
2. Launch Opencast
3. Once the indexing is complete, go ahead and stop the Opencast application.
4. Set Replica Count
   - Update the replica count by executing the following command:
   ```bash
   curl -XPUT -H "Content-Type: application/json" 'http://localhost:9200/_all/_settings' -d '{"index.number_of_replicas" : "1"}'
   ```
5. Check for Cluster Health
   - Verify that the cluster state is yellow by running:
   ```bash
   curl 'http://localhost:9200/_cluster/health'
   ```
6. Now, start the Opencast again to confirm that it functions correctly with the updated settings.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
+++ b/modules/elasticsearch-impl/src/main/java/org/opencastproject/elasticsearch/impl/AbstractElasticsearchIndex.java
@@ -472,10 +472,13 @@ public abstract class AbstractElasticsearchIndex implements SearchIndex {
     try {
       // test connection
       ClusterHealthRequest request = new ClusterHealthRequest();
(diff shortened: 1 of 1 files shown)


Arnei (approved):
Works and makes sense to me. With how most of our users will be using OpenSearch, a yellow status is unlikely to be caused by anything that would negatively impact Opencast.

PR #6498, opened 2025-02-23
Optimize search

geichelberger:
This PR enhances full-text search by:  
- Switching to a **"search as you type"** field with an **edge n-gram analyzer** for better partial term matching.  
- Adding a dedicated **search analyzer** to improve query relevance.  
- Replacing `must` with `filter` for **keyword fields**, improving performance by avoiding unnecessary scoring calculations.  
- Removing the unused `text_fuzzy` field, as fuzziness is already handled by Elasticsearch/OpenSearch.  

These changes improve search accuracy, efficiency, and overall performance.

### How to test this patch

Compare search queries with your current version and the version from the PR. Give feedback!

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/upgrade/17_to_18/event_mapping.json
+++ b/docs/upgrade/17_to_18/event_mapping.json
@@ -1,0 +1,257 @@
+{
+  "_source" : { "enabled" : true },
+  "dynamic": true,
+  "properties" : {
+
+    "uid": { "type" : "keyword", "store" : true },
+
+    "organization": { "type" : "keyword", "store" : true },
+
+    "object": { "type" : "text", "index" : false, "store" : true },
+
+    "title": {
+      "type" : "keyword",
+      "fields" : {
+        "sort": {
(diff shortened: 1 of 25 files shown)


Arnei (approved):
Changes sounds reasonable to me, code reads reasonable to me and using the search field in the admin ui now makes much more sense to me (and seems a bit faster). Also does not appear to break anything else, so I'll approve this. Ideally we get more people to test this, particularly by users who actually rely on the search functionality in their work.

gregorydlogan:
As discussed in the technical meeting today, this requires/would be better with a reindex. 
 @geichelberger is going to explore how to just munge the data we need rather than doing a full reindex.  Worst case we do require a reindex, but we're going to try and avoid that here (and going forward).

gregorydlogan:
This looks good to me, and seems to work from my limited testing.

PR #6500, opened 2025-02-25
Unify constants for episode ID role access configuration

JulianKniephoff:
#5056 introdued the feature to grant access to events by assigning the user a “magic role” based on the ID of the event.
It centralized the configuration to `etc/custom.properties` because it is needed in many different places.
However, every site of use got its own constant to refer to that property.

This patch replaces those with a central constant that lives where some of the others already live.


Arnei (approved):
Sure why not.

gregorydlogan (approved):

lkiesow (approved):
Looks good.
Can be merged unless someone brings up an issue at the next technical meeting.

PR #6540, opened 2025-03-14
Changing Logger Level for Jetty 

ciegler:
Stopping and restarting Opencast, while having it open in the browser, leads to the following Jetty Warnings: `Session <sessionID> already in cache`  and  `Create session failed` .  This warning is triggered in the `DelegatingAuthenticationEntryPoint.java`  when the server tries to create a new session (via `HttpServletRequest.getSession()`) with the sessionID it got from the Cookie.

A similar issue was already reported to Jetty (e.g https://github.com/jetty/jetty.project/issues/4888), but it seems that at least within the Jetty version (9.4.57)  we are currently using for Opencast this warning is still occurring. (Reporting this issue again is not necessary, since jetty 9 is EOL since Jan 25)

As far as I understood, the session is maintained via cookies. After seeing in the opencast.log that after restarting the server the old session can be entered and the expiration timer is reset for it, I assume that the `Create session failed` warning does not have any negative consequences for the session maintenance. 

When stopping and restarting Opencast, while having it open in the browser, the user can see that network errors are occurring, at `System warnings and notifications`  on the menu bar. After a while the user is automatically forwarded to the login page, but after logging in the user is sometimes forwarded to `health.json` or `STATS.json . To avoid this I edited the user forwarding in the `DelegatingAuthenticationEntryPoint.java`.

### DIFF
--- a/etc/org.ops4j.pax.logging.cfg
+++ b/etc/org.ops4j.pax.logging.cfg
@@ -37,6 +37,8 @@ log4j2.logger.felix.name = org.apache.felix
 log4j2.logger.felix.level = WARN
 log4j2.logger.cxf.name = org.apache.cxf
 log4j2.logger.cxf.level = WARN
+log4j2.logger.jetty.name = org.eclipse.jetty
+log4j2.logger.jetty.level = ERROR
 
 # Appenders configuration
 
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
@@ -63,7 +63,7 @@ public class DelegatingAuthenticationEntryPoint implements AuthenticationEntryPo
       // if the user attempted to access a url other than /, store this in the session so we can forward the user there
       // after a successful login
       String requestUri = null;
-      if (!request.getRequestURI().contains(".json")) {
+      if (!request.getRequestURI().contains("health") && !request.getRequestURI().contains("stats")) {
         requestUri = request.getRequestURI();
       }
       String queryString = request.getQueryString();
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
@@ -62,9 +62,12 @@ public class DelegatingAuthenticationEntryPoint implements AuthenticationEntryPo
     } else {
       // if the user attempted to access a url other than /, store this in the session so we can forward the user there
       // after a successful login
-      String requestUri = request.getRequestURI();
+      String requestUri = null;
+      if (!request.getRequestURI().contains(".json")) {
+        requestUri = request.getRequestURI();
+      }
       String queryString = request.getQueryString();
-      if (requestUri != null && !requestUri.isEmpty() & !"/".equals(requestUri)) {
+      if (requestUri != null && !requestUri.isEmpty() && !"/".equals(requestUri)) {
         if (queryString == null) {
           request.getSession().setAttribute(INITIAL_REQUEST_PATH, requestUri);
         } else {


etc/org.ops4j.pax.logging.cfg
```diff
@@ -37,6 +37,8 @@ log4j2.logger.felix.name = org.apache.felix
 log4j2.logger.felix.level = WARN
 log4j2.logger.cxf.name = org.apache.cxf
 log4j2.logger.cxf.level = WARN
+log4j2.logger.jetty.name = org.eclipse.jetty
+log4j2.logger.jetty.level = ERROR
```
mtneug, line 41: So the log messages are still there but filtered out. Is this correct?

We run two admin nodes and see this log message when the load balancer decides to move traffic to another node. We have not seen any problems, but we also use an SSO solution and Opencast will just recreate sessions if the SSO login is still valid.

modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
```diff
@@ -62,9 +62,12 @@ public void commence(HttpServletRequest request, HttpServletResponse response, A
     } else {
       // if the user attempted to access a url other than /, store this in the session so we can forward the user there
       // after a successful login
-      String requestUri = request.getRequestURI();
+      String requestUri = null;
+      if (!request.getRequestURI().contains("health") && !request.getRequestURI().contains("stats")) {
```
mtneug, line 66: I'm really not a fan of this special handling. First, just checking with `contains` could be a problem if we add new endpoints in the future where it would make sense to redirect. Second, what about every other endpoint? Just because the Admin UI requests those two path in regular intervals doesn't mean that by clicking in the UI, other endpoints will not be requested during a short offline period.

mtneug (changes requested):

etc/org.ops4j.pax.logging.cfg
ciegler: Yes, this will just filter out the WARN log messages from Jetty. I tried to circumvent the warning before filtering it out, but since this is due to how Jetty handles the `getSession()` request and we have limited options to configure Jetty itself (due to it being loaded automatically via karaf and paxweb), I could not consistently prevent these warning logs. 

Regarding the SSO (oc-remember-me cookie) : when the server is shutdown and restarted an existing `oc-remember-me` cookie will be detected. However (at least when testing this locally) the cookie signature is no longer matching the expected cookie from the restarted server. So the `oc-remember-me` cookie is canceled and the user is required to sign in again.

modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
ciegler: I can understand that. Another idea that I had was to check whether the URI is `<something>.json` in general and in this case not redirect the user to it.

etc/org.ops4j.pax.logging.cfg
mtneug: Whether sessions are remembered after a restart depends highly on your setup. You should be able to configure Opencast to accept remember-me cookies after a restart (see [mh_default_org.xml config](https://github.com/opencast/opencast/blob/c427c65999fb383253de7bae010c1870fb0f72b4/etc/security/mh_default_org.xml#L434-L450)). We don't use remember-me cookies since auth is handled by the SSO system outside of Opencast. So users will remain logged in and Jetty will log this warning.

mtneug: I'm not sure if there are other warnings Jetty will log that are now also hidden. Otherwise, maybe error messages are enough?

modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
lkiesow: I'm fine with the current change since it does improve the situation. @mtneug, can you live with this change?

If not, then I suggest removing the change here and moving it and the discussion to another pull request. The change addresses a separate issue, has nothing to do with “Changing Logger Level for Jetty” and should have been a separate patch in the first place. That would also mean we could go ahead and merge the actual part of this pull request (according to the title at least).

etc/org.ops4j.pax.logging.cfg
```diff
@@ -38,6 +38,10 @@ log4j2.logger.felix.level = WARN
 log4j2.logger.cxf.name = org.apache.cxf
 log4j2.logger.cxf.level = WARN
 
+# uncomment the following lines if you want to set Jetty log level to ERROR
+# log4j2.logger.jetty.name = org.eclipse.jetty
+# log4j2.logger.jetty.level = ERROR
+
```
lkiesow: I disagree with the changes from commit 7efff437c5c5d34fb2b7c7647a5a8b119d3d64ae. First, it's a commit modifying completely unrelated code. The configuration has nothing to do with the changes to the Java code.

But I really disagree with commenting this out. I don't understand why you want to do that? We already change log level of some components since we think it's a good idea as the warnings users would see otherwise confuse adopters and don't help at all. I think it's the same case here and therefor, adjusting the log level is a good idea and we should commit to that. I skimmed through the Jetty code and I don't think we hide anything adopters would really need to see.

Otherwise, this is a configuration file already. So adopters can already hide the Jetty logs. If you think setting the configuration is a bad idea, cluttering the configuration files with everything you could set isn't helpful at all.

So for me, I think we should commit to the change ot don't modify the configuration file. And I'm heavily in favor of the former. I already adopted this change for our production cluster.
Cluttering configuration files with

lkiesow (changes requested):

modules/kernel/src/main/java/org/opencastproject/kernel/security/DelegatingAuthenticationEntryPoint.java
mtneug: Let's get the current change in 👍

etc/org.ops4j.pax.logging.cfg
mtneug: This was commented out probably because I wondered if setting the log level to error would suppress important warnings. But I agree with @lkiesow and would also suggest including this by default. We also rolled out the change in production.

ciegler: I removed the 7efff43 commit, such that the PR only contains the changes from 1a17c0a, which were agreed upon above.

PR #6564, opened 2025-03-21
Set sender value in sent emails if set

### LINKED ISSUE #6554: Missing From-Header in Emails sent by SmtpService Class
Since one month, I have been receiving emails, without a from header from our Opencast server. After consulting with our mail server administrators, they confirmed that Opencast is sending emails without the correct From header.

This issue is often concealed by many MTAs that automatically "correct" such errors. However, according to RFC 822 and its later versions, email messages **must** contain a From field.

It seems that the SmtpService class ([source](https://github.com/opencast/opencast/blob/16.7/modules/kernel/src/main/java/org/opencastproject/kernel/mail/SmtpService.java)) reads the from email address from org.opencastproject.kernel.mail.SmtpService.cfg but does not explicitly set the From header. As a result, the email payload/message is missing this value.

gregorydlogan:
As reported in #6554, we're not setting the sender field in sent emails.  This is actually required per RFC 822.

Fixes #6554

### How to test this patch

Send some emails from @stefanosgeo's systems.  I can't reproduce, since gmail doesn't seem to care :D

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests
* [i] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)
* [x] ~explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch~

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/mail/BaseSmtpService.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/mail/BaseSmtpService.java
@@ -151,6 +151,10 @@ public class BaseSmtpService {
     this.sender = sender;
   }
 
+  public String getSender() {
+    return this.sender;
+  }
+
   public void setSsl(boolean ssl) {
     this.ssl = ssl;
   }
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/mail/SmtpService.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/mail/SmtpService.java
@@ -304,6 +304,9 @@ public class SmtpService extends BaseSmtpService implements ManagedService {
     }
 
     MimeMessage message = createMessage();
+    if (StringUtils.isNotBlank(getSender())) {
+      message.addFrom(new InternetAddress[] { new InternetAddress(getSender()) });
+    }
     addRecipients(message, RecipientType.TO, to);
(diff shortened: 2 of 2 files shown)


stefanosgeo:
Well, yes, I suppose so! At the very least, I can give it a shot. I haven't tested anything yet, and I'm not too familiar with it, but I'll definitely give it a try this week and report back

stefanosgeo (commented):
I I tested it as follows:

1. I built the pr locally, skipping Maven tests (since my local FFmpeg installation lacks some required codecs).
2. I transferred the  allinone-tar archive to my test-vm, extracted it, and started Opencast with:
```
sudo -u opencast /opt/opencast/bin/start-opencast
 ```
3. I received an email with the expected From header and a simple "Hello world" body:

```
Date: Thu, 27 Mar 2025 08:03:38 +0100 (CET)
From: rrze-opencast@*****.de
To: rrze-opencast@*****.de
Message-ID: <364652932.1.1743059018023@****.de>
Subject: Test from Opencast
```

4. I haven’t tested the send-mail WOH yet, but I assume it uses the same MIME object and should also include the From header.

I believe we can merge it.Would it be okay to file the PR against the 16.x branch?

gregorydlogan:
I'm going to close in this in favour of #6581 since this is based off my `r/17.x` tip and I can't change that.  Sigh.

PR #6566, opened 2025-03-22
Add FLOSS/Fund file

gregorydlogan:
This is part of something at organized by Apereo, attempting to get funding from FLOSS/Fund via https://floss.fund.  It's unclear to me if the URL in the file is the final one, correct one will be provided by EOD Thursday with merge being done by EOD Friday.


gregorydlogan:
Url appears to be the final one. Merging.

PR #6575, opened 2025-03-27
Add asset manager endpoint to trigger event index update

lkiesow:
This patch adds a REST endpoint to allow admins to trigger index updates for single events. While this is usually not necessary (hence limiting this to admins), this can be really helpful when fixing broken events and avoids weird workarounds.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
@@ -51,6 +51,7 @@ import org.opencastproject.rest.AbstractJobProducerEndpoint;
 import org.opencastproject.security.api.UnauthorizedException;
 import org.opencastproject.util.Checksum;
 import org.opencastproject.util.ChecksumType;
+import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.data.Option;
 import org.opencastproject.util.doc.rest.RestParameter;
 import org.opencastproject.util.doc.rest.RestParameter.Type;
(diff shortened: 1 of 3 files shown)


Arnei (approved):
Works and makes sense to me to have this as a handy tool for admins who know what they are doing.

That the endpoint lives under /assets instead of /index is highly unintuitive to me, but I would not consider that a blocker.

lkiesow:
It is because it's specific to the updates sent from the asset manager.
Katrin and I already talked about the idea of moving something like this to the index endpoint and trigger all parts of an index for a given event in succession. But that's something for later and outside the scope of this patch.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
```diff
@@ -165,6 +166,43 @@ public Response snapshot(@FormParam("mediapackage") final MediaPackageImpl media
...
+              description = "Update successfully triggered.",
+              responseCode = SC_NO_CONTENT),
+          @RestResponse(
+              description = "Not allowed to trigger update.",
+              responseCode = SC_FORBIDDEN),
+          @RestResponse(
+              description = "No such event found.",
+              responseCode = SC_NOT_FOUND),
+          @RestResponse(
+              description = "There has been an internal error and no snapshot could be taken.",
```
gregorydlogan: No snapshot is being taken in any case :)

gregorydlogan (changes requested):

rute-santos:
What if there's a workflow running on that event?

lkiesow:
> What if there's a workflow running on that event?

Then the workflow keeps running? I don't see how this effects or is effected by workflows in any way? Can you elaborate what your specific concern is?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/endpoint/AbstractAssetManagerRestEndpoint.java
lkiesow: While this is still technically correct :grin:… you are absolutely right that this doesn't make sense. I've updated the RESR docs.

PR #6619, opened 2025-04-02
Add tobira auth callback

geichelberger:
This introduces a new callback endpoint (`/tobira/callback/{token}`) that allows Tobira to retrieve user details based on the REMOTE_USER header.

### How to test this patch
Tobira test config:

```toml
[auth]
source = "callback:http://localhost:8080/tobira/callback/test123"
trusted_external_key = "tobira"
pre_auth_external_links = true
login_link = "/Shibboleth.sso/Login"
logout_link = "/Shibboleth.sso/Logout"

[auth.callback]
relevant_headers = ["REMOTE_USER"]
cache_duration="0"

```

Opencast test config org.opencastproject.tobira.impl.TobiraEndpoint.cfg:
```cfg
callbackToken=test123
allowedRolesPattern=ROLE_GR.*
```
Use the tobira graphql explorer http://127.0.0.1:3080/~graphiql

Query:

```graphql
query MyQuery {
  currentUser {
    username
    userRole
    roles
    email
    displayName
    canUseStudio
    canUseEditor
    canUpload
    canFindUnlisted
    canCreateUserRealm
  }
}
```
Headers:
```json
{
  "REMOTE_USER": "admin"
}
```

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
@@ -31,8 +31,12 @@ import org.opencastproject.db.DBSessionFactory;
 import org.opencastproject.playlists.PlaylistService;
 import org.opencastproject.search.api.SearchService;
 import org.opencastproject.security.api.AuthorizationService;
+import org.opencastproject.security.api.Role;
 import org.opencastproject.security.api.SecurityService;
+import org.opencastproject.security.api.User;
+import org.opencastproject.security.api.UserDirectoryService;
 import org.opencastproject.series.api.SeriesService;
+import org.opencastproject.userdirectory.UserIdRoleProvider;
 import org.opencastproject.util.Jsons;
 import org.opencastproject.util.doc.rest.RestParameter;
 import org.opencastproject.util.doc.rest.RestQuery;
@@ -50,6 +54,7 @@ import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 import org.osgi.service.jaxrs.whiteboard.propertytypes.JaxrsResource;
+import org.osgi.service.metatype.annotations.Designate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,6 +63,10 @@ import java.io.InputStream;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import javax.persistence.EntityManagerFactory;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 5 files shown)


modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
```diff
@@ -225,6 +250,69 @@ public Response harvest(
...
+      },
+      returnDescription = "Returns user information"
+  )
+  public Response callback(@PathParam("token") String token, @Context HttpHeaders headers) {
+
+    if (callbackToken == null || !callbackToken.equals(token)) {
+      return badRequest("Invalid token or callback disabled");
+    }
+
+    List<String> username = headers.getRequestHeader("REMOTE_USER");
```
gregorydlogan: After much faffing about, I *finally* convinced my nginx reverse proxy to pass this header.  This required adding `underscores_in_headers on;` in the server blocks.

From my googling, underscores are either not allowed, or at the very least very much frowned upon in header names.  Can we make this REMOTE-USER, or something else?

gregorydlogan (changes requested):

modules/tobira/src/main/java/org/opencastproject/tobira/impl/TobiraEndpoint.java
mtneug: `X-Remote-User` seems to be commonly used, but there is also `Remote-User` or `X-Webauth-User`. I would suggest `X-Remote-User`.

geichelberger: Maybe I should make the header configurable.

gregorydlogan: Considering this is in the tobira endpoint space I don't know if thats worth it.

If you want to move this to a more general endpoint space then sure.

geichelberger: If you use Shibboleth SP with Apache httpd, the REMOTE_USER header would work out of the box.

gregorydlogan: ... and lots of our adopters don't use Shib, much less Apache ;)

I don't care what we call it, as long as it's not going to require an adopter to dig around trying to figure out why the header isn't working.  If making it configurable makes more sense to you then go ahead :)

mtneug: > If you use Shibboleth SP with Apache httpd, the REMOTE_USER header would work out of the box.

@geichelberger are you sure this is a header? [The Shibboleth documentation](https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065335257/AttributeAccess#REMOTE_USER) explicitly states that this is not a header but an env variable from the CGI days.

geichelberger: If you set `ShibUseHeaders On`, the header `REMOTE_USER` is also added, and it's usually a CGI variable. 

These are the headers added by `ShibUseHeaders` for our setup:
```
Shib-Cookie-Name: 
Shib-Session-ID: 
Shib-Session-Index: 
Shib-Session-Expires: 
Shib-Session-Inactivity: 
Shib-Identity-Provider: 
Shib-Authentication-Method: 
Shib-Authentication-Instant: 
Shib-AuthnContext-Class: 
Shib-AuthnContext-Decl: 
Shib-Assertion-Count: 
Shib-Handler: 
subject-id: 
pairwise-id: 
eppn: 
affiliation: 
entitlement: 
oid: 
mn: 
sn: 
givenName: 
displayName: 
uid: 
mail: 
persistent-id: 
Shib-Application-ID: 
REMOTE_USER: 
```

geichelberger:
I added a configuration key with the default header name `X-WEBAUTH-USER`.

PR #6621, opened 2025-04-03
Properly handle Tobira harvest of events with streaming-only tracks

gregorydlogan:
As reported on Matrix, in streaming-only publications (ie, only `engage-streaming` video tracks) Tobira's harvester throws an exception like this:

```
2025-04-03T11:10:19,768 | ERROR | (Harvest:168) - Error reading event '2c38219d-4caa-45cc-8f92-0fc93b6f8e73' (skipping...)
java.lang.IllegalArgumentException: Cannot decode to DCMIPeriod: PT2M
        at org.opencastproject.metadata.dublincore.EncodingSchemeUtils.decodeMandatoryPeriod(EncodingSchemeUtils.java:359) ~[?:?]
        at org.opencastproject.tobira.impl.Item.lambda$new$10(Item.java:149) ~[?:?]
        at java.util.OptionalLong.orElseGet(OptionalLong.java:234) ~[?:?]
        at org.opencastproject.tobira.impl.Item.<init>(Item.java:147) ~[?:?]
        at org.opencastproject.tobira.impl.Harvest.lambda$harvest$0(Harvest.java:165) ~[?:?]
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
[...]
```

The root cause here is that `EncodingSchemeUtils`'s `decodeMandatoryPeriod` doesn't speak ISO8601 Period, despite sounding an *awful lot* like it should.  Instead it throws an exception.

So we're switching to `decodeMandatoryDuration` instead, which works as expected.

This should be in 16.x since it's probably not *that* rare for adopters to have streaming-only publications, and we should probably support that properly for Tobira as well.

### How to test this patch

I tested by removing the `Arrays.stream` chunk and just setting the duration to the `orElseGet` chunk, but I suppose you could also publish a streaming-only event too :)

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
+++ b/modules/tobira/src/main/java/org/opencastproject/tobira/impl/Item.java
@@ -30,7 +30,6 @@ import org.opencastproject.mediapackage.MediaPackageElement;
 import org.opencastproject.mediapackage.Track;
 import org.opencastproject.mediapackage.TrackSupport;
 import org.opencastproject.mediapackage.VideoStream;
-import org.opencastproject.metadata.dublincore.DCMIPeriod;
 import org.opencastproject.metadata.dublincore.DublinCore;
 import org.opencastproject.metadata.dublincore.DublinCoreCatalog;
 import org.opencastproject.metadata.dublincore.DublinCoreUtil;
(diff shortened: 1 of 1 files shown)


gregorydlogan:
Testing jar for Maxime, built against 16.10.  Just rename it from .zip to .jar, 'cause GitHub won't let me attach jars directly.
[opencast-tobira-16.10.zip](https://github.com/user-attachments/files/19591356/opencast-tobira-16.10.zip)

mpmuc84:
Thanks for the patch!

Added this to our test cluster, and it appears to fix the issue. Processing a streaming-only publication now goes through without exceptions, and Tobira can harvest the event as extected.

Issue #6565, opened 2025-03-21
Setting a user property more than once causes DB errors


gregorydlogan:
**Describe the bug**

Testing https://github.com/opencast/opencast-admin-interface/pull/825, enable terms of use and then log in with a not-`ROLE_ADMIN` user.  You will be presented with the ToU popup.  Accept it, proceed with life.

Now log in again.  You'll see the ToU modal again.  If you accept it again you get this fun stack trace:

```
Caused by: org.eclipse.persistence.exceptions.DatabaseException: 
Internal Exception: java.sql.SQLIntegrityConstraintViolationException: (conn=3) Duplicate entry 'test-mh_default_org' for key 'UNQ_oc_user_settings_0'
Error Code: 1062
Call: INSERT INTO oc_user_settings (id, setting_key, organization, username, setting_value) VALUES (?, ?, ?, ?, ?)
        bind => [5 parameters bound]
Query: InsertObjectQuery(org.opencastproject.adminui.usersettings.persistence.UserSettingDto@16aa3bc0)
        at org.eclipse.persistence.exceptions.DatabaseException.sqlException(DatabaseException.java:333) ~[?:?]
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeDirectNoSelect(DatabaseAccessor.java:917) ~[?:?]
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeNoSelect(DatabaseAccessor.java:979) ~[?:?]
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.basicExecuteCall(DatabaseAccessor.java:640) ~[?:?]
        at org.eclipse.persistence.internal.databaseaccess.DatabaseAccessor.executeCall(DatabaseAccessor.java:567) ~[?:?]
        at org.eclipse.persistence.internal.sessions.AbstractSession.basicExecuteCall(AbstractSession.java:2105) ~[?:?]
        at org.eclipse.persistence.sessions.server.ClientSession.executeCall(ClientSession.java:313) ~[?:?]
        at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:275) ~[?:?]
        at org.eclipse.persistence.internal.queries.DatasourceCallQueryMechanism.executeCall(DatasourceCallQueryMechanism.java:261) ~[?:?]
[...]

PR #6443, opened 2025-01-31
Introduce series creator field

geichelberger:
The database was missing a field to store the creator of a series, and it was only available in the index. After rebuilding an index, the information will be lost.

This is just a draft. If the solution is feasible, I will add a migration script.

Fixes https://github.com/opencast/opencast/issues/6428

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)
* [ ] explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch

PR #6622, opened 2025-04-03
User Settings Now Working

gregorydlogan:
This PR get the user settings endpoints working properly

- **The addUserSetting method now checks to see if a key for that user already exists, and updates it if it does.  Previous behaviour led to db constraint issues since it was blindly inserting new, identical keys.  This does mean that you can no longer add duplicate keys, so if you need more than one thing stored under a given key you must marshall and unmarshall on the user's end.**
- **Previous behaviour worked correctly for the *first* key set by a user, but when more than one key was set you would see constraint violations.  Adding the key to the user+org tuple resolves the issue, and matches the logic in the service.**

Note that this PR updates the indexes on a presumably empty table (since the endpoints don't really work right at all anyway).

Fixes: #6565

### How to test this patch

Use the endpoints under `/admin-ng/user-settings` to set, unset, and update user settings.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/upgrade/17_to_18/mysql5.sql
+++ b/docs/upgrade/17_to_18/mysql5.sql
@@ -1,0 +1,3 @@
+-- Drop table so it can be recreated.  Previous functionality was unused, and broken even if it was
+drop table oc_user_settings;
+
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
@@ -40,6 +40,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Function;
 
 import javax.persistence.EntityManager;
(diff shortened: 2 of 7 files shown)


gregorydlogan:
Marking this as a draft since it needs a DB migration, and thus has to go into `develop`

docs/upgrade/17_to_18/mysql5.sql
```diff
@@ -0,0 +1,3 @@
+-- Drop table so it can be recreated.  Previous functionality was unused, and broken even if it was
+drop table oc_user_settings;
```
lkiesow, line 2: Makes sense to drop this, but could also accidentally cause problems if adopters don't run the migration exactly at the correct time (thinking about adopters upgrading over multiple versions, …).

That could be easily avoided by naming the new table differently. Maybe something like `oc_frontend_user_settings`? What do you think?

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/UserSettingsService.java
```diff
@@ -217,20 +218,38 @@ private Function<EntityManager, UserSettings> getUserSettingsQuery(int limit, in
...
+          userSettingDto.setValue(value);
+          userSettingDto.setUsername(username);
+          userSettingDto.setOrganization(orgId);
+          em.persist(userSettingDto);
+          return userSettingDto.toUserSetting();
+        });
+      } catch (Exception e) {
+        logger.error("Could not add user setting username '{}' org: '{}' key: '{}' value: '{}'", username, orgId,
+            key, value, e);
+        throw new UserSettingsServiceException(e);
```
lkiesow: Never log an error if you continue to throw the caught exception. If you want to include additional information, do that in the exception string of the new exception.

Reasoning:
- The thrown exception may be caught and handled later in which case the error is no longer an error, but expected and handled behavior.
- The exception may trickle up and be caught and logged there in which case you log the same error twice so that it looks like there are multiple erros. Especially on a busy system where other logs may show up in between.

modules/admin-ui/src/main/java/org/opencastproject/adminui/usersettings/persistence/UserSettingDto.java
```diff
@@ -41,7 +41,8 @@
 @Table(name = "oc_user_settings", indexes = {
```
lkiesow: Maybe? See above…
```suggestion
@Table(name = "oc_frontend_user_settings", indexes = {
```

lkiesow (changes requested):
Didn't really test this. Just skimmed the code. The change looks reasonable, though.

What's definitely missing is an entry in the release notes folder since this migration would need to appear in the release notes even if we decide to rename the table.

Also, one of out OC18 RMs (@jduehring, @dennis531) should approve this since technically this should have gone in prior to the branch cut. Although, I don't see a poblem in this specific case as no one could use this before.

docs/upgrade/17_to_18/mysql5.sql
gregorydlogan: I'm fine with renaming it, but do we leave this in so that the defunct table *hopefully* goes away?

dennis531 (approved):
The code looks good and API works! Thank you for this fix!

gregorydlogan:
Release notes blub added.

docs/upgrade/17_to_18/mysql5.sql
jduehring: https://github.com/opencast/opencast/pull/6443 introduced DB migration scripts for mariadb and postgres. I think we could add the drop table there to avoid having multiple different scripts right?

dennis531:
We merge this PR and create another PR for combining the db migrations.

Issue #5692, opened 2024-03-25
Harmonize permission checking


KatrinIhler:
We have a couple different places that do permission checking on the basis of ACLs additionally to the XACMLAuthorizationService, for example:

- The AssetManager (right to snapshot, static file protection)
-  The Workflow Service (right to start workflow)
- The Search Service (static file protection, right to update mp in search)
- AccessControlUtil (used by the search service for...something)
- The Series Service

(This list is the result of a superficial search and may not be entirely accurate, nor is it complete.)

This makes introducing new features like the id roles (see #5056) a bit annoying. Maybe we can harmonize this a bit.

Issue created to keep track of refactoring ideas.

Arnei:
Harmonizing the AssetManager for now does not make sense I feel.
- The permission checks in the AssetManager are interwoven with it's overengineered query architecture. In my head it makes sense to first disentangle those queries.
- The permission checks use the database, likely for performance reasons. Our AuthorizationService uses the XMLs instead. So if the AssetManager were to use the AuthService, the probably performance loss may cause problems.

PR #6635, opened 2025-04-14
Unify authorization, at least partially

Arnei:
Helps with #5692.

Attempt at making more modules use the `XACMLAuthorizationService` (our only implementation of `AuthorizationService`) for their authorizations checks, instead of using their own custom code.

This is likely missing a few modules, most notably the AssetManager which has woven permission checking into its query structure.

### How to test this patch

There are probably many ways to test this patch. Create yourself a user with restricted roles and focus in particular on series, workflows and the search service.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan:
If you want this to move forward, can you rebase it against current `develop`?

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
```diff
@@ -1150,10 +1149,9 @@ protected void assertPermission(WorkflowInstance workflow, String action, String
 
     var creatorName = workflow.getCreatorName();
     var workflowCreator = creatorName == null ? null : userDirectoryService.loadUser(creatorName);
-    boolean authorized = currentUser.hasRole(GLOBAL_ADMIN_ROLE)
-            || (currentUser.hasRole(currentOrgAdminRole) && currentOrgId.equals(workflowOrgId))
-            || (currentUser.equals(workflowCreator))
-            || (authorizationService.hasPermission(mediapackage, action) && currentOrgId.equals(workflowOrgId));
+    boolean authorized = (currentUser.equals(workflowCreator))
+            || (authorizationService.hasPermission(
+                mediapackage, currentUser, currentOrg, action, currentOrgId.equals(workflowOrgId)));
```
KatrinIhler: The authorization service also knows the current org, so it would be enough to hand over the entity org, and have the authorization service make the check if that is not null or something.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -395,20 +406,16 @@ public MediaPackage getMediaPackage(String mediaPackageId)
...
-                  && !AccessControlUtil.isAuthorized(acl, currentUser, currentOrg, CONTRIBUTE.toString(),
-              mediaPackageId)
-                  && !AccessControlUtil.isAuthorized(acl, currentUser, currentOrg, WRITE.toString(), mediaPackageId)) {
-            throw new UnauthorizedException(currentUser + " is not authorized to see episode " + mediaPackageId);
-          }
+        MediaPackage mediaPackage = MediaPackageParser.getFromXml(episodeEntity.get().getMediaPackageXML());
+        User currentUser = securityService.getUser();
+        // There are several reasons a user may need to load a episode: to read content, to edit it, or add content
+        if (!authorizationService.hasPermission(mediaPackage, READ.toString())
+                && !authorizationService.hasPermission(mediaPackage, CONTRIBUTE.toString())
```
KatrinIhler, line 416: TIL we have an action called contribute. :D We should get rid of that one day.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
```diff
@@ -169,12 +183,10 @@ public void deleteMediaPackage(String mediaPackageId, Date deletionDate) throws
...
         Organization currentOrg = securityService.getOrganization();
         MediaPackage searchMp = MediaPackageParser.getFromXml(searchEntity.get().getMediaPackageXML());
-        String accessControlXml = searchEntity.get().getAccessControl();
 
         // allow ca users to retract live publications without putting them into the ACL
-        if (!(searchMp.isLive() && currentUser.hasRole(GLOBAL_CAPTURE_AGENT_ROLE)) && accessControlXml != null) {
-          AccessControlList acl = AccessControlParser.parseAcl(accessControlXml);
-          if (!AccessControlUtil.isAuthorized(acl, currentUser, currentOrg, WRITE.toString(), mediaPackageId)) {
+        if (!(searchMp.isLive() && currentUser.hasRole(GLOBAL_CAPTURE_AGENT_ROLE))) {
+          if (!authorizationService.hasPermission(searchMp, currentUser, currentOrg, WRITE.toString())) {
```
KatrinIhler: For this service, you switch the permission check over from looking at the ACL directly to handing over the media package. This impacts performance a bit, because we then read the ACL from disk instead of using the one that's in the database that we already have. There's also a Github Actions failure (workflow fails during publish-engage) where I wonder if that's related. Can you take another look there?

(And I know that the permission check on the ACL directly doesn't consider the episode ID roles bc we don't know the media package id but can't we hand that over?)

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
```diff
@@ -383,8 +382,7 @@ private void checkSearchEntityWritePermission(final String mediaPackageId) throw
         throw new NotFoundException();
       }
       AccessControlList acl = persistence.getAccessControlList(mediaPackageId);
-      if (!AccessControlUtil.isAuthorized(acl, user, securityService.getOrganization(), WRITE.toString(),
-          mediaPackageId)) {
+      if (!authorizationService.hasPermission(acl, user, securityService.getOrganization(), WRITE.toString())) {
```
KatrinIhler: This probably breaks the episode id roles for this check.

KatrinIhler (changes requested):
Code looks good in general, but there are some concerns left. :)

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/persistence/SearchServiceDatabaseImpl.java
Arnei: Thanks for pointing that out, I was not aware that that would screw with performance. I added another `hasPermission` variants for acls that also takes a mediapackageId specifically for the episode ID roles check.

modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java
Arnei: Should not be broken anymore.

modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java
Arnei: Sure, can do that. Should be changed now.

Arnei:
Not sure why the test workflow keeps failing. Can't reproduce that locally and I don't see a link between the timeouts and the PR changes.

gregorydlogan:
Hrm...  I can't reproduce this locally either, but the workflows are *failing*.  This, to me, suggests that one of the components supporting a WOH is broken.  Looking at the GHA logs, it's installing the system's `ffmpeg`, not `ffmpeg-dist` or anything from our repo.  Could that be the culprit?

gregorydlogan:
Had a dive into this.  There CI tests are reproducable locally, though I'm not exactly sure why.  The workflow is failing because the publish WOH is timing out.  Looking at the stack, I get
```
2026-08-07T15:08:49,761 | INFO  | (PublishEngageWorkflowOperationHandler:431) - Publishing media package test to search index
2026-08-07T15:13:50,924 | ERROR | (DublinCoreUtil:78) - Unable to load metadata from catalog 'http://localhost:8081/static/mh_default_org/enga
ge-player/test/0f217309-db7f-42c7-9526-739abea1b293/dublincore.xml'
org.opencastproject.util.NotFoundException: Could not copy http://localhost:8081/static/mh_default_org/engage-player/test/0f217309-db7f-42c7-9
526-739abea1b293/dublincore.xml to /home/greg/opencast/upstream/build/opencast-dist-admin/data/opencast/workspace/http_localhost_8081/static/m
h_default_org/engage-player/test/0f217309-db7f-42c7-9526-739abea1b293/dublincore-18ebd8cb-0db5-46ff-b7c4-f6f97ca4d03d.xml
        at org.opencastproject.workspace.impl.WorkspaceImpl.downloadIfNecessary(WorkspaceImpl.java:525) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.lambda$downloadIfNecessary$2(WorkspaceImpl.java:537) ~[?:?]
        at org.opencastproject.util.IoSupport.locked(IoSupport.java:418) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.get(WorkspaceImpl.java:387) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.read(WorkspaceImpl.java:422) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.loadDublinCore(DublinCoreUtil.java:75) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.lambda$loadEpisodeDublinCore$0(DublinCoreUtil.java:63) ~[?:?]
        at java.util.Optional.map(Optional.java:260) ~[?:?]
        at org.opencastproject.metadata.dublincore.DublinCoreUtil.loadEpisodeDublinCore(DublinCoreUtil.java:63) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.indexMediaPackage(SearchServiceIndex.java:285) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.indexMediaPackage(SearchServiceIndex.java:275) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceIndex.addSynchronously(SearchServiceIndex.java:254) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceImpl.lambda$process$0(SearchServiceImpl.java:264) ~[?:?]
        at org.opencastproject.security.util.SecurityUtil.runAs(SecurityUtil.java:85) ~[?:?]
        at org.opencastproject.search.impl.SearchServiceImpl.process(SearchServiceImpl.java:262) ~[?:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314) ~[?:?]
        at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:272) ~[?:?]
        at java.util.concurrent.FutureTask.run(FutureTask.java:317) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
        at java.lang.Thread.run(Thread.java:1583) [?:?]
Caused by: org.opencastproject.security.api.TrustedHttpClientException: java.net.SocketTimeoutException: Read timed out
        at org.opencastproject.kernel.security.TrustedHttpClientImpl.execute(TrustedHttpClientImpl.java:423) ~[?:?]
        at org.opencastproject.kernel.security.TrustedHttpClientImpl.execute(TrustedHttpClientImpl.java:370) ~[?:?]
        at org.opencastproject.workspace.impl.WorkspaceImpl.downloadIfNecessary(WorkspaceImpl.java:503) ~[?:?]
        ... 20 more
Caused by: java.net.SocketTimeoutException: Read timed out
```

Which is *weird*, because `curl`-ing to localhost:8081 works fine.  I *suspect* here, that the CI tests use of an otherwise unpriviledged user (`ROLE_STUDIO` only) is partially to blame here, but I'm out of time to keep digging.

Arnei:
I also looked at this a couple weeks ago and came to the conclusion that the authorization is not working as it should, meaning I failed. I could not pinpoint exactly what I did wrong though.

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -308,44 +310,101 @@ private Optional<AccessControlList> loadAcl(final URI uri) {
...
-      for (Role role : user.getRoles()) {
-        if (entry.getRole().equals(role.getName())) {
-          // immediately abort on matching deny rules
-          // (never allow if a deny rule matches, even if another allow rule matches)
-          if (!entry.isAllow()) {
-            logger.debug("Access explicitly denied for role({}), action({})", role.getName(), action);
-            return false;
+      if (entry.getAction().equals(action)) {
+        for (Role role : user.getRoles()) {
+          if (entry.getRole().equals(role.getName())) {
```
gregorydlogan, line 390: Leaving aside the rest of this PR, deny rules are not a thing in Opencast and will Soon(tm) go away via https://github.com/opencast/opencast/issues/7718#issuecomment-4782561978

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -308,44 +310,101 @@ private Optional<AccessControlList> loadAcl(final URI uri) {
...
+              return false;
+            }
+            allowed = true;
           }
-          allowed = true;
         }
       }
     }
-    logger.debug("XACML file allowed access");
+    logger.debug("ACL allowed access");
```
gregorydlogan, line 402: Since we're touching this line, can we add the MP id, and user?  Or just remove this line completely?

gregorydlogan (changes requested):
Ok, so this was bothering me, so I stepped through *far* too much.  At least one hang during publication is caused by [`loadDublinCore`](https://github.com/opencast/opencast/blob/d7926c91b8f5adc99a87c57579685f5e12d66bc0/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java#L284) when adding things in SearchServiceIndex.  That method, makes its way to [here](https://github.com/opencast/opencast/blob/2260c6aa7db465bf68af860ab9a55c616e8f511b/modules/dublincore/src/main/java/org/opencastproject/metadata/dublincore/DublinCoreUtil.java#L72), where we see a call to `workspace.read(is)`.  In the test case, that call is what hangs.  Looking at the code I suspect what's happening is that `TrustClientImpl` is *not* switching users when it makes the HTTP request to download the DC file, and then gets a 403 and fails to handle it properly.

PR #6640, opened 2025-04-17
Don't index themes anymore

### LINKED ISSUE #5264: Don't index themes
IMO, putting themes in the Elasticsearch index is total overkill, because nobody has that many themes to warrant the effort. (If you are the exception to this rule, speak up now!) Might as well just get them from the database like we do for groups, users, and ACL templates already.

(Might be useful to make a proposal for this first, though.)

Arnei:
Fixes #5264.

Removes all code that relates themes to the index. For services like the admin ui that require detailed query capabilities, a new function has been added to the themes database.

See related issue for motivation.

Of course, this patch cannot remove any actual indices from ElasticSearch/OpenSearch installations. So if you really want themes gone from your index you'll have to nuke and rebuild it, or send a query directly to your index that gets rid of them. Not doing so should not be a problem though.

### How to test this patch

This mostly affects the admin ui and the listprovider, so be sure to check that those work as before.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestSeriesEndpoint.java
```diff
@@ -229,8 +233,29 @@ public TestSeriesEndpoint() throws Exception {
     indexServiceImpl.setSecurityService(securityService);
     indexServiceImpl.setSeriesService(seriesService);
 
+    // setup theme database
+    User user = new JaxbUser("test", null, "Test User", "test@test.com", "test", new DefaultOrganization(), new HashSet<>());
+    UserDirectoryService userDirectoryService = EasyMock.createNiceMock(UserDirectoryService.class);
+    EasyMock.expect(userDirectoryService.loadUser((String) EasyMock.anyObject())).andReturn(user).anyTimes();
+    EasyMock.replay(userDirectoryService);
+
+    themesServiceDatabaseImpl = new ThemesServiceDatabaseImpl();
```
KatrinIhler, line 242: Wouldn't it be sufficient for this test to mock this as well?

modules/themes/src/test/java/org/opencastproject/themes/ThemesServiceDatabaseTest.java
```diff
@@ -121,4 +112,96 @@ public void testStoreUpdateAndDelete() throws Exception {
...
+
+    themes = themesDatabase.findThemesQuery(
+        Optional.ofNullable(Integer.MAX_VALUE),
+        Optional.ofNullable(0),
+        new ArrayList<>(),
+        Optional.empty(),
+        Optional.ofNullable("admin")
+    );
+    Assert.assertEquals(1, themes.size());
+  }
```
KatrinIhler, line 206: :heart:

modules/themes/src/main/java/org/opencastproject/themes/persistence/ThemesServiceDatabaseImpl.java
```diff
@@ -185,6 +164,99 @@ private List<Theme> getThemes() throws ThemesServiceDatabaseException {
     }
   }
 
+  public List<Theme> findThemesQuery(
```
KatrinIhler: Nitpick: Just call it findThemes(), bc adding Query implies, for me at least, that you get back a query object, not the results.

KatrinIhler (changes requested):
Looks pretty good! Love that you extended the tests as well. Just a few minor things, and I stumbled over one issues during testing that already exist in develop: When changing the theme for an existing series, the dropdown shows the ID of the theme after selection, and the series cannot be saved. I know you didn't break this, but could you take a look at that either way?

modules/admin-ui/src/test/java/org/opencastproject/adminui/endpoint/TestSeriesEndpoint.java
Arnei: Arguably yes, since in an ideal world the functionality of the themes database has already been tested by its respective module. So we could spare us the trouble of testing it here and save a couple of lines by mocking up the database instead. On the other hand, I see no harm in double testing the themes database here.

Arnei:
> Issue that already exists in develop

I'd say that is better solved in a different pull request. You're mentioning a dropdown, so might this be an admin-ui issue? I suggest opening a ticket over there https://github.com/opencast/opencast-admin-interface/issues

KatrinIhler:
> > Issue that already exists in develop
> 
> I'd say that is better solved in a different pull request. You're mentioning a dropdown, so might this be an admin-ui issue? I suggest opening a ticket over there https://github.com/opencast/opencast-admin-interface/issues

Done, see https://github.com/opencast/opencast-admin-interface/issues/1205

Would be great if you could take a look at that because it kinda gets in the way of properly testing this. :)

KatrinIhler (approved):

KatrinIhler:
Works now! Do you want to rebase this onto 18.x while there's still time?

Arnei:
What do you mean by "while there's still time"? The branch cut already happened, right? In any case, I don't think this is important enough to force into 18.

PR #6648, opened 2025-04-24
Same target tag behaviour for all WOH 

### LINKED ISSUE #5265: Harmonize the support of + and - for tags in WOHs
While the configuration of tags, both source and target, was harmonized a while ago (see https://github.com/opencast/opencast/issues/1157), there are still inconsistencies in the way these tags are then used in the WOHs. Especially the support of + and - for adding or removing a single tag in target tags or - for excluding a tag in source-tags varies greatly with multiple handlers having their own implementation for this. Instead I suggest that

1) WOHs should use a MediaPackageElementSelector instead of selecting source elements themselves (looking at you, `publication-channel-to-workspace`), this supporting the exclusion of tags with -, and

2) ConfiguredTagsAndFlavors should support special target tags with + and - and WOHs like `image-convert` should use that instead of their own implementation.

Arnei:
Fixes #5265.

All workflow operation handlers adapt the same method of handling target tags (if they support target tags).

The chosen behaviour is that of the tag WOH:
"If a target-tag starts with a '-', it will be removed from preexisting tags, if a target-tag starts with a '+', it will be added to preexisting tags. If there is no prefix, all preexisting tags are removed and replaced by the target-tags."

This means that all WOH that did not support "+" and "-" now do.
!THIS CONSTITUTES A BEHAVIOUR CHANGE FOR SOME WORKFLOW OPERATION HANDLERS!
In particular, some WOH were adding tags instead of replacing them, even though they were not prefaced with a "+". Adopters are advised to check their workflows.

Also replace some entwine stuff in the image WOH.

### How to test this patch

Create a test event and run all the workflows you have, especially if they are custom. Check if assets are tagged correctly.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan (changes requested):
Didn't get to actually reviewing the code, but I *think* the following classes need to be modified to match this behaviour:

- MultiEncodeWorkflowOperationHandler
- ProcessSmilWorkflowOperationHandler
- CutMarksToSmilWorkflowOperationHandler
- VideoSegmenterWorkflowOperationHandler

I went through everyhing extending `AbstractWorkflowOperationHandler`, think I got the rest of them...

Arnei:
Thanks for looking through that, I changed the behaviour for CutMarksToSmilWorkflowOperationHandler and VideoSegmenterWorkflowOperationHandler to fit the new behaviour.

I deliberately did not touch MultiEncodeWorkflowOperationHandler and ProcessSmilWorkflowOperationHandler, because they did non-standard stuff with their target tags, like grouping them by in blocks delimited by `;`. As they require this behaviour to function properly, it does not make sense to change them I think.

PR #6649, opened 2025-04-25
Introduce configuration "atLeastOne" for getting tags and flavors

### LINKED ISSUE #3530: `source-flavor` and `source-tags` are falsely required for some operations
**Describe the bug**

Many operations now use `.getTagsAndFlavors` to parse `source-flavor` and `source-tags`. When there should only be one flavor/tag, `Configuration.one` can be used. However, this strictly requires there to be exactly one configured flavor/tag. This is not always desired.

In my particular case, I configured `segmentpreviews` only with `source-tags` and no `source-flavor` which caused the following exception:

```
2022-03-15 11:03:26,314 | ERROR | (WorkflowOperationWorker:141) - Workflow operation 'operation:'segmentpreviews', position:132, state:'FAILED'' failed
org.opencastproject.workflow.api.WorkflowOperationException: Configuration key 'source-flavor' must be set
    at org.opencastproject.workflow.api.AbstractWorkflowOperationHandler.getTagsAndFlavors(AbstractWorkflowOperationHandler.java:428) ~[!/:?]
    at org.opencastproject.workflow.handler.composer.SegmentPreviewsWorkflowOperationHandler.createPreviews(SegmentPreviewsWorkflowOperationHandler.java:172) ~[?:?]
    at org.opencastproject.workflow.handler.composer.SegmentPreviewsWorkflowOperationHandler.start(SegmentPreviewsWorkflowOperationHandler.java:145) ~[?:?]
    at org.opencastproject.workflow.impl.WorkflowOperationWorker.start(WorkflowOperationWorker.java:193) ~[!/:?]
    at org.opencastproject.workflow.impl.WorkflowOperationWorker.execute(WorkflowOperationWorker.java:116) [!/:?]
    at org.opencastproject.workflow.impl.WorkflowServiceImpl.runWorkflowOperation(WorkflowServiceImpl.java:804) [!/:?]
    at org.opencastproject.workflow.impl.WorkflowServiceImpl.process(WorkflowServiceImpl.java:1841) [!/:?]
    at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2280) [!/:?]
    at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2246) [!/:?]
```

I'm not sure if the best way would be to change the meaning of `Configuration.one` to one or zero. Maybe add another enum entry?

**Environment Information**
 - Opencast version with minor release 11.4

Arnei:
My suggestion on how to fix #3530.

Most workflow operation handlers use `.getTagsAndFlavors` to parse tags and flavors (and if they don't they probably should). The `Configuration` arguments determine whether there should be zero, exactly one or an arbitrary amount of e.g. source flavors.

This commit introduces a new `Configuration` option called "atLeastOne", so that there for example must be at least one source flavor or more, but not zero.

This commit also introduces a new parameter list for `.getTagsAndFlavors` where the `Configuration` for both source tags and source flavors can be specified together. This means that the configuration will be applied to the sum of source tags and source flavors. I imagine this will mostly be used in combination with "atLeastOne", for WOHs that want at least one source flavor or tag specified, but don't particularly care which one.

### How to test this patch

Modify the segmentpreviews operation in the fast workflow with different combinations of source flavors and source tags. (Be sure to choose a presentation video long enough for segment-video to generate segments).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
+++ b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
@@ -193,7 +193,7 @@ public class SegmentPreviewsWorkflowOperationHandler extends AbstractWorkflowOpe
     // Read the configuration properties
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(wi,
         Configuration.atLeastOne, Configuration.many, Configuration.one);
-    MediaPackageElementFlavor sourceVideoFlavor = tagsAndFlavors.getSingleSrcFlavor();
+    List<MediaPackageElementFlavor> sourceVideoFlavors = tagsAndFlavors.getSrcFlavors();
     List<String> sourceTagSet = tagsAndFlavors.getSrcTags();
     ConfiguredTagsAndFlavors.TargetTags targetImageTags = tagsAndFlavors.getTargetTags();
     MediaPackageElementFlavor targetImageFlavor = tagsAndFlavors.getSingleTargetFlavor();
@@ -208,7 +208,9 @@ public class SegmentPreviewsWorkflowOperationHandler extends AbstractWorkflowOpe
 
     // Select the tracks based on the tags and flavors
     TrackSelector trackSelector = new TrackSelector();
-    trackSelector.addFlavor(sourceVideoFlavor);
+    for (MediaPackageElementFlavor flavor : sourceVideoFlavors) {
+      trackSelector.addFlavor(flavor);
+    }
     for (String tag : sourceTagSet) {
       trackSelector.addTag(tag);
     }
--- a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
+++ b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
@@ -225,7 +225,7 @@ public class SegmentPreviewsWorkflowOperationHandler extends AbstractWorkflowOpe
 
     if (videoTrackSet.size() == 0) {
       logger.debug("Mediapackage {} has no suitable tracks to extract images based on tags {} and flavor {}",
-              mediaPackage, sourceTagSet, sourceVideoFlavor);
+              mediaPackage, sourceTagSet, sourceVideoFlavors);
       return createResult(mediaPackage, Action.CONTINUE);
     } else {
 
--- a/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
+++ b/modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
@@ -192,7 +192,7 @@ public class SegmentPreviewsWorkflowOperationHandler extends AbstractWorkflowOpe
 
     // Read the configuration properties
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(wi,
-        Configuration.many, Configuration.one, Configuration.many, Configuration.one);
+        Configuration.atLeastOne, Configuration.many, Configuration.one);
     MediaPackageElementFlavor sourceVideoFlavor = tagsAndFlavors.getSingleSrcFlavor();
     List<String> sourceTagSet = tagsAndFlavors.getSrcTags();
     ConfiguredTagsAndFlavors.TargetTags targetImageTags = tagsAndFlavors.getTargetTags();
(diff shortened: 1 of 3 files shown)


modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/AbstractWorkflowOperationHandler.java
```diff
@@ -498,21 +530,15 @@ protected ConfiguredTagsAndFlavors getTagsAndFlavors(WorkflowInstance workflow,
...
-          targetFlavorList.add(flavor);
+      case atLeastOne:
+        targetFlavorList = getFlavors(operation, TARGET_FLAVORS, TARGET_FLAVOR);
+        if (targetTagList.isEmpty()) {
+          throw new WorkflowOperationException("Configuration key '" + TARGET_FLAVORS + "' or '" + TARGET_FLAVOR + "' must be set");
         }
         break;
+      case many:
+        targetFlavorList = getFlavors(operation, TARGET_FLAVORS, TARGET_FLAVOR);
+        break;
```
gregorydlogan, line 549:
```suggestion
        targetTagMap = parseTargetTagsByType(targetTagList);
        break;
```

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/AbstractWorkflowOperationHandler.java
```diff
@@ -498,21 +530,15 @@ protected ConfiguredTagsAndFlavors getTagsAndFlavors(WorkflowInstance workflow,
...
-            flavor = MediaPackageElementFlavor.parseFlavor(elem);
-          } catch (IllegalArgumentException e) {
-            throw new WorkflowOperationException(elem + " is not a valid flavor!");
-          }
-          targetFlavorList.add(flavor);
+      case atLeastOne:
+        targetFlavorList = getFlavors(operation, TARGET_FLAVORS, TARGET_FLAVOR);
+        if (targetTagList.isEmpty()) {
+          throw new WorkflowOperationException("Configuration key '" + TARGET_FLAVORS + "' or '" + TARGET_FLAVOR + "' must be set");
         }
```
gregorydlogan, line 545:
```suggestion
        }
        targetTagMap = parseTargetTagsByType(targetTagList);
```

gregorydlogan (changes requested):
This fixes the ~compile~test issues.  Still working on testing, but it looks like it should work.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
```diff
@@ -192,7 +192,7 @@ private WorkflowOperationResult createPreviews(final MediaPackage mediaPackage,
 
     // Read the configuration properties
     ConfiguredTagsAndFlavors tagsAndFlavors = getTagsAndFlavors(wi,
-        Configuration.many, Configuration.one, Configuration.many, Configuration.one);
+        Configuration.atLeastOne, Configuration.many, Configuration.one);
     MediaPackageElementFlavor sourceVideoFlavor = tagsAndFlavors.getSingleSrcFlavor();
```
jduehring: Line 196 still leads to an error if no `source-flavor` is specified, even though one `source-tag` should be sufficient.

I suggest to handle it like the tags and get a list of all flavors (including empty list, if no flavors exist) and adding these to the TrackSelector.

jduehring:
```suggestion
    List<MediaPackageElementFlavor> sourceVideoFlavors = tagsAndFlavors.getSrcFlavors();
```

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
jduehring:
```suggestion
    for (MediaPackageElementFlavor flavor : sourceVideoFlavors) {
      trackSelector.addFlavor(flavor);
    }
```

jduehring (commented):
I tried to test different combinations of `source-flavors` and `source-tags` as you suggested. Seemed to mostly work fine.
But only specifiying one `source-tag` still leads to an error, even though this should work if I understood the initial issue correctly. 

We should also probably check the docs and describe in the WOHs how many tags/flavors are expected/required in each operation.

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/SegmentPreviewsWorkflowOperationHandler.java
Arnei: Thanks for the review. I applied your suggestions.

modules/workflow-service-api/src/main/java/org/opencastproject/workflow/api/AbstractWorkflowOperationHandler.java
Arnei: Could you elaborate?

PR #6689, opened 2025-05-13
Remove queries from the asset manager

### LINKED ISSUE #5966: Remove AssetManager Queries
The code for the AssetManager queries is incredibly convoluted and difficult to understand because it's spread out over multiple classes that inherit from one another and wrap each other. We can probably replace that with a few named queries and be happier for it.

Arnei:
Fixes #5966.

Removes all query related code from the asset manager. This means this asset manager cannot be accessed via queries anymore. Instead, the asset manager is exposing new methods which can be used instead.

What this boils down to is dumbing down the asset manager. The query language we used was powerful but complex. In removing it, we lose the power to create whatever query we want or need on a dime. Now, if there is something you want to do with the asset manager but there is no method for it, that method will have to be written first.
On the other hand, writing new methods should now be a lot easier. Named queries are less powerful, but comparatively easy to understand. Therefore accessing and working with the asset manager should be easier now.

This also allows us to get rid of the query-dsl dependency, which hopefully makes upgrading other depdencies easier.

### How to test this patch

Do whatever you can think of that would affect assets. Mostly this will be running workflows, but creativity is encouraged.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


Arnei:
I'm failing to recreate the failing test in the workflows. I did create the test user with ROLE_STUDIO and ran the curl command, but it returns 200, which makes debugging difficult. @gregorydlogan do you maybe have any idea how I could better recreate this?

Edit: T'was probably to do with authentication. Rolled back my functional authentication changes.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
```diff
@@ -76,7 +76,161 @@
...
+                + "ORDER BY s.version DESC"
+        ),
+        @NamedQuery(
+            name = "Snapshot.findLatest",
+            query = "SELECT s FROM Snapshot s "
+                + "WHERE s.mediaPackageId = :mediaPackageId "
+                + "AND (:organizationId IS NULL OR s.organizationId = :organizationId) "
+                + "AND s.version = ( "
+                + "  SELECT MAX(s2.version) FROM Snapshot s2 WHERE s2.mediaPackageId = s.mediaPackageId "
+                + ")"
```
lkiesow, line 94: Not really that important, but maybe this is simpler:
```sql
ORDER BY s.version DESC
LIMIT 1
```

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
```diff
@@ -76,7 +76,161 @@
...
+        ),
+        @NamedQuery(
+            name = "Snapshot.findForIndexRebuild",
+            query = "SELECT s FROM Snapshot s "
+                + "WHERE s.version = ( "
+                + "  SELECT MAX(s2.version) FROM Snapshot s2 "
+                + "  WHERE s2.mediaPackageId = s.mediaPackageId "
+                + ") "
+                + "ORDER BY s.mediaPackageId DESC "
+        ),
```
lkiesow, line 162: <del>Not yet sure if that's a problem, but wouldn't this load all events at once from the database? Couldn't that lead to memory problems if there are 50000+ events?</del>

*Named query parameters in the JPA functions seem to limit this.*

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
```diff
@@ -236,43 +242,64 @@ public void setAvailability(final VersionImpl version, final String mpId, final
...
 
   public Optional<AssetDtos.Full> findAssetByChecksum(final String checksum) {
     return db.execTx(em -> {
-      final Tuple result = AssetDtos.baseJoin(em)
-          .where(QAssetDto.assetDto.checksum.eq(checksum))
-          .singleResult(Full.select);
-      var dtoOpt = Opt.nul(result).map(Full.fromTuple);
-      return dtoOpt.isSome() ? Optional.of(dtoOpt.get()) : Optional.empty();
+      return namedQuery.findOpt(
+              "Asset.findByChecksum",
```
lkiesow: This method is used only in a test which should probably use `findAssetByChecksumAndStoreAndOrg` instead. Getting rid of the test gets rid of this method and the new named query `Asset.findByChecksum`.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/AssetDto.java
```diff
@@ -23,23 +23,68 @@
...
+            + "AND a.mediaPackageElementId = :mpeId "
+            + "AND s.version = :version "
+            + "ORDER BY s.version DESC"
+    ),
+    @NamedQuery(
+        name = "Asset.findByChecksum",
+        query = "SELECT a FROM Asset a "
+            + "INNER JOIN a.snapshot s "
+            + "WHERE a.checksum = :checksum"
+    ),
```
lkiesow: This is probably unnecessary since it's used only in one test. See comment at `findAssetByChecksum`.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/AssetDto.java
```diff
@@ -23,23 +23,68 @@
...
+    @NamedQuery(
+        name = "Asset.findByChecksumStorageIdAndOrganizationId",
+        query = "SELECT a FROM Asset a "
+            + "INNER JOIN a.snapshot s "
+            + "WHERE a.checksum = :checksum "
+            + "AND a.storageId = :storageId "
+            + "AND s.organizationId = :orgId "
+    ),
```
lkiesow, line 65: It seems weird to join only to filter and discard the data and then get the snapshot later anyway. See comment in `findAssetByChecksumAndStoreAndOrg`.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
```diff
@@ -348,16 +375,273 @@ public long countEvents(final String organization) {
...
     return db.execTx(em -> {
-      final Tuple result = AssetDtos.baseJoin(em)
-          .where(QAssetDto.assetDto.checksum.eq(checksum)
-              .and(QAssetDto.assetDto.storageId.eq(storeId))
-              .and(QAssetDto.assetDto.snapshot.organizationId.eq(orgId)))
-          .singleResult(Full.select);
-      var dtoOpt = Opt.nul(result).map(Full.fromTuple);
-      return dtoOpt.isSome() ? Optional.of(dtoOpt.get()) : Optional.empty();
+      return namedQuery.findOpt(
+              "Asset.findByChecksumStorageIdAndOrganizationId",
```
lkiesow, line 379: Looking at `Asset.findByChecksumStorageIdAndOrganizationId`, it joins the assets table with the snapshot table only to filter for one snapshot property (organization identifier) and then discard the snapshot data and return just an asset. What do we do with the asset? We use it to run another query to get the snapshot data :see_no_evil: 

If we need the join, can we use all the data we get?

Given that the checksum should usually be unique (check this assumption!), we could not join, but get the list of assets with the given checksum (should usually just be one), and use the first(?) one with a snapshot that belongs to the correct organization.

All of this is a bit weird, since `checksum` is not marked unique but is used as if it was (at least in combination with the organization).

Also, is it just me, or is `AssetDtos.Full` and `AssetDtos.Medium` a weird concept? Why not just use a tuple of `Asset` and `Snapshot` but copy over half of the snapshot data to another object?

lkiesow: Actually :sweat_smile: … it seems like `AssetDtos.Full` is only ever used here in non-test scenarios. And the result of this method is only used to use snapshot data and the media package element identifier.. So, maybe we can return just that.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/AssetDtos.java
```diff
@@ -63,6 +40,14 @@ public static class Medium {
     private final String organizationId;
     private final String owner;
 
+    public Medium(AssetDto assetDto, Availability availability, String organizationId) {
```
lkiesow, line 43: The more I look at code related to this class the more I think we should just have named queries which then return a `Pair<Snapshot, Asset>` instead of building partial classes we then later convert, discard, …
Do we want to get rid of this entirely?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
```diff
@@ -236,43 +242,64 @@ public void setAvailability(final VersionImpl version, final String mpId, final
...
+            Pair.of("mpId", mpId),
+            Pair.of("version", version.value()))
+          .apply(em)
+          .map(result -> new SnapshotDtos.Medium(
+              result,
+              Availability.valueOf(result.getAvailability()),
+              result.getStorageId(),
+              result.getOrganizationId(),
+              result.getOwner()
+          ));
```
lkiesow, line 276: Why? A `SnapshotDtos.Medium` is just a `Snapshot` with some data reordered? Why don't we just use `Snapshot`?

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
```diff
@@ -76,7 +76,161 @@
         @NamedQuery(name = "Snapshot.countByMediaPackage", query = "select count(s) from Snapshot s "
                 + "where s.mediaPackageId = :mediaPackageId"),
         @NamedQuery(name = "Snapshot.countByMediaPackageAndOrg", query = "select count(s) from Snapshot s "
-                + "where s.mediaPackageId = :mediaPackageId and s.organizationId = :organizationId")})
+                + "where s.mediaPackageId = :mediaPackageId and s.organizationId = :organizationId"),
+        @NamedQuery(
+            name = "Snapshot.findMediumByMpIdAndVersion",
```
lkiesow:
```suggestion
            name = "Snapshot.findByMpIdAndVersion",
```

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
```diff
@@ -348,16 +375,273 @@ public long countEvents(final String organization) {
...
+
+  public List<Snapshot> getSnapshots(String mediaPackageId, String orgId, String orderByVersion) {
+    return db.execTx(em -> {
+      String namedQueryName = "Snapshot.findByMpId";
+      if ("ASC".equals(orderByVersion)) {
+        namedQueryName = "Snapshot.findOldestVersionFirst";
+      }
+      if ("DESC".equals(orderByVersion)) {
+        namedQueryName = "Snapshot.findLatestVersionFirst";
+      }
```
lkiesow: We never promised that queries are unsorted if you don't pass any sort parameter → you can have one named query less:

```suggestion
      // Optional: check parameter
      if (!List.of("ASC", "DESC").contains(orderByVersion)) {
        // Exception here
      }

      String namedQueryName = "ASC".equals(orderByVersion) ? "Snapshot.findOldestVersionFirst" : "Snapshot.findLatestVersionFirst";
```

modules/scheduler-impl/src/test/java/org/opencastproject/scheduler/impl/SchedulerServiceImplTest.java
```diff
@@ -852,78 +842,79 @@ public void testAddMultipleEventsConflict() throws Exception {
...
-    try {
-      schedSvc.updateEvent(mediaPackageId, Opt.<Date> none(), Opt.<Date> none(), Opt.<String> none(),
-              Opt.<Set<String>> none(), Opt.<MediaPackage> none(), Opt.<Map<String, String>> none(),
-              Opt.<Map<String, String>> none());
-      fail();
-    } catch (NotFoundException e) {
-      Assert.assertNotNull(e);
-    }
-  }
+// TODO: Figure out why getMediaPackage should throw an exception if it correctly returns an existing mediapackage
```
lkiesow: If we can't figure that out: Drop the test. It wouldn't be our first test with errors in the test.
In any case, we shouldn't just comment this out.

lkiesow (changes requested):
That you for doing this. I hated the review! But that you nonetheless :D 

This generally looks good. Most things I found probably existed before. I also did some local testing where I ran this by processing a few videos → no problems.

I'm going to do a few additional tests with different users in the next days.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
Arnei: Iirc I can't do that because JPQL does not support "LIMIT" like that.

I could instead add limit as a parameter to the related entity manager query, but I thought it would be bad style to split the logic of the query like that. I do not know which approach results in more efficient db queries.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/SnapshotDto.java
Arnei: And then contrary to my answer to the previous comment, here limit (and offset) are set in the entity manager query because they are supposed to be controlled ^^'.

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/Database.java
Arnei: > Looking at Asset.findByChecksumStorageIdAndOrganizationId, it joins the assets table with the snapshot table only to filter for one snapshot property (organization identifier) and then discard the snapshot data and return just an asset. What do we do with the asset? We use it to run another query to get the snapshot data 🙈

I might be heavily misunderstanding what you're getting at, but the snapshot data is not discarded, is it? It is contained in the result and used in the `.map` right after this query is run. I don't see a second query to get the snapshot data.

> Also, is it just me, or is AssetDtos.Full and AssetDtos.Medium a weird concept? Why not just use a tuple of Asset and Snapshot but copy over half of the snapshot data to another object?

I honestly did not bother understanding what the purpose of all the ".Full" and ".Medium" classes was, so I can't speak on that :').

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/persistence/AssetDtos.java
Arnei: That is possibly a very good idea, but not something I wanted to address in this PR, for fear of losing sight of me goal of removing query-dsl and instead getting stuck trying to rewrite everything.

lkiesow (approved):
Tested this again and it looks good.

Ignoring `AssetDtos.Full` and `AssetDtos.Medium` is fair since it's not an issue created by this pull request. I'll create an OSDN issue for that :grin:

PR #6696, opened 2025-05-14
Enable feature "Episode ID Roles" by default

LukasKalbertodt:
This is a very useful feature to have, especially for external systems. Especially especially so with JWT, which depends on this to make the `oc` claim work. For external apps, it's extremely useful to be able to rely on the presence of this feature and it will be annoying to tell everyone to enable it manually. For example, Tobira is already making use of this.

CC @lkiesow @Arnei  (since you implemented this in #5056)

### How to test this patch

Build & start Opencast, then somehow create a user with a magic role (permanent or just a JWT session) and see that the user has access to the event.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/configuration/episode-id-roles.md
+++ b/docs/guides/admin/docs/configuration/episode-id-roles.md
@@ -12,11 +12,11 @@ The `<ACTION>` will be capitalized, but special characters may not be converted 
 Setup
 --------------------
 
-Enable `episode.id.role.access` in `etc/custom.properties`.
+`episode.id.role.access` in `etc/custom.properties` is enabled by default.
 
 To make this work for the Admin UI and External API, the Elasticsearch Index needs to be updated with modified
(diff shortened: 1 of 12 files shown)


lkiesow:
Sometimes having less options (being more consistent) is nice. Would it hurt us to always enable this and remove the configuration completely?

Also, maybe we could mention in the release notes that you can do the reindexing at runtime while people are already using the system and the new roles will then be available eventually.

LukasKalbertodt:
> Sometimes having less options (being more consistent) is nice. Would it hurt us to always enable this and remove the configuration completely?

Hell yeah, I'm absolutely in favor of doing that. I was just not sure if that'd be fine for the community. If you think it's fine, I can adjust this PR.

> Also, maybe we could mention in the release notes that you can do the reindexing at runtime while people are already using the system and the new roles will then be available eventually.

I didn't know about that (in fact, I don't know any details about what exactly needs reindexing), but I can add it to the changelog.

LukasKalbertodt:
Rebased

KatrinIhler (approved):

PR #6702, opened 2025-05-15
Fix whispercpp operation when using the language from metadata

### LINKED ISSUE #6293: Fix incorrect language parameter for WhisperCPP
The same that was fixed here [#5950](https://github.com/opencast/opencast/pull/5950) should be implemented for WhisperCpp. The bug was discovered during the review of PR #6247.

Here the description copied from #5950: 
>The whisper --language option only accepts 2-letter language codes. However, the language codes provided by Opencast are 3-letter codes. This fix converts the language codes before using them with whisper.

marwyg:
Fixes #6293 

When using the language string from the event metadata, the speechtotext operation using the whisperc++ engine won't work. Now it should be fixed. I used the same logic that already works in the whisper engine version.

### Why in 17.x
It's a Bugfix and it shouldn't alter any other logic.

### How to test

1. Install whisperc++. I've done it with the quick start instructions from the github repository (it's nice and fast).
https://github.com/ggml-org/whisper.cpp?tab=readme-ov-file#quick-start

2. Edit Opencasts whispercpp config: `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg`
The configuration depends on the installation path of your whisperc++.
Configure the root path and the model path to something like this:
``` 
whispercpp.root.path=/path/to/cloned/github/project/directory/whisper.cpp/build/bin/whisper-cli
whispercpp.model=/path/to/cloned/github/project/directory/whisper.cpp/models/ggml-base.en.bin
```

3. Add the speechtotext operation somewhere in your workflows. I've put the following two operations in the default schedule-and-upload Workflow right before the "partial-publish" Operation. (Please notice that I only run the speechtotext operation on the presenter track. Don't forget this when uploading a video. You also can change the source-flavors configuration here)
```
    <operation
        id="speechtotext"
        description="Create subtitles with WhisperCPP">
      <configurations>
        <configuration key="source-flavor">presenter/source</configuration>
        <configuration key="target-flavor">captions/source</configuration>
        <configuration key="limit-to-one">true</configuration>
      </configurations>
    </operation>

    <operation id="tag">
      <configurations>
        <configuration key="source-flavors">captions/source</configuration>
        <configuration key="target-tags">+archive</configuration>
      </configurations>
    </operation>
```

4. Upload an event and set a language in the metadata.
5. The Workflow should run without errors and you should find a subtitle file in the Assets/Archive. You also should see your selected language in the tags.
image

### DIFF
--- a/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
+++ b/modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/engine/WhisperCppEngine.java
@@ -184,9 +184,6 @@ public class WhisperCppEngine implements SpeechToTextEngine {
   /** Path to the executable */
   protected String ffmpegBinary = DEFAULT_FFMPEG_BINARY;
 
-  /** The LangCodeUtil instance, contains language code mappings */
-  private LangCodeUtil langCodeUtil;
-
   @Override
   public String getEngineName() {
     return engineName;
@@ -279,9 +276,6 @@ public class WhisperCppEngine implements SpeechToTextEngine {
 
     ffmpegBinary = Objects.toString(cc.getBundleContext().getProperty(FFMPEG_BINARY_CONFIG_KEY), DEFAULT_FFMPEG_BINARY);
     logger.debug("ffmpeg binary set to {}", ffmpegBinary);
-
-    langCodeUtil = LangCodeUtil.getInstance();
-    logger.debug("LangCodeUtil initialized");
 
     logger.debug("Finished activating/updating speech-to-text service");
   }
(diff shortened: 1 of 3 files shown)


LukasKalbertodt (approved):
Thank you for fixing this bug! I just tested it by cherry picking your commit on top of 17.3 and it works as expected! Details:

I uploaded a single stream presentation video with english audio track and specified the event language as "englisch" while uploading via admin UI. I added this workflow operation:

```xml
<operation
        id="speechtotext"
        description="Generates subtitles via Whisper">
        <configurations>
            <configuration key="source-flavor">*/source</configuration>
            <configuration key="target-flavor">captions/source</configuration>
        </configurations>
    </operation>
```

Without this patch, processing failed bc Opencast passed `--language eng` to whispercpp, which whispercpp doesn't like and stopped. With this patch, `--language en` is passed and everything works!

I also skimmed over the code and I haven't found any problems (thanks for deduplicating the code instead of just copy&pasting). This wasn't an in-depth code review though.

geichelberger (changes requested):
What's the reason for implementing LangCodeUtil as a singleton?

marwyg:
The language conversion was originally used in the other Whisper engine. I just wanted to avoid code duplication, so I outsourced it. Is there a better way to do that? Because I had no particular reason to do it as a singleton.

marwyg:
@geichelberger I removed the singleton pattern and used a static initializer instead.

marwyg:
There was a checkstyle error and a wrong comment. I fixed both.

PR #6739, opened 2025-05-24
Revert PR 6382 Add User Info to Roles for 17.x

KatrinIhler:
This reverts #6382 for 17.x. This is the backend change that needs to go with the revert in the frontend (see https://github.com/opencast/opencast-admin-interface/pull/1306). I'm going to re-introduce this change to develop.


KatrinIhler:
Since I need this for an Admin UI release, I'm going to make an exception and merge this without discussion. We can still talk about this in the next Dev meeting and re-introduce the change to 17.x if needed.

Issue #6716, opened 2025-05-16
Improve Opencast Docs


KatrinIhler:
This can mean a number of things for example:
 
- Finish https://github.com/opencast/opencast/pull/5474
- Merge adopters and developers guides to enable searching across all documentation easier?
- Look at general structure
- Look at UI
- Reference videos on OC explore (have playlist containing videos for beginners/certain other topics)
- See #1961 (attention: might be out-of-date)
- See issues tagged with [documentation](https://github.com/opencast/opencast/issues?q=is%3Aissue%20state%3Aopen%20label%3Adocumentation)
- Weed out out-of-date information (e.g. references to the old mailing lists, releasemanager docs)
    - Mention that a RM should also cut Admin UI & Editor releases as needed
- Mention Admin UI, Editor, Tobira, ... 

**Advice:** Don't do everything at once, as that makes reviews harder!

PR #6742, opened 2025-05-26
Splitting the Changelog into Multiple Files and other Minor Docs Improvements

KatrinIhler:
This splits the changelog file into multiple files, one per major version of Opencast. The reason for this is that the original file had grown to a size that was difficult to handle for an IDE, making edits to it pretty cumbersome and annoying. This also means that we can hide the changelog of older versions in a subfolder.

The reason I'm aiming this against 16.x is that it would make the merging of the branches more difficult if they contained different structures for the changelog. I think it's better to fix merge conflicts once and then hopefully be done with it. 

I also fixed a few other warnings I got when building the docs, moved some entries in the navigation, and reworded the landing page a bit as some things were out-of-date (e.g. Opencast no longer comes with its own Capture Agent). I also felt like we didn't need to reproduce so much of the navigation on the landing page itself, opting to highlighting just a few important things instead.

In the future it would probably be better if we referenced stuff like the Editor, Tobira or the Admin UI repo on the landing page, but I didn't have time for that.

This somewhat relates to #6716  and https://github.com/opencast/helper-scripts/issues/77.


docs/guides/admin/docs/index.md
```diff
@@ -3,37 +3,27 @@ Opencast {{ opencast_major_version() }} Administration Guide
 
 Welcome to the Opencast Universe! Opencast is an open-source enterprise level lecture recording system. The core of the
 system delivers functionality for scheduling, media encoding, editing and content delivery. For lecture capture,
-Opencast provides capture agent software and third party appliances are available. An awesome community provides new
-features and support.
+Opencast third party appliances are available. An awesome community provides new features and support.
```
mtneug, line 6: pyCA is third party?

mtneug:
Looks good. Not sure about pyCA.

docs/guides/admin/docs/index.md
KatrinIhler: I figured this referenced the internal Capture Agent that Opencast used to have, but yeah, good point. No idea if this meant PyCA.

KatrinIhler:
Repointed to 17 after discussion in the Dev meeting

docs/guides/admin/docs/releasenotes.md
```diff
@@ -71,8 +71,8 @@ configure asset upload options in more detail [[#6362](https://github.com/openca
 - Waveform peaks are now more visible, but that behavior is configurable.
   [[#6028](https://github.com/opencast/opencast/pull/6028)]
 
-For more details, please take a look at the [full changelog](changelog.md). If you want to update Opencast from a
-previous version, you should also read the [upgrade guide](upgrade.md).
+For more details, please take a look at the [full changelog](changelog/opencast{{ opencast_major_version() }}.md).
```
gregorydlogan: `changelog/opencast{{ opencast_major_version() }}.md` resolves to `changelog/opencast17.md` at runtime, which generates `WARNING -  Doc file 'releasenotes.md' contains a link 'changelog/opencast17.md', but the target is not found among documentation files.`

gregorydlogan (changes requested):
Looks good to me aside from this warning.  Incidentally, this fixes another warning in the current https docs :)

docs/guides/admin/docs/releasenotes.md
KatrinIhler: Good catch, fixed!

PR #6802, opened 2025-06-16
Add support for extra data in adopter registration

gregorydlogan:
This PR does a few things:
- It adds support for arbitrary extra data to be sent as part of the normal adopter registration data.  This enables (easier) extensions for things like tracking integrations like Tobira.
- Adds a /stats endpoint for Tobira to put its stats information.
- Cleans up the adopter registration code, renaming classes and removing redundant interfaces and such.

Ideally this would go into a `r/17.x` release, however if it can't then `develop` would also work.  This code adds a database table, but should not require changing existing schema.

### How to test this patch

Start an instance of https://github.com/opencast/adopter-registration-server locally, setting `org.opencastproject.adopter.registration.server.url=http://localhost:5000/` in `custom.properties` prior to starting.  Register as normal, restart Opencast (or the adopter registration bundle).  Manual insert of data is required in the `/tobira/stats` endpoint, use the REST docs.  The tobira parameters supported are `num_realms, num_blocks, version, config`.  Omitted fields are silently ignored.

### Your pull request should…
* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
Example Tobira blob for use with the REST docs:
`{"num_realms": 2, "num_blocks": 9, "version": {"identifier": "dunno", "build_time_utc": "imastring", "git_commit_hash": "abc123", "git_was_dirty": true}, "config": {"auth_mode": "soemthing", "login_link_overridden": true, "logout_link_overridden": false, "uses_pre_auth": true, "has_narrow_logo": false, "download_button_shown": true}}`

owi92:
I have not looked at the code, but I have tested this using both your provided blob (thank you), and "real" data from my local Tobira. I can confirm that this works with the REST docs form.

One thing to note is that Tobira itself sends these stats as raw JSON, so this will currently end in a 415 response. But that is an easy fix on our end, if you want or need to stick with url encoding.

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -625,6 +627,7 @@
   <feature name="opencast-presentation" version="${project.version}">
     <feature>opencast-core</feature>
 
+    <bundle start-level="82">mvn:org.opencastproject/opencast-adopter-registration-api/${project.version}</bundle>
```
KatrinIhler, line 628: Some of these seem to be duplicates.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/AdopterRegistrationServiceImpl.java
```diff
@@ -21,80 +21,470 @@
...
+          StatisticData statisticData = collectStatisticData(adopter.getAdopterKey(), adopter.getStatisticKey());
+          sender.sendStatistics(statisticData.jsonify());
+          db.exec(em -> {
+            TypedQuery<AdopterRegistrationExtra> q = em.createNamedQuery("AdopterRegistrationExtra.findAll",
+                AdopterRegistrationExtra.class);
+            q.getResultList().forEach(extra -> {
+              try {
+                Map<String, Object> data = Map.of("statistic_key", statisticData.getStatisticKey(),
+                    "data", gson.fromJson(extra.getData(), Map.class));
+                sender.sendExtraData(extra.getType(), gson.toJson(data));
```
KatrinIhler, line 322: Question: Since this is sent separately, can we associate that data with the general adopter's data afterwards?

KatrinIhler (changes requested):
Two minor things. Otherwise looks reasonable. I guess we can put this into 17.

modules/adopter-registration-impl/src/main/java/org/opencastproject/adopter/registration/AdopterRegistrationServiceImpl.java
gregorydlogan: Yes. The statistics uuid is the statistics object's uuid. The statistics object has the adopter object's uuid :)

assemblies/karaf-features/src/main/feature/feature.xml
gregorydlogan: You'd think Intellij would keep these in order by default.  Good catch, ty.

KatrinIhler (approved):

LukasKalbertodt:
> One thing to note is that Tobira itself sends these stats as raw JSON, so this will currently end in a 415 response. But that is an easy fix on our end, if you want or need to stick with url encoding.

Tobira can of course easily fix this, but then it means that data sent by older Tobiras will be ignored right? Is there any good reason to switch to `application/x-www-form-urlencoded` here? If not, I would suggest to change it back in OC, to be able to gather from older Tobiras and not create needless breakage.

---

On a different note: does this PR also solve the long standing issue with multi-node systems? I.e. with this, does Tobira adopter tracking finally fully work? :O that would be great

gregorydlogan:
@LukasKalbertodt yes, multinode is now fixed using the magic of schlepping a bunch of json into a database column :D

Will investigate why the application type changed.  I can't think of a solid reason here.

Edit: It didn't, additional support for the urlencoded stuff was added.  Enquiring in chat to see what else is going on here.

PR #6813, opened 2025-06-20
Remove series load in paella player

### LINKED ISSUE #6812: Paella Player fails when series is missing
**Describe the bug**

The paella player fetches the series from Opencast if the event has a linked series. If the user has for some reasons no permissions on the series, the player shows a cryptic error message, see screenshot. In our case, the role based event ALCs are missing in the `searchable_acl` property of the series.  

**Expected behavior**

The player should load the event and log a warning to the browser console.

**Screenshot**

Image

**Stack trace**
```
paella-core - Error: TypeError: n.result[0] is undefined <anonymous code>:1:145535
    overrideMethod <anonymous code>:1
    Ir Log.js:57
    error Log.js:85
    value Log.js:154
    t Paella.js:685
    f Paella.js:2
    _ Paella.js:2
    L Paella.js:2
    Rn Paella.js:2
    s Paella.js:2
```

**Server environment:**
 - Version 17.4

dennis531:
I see no reason why this series fetch from Opencast is needed because the fetched episode data contains the series title. This series fetch was introduced in https://github.com/opencast/opencast/commit/4aaca26a0f9ebcb84ed440bdd23cbfbe5d59298.

This patch fixes #6812.

### DIFF
--- a/modules/engage-paella-player-7/package-lock.json
+++ b/modules/engage-paella-player-7/package-lock.json
@@ -19,7 +19,7 @@
         "@babel/core": "^7.25.7",
         "@babel/eslint-parser": "^7.25.9",
         "@babel/preset-env": "^7.26.0",
-        "@playwright/test": "^1.48.2",
+        "@playwright/test": "^1.53.1",
         "babel-loader": "^9.2.1",
         "copy-webpack-plugin": "^12.0.2",
         "css-loader": "^7.1.2",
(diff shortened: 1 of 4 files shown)


mtneug:
There is a plugin for showing other videos in a series (deactivated by default). Not sure, if this data is needed there.

dennis531:
> There is a plugin for showing other videos in a series (deactivated by default). Not sure, if this data is needed there.

If I see it correctly, the deleted series data is not passed to the plugin or has any dependencies on the plugin. The deleted code simply fetches the series to get its title. However, this step is unnecessary because the episode already contains the series title.

gregorydlogan (approved):
This works as expected and does not break the other-videos-in-series bits.

Tests appear to be failing because the great googly moogly has upgraded chrome and we don't have the relevant updated bits in our testing stuff.

gregorydlogan:
Fixed the playwright issue, pushed to this PR since this PR also exposed that things were broken.  Once these pass I can merge this, unless someone else gets to it first.

PR #6815, opened 2025-06-20
Remove full-text search escaping

geichelberger:
The search index uses `multi_match` for full-text search, which is based on match, and does not use the powerful query string syntax based on Apache Lucene query syntax. We must remove the escaping, or we will not get exact matches for keyword fields, like searching for the media package ID.

Fixup PR https://github.com/opencast/opencast/pull/6498


gregorydlogan (changes requested):
The whole `QueryPreprocessor` class is now unused, remove this too?

PR #6821, opened 2025-06-24
Remove fast workflow prior to release

gregorydlogan:
This PR removes the fast workflow as part of the release process.  This makes the removal transparent to development since it happens just before the final build in GHA.

Targetting r/18.x since we don't have an 18.0, happy to backport to wherever if we want to apply this more broadly.

### DIFF
--- a/.github/make-release
+++ b/.github/make-release
@@ -49,6 +49,7 @@ test "$(jq .consistent < version.json)" = "true"
 
 # commit changes
 git add pom.xml modules/*/pom.xml assemblies/pom.xml assemblies/*/pom.xml
+git rm -f etc/workflows/fast.xml
 git commit -m "Opencast ${VERSION}"
 git tag "${VERSION}" -m "Opencast ${VERSION}"
 
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -30,7 +30,6 @@ jobs:
 
       - name: build opencast
         run: |
-          rm -f etc/workflows/fast*
           mvn clean install \
             --batch-mode \
             -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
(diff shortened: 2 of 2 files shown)


mtneug:
ok so removing fast was a consensus? That means I need to remove the workflow from the container images as well or do people want that in there for quick testing with the container images?

gregorydlogan:
It was under-discussed, but I figured I'd file this so we have the discussion.  I don't really care either way, but I can definitely see the argument for removing this as a default.

lkiesow:
-1 to this specific implementation of the idea.

I like the idea of removing the workflow from the release, but this will only remove it from the binary distributions we upload to GitHub. If you download the source from the same release page, you will still get it. That may easily cause confusion. I expect the binary to be the same as me downloading the source and compiling it.

If we want to remove the workflow, we should remove it right before creating the release commit.

gregorydlogan:
I mean, that's easily enough done.  Can discuss further at the next meeting.

.github/make-release
```diff
@@ -49,6 +49,7 @@ test "$(jq .consistent < version.json)" = "true"
 
 # commit changes
 git add pom.xml modules/*/pom.xml assemblies/pom.xml assemblies/*/pom.xml
+git rm -f etc/workflows/fast.xml
```
lkiesow, line 52: Actually :see_no_evil:…

This is part of the [Cut Release Tag](https://github.com/opencast/opencast/blob/develop/.github/workflows/cut-release-tag.yml) workflow, which is supposed to replace release managers manually cutting releases, but I never finished that. It still has a bug and I didn't fix that. But let me put that on the OSDN board.

gregorydlogan: I wondered about that :)  I haven't been an RM in long enough that I wasn't sure if this was in use.

lkiesow: It's now on the OSDN board → https://github.com/opencast/opencast/issues/6829
Hopefully someone can pick that up and finish it.

gregorydlogan:
I'm going to decline this since this won't be moving forward.

PR #6825, opened 2025-06-25
HttpClient Timeouts in Stud.IP User Directory

lkiesow:
If the Stud.IP user directory is configured, but the LMS doesn't answer HTTP requests, the HTTP client will just hang, blocking all kind of internal Opencast functionality.

This patch adds a simple timeout configuration as a mitigation. Opencast will still be slowed down at several places, but it will at least continue to function.

### How to test this patch

Enable the Stud.IP user directory plugin and configure the provider by creating `etc/org.opencastproject.userdirectory.studip-default.cfg` containing:

```properties
org.opencastproject.userdirectory.studip.org=mh_default_org
org.opencastproject.userdirectory.studip.url=http://localhost:8000
org.opencastproject.userdirectory.studip.token=sometoken
```

Run a non-answering fake LMS:
```
while true; do nc -l 8000; done
```

Create a user not called `admin` and log-in. The login will take ~10s since you run into the timeout.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/userdirectory-studip/src/main/java/org/opencastproject/userdirectory/studip/StudipUserProviderInstance.java
+++ b/modules/userdirectory-studip/src/main/java/org/opencastproject/userdirectory/studip/StudipUserProviderInstance.java
@@ -38,11 +38,12 @@ import com.google.common.cache.LoadingCache;
 import com.google.common.util.concurrent.ExecutionError;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClients;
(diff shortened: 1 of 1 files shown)


mtneug:
Since all LMS user providers are similar, does this also affect Moodle, Sakai, etc.?

lkiesow:
I didn't check, but I guess it's likely.

dennis531:
Would it make sense to implement configurable timeouts so that different universities can adjust them according to their response times?

dennis531 (approved):
The code looks good. I tested this patch against my local development Stud.IP and the user directory works as expected. Thank you for this Fix!

lkiesow:
> Would it make sense to implement configurable timeouts so that different universities can adjust them according to their response times?

I thought about it and didn't do it since this is basically just a safeguard and shouldn't affect normal operations. If your LMS repeatedly answers slower, your Opencast hangs and your system is really no longer working properly. Going too low would, on the other hand, easily cause problems with some requests being lost, causing users to not get their actual roles.

What this essentially means is that the window for a sensible configuration is actually pretty narrow, and I don't really see anyone starting to tweak this. Let's rather go with fewer options and a simpler configuration. If the need arises, we can always simply add the options later.

PR #6794, opened 2025-06-10
Update Whisper.cpp defaults to reflect new binary names

lkiesow:
The default binary names of Whisper.cpp have changed. This patch adjusts the Opencast code to reflect the new binary names. For more details, see the deprecation warning at
https://github.com/ggml-org/whisper.cpp/tree/v1.7.5/examples/deprecation-warning

Starting with Opencast 18, this change will also be included in Opencast's package repository. The additional change to the default model folder is done to also match the packages.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [x] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)
* [x] explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch

PR #6838, opened 2025-06-27
Add Voice Activity Detection (VAD) configs to whisperCpp

dennis531:
This PR adds configurations for the Voice Activity Detection (VAD) of WhisperC++, see https://github.com/ggml-org/whisper.cpp?tab=readme-ov-file#voice-activity-detection-vad.

### How to test this patch
1. Install WhisperC++ and configure Opencast according to https://docs.opencast.org/r/17.x/admin/#configuration/transcription.configuration/whispercpp/#enable-whisperc-engine.
2. Configure VAD in `org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg`
3. Upload a video and see the generated subtitles in the player.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
@@ -61,7 +61,7 @@
 #whispercpp.vad=false
 
 # VAD model path
-#whispercpp.vad-model=/usr/share/ggml/silero-v5.1.2-ggml.bin
+#whispercpp.vad-model=/usr/share/whisper.cpp/models/ggml-silero-v5.1.2.bin
 
 # VAD threshold for speech recognition
 #whispercpp.vad-thold=0.50
--- a/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
@@ -73,7 +73,9 @@
 #whispercpp.vad-min-silence-dur=100
 
 # VAD max speech duration (auto-split longer) in s
-#whispercpp.vad-max-speech-dur=FLT_MAX
+# Needs to be a valid floating-point number.
+# Leave empty for FLT_MAX, i.e., the maximum representable finite floating-point number.
+#whispercpp.vad-max-speech-dur=31536000
 
 # VAD speech padding (extend segments) in ms
 #whispercpp.vad-speech-pad=30
--- a/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
+++ b/etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
@@ -53,3 +53,30 @@
 # This is done by Opencast itself before handing over the audio track.
 # Default: true
 #whispercpp.auto-encode=true
+
+
+## Optional settings for Voice Activity Detection (VAD)
+
+# Enable Voice Activity Detection (VAD)
+#whispercpp.vad=false
+
+# VAD model path
+#whispercpp.vad-model=/usr/share/ggml/silero-v5.1.2-ggml.bin
+
+# VAD threshold for speech recognition
+#whispercpp.vad-thold=0.50
+
+# VAD min speech duration (0.0-1.0) in ms
+#whispercpp.vad-min-speech-dur=250
+
+# VAD min silence duration (to split segments) in ms
+#whispercpp.vad-min-silence-dur=100
+
+# VAD max speech duration (auto-split longer) in s
+#whispercpp.vad-max-speech-dur=FLT_MAX
+
+# VAD speech padding (extend segments) in ms
+#whispercpp.vad-speech-pad=30
+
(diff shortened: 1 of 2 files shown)


lkiesow:
Can we shift this to the `r/18.x` branch? We agreed to package the newer version for OC18 since they also renamed the binaries, … which may be confusing.

mtneug:
I include the new Whisper.cpp versions in 17.x container images, but rename the binary to the old name. 18.x images use the new name. So users of the container images could use these options. What do you think? Still redirect to `r/18.x`?

etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
```diff
@@ -54,3 +54,30 @@
 # This is done by Opencast itself before handing over the audio track.
 # Default: true
 #whispercpp.auto-encode=true
+
+
+## Optional settings for Voice Activity Detection (VAD)
+
+# Enable Voice Activity Detection (VAD)
+#whispercpp.vad=false
```
mtneug, line 61: Should we enable this by default? Currently, there is only the Silero model, so users don't really have options to choose from, like with the base Whisper model. Moreover, the Silero model is < 1mb, and we could easily always include it as a hard dependency. Opinions from RPMs/DEBs perspective @lkiesow, @gregorydlogan?

lkiesow: Technically this would mean that you couldn't run with the old configuration anymore, which we shouldn't do on a release branch. But we break that rule at other places as well, so if the RM (@dennis531) is fine with that, I'm fine with that :shrug: 

In that case, we definitely shouldn't target `r/17.x`, though.

> Opinions from RPMs/DEBs perspective

I'll definitely package the model for OC18.

gregorydlogan: I've packed with silero model as it's own, optional dependency. I'm fine making it a hard requires until there are more vad models, but I figured if you're not using VAD then you don't technically need the model. Don't much care though 🤷

dennis531: If I understand it correctly, you propose to set this parameter to `whispercpp.vad=true` per default.  In OC 18, the vad model will always be installed in the path `/usr/share/ggml/silero-v5.1.2-ggml.bin`?

I think I am fine with this change, since OC 18 hasn't been released yet.

mtneug: Let's postpone this decision for OC 19 and give this some time to bake in OC 18. We can always enable this by default in later versions.

mtneug:
@dennis531, we discussed this in the dev meeting and propose to (1) redirect this to `r/18.x` and (2) have a generic option possible in `r/17.x` to pass any flags to Whisper.cpp [similar to how the "normal" Whisper can be configured](https://github.com/opencast/opencast/blob/3efb11e04b6518c7d4cbce54d45ac1eaef7e0b4f/etc/org.opencastproject.speechtotext.impl.engine.WhisperEngine.cfg#L10-L12). In this way, people could still configure VAD in 17.x, but we don't advertise official support.

etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
```diff
@@ -54,3 +54,30 @@
...
+
+# VAD model path
+#whispercpp.vad-model=/usr/share/ggml/silero-v5.1.2-ggml.bin
```
mtneug: This should be changed to following #6794. Filename is the one used by the download script.
```suggestion
#whispercpp.vad-model=/usr/share/whisper.cpp/models/ggml-silero-v5.1.2.bin
```

etc/org.opencastproject.speechtotext.impl.engine.WhisperCppEngine.cfg
```diff
@@ -54,3 +54,30 @@
...
+#whispercpp.vad-thold=0.50
+
+# VAD min speech duration (0.0-1.0) in ms
+#whispercpp.vad-min-speech-dur=250
+
+# VAD min silence duration (to split segments) in ms
+#whispercpp.vad-min-silence-dur=100
+
+# VAD max speech duration (auto-split longer) in s
+#whispercpp.vad-max-speech-dur=FLT_MAX
```
mtneug: `FLT_MAX` is from the usage output, but apparently not a valid value. Whisper.cpp will fail with
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stof
Aborted
```
IMO we should somehow indicate that this needs to be a valid float
```suggestion
# Needs to be a valid floating-point number. Leave empty for FLT_MAX, i.e., the maximum representable finite floating-point number.
#whispercpp.vad-max-speech-dur=FLT_MAX
```

mtneug (changes requested):
This works as expected. I have some smaller complaints regarding the config file.

mtneug (approved):

PR #6842, opened 2025-06-27
Faster Tests (Build JavaDocs only once)

lkiesow:
This tests build Opencast with two Java versions, but they run the integration tests only with one version. The JavaDocs are also built with two versions.

It's somewhat unlikely that the JavaDocs break with a Java version update, nor do we really care that much. That is why this patch makes them run only once on the version that is not running the integration tests, so that they can run in parallel. This should make tests finish about 5 minutes faster.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-build.yml
+++ b/.github/workflows/test-opencast-build.yml
@@ -75,6 +75,7 @@ jobs:
 
         #This has to happen in a separate step or else various service classes aren't visible
       - name: build javadocs
+        if: matrix.java == 23
         run: |
           mvn javadoc:javadoc javadoc:aggregate -Pnone \
             --batch-mode \


gregorydlogan:
So the issue here is that the jump from 11 to 17 exposed issues where the javadocs *didn't* build in 17, but did in 11.  Do I think that's likely to happen again?  Unclear, but there was a reason behind doing it multiple times.

mtneug:
But I would guess JavaDocs for 17 would build with 11? So, targeting the latest supported JDK should be enough?

gregorydlogan:
Yeah, good point. Regardless I'm not a -1 to this, just giving context.

PR #6883, opened 2025-07-02
Limit OpenMP Threads only for Tesseract

mtneug:
Currently, Opencast sets the environment variable `OMP_THREAD_LIMIT` to 1, thus limiting all spawned subprocesses using OpenMP to only use one thread. This was initially introduced by #2051 to limit Tesseract that could otherwise lock up systems. However, this also limits any other binary built upon OpenMP, e.g. Whisper.cpp.

This PR removes the default from `setenv` and sets the environment variable only when spawning the Tesseract subprocess. As such, the environment variable is now hard-coded. If this is undesired, a configuration option could be introduced.

### How to test this patch

* Configure the system for Whisper.cpp
* Check before and after this patch how many threads are used. The default should be 4.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):
Considering I've never heard of any adopters modifying this setting, I think hardcoding it is reasonable.

PR #6884, opened 2025-07-02
Allow to escape colons in workflow variable syntax

mtneug:
Workflow variables have the following syntax:

    ${var:default}

A colon is used to separate the variable name from the default value. However, the variable itself might contain a colon and there is currently no way to use such a variable. A specific use case is the `analyze-mediapackage` operation with `set-tag-variables` enabled. For subtitle tracks, this might add a workflow property such as

    captions_source_hastag_lang:en

This PR changes the regex to parse variables and allows to escape colons in the following way:

    ${var-with-\:-colon:default-with\:-colon}

This is breaking in the sense that the string `\:` is now replaced in the variable name and default value with a `:`, but I don't think anybody used `\:` at the moment. So this should not be breaking in practice.

### How to test this patch

Use a variable with a colon in a workflow.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):
This seems fine to me.  I can potentially people wanting this in 18.x rather than 17.x since it's breaking, but IMO it's fine in 17.  Let's discuss in the technical meeting.

mtneug:
@KatrinIhler have you decided whether to put this in 17 or not?

KatrinIhler:
> @KatrinIhler have you decided whether to put this in 17 or not?

If you add release notes for this this can go into 17 imo.

mtneug:
I added a release note.

PR #6885, opened 2025-07-03
Fix snapshot deletion for remote only MediaPackages

mtneug:
Deletion of snapshots will currently fail if all snapshots live in remote asset stores. When trying to delete from the local asset store the base directory of the MediaPackage will not be found leading to an exception when trying to concatenate a path. This patch adds a null check that will skip the deletion from the local asset store if the base directory cannot be found.

### How to test this patch

1. Configure S3 as asset store.
2. Move all snapshots to S3, e.g. with the `move-storage` operation.
3. Delete snapshots, e.g. with the `asset-delete` operation including `keep-last-snapshot: true`.
4. Step 3 will fail without this patch.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan (approved):
LGTM

PR #6902, opened 2025-07-14
Remove annotations module

gregorydlogan:
This PR, as discussed in #6515 removes the annotations service.  Placing this into `r/18.x` since the ticket says to put it there, however we are after the branch cut so perhaps this should go into `develop`.  No strong feelings either way.

### How to test this patch

Test as usual, ~afaik~ nothing is using this service ~aside from (maybe?) the annotation tool?  Will discuss at upcoming technical meeting~.

### Your pull request should…

* [x] have a concise title
* [x] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [x] include migration scripts and documentation, if appropriate
* [x] pass automated tests
* [x] have a clean commit history
* [x] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)
* [ ] ~explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch~

Issue #5615, opened 2024-02-20
Major Changes Upcoming in Opencast 18


gregorydlogan:
This ticket will contain a list of major upcoming changes in OC 16 which adopters should budget time and space to handle when they upgrade

- Replacing the Solr Search Index with Elasticsearch #5597
- Removing Paella 6  #5605 
- Removing/demoting generally unused modules
  - This would be things like the Animate service
 - Removing old admin ui (`admin-ng`)
    -  This removes the integrated editor as well

Arnei:
Potentially removing the "old" admin ui (`admin-ng`), which would also mean removing the integrated editor.

mtneug:
Here are some potential modules that could be removed in 16:

* `admin-ui-frontend`
* `animate-api`
* `animate-impl`
* `animate-remote`
* `animate-workflowoperation`
* `annotation-api` (["...Planned removal is the Opencast release in December 2017..." ups](https://github.com/opencast/opencast/blob/8f2e828bcacfc15e80850003f441136dbc220b9a/modules/annotation-impl/src/main/java/org/opencastproject/annotation/impl/AnnotationRestService.java#L73-L74))
* `annotation-impl`
* `dictionary-hunspell` (other dictionaries?)
* `engage-paella-player`
* [`fileupload` (is this used?)](https://github.com/opencast/opencast/pull/6520)
* `presets` (is this used?)
* `search-service-feeds`
* `solr`
* [`sox-api` (use FFmpeg)](https://github.com/opencast/opencast/pull/6682)
* [`sox-impl`](https://github.com/opencast/opencast/pull/6682)
* [`sox-remote`](https://github.com/opencast/opencast/pull/6682)
* [`sox-workflowoperation`](https://github.com/opencast/opencast/pull/6682)
* `usertracking-api` (is this used?)
* `usertracking-impl`

gregorydlogan:
Per discussion at the technical meeting on 2024-02-27, all of the above could be removed aside from `admin-ui-frontend`, which should be moved to a *plugin* and disabled by default.

gregorydlogan:
Lars unclear if this is useful, but we built this prior to 16.x cut.  Pretty sure this isn't useful at this point?

mtneug:
Some modules have not been removed. So rename issue to Opencast 17?

ziegenberg:
We might need to update the list of removed and potentially-to-be-removed modules and rename the issue again... We're getting there, slowly, but we are getting there.

[...]

PR #6904, opened 2025-07-14
Remove usertracking service

gregorydlogan:
This PR, as discussed in #6515 removes the usertracking service.  Placing this into `r/18.x` since the ticket says to put it there, however we are after the branch cut so perhaps this should go into `develop`.  No strong feelings either way.

### How to test this patch

Test as usual, afaik nothing is using this service.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


gregorydlogan:
... I can't seem to even turn the relevant REST calls on from our Paella config, so maybe Paella 7 removed the front end of this entirely already?

miesgre:
> ... I can't seem to even turn the relevant REST calls on from our Paella config, so maybe Paella 7 removed the front end of this entirely already?

You will need to remove this plugin: https://github.com/opencast/opencast/blob/develop/modules/engage-paella-player-7/src/plugins/org.opencast.paella.opencast.userTrackingDataPlugin.js

and the config:
https://github.com/opencast/opencast/blob/develop/etc/ui-config/mh_default_org/paella7/config.json#L644-L649

gregorydlogan:
Thanks @miesgre, totally forgot about those!

Arnei (approved):
Same comments as in https://github.com/opencast/opencast/pull/6902#pullrequestreview-3046744316

gregorydlogan:
Rebased onto current develop.

smarquard:
This doesn't seem to be discussed in #6515 - is that the right reference?

We're using this service. What replaces it?  The argument about "it puts a large load on your database" and therefore should go is really spurious. You can configure what events to track. Unless this module is causing some other issue with code maintainability, it should stay. Otherwise it's imposing an additional burden on OC users who need these statistics.

mtneug:
> This doesn't seem to be discussed in #6515 - is that the right reference?

Number swap. #5615 is the correct reference.

This proposal is quite old and was first meant for 16.x. We also discussed this during the last OC summit, basically asking people what on that list is still really used. `usertracking` is on there because of issues in the implementation. I'm fine with leaving this in if it's useful to you. Currently, this is an opt-in plugin anyways.

There are no replacements at the moment, but #6916 tracks potential work in OSDN for a replacement.

gregorydlogan:
No particular objection to keeping this around, I was just cleaning out my TODOs.  I'll leave this open so that it shows up in my filters from time to time.

PR #6937, opened 2025-07-25
Fix Partial path traversal vulnerability from remote UIConfigRest

odaysec:
https://github.com/opencast/opencast/blob/2fd7316569c6856b2db8ba3334f673501fe1c818/modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java#L173-L173

A common way to check that a user-supplied path `SUBDIR` falls inside a directory `DIR` is to use `getCanonicalPath()` to remove any path-traversal elements and then check that `DIR` is a prefix. However, if `DIR` is not slash-terminated, this can unexpectedly allow accessing siblings of `DIR`.

fix the partial path traversal vulnerability, we need to ensure that the base path used in the prefix check is slash-terminated before comparing it to the canonical path of the requested file. This prevents attackers from accessing sibling directories by crafting a path that starts with the base path but is not actually within it. The best way to do this is to append a file separator (`File.separator`) to the base path after calling `getCanonicalPath()`, and then check that the canonical path of the requested file starts with this slash-terminated base path. This change should be made in the `getConfigFile` method, specifically where the prefix check is performed (lines 169–173). No additional dependencies are required.

#### References
[Partial Path Traversal](https://owasp.org/www-community/attacks/Path_Traversal)
[ESAPI Security_API](https://github.com/ESAPI/esapi-java-legacy/blob/develop/documentation/GHSL-2022-008_The_OWASP_Enterprise_Security_API.md)
 [CWE-23](https://cwe.mitre.org/data/definitions/23.html)

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
+++ b/modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
@@ -166,12 +166,11 @@ public class UIConfigRest {
     File configFile = Paths.get(uiConfigFolder, orgId, component, filename).toFile();
 
     try {
-      final String basePath = new File(uiConfigFolder, orgId).getCanonicalPath();
-      final String basePathWithSep = basePath.endsWith(File.separator) ? basePath : basePath + File.separator;
+      final String basePath = new File(uiConfigFolder, orgId).getCanonicalPath() + File.separator;
       final String configFileCanPath = configFile.getCanonicalPath();
 
       // is configFile a subdirectory of basePath (additional directory traversal protection), if not stop
-      if (!configFileCanPath.startsWith(basePathWithSep)) {
+      if (!configFileCanPath.startsWith(basePath)) {
         logger.debug("Directory traversal prevented (trying to access '{}')", configFile.getPath());
         throw new AccessDeniedException(configFileCanPath);
       }
--- a/modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
+++ b/modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
@@ -167,10 +167,11 @@ public class UIConfigRest {
 
     try {
       final String basePath = new File(uiConfigFolder, orgId).getCanonicalPath();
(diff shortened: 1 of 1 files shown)


modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
```diff
@@ -167,10 +167,11 @@ public Response getConfigFile(@PathParam("component") String component, @PathPar
 
     try {
       final String basePath = new File(uiConfigFolder, orgId).getCanonicalPath();
+      final String basePathWithSep = basePath.endsWith(File.separator) ? basePath : basePath + File.separator;
```
lkiesow: Since `getCanonicalPath` should never end with a file separator, you can actually simplify your code:

```suggestion
      final String basePath = new File(uiConfigFolder, orgId).getCanonicalPath() + File.separator;
```

modules/user-interface-configuration/src/main/java/org/opencastproject/uiconfig/UIConfigRest.java
```diff
@@ -167,10 +167,11 @@ public Response getConfigFile(@PathParam("component") String component, @PathPar
...
       final String configFileCanPath = configFile.getCanonicalPath();
 
       // is configFile a subdirectory of basePath (additional directory traversal protection), if not stop
-      if (!configFileCanPath.startsWith(basePath)) {
+      if (!configFileCanPath.startsWith(basePathWithSep)) {
```
lkiesow: In combination with the change above:
```suggestion
      if (!configFileCanPath.startsWith(basePath)) {
```

lkiesow (changes requested):
Good catch. This is actually a nice find. Thank you!

While this is pretty unlikely to actually cause a security issue for an adopter due to the structure of Opencast, this should probably go into the legacy release branch as well. Can you change this pull request to target the `r/17.x` branch instead of `develop`? That might also need y rebase of your branch.

Once merged, I'll then also release a minor security advisory for this.

Also, if you have any questions about filing the ICLA, let me know.

PR #6938, opened 2025-07-29
Reactivate user role filter for admin ui

Arnei:
Helps with https://github.com/opencast/opencast-admin-interface/issues/1287.

Allows users of the admin interface to filter the users table by roles once again. This was originally removed in https://github.com/opencast/opencast/pull/3172.

The filter was originally removed due to performance concerns and to this day it is still not terribly performant. On my local testing instance with about 20000 users, a request with a role filter could take between 350 ~ 400ms, compared to 80 ~ 120ms without a role filter. However, we are allowing admin interface users to use the textFilter (aka the search bar) to filter by role, which has the exact same performance concerns as using the role filter directly. Therefore I argue that we should either allow users to use the role filter, or disable the textFilter either just for roles or alltogether.

Since allowing role filters hinges on performance, I would like the help of adopters with many users (thousands or more) and ideally multiple user providers to test this.

### How to test this patch

After installing this, open the admin interface. Go to the users table and set a role filter. Open your browsers dev tools(F12), open the network tab and observe the time the GET users.json... request takes.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
@@ -83,7 +83,6 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


mtneug:
> On my local testing instance with about 20000 users, a request with a role filter could take between 350 ~ 400ms, compared to 80 ~ 120ms without a role filter.

That doesn't sound too bad. We have/had APIs that take multiple seconds (starting workflows, getting events with (extended) metadata, ...).

To limit the amount of work Opencast has to do, maybe don't fetch users every 5s? This doesn't change so often after all.

Arnei:
I agree it does not sound too bad, but it is also more of a best case scenario than a worst case scenario. Therefore I would really like some numbers from other Opencast instances.

We could definitely lower the refresh rate for the users table. My main concern with that is that it could confuse users if different tables have different refresh rates (which is admittedly not that big a concern).

gregorydlogan:
I'm going to merge this so that we can gather some real world data - worst case this is going to cause some slowdowns and force adopters to complain, best case no one notices.

Issue #5335, opened 2023-10-12
Non-ASCII characters in JWTs are broken


LukasKalbertodt:
**Describe the bug**

Data from a JWT that uses non-ASCII characters gets mangled somewhere by Opencast.

**To Reproduce**

1. [Configure JWT](https://docs.opencast.org/develop/admin/#configuration/security.jwt/#configuration-for-jwt-based-authentication-and-authorization): 
  - Add `opencast-security-jwt/14.0.0.SNAPSHOT, \` to `org.apache.karaf.features.cfg`
  - Configure `mh_default_org.xml` by following the link. The relevant part of the config:
    <details>

    ```xml
    <!-- General JWT header extraction filter
    <bean id="jwtHeaderFilter" class="org.opencastproject.security.jwt.JWTRequestHeaderAuthenticationFilter">
      <property name="principalRequestHeader" value="Authorization"/>
      <property name="principalPrefix" value="Bearer "/>
      <property name="authenticationManager" ref="authenticationManager" />
      <property name="loginHandler" ref="jwtLoginHandler" />
      <property name="exceptionIfHeaderMissing" value="false" />
    </bean>-->

    <!-- General JWT request parameter extraction filter-->
    <bean id="jwtRequestParameterFilter" class="org.opencastproject.security.jwt.JWTRequestParameterAuthenticationFilter">
      <property name="parameterName" value="jwt" />
      <property name="authenticationManager" ref="authenticationManager" />
      <property name="loginHandler" ref="jwtLoginHandler" />
      <property name="exceptionIfParameterMissing" value="false" />
    </bean>

    <!-- JWT login handler-->
    <bean id="jwtLoginHandler" class="org.opencastproject.security.jwt.DynamicLoginHandler">
      <property name="userDetailsService" ref="userDetailsService" />
      <property name="userDirectoryService" ref="userDirectoryService" />
      <property name="securityService" ref="securityService" />
      <property name="userReferenceProvider" ref="userReferenceProvider" />
      <property name="jwksCacheExpiresIn" value="1440" />
      <property name="secret" value="banana" />
      <property name="expectedAlgorithms" ref="jwtExpectedAlgorithms" />
      <property name="claimConstraints" ref="jwtClaimConstraints" />
      <property name="usernameMapping" value="['username'].asString()" />
      <property name="nameMapping" value="['name'].asString()" />
      <property name="emailMapping" value="'mail@invalid'" />
      <property name="roleMappings" ref="jwtRoleMappings" />
[...]

PR #6965, opened 2025-08-03
Remove reencoding to utf-8 of JWTs

jduehring:
This PR fixes https://github.com/opencast/opencast/issues/5335 . 

Before this patch some parts of the JWT like `name` and `mail` would be interpreted as ISO_8859_1 encoded and reencoded to UTF-8 to "ensure UTF_8 encoding". However if the string is already correctly UTF-8 encoded, this could have led to some characters being misinterpreted.

This patch removes the reencoding as to my understanding the auth0 library always decodes JWTs as UTF-8 anyways according to JWT specification.

### How to test this patch

See https://github.com/opencast/opencast/issues/5335 for a detailed explanation how to set up and test jwt in Opencast. 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
+++ b/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
@@ -48,7 +48,6 @@ import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.util.Assert;
 
-import java.nio.charset.StandardCharsets;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


LukasKalbertodt (approved):
Code looks good and I tested it locally and it indeed fixes the bug. Thanks!

I tried looking at the git history to figure out why this encoding code was added in the first place, but couldn't find out anything since it was added with basically the first JWT commit. 
@joneug do you remember why you added this encoding code? I don't see how ISO‑8859‑1 would ever be relevant here.

But yeah, unless anyone finds a good reason for this code, we can merge this after technical meeting discussion.

joneug:
I implemented the `security-jwt` module based on the `security-aai` module. There, the `DynamicLoginHandler` was implemented with this special treatment for the user's email address and name. That's how this made it into the `security-jwt` module. I agree that this special treatment is not necessary for JWT-based authentication.

https://github.com/opencast/opencast/blob/1b993fcc9294cfdb4b15f10dd5ebd4459f5716a1/modules/security-aai/src/main/java/org/opencastproject/security/aai/DynamicLoginHandler.java#L168-L207

PR #6975, opened 2025-08-07
Update Hamcrest dep

gregorydlogan:
This PR extends #6778 to include a required update to the assembly configs.  The parent PR does not boot without this change since the newer hamcrest build needs the newer caffeine version.  A better approach would be to have these be packed on their own, but outright removing the line in the assembly config omits the library completely, which also doesn't work.

If someone has a better way of doing this please let me know, but in the meantime this will work.

### How to test this patch

Turn on the graphql plugin and make sure things work

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


geichelberger (approved):
LGTM, GraphQL is still working.

PR #6980, opened 2025-08-13
Fix javadoc build

gregorydlogan:
This PR fixes the database javadoc build so that releases can happen.  Unclear what `checker-qual` is doing to break the build, but it was munging the javadocs options file, which lead to bad builds.  This workaround likely breaks whatever `checker-qual` is doing in our javadocs (if anything), but would not break the actual compiled code itself.

This should go into `r/17.x` since we would like to be able to push new builds into maven central.  This won't fix the previous builds, but new ones should work.

### How to test this patch

`./mvnw deploy`, which should fail with errors about missing classes in `org.osgi.framework` in the `db` module without this, and fail with HTTP 40(1|3) errors with this since you don't have permissions to deploy things (and the deploy is still broken regardless, that's next :)).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/db/pom.xml
+++ b/modules/db/pom.xml
@@ -98,6 +98,18 @@
         </configuration>
       </plugin>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
+          <detectJavaApiLink>true</detectJavaApiLink>
+          <maxmemory>512m</maxmemory>
+          <quiet>false</quiet>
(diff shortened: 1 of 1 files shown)


lkiesow:
> …so that releases can happen.

To clarify: This is just about Maven Central, correct? The regular builds are fine?
Also, this is not caught by the regular build tests? We do run `mvn site`, don't we?

gregorydlogan:
The regular builds work fine, running `javadoc:javadoc javadoc:aggregate`, as does `site` though we don't do that one in CI.  Happy to move to building the site rather than just the javadocs if you'd rather.

PR #6984, opened 2025-08-14
Fix index rebuild stuff (again)

KatrinIhler:
This was initially just trying to fix #3274 for events which was broken by #4120. But the longer I looked at the code the more I found. So this now does the following:

- Remove code duplication that was introduced by #4120 by merging the methods responsible for updating objects in the index for pretty much all relevant services
- As a result, the AssetManager now indexes extended metadata during the index rebuild again
- Inadvertently, this also fixes two other instances where the code had diverged:
  - The Workflow Service now also indexes the workflow definition during the index rebuild
  - A small patch by Lars from last year had only been applied to one method in the Comments Service
- In the Scheduler Service, I removed the request to the Asset Manager during the index rebuild. In my opinion, this is unnecessary since the Scheduler should only re-index the data it's responsible for and leave the rest to the Asset Manager. This should improve performance as we no longer need to read the catalogs from disk. However, this might be controversial.
-  Very minor code cleanup for stuff that was flagged by Intellij.

I did my best to provide a clear commit history for easier review. I'm uncertain what we should aim this at because it's a little more than a bug fix now. Opinions?

Also, short rant at the end:

In my opinion, #4120 _should not have been merged_ like this. The idea was good but the implementation is subpar. In addition to these issues (and the ones that have been already fixed in the meantime, like #5150 ), we no longer do locking during the index rebuild, which is not a problem when the system is idle, but could lead to lost updates when it is not. But I didn't want to spend more time than reasonable on this since things will most likely change because of the new data model anyway. Really, I just wanted to index extended metadata :sob: 

:warning: **I thought a great deal about this, but I didn't have time to test this yet before my vacation!**

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -534,65 +534,15 @@ public class AssetManagerImpl extends AbstractIndexProducer implements AssetMana
    *         The newest snapshot of the event to update
    */
   private void updateEventInIndex(Snapshot snapshot) {
-    final MediaPackage mp = snapshot.getMediaPackage();
-    String eventId = mp.getIdentifier().toString();
-    final String organization = securityService.getOrganization().getId();
+    final String eventId = snapshot.getMediaPackage().getIdentifier().toString();
+    final String orgId = securityService.getOrganization().getId();
     final User user = securityService.getUser();
+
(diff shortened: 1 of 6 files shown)


jduehring:
I haven't looked closely at the code, but I tested it locally with extended metadata. The metadata appears to be indexed correctly, and a text search through the UI returned the expected results.

wsmirnow (approved):
I had a closer look at it and tested it locally. Works for me. Thank you @KatrinIhler

gregorydlogan (approved):
This looks good to me, and works.  This gets us indexed extended metadata, plus it also cuts down on the duplication.

PR #6987, opened 2025-08-19
Fix code coverage tooling

gregorydlogan:
This PR replaces cobertura with jacoco.  Cobertura was EOL with JDK11, and jacoco seems to be the new hoteness.  I also switched from calling building javadocs directly to building the full maven site, which builds both the javadocs and the coverage reports.  Jacoco also has a `check` goal, which I've put into the pom but commented out.  This goal can be used to enforce per-module coverage levels.

Open issue:
- Aggregate project reports (ie, the toplevel one) doesn't pick up the individual module coverage levels for some reason.

I've filed this against 17, but I don't actually care if it's 17 or 18.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


.github/workflows/test-opencast-site.yml
```diff
@@ -0,0 +1,125 @@
...
+        env:
+          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+        run: |
+          install -dm 700 ~/.ssh/
+          echo "${DEPLOY_KEY}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+          ssh-keyscan github.com >> ~/.ssh/known_hosts
+
+        #This has to happen in a separate step or else various service classes aren't visible
+      - name: build maven site
```
gregorydlogan, line 71: Bringing this up tomorrow: This takes ~860MB *per branch*.  Thus this storage approach seems... suboptimal in terms of developer experience.  I'd like thoughts on where it *should* live.

PR #7032, opened 2025-09-02
Filter Live Streams in the Search Index

ciegler:
Similar to a previous PR (https://github.com/opencast/opencast/pull/6331) this adds a configurable option to filter out live events in the LTI series subtool (ltitools/index.html?subtool=series). 
However this PR also configures the search index in such a way that "live" is now a new and searchable field in the search index. This fixes the wrong event and page number which resulted in filtering out live events in the frontend only. With this live events can be filtered out when requesting the search index.

To configure whether or not live events should be filtered a config file was added in  etc/ui-config/mh_default_org/ltitools/ with the option to either `excludeLiveStreams` or to show `onlyLiveStreams`. Setting both values to true or false at the same time leads to not filtering the results at all. 

### How to test this patch

Testing the search index only can be done via the REST API Docs /search endpoint. Testing the filter in the LTI series tools frontend can be done by changing the values in etc/ui-config/mh_default_org/ltitools/config.json and reloading the LTI fronted.


modules/search-service-impl/src/main/resources/search-mapping.json
```diff
@@ -66,6 +66,9 @@
       "analyzer": "autocomplete_analyzer",
       "search_analyzer": "whitespace_analyzer"
     },
+    "live": {
+      "type": "keyword"
```
geichelberger: Shouldn't that be a boolean?

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -401,6 +408,10 @@ public Response getEpisodes(
       );
     }
 
+    if (StringUtils.isNotEmpty(live)) {
+      query.must(QueryBuilders.termQuery("live", live));
```
geichelberger:
```suggestion
      query.filter(QueryBuilders.termQuery("live", live));
```
This is simple binary logic, so we can use `filter` instead of `must`, which executes faster as no relevance score needs to be computed, and Elastic/OpenSearch can cache frequently used filters.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
```diff
@@ -339,7 +345,8 @@ public Response getEpisodes(
       @QueryParam("sort") String sort,
       @QueryParam("limit") String limit,
       @QueryParam("offset") String offset,
-      @QueryParam("sign") String sign
+      @QueryParam("sign") String sign,
+      @QueryParam("live") String live
```
geichelberger:
```suggestion
      @QueryParam("live") Boolean live
```

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
```diff
@@ -75,6 +76,8 @@ public class SearchResult {
 
   private String id = null;
 
+  private String live = null;
```
geichelberger:
```suggestion
  private Boolean live = null;
```

etc/ui-config/mh_default_org/ltitools/config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "excludeLiveStreams" : false,
+  "onlyLiveStreams" : false
+}
```
geichelberger: Newline is missing.

modules/lti/src/components/Series.tsx
```diff
@@ -131,20 +138,22 @@ class TranslatedSeries extends React.Component<SeriesProps, SeriesState> {
     loadCurrentPage(pageNumber: number = 1) {
         const qs = parsedQueryString();
 
+
```
geichelberger, line 140:
```suggestion
```

geichelberger (changes requested):

ciegler:
Thank you for the feedback!  
I applied all your requested changes in the commit "Applying change requests in PR".   

I also received the feedback that it would be nice to be able to filter the live events in the lti tools via the query parameters, like every other filter option in the lti tools.  I added this in the second commit "Adding filter via query parameters".  Furthermore it was pointed out that the config file itself could be confusing, since the live filter option is only available for the series subtool. To make that clearer I added the key "series" to the config and reorganized the live filter options as values of "series".  These changes are also just in the second commit.

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
```diff
@@ -252,7 +260,8 @@ public Map<String, Object> dehydrateEpisode() {
         DUBLINCORE, SearchResult.dehydrateDC(this.dublinCore),
         ORG, this.orgId,
         TYPE, this.type.name(),
-        MODIFIED_DATE, DateTimeFormatter.ISO_INSTANT.format(this.modified)));
+        MODIFIED_DATE, DateTimeFormatter.ISO_INSTANT.format(this.modified),
+        LIVE, this.live));
```
gregorydlogan, line 264: So this puts the live flag into the index, but there's no explicit handling of the flag on the other (`rehydrate()`) end.  Even if it's not required (since the queries are looking at the index directly), I think we should reconstruct the live flag on the other end too.

modules/search-service-impl/src/main/java/org/opencastproject/search/endpoint/SearchRestService.java
gregorydlogan: Yep.  You've changed the REST docs, but not the Java side of things.  It will still work since Boolean->String is less restrictive, but docs and parameters should match.

modules/lti/src/components/Series.tsx
```diff
@@ -130,21 +137,24 @@ class TranslatedSeries extends React.Component<SeriesProps, SeriesState> {
 
     loadCurrentPage(pageNumber: number = 1) {
         const qs = parsedQueryString();
+        console.log(qs);
```
gregorydlogan:
```suggestion
```

gregorydlogan (changes requested):
Code works as is, very minor nits I'd like to see fixed.  Should make the 19.x branch.

modules/lti/src/components/Series.tsx
ciegler: I deleted the console log, sorry for not checking before pushing.

modules/search-service-api/src/main/java/org/opencastproject/search/api/SearchResult.java
ciegler: Maybe I did not fully understand the rehydrate method in the SearchResult. As far as I understood is that `rehydrate()` initializes a new SearchResult() if it finds the corresponding ACLs, DC and mediapackage. Since `live` is depending on the mediapackage and also set when initializing a new SearchResult() I assumed that the case was already covered.

PR #7040, opened 2025-09-08
Make speech to text fully asynchronous

### LINKED ISSUE #6967: Have the option to run Whisper fully asynchronously
The current speech-to-text woh async=true configuration makes it possible to continue the workflow and attach the captions later in the same workflow via the speech-to-text-attach woh by setting a workflow configuration with the job ids and waiting for them. This does not save a lot of time.

What we would like to have is to be able to attach the captions asynchronously using another workflow so that we can publish the recording to students without captions and not "lock" the recording until the captions are done i.e. we would like to be able to run another workflow on it if necessary. We are running whisper-ctranslate2 on CPUs and a 3-hour class takes 30-50 minutes to caption.

I'm currently working on this. This is how I'm planning to implement:

- Maintain backwards compatibility with the current async mode.
- But also add a db control table that keeps track of status and a quartz job (on admin) that will check that table and start a workflow if captions running in async mode are done.

rute-santos:
Whisper can now run fully asynchronously i.e. a workflow starts the subtitle generation and finishes execution; the generated subtitles are attached later by running another workflow.

This is useful when we want to publish the recording quickly without having to wait for the subtitles, and attach them later.

Changes:

- New service that runs on the admin node and tracks those jobs (new db table).
- Make sure input files and arguments are not deleted at the end of the starting workflow.
- Add quartz job to schedule workflow to attach subtitles when all STT jobs have finished.

Closes #6967.

### How to test this patch

1 - Create a workflow to attach the subtitles. Operation example:

```
- id: speechtotext-attach
  description: Attach subtitles
  configurations:
    - target-flavor: captions/delivery
    - target-element: track
    - target-tags: engage-download

- id: publish-engage-aws
  description: Publish media package
  configurations:
    - download-source-flavors: captions/delivery,security/*
    - strategy: merge
```

2 - Add it to the quartz job configuration in etc/org.opencastproject.speechtotext.async.impl.SpeechToTextWorkflowSchedulerQuartz

3 - Run a workflow that starts Whisper transcriptios after the media package is published. Operation example:

```
- id: speechtotext
  description: Starts Whisper transcription
  configurations:
    - source-flavor: '*/delivery'
    - target-flavor: captions/delivery
    - limit-to-one: true
    - track-selection-strategy: try_presenter_first
    - async: true
```

4 - Verify that the subtitles are attached to the media package after the second workflow ran.


modules/speech-to-text-async-impl/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+      <groupId>org.opencastproject</groupId>
+      <artifactId>opencast-kernel</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <!-- Persistence -->
+    <dependency>
+      <groupId>org.eclipse.persistence</groupId>
+      <artifactId>jakarta.persistence</artifactId>
+    </dependency>
+    <!-- dependency>
```
lkiesow: This seems like an accidental left over?

rute-santos: Yes, removing...

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Quartz job to start workflows to asynchronously attach captions generated by the SpeechToText service.
+ */
+@Component(property = { "service.description=Speech to Text Workflow Scheduler" }, immediate = true, service = {
+        ManagedService.class })
+public class SpeechToTextWorkflowSchedulerQuartz extends AbstractScanner implements ManagedService {
```
gregorydlogan, line 83:
```suggestion
public class SpeechToTextWorkflowSchedulerQuartz extends AbstractScanner {
```

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+      job.setVolatility(true);
+      job.getJobDataMap().put(JOB_PARAM_PARENT, this);
+      quartz.addJob(job, true);
+    } catch (org.quartz.SchedulerException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void updated(Dictionary<String, ?> properties) throws ConfigurationException {
```
gregorydlogan, line 148: Because we're not doing `ManagedService`
```suggestion
  @Modified
  public void modified(Map<String, Object> properties) {
```

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+import org.opencastproject.util.data.Option;
+import org.opencastproject.workflow.api.ConfiguredWorkflow;
+import org.opencastproject.workflow.api.WorkflowDatabaseException;
+import org.opencastproject.workflow.api.WorkflowDefinition;
+import org.opencastproject.workflow.api.WorkflowInstance;
+import org.opencastproject.workflow.api.WorkflowService;
+
+import org.apache.commons.lang3.StringUtils;
+import org.osgi.service.cm.ConfigurationException;
+import org.osgi.service.cm.ManagedService;
```
gregorydlogan, line 48:
```suggestion
```

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextAsyncTrackerImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import org.opencastproject.job.jpa.JpaJob;
+import org.opencastproject.speechtotext.async.api.SpeechToTextAsyncException;
+import org.opencastproject.speechtotext.async.api.SpeechToTextAsyncTracker;
+import org.opencastproject.speechtotext.async.persistence.SpeechToTextControl;
+import org.opencastproject.speechtotext.async.persistence.SpeechToTextDatabase;
+
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+@Component(property = { "service.description=Speech to Text Async Tracker" }, immediate = true, service = {
```
gregorydlogan: Can we reformat this so it reads nicer?

assemblies/karaf-features/src/main/feature/feature.xml
gregorydlogan, line 1: You've added both an API and an impl bundle here, which isn't *wrong*, but are we likely to see a different implementation of this API at any point?  I would be tempted to collapse them together.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
gregorydlogan: `ManagedService` should be avoided if possible in favour of declarative services

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+      }
+
+      String cronExpression = (String) properties.get(PARAM_KEY_CRON_EXPR);
+      if (StringUtils.isBlank(cronExpression) || !CronExpression.isValidExpression(cronExpression)) {
+        throw new ConfigurationException(PARAM_KEY_CRON_EXPR, "Cron expression must be valid");
+      }
+      setCronExpression(cronExpression);
+      logger.info("{} : {}", PARAM_KEY_CRON_EXPR, cronExpression);
+
+      attachWorkflowDef = (String) properties.get(WORKFLOW);
```
gregorydlogan, line 169: So if this key is unset we just roll with it?  What's the usecase for *not* having a workflow ID set?

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+  }
+
+  /**
+   * Start a workflow to attach subtitles for all STTs that have finished (one per workflow id that created them).
+   *
+   * @throws SpeechToTextAsyncException
+   */
+  void handleTranscriptionFinished() throws SpeechToTextAsyncException {
+    if (attachWorkflowDef == null) {
```
gregorydlogan, line 305: If we enforce this being set during `activate` we don't need to check it here.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,403 @@
...
+      Workflows workflows = wfUtil != null ? wfUtil : new Workflows(assetManager, workflowService);
+      Set<String> mpIds = Collections.singleton(mpId);
+      Map<String, String> wfConfig = new HashMap<String, String>();
+      wfConfig.put(JOBS_WORKFLOW_CONFIGURATION,
+              jobs.stream().map(j -> String.valueOf(j.getId())).collect(Collectors.joining(",")));
+      List<WorkflowInstance> wfList = workflows
+              .applyWorkflowToLatestVersion(mpIds, ConfiguredWorkflow.workflow(wfDef, wfConfig));
+      WorkflowInstance wf = wfList.size() > 0 ? wfList.get(0) : null;
+      logger.info("Workflow {} {}started for mp {}{}", attachWorkflowDef, wf != null ? "" : "NOT ", mpId,
+              wf != null ? ": " + wf.getId() : "");
```
gregorydlogan: While this is more verbose, it makes for *way* easier reading
```suggestion
      if (wf != null) {
        logger.info("Workflow {} started for mp {}: {}", attachWorkflowDef, mpId, wf.getId());
      } else {
        logger.info("Workflow {} NOT started for mp {}", attachWorkflowDef, mpId);
      }
```

gregorydlogan (changes requested):
A few issues, the biggest one being that the unit tests are failing for me:

```
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.169 s <<< FAILURE! -- in org.opencastproject.speechtotext.impl.Speec
hToTextServiceImplTest                                                                                                                        
[ERROR] org.opencastproject.speechtotext.impl.SpeechToTextServiceImplTest.testVoskSpeechToText -- Time elapsed: 0.080 s <<< FAILURE!          
java.lang.AssertionError:                                                                                                                     
                                                                                                                                              
  Unexpected method call EasyMock for interface org.opencastproject.serviceregistry.api.ServiceRegistry -> ServiceRegistry.getCurrentJob()    
        at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:46)                                                  
        at org.easymock.internal.ObjectMethodsFilter.invoke(ObjectMethodsFilter.java:100)                                                     
        at jdk.proxy2/jdk.proxy2.$Proxy15.getCurrentJob(Unknown Source)                                                                       
        at org.opencastproject.speechtotext.impl.SpeechToTextServiceImpl.transcribe(SpeechToTextServiceImpl.java:189)                         
        at org.opencastproject.speechtotext.impl.SpeechToTextServiceImplTest.testVoskSpeechToText(SpeechToTextServiceImplTest.java:152)
[...]
```

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: I tried and I wasn't successful with that. All other classes that extend AbstractScanner also implement ManagedService so maybe this is a change that needs to be done to all at once?

assemblies/karaf-features/src/main/feature/feature.xml
rute-santos: I always separate them... but please let me know if you feel strongly about this and I will change :)

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: Fixed.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: Fixed.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
gregorydlogan: Ok, fair. Didn't catch that. It's in my list to remove ManagedService, so let's ignore this and I'll deal with it later.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,494 @@
...
+    logger.info("Updating!");
+    unschedule();
+
+    if (properties != null) {
+      logger.debug("Updating configuration...");
+
+      setEnabled(properties.get(PARAM_KEY_ENABLED) != null
+          && Boolean.valueOf(properties.get(PARAM_KEY_ENABLED).toString()));
+      if (!isEnabled()) {
+        logger.info("Speech to text workflow scheduler quartz job is disabled");
```
gregorydlogan, line 158: Why would you ever want this to be disabled?  Allowing the user to modify this seems like it can only lead to pain.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,494 @@
...
+      }
+      logger.info("Attach workflow definition is {}", attachWorkflowDef);
+
+      abandonAfterMs = DEFAULT_ABANDON_AFTER_SECS * 1000;
+      if (properties.get(ABANDON_AFTER_SECS) != null) {
+        try {
+          abandonAfterMs = Integer.parseInt((String) properties.get(ABANDON_AFTER_SECS)) * 1000;
+        } catch (Exception e) {
+          // Log and use default
+          logger.warn("Invalid configuration for {}: '{}'", ABANDON_AFTER_SECS, properties.get(ABANDON_AFTER_SECS), e);
```
gregorydlogan: Should list the fallback default as well.

```suggestion
          logger.warn("Invalid configuration for {}: '{}', defaulting to '{}'", ABANDON_AFTER_SECS, properties.get(ABANDON_AFTER_SECS), DEFAULT_ABANDON_AFTER_SECSe);
```

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
```diff
@@ -0,0 +1,494 @@
...
+
+    try {
+      // STTs that are stuck in progress for a long time will be transitioned to error so that they will be retried.
+      database.transitionStatusByDate(SpeechToTextControl.Status.TranscriptionError, nowMinusInterval,
+              SpeechToTextControl.Status.InProgress);
+
+      // STTs that had a workflow to attach subtitles started are transitioned to done because afaik everything worked
+      // as supposed to
+      database.transitionStatusByDate(SpeechToTextControl.Status.Done, nowMinusInterval,
+              SpeechToTextControl.Status.WorkflowInProgress);
```
gregorydlogan, line 290: The count here should be logged.  Ideally we'd log the individual control/mediapackage/workflow ID at debug too, but looking at the queries I'm not sure how reasonable that is.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/endpoint/SpeechToTextServiceRestEndpoint.java
```diff
@@ -102,7 +102,9 @@ public class SpeechToTextServiceRestEndpoint extends AbstractJobProducerEndpoint
           @RestParameter(name = "language", isRequired = false, type = STRING,
               description = "Language of the media file."),
           @RestParameter(name = "translate", isRequired = false, type = BOOLEAN,
-              description = "Enable translation to english") },
+              description = "Enable translation to english"),
+          @RestParameter(name = "async", isRequired = false, type = BOOLEAN,
+              description = "If job will run outside the scope of a workflow") },
```
gregorydlogan, line 107: Include the default (false?) too please.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
```diff
@@ -388,11 +394,31 @@ private void createSubtitleAsync(WorkflowInstance workflow, List<Track> tracks,
...
+      }
       for (var track: tracks) {
-        var job = speechToTextService.transcribe(track.getURI(), languageCode, translate);
+        // Make a copy of the track so that it isn't deleted by cleanup at the end of the workflow
+        String path = track.getURI().getPath();
+        String extension = path.lastIndexOf(".") > -1 ? path.substring(path.lastIndexOf(".")) : "";
+        String fileName = String.format("%d-%s%s", workflow.getId(), track.getIdentifier(), extension);
+        URI inputUri = workspace.putInCollection(STT_ASYNC_COLLECTION, fileName, workspace.read(track.getURI()));
+
+        var job = speechToTextService.transcribe(inputUri, languageCode, translate, true);
```
gregorydlogan, line 411: We default to doing this async, but we don't have any changes to the workflows to doing the actual attach?  Or am I misunderstanding something?

gregorydlogan (changes requested):
This needs a release notes entry so it gets picked up on by the RMs as well.

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: When Whisper is run synchronously, which I believe is the majority of cases, you don't need a quartz job that does nothing running all the time.

modules/speech-to-text-workflowoperation/src/main/java/org/opencastproject/workflow/handler/speechtotext/SpeechToTextWorkflowOperationHandler.java
rute-santos: There are 2 ways to run it asynchronously:

1. Have the speechtotext woh and speechtotext-attach woh in the same workflow - this is how it was implemented before and was kept as is. The speechtotext-attach woh in the same workflow will wait until subtitles are ready.
2. Have the speechtotext woh in one workflow and the speechtotext-attach in another, started by the quartz scheduler action when the subtitles are ready - this is what this PR implements.

I didn't provide a sample workflow for that; I only updated the documentation (whisper.md) as it seems there's no sample workflow that uses speechtotext? Please let me know if this is not enough!

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: Added debug logging with counter.

modules/speech-to-text-impl/src/main/java/org/opencastproject/speechtotext/impl/endpoint/SpeechToTextServiceRestEndpoint.java
rute-santos: Done!

modules/speech-to-text-async-impl/src/main/java/org/opencastproject/speechtotext/async/impl/SpeechToTextWorkflowSchedulerQuartz.java
rute-santos: Done!

gregorydlogan (approved):

PR #7041, opened 2025-09-08
Add default ACL template config key

Arnei:
Adds the config key used by the following frontend PR: https://github.com/opencast/opencast-admin-interface/pull/1429.

As the default for the config key is "None", this will not break frontend.

Update 12.09: Now also adds a new admin ui endpoint to fetch templates by name. Still will not break the frontend.

### How to test this patch

Just a config key, not really much to test here. Better test https://github.com/opencast/opencast-admin-interface/pull/1429

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


mtneug:
I'm not a fan of using dynamic IDs from the DB in configuration files. This makes automation hard and requires a second config roleout after cluster initialisation. Can't you use the ACL `name`, e.g. `authenticated` or `public`? These are (somewhat) stable as they can be set before starting the cluster.

etc/listproviders/acl.default.create.properties
```diff
@@ -22,6 +22,11 @@ list.name=ACL.DEFAULTS
 # Default: No actions
 #default_actions=annotate,annotate-admin
 
+# Template that will automatically be applied to new events/series
```
lkiesow: This definitely needs a better explanation. You probably didn't notice since you know what this is about, but this really confused me and I only kind of know what you are talking about since you are referencing the ACL endpoint.

What type of template are you talking about? It's an ACL template? But the endpoint doesn't actually returns a template. It returns pre-defined ACL.

So, this should probably be the default ACL config key.

The same also goes for at least the title of [the corresponding admin interface pull request](https://github.com/opencast/opencast-admin-interface/pull/1429). Though it's less confusing than here since you have someexplaination.

Arnei: Thanks for pointing that out, will try and improve the description.

Arnei:
> I'm not a fan of using dynamic IDs from the DB in configuration files. This makes automation hard and requires a second config roleout after cluster initialisation. Can't you use the ACL `name`, e.g. `authenticated` or `public`? These are (somewhat) stable as they can be set before starting the cluster.

I've been going with IDs because that is what our endpoints use. If names are preferred, I would add an endpoint that uses names instead of IDs and then use that. Does that sound good?

mtneug:
> I've been going with IDs because that is what our endpoints use. If names are preferred, I would add an endpoint that uses names instead of IDs and then use that. Does that sound good?

Sounds good to me.

PR #7042, opened 2025-09-08
Fix 17.x site build

gregorydlogan:
The site plugin requires install to have run, lest it be unable to find
artifacts.  This makes sense since install is what actually puts them
into your repo.  The catch here is that install + site in the same build
hits issues with checkerqual, which conflicts with site.  The solution
is a two pass approach, which avoids the conflict.

We deliberately disabled the first build-only test pass since we don't care
about the results of the tests *and* we still run the tests in the site
pass.  If we don't build the tests during the site pass we don't get the
code coverage metrics that we're after with the site build in the first
place.

This should go into 17 since this has broken the build in 17, and having
coverage data for 17 would be useful.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/.github/workflows/test-opencast-site.yml
+++ b/.github/workflows/test-opencast-site.yml
@@ -67,17 +67,27 @@ jobs:
         run: |
           .github/check-jacoco-deps.sh
 
+      - name: build opencast, skipping tests which run during site
+        run: |
+          ./mvnw install -DskipTests -T1C -Pnone \
+            --batch-mode \
+            -Dsurefire.rerunFailingTestsCount=2 \
+            -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn \
+            -Dhttp.keepAlive=false \
+            -Dmaven.wagon.http.pool=false \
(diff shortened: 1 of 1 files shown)


gregorydlogan:
The checkerqual stuff I was talking about was fixed in https://github.com/opencast/opencast/commit/e9e1f29862e8b33bb2f1bf9105902d4b5b1e0d62

.github/workflows/test-opencast-site.yml
```diff
@@ -67,17 +67,27 @@ jobs:
         run: |
           .github/check-jacoco-deps.sh
 
+      - name: build opencast, skipping tests which run during site
+        run: |
+          ./mvnw install -DskipTests -T1C -Pnone \
+            --batch-mode \
+            -Dsurefire.rerunFailingTestsCount=2 \
```
lkiesow: We don't actually need this since we don't run any tests, do we?
We could skip the style checker, though.
```suggestion
            -Dcheckstyle.skip=true \
```

gregorydlogan: Yep, great catch

lkiesow (approved):

PR #7052, opened 2025-09-16
Replace java-jwt with nimbus-jose-jwt

### LINKED ISSUE #6093: Support `EdDSA` JWT algorithm
Just a quick feature request so that this isn't forgotten.

Arguably the best algorithm for JWTs to use is `EdDSA`. It has very strong security and is fast for signing and verifying. Compare [this blog post](https://www.scottbrady91.com/jose/jwts-which-signing-algorithm-should-i-use) for example. Opencast does not support this, unfortunately. Even more unfortunately, the library we use `java-jwt` by auth0 also does not support it yet, as that library targets Java 8 or something, which crypto module does not support the relevant crypto primitives yet. Newer Java versions absolutely do. 

So either we wait until `java-jwt` gets around to implementing that or we switch to a different library. Given that this issue is not super high priority, I suspect the former is gonna happen.

Arnei:
Fix #6093.

The goal of this commit is to introduce a jwt library capable of handling EdDSA signed jwts. java-jwt does not do this and it is currently unclear when or even if it ever will be able to. nimbus-jose-jwt has been chosen because it:
- supports EdDSA (through google-crypto-tink)
- has a long-standing history
- is actively developed
- is easy to use

### How to test this patch

Ideally you already have a jwt setup, so you can test if it still works. Be aware that the configuration has changed slightly (no functional changes, but one variable has been renamed and explicit string transformations are now unnecessary).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


etc/security/mh_default_org.xml
```diff
@@ -780,7 +780,18 @@
...
+    <value>HS512</value>
     <value>RS256</value>
+    <value>RS384</value>
+    <value>RS512</value>
+    <value>ES256</value>
+    <value>ES256K</value>
+    <value>ES384</value>
+    <value>ES512</value>
+    <value>EdDSA</value>
+    <value>Ed25519</value>
```
LukasKalbertodt: Why are `EdDSA` and `Ed25519` valid values here? `Ed25519` is a subset of `EdDSA`, or rather one specific curve. `EdDSA` just describes all signature schemes with twisted edwards curves, which also includes `Ed448` (another specific curve, hardly supported). I do get that it might make sense to explicitly want to allow only a single specific curve or to generally allow a family, but given that only `EdDSA` is a valid `alg` value, this might confuse people. 

I would probably just use `EdDSA` and not give users the option to only specify a single curve. This configuration is weird anyway, as there is little benefit to restricting algorithms like this, as far as I know.

etc/security/mh_default_org.xml
```diff
@@ -765,13 +765,13 @@
     <property name="securityService" ref="securityService" />
     <property name="userReferenceProvider" ref="userReferenceProvider" />
     <property name="jwksUrl" value="https://auth.example.org/.well-known/jwks.json" />
-    <property name="jwksCacheExpiresIn" value="1440" />
+    <property name="jwksTimeToLive" value="3600000" />
```
LukasKalbertodt, line 768: Why the unrelated change of default duration? 

And as a different issue: this seems like a breaking change that's not really required. We could keep the old name and unit (seconds) in this file to make updating easier. Probably not super important because there are other breaking changes that we cannot avoid, but still: fewer breaking changes are nice when there is no good reason to make htem.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
```diff
@@ -21,55 +21,73 @@
...
+
+    if (alg.equals(JWSAlgorithm.RS256) || alg.equals(JWSAlgorithm.RS384) || alg.equals(JWSAlgorithm.RS512)) {
+      RSAKey rsaKey = jwkSet.get(0).toRSAKey();
+      return verify(jwt, claimConstraints, new RSASSAVerifier(rsaKey));
+    } else if (alg.equals(JWSAlgorithm.ES256) || alg.equals(JWSAlgorithm.ES256K) || alg.equals(JWSAlgorithm.ES384)
+        || alg.equals(JWSAlgorithm.ES512)) {
+      ECKey ecKey = jwkSet.get(0).toECKey();
+      return verify(jwt, claimConstraints, new ECDSAVerifier(ecKey));
+    } else if (alg.equals(JWSAlgorithm.EdDSA) || alg.equals(JWSAlgorithm.Ed25519)) {
+      JWK jwk = jwkSet.get(0).toPublicJWK();
```
LukasKalbertodt: All this code only tries the first key of the JWKS? Am I reading this correctly? Why is that? Was that the previous behavior as well?

LukasKalbertodt (commented):
I haven't really looked at the code, but I tested this. And it works with Tobira! The old ES256 key still works, but Ed25519 now works too.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
mtneug: @joneug do you know what the previous behavior was? If we only supported the first entry, IMO this would be a bug and we should support JWKS with multiple entries.

joneug: In the current implementation, all keys are fetched from the provider.

https://github.com/opencast/opencast/blob/8046ae9b7fd3a67bc9581a308e3710c95cebcabd/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java#L63-L73

The verification logic then considers all of them:

https://github.com/opencast/opencast/blob/8046ae9b7fd3a67bc9581a308e3710c95cebcabd/modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java#L126-L154

etc/security/mh_default_org.xml
Arnei: Both are valid values because I did not think about this too much. I simply saw that the nimbus library offered both values as alg constant, so I added them. I'll change it to just `EdDSA` then.

etc/security/mh_default_org.xml
Arnei: I changed this because
- Internally this was previously minutes, but is now milliseconds.
- An hour (3600000 ms) was recommended by the nimbus-jose-jwt docs as a reasonable default.

I have no hard opinion on this however. If you think it is better, I'll change it back to the old value (a day) and parse it internally from minutes to milliseconds.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
Arnei: Thanks for pointing that out, should be fixed now.

PR #7060, opened 2025-09-18
Allow `mvnw` to work with symlinks

lkiesow:
We have `mvnw` to make running Maven easier. When compiling just one module, it is somewhat annoying to type `../../mvnw` however. To solve that, just create a symlink… except that doesn't work since the script then thinks the symlink directory is the project's base directory.

That is fixed by this patch. It resolves symlink and will use the actual path to the `mvnw` binary to determine the project's base path.

### How to test this patch

- Create a symlink: `ln -s /path/to/mvnw`
- Make sure the link is in the `PATH`
- Run `mvnw`

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/mvnw
+++ b/mvnw
@@ -192,7 +192,7 @@ log() {
   fi
 }
 
-BASE_DIR=$(find_maven_basedir "$(dirname "$0")")
+BASE_DIR=$(find_maven_basedir "$(dirname "$(readlink -f "$0")")")
 if [ -z "$BASE_DIR" ]; then
   exit 1
 fi


mtneug:
I'm not quite following. Where do you actually suggest put the symlinks? Within each module folder? Always or only if you are working on it? If you have to type out the commands to create the symlink, wouldn't `../../mvnw` be simpler?

To be clear: no complains about the patch. If you think it's useful, let's merge it.

lkiesow:
This should explain what I want, though I wouldn't suggest `/bin`. But since it is in the `PATH` by default, let's just use that:

```sh
cd /bin
ln -s /path/to/opencast/mvnw

# Now you can still use mvnw in the main project directory
cd /path/to/opencast
mvnw clean install

# But this also works in a module directory with the exact same command
cd /path/to/opencast/modules/admin-ui
mvnw clean install
```

Without this patch, this would have failed. Sure, you could previously call `../../mvnw` in the module directory, but that's annoying.

PR #7068, opened 2025-09-26
Replace custom classes with standard library

Arnei:
Removes various classes from common/src/main/java/org/opencastproject/util. These classes were implementing concepts that by now are well represented in the Java standard libraries. As such, this pull request refactors the code that was using those classes with classes from the standard libraries.

The goal here is to make our code easier to understand. Developers should not need to remember different names for classes and methods that are functionally the same, or have to worry about if one of our classes is ever so slightly different from their standard library counterpart.

This is only a refactoring and should result in no functional changes.

### How to test this patch

This touches quite a few files, so I don't think there is a feasible way to target test this. Ideally someone can install this on their test system (that they actually use) and report after a week that nothing has broken.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan:
I don't see a re-request button @lkiesow, for likely obvious reasons.  Wanna try again?

gregorydlogan (approved):
This looks sane, so let's merge it in so people can actually for real test this.

PR #7108, opened 2025-10-08
improve docs for execute-once WOH and execute service

reiferschris:
This PR improves the docs for the execute-once WOH. It now mentions the existence of  `#tags` for parameter substitution. In the execute service docs it adds the missing `s` to the `tags` because the code actually matches that. See https://github.com/opencast/opencast/blob/develop/modules/execute-impl/src/main/java/org/opencastproject/execute/impl/ExecuteServiceImpl.java#L369

### How to test this patch

Use the execute-once WOH e.g. with `#{tags(engage-streaming)}` to select the first track that matches.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


mtneug (approved):

mtneug:
@reiferschris is it correct that you didn't file an ICLA?

CC @lkiesow

reiferschris:
I did file it now :)

gregorydlogan:
Can we rebase this to 17?  I'd imagine it also affects older versions as well.

reiferschris:
> Can we rebase this to 17? I'd imagine it also affects older versions as well.

It certainly affects all versions from 13.x on. 
Is this a question for me or a very polite way of asking me to do this?
If so I'd appreciate some guidance on how I would have to do this properly. 

I had my branch `patch-1` checked out and attempted `git rebase r/17.x` but this leaves me with two commits that appear to be from `develop` between my commit and the actual head of `r/17.x`

gregorydlogan:
Yep, you're on the right path.

You want `git rebase -i`, which will drop you to a text editor with a list of commit's. Remove the commit's you don't want and then it will do what you're looking for!

reiferschris:
Thanks for the hint Greg, but unfortunately this doesn't seem to work for me. I tried `git rebase -i r/17.x` but I end up with 500+ changes marked with pick. when I try to apply this I get tons of merge conflicts. I tried resolving them manually but gave up after 25+ or so, that's way to tedious.
Can we either just merge this to develop or if it's really necessary I can try another pull request.
I doubt that anyone but me noticed this documentation error in the last years anyways.

mtneug:
Let's fix it in `develop`. If we want, we can always cherry-pick it to earlier versions.

PR #7113, opened 2025-10-09
Re-Introduce Event Handlers to AssetManager

KatrinIhler:
#6689 (specifically bb33272) removed the event handlers from the AssetManager, but we still need those to trigger certain changes when taking snapshots or deleting episodes, e.g. for the live schedule service. So I re-introduced them.

This also restores the index update when deleting events.

### DIFF
--- a/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
+++ b/modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
@@ -66,6 +66,7 @@ import org.opencastproject.mediapackage.MediaPackageElements;
 import org.opencastproject.mediapackage.MediaPackageParser;
 import org.opencastproject.mediapackage.MediaPackageSupport;
 import org.opencastproject.message.broker.api.assetmanager.AssetManagerItem;
+import org.opencastproject.message.broker.api.update.AssetManagerUpdateHandler;
 import org.opencastproject.metadata.dublincore.DublinCores;
 import org.opencastproject.metadata.dublincore.EventCatalogUIAdapter;
 import org.opencastproject.security.api.AccessControlEntry;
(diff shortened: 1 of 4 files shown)


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1543,6 +1569,18 @@ public static Snapshot rewriteUris(Snapshot snapshot, Function<MediaPackageEleme
             mpCopy);
   }
 
+  public void fireEventHandlers(AssetManagerItem item) {
+    while (handlers.size() != 2) {
+      logger.warn("Expecting 2 handlers, but {} are registered.  Waiting 10s then retrying...", handlers.size());
+      try {
+        Thread.sleep(10000L);
+      } catch (InterruptedException e) { /* swallow this, nothing to do */ }
```
karendolan, line 1579: It would add clarity to move "2" and "10000L" into constants to clarify their purpose
```
public static final int EXPECTED_HANDLER_COUNT = 2;
public static final long HANDLER_EXECUTE_INTERVAL = 10000L; // 10 seconds
```

gregorydlogan:
@KatrinIhler I'm going to cut 19 soonish, but IMO this should be in 19.  Can you rebase and address the review comments here when you get a chance?

KatrinIhler:
> @KatrinIhler I'm going to cut 19 soonish, but IMO this should be in 19. Can you rebase and address the review comments here when you get a chance?

Done!

modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
KatrinIhler: I created one for the handler count, but imo the thread interval is too unimportant to have its own constant, so I left it as is.

mtneug:
> Done!

This is not yet pointing to `r/19.x`

KatrinIhler:
> This is not yet pointing to `r/19.x`

Whoops, did the rebase, but forgot to change the target. Thank you!

Issue #6992, opened 2025-08-20
Extended Capture Agent API


JulianKniephoff:
On the Roadmap: https://github.com/opencast/roadmap/issues/25:

> The idea for this extension is to provide an optional set of capture agent metadata which is sent to agents with each event and which options have a clearly specified meaning.
This project is meant to define and implement the API (how the metadata will end up in the scheduled events API), define a set of metadata values and their meaning and implement a sensible way to configure or automatically set these values.

Now, I wasn't involved in the initial planning of this project; @lkiesow, @KatrinIhler please correct me if I misinterpreted anything, or add anything you deem important!

Anyway, this is my understanding of the project, with a few more technical details for this more developer-centric context, and so that whoever ends up having to do this actually knows **what** they have to do:

The goal of this project is to define/standardize/specify some of the properties and capabilities that are communicated between a capture agent and Opencast, so that Opencast can know and subsequently make use of/control some of the (presumably) more modern features of CAs. This has to happen in a backwards compatible way.

As far as I understood, this will mostly touch two places in [“the protocol”](https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/):

- The [`/calendar`-endpoint](https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/#getting-the-calendarschedule), which can carry [arbitrary data in an (ical-)attachment](https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/#capture-agent-configuration-file) (there is also a corresponding `json` endpoint we will have to adapt as well, I presume, and it has a corresponding key for that kind of data). This is what CAs query to get information about the scheduled events, and it could contain certain directives to control the behavior of the CA during the recording/streaming of each event.
- The [`/configuration`-endpoint](https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/#setting-agent-capabilities) which the CAs can call with an arbitrary JSON object to communicate their capabilities to Opencast. This is already used to communicate the available inputs, and can similarly be used to **constrain** the new things we want to pass back to the CA in the endpoints above. (See also below for an example how these interact.)

What new properties/capabilities we want is still up for debate, and moderating this debate will be part of this project, but here is a candidate set that seems pretty ... set, shamelessly stolen from a note from @lkiesow :

> - `record`
>     - `0`: Do not record
>     - `1` (default): Record and upload
>     - May be set to `0` only if `stream` is set to `1`
> - `stream`
[...]

PR #7114, opened 2025-10-13
Extended capture agent api

Arnei:
Basic demonstration implementation for #6992.

Show all places in the code that need to be modified in order to allow for new capture agent capabilties in Opencast. Implements the properties/capabilties that #6992 describes as "set":

> - `record`
>     - `0`: Do not record
>     - `1` (default): Record and upload
>     - May be set to `0` only if `stream` is set to `1`
> - `stream`
>     - `0` (default): Do not send a live stream
>     - `1`: Stream the event
> - `layout`
>     - Alphanumeric string specifying a layout
> - `camera-position`
>     - Alphanumeric string specifying a camera position

Related Admin UI PR: https://github.com/opencast/opencast-admin-interface/pull/1565

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
@@ -304,6 +304,16 @@ public class CaptureAgentsEndpoint {
         parsedCapabilities.add("stream", generateJsonDevice(parsedStream, TRANSLATION_KEY_PREFIX + "STREAM."));
       }
 
+      String record = (String) agent.getCapabilities().get(CaptureParameters.CAPTURE_DEVICE_RECORD);
+      if (record == null || record.isEmpty()) {
+        parsedCapabilities.add("stream", new JsonArray());
+      } else {
+        String[] parsedRecord = Arrays.stream(record.split(","))
+            .map(String::trim)
+            .toArray(String[]::new);
+        parsedCapabilities.add("record", generateJsonDevice(parsedRecord, TRANSLATION_KEY_PREFIX + "RECORD."));
+      }
+
       json.add("parsedCapabilities", parsedCapabilities);
     }
 
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
@@ -59,6 +59,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -131,10 +132,10 @@ public class CaptureAgentsEndpoint {
           @RestParameter(name = "filter", isRequired = false, description = "The filter used for the query. They should be formated like that: 'filter1:value1,filter2:value2'", type = STRING),
           @RestParameter(defaultValue = "100", description = "The maximum number of items to return per page.", isRequired = false, name = "limit", type = RestParameter.Type.STRING),
           @RestParameter(defaultValue = "0", description = "The page number.", isRequired = false, name = "offset", type = RestParameter.Type.STRING),
-          @RestParameter(defaultValue = "false", description = "Define if the inputs should or not returned with the capture agent.", isRequired = false, name = "inputs", type = RestParameter.Type.BOOLEAN),
+          @RestParameter(defaultValue = "false", description = "Define if the parsed capabilities should or not returned with the capture agent.", isRequired = false, name = "withParsedCapabilities", type = RestParameter.Type.BOOLEAN),
           @RestParameter(name = "sort", isRequired = false, description = "The sort order. May include any of the following: STATUS, NAME OR LAST_UPDATED.  Add '_DESC' to reverse the sort order (e.g. STATUS_DESC).", type = STRING) }, responses = { @RestResponse(description = "An XML representation of the agent capabilities", responseCode = HttpServletResponse.SC_OK) }, returnDescription = "")
   public Response getAgents(@QueryParam("limit") int limit, @QueryParam("offset") int offset,
-          @QueryParam("inputs") boolean inputs, @QueryParam("filter") String filter, @QueryParam("sort") String sort) {
+        @QueryParam("withParsedCapabilities") boolean withParsedCapabilities, @QueryParam("filter") String filter, @QueryParam("sort") String sort) {
     Optional<String> filterName = Optional.empty();
     Optional<String> filterStatus = Optional.empty();
     Optional<Long> filterLastUpdated = Optional.empty();
@@ -211,7 +212,7 @@ public class CaptureAgentsEndpoint {
     // Run through and build a map of updates (rather than states)
     List<JsonObject> agentsJSON = new ArrayList<>();
     for (Agent agent : filteredAgents) {
-      agentsJSON.add(generateJsonAgent(agent, inputs, false));
+      agentsJSON.add(generateJsonAgent(agent, withParsedCapabilities, false));
     }
 
     return okJsonList(agentsJSON, offset, limit, total);
@@ -266,13 +267,13 @@ public class CaptureAgentsEndpoint {
    *
    * @param agent
    *          The target capture agent
-   * @param withInputs
-   *          Whether the agent has inputs
+   * @param withParsedCapabilities
+   *          Add capabilities as individual fields and pre-parse them
    * @param details
    *          Whether the configuration and capabilities should be serialized
    * @return A {@link JsonObject} representing the capture agent
    */
-  private JsonObject generateJsonAgent(Agent agent, boolean withInputs, boolean details) {
+  private JsonObject generateJsonAgent(Agent agent, boolean withParsedCapabilities, boolean details) {
     JsonObject json = new JsonObject();
     String status = AgentState.TRANSLATION_PREFIX + agent.getState().toUpperCase();
     json.addProperty("Status", safeString(status));
@@ -280,13 +281,30 @@ public class CaptureAgentsEndpoint {
     json.addProperty("Update", safeString(toUTC(agent.getLastHeardFrom())));
     json.addProperty("URL", safeString(agent.getUrl()));
 
-    if (withInputs) {
-      String devices = (String) agent.getCapabilities().get(CaptureParameters.CAPTURE_DEVICE_NAMES);
-      if (devices == null || devices.isEmpty()) {
-        json.add("inputs", new JsonArray());
+    if (withParsedCapabilities) {
+      JsonObject parsedCapabilities = new JsonObject();
+
+      String inputs = (String) agent.getCapabilities().get(CaptureParameters.CAPTURE_DEVICE_NAMES);
+      if (inputs == null || inputs.isEmpty()) {
+        parsedCapabilities.add("inputs", new JsonArray());
       } else {
-        json.add("inputs", generateJsonDevice(devices.split(",")));
+        String[] parsedInputs = Arrays.stream(inputs.split(","))
+            .map(String::trim)
+            .toArray(String[]::new);
+        parsedCapabilities.add("inputs", generateJsonDevice(parsedInputs, TRANSLATION_KEY_PREFIX + "INPUTS."));
       }
+
+      String stream = (String) agent.getCapabilities().get(CaptureParameters.CAPTURE_DEVICE_STREAM);
+      if (stream == null || stream.isEmpty()) {
+        parsedCapabilities.add("stream", new JsonArray());
+      } else {
+        String[] parsedStream = Arrays.stream(stream.split(","))
+            .map(String::trim)
+            .toArray(String[]::new);
+        parsedCapabilities.add("stream", generateJsonDevice(parsedStream, TRANSLATION_KEY_PREFIX + "STREAM."));
+      }
+
+      json.add("parsedCapabilities", parsedCapabilities);
     }
 
     if (details) {
@@ -326,13 +344,13 @@ public class CaptureAgentsEndpoint {
    *          an array of devices String
    * @return A {@link JsonArray} representing the devices
    */
-  private JsonArray generateJsonDevice(String[] devices) {
+  private JsonArray generateJsonDevice(String[] devices, String translationPrefix) {
     JsonArray jsonDevices = new JsonArray();
 
     for (String device : devices) {
       JsonObject jsonDevice = new JsonObject();
       jsonDevice.addProperty("id", device);
-      jsonDevice.addProperty("value", TRANSLATION_KEY_PREFIX + device.toUpperCase());
+      jsonDevice.addProperty("value", translationPrefix + device.toUpperCase());
       jsonDevices.add(jsonDevice);
     }
 
--- a/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
+++ b/modules/admin-ui/src/main/java/org/opencastproject/adminui/endpoint/CaptureAgentsEndpoint.java
@@ -324,6 +324,16 @@ public class CaptureAgentsEndpoint {
         parsedCapabilities.add("layout", generateJsonDevice(parsedLayout, TRANSLATION_KEY_PREFIX + "LAYOUT."));
       }
 
+      String cameraPosition = (String) agent.getCapabilities().get(CaptureParameters.CAPTURE_DEVICE_CAMERA_POSITION);
+      if (cameraPosition == null || cameraPosition.isEmpty()) {
+        parsedCapabilities.add("cameraPosition", new JsonArray());
+      } else {
+        String[] parsedCameraPosition = Arrays.stream(cameraPosition.split(","))
+            .map(String::trim)
+            .toArray(String[]::new);
+        parsedCapabilities.add("cameraPosition", generateJsonDevice(parsedCameraPosition, TRANSLATION_KEY_PREFIX + "CAMERA_POSITION."));
+      }
+
       json.add("parsedCapabilities", parsedCapabilities);
     }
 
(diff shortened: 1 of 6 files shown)


gregorydlogan (commented):
For the actual implementation, I have to wonder if moving the various marshalling implementations into a central class (say, AgentImpl) would make sense.  Because as this code is right now, the CA admin endpoints don't reflect any of this.

It would also be nice to reduce some of the duplications where we do the same thing with different string constants.  I'm not sure how to do that without introducing other, *different* issues though.

Arnei:
> Because as this code is right now, the CA admin endpoints don't reflect any of this.

What do you mean by that? Is there a place in the code I should have changed but didn't?

gregorydlogan:
Giant, giant info dump here, but I don't know how better to do this.  Vendors, don't do anything yet, we'll let you know when this is stable enough to start work.

From the Manchester CA API discussion regarding CA API 2.0:

The goal here is to take the existing CA API and *extend* it, rather than replace it.  Ideally 1.0 agents should transparently work with no changes required - we really don't want to break compat with existing hardware which may or may not be field upgradable firmware wise.  All names, numbers, and encoding schemes below are subject to change still.  I wrote this with the intention of combining this and [the existing API](https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/) page once it's complete.  When reading this, assume that all of the 1.0 stuff is in the doc as well.

## Semi-decided

### Agent registration and configuration data sent to Opencast (Agent->Core communication)

Properties/Capabilities are now required.  Agents which only register their state are not API 2.0 agents.

When registering its capabilities, the CA MUST include the following when registering with the core
  - The Vendor name, as an ASCII string of not more than 256 characters named `vendor.name`
  - The Model name, as an ASCII string of not more than 256 characters named `vendor.model`
  - The Firmware version, as an ASCII string of not more than 256 characters named `vendor.firmware`
  - The Hardware revision, as an ASCII string of not more than 256 characters named `vendor.hardware`

When registering, the CA MAY include the following when registering with the core 
- The CAs inputs, as an ASCII string of not more than 256 characters named `capture.device.names`
- Whether the device supports starting a local capture in a paused state, as a binary digit named `capture.local.startpaused`
- Whether the device supports streaming, as a binary digit named `capture.stream.capable`
- If supporting streaming (`capture.stream.capable=1`): 
  - Whether the device starts streaming at capture start, as a binary digit named `capture.stream.startpaused`
    - Agents which support streaming *and* not starting immediately SHOULD implement a filler 'Recording coming soon' option.
  - Agents MAY provide a list of bitrates+frame sizes supported by the CA for streaming, as a list of ASCII strings named `capture.stream.configuration`
      - Each list entry must not be more than 32 ASCII characters
      - These entries are vendor defined
      - Omitting this list means the stream settings are configured on the device itself, or these settings are non-configurable
      - CAs MUST NOT send this as an empty list, omit it instead
- If the agent supports controlling PTZ cameras: 
  - A list of camera positions, named `capture.device.positions`
  - Each list entry must be not more than 256 ASCII characters
  - CAs MUST NOT send this as an empty list, omit it instead
- If the agent supports frame composition locally (ie, combining any two or more inputs): 
  - A list of content selection options, named `capture.device.content`
  - Each list entry must be not more than 256 ASCII characters
  - CAs MUST NOT send this as an empty list, omit it instead
  - NB: We said inputs, that includes audio.  A combination like "Input A + Input C (Mute)" is reasonable.
  - NB: The definition of combinations (or lack thereof) is deliberately left to the vendors here A vendor may choose not to support muted capture, for example.
- Any vendor specific extensions, as ASCII strings of not more than 256 characters named with prefix `X-$VENDOR_NAME`, eg: `X-pyCA`
  - The key name will be validated against the regular expression `[A-Za-z0-9]{0,256}`
  - Keys not matching the expression above will be discarded

### Agent properties (`org.opencastproject.capture.agent.properties`, parsed by the device from the ical data):

- The config key `capture.local.startpaused` MUST be parsed as a binary digit, with 1 starting recording at event start time and 0 allowing the device's configuration to take precedence
- If supporting streaming (`capture.stream.capable=1`):
  - `capture.stream` MUST be parsed as a binary digit, with 1 starting the stream at event start time, and 0 not starting the stream
  - `capture.stream.target` MUST be parsed as an ASCII string, and be the url which the CA sends its stream to
    - The CA MAY provide support for interpolation of variables (ie, the event ID) in this string
  - `capture.stream.configuration` MAY be parsed and matched in a vendor defined way
    - If the configuration does not match something the CA can provide (for example, if a CA has been swapped for a different model) then the CA should use its default factory bitrate and frame size
- If supporting `capture.stream.startpaused=1`:
  - `capture.stream.startpaused` MUST be parsed as a binary digit, with 0 starting the stream at event start time, and 1 not starting the stream or providing 'coming soon frames'
- If supporting frame composition locally (`capture.device.content is present and not empty`):
  - `capture.device.content` MUST be parsed as an ASCII string and honoured by the CA
    - If the content configuration does not match what the CA can provide (for example, if a CA has been swapped for a different model) then the CA should use its default factory content configuration
  - Inputs not specified in the content configuration MUST NOT be included in the resulting capture and MUST NOT be streamed to the stream target, if supported
- If supporting PTZ camera control
  - `capture.device.position` MUST be parsed as an ASCII string and honoured by the CA
    - If the position does not match what the CA can provide (for example, if a CA has been swapped for a different model, or the position is not supported by the camera) then the CA should discard the instruction
- Vendor specific extensions (X-$VENDOR_NAME) should be parsed and handled as desired
- Properties not matching the above should be silently ignored
- An `org.opencastproject.capture.agent.properties` file MUST be attached to the mediapackage on ingest containing:
  - All properties from the version attached to the ical event
  - All agent properties passed to the core during registration (vendor name, model, firmware and hardware rev)
  - Any other properties the vendor desires to include
    - These must be UTF8 encoded, and suitable for default decoding in a Java Properties object

#### New possible configurations that need thinking about:

- Stream-native devices (ie, auto stream, not auto local capture)

### Open questions:

- Is CA API 2.0 a sane name?  We're going to be talking about this a lot, we should pick something nice to deal with.
- If a CA starts neither streaming, nor local capture, what actually happens when a recording starts?
- We have `capture.device.names`, and `capture.device.content`, how do they interact?  If I don't pick an input that's in a content configuration, what happens?  Ie, `capture.device.names=A,B`, and `capture.device.content=A+B+C`. 
  - Immediate thought: Support either, but not both for a given agent
  - If an agent defines only `capture.device.names` -> We assume no support for content definitions, return only `capture.device.names` in the ical
  - If an agent defines `capture.device.content` -> We supply both content definitions and device names in the ical, but device.names is *always* the full set of device names
    - This covers the case where an event scheduled for a 2.0 agent gets recorded on a 1.0 agent instead - then the agent still gets as many inputs as it can -> no lost data
  - A 1.0 agent would then be unchanged, and only a 2.0+ agent would get the special handling
  - Leaving this vendor defined is a recipe for frustration IMO, from OC developer, OC adopter, **and** vendor perspectives
- I used "ASCII string" above, but does that cover common use cases in the community?  I have no strong feelings here.
- I used a lot of 256 character limits.  These are (probably) something the DB already handles, but can the UI?  256 is *way* more than I would hope is necessary, but maybe we should shorten these down to something that both renders nicely and is reasonable database wise?  If I replace 256 with 32, is that reasonable? 
  - Caveat here that the vendor specific extensions may need a special case here since `X-pyCA` is a lot shorter than say `X-WOLFVISION`
  - How on earth are we going to handle rendering these?  Just... don't?
- We should consider how we would render lists.  Dropdowns?
- I don't like the name `capture.device.content`, but I can't think of a better one.  Anyone got an idea?
- If `capture.device.position` or `capture.device.content` don't match what the current CA can do, then we're *technically* in an error state.  There isn't a way to surface this to the core right now.  Not sure what to do here.
- Confidence monitoring (cf Peekaboo) was long ago desired.  Is this something we want to tackle for 2.0?

Things I forgot initially, but vendors have brought up or I came up with today:
- None of the key names above are final.  Feel free to suggest options.
- If an event does not auto start local capture, and does not get manually started then the agent needs to mark the event as skipped somehow.
- The not-auto-starting thing seems to be common amongst the vendors involved thus far.  Perhaps we don't need the `capture.*.startpaused` capabilities flags - support for this becomes a *requirement* for a 2.0 agent.
- The concept of a backup stream URL has been raised.  Perhaps `capture.stream.target` should be a *list* above, and the agent is responsible for failing over when needed.
- How are we versioning things going forward.  As much as I want this to transparently handle 1.0 agents, for a hypothetical 3.0 version that might not be doable - it might not even be doable for *this* version if someone can point out a breaking issue.  I don't personally like the header based versioning in the external API, but I also don't have a better suggestion short of just adding version number in to the rest endpoint cf `/calendar/2.0/{agent_name}`

geichelberger:
> Whether the device supports starting a local capture in a paused state, as a binary digit named capture.local.startpaused

For me, the paused state is when you pause a recording or playback; wouldn't 'idle' be more accurate? Like it's currently shown in the admin-ui?

gregorydlogan:
> > Whether the device supports starting a local capture in a paused state, as a binary digit named capture.local.startpaused
> 
> For me, the paused state is when you pause a recording or playback; wouldn't 'idle' be more accurate? Like it's currently shown in the admin-ui?

I get where you're coming from, but it's not *quite* right.  The agent *isn't* idle in this case - no other recording should be schedulable in the same time slot.  It's just not actually recording to storage/streaming actual capture data.

What about changing both `startpaused` keys to something like `capture.(local|stream).autostart`, with `1` being automatically starting, and `0` being a manual start?

ypatios:
> > > Whether the device supports starting a local capture in a paused state, as a binary digit named capture.local.startpaused
> > 
> > 
> > For me, the paused state is when you pause a recording or playback; wouldn't 'idle' be more accurate? Like it's currently shown in the admin-ui?
> 
> I get where you're coming from, but it's not _quite_ right. The agent _isn't_ idle in this case - no other recording should be schedulable in the same time slot. It's just not actually recording to storage/streaming actual capture data.

as far as the human-readable terminology about the state of a "paused" ca is concerned, I would pick `ready` as opposed to `idle`. Even more descriptive would be `engaged`, as it denotes that the CA is "ready-to-fire" for that particular event but _not available_ for anything else.

geichelberger:
> I get where you're coming from, but it's not quite right. The agent isn't idle in this case - no other recording should be schedulable in the same time slot. It's just not actually recording to storage/streaming actual capture data.

Oh, I misread that. I thought the CA was simply in a state that allows manual recordings, but it's actually scheduled, but not yet started or acknowledged. 

I like the `autostart` idea.

ppettit:
thanks for writing all this down greg!

> ### Agent registration and configuration data sent to Opencast (Agent->Core communication)
> 
> Properties/Capabilities are now required. Agents which only register their state are not API 2.0 agents.

should there be an explicit "i support up to version X" key so the UI doesn't try to offer things that the device is just going to ignore?
 
> When registering its capabilities, the CA MUST include the following when registering with the core
>     * The Vendor name, as an ASCII string of not more than 256 characters named `vendor.name`
>     * The Model name, as an ASCII string of not more than 256 characters named `vendor.model`
>     * The Firmware version, as an ASCII string of not more than 256 characters named `vendor.firmware`
>     * The Hardware revision, as an ASCII string of not more than 256 characters named `vendor.hardware`

what happens if these are omitted?  i assume it doesn't really matter? maybe MUST is too strong for these?
 
> When registering, the CA MAY include the following when registering with the core
>     * Whether the device supports streaming, as a binary digit named `capture.stream.capable`

are there (or could there be!) devices that stream but don't record? if so maybe a `capture.local.capable` is needed too?

>     * If the agent supports controlling PTZ cameras:
>       * A list of camera positions, named `capture.device.positions`
>       * Each list entry must be not more than 256 ASCII characters
>       * CAs MUST NOT send this as an empty list, omit it instead

would "enable tracking" best be handled as a camera "position" (or vendor extension) or should we define something like `capture.device.tracking`? 
 
> ### Agent properties (`org.opencastproject.capture.agent.properties`, parsed by the device from the ical data):
>     * The config key `capture.local.startpaused` MUST be parsed as a binary digit, with 1 starting recording at event start time and 0 allowing the device's configuration to take precedence

i prefer `autostart` as you mention in a later comment  

>     * An `org.opencastproject.capture.agent.properties` file MUST be attached to the mediapackage on ingest containing:

> #### New possible configurations that need thinking about:
>     * Stream-native devices (ie, auto stream, not auto local capture)

ha, i mentioned that above ;)

> ### Open questions:
> 
>     * Is CA API 2.0 a sane name?  We're going to be talking about this a lot, we should pick something nice to deal with.

i guess so! but if it is not a breaking change should it really be 2.0?

>     * If a CA starts neither streaming, nor local capture, what actually happens when a recording starts?

i think that is just up to the device, whether it wants to "start" the recording in a paused state in order to set up the inputs etc, or only do that when it is actually required? but i don't know if the device should send `recording` or `paused` state or not? 
 
>     * We have `capture.device.names`, and `capture.device.content`, how do they interact?  If I don't pick an input that's in a content configuration, what happens?  Ie, `capture.device.names=A,B`, and `capture.device.content=A+B+C`.
>       * Immediate thought: Support either, but not both for a given agent
>       * If an agent defines only `capture.device.names` -> We assume no support for content definitions, return only `capture.device.names` in the ical
>       * If an agent defines `capture.device.content` -> We supply both content definitions and device names in the ical, but device.names is _always_ the full set of device names
>         
>         * This covers the case where an event scheduled for a 2.0 agent gets recorded on a 1.0 agent instead - then the agent still gets as many inputs as it can -> no lost data
>       * A 1.0 agent would then be unchanged, and only a 2.0+ agent would get the special handling
>       * Leaving this vendor defined is a recipe for frustration IMO, from OC developer, OC adopter, **and** vendor perspectives

perhaps there could be a `capture.device.names=content` which would then use the `capture.device.content=xxx`? so the composited video is just another input? i was sort of assuming the string in `capture.device.content` would just be a preset already defined on the device but maybe that is not useful? i assumed configuring a composited layout in a string would be pretty unwieldy! perhaps i am just not really understanding the requirements.
 
>     * I used "ASCII string" above, but does that cover common use cases in the community?  I have no strong feelings here.
> 
>     * I used a lot of 256 character limits.  These are (probably) something the DB already handles, but can the UI?  256 is _way_ more than I would hope is necessary, but maybe we should shorten these down to something that both renders nicely and is reasonable database wise?  If I replace 256 with 32, is that reasonable?

we have already had to increase database field size for workflow properties so i'd be inclined to support longer rather than shorter strings...
       
>     * I don't like the name `capture.device.content`, but I can't think of a better one.  Anyone got an idea?

i don't either, how about `layout`?  or `composite`?

>     * If `capture.device.position` or `capture.device.content` don't match what the current CA can do, then we're _technically_ in an error state.  There isn't a way to surface this to the core right now.  Not sure what to do here.

i would say if these do not make sense to a device it should just record everything it can.

>     * Confidence monitoring (cf Peekaboo) was long ago desired.  Is this something we want to tackle for 2.0?

at the moment we have `address=http(s)://<ca-web-ui>` in `POST ${CAPTURE-ADMIN-ENDPOINT}/agents/<name>` I guess there could be something in there (or would it go into `configuration`?)  that had a list of preview links. would probably need to include MIME types or something so the UI could know how to display (is it an image, audio stream, video stream?) and display names too. sounds like it could get complicated and the device would need to handle authentication somehow too if it were to be displayed in the admin UI. it's definitely not something manchester needs, our monitoring tools are outside of OC.

> Things I forgot initially, but vendors have brought up or I came up with today:
>     * If an event does not auto start local capture, and does not get manually started then the agent needs to mark the event as skipped somehow.

yes, at the moment if a recording is scheduled and the device starts but is not ingested then the event gets stuck in `recording` state
 
>     * The not-auto-starting thing seems to be common amongst the vendors involved thus far.  Perhaps we don't need the `capture.*.startpaused` capabilities flags - support for this becomes a _requirement_ for a 2.0 agent.

it could still just be ignored by a 1.0 agent anyway

>     * How are we versioning things going forward.  As much as I want this to transparently handle 1.0 agents, for a hypothetical 3.0 version that might not be doable - it might not even be doable for _this_ version if someone can point out a breaking issue.  I don't personally like the header based versioning in the external API, but I also don't have a better suggestion short of just adding version number in to the rest endpoint cf `/calendar/2.0/{agent_name}`

i can't see any breaking change so far. as we currently don't have a version in the url i wouldn't bother adding one unless there really is a breaking change

mtneug:
@gregorydlogan thank you for the write-up. I think it's mostly clear, but I think each new option could benefit from an example.

+1 for `capture.(local|stream).autostart`
+1 for `capture.device.layout`
+1 for streaming-only events and devices
+1 for naming this CA API 2.0. Even if there is technically no breaking change, this still indicates a significant change and hopefully prompts vendors to support this new version.

ASCII is probably fine. The only option I'm unsure is `capture.stream.target`. I assume this is a URL (or list of URLs). If so, this could be non-ASCII right?

gregorydlogan:
Brought up in the architecture meeting today:
Quarkus does not support digest auth.  Given that we want to continue supporting existing CAs, we will have to build our own support for digest, however this code would then need to be maintained.  Does it make sense to mark digest auth as deprecated as part of CA API 2.0, and point new integrations at... something else?

mtneug:
> Does it make sense to mark digest auth as deprecated as part of CA API 2.0, and point new integrations at... something else?

I don't think we have officially stated this, but the recommendation given in the recent past has been to move to basic auth anyways. Still, I don't see digest auth being removed soon.

gregorydlogan:
I have not read or handled the other comments, this is still pending.  But, further discussion with other stakeholders today brought up that agents should either support *multiple* agent states simultaneously, or combined states.  If an agent is currently uploading *and* recording, the API can only handle a single state currently.  Most vendors place more importance on uploading, but then recordings don't get marked as recording.  IMO recording states should have handled this, but I see how adopters get this confused.

IMO, we add combined states to the agent status.  The valid states then are IDLE, CAPTURING, UPLOADING, and CAPTURING_AND_UPLOADING.

Input welcome.

gregorydlogan:
In case this comes up at DACH, I'm currently working on this in my fork under a branch with the same name.

mtneug:
> In case this comes up at DACH, I'm currently working on this in my fork under a branch with the same name.

This was talked about only in the context of https://github.com/opencast/opencast-quarkus/issues/25 ([see comment there](https://github.com/opencast/opencast-quarkus/issues/25#issuecomment-5537995694)).

PR #7124, opened 2025-10-20
Mark system as broken on admin circular reference

gregorydlogan:
This PR deliberately throws an exception in one of the cases where #4117 has occurred.  This doesn't make the experience better since things will fail and the service will go into an ERROR state, *but* this hopefully means the user will investigate what's going on.  The system is broken in this case regardless, so a restart is basically *required* anyway.

### How to test this patch

Somehow figure out how to trigger #4117, and at the same time don't fix the underlying issue (which is a Karaf bug of somesorts).

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


mtneug:
> so a restart is basically required anyway.

I wonder, can/should we crash Opencast and let the init/orchestration system restart Opencast?

gregorydlogan:
I think that would be reasonable as well. Will bring it up tomorrow.

Or could we grab the relevant bundles from Karaf and force them to restart?

gregorydlogan:
I won't have time to address this until the new year, and still think we should do *something* about it.  Bumping to 18 so this doesn't get lost.

gregorydlogan:
Dropping in favour of #7128

PR #7167, opened 2025-11-04
Fix captions not published from editor

Arnei:
When adding captions to an event through the editor, and the event was not themed, the captions would not be published. This fixes that.

Fixes the issue by adding another tag operation to the theming workflow that aims to ensure that captions end up with the expected flavor (captions/shifted) after the workflow has run.

Targeting legacy because I am considering this a bugfix. But since this is technically a configuration change, as workflows are technically configuration, I have no problem retargeting this.

### How to test this patch

- Configure the editor to allow for subtitle editing.
- Upload a test event to Opencast that does not belong to a themed series and without subtitles.
- Open the test event in the editor. Add a subtitle. Start the "publish" workflow from the editor.
- Let the workflow run.
- Check if the search endpoint (/search/episode.json?id={id}) returns a subtitle track.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/workflows/partial-theming.xml
+++ b/etc/workflows/partial-theming.xml
@@ -40,6 +40,17 @@
     </operation>
 
     <operation
+        id="tag"
+        if="NOT(${theme_active})"
+        description="If no shifting was necessary, mark the trimmed captions as shifted">
+      <configurations>
+        <configuration key="source-flavors">captions/trimmed</configuration>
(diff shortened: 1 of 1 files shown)


gregorydlogan:
@KatrinIhler thoughts on merging this into 17?  IMO this is a bug fix, and things will continue to (not) work regardless of whether the adopter's config is updated.  Tested locally and this fixes the issue.

PR #7170, opened 2025-11-06
Add audio merging in partial ingest workflow

marwyg:
These changes ensure, that Opencast correctly handles media where partial ingests include overlapping audio. Therefore it fixes the BigBlueButton-Opencast Integration, because BBB is using LiveKit since version 3.0, which results in multiple overlapping audio snippets.

Note: 
a) This PR adds an encoding profile.
b) The workflow operation should function like before with one benefit: Now it can handle audio that overlaps.

### How to test this patch

You have to ingest multiple audio tracks. It would be good if different combinations would be tested like: 
1. presenter has video tracks, presentation has audio tracks
2. presenter has video and audio tracks, presentation has nothing
3. ...

I created a script in the helper-scripts repository for testing purposes:
https://github.com/marwyg/helper-scripts/tree/complex-partial-ingest/ingest/complex-partial-ingest

There is also a PR for this:
https://github.com/opencast/helper-scripts/pull/82

In this repository you will find a partial-ingest.sh script, where you can configure your opencast stuff. In the directory is also a workflow you can use for the testing (partial-import-test.xml). The workflow simply runs the workflow operations and ends, so you can check the result in the assets. You can change up the encoding profiles in the workflow or use the encoding profiles provided in the same directory as the workflow. The file is named `partial-import-encoding-profiles.properties`.

I created some audio and video files. I configured the ingests in a way, that the presenter tracks will include lower pitched audios and the presentation higher pitched once.

I also added some video snippets. I am ingesting it in such a way, that the presenter tracks will start with the blue-yellow snippet and end with the purple one. The presentation tracks will be vice versa.

There is a new endpoint for the audio merge part, because the merge jobs will be offloaded to the workers. To test this locally, you can use the multi_node_setup script: https://github.com/marwyg/helper-scripts/tree/complex-partial-ingest/multi-node-test-setup: https://github.com/opencast/helper-scripts/tree/master/multi-node-test-setup It will build an admin, presentation and worker node and you can start them all locally and let them run simultaneously.

--

When I tested this I got the following results, but please feel free to add more test cases. Maybe I forgot some:

legend: v = video, a = audio, p = presenter, s = presentation (s for slides)
for example: vp - vas -> presenter with video, presentation with video and audio

- vap - vas
	- 2 Tracks, both are muxed video and audio
- ap - as
	- 2 tracks, both audio
- ap
	- 1 presenter audio track
- as
	- 1 presentation audio track
- vp - vs
	- 2 tracks, both video only
- vp
	- 1 presenter video track
- vs
	- 1 presentation video track
- vas
	- 1 presentation track, audio and video muxed
- vap
	- 1 presenter track, audio and video muxed
- ap - vs
	- 1 presentation track, audio and video muxed
- ap - vas
	- 2 tracks, presenter = audio, presentation = audio and video muxed
- vp - as
	- 1 presenter track, audio and video muxed
- vp - vas
	- 2 tracks, presenter = video, presentation = audio and video muxed
- vap - as
	- 2 tracks, presenter = video and audio muxed, presentation = audio
- vap - vs
	- 2 tracks, presenter = video and audio muxed, presentation = video
	
	
edit: 

1) I added a default encoding profile as fallback, therefore the workflow parameter is optional now.
2) The reason for this going into 17 is, that this is an important Bugfix, because the BBB-OC Integration no longer works, without these changes.

### DIFF
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -31,6 +31,7 @@ import static org.opencastproject.util.data.Tuple.tuple;
 import org.opencastproject.composer.api.ComposerService;
 import org.opencastproject.composer.api.EncoderException;
 import org.opencastproject.composer.api.EncodingProfile;
+import org.opencastproject.composer.api.EncodingProfileImpl;
 import org.opencastproject.composer.api.LaidOutElement;
 import org.opencastproject.composer.api.VideoClip;
 import org.opencastproject.composer.layout.Dimension;
@@ -166,6 +167,7 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
   private static final int PROCESS_SMIL_FAILED = 19;
   private static final int MULTI_ENCODE_FAILED = 20;
   private static final int NO_STREAMS = 23;
+  private static final int AUDIO_MERGE_FAILED = 24;
 
   /** The logging instance */
   private static final Logger logger = LoggerFactory.getLogger(ComposerServiceImpl.class);
@@ -254,6 +256,9 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
   /** Path to the FFmpeg binary */
   private String ffmpegBinary = FFMPEG_BINARY_DEFAULT;
 
+  /** Default audio-merge encoding profile if no one is set **/
+  private EncodingProfile defaultAudioMergeEncodingProfile;
+
   /** Creates a new composer service instance. */
   public ComposerServiceImpl() {
     super(JOB_TYPE);
@@ -272,6 +277,16 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
     ffmpegBinary = StringUtils.defaultString(cc.getBundleContext().getProperty(CONFIG_FFMPEG_PATH),
             FFMPEG_BINARY_DEFAULT);
     logger.debug("ffmpeg binary: {}", ffmpegBinary);
+
+    // creating default encoding profile in case none is set
+    EncodingProfileImpl profile = new EncodingProfileImpl("audiomerge.work", "audiomerge", null);
+    profile.setOutputType(EncodingProfile.MediaType.Audio);
+    profile.setSuffix("-merged.mkv");
+    profile.setApplicableType(EncodingProfile.MediaType.AudioVisual);
+    profile.addExtension("ffmpeg.command", "#{audioMergeCommand} -c:a flac #{out.dir}/#{out.name}#{out.suffix}");
+    this.defaultAudioMergeEncodingProfile = profile;
+    logger.info("Default audio merge encoding profile set: " + profile.getIdentifier());
+
     logger.info("Activating composer service");
   }
 
@@ -974,7 +989,20 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
       audioFiles.add(loadTrackIntoWorkspace(job, "audio_merge", t, false));
     }
 
-    EncodingProfile encodingProfile = getProfile(job, profileId);
+    // try to get encoding profile from file, use default as fallback
+    EncodingProfile encodingProfile = profileScanner.getProfile(profileId);
+    if (encodingProfile == null) {
+      final String msg = format("Profile %s is unknown.", profileId);
+      if ("audiomerge.work".equals(profileId)) {
+        logger.info(msg + " Using default");
+        encodingProfile = defaultAudioMergeEncodingProfile;
+      } else {
+        logger.error(msg);
+        incident().recordFailure(job, PROFILE_NOT_FOUND, Collections.map(tuple("profile", profileId)));
+        throw new EncoderException(msg);
+      }
+    }
+
     final EncoderEngine encoderEngine = getEncoderEngine();
 
     String audioMergeCommand = buildAudioMergeCommand(audioStarTimes, audioFiles);
@@ -993,7 +1021,7 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
       params.put("tracks", StringUtils.join(trackList, ","));
       params.put("profile", encodingProfile.getIdentifier());
       params.put("properties", properties.toString());
-      incident().recordFailure(job, CONCAT_FAILED, e, params, detailsFor(e, encoderEngine));
+      incident().recordFailure(job, AUDIO_MERGE_FAILED, e, params, detailsFor(e, encoderEngine));
       throw e;
     } finally {
       activeEncoder.remove(encoderEngine);
--- a/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
+++ b/modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
@@ -221,7 +221,7 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
 
   /** List of available operations on jobs */
   enum Operation {
-    Encode, Image, ImageConversion, Mux, Trim, Composite, Concat, ImageToVideo, ParallelEncode, Demux, ProcessSmil, MultiEncode
+    Encode, Image, ImageConversion, Mux, Trim, Composite, Concat, ImageToVideo, ParallelEncode, Demux, ProcessSmil, MultiEncode, MergeAudioTracks
   }
 
   /** tracked encoder engines */
@@ -919,6 +919,22 @@ public class ComposerServiceImpl extends AbstractJobProducer implements Composer
   }
 
   @Override
+  public Job mergeAudioTracks(String profileId, List<Long> audioStartTimes, List<Track> audioTracks) throws MediaPackageException, EncoderException {
+    try {
+      ArrayList<String> arguments = new ArrayList<String>();
+      arguments.add(0, profileId);
+      String audioStartsStringified = audioStartTimes.stream().map(String::valueOf).collect(Collectors.joining(","));
+      arguments.add(1, audioStartsStringified);
+      for (int i = 0; i < audioTracks.size(); i++) {
+        arguments.add(i + 2, MediaPackageElementParser.getAsXml(audioTracks.get(i)));
+      }
+      return serviceRegistry.createJob(JOB_TYPE, Operation.MergeAudioTracks.toString(), arguments);
+    } catch (ServiceRegistryException e) {
+      throw new EncoderException("Unable to create 'Merge Audio Tracks' job", e);
+    }
+  }
+
+  @Override
   public Job concat(String profileId, Dimension outputDimension, float outputFrameRate, boolean sameCodec, Track... tracks) throws EncoderException,
           MediaPackageException {
     ArrayList<String> arguments = new ArrayList<String>();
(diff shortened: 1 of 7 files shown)


modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/PartialImportWorkflowOperationHandler.java
```diff
@@ -115,11 +115,16 @@ public class PartialImportWorkflowOperationHandler extends AbstractWorkflowOpera
   private static final String TARGET_PRESENTATION_FLAVOR = "target-presentation-flavor";
 
   private static final String CONCAT_ENCODING_PROFILE = "concat-encoding-profile";
+  private static final String AUDIO_MERGE_ENCODING_PROFILE = "audio-merge-encoding-profile";
```
mtneug, line 118: Is this already documented in the workflow operation documentation?

mtneug (changes requested):
Why should this go into 17, the legacy release?

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/PartialImportWorkflowOperationHandler.java
marwyg: Thank you for taking a look! :) 
Regarding the version, I maybe confused something, I will check on that. And regarding the docs, you are right, I forgot to update them, thank you for pointing out! :) I will update them

gregorydlogan:
I'm not sure this can go into either current release. The upgraded code needs to work without any config file changes, which means the encoding profile would be missing...

modules/composer-workflowoperation/src/main/java/org/opencastproject/workflow/handler/composer/PartialImportWorkflowOperationHandler.java
marwyg: Its documented now

marwyg:
1) I added a default encoding profile as fallback, therefore the workflow parameter is optional now.
2) The reason for this going into 17 is, that this is an important Bugfix, because the BBB-OC Integration no longer works, without these changes.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -918,6 +933,22 @@ public Job concat(String profileId, Dimension outputDimension, boolean sameCodec
...
   }
 
+  @Override
+  public Job mergeAudioTracks(String profileId, List<Long> audioStartTimes, List<Track> audioTracks) throws MediaPackageException, EncoderException {
+    try {
+      ArrayList<String> arguments = new ArrayList<String>();
+      arguments.add(0, profileId);
+      String audioStartsStringified = audioStartTimes.stream().map(String::valueOf).collect(Collectors.joining(","));
+      arguments.add(1, audioStartsStringified);
+      for (int i = 0; i < audioTracks.size(); i++) {
```
gregorydlogan, line 943: I'm not a fan of this approach, but `getAsXml` throws exceptions so we can't turn this into a functional call :(

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -941,6 +972,78 @@ public Job concat(String profileId, Dimension outputDimension, float outputFrame
     }
   }
 
+  private Option<Track> mergeAudioTracks(Job job, String profileId, List<Long> audioStarTimes, List<Track> audioTracks) throws EncoderException {
```
gregorydlogan:
```suggestion
  private Option<Track> mergeAudioTracks(Job job, String profileId, List<Long> audioStartTimes, List<Track> audioTracks) throws EncoderException {
```

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -941,6 +972,78 @@ public Job concat(String profileId, Dimension outputDimension, float outputFrame
...
+    List<File> audioFiles = new ArrayList<>();
+    for (Track t : audioTracks) {
+      if (t.hasVideo()) {
+        throw new EncoderException(String.format("There was at least one video in the audio track list: '{}'", t.getURI()));
+      }
+      audioFiles.add(loadTrackIntoWorkspace(job, "audio_merge", t, false));
+    }
+
+    // try to get encoding profile from file, use default as fallback
+    EncodingProfile encodingProfile = profileScanner.getProfile(profileId);
```
gregorydlogan, line 993: Default defined by this PR (`encode.partial-import`) is not included in the PR.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
```diff
@@ -652,6 +653,51 @@ public Response concat(@FormParam("sourceTracks") String sourceTracksXml, @FormP
...
+
+    // Deserialize the audio tracks
+    List<? extends MediaPackageElement> tracks = MediaPackageElementParser.getArrayFromXml(audioTracksXml);
+    if (tracks.size() < 2) {
+      return Response.status(Response.Status.BAD_REQUEST).entity("At least two tracks must be set to merge").build();
+    }
+
+    for (MediaPackageElement elem : tracks) {
+      if (!Track.TYPE.equals(elem.getElementType())) {
+        return Response.status(Response.Status.BAD_REQUEST).entity("Source tracks must be of type 'track'").build();
```
gregorydlogan, line 684: We check later if they have video and throw an exception, should we check that here too?

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
```diff
@@ -652,6 +653,51 @@ public Response concat(@FormParam("sourceTracks") String sourceTracksXml, @FormP
...
+      }
+    }
+
+    List<Long> audioStartTimes = Arrays.stream(audioStartTimesStringified.split(","))
```
gregorydlogan, line 691: Similarly, should we check the start time count matches the track count?

modules/composer-service-remote/src/main/java/org/opencastproject/composer/remote/ComposerServiceRemoteImpl.java
```diff
@@ -524,6 +524,35 @@ public Job composite(Dimension compositeTrackSize, Option<LaidOutElement<Track>>
     }
   }
 
+  public Job mergeAudioTracks(String profileId, List<Long> audioStartTimes, List<Track> audioTracks) throws EncoderException, MediaPackageException {
```
gregorydlogan, line 527: Note for future re-review: This *must* be tested in a distributed install prior to merge!

gregorydlogan (changes requested):
Biggest issue here is the lack of encoding profile.  Didn't get through `PartialImportWorkflowOperationHandler` yet either.

Given that fixing this *requires* workflow and encoding changes, I don't know that this can go into 17.x, or 18.x for that matter.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
```diff
@@ -272,6 +277,16 @@ public void activate(ComponentContext cc) {
...
     logger.debug("ffmpeg binary: {}", ffmpegBinary);
+
+    // creating default encoding profile in case none is set
+    EncodingProfileImpl profile = new EncodingProfileImpl("audiomerge.work", "audiomerge", null);
+    profile.setOutputType(EncodingProfile.MediaType.Audio);
+    profile.setSuffix("-merged.mkv");
+    profile.setApplicableType(EncodingProfile.MediaType.AudioVisual);
+    profile.addExtension("ffmpeg.command", "#{audioMergeCommand} -c:a flac #{out.dir}/#{out.name}#{out.suffix}");
+    this.defaultAudioMergeEncodingProfile = profile;
+    logger.info("Default audio merge encoding profile set: " + profile.getIdentifier());
```
marwyg, line 288: Here is the fallback encoding profile defined.

marwyg: Thank you for the review Greg! :) I will go through your points.

This PR should not change any behavior and there also shouldn't be any necessary changes to be made in the workflow or in the encoding profiles. The new workflow option is optional (or at least it should be, I will take a second look) and the encoding profile above will be created and picked, if it's not present in any encoding-profiles file.

Regarding the remote implementation, I also tested this, but I didn't add a description at the top of this PR. I will update the description with an "how to test the remote impl" part.

gregorydlogan: Sounds good.  I think i'm tripping over the encoding profile stuff since https://github.com/opencast/helper-scripts/pull/82 uses the `partial-import-preencode`, `audiomerge.work`, `trim.work`, and `encode.partial-import` encoding profiles, but I don't have any of those defined.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
marwyg: Fixed the typo

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
marwyg: I will add some encoding profiles, so it can be tested. But as described below, the new encoding profile for the audio merge should not be necessary. A default will be set in the code, if it's not defined in any encoding file.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
marwyg: I think it makes sense, I added error cases so it fails with a bad request when videos were provided.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/endpoint/ComposerRestService.java
marwyg: I think this also does make sense. I added an error case for this.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
marwyg: I edited the description of this PR. I added missing encoding profiles to the helper-scripts repository (my fork). I also mentioned the multi_node_setup Script. With this script, the remote impl can be tested (I did it like this locally).

marwyg:
@gregorydlogan I took care of your points and also edited the Description.

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
gregorydlogan: Sigh.  Now it's complaining about a missing `import.preview` profile.  Don't suppose you've got one of those kicking around as well?

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
gregorydlogan: This 'defined in the code' part does not work.  If I remove the profile from the config file it bails with
```
2025-11-24 16:14:31,993 | ERROR | (WorkflowOperationWorker:140) - Workflow operation 'operation:'partial-import, state:'FAILED'' failed
org.opencastproject.workflow.api.WorkflowOperationException: org.opencastproject.workflow.api.WorkflowOperationException: Force encoding profile 'encode.partial-import' was not found
        at org.opencastproject.workflow.handler.composer.PartialImportWorkflowOperationHandler.start(PartialImportWorkflowOperationHandler.java:203) ~[?:?]
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.start(WorkflowOperationWorker.java:212) ~[!/:?]
        at org.opencastproject.workflow.impl.WorkflowOperationWorker.execute(WorkflowOperationWorker.java:117) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.runWorkflowOperation(WorkflowServiceImpl.java:722) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl.process(WorkflowServiceImpl.java:1755) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2116) [!/:?]
        at org.opencastproject.workflow.impl.WorkflowServiceImpl$JobRunner.call(WorkflowServiceImpl.java:2082) [!/:?]
Caused by: org.opencastproject.workflow.api.WorkflowOperationException: Force encoding profile 'encode.partial-import' was not found
        at org.opencastproject.workflow.handler.composer.PartialImportWorkflowOperationHandler$1.apply(PartialImportWorkflowOperationHandler.java:679) ~[?:?]
        at org.opencastproject.workflow.handler.composer.PartialImportWorkflowOperationHandler$1.apply(PartialImportWorkflowOperationHandler.java:673) ~[?:?]
```

gregorydlogan (changes requested):

modules/composer-ffmpeg/src/main/java/org/opencastproject/composer/impl/ComposerServiceImpl.java
marwyg: This should work I hope :)

profile.import.preview.name = Extract an image
profile.import.preview.input = visual
profile.import.preview.output = image
profile.import.preview.suffix = -image.jpg
profile.import.preview.ffmpeg.command = -ss #{time} -i #{in.video.path} -r 1 -frames:v 1 #{out.dir}/#{out.name}#{out.suffix}

marwyg:
In your error message its complaining about the "Force" encoding profile. Could it be, that you mixed something up? 

```xml
    <operation id="partial-import"
               description="Put videos and audio together"
               fail-on-error="true"
               exception-handler-workflow="partial-error">
      <configurations>
        <configuration key="source-presenter-flavor">presenter/source</configuration>
        <configuration key="source-presentation-flavor">presentation/source</configuration>
        <configuration key="source-smil-flavor">smil/source+partial</configuration>
        <configuration key="target-presentation-flavor">presentation/unified</configuration>
        <configuration key="target-presenter-flavor">presenter/unified</configuration>
        <configuration key="preencode-encoding-profile">partial-import-preencode</configuration>
        <configuration key="concat-encoding-profile">concat.work</configuration>
        
        <!-- this is optional, if this config is removed completly, the operation should take the default one (from the code) -->
        <configuration key="audio-merge-encoding-profile">audiomerge.work</configuration> 
        
        <configuration key="trim-encoding-profile">trim.work</configuration>
        
        <!-- did you maybe removed this one accidentally?? -->
        <configuration key="force-encoding-profile">encode.partial-import</configuration>
      </configurations>
    </operation>
```

gregorydlogan:
While I can't guarantee I haven't missed something in earlier testing, latest testing now complains about a missing profile called `import.image-frame`.

If this is working for you then I would ask that you diff your *entire* encoding and workflow directories against upstream's `r/17.x`.  Maybe we're missing just list last profile, or maybe there are more.  Either way, then we'd know.  And please include the relevant missing profile(s) in https://github.com/opencast/helper-scripts/pull/82 so that they're in the helper scripts too.  The settings for `import.preview` aren't in there, and I was *very* confused until I caught that.

KatrinIhler:
Talked to Martin, he will repoint to 18 when he has the time.

gregorydlogan:
Closing this since 17 is now done.  Please refile when ready.

PR #7175, opened 2025-11-07
Handle org wildcard for catalog ui adapters

KatrinIhler:
We allow setting catalog ui adapters for all organizations with *, but not all services where handling this case correctly. This should fix this.


modules/asset-manager-impl/src/main/java/org/opencastproject/assetmanager/impl/AssetManagerImpl.java
```diff
@@ -1586,7 +1589,7 @@ private Function<Optional<Event>, Optional<Event>> getEventUpdateFunction(Snapsh
       // extended metadata
       event.resetExtendedMetadata();  // getting rid of old data
       for (EventCatalogUIAdapter extendedCatalogUIAdapter : extendedEventCatalogUIAdapters.getOrDefault(orgId,
-              Collections.emptyList())) {
```
geichelberger, line 1589: If an event catalog is for a specific organization and one for all, we would still miss the wildcard one.

KatrinIhler: Good point, I didn't consider that scenario. Fixed!

gregorydlogan (approved):
Nominal approval, though I'm by no means an extended metadata expert.

PR #7181, opened 2025-11-13
Fix HEAD requests to static files returning wrong status codes

### LINKED ISSUE #7179: Opencast incorrectly replies 302 to HEAD requests for static files (even if access is allowed)
**Describe the bug**

When sending a `HEAD` request to Opencast with a path `/static/...` and without any auth data (i.e. without cookie or basic auth), it replies 302. This is different from the status code a GET request would return, which is incorrect behavior. A few particular cases:

- If the asset is protected (ROLE_ANONYMOUS has no access), 302 is returned, but 403 should be returned (as returned by GET)
- If the asset is public (ROLE_ANONYMOUS has access), 302 is returned, but 200 should be returned (as returned by GET)
- If static file auth is disabled, 302 is returned, but 200 should be returned (as returned by GET)

This is a problem for octoka, as it needs (in some configurations) acquire whether Opencast would have allowed access. Of course, it can just send a GET request instead, but that would lead to Opencast opening the file and preparing to send it, potentially incurring a large overhead.

**To Reproduce**

```
curl -v -X HEAD 'https://develop.opencast.org/static/mh_default_org/engage-player/ID-cats/c76d9938-368e-456f-88ff-ecd7d269fff1/video-of-a-tabby-cat_1.000s-player.jpg'
```

**Expected behavior**

The same status could as GET would have returned.

LukasKalbertodt:
I'm not sure if this is the best way to do it, but I just overwrite `doHead` and do everything that `doGet` would do, except add a body. This still opens the file and does some other logic, but it is required to correctly set status code and headers.

Fixes #7179

It is targeting a legacy branch as this is a bug fix, that `octoka` will rely on. Without this being backported to 17, it's harder for people to use octoka.

### How to test this patch

Upload some test videos in different visibilities and then send HEAD requests to static file routes. 

To review the code, enable "hide whitespace", which makes the diff considerably smaller.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
+++ b/modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
@@ -178,6 +178,16 @@ public class StaticResourceServlet extends HttpServlet {
   @Override
   protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
     logger.debug("Looking for static resource '{}'", req.getRequestURI());
+    impl(req, resp, true);
+  }
+
+  @Override
+  protected void doHead(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    logger.trace("HEAD request '{}'", req.getRequestURI());
+    impl(req, resp, false);
+  }
+
+  private void impl(HttpServletRequest req, HttpServletResponse resp, boolean withBody) throws IOException {
     String path = req.getPathInfo();
     if (path == null || path.contains("..")) {
       resp.sendError(HttpServletResponse.SC_FORBIDDEN);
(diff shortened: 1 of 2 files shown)


modules/static/src/main/java/org/opencastproject/fsresources/StaticResourceServlet.java
```diff
@@ -221,14 +231,16 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IO
     ArrayList<Range> ranges = parseRange(req, resp, eTag, file.lastModified(), file.length());
 
     if ((((ranges == null) || (ranges.isEmpty())) && (req.getHeader("Range") == null)) || (ranges == FULL_RANGE)) {
```
gregorydlogan, line 233: Maybe this is me being picky, but does supporting any kind of logic around ranges make sense for HEAD requests?  Why not just check withBody here, and if it's false return?

I would probably also rename `withBody` to `isHead` or `isGet`, adjusting the logic as needed, but realistically that doesn't matter much.

LukasKalbertodt: According to the spec, `HEAD` requests should reply with exactly the same headers as a GET request would get. So we kind of have to do all this involved logic here in order to correctly set headers. 

But yes, that's why I asked whether it makes sense to implement a better API for getting these kinds of "yes/no" auth answers. But I guess it still makes sense to fix HEAD requests.

gregorydlogan: We do what the spec says, even if we think the spec is dumb.

This is good to go, will try and get it in soon

LukasKalbertodt: Ok actually, on re-reading the spec, it allows for some leeway:

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, a server MAY omit header fields for which a value is determined only while generating the content. [...] These minor inconsistencies are considered preferable to generating and discarding the content for a HEAD request, since HEAD is usually requested for the sake of efficiency.

https://httpwg.org/specs/rfc9110.html#HEAD

So I guess we could do what you suggest. But we still have to open the file to determine the status code, so I'm not sure if we save really all that much by excluding range logic on HEAD. The main benefit can only be gotten a new API. So I'd still suggest keeping it like this.

PR #7182, opened 2025-11-13
Fix APIs replying 302 for unauthorized requests

### LINKED ISSUE #3244: Reply 401 or 403 to unauthenticated API requests instead of forwarding to login page
Opencast's response to requests that are not authenticated is currently a 302 forward to the login page. This behavior makes sense for routes that humans might navigate to (e.g. Admin UI). However, for API endpoints, this behavior is inconvenient and semantically incorrect. Code that talks to those APIs (oftentimes JS) currently has to add extra checks for catch unexpected redirects. Often, code forgets to add this check, resulting in weird errors (e.g. trying to parse the login page as JSON). Further, just treating all redirects as "not authenticated" is also not great as there might be some legitimate reasons for the redirect.

I think Opencast should change the behavior in those cases (API requests). There are a couple open questions:

- How to distinguish between API requests and "human" requests?
  - Define this per endpoint/path? 
  - Or check something like the `Accept` header? However, at least Chrome includes `application/xml` in that header by default and some APIs respond with XML, so there might be some ambiguity there.

- How easy would it be to change this?

- As it changes existing behavior: is the change ok? Does someone rely on the current behavior? This would have to be brought up on mail or in technical meeting.

LukasKalbertodt:
Fixes #3244

Deciding when to reply 403 vs 302 is done purely on a path base. See the linked issue for arguments about this. The paths are configurable in the security config, so even if I missed something now, adopters can just add other paths they seem important.

The path `/` is not listed there, as it always replies 200 with a JS redirect. 

### How to test this patch

Run Opencast and just `curl` against a bunch of different paths. 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/security/mh_default_org.xml
+++ b/etc/security/mh_default_org.xml
@@ -525,8 +525,33 @@
   </bean>
 
   <!-- Redirects unauthenticated requests to the login form -->
-  <bean id="userEntryPoint" class="org.springframework.security.web.authentication.LoginUrlAuthenticationEntryPoint">
+  <bean id="redirectingEntryPoint" class="org.springframework.security.web.authentication.LoginUrlAuthenticationEntryPoint">
     <property name="loginFormUrl" value="/login.html" />
+  </bean>
+
+  <!-- Entry point that redirects to /login for certain paths, replying 403 otherwise -->
+  <bean id="userEntryPoint" class="org.opencastproject.kernel.security.UserEntryPoint">
+    <property name="redirectingEntryPoint" ref="redirectingEntryPoint" />
+    <property name="redirectingPathPatterns">
+      <list>
+        <value>**/index.html</value>
+        <value>/admin-ui/**</value>
+        <value>/admin-ng</value>
+        <value>/admin-ng/</value>
+        <value>/rest_docs.*</value>
+        <value>/rest-docs/**</value>
+        <value>/docs.*</value>
+        <value>/engage/ui/**</value>
+        <value>/studio</value>
+        <value>/studio/</value>
+        <value>/editor-ui</value>
+        <value>/editor-ui/</value>
+        <value>/engage-player</value>
+        <value>/engage-player/</value>
+        <value>/ltitools</value>
+        <value>/ltitools/</value>
+      </list>
+    </property>
   </bean>
 
   <!-- Redirect unauthenticated requests to custom login url with configurable redirect query parameter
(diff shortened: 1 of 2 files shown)


geichelberger:
The same can be achieved without Java code. This is an example of a custom Opencast distribution:

```
  <sec:http pattern="/v1/**" entry-point-ref="ep403" servlet-api-provision="true" realm="Opencast" create-session="ifRequired">
    <sec:intercept-url pattern="/v1/info/me.json" method="GET" access="ROLE_ANONYMOUS" />
    <sec:intercept-url pattern="/v1/camera" method="POST" access="ROLE_ADMIN" />
    <sec:intercept-url pattern="/v1/camera/**" method="GET" access="ROLE_ADMIN,ROLE_LOCATION" />
    <sec:intercept-url pattern="/v1/camera/**" method="PUT" access="ROLE_ADMIN,ROLE_LOCATION" />
    <sec:intercept-url pattern="/v1/capture/**" access="ROLE_ADMIN, ROLE_LOCATION" />
    <sec:intercept-url pattern="/v1/preview/**" method="GET" access="ROLE_LOCATION" />

    <!-- Authentication filter chain -->
    <sec:custom-filter position="BASIC_AUTH_FILTER" ref="authenticationFilters" />

    <!-- Opencast is shipping its own implementation of the anonymous filter -->
    <sec:custom-filter ref="anonymousFilter" position="ANONYMOUS_FILTER" />

    <!-- Enables "remember me" functionality -->
    <sec:remember-me key="opencast" user-service-ref="userDetailsService" />

    <!-- If any URLs are to be exposed to anonymous users, the "sec:anonymous" filter must be present -->
    <sec:anonymous enabled="false" />

  </sec:http>

  <bean id="ep403" class="org.springframework.security.web.authentication.Http403ForbiddenEntryPoint"/>

```

LukasKalbertodt:
@geichelberger I mentioned that in the linked issue, but said that it would probably not be ideal as it would move around lots of lines, leading to annoying merge conflicts when people update their configurations. At the end of the day, I don't really care which solution we use, but I would suspect this is easier to update. And I probably don't have the time to implement the other solution in time for the branch cut.

gregorydlogan:
I'm going to merge this as is.  IMO, minimizing the inter-branch merging is very nice for devs, and adopters.  And, at least to my eyes, this approach is clearer.

PR #7189, opened 2025-11-15
Vastly simplify JWT setup for common cases

### LINKED ISSUE #7034: Simplify setting up JWT
JWT gets more and more important in the OC ecosystem, for "app use cases" (studio, editor) to serving static files. Tobira already requires JWT for many functions for years, with LMSes also implementing support soon. But configuring JWT in Opencast is time-consuming and annoying. That goes for admins and developers. I helped many people with this, who tripped over some parts of the configuration.

In my mind, the ideal case would be: the admin only has to configure the JWKS URL (or alternatively a symmetric secret) and then everything works. With #6177 we have a standard that external applications can follow so for those, no additional configuration should be needed. In fact, for [octoka](https://github.com/opencast/octoka), this is already true: the JWKS URL is the only required config.

Of course, there are surely use cases where the current flexibility is required and I don't want to get rid of it. Just make the common case simple.

- `jwt{Header/RequestParameter}Filter`: it's fine to just enable both by default with the standard config
- `jwtLoginHandler`:
  - `jwksUrl` and `secret`: only things that actually need to be configured
  - Mapping of name, username, email: default is fine, is defined like that in the standard #6177 
  - Everything else: default is fine
- `expectedAlgorithms`: I don't think there is a security problem with just accepting almost all by default. Obviously, infamously, `none` should be always rejected, but Opencast shouldn't even allow you to configure that. But we can just allow many secure-enough algorithms by default, certainly `ES*` and `EdDSA`. I don't think that's a problem for security because the public key itself also defines an algorithm, and the public key you already fully trust.
- `claimConstraints`: I don't think we need any by default. Opencast should already internally check `exp` and similar "known claims". Apart from that I don't see any constraints we need for the common case.
- `roleMappings`: there is already `ocStandardRoleMapping` which should take care of everything for the standard case.
- All other JWT related things in the security config should be fine to enable by default.

It might also be a good idea to have a separate config file for this, at least the most important things? Then everything important is close to one another. And since the security config is changed upstream fairly recently, it reduces the amount of merges admins have to do when updating.

LukasKalbertodt:
Fixes #7034 

With this PR, OC administrators only need to change a single line (`jwksUrl) in the security config to make common cases (Tobira, LMS) work. This does not only benefit admins, but also devs who would otherwise have to adjust the security config every time they test something with JWT. This is not only annoying, but also error prone. In the past, quite a lot of human life time was lost on this non-trivial and confusing configuration. I know because I have helped many people over the past few years setting up JWT.

You should probably read this PR commit by commit. Basically each commit makes the configuration a bit easier.

I don't think this PR is a _perfect_ solution, in that it enables some logic to run even if not necessary. But:
- Most OC installations will actually configure JWT, and then there is no useless overhead.
- Implementing it in a way that admins only have to configure a single line, and only then all the machinery is enabled, is _way_ harder than this PR. 
- The wasted CPU cycles due to this are minimal, likely swamped by many other factors. And if an admin so desires, they can still disable it.

While I hope that this PR gets fully merged, I'm also fine with removing some parts, if there are parts that are too controversial to merge before the OC 19 branch cut. Every bit helps.

CC @mtneug @joneug 

### How to test this patch

Not sure how much testing is necessary here. Ideally of course, it is tested with a JWT provider like Tobira. I did that and it just works after changing this one line. 

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/docs/guides/admin/docs/configuration/security.jwt.md
+++ b/docs/guides/admin/docs/configuration/security.jwt.md
@@ -102,79 +102,44 @@ An external application wants to let a user use Opencast Studio to record and up
 Spring Security Configuration
 -----------------------------
 
-In order to active JWT-based authentication and authorization, you will need to uncomment and adapt the following
-sections found in `etc/security/mh_default_org.xml`. Some of the options are configured with the
-[Spring Expression Language (SpEL)](https://docs.spring.io/spring-framework/docs/3.0.x/reference/expressions.html).
+In order to active JWT-based authentication and authorization, you will need to adjust your security config (e.g. `etc/security/mh_default_org.xml`).
 
-* Enable the `preauthAuthProvider`.
-```xml
-<!-- Uncomment this if using Shibboleth or JWT authentication -->
-<sec:authentication-provider ref="preauthAuthProvider" />
-```
-* Enable the `userReferenceProvider`.
-```xml
-<!-- Uncomment to enable external users e.g. used together with shibboleth or JWT -->
-<osgi:reference id="userReferenceProvider" cardinality="1..1"
-                interface="org.opencastproject.userdirectory.api.UserReferenceProvider" />
-```
-* Add the configured `jwtHeaderFilter` and/or `jwtRequestParameterFilter` to the filters list
-  of the `preAuthenticationFilters` bean.
-```xml
-<!-- Uncomment the line below to support JWT. -->
-<ref bean="jwtHeaderFilter" />
-<!-- Additionally/alternatively uncomment this to support passing a JWT in a URL parameter. -->
-<ref bean="jwtRequestParameterFilter" />
-```
-* Configure the `jwtHeaderFilter` and/or `jwtRequestParameterFilter` beans.
-```xml
-<!-- General JWT header extraction filter -->
-<bean id="jwtHeaderFilter" class="org.opencastproject.security.jwt.JWTRequestHeaderAuthenticationFilter">
-  <!-- Name of the HTTP request header that contains the JWT (default: SM_USER) -->
-  <property name="principalRequestHeader" value="Authorization"/>
-  <!-- Prefix string occurring before the JWT value within the configured header (default: null) -->
-  <property name="principalPrefix" value="Bearer "/>
-  <property name="authenticationManager" ref="authenticationManager" />
-  <property name="loginHandler" ref="jwtLoginHandler" />
-  <!-- Throws an exception if a request is missing the configured header (default: true) -->
-  <property name="exceptionIfHeaderMissing" value="false" />
-</bean>
+If you just want to be able to accept JWTs from Tobira or LMS, you only have to fill out a single value: `jwksUrl` (just search for that string, it's part of the `jwtLoginHandler` bean configuration). Just set the corresponding Tobira/LMS URL there (check their docs to find the URL). All other JWT configuration can usually be kept as is, as the defaults work fine for most cases.
+(Note: currently you can only easily connect either Tobira or an LMS. Until Opencast supports multiple URLs out of the box, you have to use an external tool for aggregating JWKS.)
 
-<!-- General JWT request parameter extraction filter -->
-<bean id="jwtRequestParameterFilter" class="org.opencastproject.security.jwt.JWTRequestParameterAuthenticationFilter">
-  <!-- Name of the request parameter that contains the JWT (default: jwt) -->
-  <property name="parameterName" value="jwt" />
-  <property name="authenticationManager" ref="authenticationManager" />
-  <property name="loginHandler" ref="jwtLoginHandler" />
-  <!-- Throws an exception if a request is missing the configured parameter (default: true) -->
-  <property name="exceptionIfParameterMissing" value="false" />
-</bean>
-```
-* Configure the `jwtLoginHandler`. For the JWT validation, either configure the `secret` property for JWTs signed with
-  a symmetric algorithm or configure the `jwksUrl` for JWTs signed with an asymmetric algorithm. The `jwksUrl` should
-  provide the [JSON Web Key Set (JWKS)](https://datatracker.ietf.org/doc/html/rfc7517). If both `secret` and `jwksUrl`
-  are specified an exception is thrown and the login mechanism will not be activated. So make sure to configure only one
-  of them.
+If you want to interface with an OIDC provider or do something else entirely, configuring JWT is a bit more involved.
+
+### Advanced configuration
+
+Some of the options are configured with the [Spring Expression Language (SpEL)](https://docs.spring.io/spring-framework/docs/3.0.x/reference/expressions.html).
+In the following, certain IDs (like `jwtHeaderFilter`) are mentioned: just search in the security config file for these IDs to see where they can be configured.
+
+A few points of interest:
+
+* `jwtHeaderFilter` and `jwtRequestParameterFilter` define how JWTs are extract from incoming requests.
+  By default, `jwt` query parameter and the `Authorization` header (with prefix `Bearer `) are checked.
+
+* The main configuration is in `jwtLoginHandler`.
+  Most importantly, you have to set one of `secret` or `jwksUrl` to enable the login handler, depending on whether you are using a symmetric or asymmetric algorithm.
+  The `jwksUrl` should provide the [JSON Web Key Set (JWKS)](https://datatracker.ietf.org/doc/html/rfc7517)
+
 ```xml
-<!-- JWT login handler -->
 <bean id="jwtLoginHandler" class="org.opencastproject.security.jwt.DynamicLoginHandler">
   <property name="userDetailsService" ref="userDetailsService" />
   <property name="userDirectoryService" ref="userDirectoryService" />
   <property name="securityService" ref="securityService" />
   <property name="userReferenceProvider" ref="userReferenceProvider" />
-  <!-- JWKS URL to use for JWT validation (asymmetric algorithms) (default: null) -->
-  <property name="jwksUrl" value="https://auth.example.org/.well-known/jwks.json" />
-  <!-- How many minutes to cache a fetched JWKS before re-fetching (default: 1440) -->
-  <property name="jwksCacheExpiresIn" value="1440" />
-  <!-- Secret to use for JWT validation (symmetric algorithms) (default: null) -->
-  <property name="secret" value="***" /> <-- Change this
+  <!-- Set either `jwksUrl` or `secret` to enable this login handler -->
+  <!-- <property name="jwksUrl" value="https://replace-me.invalid/.well-known/jwks.json" /> -->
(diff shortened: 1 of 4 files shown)


docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -102,79 +102,43 @@ An external application wants to let a user use Opencast Studio to record and up
 Spring Security Configuration
 -----------------------------
 
-In order to active JWT-based authentication and authorization, you will need to uncomment and adapt the following
-sections found in `etc/security/mh_default_org.xml`. Some of the options are configured with the
-[Spring Expression Language (SpEL)](https://docs.spring.io/spring-framework/docs/3.0.x/reference/expressions.html).
+In order to active JWT-based authentication and authorization, you will need to adjust your security config (found in `etc/security/mh_default_org.xml`).
```
gregorydlogan: It's an edge case, but there *are* probably adopters with both jwt and multiple orgs. The docs should probably reflect that...

LukasKalbertodt: I only kept this default path because the previous docs also mentioned it without additional information. I assumed multi-tenant people already know that they have to change this. But I can still adjust the docs

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
```diff
@@ -151,6 +151,11 @@ public void afterPropertiesSet() {
 
   @Override
   public String handleToken(String token) {
+    if (jwkProvider == null && secret == null) {
+      logger.debug("neither jwksURL nor secret set: ignoring JWT");
```
mtneug, line 159: Having this here as a debug message is fine, but I would like this to be printed once as info/warning during startup just so that people know it is off without touching the logging config.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
```diff
@@ -125,7 +125,7 @@ public static SignedJWT verify(String token, String secret, List<String> claimCo
   public static SignedJWT verify(SignedJWT jwt, List<String> claimConstraints, JWSVerifier... verifiers)
           throws JOSEException {
     Assert.notNull(jwt, "A decoded JWT must be set");
-    Assert.notEmpty(claimConstraints, "Claim constraints must be set");
+    Assert.notNull(claimConstraints, "Claim constraints must be set");
```
mtneug, line 128: Maybe require `ocStandardRoleMappings` to the `true` before no longer requiring claim constraints?

etc/security/mh_default_org.xml
```diff
@@ -746,85 +746,76 @@
...
-    <value>['domain'] eq 'example.org'</value>
-    <value>['affiliation'].contains('faculty@example.org')</value>
-  </util:list>-->
-
-  <!-- The mapping from JWT claims to Opencast roles
+    <!-- The following are just examples if you want to add additional constraints -->
+    <!-- <value>containsKey('domain')</value> -->
+    <!-- <value>['iss'] eq 'https://auth.example.org'</value> -->
+    <!-- <value>['username'] matches '.*@example\.org'</value> -->
+    <!-- <value>['affiliation'].contains('faculty@example.org')</value> -->
```
mtneug, line 805: Why remove the other examples? Only have one "example type"?

etc/security/mh_default_org.xml
```diff
@@ -746,85 +746,76 @@
...
-    <property name="secret" value="***" />
+    <!-- Set either `jwksUrl` or `secret` to enable this login handler -->
+    <!-- <property name="jwksUrl" value="https://replace-me.invalid/.well-known/jwks.json" /> -->
+    <!-- <property name="secret" value="***" /> -->
     <property name="expectedAlgorithms" ref="jwtExpectedAlgorithms" />
     <property name="claimConstraints" ref="jwtClaimConstraints" />
     <property name="usernameMapping" value="['sub']" />
     <property name="nameMapping" value="['name']" />
-    <property name="emailMapping" value="['email']" />
+    <property name="emailMapping" value="['email'] ?: (['sub'] + '@jwt.invalid')" />
```
mtneug: I'm not really a fan of this. For us, `sub` is scoped, e.g. something like `username@example.edu` meaning the mail would be `username@example.edu@jwt.invalid`. I don't think you can (easily) inforce `sub` to be a valid local part of an email (well you can take a hash or something).

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -102,79 +102,43 @@ An external application wants to let a user use Opencast Studio to record and up
...
-  <!-- Name of the HTTP request header that contains the JWT (default: SM_USER) -->
-  <property name="principalRequestHeader" value="Authorization"/>
-  <!-- Prefix string occurring before the JWT value within the configured header (default: null) -->
-  <property name="principalPrefix" value="Bearer "/>
-  <property name="authenticationManager" ref="authenticationManager" />
-  <property name="loginHandler" ref="jwtLoginHandler" />
-  <!-- Throws an exception if a request is missing the configured header (default: true) -->
-  <property name="exceptionIfHeaderMissing" value="false" />
-</bean>
+If you just want to be able to accept JWTs from Tobira or LMS, you only have to fill out a single value: `jwksUrl` (just search for that string, it's part of the `jwtLoginHandler` bean configuration). Just set the corresponding Tobira/LMS URL there (check their docs to find the URL). All other JWT configuration can usually be kept as is, as the defaults work fine for most cases.
```
mtneug, line 107: That "Tobira or LMS" is really a "Tobira xor LMS". I haven't looked at how multiple JWKSes can be combined to accept keys from Tobira AND an LMS. Maybe mention this?

mtneug (changes requested):
Just some smaller comments from my side. I'm fine with this change in general.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
LukasKalbertodt: Mh I get what you're saying, but some people might not want to configure JWT. And then every request with JWT attached would show a warning in the log, which might blow up the log. Yes, in theory there shouldn't really be many requests with JWT incoming, but external apps might send JWTs regardless?

Currently, people also don't expect JWT to work out of the box, so why would they need to be reminded of that now?

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
LukasKalbertodt: Mhhh why? That mapping doesn't add any constraints? I don't quite see the connection here..

etc/security/mh_default_org.xml
LukasKalbertodt: Yep, that was my intention. Do you think we should keep all the old ones?

etc/security/mh_default_org.xml
LukasKalbertodt: Mh fair. But what do you suggest? Not all JWTs will have emails. Just make the mail optional? Or just a fixed mail `no-mail@jwt.invalid`?

docs/guides/admin/docs/configuration/security.jwt.md
LukasKalbertodt: Yeah, you noticed the problem here :D I have also never looked into this. We might want to just support specifying multiple JWKS URLs, as octoka does. But yep, I can mention this here.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
mtneug: No, that wasn't my suggestion. Leave this here as debug message, but also add a log line after JWT configuration with a single info/warn message per startup.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/JWTVerifier.java
mtneug: Ah sorry. I was thinking about the role mappings. Ignore this comment.

modules/security-jwt/src/main/java/org/opencastproject/security/jwt/DynamicLoginHandler.java
LukasKalbertodt: Oh I see. I mean, it might still bother people who do not plan on setting JWT up. But sure, I'm fine with that.

etc/security/mh_default_org.xml
mtneug: I don't know if Opencast requires the mail to be unique. Maybe a fixed mail could actually work 🤔

docs/guides/admin/docs/configuration/security.jwt.md
mtneug: My guess would be, there is some project that allows aggregating multiple JWKS URLs. If not, this shouldn't be too hard to implement. But having this in Opencast directly would simplify things for adopters.

docs/guides/admin/docs/configuration/security.jwt.md
```diff
@@ -102,79 +102,44 @@ An external application wants to let a user use Opencast Studio to record and up
 Spring Security Configuration
 -----------------------------
 
-In order to active JWT-based authentication and authorization, you will need to uncomment and adapt the following
-sections found in `etc/security/mh_default_org.xml`. Some of the options are configured with the
-[Spring Expression Language (SpEL)](https://docs.spring.io/spring-framework/docs/3.0.x/reference/expressions.html).
+In order to active JWT-based authentication and authorization, you will need to adjust your security config (e.g. `etc/security/mh_default_org.xml`).
```
owi92, line 105: I suppose this should say
```suggestion
In order to activate JWT-based authentication and authorization, you will need to adjust your security config (e.g. `etc/security/mh_default_org.xml`).
```

That typo was there before, but since you're changing this part anyway, maybe also fix that.

owi92 (approved):
You already have code reviews but I wanted to test this, and yeah... it works and is a huge improvement.

mtneug (approved):
Works as intendet.

docs/guides/admin/docs/configuration/security.jwt.md
mtneug: Oh, I missed this comment. @LukasKalbertodt can you apply this change in another PR?

LukasKalbertodt: https://github.com/opencast/opencast/pull/7196

mtneug: Thanks -> merged

PR #7204, opened 2025-11-24
Create YouTube playlists with non-public privacy status

JulianKniephoff:
This is part of opencast/roadmap#17 and adds a new configuration option to the YouTube publication service which controls the privacy status of any playlists newly created by the service. It does **not** affect existing playlists!

Some questions for @ypatios, who commissioned this feature:

- Is this enough for you or do you need this for videos as well? [Here](https://github.com/opencast/roadmap/issues/17#issuecomment-3280869877) we talked only about playlists, but in your [original post](https://github.com/opencast/roadmap/issues/17#issuecomment-2797040594) you mentioned videos as well. Note that you can already create public and private videos. What's missing is “just” an option to upload unlisted videos.
- I hope putting this towards `develop` is fine?
- Is this service level configurability enough for you or do you need this per workflow run?

PR #7205, opened 2025-11-24
Set proper language in YouTube publication

JulianKniephoff:
Part of opencast/roadmap#17. This adds the possibility to transfer the language in the media package's metadata to YouTube during publication. To this end, a group of configuration option is added.

The first controls where the language ends up in YouTube. YouTube knows about two languages (apart from captions) per video: the language of the metadata (description, title) and the default audio language. Since Opencast only knows one language in most cases, I opted to make it configurable whether we use that for the metadata language, the audio language, both, or neither.

A possible extension of this feature would maybe allow you to configure independent metadata fields for these two languages.

The rest of the settings is a list of rules to translate whatever data people have in their language DC field to the BCP-47 codes YouTube expects. This is done using regular expressions. For every BCP-47 code you might want to upload, you provide a regex that matches whatever you want to see in the DC catalog for the language.

This is a bit roundabout, but I couldn't see any other way to deal with the fact that there are no restrictions in Opencast as to what the language field can contain.

Note that this means that videos in languages that aren't matched by any of these patterns won't have its language set in YouTube!

As in #7204 I have a few questions specifically for @ypatios:

- Is this roughly what you had in mind?
- I hope putting this towards `develop` is fine?
- Is this service level configurability enough for you or do you need this per workflow run?

### DIFF
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
@@ -142,6 +142,21 @@ public class YouTubePublicationAdapter {
   }
 
   /**
+   * Gets the language for the episode of the media package
+   *
+   * @return the language of the episode
+   */
+  public String getEpisodeLanguage() {
+    if (dcEpisode != null) {
+      return dcEpisode.getFirst(DublinCore.PROPERTY_LANGUAGE);
+    }
+    if (dcSeries != null) {
+      return dcSeries.getFirst(DublinCore.PROPERTY_LANGUAGE);
+    }
+    return null;
+  }
+
+  /**
    * Parse Dublincore metadata from the workspace
    *
    * @param catalog
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
@@ -22,6 +22,7 @@
 package org.opencastproject.publication.youtube;
 
 import org.opencastproject.mediapackage.Catalog;
+import org.opencastproject.mediapackage.EName;
 import org.opencastproject.mediapackage.MediaPackage;
 import org.opencastproject.mediapackage.MediaPackageElements;
 import org.opencastproject.metadata.dublincore.DublinCore;
@@ -147,11 +148,21 @@ public class YouTubePublicationAdapter {
    * @return the language of the episode
    */
   public String getEpisodeLanguage() {
-    if (dcEpisode != null) {
-      return dcEpisode.getFirst(DublinCore.PROPERTY_LANGUAGE);
-    }
+    return getEpisodeProperty(DublinCore.PROPERTY_LANGUAGE);
+  }
+
+  /**
+   * Gets a property for the episode of the media package
+   *
+   * @param property the property to get
+   * @return the property value
+   */
+  private String getEpisodeProperty(EName property) {
     if (dcSeries != null) {
-      return dcSeries.getFirst(DublinCore.PROPERTY_LANGUAGE);
+      return dcSeries.getFirst(property);
+    }
+    if (dcEpisode != null) {
+      return dcEpisode.getFirst(property);
     }
     return null;
   }
--- a/etc/org.opencastproject.publication.youtube.YouTubeV3PublicationServiceImpl.cfg
+++ b/etc/org.opencastproject.publication.youtube.YouTubeV3PublicationServiceImpl.cfg
@@ -3,6 +3,16 @@ org.opencastproject.publication.youtube.enabled=false
 org.opencastproject.publication.youtube.category=CHANGE_ME
 org.opencastproject.publication.youtube.keywords=CHANGE_ME
 
+# The following group of properties determines how to extract the audio
+# and metadata language that is transferred to YouTube from the media package.
+# `languageTarget` can be either `audio`, `metadata`, or `both` and determines whether
+# the language is set for the metadata or the video's audio track.
+# If it's unset, no language will be transferred.
+#org.opencastproject.publication.youtube.languageTarget=both
+# For each potential BCP-47 language code that you might want to set in YouTube,
+# we need a pattern to match against the language in the media package.
+#org.opencastproject.publication.youtube.languagePatterns.fr=fra
+
 # The following property corresponds to public or private status at YouTube
 org.opencastproject.publication.youtube.makeVideosPrivate=true
 org.opencastproject.publication.youtube.defaultPlaylist=CHANGE_ME
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/VideoUpload.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/VideoUpload.java
@@ -34,6 +34,8 @@ public class VideoUpload {
 
   private final String title;
   private final String description;
+  private final String metadataLanguage;
+  private final String audioLanguage;
   private final String privacyStatus;
   private final File videoFile;
   private final MediaHttpUploaderProgressListener progressListener;
@@ -42,6 +44,8 @@ public class VideoUpload {
   /**
    * @param title may not be {@code null}.
    * @param description may be {@code null}.
+   * @param metadataLanguage may be {@code null}.
+   * @param audioLanguage may be {@code null}.
    * @param privacyStatus may not be {@code null}.
    * @param videoFile may not be {@code null}.
    * @param progressListener may be {@code null}.
@@ -50,6 +54,8 @@ public class VideoUpload {
   public VideoUpload(
       final String title,
       final String description,
+      final String metadataLanguage,
+      final String audioLanguage,
       final String privacyStatus,
       final File videoFile,
       final MediaHttpUploaderProgressListener progressListener,
@@ -57,6 +63,8 @@ public class VideoUpload {
   ) {
     this.title = title;
     this.description = description;
+    this.metadataLanguage = metadataLanguage;
+    this.audioLanguage = audioLanguage;
     this.privacyStatus = privacyStatus;
     this.videoFile = videoFile;
     this.progressListener = progressListener;
@@ -77,6 +85,22 @@ public class VideoUpload {
    */
   public String getDescription() {
     return description;
+  }
+
+  /**
+   * The video's language.
+   * The value may be {@code null}.
+   */
+  public String getMetadataLanguage() {
+    return metadataLanguage;
+  }
+
+  /**
+   * The video's audio language.
+   * The value may be {@code null}.
+   */
+  public String getAudioLanguage() {
+    return audioLanguage;
   }
 
   /**
--- a/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeAPIVersion3ServiceImpl.java
+++ b/modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubeAPIVersion3ServiceImpl.java
@@ -77,6 +77,8 @@ public class YouTubeAPIVersion3ServiceImpl implements YouTubeAPIVersion3Service 
     final VideoSnippet snippet = new VideoSnippet();
     snippet.setTitle(videoUpload.getTitle());
     snippet.setDescription(videoUpload.getDescription());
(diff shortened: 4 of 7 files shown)


modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
```diff
@@ -141,6 +141,21 @@ public String getEpisodeDescription() {
...
+  public String getEpisodeLanguage() {
+    if (dcEpisode != null) {
+      return dcEpisode.getFirst(DublinCore.PROPERTY_LANGUAGE);
+    }
+    if (dcSeries != null) {
+      return dcSeries.getFirst(DublinCore.PROPERTY_LANGUAGE);
+    }
+    return null;
+  }
+
```
JulianKniephoff, line 169: I'd actually love some input from the OG Opencasters: Is this kind of defaulting to the series catalogue something that we want? And/or is it even necessary to implement it here or would that be handled somewhere else already? 🤔

Should we also consider this for https://github.com/opencast/opencast/pull/7206?

gregorydlogan: Short term: Apathy.  To my mind, it makes sense to have this kind of defaulty behaviour.  Whether this is the right place, or whether it will be done somewhere else already without this, I'm not sure.  Not sure if this is done elsewhere tbh.

Long term: This kind of behaviour should be specified in the data model discussion.  To my mind the logic in the future should match this, but be done *not in the publication handler/adapter*.

mtneug: As Opencast admin, you can already copy metadata from the series to the episode catalog if they are not defined there using the `series` workflow operation ([see `copy-metadata` configuration key](https://docs.opencast.org/r/19.x/admin/#workflowoperationhandlers/series-woh/)). The default workflows don't seem to do that currently.

I don't really have an opinion if this defaulting behavior should be in the codebase. Although, that can somewhat blow up if it's added everywhere. I guess it makes sense here.

ypatios: thx @JulianKniephoff 
I also find this behaviour (to default to the respective series metadata, if available) sensible.

ypatios (approved):
content(audio)-language and metadata-language are a cause of confusion, for sure, but the way you handled it I think is very well-thought and gives enough flexibility.
in general I have the tendency that less is more, especially when it comes to code :D, so this solution might not be ideal on that basis. ideally we would add this, slightly more complex code initially but then possibly remove it (eg fix the option for both) if it proves to be unnecessary in practice. hope this makes sense haha.
thx.

modules/publication-service-youtube-v3/src/main/java/org/opencastproject/publication/youtube/YouTubePublicationAdapter.java
JulianKniephoff: I'll leave it in then, just because it's less work now, and also the YouTube code already does something similar to it, so it's at least locally consistent, and changing this is out of the scope of this PR, IMO. Thanks for y'all's input. 😊

ypatios:
We're testing this currently and ran into the following error:
```
2026-03-23T11:55:45,583 | ERROR | (YouTubeV3PublicationServiceImpl:396) - failed publishing to YouTube
java.lang.NullPointerException: Cannot invoke "java.util.Map.entrySet()" because "this.languagePatterns" is null
	at org.opencastproject.publication.youtube.YouTubeV3PublicationServiceImpl.publish(YouTubeV3PublicationServiceImpl.java:318) [!/:?]
	at org.opencastproject.publication.youtube.YouTubeV3PublicationServiceImpl.process(YouTubeV3PublicationServiceImpl.java:518) [!/:?]
	at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:314) [!/:?]
	at org.opencastproject.job.api.AbstractJobProducer$JobRunner.call(AbstractJobProducer.java:273) [!/:?]
```

this pointed to the new language pattern mapping and checked the config again:
https://github.com/opencast/opencast/pull/7205/changes#diff-cdb5196d845284242d754b4aa4e6be9a84c842fc4fcfda45093ebb0e51f68aaeR14

I assumed that in our case the following would work, but I might be wrong:
```
# The following group of properties determines how to extract the audio
# and metadata language that is transferred to YouTube from the media package.
# `languageTarget` can be either `audio`, `metadata`, or `both` and determines whether
# the language is set for the metadata or the video's audio track.
# If it's unset, no language will be transferred.
#org.opencastproject.publication.youtube.languageTarget=both
# For each potential BCP-47 language code that you might want to set in YouTube,
# we need a pattern to match against the language in the media package.
# [CUSTOMIZED]: all imoox-supported languages, as per requirement
org.opencastproject.publication.youtube.languagePatterns.en=en
org.opencastproject.publication.youtube.languagePatterns.de=de
org.opencastproject.publication.youtube.languagePatterns.id=id
org.opencastproject.publication.youtube.languagePatterns.bs=bs
org.opencastproject.publication.youtube.languagePatterns.ca=ca
org.opencastproject.publication.youtube.languagePatterns.es=es
org.opencastproject.publication.youtube.languagePatterns.fr=fr
org.opencastproject.publication.youtube.languagePatterns.hr=hr
org.opencastproject.publication.youtube.languagePatterns.it=it
org.opencastproject.publication.youtube.languagePatterns.nl=nl
org.opencastproject.publication.youtube.languagePatterns.pl=pl
org.opencastproject.publication.youtube.languagePatterns.pt=pt
org.opencastproject.publication.youtube.languagePatterns.sl=sl
org.opencastproject.publication.youtube.languagePatterns.fi=fi
org.opencastproject.publication.youtube.languagePatterns.sv=sv
org.opencastproject.publication.youtube.languagePatterns.tr=tr
org.opencastproject.publication.youtube.languagePatterns.el=el
org.opencastproject.publication.youtube.languagePatterns.mk=mk
org.opencastproject.publication.youtube.languagePatterns.he=he
org.opencastproject.publication.youtube.languagePatterns.ar=ar
```


Our metadata use 2-char lang pattern; see below:
```
<?xml version="1.0" encoding="UTF-8"?><dublincore xmlns="http://www.opencastproject.org/xsd/1.0/dublincore/" xmlns:dcterms="http://purl.org/dc/terms/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <dcterms:created>2026-03-23T10:50:30.592Z</dcterms:created>
    <dcterms:extent xsi:type="dcterms:ISO8601">PT4.160S</dcterms:extent>
    <dcterms:isPartOf>b24d63f0-0a2b-4eee-a62c-2430f5004c91</dcterms:isPartOf>
    <dcterms:language>de</dcterms:language>
    <dcterms:license>CC-BY</dcterms:license>
    <dcterms:publisher>Me Myself</dcterms:publisher>
    <dcterms:temporal xsi:type="dcterms:Period">start=2026-03-23T10:50:30Z; end=2026-03-23T10:50:30Z; scheme=W3C-DTF;</dcterms:temporal>
    <dcterms:title>test yt upload only vid</dcterms:title>
</dublincore>
```

if @Arnei or @JulianKniephoff have a quick hint, that would be great :) else I'll try to troubleshoot this myself.
thx :)

JulianKniephoff:
I'm having a look right now. @ypatios can you post the entire config?

JulianKniephoff:
And also maybe the log?

Like, the only reason `languagePatterns` would be `null` there is for the YT service not to initialize correctly. Either it fails or it's disabled in the config. 🤔

ypatios:
> Like, the only reason `languagePatterns` would be `null` there is for the YT service not to initialize correctly. Either it fails or it's disabled in the config. 🤔

thx for the pointers, @JulianKniephoff. We created a new test yt-account for this; I'm not sure if it works yet, so I'll have a look at the logs, maybe restart oc, see what happens.
it's definitely activated in the config – which I'm posting here nevertheless

```
org.opencastproject.publication.youtube.enabled=true

org.opencastproject.publication.youtube.category=Education
org.opencastproject.publication.youtube.keywords=iMooX,MOOC,Online Course,Education,Online Education,Educational Video,tugraz,Austria,Digitale Bildung,Lernvideo,eLearning,TEL,Open Education,Open Educational Resources,OER,Learning,Teaching,Gratis,Freie Bildungsressourcen,Bildung,Bildungsvideo,Lernen,Lehre,Unterricht,Hochschule,Universität,free,digital

# The following group of properties determines how to extract the audio
# and metadata language that is transferred to YouTube from the media package.
# `languageTarget` can be either `audio`, `metadata`, or `both` and determines whether
# the language is set for the metadata or the video's audio track.
# If it's unset, no language will be transferred.
#org.opencastproject.publication.youtube.languageTarget=both
# For each potential BCP-47 language code that you might want to set in YouTube,
# we need a pattern to match against the language in the media package.
# [CUSTOMIZED]: all imoox-supported languages, as per requirement
org.opencastproject.publication.youtube.languagePatterns.en=en
org.opencastproject.publication.youtube.languagePatterns.de=de
org.opencastproject.publication.youtube.languagePatterns.id=id
org.opencastproject.publication.youtube.languagePatterns.bs=bs
org.opencastproject.publication.youtube.languagePatterns.ca=ca
org.opencastproject.publication.youtube.languagePatterns.es=es
org.opencastproject.publication.youtube.languagePatterns.fr=fr
org.opencastproject.publication.youtube.languagePatterns.hr=hr
org.opencastproject.publication.youtube.languagePatterns.it=it
org.opencastproject.publication.youtube.languagePatterns.nl=nl
org.opencastproject.publication.youtube.languagePatterns.pl=pl
org.opencastproject.publication.youtube.languagePatterns.pt=pt
org.opencastproject.publication.youtube.languagePatterns.sl=sl
org.opencastproject.publication.youtube.languagePatterns.fi=fi
org.opencastproject.publication.youtube.languagePatterns.sv=sv
org.opencastproject.publication.youtube.languagePatterns.tr=tr
org.opencastproject.publication.youtube.languagePatterns.el=el
org.opencastproject.publication.youtube.languagePatterns.mk=mk
org.opencastproject.publication.youtube.languagePatterns.he=he
org.opencastproject.publication.youtube.languagePatterns.ar=ar

# This property contains a regular expression that the license of a published
# media package has to match in order to be given the creative commons license
# on YouTube.
#org.opencastproject.publication.youtube.ccLicenses=CC.*

# The following property corresponds to public or private status at YouTube
org.opencastproject.publication.youtube.makeVideosPrivate=true
# The following property corresponds to the privacy status of any newly created playlist.
# Acceptible values are public, private and unlisted
# [CUSTOMIZED]: new playlists set to private per default, as per requirement
org.opencastproject.publication.youtube.playlistPrivacy=private
org.opencastproject.publication.youtube.defaultPlaylist=CHANGE_ME

# The next property reflects a YouTube rule concerning playlist and video titles.
# Those YouTube properties correspond to Opencast's series and recording titles.
# The current value is equal to YouTube's current limit.
# [CUSTOMIZED]
org.opencastproject.publication.youtube.maxFieldLength=90

# Name for the data store
org.opencastproject.publication.youtube.credentialDatastore=store

# Auxillary configs live under storage.dir
org.opencastproject.publication.youtube.clientSecretsV3=/home/opencast/youtube/client_secret.json
org.opencastproject.publication.youtube.dataStore=/home/opencast/youtube/data-store

# The load on the system introduced by creating a publish job. Each job involves copying the output file to Youtube
# which can be expensive depending on file size. Since this will fairly quickly add up, these should be relatively
# expensive, while not crippling the system.
# Default: 0.1
#job.load.youtube.publish=0.1

# The load on the system introduced by creating a retract job. Each job involves instructing Youtube to delete the
# media. This is a quick and inexpensive operation, so we can run a lot of these in parallel.
# Default: 0.1
#job.load.youtube.retract=0.1
```

And here the logs from the last startup (respective lines of modules during initiation):
```
2026-03-23T09:09:54,660 | INFO  | (BundleInfoLogger:120) - Bundle opencast-distribution-workflowoperation, id 100, version 15.6.0, build number cf7a306
2026-03-23T09:09:54,734 | INFO  | (BundleInfoLogger:120) - Bundle opencast-publication-service-api, id 149, version 15.6.0, build number cf7a306
2026-03-23T09:09:54,738 | INFO  | (BundleInfoLogger:120) - Bundle opencast-publication-service-youtube-v3, id 152, version 15.6.0, build number cf7a306
```

the 4th module (remote) doesn't get initiated prob due to allinone variant of test system.

JulianKniephoff:
> the 4th module (remote) doesn't get initiated prob due to allinone variant of test system.

Yup, and it shouldn't be necessary for this to work, either.

gregorydlogan (approved):

PR #7206, opened 2025-11-24
Set YouTube license according to metadata

JulianKniephoff:
This is part of opencast/roadmap#17, and it allows you to set the license of a YouTube video while publishing there, according to the videos metadata in Opencast.

Since there are only two licenses on YouTube, namely YouTube Standard and Creavie Commons, and the former is the standard, all we need is a way to identify the CC licenses. Since the license field in the DC catalogs is basically just text, we provide a regex that must be matched in order for the video to have its license set to CC. This is similar to how we do languages in #7205.

As with the other YouTube PRs, some Questions for @ypatios:

- Is this what you had in mind?
- I hope putting this towards `develop` is fine?
- Is this service level configurability enough for you or do you need this per workflow run?


ypatios (approved):
confirmed, there are only two(2) options for licences (sorry, I use the British spelling but in the code it's good that u used US spelling for consistency), and the solution with regex is optimal (as we can have different types of CC licences in OC).
thx.

gregorydlogan (approved):

PR #7214, opened 2025-11-27
Deprecate Paella 7 with OC 20

gregorydlogan:
As discussed in the roadmap meeting today, Paella 7 feature development has ceased.  I therefore propose that we deprecate Paella 7 with Opencast 20, and remove it with Opencast 21.


gregorydlogan:
As discussed in the technical meeting today, given that we want to remove paella 8 with 21 we need to update default the release *before* we remove 7.  Since we don't want to spring a surprised default change with 19, that makes OC 20 when the default will change.

PR #7245, opened 2025-12-03
Store episode catalog in snapshot to decrease re-indexing time

JamesUoM:
A lot of the time re-indexing the Asset Manager is spent reading episode catalog files from disk. Depending on the storage type this can be prohibitively slow. This pull request adds the episode catalog to the snapshot entry in the database. During indexing, it will attempt to load the catalog from the database, if not found, it will read it from storage. All new snapshots will attempt to store the episode catalog.

A new REST endpoint `PUT /asset/snapshot/catalog` can be used to populate previous snapshots (latest version).

This PR has been running at Manchester for several years under 3.x, then ported to 12.x then 16.x. I've had to change the database access routines to match changes since 16.x.

### How to test this patch

* Create some asset snapshots before applying the PR, so you have some without the episode catalog. If you can create enough you could time the AssetManager re-index for comparison with the PR.
* Against an existing deployment, you will need to run the appropriate update script for your database to modify the snapshot table.
* Create some new assets, and re-index. The PR will fallback to reading from disk if the catalog is not in the snapshot 
* run  `PUT /asset/snapshot/catalog` and reindex - you could remove some episode catalogs to prove it is reading the catalog from.

With luck you will see any improvement in asset manager re-indexing speed.


docs/upgrade/19_to_20/mariadb.sql
```diff
@@ -0,0 +1,2 @@
+-- add catalog storage to snapshots
+ALTER TABLE `oc_assets_snapshot` ADD `episode_xml` longtext DEFAULT NULL;
```
mtneug, line 2: Call it `episode_catalog` instead of `episode_xml`? I first thought this was the mediapackage manifest XML that is already in the DB.

JamesUoM: I called it `_xml`  to be consistent with the mediapackage column naming :shrug:

gregorydlogan:
I haven't really reviewed this past reading the release notes blurb, but I have to wonder if it makes sense to have a script which runs through the media packages in the system and does the relevant database inserts, rather than doing this in an on-change manner.

My reasoning is that if we leave this as is we're going to have the code constantly checking the db, then the filesystem. Forever. When could we safely remove this in the future? Is there a usecase where the relevant catalogs are *supposed to be* on disk? Supporting this fallback adds complexity for (imo) little gain. But then I'm also not an adopter with thousands of catalogs in inaccessible cold storage...

JamesUoM:
> I haven't really reviewed this past reading the release notes blurb, but I have to wonder if it makes sense to have a script which runs through the media packages in the system and does the relevant database inserts, rather than doing this in an on-change manner.

Agree that the endpoint is just a "use-once" feature but a script has to support the different asset stores. I wrote the endpoint, so I didn't have re-do all the file/object access code. Maybe move it to a migration bundle?

The code will always need to check the DB to get the catalog - it's only when it fails it does something more than throwing an exception.

gregorydlogan:
A migration bundle isn't unprecedented, and could be handy for some adopters.  Do you have time available to implement that?

JamesUoM:
> A migration bundle isn't unprecedented, and could be handy for some adopters. Do you have time available to implement that?

More than doing it via a script...

mtneug:
A migration script for an update is not unprecedented for Opencast. The question is if this should be mandatory or not, but in general I'm fine with a migration script instead of a full Java module implementation.

JamesUoM:
> A migration script for an update is not unprecedented for Opencast. The question is if this should be mandatory or not, but in general I'm fine with a migration script instead of a full Java module implementation.

For the reasons given above I'm not writting a migration script.

JamesUoM:
I think add ingthe bundle as part of "migration plugin" would be the tidiest way, it is then easy for users to enable/disable it.

gregorydlogan:
@JamesUoM are you likely to have more time to work on this prior to the branch cut for 21?

JamesUoM:
> @JamesUoM are you likely to have more time to work on this prior to the branch cut for 21?

Updated for a 21 release

rute-santos:
I've tested this, but in our OC 15.x codebase, and it works great. Thanks @JamesUoM !

PR #7249, opened 2025-12-03
Add JWT support to `/paella7/ui` (with token refresh via iframe host)

LukasKalbertodt:
This makes the built-in player capable of authorizing requests via JWT. Well, it is a "first viable product", see todo-list below. 
- The `jwt` query parameter can be passed (i.e. `/paella7/ui/watch.html?id=...&jwt=...`)
- That given JWT is used when requesting the search API to get event information.
- A service worker is installed, which uses [`jwtify`](https://github.com/opencast/jwtify) to intercept all requests to static files (e.g. thumbnail, video, ...) and authenticate them.
- If the `jwtRefresh=true` parameter is given, the service worker can request fresh JWTs by posting a message to the iframe host, which can then reply with a fresh JWT. 
- If refresh is not enabled, it is assumed that the initial JWT is valid for a long time and that is used for all requests. Once it expires, static file requests will start returning 403, breaking the page.

## Documentation for integration developer

... ok let's be real, I am talking about @ferishili here :P

**Note**: this is a draft PR and thus, this "protocol" might still change a bit. So don't release/ship any code implementing this yet, until we decided on this.

The iframe route accepts two new query parameters:
- `jwt`: an initial JWT which is used to load data from the search API, and potentially for other early requests like the thumbnail or the first video request. 
- `jwtRefresh`: if `true`, refresh is enabled.

If token refresh is enabled, the iframe will send the following message:

```js
{
    type: "oc-event-jwt-request", // this is always this exact string
    event: "19be2fb4-0241-4947-a6e0-28873e857a7d", // the event ID
}
```

You (the host of the iframe) are then supposed to answer with this message:

```js
{
    type: "oc-event-jwt", // always this exact string
    event: "19be2fb4-0241-4947-a6e0-28873e857a7d", // event ID, the same that was passed to you
    jwt: "ey...", // the JWT 
}
```

The JWT needs to give read access to the event in question, and do it in a way that `octoka` understands, which basically means: via the `"oc": { "e:<id>": ["read"] }` claim.

Here is an example for dealing with reacting to messages:

<details>

It is assumed that `fetchJwt` is `async (eventId: string) => string`.

```js
  window.addEventListener("message", ev => {
    // TODO: make sure it's from the expected iframe! Excercise for the reader.
    if (typeof ev.data === "object" && ev.data?.type === "oc-event-jwt-request" && typeof ev.data?.event === "string") {
      fetchJwt(ev.data.event).then(jwt => {
        ev.source.postMessage({
          type: "oc-event-jwt",
          event: ev.data.event,
          jwt,
        }, '*');
      });
    }
  });
```

</details>

## TODO

While the central logic already works, there is still lots to do before we can merge it:
- [ ] Make whole thing depend on config variable: I assume we probably don't want to enable this outright for everyone, but make it opt-in first. 
  - This might be a bit tricky, as fetching this config value must happen before registering the service workers, which must happen before the first static file request is sent.
  - In Tobira, I also made sure that when this is disabled, all current service workers are uninstalled, as otherwise it might stay in a broken state.
- [ ] Disable all of this new stuff when no JWT is given as query parameter
  - Or is this enough as a "guard" and the previous point about configuring this can be ignored?
- [ ] Maybe make "protocol" immediately able to deal with multiple event IDs? 
  - `jwtify` already uses "request batches" in its API, and it might be useful to also have that for this iframe<->host protocol so that the host can reduce the number of fetches it needs to do. On the other hand, I cannot currently imagine when that would ever happen? When would an iframe feature content from multiple videos?
- [ ] Implement for Paella 8 as well
- [ ] Correctly set `trustedHosts`
  - This is a protection by jwtify to not send JWTs to random hosts, which might have been injected into the media package by an attacker. 
  - Decide: is that really an attack worth protecting against? In the worst case, the attacker gets a JWT that grants access to a single event, but the attacker must have already manipulated said event for that to work. Soooo maybe not a big deal? We could also say that "being able to put non-Opencast URLs into media packages" is the core problem and that everyone else should be able to trust media packages.
  - If we decide to protect against it: get a list of trusted hosts in the code and pass it to jwtify. Again, might be tricky timing wise. Maybe jwtify needs to be adjusted.
  - If we decide to ignore this, we need to adjust jwtify to not require this trustedHosts list.
- [ ] Check if other non-static-file requests need to be authenticated. I just dealed with the search API requests, but haven't gone through the code to see if there are other cases.

## Open Problems & Questions
- The search endpoint is requested with the JWT. But unlike static files, which will be served by octoka, Opencast currently still requires a username and name. Typically, static file JWTs do not include them, as they are useless for the task. Either the host site (e.g. LMS) needs to pass these two values in the JWT, or Opencast needs to be able to accept the JWT even without that info. For local testing, 
  - I configured the `usernameMapping` in the security config to `['sub'] ?: 'unknown-jwt-user'`. I wonder: is that safe for production systems? The problem with the username is, that it is (sometimes?) used to look up additional permissions?! Someone needs to look into that.

### How to test this patch

(This is still draft and no one should do a full test yet. This is mostly for Farbod to play around with now.) 

Build it, configure JWT (luckily much easier now :>), upload videos, and well... start experimenting with the `/paella7/ui/watch.html` URL. Pass a valid JWT to it and you should see the thumbnail loading even if you are not logged in. And then you probably have to build your own iframe stuff to check the refresh code.

### DIFF
--- a/modules/engage-paella-player-7/src/index.js
+++ b/modules/engage-paella-player-7/src/index.js
@@ -52,3 +52,21 @@ window.onload = async () => {
     paella.log.error(error);
   }
 };
+
+
+// Setup service worker to enable static file auth via JWT. We pass JWT and
+// video ID via query parameters in order to have them available in the SW
+// immediately. I know that typically, you want an unchanging URL for your SWs,
+// but that only applies to service workers that want to have a persistent state
+// as far as I can tell. This query parameter passing method is known and used.
+// Also, the SW is super tiny, so the additional fetches don't really matter.
+const params = new URLSearchParams(window.location.search);
+const swUrl = `sw.js?${new URLSearchParams({
+  jwt: params.get('jwt'),
+  videoId: params.get('id'),
+})}`;
+
+navigator.serviceWorker
+  .register(swUrl, { updateViaCache: 'none' })
+  // eslint-disable-next-line no-console
+  .catch(e => console.error('Failed to register service worker', e));
--- a/modules/engage-paella-player-7/src/index.js
+++ b/modules/engage-paella-player-7/src/index.js
@@ -61,12 +61,47 @@ window.onload = async () => {
 // as far as I can tell. This query parameter passing method is known and used.
 // Also, the SW is super tiny, so the additional fetches don't really matter.
 const params = new URLSearchParams(window.location.search);
+let refreshEnabled = params.get('jwtRefresh') === 'true';
+if (refreshEnabled && window.self === window.top) {
+  // eslint-disable-next-line no-console
+  console.warn('jwtRefresh=true given, but not running in iframe');
+  refreshEnabled = false;
+}
 const swUrl = `sw.js?${new URLSearchParams({
   jwt: params.get('jwt'),
   videoId: params.get('id'),
+  refresh: refreshEnabled,
 })}`;
 
 navigator.serviceWorker
   .register(swUrl, { updateViaCache: 'none' })
   // eslint-disable-next-line no-console
   .catch(e => console.error('Failed to register service worker', e));
+
+
+
+// If JWT refresh is enabled, we need to forward messages between the host of
+// this iframe and the SW (as they cannot communicate directly).
+if (refreshEnabled) {
+  // Forward messages coming from the iframe host to the SW.
+  window.addEventListener('message', ev => {
+    if (ev.source !== window.parent) {
+      return;
+    }
+
+    const d = ev.data;
+    if (typeof d === 'object' && d?.type === 'oc-event-jwt' && typeof d?.jwt === 'string') {
+      const { type, jwt } = d;
+      navigator.serviceWorker.controller?.postMessage({ type, jwt });
+    }
+  });
+
+  // Forward messages coming from the SW to the iframe host.
+  navigator.serviceWorker.addEventListener('message', ev => {
+    const d = ev.data;
+    if (typeof d === 'object' && d?.type === 'oc-event-jwt-request' && typeof d?.event === 'string') {
+      const { type, event } = d;
+      window.parent.postMessage({ type, event }, '*');
+    }
+  });
+}
--- a/modules/engage-paella-player-7/src/js/sw/index.js
+++ b/modules/engage-paella-player-7/src/js/sw/index.js
@@ -21,20 +21,73 @@
 
 import { setUpServiceWorker } from '@opencast/jwtify';
 
-// Get initial parameters (containing video ID and JWT)
+// Get initial parameters
 const params = new URLSearchParams(self.location.search);
 const jwt = params.get('jwt');
 const videoId = params.get('videoId');
+const refreshEnabled = params.get('refresh') === 'true';
 
 // Only of those are set do we actually register anything.
 if (jwt && videoId) {
-  const fetchJwts = async (eventIds) => {
-    return new Map(eventIds.has(videoId) ? [[videoId, jwt]] : []);
+  let lastClientId = null;
+  let resolveRefresh = () => {};
+  if (refreshEnabled) {
+    // Remember the last client that sent a request. We will always ask it for
+    // fresh JWTs.
+    self.addEventListener('fetch', event => {
+      lastClientId = event.clientId;
+    });
+
+    // Receive messages from clients, which will send JWTs back.
+    self.addEventListener('message', e => {
+      const d = e.data;
+      if (typeof d === 'object' && d?.type === 'oc-event-jwt' && typeof d?.jwt === 'string') {
+        resolveRefresh(d.jwt);
+      }
+    });
+  }
+
+  const fetchJwts =  async (eventIds) => {
+    if (eventIds.size !== 1 || !eventIds.has(videoId)) {
+      // eslint-disable-next-line no-console
+      console.warn('event IDs requested by jwtify do not match video ID');
+      return new Map();
+    }
+
+    if (!refreshEnabled) {
+      // eslint-disable-next-line no-console
+      console.warn('JWT refresh is not enabled, but jwtify asked for a new JWT, meaning the old one is expired');
+      return new Map();
+    }
+
+    const client = await self.clients.get(lastClientId) ?? self.clients.matchAll()[0];
+    if (!client) {
+      // eslint-disable-next-line no-console
+      console.warn('No client connected to SW, but a JWT is needed for a request... weird');
+      return new Map();
+    }
+
+    // Get fresh JWT by asking the host frame (by first asking the client).
+    const jwt = await new Promise(resolve => {
+      resolveRefresh = resolve;
+      client.postMessage({
+        type: 'oc-event-jwt-request',
+        event: videoId,
+      });
+
+      // After some timeout we give up, as there might not be anyone answering.
+      setTimeout(() => resolve(null), 3000);
+    });
+
+    return new Map(jwt ? [[videoId, jwt]] : []);
   };
 
-  setUpServiceWorker({
+  const handle = setUpServiceWorker({
     getJwts: fetchJwts,
     trustedOcOrigins: ['https://stable.opencast.org', 'http://localhost:8080', 'http://localhost:8081'], // TODO
     debugLog: true, // TODO: remove
   });
+
+  // Add our initial JWT to the cache so that it can be used for immediate requests.
+  handle.cache.add(videoId, jwt);
 }
--- a/modules/engage-paella-player-7/src/js/sw/index.js
+++ b/modules/engage-paella-player-7/src/js/sw/index.js
@@ -1,0 +1,40 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+import { setUpServiceWorker } from '@opencast/jwtify';
+
+// Get initial parameters (containing video ID and JWT)
+const params = new URLSearchParams(self.location.search);
+const jwt = params.get('jwt');
+const videoId = params.get('videoId');
+
+// Only of those are set do we actually register anything.
+if (jwt && videoId) {
+  const fetchJwts = async (eventIds) => {
+    return new Map(eventIds.has(videoId) ? [[videoId, jwt]] : []);
+  };
+
+  setUpServiceWorker({
+    getJwts: fetchJwts,
+    trustedOcOrigins: ['https://stable.opencast.org', 'http://localhost:8080', 'http://localhost:8081'], // TODO
+    debugLog: true, // TODO: remove
+  });
+}
--- a/modules/engage-paella-player-7/package-lock.json
+++ b/modules/engage-paella-player-7/package-lock.json
@@ -6,6 +6,7 @@
     "": {
       "name": "paella-7-opencast",
       "dependencies": {
+        "@opencast/jwtify": "^0.1.3",
         "paella-basic-plugins": "^1.50.3",
         "paella-core": "^1.50.4",
         "paella-mp4multiquality-plugin": "^1.47.1",
@@ -1977,6 +1978,12 @@
       "dependencies": {
         "eslint-scope": "5.1.1"
       }
+    },
+    "node_modules/@opencast/jwtify": {
+      "version": "0.1.3",
+      "resolved": "https://registry.npmjs.org/@opencast/jwtify/-/jwtify-0.1.3.tgz",
+      "integrity": "sha512-aC2Z/3lRR5OPGG1FldpNlXE/cKJIeCzlsFNOSUfAKWQ62glil9fO15BaHkGb2uBtiR7VI17q7M5dtDZVpShoFw==",
+      "license": "MIT"
     },
     "node_modules/@playwright/test": {
       "version": "1.55.1",
--- a/modules/engage-paella-player-7/package.json
+++ b/modules/engage-paella-player-7/package.json
@@ -36,6 +36,7 @@
     "xml-loader": "^1.2.1"
   },
   "dependencies": {
+    "@opencast/jwtify": "^0.1.3",
     "paella-basic-plugins": "^1.50.3",
     "paella-core": "^1.50.4",
     "paella-mp4multiquality-plugin": "^1.47.1",
--- a/modules/engage-paella-player-7/src/js/PaellaOpencast.js
+++ b/modules/engage-paella-player-7/src/js/PaellaOpencast.js
@@ -105,7 +105,12 @@ const initParams = {
   repositoryUrl: getUrlFromOpencastServer('/search/episode.json'),
 
   getManifestUrl: (repoUrl, videoId) => {
-    return `${repoUrl}?id=${videoId}`;
+    let out =  `${repoUrl}?id=${videoId}`;
+    const jwt = new URLSearchParams(location.search).get('jwt');
+    if (jwt) {
+      out += `&jwt=${jwt}`;
+    }
+    return out;
   },
 
   getManifestFileUrl: (manifestUrl) => {
--- a/modules/engage-paella-player-7/src/js/PaellaOpencast.js
+++ b/modules/engage-paella-player-7/src/js/PaellaOpencast.js
@@ -163,7 +163,7 @@ const initParams = {
       const data = await fetch(getUrlFromOpencastServer('/info/me.json'));
       const me = await data.json();
 
-      if (!me.roles.includes('ROLE_USER')) {
+      if (!me.roles.includes('ROLE_USER') && !new URLSearchParams(location.search).get('jwt')) {
         player.log.info('Video not found and user is not authenticated. Try to log in.');
         await redirectToAuthPage();
         // This Error should not happen, as the user is redirected to the auth page.
--- a/modules/engage-paella-player-7/webpack.config.js
+++ b/modules/engage-paella-player-7/webpack.config.js
@@ -33,7 +33,13 @@ module.exports = function (env) {
   };
 
   return {
-    entry: './src/index.js',
+    entry: {
+      main: './src/index.js',
+      sw: {
+        import: './src/js/sw/index.js',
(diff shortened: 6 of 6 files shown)


modules/engage-paella-player-7/src/index.js
```diff
@@ -52,3 +52,56 @@ window.onload = async () => {
...
+
+    const d = ev.data;
+    if (typeof d === 'object' && d?.type === 'oc-event-jwt' && typeof d?.jwt === 'string') {
+      const { type, jwt } = d;
+      navigator.serviceWorker.controller?.postMessage({ type, jwt });
+    }
+  });
+
+  // Forward messages coming from the SW to the iframe host.
+  navigator.serviceWorker.addEventListener('message', ev => {
```
mtneug, line 100: Question: Can we be sure this is from the jwtify SW? Should this be somehow verified like the check above for the other direction?

modules/engage-paella-player-7/src/js/sw/index.js
```diff
@@ -0,0 +1,93 @@
...
+      // After some timeout we give up, as there might not be anyone answering.
+      setTimeout(() => resolve(null), 3000);
+    });
+
+    return new Map(jwt ? [[videoId, jwt]] : []);
+  };
+
+  const handle = setUpServiceWorker({
+    getJwts: fetchJwts,
+    trustedOcOrigins: ['https://stable.opencast.org', 'http://localhost:8080', 'http://localhost:8081'], // TODO
```
mtneug, line 87: This is tricky. You definitely want to limit where the JWT is attached, but it's not so easy to determine what origins are ok. E.g. when you have `https://stable.opencast.org` as main Opencast URL to talk to its APIs, you could have `https://files.opencast.org` for serving static files. I don't see a way to determine this list other than letting users configure this. At the same time, I wouldn't remove this guard. Paella Player might make all kinds of requests to other origins, e.g. to Matomo.

mtneug:
> ... ok let's be real, I am talking about @ferishili here :P

While I'm not implementing the host side for LMS plugins, I still requested this protocol description. IMO the final protocol should be documented in the Opencast dev documentation. We serve a custom player and need to correctly implement the iFrame side.

> * The search endpoint is requested with the JWT. But unlike static files, which will be served by octoka, Opencast currently still requires a username and name. Typically, static file JWTs do not include them, as they are useless for the task. Either the host site (e.g. LMS) needs to pass these two values in the JWT, or Opencast needs to be able to accept the JWT even without that info. For local testing,
>   
>   * I configured the `usernameMapping` in the security config to `['sub'] ?: 'unknown-jwt-user'`. I wonder: is that safe for production systems? The problem with the username is, that it is (sometimes?) used to look up additional permissions?! Someone needs to look into that.

Given a username, all role providers are asked to contribute roles for this user. We probably need something like a special `nobody` user that only exists in-memory and has no role except the one given by the JWT during the single request cycle. In particular, I wouldn't store it as a user reference.

mtneug:
Overall, I think the proposed protocol is usable and covers various concerns.

The one thing that's still left open for me is the "embed public video to static site" use case. In theory, you could create a JWT without an expiration date for that specific video. What do you think?

LukasKalbertodt:
>  IMO the final protocol should be documented in the Opencast dev documentation

Yes.

> Given a username, all role providers are asked to contribute roles for this user. We probably need something like a special nobody user that only exists in-memory and has no role except the one given by the JWT during the single request cycle. In particular, I wouldn't store it as a user reference.

I created https://github.com/opencast/opencast/issues/7253 

> The one thing that's still left open for me is the "embed public video to static site" use case. In theory, you could create a JWT without an expiration date for that specific video. What do you think?

JWT without exp is dangerous, as this means the owner can effectively never make the video private again. I would just say to not include a JWT at all. Opencast will grant all requests because it knows, the video is public. When using octoka, it itself cannot grant access, but will fall back to Opencast, so all is good. Only problem is when you want to run octoka without Opencast fallback.

mtneug:
> Only problem is when you want to run octoka without Opencast fallback.

That is the mode we are targeting :)

Well, embedding the Tobira player could be an official alternative and we will find a custom way if we need that.

ferishili:
FYI: I just got this running in the ILIAS Opencast plugin 🎉

<del>One thing that still needs to be addressed is that the segment thumbnails fail to load in the Opencast Paella Player on the first hover. After that initial hover, they display correctly.</del>
**[UPDATE]**: This problem was gone in today’s tests (magic!). I believe the cause was the username mapping issue, which I fixed not in Opencast itself, but by adding it to the claims sent from the ILIAS Opencast plugin.

ferishili:
A major issue:

The search endpoint now requires a JWT. That part is fine, but it only accepts the token in the header, not as a query parameter, which does not work for our use case.

In ILIAS, we call this endpoint directly for livestreams to retrieve livestream information. Now it fails when called without JWT claims, and when we include a JWT, it returns an empty result.

gregorydlogan:
This should be fixable I would think, though not in 17. Are there other endpoints which do it this way that we can crib from?

ferishili:
So far, this only affects the search endpoint, specifically `/search/episode.json`, if that helps.
I am trying to go step by step for this, from the LMS point of view, I might bump to something else!

We only support OC 18 and higher for functional JWT, so Opencast 17 does not matter here!

ypatios:
guys, I was looking for sth else and stumbled upon this draft pr and I have to say...
this is exemplary: from the super-informative description (a pr that includes its own project management haha) to the discussion and testing, what can I say: just reading through it is exhilarating.
hats off

ferishili:
With building and testing this PR on my test Opencast instance, I was able to verify two major issues:

### First issue

With the simplified JWT configuration introduced in **19.x**, it is still necessary to provide a fallback value for the `nameMapping` property:

```
<property name="nameMapping" value="['name'] ?: 'unknown-jwt-user'" />
```

Without this fallback, requests to `/play` via an iframe always result in:

```
HTTP ERROR 500
java.lang.IllegalArgumentException: Extracted name is blank
```

This indicates that the extracted JWT `name` claim is still required and cannot be empty, even with the simplified configuration.

Screenshot 2026-04-08 at 15 20 48

---

### Second issue

After setting the fallback value for `nameMapping` (e.g. `unknown-jwt-user`) and successfully reaching `/play` inside an iframe, the player displays the well-known large red **Paella Player error screen**.

Further investigation shows that the issue is caused by the request:

```
search/episodes.json?id=...
```

which consistently returns an empty array instead of the expected episode data.

As a result, the player cannot load the recording.

Screenshot 2026-04-08 at 15 22 37


PS: This issue is currently blocking all LMS integrations, so resolving it as soon as possible would be greatly appreciated.

JulianKniephoff:
@ferishili just for my own understanding: The first issue in your last comment is basically https://github.com/opencast/opencast/issues/7253, right?

JulianKniephoff:
Also @ferishili can you provide a bit more context and/or a few more details about the second issue? I can query the search endpoint with a jwt query parameter just fine. 🤔

ferishili:
> Also @ferishili can you provide a bit more context and/or a few more details about the second issue? I can query the search endpoint with a jwt query parameter just fine. 🤔

The second issue, **although I assume both issues are related**, is that the search service still tries to match user roles against the Event/Series ACLs in the same way as before. However, this approach is no longer suitable when using JWT-based authorization.

In the JWT case, the search service receives roles such as `"e:{identifier}" => ["read"]`, but these do not match the existing event ACL roles like `ROLE_USER`, `ROLE_ADMIN`, etc. As a result, access checks fail.

As a workaround in ILIAS that I did, I created a dedicated JWT user in Opencast:

* Username: `unknown-jwt-user`
* Email: `no-mail@jwt.invalid`
* Role: `ROLE_JWT_USER` (which was also the default suggested role)

I then added the role `ROLE_JWT_USER` to the ACLs of the events.

This workaround functions, but it unfortunately keeps the JWT integration role-based. In practice, this means that only newly uploaded events that include the effective role `ROLE_JWT_USER` in their ACLs are accessible. Older events that do not contain this role return the message:

> `Event not found or user does not have access`

mtneug:
The JWT is subject to a role mapping which is configurable. Something like `"e:{identifier}" => ["read"]` should (now by default) add `ROLE_EPISODE_{identifier}_read` to the user role list. This should give this user access to this event even in search.

ferishili:
> The JWT is subject to a role mapping which is configurable. Something like `"e:{identifier}" => ["read"]` should (now by default) add `ROLE_EPISODE_{identifier}_read` to the user role list. This should give this user access to this event even in search.

That is true, but keep in mind that the initial call is to `/play/[id]` + jwt via an iframe! therefore whatever is missing is between opencast player (?)  <-> opencast search service!

JulianKniephoff:
`/play` is basically a redirect, and it doesn't pass through arbitrary URL parameters. Funnily enough, I stumbled over this in an unrelated context the other day. Did you verify that the `jwt` query parameter actually reaches the player? 🤔

ferishili:
> `/play` is basically a redirect, and it doesn't pass through arbitrary URL parameters. Funnily enough, I stumbled over this in an unrelated context the other day. Did you verify that the `jwt` query parameter actually reaches the player? 🤔

If I remember correctly, when I was implementing the integration for ILIAS, the `/play` was not working properly, as you rightly mentioned. Because of that, I eventually switched to `/paella7/ui/watch.html`. However, if the issue could be fixed, it would give us the major advantage of no longer having to care about whether the installation uses Paella 7, 8, 9, and so on.

But first things first: let's fix the search service so it can handle JWT roles within Opencast after calling either `/play` or `/paella7/...` in an iframe!

JulianKniephoff:
I want to tackle the `/play` thing as well, but I would say that's out of scope for this PR. So you are saying it still didn't work after you switched to `/paella7/...`?.

ferishili:
> So you are saying it still didn't work after you switched to `/paella7/...`?

😅 like 100 times so far! Search Service returns `[]`!

gregorydlogan:
I'm assuming this moves to 20+?

JulianKniephoff:
Is Paella 7 even still in OC 20? I know @snoesberger wants this in 19, right? ;D Would that be against our rules rn? I always forget. 🫥

snoesberger:
> Is Paella 7 even still in OC 20? I know @snoesberger wants this in 19, right? ;D Would that be against our rules rn? I always forget. 🫥

Yes, Paella 7 is still in OC 20, but Paella 8 is now the default player. We'll skip OC 19 and update directly to OC 20, therefore, from our point of view, this PR could also go into OC 20.

gregorydlogan:
Punted this to 20.x, however conflicts.  And draft status...

KatrinIhler:
Note from internal elan meeting: still needed bc partitioned cookies not a solution? (as told by @ferishili ) @mtneug can you elaborate on that?

LukasKalbertodt:
To explain our question a bit further: can you explain how partitioned cookies do _not_ solve all use cases? Apparently you tested something where they don't work. It would be great if you could explain the situation. Knowing that CHIPS don't solve everything and that this solution is actually still needed, we can hopefully finally get this PR over the finish line.

mtneug:
> Note from internal elan meeting: still needed bc partitioned cookies not a solution? (as told by @ferishili ) @mtneug can you elaborate on that?

> To explain our question a bit further: can you explain how partitioned cookies do _not_ solve all use cases? Apparently you tested something where they don't work. It would be great if you could explain the situation. Knowing that CHIPS don't solve everything and that this solution is actually still needed, we can hopefully finally get this PR over the finish line.

ok this probably got lost (changed) in translation...

I don't remember saying that CHIPS does not solve the third party cookie problem for Opencast (LTI) sessions. It probably should. Although, there is the problem of limited implementation in older Safari versions (since Safari intentionally misinforms about the macOS version in the user string, it's a bit hard to estimate how may users are using older Safari versions).

What we actually tested was the JWT implementation of the Opencast Moodle plugins against an upstream Opencast version, i.e. without this patch applied. This obviously does not work since the token exchange between Moodle and the embedded player is not happening. We also identified some problems in the Moodle plugin that should be fixed at this point. Finally, I think there is an issue in octoka when serving files directly on `/`, i.e. without a path prefix. This is rejected. Not a big deal; I suspect the fix is simple and I would guess most use the default `/static` prefix.

That being said, this patch is at all not about CHIPS but rather about attaching JWTs to requests for static content. This is still necessary. octoka does and should not use cookies. For us, cookies are not even an option here with or without CHIPS since the serving domain for static content is not shared with Opencast.

LukasKalbertodt:
The main question in need of answering is: do we need this PR's solution in Opencast's Paella player? Tobira needs it (and has it), octoka will also go nowhere, these things are clear. But is it needed for the built-in Paella player? 

When I say "this solution", I am referring to something that makes the requests generated by Paella for static files contain a JWT. (I looked into it and figured the web worker solution is the best technical implementation of that.)

If we can always have some kind of session cookie that would be used to authenticate all requests Paella performs, then the answer is "no, we don't need this solution", I think. The session can be created via JWT, LTI or some other means, that doesn't matter. And the session cookie can be isolated to an iframe, that would also be totally fine.

However, if
- due to browser cookie limitations/support, this session cookie doesn't always work, or
- because Paella loads data from domains the cookie is not sent to,
... then we do need this solution.

> For us, cookies are not even an option here with or without CHIPS since the serving domain for static content is not shared with Opencast.

This sounds very much like the second point above. So just to be clear: you are saying we need this, i.e. a way that all static file requests generated by Paella in Opencast contain a JWT, correct?

mtneug:
> This sounds very much like the second point above. So just to be clear: you are saying we need this, i.e. a way that all static file requests generated by Paella in Opencast contain a JWT, correct?

Yes.

With the JWT implementation in Moodle (and I think ILIAS as well), the Opencast player is loaded directly as iFrame and the assumption from the LMS is a token exchange with the embedded site, i.e. the same approach as within Tobira.

LukasKalbertodt:
Sorry for asking so much/being pedantic, but I really want to make sure we are all on the same page. Because:

> With the JWT implementation in Moodle (and I think ILIAS as well), the Opencast player is loaded directly as iFrame and the assumption from the LMS is a token exchange with the embedded site, i.e. the same approach as within Tobira.

As far as I can tell, that's _not_ an argument for "this solution". Because if LMS want to pass a JWT to OC, that would also work if the Paella iframe accepts one via URL query parameter, then creates a session (via cookie) and from then on uses that session cookie to perform all requests. 

Again, only if we said (a) session cookie doesn't always work in that context or (b) some requests are for different domains that the cookie is not sent to, ... then we can say for sure we need "this solution".

I hope it's clear what I'm getting at? So was your leading "Yes" independent of the second paragraph?

mtneug:
As straight as I can to your questions:

> I hope it's clear what I'm getting at?

The content of the question yes, the motivation no (see below).

> So was your leading "Yes" independent of the second paragraph?

For *my* services personally, the "Yes" is independent as they fall in the (b) scenario.

---

I don't know what to say. At this point I'm very confused why we argue about the approach again. I thought we already agreed upon the approach here. This draft served as the Opencast side for the implementation of both the ILIAS and Moodle plugins. Now that the first implementations landed in the LMSes, we again question the whole premise?

For me, token-signed URLs were never just about replacing cookies. It's an industry best practice for good reason. My main motivation is to better decouple the video application (i.e. Opencast) from the delivery system (i.e. HTTP web/caching servers / octoka). This separation is intentional and allows multimedia streaming systems to scale.

Currently, static content served by Opencast is secured via user session cookies (default setting). However, it was not always like this and implemented only later because few used the old still existing stream security implementation. BTW this separation [is still described in our docs](https://docs.opencast.org/r/20.x/admin/#configuration/stream-security/stream-security-overview/#content-security-in-opencast).

Using session cookies as the authorization method for the delivery system as well defeats the whole purpose. The delivery system should not care about users as it has no user data.

ok so now for the embedded player served by Opencast (i.e. the front-end bits). This is part of the video application and thus secured by session cookies. Yes, the LMS can use JWTs only to establish a session together with a (partitioned) cookie to load the player. Regardless, the player still must make requests containing (short lived) JWTs to the delivery system to retrieve the static content. I previously raised the question if generating JWTs for this purpose should be done by Opencast itself, thus keeping the LMS out of it. This would allow embedding the player within other contexts, e.g. a CMS. However, the approach you proposed ended up with the embedding host generating the tokens and providing them to the player. This is what @ferishili now implemented. The Opencast side is still missing.

LukasKalbertodt:
Thanks, that helps tremendously! 

Why am I asking these questions again? In my head, seeing CHIPs solve some of these issues did in fact question this whole approach again (especially given its complexity). But it seems I "overreacted" and CHIPs were never in a position to invalidate all of this, since this reaches farther than just "avoid 3rd party cookies". While lack of time is the main reason this PR has been idle for so long, this thought in my head "maybe CHIPs makes all of this useless" also contributed. 

This conclusion that "we definitely still need this" actually helps, as now it's "just" a matter of fixing the remaining issues with this PR, not rethinking everything. Next week we can also chat in person, but I would expect this PR to finally be moving again soonish.

LukasKalbertodt:
**Potential bug in Safari when 302 redirects are involved**

We noticed something weird with this service worker approach while updating Tobira for the ETH and enabling it there (Tobira essentially uses the same technique). I haven't time to fully dive into this, but it's relevant for this PR, so here it goes:

It was noticed that on Safari, Videos (sometimes) loaded very slowly and jumping to the end of the video was also very slow. This already smelled of "`Range`-header issue" to me and looking into it kind of confirmed that. I only tested via browerstack since I don't have access to a mac, but there I saw that when loading the video the following happened:
- A `GET` request to the `dist` subdomain was sent, with range headers
- The server replied with a `302` redirect to `dist01` subdomain
- Then Safari sent a new request to the `dist01` server, but this time _without_ range headers!

This always led to Safari downloading the whole video file. There are still many things unclear about that, mostly due to my unfamiliarity with Safaris dev tools, and the fact that the latter are pretty shitty. There were many other requests with range headers to `dist01` as well. Some of those were marked red, which I don't know what it means. I couldn't really figure out which requests were sent by the JS, and which by the service worker. I also didn't see the CORS preflight requests, which should happen. Also, sometimes the video didn't load at all, sometimes it took some time, sometimes it was fast but jumping to the end was slow. Only one thing I'm certain of: this weird "download whole video" happened every time, as long as I checked `disable cache`. And as soon as we disabled the service worker solution in Tobira again, the problem was gone and range headers in the redirect-follow were there again. And: Firefox and Chrome worked perfectly either way.

Again, lots of questions still. Maybe this is a Safari bug. Maybe it has to do with CORS, maybe we need to adjust the CORS reply and allow more headers. But it seems to be related to the redirect for sure.

I will look into it in the future, but I'm happy about input of course.

mtneug:
Interesting observations. Maybe some things to try:

* Does it make a difference if you use `307` instead of `302`?
* I assume this is pseudo streaming? Does using HLS make a difference (with range headers)? If HLS is not already set up, I guess this is a bit more involved to test. You could package content directly with FFmpeg and put it on the distribution servers instead of changing how Opencast operates.
* IIRC CORS requests are a bit hidden in the Chrome dev tools. Maybe it's similar in Safari and has to be enabled. In any case, the server should see them and thus presumably include them in the HTTP log.

As an aside: HLS (and DASH) have content steering which is intended for CDN switching, but [some are apparently also using it to switch between edge servers](https://www.youtube.com/watch?v=uwZpykPPNSQ). So, no `dist` that redirects but a server that informs HLS clients about preferred pathways.

PR #7260, opened 2025-12-05
Try to read series acl from database

JamesUoM:
To improve the re-indexing of series try to read the series acl from the database instead of from file. The storage solution for the xamcl files can greatly impact the amount of time to re-index the series.

### For Discussion
This assumes that the database is the SPOT for series ACL and that the current ACL should always be applied. This could affect orgs where the series ACL has been changed over time and recordings have been published with different sets of ACLs ( depends on whether auto-publishing of new metadata is in effect? ). Should this feature be optional?

### How to test this patch

* before applying the PR use the helper-scripts to generate at least 1k series
* re-index series - note time spent
* apply patch
* re-index series - note time spent
* with the default setup of H2 and local disk ~10% improvement, this improves more with mariadb and shared/object storage

### DIFF
--- a/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
+++ b/modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
@@ -39,6 +39,8 @@ import org.opencastproject.security.api.AuthorizationService;
 import org.opencastproject.security.api.Role;
 import org.opencastproject.security.api.SecurityService;
 import org.opencastproject.security.api.User;
+import org.opencastproject.series.api.SeriesException;
+import org.opencastproject.series.api.SeriesService;
 import org.opencastproject.util.MimeTypes;
 import org.opencastproject.util.NotFoundException;
 import org.opencastproject.util.data.Tuple;
@@ -90,6 +92,9 @@ public class XACMLAuthorizationService implements AuthorizationService {
 
   /** The serializer for media pacakge */
   private MediaPackageSerializer serializer;
+
+  /** The series service */
+  private SeriesService seriesService;
 
   private static final String CONFIG_MERGE_MODE = "merge.mode";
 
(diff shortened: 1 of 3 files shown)


modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -196,7 +201,19 @@ private Optional<AccessControlList> getAclByFlavor(MediaPackage mp, MediaPackage
       } catch (URISyntaxException e) {
         logger.warn("URI {} syntax error, skip decoding", uri);
       }
-      acl = loadAcl(uri);
+      // UOM: Try to get ACLs from database else fall back to url
```
Arnei, line 204: No need to mention UOM

modules/authorization-xacml/src/test/java/org/opencastproject/authorization/xacml/XACMLSecurityTest.java
```diff
@@ -114,9 +121,10 @@ public void setUp() throws Exception {
             () -> new FileInputStream(uri.getValue().getPath())).anyTimes();
     workspace.delete(EasyMock.anyObject(URI.class));
     EasyMock.expectLastCall().anyTimes();
-    EasyMock.replay(securityService, workspace);
+    EasyMock.replay(securityService, seriesService, workspace);
```
Arnei, line 124: This fixes the tests, but the new code paths are not actually tested I believe?

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -196,7 +201,19 @@ private Optional<AccessControlList> getAclByFlavor(MediaPackage mp, MediaPackage
...
+      if (xacmlPolicyFlavor == XACML_POLICY_SERIES) {
```
Arnei, line 205: `equals` would probably be more consistent?

modules/authorization-xacml/src/main/java/org/opencastproject/authorization/xacml/XACMLAuthorizationService.java
```diff
@@ -196,7 +201,19 @@ private Optional<AccessControlList> getAclByFlavor(MediaPackage mp, MediaPackage
...
+        try {
+          acl = Optional.ofNullable(seriesService.getSeriesAccessControl(mp.getSeries()));
+          if (acl.isEmpty()) {
+            throw new NotFoundException("Can't find series in database");
```
Arnei, line 209: This reads like control-flow-exception-throwing, which is (arguably) an anti-pattern.

Arnei (changes requested):
Can confirm this makes the index rebuild slightly faster (21 seconds instead of 25 seconds for a local AllInOne install with ~10k series).

I am worried about changing this in XACMLAuthService. This would affect *a lot* of services, no just re-indexing. This could cause unintended side-effects for callers that expect the current behaviour. For example, the workflow service at one point gets the ACL both from the SeriesService and the AuthService and compares them. With this change, that comparison becomes a no-op.  https://github.com/opencast/opencast/blob/b694be61a9ab01fa83e51758d1a4c2b9ab5fcfd4/modules/workflow-service-impl/src/main/java/org/opencastproject/workflow/impl/WorkflowServiceImpl.java#L1211

This also causes the widely used XACMLAuthService to now have a dependency on series service. Which works, but the additional coupling this introduces is annoying me :').

So maybe it would be better to change this just for re-indexing, not in general?

PR #7271, opened 2025-12-08
Consistent Workflow Styling

ciegler:
The styling of the yaml workflows is not in itself consistent, which does not look good. To prevent style-chaos I suggest that the community workflows are consistent in their styling. 

I would suggest the following styling (derived from how the majority of the workflows was already styled):
- only use quotes when they are necessary (always before a `*`, since they are not interpreted correctly without quotes)
- use single quotes only
- only use parenthesis when necessary
- white space after logical operators and variables in the operation condition
- empty line after each workflow operation
- for flavors and tags use comma separated lists without white spaces

This PR includes the following changes
- fixing typos
- removing unnecessary quotes
- use single quotes only
- removing unnecessary parenthesis
- using white spaces in the operation conditions
- fixing `OR $ ${variable}` typo in `partial-preview` and `partial-publish`

### DIFF
--- a/etc/workflows/duplicate-event.yaml
+++ b/etc/workflows/duplicate-event.yaml
@@ -9,51 +9,49 @@ state-mappings:
   - state: failing
     value: EVENTS.EVENTS.STATE_MAPPING.DUPLICATING
 configuration_panel_json: |-
-  [
-    {
-      "description": "Title of Duplicate:",
-      "fieldset": [
-        {
-          "name": "mpTitle",
-          "type": "text",
-          "value": " "
-        }
-      ]
-    },
-    {
-      "description": "Series Id:",
-      "fieldset": [
-          {
-          "name": "seriesId",
-          "type": "text",
-          "value": " "
-        }
-      ]
-    },
-    {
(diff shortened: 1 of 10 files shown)


mtneug:
I'm very surprised that `OR $ ${variable}` doesn't show up as an exception failing the whole workflow. This should be a syntax error as I understand the grammar.

gregorydlogan:
It would be nice to have this in `r/19.x` as well since things will be forward ported from there for a full year still.

ciegler:
Ok, thanks for your input. Then I will rebase the PR to `r/19.x`  and I will also open an additional PR to fix the `OR $` syntax error in `r/18.x`.

ciegler:
Alright, I'm sorry, I made a mistake in the first rebase, but now this branch is rebased with `r/19.x` and the PR is also against `r/19.x`.

ciegler:
For the Syntax error, which is also present in `r/18.x` I opened an additional PR: https://github.com/opencast/opencast/pull/7273

gregorydlogan (approved):

Issue #7241, opened 2025-12-02
editor-WOH and vtt: Adjust subtitle/chapter timestamps when they overlap with a removed section


majosch:
**Describe the bug**

Instead of deleting overlapping subtitle/chapter segments, adjust their timestamps when they overlap with a removed section.

**To Reproduce**

1a. Create subtitles in the Editor UI

`captions/source`
```
WEBVTT

00:00.000 --> 00:30.000
S1

00:30.000 --> 01:00.000
S2

01:00.000 --> 01:30.000
S3

01:30.000 --> 02:00.000
S4

02:00.000 --> 02:37.084
S5
```

1b. Create chapters in the Editor UI

`chapters/source`
```
WEBVTT

00:00.000 --> 00:30.000
C1

00:30.000 --> 01:00.000
C2

01:00.000 --> 01:30.000
C3

01:30.000 --> 02:00.000
C4

02:00.000 --> 02:37.082
C5
```
2. Mark some segements for deletion in der Editor UI, that overlap with your subtitles/chapters
 
<img width="1904" height="1007" alt="Image" src="https://github.com/user-attachments/assets/5fe61d8e-9538-48e1-92b2-e117ed2ee6af" />

3. Save and process video
4. Chapters and subtitles, that are overlap with deleted sections are removed entirely

`captions/delivery`
```
WEBVTT

00:00:00.000 --> 00:00:30.000 
S1

00:01:03.096 --> 00:01:33.096 
S4
```

`chapters/delivery`
```
WEBVTT

00:00:00.000 --> 00:00:30.000 
C1

00:01:03.096 --> 00:01:33.096 
C4
```
**Expected behavior**

`chapters/delivery`
```
WEBVTT

00:00:00.000 --> 00:00:30.000 
C1

00:00:30.000 --> 00:00:39.300
C2

00:00:39.300 --> 00:01:03.096
C3

00:01:03.096 --> 00:01:33.096 
C4

00:01:33.096  --> 00:02:00.124 
C5
```

Arnei:
I think we want different behaviour for subtitles and chapters here.

For subtitles I think the current behaviour is arguably reasonable. While we can easily adjust the timestamps for subtitles, we cannot easily adjust their content. This could lead to situations where certain words are removed from the audio, but are still present in the subtitles which is undesirable. Therefore we err on the side of caution and remove such subtitles alltogether.

[...]

PR #7275, opened 2025-12-10
Add new cutting "mode" to editor for vtt files

Arnei:
Addresses #7241.

Currently, the editor removes cues from vtt files when the cue overlaps with a cut segment. For example, if there is a cue that goes from 10 to 16, and there is a cut from 13 to 20, that cue would be removed. This patch adds an alternative "Shorten" mode to the editor. With this active, cues are not removed but their duration is reduced if possible. If the above example was processed in shorten mode, the cue would remain and go from 10 to 13 instead. "Shorten" mode is intended to be used with chapters, which have slightly different semantics from normal subtitles.

### How to test this patch

You could use the new chapter editor in the editor to generate and cut chapters with the new mode. Or you could configure "captions" as a flavor for Shorten mode and test cutting subtitles with it (which you could also provide through the editor).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg
+++ b/etc/org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg
@@ -13,6 +13,14 @@ video.fade = 0.2
 # Default: 2000 milliseconds
 #segments.min.cut.duration = 2000
 
+# The main flavor of the vtt files which should be cut in "Shorten" mode.
+# The default mode will remove cues from vtt files if they overlap too much with a cut segment. This is not desirable
+# for e.g. chapters. "Shorten" mode will instead keep the cue, but reduce its duration by the overlap amount. A cue that
+# overlaps a cut segment completely is still removed.
+# Multiple type can be provided as a comma separated list, e.g. chapters, captions
+# Default: chapters
+#vtt.shorten.flavor.types = chapters
+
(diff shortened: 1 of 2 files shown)


majosch:
Just tested this on a local container environment (with captions and chapters) and works as intended.

etc/org.opencastproject.videoeditor.impl.VideoEditorServiceImpl.cfg
```diff
@@ -13,6 +13,12 @@ video.fade = 0.2
 # Default: 2000 milliseconds
 #segments.min.cut.duration = 2000
 
+# The main flavor of the vtt files which should be cut in "Shorten" mode.
+# The default mode will remove cues from vtt files if they overlap too much with a cut segment. This is not desirable
+# for e.g. chapters. "Shorten" mode will instead keep the cue, but reduce its duration by the overlap amount. A cue that
+# overlaps a cut segment completely is still removed.
+vtt.shorten.flavor.types = chapters
```
gregorydlogan: Reminder from Dec 16: It should be clarified that this can be a list, and what the options (and formatting!) for that list are.

gregorydlogan (changes requested):
I can't say that I *really* understand the difference between a chapter, and a segment are, but the code looks right to me.  Nit flagged from the Dec 16 technical meeting

PR #7310, opened 2026-01-06
Translated the lang-da_DK.json correct.

mjencabh:
Updated the Danish localization.
This commit adds a complete Danish translation


mjencabh:
I have filled out the ICLA online now.

gregorydlogan:
I *think* I've applied this into the codebase.  We will see when the next translation sync happens.

gregorydlogan:
This is done, thanks for the translations!

PR #7311, opened 2026-01-06
Deactivate heartbeat on non-dispatching nodes

### LINKED ISSUE #6706: Deactivate Service Registry Heartbeat for everything but the Admin
We at elan usually deactivate the service registry heartbeat for everything but the Admin node via Ansible. In our opinion, only the Admin (or rather, the node doing job dispatching) needs to be able to talk (and dispatch) to all other nodes in the cluster. 

This is not just for performance reasons as described in https://github.com/opencast/opencast/issues/1538. 
If one spins up a new worker that can't talk to other nodes because of e.g. firewall settings, but _can_ talk to the database, not only does that worker not work (not so bad), but this can pull down the whole cluster because that worker will keep marking other nodes (and thus their services) as offline, seriously impacting job dispatching across the whole system. We've seen that happen quite a few times over the years and before https://github.com/opencast/opencast-admin-interface/pull/1178, this wasn't even visible in the UI.

It would be great if we could set up the service registry so that only the node doing job dispatching does the heartbeat as well.

Arnei:
Fixes #6706.

With this patch, only dispatching nodes (usually admin or allinone) will have a heartbeat. Conversely non-dispatching nodes will not have a heartbeat anymore and cannot be configured to have one either.

Reasons:
- There is no use case for checking for heartbeat on non-dispatching nodes.
- A non-dispatching node could bring down the whole cluster if it could not talk to every other node, by repeatedly marking the unreachable nodes as offline. This is undesirable.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg
+++ b/etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg
@@ -6,3 +6,10 @@
 # usually not be activated on these nodes to avoid concurrency problems.
 # Default: 0
 #dispatch.interval=0
+
+# The interval in seconds between checking if the hosts in the service registry hosts are still alive.
+# Only effective on dispatching nodes, meaning nodes with a dispatch.interval higher than 0. Non-dispatching nodes
+# do not have a heartbeat and changing the interval will have no effect on them.
+# Set to 0 to disable checking if hosts are still alive and able to be dispatched to.
+# Default: 60
+#heartbeat.interval=60
--- a/etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg
+++ b/etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg
@@ -6,10 +6,3 @@
 # usually not be activated on these nodes to avoid concurrency problems.
 # Default: 0
 #dispatch.interval=0
-
-# The interval in seconds between checking if the hosts in the service registry hosts are still alive.
-# Only effective on dispatching nodes, meaning nodes with a dispatch.interval higher than 0. Non-dispatching nodes
-# do not have a heartbeat and changing the interval will have no effect on them.
-# Set to 0 to disable checking if hosts are still alive and able to be dispatched to.
-# Default: 60
-#heartbeat.interval=60
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
@@ -110,10 +110,6 @@ public class JobDispatcher {
   /** Multiplicative factor to transform dispatch interval captured in seconds to milliseconds */
   static final long DISPATCH_INTERVAL_MS_FACTOR = 1000;
 
-  /** Configuration key for the interval to check whether the hosts in the service registry are still alive,
-   * in seconds */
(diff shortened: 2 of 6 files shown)


etc/org.opencastproject.serviceregistry.impl.JobDispatcher.cfg
```diff
@@ -6,3 +6,10 @@
 # usually not be activated on these nodes to avoid concurrency problems.
 # Default: 0
 #dispatch.interval=0
+
+# The interval in seconds between checking if the hosts in the service registry hosts are still alive.
+# Only effective on dispatching nodes, meaning nodes with a dispatch.interval higher than 0. Non-dispatching nodes
+# do not have a heartbeat and changing the interval will have no effect on them.
+# Set to 0 to disable checking if hosts are still alive and able to be dispatched to.
+# Default: 60
+#heartbeat.interval=60
```
gregorydlogan: I would argue that this should just be removed.  Having this exposes the ability to break the system by enabling job dispatch but *disabling* the heartbeat.  Even altering the interval struggles (IMO) to find a usecase - who's actually changing this from 60, or 0?

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -225,6 +232,25 @@ public void updated(Dictionary properties) {
       }
     }
 
+    long heartbeatInterval = DEFAULT_HEART_BEAT;
+    String heartbeatIntervalString = StringUtils.trimToNull((String) properties.get(OPT_HEARTBEATINTERVAL));
+    if (StringUtils.isNotBlank(heartbeatIntervalString)) {
```
gregorydlogan: If we remove the config, we *could* remove a bunch of the logic here.  I'm open to the idea of keeping it around in case someone does need it for something though.

gregorydlogan (commented):
In general I like this change: Non-dispatching nodes doing heartbeat checks do cause problems if inter-node connectivity is disrupted, even if that disruption doesn't *actually* break things.  I do have some concerns with this implementation though:

1) We've made the hearbeat check contingent on the dispatcher being active, without making that clear in the docs (aside from the release notes).  Doubly concerning here is that this is not clear in the config.  If someone turns dispatch off, this also disables the heartbeat!
2) We expose a configuration value that, I suspect, is either left at its default, or turned off.  Considering that it is *required* in dispatching situations, and likely turned off when we're not dispatching, I would argue that we should remove the ability to change the heartbeat interval.

Arnei:
> We've made the hearbeat check contingent on the dispatcher being active, without making that clear in the docs (aside from the release notes). Doubly concerning here is that this is not clear in the config. If someone turns dispatch off, this also disables the heartbeat!

Maybe I've missed it, but we don't mention the heartbeat in the Opencast docs at all, do we? Where would you expect this to be mentioned?

> We expose a configuration value that, I suspect, is either left at its default, or turned off. Considering that it is required in dispatching situations, and likely turned off when we're not dispatching, I would argue that we should remove the ability to change the heartbeat interval.

I was not sure if there was still a use case for changing the heartbeat interval. Glad to hear there is not, will simplify the configuration.

KatrinIhler:
+1 to Greg's suggestion to just rip out the config. But I also don't know which docs you mean @gregorydlogan :D

gregorydlogan:
Exactly. This is fairly undocumented. We either need to add docs, or remove the config options imo :)

Arnei:
And it is removed.

Issue #7049, opened 2025-09-12
Opencast stops database before shutdown


lkiesow:
When shutting down Opencast, it seems that the database connection is shut down before other parts of Opencast are shut down. This could lead to data loss since still changing states cannot be persisted.

To reproduce, just shut down Opencast and take a look at the logs:

```
2025-09-12T11:43:07,505 | INFO  | (Activator:211) - Shutting down database
2025-09-12T11:43:07,505 | INFO  | (Activator:215) - Shutting down connection pool
2025-09-12T11:43:07,535 | WARN  | (WebContainerManager:233) - Can't get a WebContainer service from [org.osgi.service.http.HttpService, org.ops4j.pax.web.service.WebContainer]
2025-09-12T11:43:09,177 | WARN  | (JobDispatcher$JobDispatcherRunner:329) - Error dispatching jobs
java.lang.reflect.UndeclaredThrowableException: null
	at jdk.proxy9.$Proxy63.createEntityManager(Unknown Source) ~[?:?]
	at org.opencastproject.db.DBSessionImpl.execChecked(DBSessionImpl.java:78) ~[?:?]
	at org.opencastproject.db.DBSessionImpl.exec(DBSessionImpl.java:66) ~[?:?]
	at org.opencastproject.serviceregistry.impl.JobDispatcher$JobDispatcherRunner.run(JobDispatcher.java:284) ~[?:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572) ~[?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
	at java.lang.Thread.run(Thread.java:1583) [?:?]
Caused by: java.lang.reflect.InvocationTargetException
	at jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:580) ~[?:?]
[...]

PR #7312, opened 2026-01-06
Add @Deactivate method to JobDispatcher.java

Arnei:
Adds an osgi deactivate method to the JobDispatcher to properly shut down the JobDispatcher Runnable when Opencast shuts down.

This is supposed to help with #7049, which assumes that:

> When shutting down Opencast, it seems that the database connection is shut down before other parts of Opencast are shut down.

This should not be the case. Based on the starting levels in /assemblies/karaf-features/src/main/feature/feature.xml, the db module should not only start before all other Opencast modules, but also be shut down after all other Opencast modules. Therefore, if a module is still trying to use the db after being shut down, it likely has not properly closed its open resources.
Because the stacktrace in #7049 shows the JobDispatcher not being shut down, this patch adds a deactivate method similar to the one in ServiceRegistryJpaImpl.java to ensure that the JobDispatcher thread is properly shut down.

Unfortunately I was not able to reproduce the stacktrace shown in #7049, so I do not actually know if this patch provides an effective fix.

### How to test this patch

Shut down an Opencast with this patch installed and observe if it properly shuts down (ideally while a bunch of jobs are running).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
@@ -57,6 +57,7 @@ import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
 import org.osgi.service.component.annotations.Modified;
 import org.osgi.service.component.annotations.Reference;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -242,6 +243,24 @@ public void updated(Dictionary properties) {
     }
   }
 
+  @Deactivate
+  public void deactivate() {
```
Bennit99: test

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/JobDispatcher.java
```diff
@@ -242,6 +243,24 @@ public void updated(Dictionary properties) {
...
+    logger.info("Deactivate Job Dispatcher");
+
+    // Wait for runnable to stop before stopping
+    if (scheduledExecutor != null) {
+      try {
+        scheduledExecutor.shutdownNow();
+        if (!scheduledExecutor.isShutdown()) {
+          logger.info("Waiting for Job Dispatcher to terminate");
+          scheduledExecutor.awaitTermination(10, TimeUnit.SECONDS);
```
KatrinIhler, line 256: It could be useful to log whether termination was successful or if we just hit the timeout and gave up (look at the boolean that awaitTermination returns).

KatrinIhler:
I suspect that the job failures have nothing to do with your changes, but I can't restart the failed jobs :(

KatrinIhler (changes requested):
Minor thing, see comment. Otherwise ready for merge.

PR #7320, opened 2026-01-14
Add thumbnail support to YouTube publication service

JulianKniephoff:
This PR adds support for uploading a custom thumbnail when publishing to YouTube.

Two new configuration options have been added to the `publish-youtube` workflow operation:
- `thumbnail-flavor`: Specifies the flavor of the attachment to use as a thumbnail.
- `thumbnail-tags`: Specifies the tags of the attachment to use as a thumbnail.

**Note:** To test thumbnail uploads, the YouTube account must be verified with a phone number.

For details on thumbnail restrictions (size, format, etc.), please refer to the updated documentation or the [YouTube Data API documentation](https://developers.google.com/youtube/v3/docs/thumbnails/set).

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
@@ -10,7 +10,15 @@ Description
 The publish YouTube operation publishes a single stream to YouTube. This stream must meet YouTube's format
 requirements, and may consist of one audio and/or video stream. If you want to publish both your presenter and presentation
 streams we suggest using the [Composite](composite-woh.md) workflow operation handler to prepare a composite file
-with both streams inside of it.  
+with both streams inside of it.
+
+### Thumbnails
+
+YouTube has specific restrictions for custom thumbnails:
+- The image must be in a supported format (e.g., JPG, PNG).
+- The file size must be less than 2MB.
+- The YouTube account must be verified with a phone number to upload custom thumbnails.
+- For more details, see the [YouTube API documentation](https://developers.google.com/youtube/v3/docs/thumbnails/set).
 
 
 Parameter Table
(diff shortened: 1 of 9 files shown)


docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
```diff
@@ -20,6 +28,8 @@ Parameter Table
 |---------------------------|------------------------------------------------------------------------------|
 |source-flavors             |The flavors to publish to YouTube                                             |
 |source-tags                |The tags to publish to YouTube                                                |
+|thumbnail-flavor           |The flavor of the thumbnail to publish to YouTube                             |
```
ypatios: i understand that this would be opening a entirely new (or rather old) can of worms, but i must note the creative use of the flavor/tag config here :D  
if we accept this concept, logically the captions would need their own key -> captions-flavor etc. — which is an understandable convention.

ypatios (approved):
looks good :) didn't realise there were that many files to be changed.
I left a comment regarding the wf-operation config, but it's just a note rather than a request for change. thx for the thorough work!

docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
JulianKniephoff: That's exactly how I'm planning to implement it for captions as well, yes. I'm not a big fan of it, either, but I wasn't aware that there was prior discussion/baggage about the use of these kinds of options? 🤔 Of course we could just have `source-{tags,flavors}` and then have some **other** heuristic to pick out the video track vs. thumbnail vs. caption tracks from the selected set of MP elements, but I'm even less of a fan of that, still.

Roping in @gregorydlogan here for the historical and QA perspective. 😅

gregorydlogan: Give that this is a *terminal* WOH, in the sense that no one else is going to be re-using those files, I don't have an issue with it.  The baggage comes more from making sure that there's a sane flow of flavours and tags from the beginning to the end of the stream of WOHs.  If inspect output one flavour, compose used a different one, and then something else used yet another it would get confusing, but the YouTube distribution stuff is probably last, and probably not producing outputs that other things might want.

JulianKniephoff: Then I'll leave it like this for this PR. Still, I'd be interested in what people think a more idiomatic solution could look like. 🤔

Issue #7323, opened 2026-01-20
Admin UI Search: Certain use cases not working anymore


Arnei:
https://github.com/opencast/opencast/pull/6498 has considerably improved text search in the admin ui. However, it has also negatively impacted some use cases. This issue is supposed to document the use cases and allow for more public discussion on how to address them or if to address them at all, as I don't feel at all confident in just deciding the best way forward by myself.

I placed this issue in the backend because even though the use cases are in the frontend, they were affected by changes in the backend and will likely have to be fixed in the backend as well.

### Use Case 1: Partial String Search
"We have a series title of the format 'LMS:12345:SomeText' and need to be able to find the series by '12345'. This is not possible anymore."

This is likely due to our use of edge-ngram tokenization, which prevents us from finding partial strings. Our tokenization can prevent a lot of false positives (I am not really looking for 'Administrator' when searching for 'ist'). But maybe searching for partial strings is important enough to our users that we should go back to a more lax tokenization that allows more false positives?

### Use Case 2: Long String Search in Extended Metadata
"We have a URL in the extended metadata of our events, and we want to be able to find an event by that URL. This is not possible anymore."

This is a problem specific to extended metadata. Unlike our common metadata, they do not have elasticsearch mappings for each field, so there is no keyword to find them by. Instead they are all just crammed into the text field in elasticsearch. This means strings get cut off by our edge-ngram tokenization (which is currently configured to 20 characters max). 

One way to go about this might be to have a different field for extended metadata, i.e. instead of adding them to `text` they are added to a new field `text_extended`. That field would then also need to be search, but could be configured to be of `type: keyword` to allow for finding very long strings.

### Use Case 3: Too many results for ID search if that ID is in an unexpected field
"We have our capture agent ids in our `location` metadata field. The ids are very similar, for example 'text123', 'text124', 'text223'. When searching for events of a specific capture agent, events from other capture agents also show up. That did not happen before."

This is likely because we now actually use fuzzy search (which allows finding similar strings and covering typos, e.g. finding 'sick' when searching for 'sic').

We could disable fuzzy search again. But we could also say that our search is working as intended, and users should use the provided capture agent filter in the admin ui if they want all events that belong to one specific capture agent.

KatrinIhler:
@Arnei I talked with Lars about this in the OSDN meeting today and he said that in case you run into limitations regarding client hours while working on this, you can use up to 30h of OSDN. :)

LukasKalbertodt:
[...]

PR #7324, opened 2026-01-20
Use ngram tokens in index

Arnei:
Partially addresses #7323.

This patch attempts to address use cases 1 and 2 from the linked issue. It should allow users to search for partial strings again, without completely disregarding the benefits of the new behaviour.

This is a draft because I am by no means an expert on proper elasticsearch configuration and not sure this is a good approach actually. So I am looking for feedback! For example, I am not too happy with using ngrams analyzers as that will likely result in a lot of false positives in certain cases.

I am aware that this could not go into r/18.x, I just selected it because it is the base branch for this one. As this patch would require an index migration, it must go into develop and index migration scripts would need to go into the appropriate folders.

### How to test this patch

Test with an OC 18 installation. Use the migration script to update your index (beware that the migration script has no "undo" functionality!).

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/docs/upgrade/17_to_18/event_mapping.json
+++ b/docs/upgrade/17_to_18/event_mapping.json
@@ -235,7 +235,13 @@
     "text": {
       "type" : "search_as_you_type",
       "analyzer": "autocomplete_analyzer",
-      "search_analyzer": "whitespace_analyzer"
+      "search_analyzer": "whitespace_analyzer",
+      "copy_to": "text_technical"
+    },
+    "text_technical": {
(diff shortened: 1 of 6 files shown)


geichelberger:
There is an alternative to achieve good search for text content, and even support finding more technical content. One part of that would be to use wildcards like `*.search, *.technical` for defining the fields you want to search in. For those subfields, we can define the best way to index the data and boost the results of full-text search (search-as-you-type field).

Arnei:
This is outdated

PR #7328, opened 2026-01-21
Read series ACLs from db to improve performance

JamesUoM:
Series ACLs are read from file, this is expensive if they are not stored on local disk. Improve performance reading them from the database. Reading from ACLs disk can affect the time it takes to index the  AssetManager.

**Does this work for everyone? Is anyone requiring events in the same series to have different series ACLs?**

### How to test this patch

Create a large number of example series and rebuild the AssetManager Index.

### DIFF
--- a/modules/authorization-xacml/pom.xml
+++ b/modules/authorization-xacml/pom.xml
@@ -28,6 +28,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.opencastproject</groupId>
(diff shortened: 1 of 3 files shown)


JamesUoM:
Duplicate

mtneug:
@JamesUoM I can't find the duplicate. Is this closed correctly?

mtneug:
Never mind. I haven't searched good enough. Duplicate of #7260.

PR #7332, opened 2026-01-22
Get Extended Metadata of Events from API via the Index

### LINKED ISSUE #7473: Don't read Extended Metadata from Disk for External API requests
See #7332. Issue created for representation on OSDN board.

KatrinIhler:
Instead of reading extended metadata catalogs from disk, we can use the fact that extended metadata are in the index (#3274) to get them from there when requesting metadata from the External API. This should be significantly faster.

The returned format should be the same, which made the code a bit uglier, but means no new API version. To test, configure extended metadata and request an event from the API with `withmetadata` set.

Indexing extended metadata was first introduced in OC 13, which required an index rebuild anyway, but since there were some issues with that in the recent past (see #6984), this might require another index rebuild for the Asset Manager depending on when the last index rebuild was run.

One small caveat: The extended metadata in the index currently don't consider that one could configure two metadata fields with the same id for the same catalog, but for different namespaces. As this should be very rare, I'm not touching that right now bc we want to move away from dublincore anyway.

Fixes #7473

### DIFF
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -1637,7 +1637,7 @@ public class EventsEndpoint implements ManagedService {
 
     if (!extendedCatalogUIAdapters.isEmpty()) {
       Map<String, Map<String, List<String>>> extendedMetadata = event.getExtendedMetadata();
-      for (EventCatalogUIAdapter catalogUIAdapter : catalogUIAdapters) {
+      for (EventCatalogUIAdapter catalogUIAdapter : extendedCatalogUIAdapters) {
         if (extendedMetadata.containsKey(catalogUIAdapter.getFlavor().toString())) {
           Map<String, List<String>> extendedMetadataByType = extendedMetadata.get(
               catalogUIAdapter.getFlavor().toString());
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -1654,7 +1654,7 @@ public class EventsEndpoint implements ManagedService {
               metadataField.setValue(null);
             }
           }
-          metadataList.add(catalogUIAdapter.getFlavor().toString(), catalogUIAdapter.getUITitle(), dublinCoreMetadata);
+          metadataList.add(catalogUIAdapter, dublinCoreMetadata);
         }
       }
     }
--- a/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
+++ b/modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
@@ -53,6 +53,7 @@ import org.opencastproject.external.util.AclUtils;
 import org.opencastproject.external.util.ExternalMetadataUtils;
 import org.opencastproject.index.service.api.IndexService;
 import org.opencastproject.index.service.catalog.adapter.DublinCoreMetadataUtil;
+import org.opencastproject.index.service.catalog.adapter.events.CommonEventCatalogUIAdapter;
 import org.opencastproject.index.service.exception.IndexServiceException;
 import org.opencastproject.index.service.impl.util.EventHttpServletRequest;
 import org.opencastproject.index.service.impl.util.EventUtils;
@@ -153,6 +154,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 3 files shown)


modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1275,6 +1286,18 @@ protected JsonObject eventToJSON(Event event, Boolean withAcl, Boolean withMetad
...
+      Map<String, Map<String, List<String>>> extendedMetadata = event.getExtendedMetadata();
+      for (String type: extendedMetadata.keySet()) {
+        Map<String, List<String>> extendedMetadataByType = extendedMetadata.get(type);
+        for (String name: extendedMetadataByType.keySet()) {
+          List<String> values = extendedMetadataByType.get(name);
+          String fieldName = SeriesIndexSchema.EXTENDED_METADATA_PREFIX.concat(type + "_" + name);
+          json.addProperty(fieldName, safeString(values));
+        }
+      }
+    }
```
mtneug: I'm very much against including this in the main event JSON object especially if this is not nested and all extended metadata fields are just at the root level blowing up the object.

Instead, use `withMetadata` and improve the `getEventMetadata` helper method to fetch extended metadata from the index. And yes, the returned JSON structure is harder to navigate, but this

1. would be aligned to the current API design,
2. would not require an API version increase (and could therefore be included in OC 19) and
3. would also improve other endpoints such as `/api/events/{event_id}/metadata`.

mtneug (changes requested):

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
KatrinIhler: I see were you're coming from, but

- We are only bloating up the object if extended metadata is explicitly requested, and also most people don't have that many extended metadata fields

- get-metadata currently doesn't even read the common metadata from the index, but from disk, and changing that is out of scope for this PR. So I didn't want to change that for extended metadata only bc that can be confusing (and also get-metadata would still not perform well, which is the whole reason I wrote this in the first place. :sweat_smile:)
I'm also not sure how much we can rely on extended metadata actually being in the index since index rebuild was documented as optional for that feature, so I didn't want to completely remove the "old" way of getting them. But I agree that the design now is messy.

I guess changing the root level thing is not hard and I could do. But not sure how to fix the other thing.

mtneug: > We are only bloating up the object if extended metadata is explicitly requested, and also most people don't have that many extended metadata fields

Why these assumptions? One special tenant at educast has two catalogs with 112 and 3 fields, respectively. We were mapping a JSON structure that followed a specific metadata standard to the flat Opencast catalog structure so some of the fields had indices. But still.

> get-metadata currently doesn't even read the common metadata from the index, but from disk, and changing that is out of scope for this PR. So I didn't want to change that for extended metadata only bc that can be confusing (and also get-metadata would still not perform well, which is the whole reason I wrote this in the first place. 😅)

IIRC, it should use the data from the index for common but not for extended metadata. I might be wrong; it has been a while since I looked at this code.

Still, this is no argument against improving this method and therefore benefitting the existing endpoints.

> I'm also not sure how much we can rely on extended metadata actually being in the index since index rebuild was documented as optional for that feature, so I didn't want to completely remove the "old" way of getting them. But I agree that the design now is messy.

ok then include the change in develop only and tell people to do the reindex.

> I guess changing the root level thing is not hard and I could do. But not sure how to fix the other thing.

It's not only the root level thing. What you are proposing is:

```
GET https://develop.opencast.org/api/events/ID-espresso-video?withmetadata=true&includeExtendedMetadata=true
```

should result in the following:

```json
{
  "identifier": "ID-espresso-video",
  "title": "Espresso",
  
  // other well known keys for common metadata etc.

  // other arbitrary keys for extended metadata

  "metadata": [
    {
      "flavor": "dublincore/episode",
      "title": "EVENTS.EVENTS.DETAILS.CATALOG.EPISODE",
      "fields": [
        {
          "id": "title",
          "label": "EVENTS.EVENTS.DETAILS.METADATA.TITLE",
          "value": "Espresso",
          "type": "text",
          "readOnly": false,
          "required": true
        },
        // other field objects with known structure for common metadata
      ]
    },
    {
      "flavor": "dublincore/extended",
      "title": "my-extended",
      "fields": [
        // field objects with known structure for extended metadata
      ]
    }
  ]
}
```

Even if you nest the `other arbitrary keys for extended metadata` into another object, this still duplicates the returned data with use of the `withmetadata` flag. Then there is the problem that you don't know the key names. This is the main reason why I dislike mixing known keys and arbitrary keys at the same level. This makes parsing the JSON structure harder and could lead to clashes if we later add new known keys.

We already have a flag to include metadata fields. We should use that instead of adding another flag that somehow does the same faster and in another structure.

KatrinIhler: > IIRC, it should use the data from the index for common but not for extended metadata.

Goddammit, you're right. I was so certain we read those from disk as well. Oh well, back to the drawing board...

KatrinIhler: Rewritten as per your initial suggestion :D

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1628,20 +1630,37 @@ protected Optional<MetadataList> getEventMetadataById(String id) throws IndexSer
...
+            if (extendedMetadataByType.containsKey(metadataField.getInputID())) {
+              List<String> values = extendedMetadataByType.get(metadataField.getInputID());
+              // values are from index, but same format as in catalog
+              DublinCoreMetadataCollection.setValueFromDCCatalog(values, metadataField);
+            } else {
+              // overwrite any default values contained in raw fields to accurately reflect event state
+              metadataField.setValue(null);
+            }
+          }
+          metadataList.add(catalogUIAdapter.getFlavor().toString(), catalogUIAdapter.getUITitle(), dublinCoreMetadata);
```
Arnei, line 1657:
```suggestion
          metadataList.add(catalogUIAdapter, dublinCoreMetadata);
```

Shorter and should work just the same?

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
```diff
@@ -1628,20 +1630,37 @@ protected Optional<MetadataList> getEventMetadataById(String id) throws IndexSer
...
-    if (catalogUIAdapters.size() > 0) {
-      MediaPackage mediaPackage = indexService.getEventMediapackage(event);
+
+    // Get configured extended metadata fields & fill with values from index
+    Set<EventCatalogUIAdapter> extendedCatalogUIAdapters = getEventCatalogUIAdapters().stream()
+        .filter(c -> !(c instanceof CommonEventCatalogUIAdapter)).collect(Collectors.toSet());
+
+    if (!extendedCatalogUIAdapters.isEmpty()) {
+      Map<String, Map<String, List<String>>> extendedMetadata = event.getExtendedMetadata();
       for (EventCatalogUIAdapter catalogUIAdapter : catalogUIAdapters) {
```
Arnei, line 1640: Why do we have to iterate over all catalogUIAdapters here? We are only interested in the catalogUIAdapter for extended metadata, right? We got those earlier with extendedCatalogUIAdapters, so we could just iterate over those?

Arnei (changes requested):
Works and results look reasonable (in line with what Matthias suggested). Can't make any comments how much faster this might be.

Nitpicks below:

modules/external-api/src/main/java/org/opencastproject/external/endpoint/EventsEndpoint.java
KatrinIhler: True, that was an oversight. Thanks, fixed!

Arnei (approved):

PR #7334, opened 2026-01-26
Refactor docker compose service configuration

JulianKniephoff:
- Renamed elasticsearch service to opensearch in all docker compose files
- Removed container_name from all services

This allows container and volume names to be overridden using the -p option of docker compose for better flexibility. For example, this enables running multiple instances of the same services simultaneously for different projects or testing scenarios.


JulianKniephoff:
I have a slight déjà vu suggesting this, but I couldn't find any previous discussion, so I'll just try (again)! 😅

mtneug (approved):

Issue #7335, opened 2026-01-26
The IngestService throws a SAXParseException when HTTPS is enabled in Opencast and the series.xml file is present


grexca:
**Describe the bug**

I'm currently using Opencast 17.5 with **HTTPS enabled**, and it's working fine.

I've tried updating to versions 17.7 and 17.10, but the ingest call fails because these opencast releases throw an exception related to the XML parser. The same issue also occurs with a clean installation of Opencast 19.0.

If I **disable HTTPS**, all the tested releases work correctly with the /ingest endpoint using the same mediapackage.

With https enabled,  calling /ingest/addZippedMediaPackage/{workflowDefinitionId} fails.
Here are the log lines:

```
WARN  | (IngestRestService:1344) - Unable to ingest mediapackage
    org.opencastproject.ingest.api.IngestException: java.lang.RuntimeException: Unable to read DublinCore catalog, XML parsing failed.
    org.xml.sax.SAXParseException: DOCTYPE is disallowed when the feature "http://apache.org/xml/features/disallow-doctype-decl" set to true.
```

This issue only occurs when HTTPS is enabled. If I reconfigure opencast to disable HTTPS, everything works as expected.

It makes no difference whether I use a PyCa capture agent or a Galicaster capture agent.

NOTE: If I remove the "series.xml" file and update the the manifest, the /addZippedMediaPackage/{workflowDefinitionId} call works well even with HTTPS enabled

**To Reproduce**

1. Edit the  'org.ops4j.pax.web.cfg' file and configure opencast to use https
2. Try to ingest a mediapackage from a PyCA or Galicaster Capture Agent using the API Call /ingest/addZippedMediaPackage/{workflowDefinitionId}
3. Look for the "SAXParseException : DOCTYPE is disallowed when ..." exception in the log file.

**Expected behavior**
Same as when https is disabled: No errors when parsing the XML.

**Server environment:**
 - Opencast docker.
 - OS: Ubuntu 24.04.3 LTS
 - Java: openjdk 21.0.9 2025-10-21 LTS
 - Distribution: [admin+presentation+worker]
- Installation method: docker
 - Version [19.0, 17.10, 17.7]

**Additional context**

gregorydlogan:
[...]

PR #7338, opened 2026-01-28
Remove internal SSL documentation

gregorydlogan:
This PR removes the documentation around using Opencast to terminate the SSL connections directly.  This does *not* change how reverse proxies work.  Chances are if you really need to enable ssl in java directly then you can figure this out on your own, and removing these docs prevents people who do not actually want to do this from accidentally messing up their configurations.

Putting this into legacy because legacy will be around for a long time still.  Happy to move it up to stable if people would rather.  Does touch the main config file but only commented out text.

Relates to #7335, though does not close it.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests
* [ ] ~include a [release note](https://docs.opencast.org/develop/developer/#participate/development-process/#release-notes) if a feature~


etc/org.ops4j.pax.web.cfg
```diff
@@ -20,6 +20,8 @@ org.osgi.service.http.enabled=true
 
 # Whether Opencast itself should handle HTTPS traffic.
 # Even if you set this to 'false',you can still use an HTTP proxy to handle SSL.
+# Note: This is NOT required to use https with Opencast.
+# Unless you know exactly what you are doing, use a reverse proxy instead.
 org.osgi.service.http.secure.enabled=false
```
mtneug: What do you think of just removing this option completely from the config file? We already don't include all options for `org.ops4j.pax.web`. People can always add it back, but that's of course harder as they really need to research what options are available.

gregorydlogan: +1

Removed this, plus some other SSL related things that only make sense if this part is there anyway.

mtneug (approved):

PR #7371, opened 2026-02-06
Spring Security 5 Upgrade

rute-santos:
### Upgrade to Spring Security 5

**Notes**

This has been tested with Harvard-DCE's code base, which is somewhere between Opencast 15 and 16, but we only tested the authentication methods that we use:

- User/password
- Digest
- CAS
- JWT
- LTI 1.1

I also tested the basics with the community latest develop.

Not tested: LDAP, Shibboleth, anything else?

LTI: Added a custom filter to process LTI 1.1 requests since spring-security-oauth library has been deprecated. The Lti11AuthenticationFilter.java verifies the signature for LTI 1.1 POST requests. The other related classes were kept as is, but do not implement the old interfaces anymore.

In modules/kernel/src/main/java/org/opencastproject/kernel/security/CustomPasswordEncoder.java, the support for very old md5 based passwords was removed. Anyone still using them?


assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -613,15 +607,26 @@
...
   </feature>
 
+  <!-- This is included in the etc/org.apache.karaf.features.cfg generated from
+       assemblies/dist-xxx/pom.xml -->
   <feature name="opencast-security-cas" version="${project.version}">
-    <bundle start-level="82">mvn:commons-collections/commons-collections/3.2.1</bundle>
-    <bundle start-level="82">mvn:org.apache.velocity/velocity/1.7</bundle>
-    <bundle start-level="82">mvn:org.apache.santuario/xmlsec/${xmlsec.version}</bundle>
-    <bundle start-level="82">mvn:org.bouncycastle/bcprov-jdk15on/1.51</bundle>
+    <!-- Included the bundles below to fix errors like:
```
JamesUoM: This comment can be removed or moved to a Github comment.
Any missing dependency causes similar errors

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -613,15 +607,26 @@
...
+    <!-- Picked the same version used by karaf in the 'shell' feature -->
+    <bundle start-level="82">mvn:org.bouncycastle/bcutil-jdk18on/1.78.1</bundle>
+    <bundle start-level="82">mvn:org.bouncycastle/bcprov-jdk18on/1.78.1</bundle>
+    <bundle start-level="82">mvn:org.bouncycastle/bcpkix-jdk18on/1.78.1</bundle>
     <bundle start-level="82">mvn:org.opencastproject/opencast-security-cas-client-wrapper/${project.version}</bundle>
-    <bundle start-level="82">mvn:org.springframework.security/spring-security-core/3.1.7.RELEASE</bundle>
-    <bundle start-level="82">mvn:org.springframework.security/spring-security-web/3.1.7.RELEASE</bundle>
-    <bundle start-level="82">mvn:org.springframework.security/spring-security-cas/3.1.7.RELEASE</bundle>
+    <bundle start-level="82">mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.spring-security-cas/5.7.12_4</bundle>
+    <!-- Added due to karaf error: [caused by: Unable to resolve org.apache.servicemix.bundles.spring-security-cas/5.7.12.4:
```
JamesUoM: Same as above. This comment can be removed or moved to a Github comment.
Any missing dependency causes similar errors

modules/security-cas-client-wrapper/pom.xml
```diff
@@ -16,24 +16,35 @@
   </properties>
   <dependencies>
     <dependency>
-      <groupId>org.apache.velocity</groupId>
-      <artifactId>velocity</artifactId>
+      <groupId>org.jasig.cas.client</groupId>
+      <artifactId>cas-client-core</artifactId>
     </dependency>
+    <!-- These are dependencies of cas-client-core gotten with 'mvn depdency:tree' -->
```
JamesUoM: typo ;)

etc/security/mh_default_org.xml
```diff
@@ -903,39 +919,6 @@
...
-  <osgi:reference id="userReferenceProvider" cardinality="1..1"
-                  interface="org.opencastproject.userdirectory.api.UserReferenceProvider"  />
-
-  <osgi:reference id="userDirectoryService" cardinality="1..1"
-                  interface="org.opencastproject.security.api.UserDirectoryService" />
-
-  <!-- Uncomment this as an alternative to the userDetailsMapper -->
-  <!-- Make sure you provide the same instanceId you used in org.opencastproject.userdirectory.ldap-….cfg -->
-  <!--
-  <osgi:reference id="authoritiesPopulator" cardinality="1..1"
```
JamesUoM, line 928: I've successfully configured Spring to use our ldap service and I can "login" with my credentials. However, there is no attribute or role mapping; the code removed here references the service provided by:
`modules/userdirectory-ldap/src/main/java/org/opencastproject/userdirectory/ldap/OpencastLdapAuthoritiesPopulator.java` 
with a filter that correlates to the id set is the `etc/org.opencastproject.userdirectory.ldap-<id>.cfg`, the docs have you create from the template, the id is explicitly set in the cfg. This allows multple ldap services to be used.

There must be some method of making Spring aware of this `LdapAuthoritiesPopulator` service; define a bean rather than an OSGI service? The other service reference in this section probably need a similar solution.

The ldap docs need updating as they refer to the `<osgi:reference>` element

JamesUoM (changes requested):

etc/security/mh_default_org.xml
rute-santos: Hi @JamesUoM ,
Thanks for testing! Yes, unfortunately it seems there are pieces missing that need to be fixed for ldap to work...
I had to refresh my knowledge of the ldap configuration - we used it at Harvard at some point a long time ago... :)
Given that:
- Each _OpencastLdapAuthoritiesPopulator_ and _LdapUserProviderInstance_ pair is created dynamically by the _LdapUserProviderFactory_ factory service based on the defined _etc/org.opencastproject.userdirectory.ldap-ID.cfg_
- The defined IDs can be anything and are not known in advance
- There can any number of IDs

It isn't possible to just add another reference in the _SpringSecurityConfigurationArtifactInstaller_ as we did with the other services that are known in advance.
I believe we will have to:
- Know which IDs are expected for each organization (maybe a configuration file for _SpringSecurityConfigurationArtifactInstaller_ where we have something like "ORG_ID=ID_1,ID_2"?)
- Delay the xml parsing and instantiation of the spring context until all _LdapUserProviderInstance_ + _OpencastLdapAuthoritiesPopulator_ pairs have registered and thus can be added with _orgAppContext.getBeanFactory().registerSingleton("someprefix_ID", authoritiesPopulatorInstance)_
- I'm not sure what to do if/when they unregister though...

Does this sound reasonable or can you think of a simpler way? I can make the changes, but I doubt I will be able to test them properly...

JamesUoM: Naive question: so we can't configure an ldap reference with `cardinality="0..N"`, and handle them dynamically?

modules/kernel/src/main/java/org/opencastproject/kernel/security/SpringSecurityConfigurationArtifactInstaller.java
```diff
@@ -94,32 +151,57 @@ public boolean canHandle(File artifact) {
...
+    orgAppContext.getBeanFactory().registerSingleton("securityService", this.securityService);
+    orgAppContext.getBeanFactory().registerSingleton("userDirectoryService", this.userDirectory);
+    orgAppContext.getBeanFactory().registerSingleton("userDetailsService", this.userDetailsService);
+    orgAppContext.getBeanFactory().registerSingleton("userReferenceProvider", this.userReferenceProvider);
+    orgAppContext.getBeanFactory().registerSingleton("oAuthConsumerDetailsService", this.oAuthConsumerDetailsService);
+    orgAppContext.getBeanFactory().registerSingleton("ltiLaunchAuthenticationHandler",
+            this.ltiLaunchAuthenticationHandler);
+
+    XmlBeanDefinitionReader xmlBeanDefinitionReader = new XmlBeanDefinitionReader(orgAppContext);
+    // So that it finds META-INF/spring.handlers
```
JamesUoM, line 288: Where is `spring.handlers` ?

etc/security/mh_default_org.xml
JamesUoM: > Know which IDs are expected for each organization (maybe a configuration file for SpringSecurityConfigurationArtifactInstaller where we have something like "ORG_ID=ID_1,ID_2"?)

Could we record them where they are now in the `<organization>.xml` as properties of a custom element or bean and then SpringSecurityConfigurationArtifactInstaller could parse the xml and wait for the services to be registered?

rute-santos: > Naive question: so we can't configure an ldap reference with `cardinality="0..N"`, and handle them dynamically?

This is more or less what we need to do, but making sure all the optional references are available before we parse the xml to create the security context.

rute-santos: > > Know which IDs are expected for each organization (maybe a configuration file for SpringSecurityConfigurationArtifactInstaller where we have something like "ORG_ID=ID_1,ID_2"?)
> 
> Could we record them where they are now in the `<organization>.xml` as properties of a custom element or bean and then SpringSecurityConfigurationArtifactInstaller could parse the xml and wait for the services to be registered?

Interesting idea. I'm not sure because the parsing is done by _XmlBeanDefinitionReader_, which will throw an exception if the bean is not found.

modules/kernel/src/main/java/org/opencastproject/kernel/security/SpringSecurityConfigurationArtifactInstaller.java
rute-santos: This is inside the spring jar (spring-beans I believe).

etc/security/mh_default_org.xml
rute-santos: These are the changes I had in mind: https://github.com/harvard-dce/opencast-contrib/commit/617ec148dde7d61d5c8038c403dd9135e4ef921f
I have not even tested them so they might not work at all...

JamesUoM: I've had a look at this commit, it just needed this addition to  `LdapUserProviderFactory.java` as the cfg defaults to the first org if not set:

ln 379  `        organization = org.getId();`

Logs show it should be installing the ldapPopulator as a bean:

```11:30:49,917 | INFO  | (SpringSecurityConfigurationArtifactInstaller:153) - Set LtiLaunchAuthenticationHandler
11:30:49,917 | INFO  | (SpringSecurityConfigurationArtifactInstaller:147) - Set setOAuthConsumerDetailsService
11:30:49,917 | INFO  | (SpringSecurityConfigurationArtifactInstaller:117) - Set SecurityFilter
11:30:49,917 | INFO  | (SpringSecurityConfigurationArtifactInstaller:123) - Set SecurityService
11:30:49,918 | INFO  | (SpringSecurityConfigurationArtifactInstaller:135) - Set UserDetailsService
11:30:49,918 | INFO  | (SpringSecurityConfigurationArtifactInstaller:129) - Set UserDirectoryService
11:30:49,918 | INFO  | (SpringSecurityConfigurationArtifactInstaller:141) - Set UserReferenceProvider
11:30:49,920 | INFO  | (SpringSecurityConfigurationArtifactInstaller:203) - mh_default_org = [uom]
11:30:49,921 | TRACE | (SpringSecurityConfigurationArtifactInstaller:223) - Install /opt/opencast/develop/build/opencast-dist-develop-20-SNAPSHOT/etc/security/mh_default_org.xml
11:30:49,922 | INFO  | (SpringSecurityConfigurationArtifactInstaller:256) - Creating security context for organization mh_default_org
11:30:49,952 | TRACE | (SpringSecurityConfigurationArtifactInstaller:281) - Registering bean ldapAuthoritiesPopulator_uom
11:30:50,655 | INFO  | (SpringSecurityConfigurationArtifactInstaller:294) - registered 48 items in [...]
```

It seems to work! I had done my own hacky attempt yesterday that didn't work, but I just spotted a typo in my .cfg which set default roles: ROLE_ADMIN, ROLE_UOM. ROLE_STAFF is mapped from an LDAP attribute, and so is my screen name and email.

image

rute-santos: This is great! I'm going to do a quick test and update the documentation and then will add the commit to this PR.

JamesUoM: In the spirit of exploration, I did some testing to see if we could define the ldap provider ids in the xml. As we already have a customised `UserDetailsContextMapper` bean, I added a List property to this:

```
  <bean id="userDetailsMapper" class="org.opencastproject.security.ldap.OpencastUserDetailsMapper">
    <constructor-arg ref="userDetailsService" />
    <property name="ldapProviderIds">
      <list>
        <value>uom</value>
        <value>other</value>
      </list>
    </property>
  </bean>
```
Then since the kernel is unaware of our custom OpencastUserDetailsMapper I changed it to extend HashSet; the `setLdapProviderIds`, copies the list to `(Set)this`. Once the xml has been parsed and  `orgAppContext.refresh()` called, I can get the ldap ids using:

```
Set<String> ldapProviderIds = (Set<String>)orgAppContext.getBean("userDetailsMapper");
```
After which we could wait for the ldapAuthoritiesPopulator services to appear - not tested.

However, it feels a bit too hacky to just encapsulate the ldap ids in the organization xml and we should proceed with your solution.

rute-santos: Hi @JamesUoM ,
It like the idea of having everything in one place, but I think this is kind of a chicken and egg problem because parsing the security xml requires the osgi service references to be already in place. We would have to parse it twice or something, which would probably be messy.
I added the changes to this PR, fixed the org name bug you pointed out, and updated the documentation. I could not test it completely... due to my inability of creating a working ldap container with users and groups and such, but I believe the basics are working. I hope other people also chime in to test!

JamesUoM: I completely agree, and though it's another config file, it is the least worst solution.

JamesUoM:
Tested with LDAP, simple attribute to role mapping working.

etc/security/mh_default_org.xml
```diff
@@ -899,43 +915,12 @@
...
-
-  <!-- Obtain services from the OSGI service registry -->
-  <osgi:reference id="userDetailsService" cardinality="1..1"
-                  interface="org.springframework.security.core.userdetails.UserDetailsService" />
-
-  <osgi:reference id="securityService" cardinality="1..1"
-                  interface="org.opencastproject.security.api.SecurityService" />
-
-  <!-- For enabling external users e.g. used together with shibboleth and JWT -->
-  <osgi:reference id="userReferenceProvider" cardinality="1..1"
```
JamesUoM, line 919: This is refereed to in the AAI (shib) docs and should be removed from there, `docs/guides/./admin/docs/configuration/security.aai.md`

rute-santos: Good catch!

etc/security/mh_default_org.xml
```diff
@@ -720,44 +731,44 @@
   <bean id="attributeMapper" class="org.opencastproject.security.aai.api.AttributeMapper">
     <property name="useHeader" value="true" />
     <property name="multiValueDelimiter" value=";" />
-    <property name="attributeMap" ref="attributeMap" />
-    <property name="aaiAttributes" ref="aaiAttributes" />
+    <property name="attributeMap">
+      <map map-class="java.util.HashMap">
```
JamesUoM: `map-class` attribute is not available to `map` element

rute-santos: Removed!

snoesberger:
Thanks for that PR!
We installed this PR in our test environment, which is configured for Shibboleth authentication. I was able to log in successfully to Opencast, and the attribute to role mapping worked too.

lgehrlein:
Hey, I can confirm that this branch also works with our shibboleth configuration using the dynamic login handler.

But while continuing my testing with moodle plugins and especially the filter https://github.com/Opencast-Moodle/moodle-filter_opencast I came across the following error when trying to access a video or the studio using LTI:

```
HTTP ERROR 500 java.lang.NoClassDefFoundError: net/oauth/OAuthValidator
URI:	/lti
STATUS:	500
MESSAGE:	java.lang.NoClassDefFoundError: net/oauth/OAuthValidator
SERVLET:	/lti
CAUSED BY:	java.lang.NoClassDefFoundError: net/oauth/OAuthValidator
CAUSED BY:	java.lang.ClassNotFoundException: net.oauth.OAuthValidator not found by net.oauth.core.oauth_Bundle-Version_20100527_Export-Package__ [67]
```

Is that to be expected? ~As I am seeing LTI version 1.0 in the request to opencast from the plugin and support for that seems to be removed in this PR if I understood correctly?~

I am using a clean build without any patches from this branch.

Edit: I just learned the LTI version and other launch parameters do not vary between 1.0 and 1.1, so the plugin should still work.

lgehrlein:
I got LTI to work again by adding the oauth-provider to the karaf feature.xml:

```xml
    <bundle>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.oauth-provider/20100527_1</bundle>
```

rute-santos:
> I got LTI to work again by adding the oauth-provider to the karaf feature.xml:
> 
> ```
>     <bundle>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.oauth-provider/20100527_1</bundle>
> ```

I'm glad you did! We tested LTI with a very basic setup with Canvas and I confess I know nothing about Moodle or the plugin.
Can you create a PR with your change and I will merge here? Or tell me exactly where I should add it?
Thanks!

assemblies/karaf-features/src/main/feature/feature.xml
```diff
@@ -66,6 +68,8 @@
     <bundle>mvn:org.codehaus.jettison/jettison/${jettison.version}</bundle>
     <bundle>mvn:org.opencastproject/android-mms/1.2</bundle>
     <bundle>wrap:mvn:com.googlecode.json-simple/json-simple/${json-simple.version}</bundle>
+    <!-- Used by the Lti filter -->
```
lgehrlein, line 71: I added the following line here to fix the LTI problem:
```suggestion
    <!-- Used by the Lti filter -->
    <bundle>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.oauth-provider/20100527_1</bundle>
```

KatrinIhler:
Merged as-is as discussed in the Board Meeting., since this fixes multiple important security issues. Institutions are encouraged to do more testing pior to the release of OC 21 in November.

PR #7372, opened 2026-02-07
Add endpoint for batch-creating users

lkiesow:
If you have to create a large set of users, using the current endpoint which can only create a single user at a time is very slow. For example, looking at the demo servers, it takes develop.opencast.org about 1:45h each nicht to create the 25000 users.

This patch allows you to send a single request which will create a bunch of users at once, speeding up the process significantly.

Unfortunately, the users are still added to the database one after another. But changing that would probably require larger changes. But even like this, the process becomes quite a bit faster.

This targets the legacy branch to allow all demo servers to use the same scripts. It's code related to an internal API not usually used in day to day operations with low risk of breaking anything critical, even if this patch was buggy.

### How to test this patch

```
❯ curl -u admin:opencast \
  -X POST \
  -H 'Content-type: application/json' \
  http://127.0.0.1:8080/user-utils/users.json \
  --data-raw '@users.json'
```

with `users.json` being something like:
```json
[
{"username":"a","password":"b"},
{"username":"c","password":"d","roles":["ROLE_XY"]}
]
```

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests
* [x] include a [release note](https://docs.opencast.org/develop/developer/#participate/development-process/#release-notes) if a feature

### DIFF
--- a/modules/userdirectory/pom.xml
+++ b/modules/userdirectory/pom.xml
@@ -120,6 +120,10 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-reload4j</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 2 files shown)


gregorydlogan (approved):

gregorydlogan:
Slight change to how things are tested: `--data-raw '@users.json'` became `--data-binary @users.json`, but it worked fine after.

Issue #7368, opened 2026-02-06
Use StorageUsage in metrics


lkiesow:
We seemingly gather information about available storage space for workspace, working file repository and asset manager by implementing the `StorageUsage` class. Maybe we could include those values in the metrics endpoint?

PR #7376, opened 2026-02-09
Add metric for storage usage

marwyg:
Issue: #7368

This PR simply uses the available methods of the StorageUsage interface for new metrics. The new metrics:

- Total Space
- Usable Space
- Used Space

for

- Asset Manager
- Workspace
- Working File Repository

Here an example:

```properties
# TYPE opencast_storage_available gauge
# HELP opencast_storage_available Shows available and used up space of different directories of Opencast.
opencast_storage_available{part="OsgiFileSystemAssetStore_usedSpace"} 0.0
opencast_storage_available{part="WorkingFileRepositoryRestEndpoint_usedSpace"} 0.0
opencast_storage_available{part="WorkspaceImpl_totalSpace"} 1.021414735872E12
opencast_storage_available{part="WorkingFileRepositoryRestEndpoint_usableSpace"} 9.23058225152E11
opencast_storage_available{part="OsgiFileSystemAssetStore_usableSpace"} 9.23058225152E11
opencast_storage_available{part="WorkspaceImpl_usedSpace"} 0.0
opencast_storage_available{part="WorkingFileRepositoryRestEndpoint_totalSpace"} 1.021414735872E12
opencast_storage_available{part="WorkspaceImpl_usableSpace"} 9.23058225152E11
opencast_storage_available{part="OsgiFileSystemAssetStore_totalSpace"} 1.021414735872E12
```

edit:

the new version looks like this:

```properties
# TYPE opencast_storage gauge
# HELP opencast_storage Shows available and used up space of different directories of Opencast.
opencast_storage{part="WorkspaceImpl",kind="usable"} 9.22632798208E11
opencast_storage{part="OsgiFileSystemAssetStore",kind="used"} 0.0
opencast_storage{part="WorkspaceImpl",kind="total"} 1.021414735872E12
opencast_storage{part="OsgiFileSystemAssetStore",kind="usable"} 9.22632798208E11
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="total"} 1.021414735872E12
opencast_storage{part="WorkspaceImpl",kind="used"} 0.0
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="usable"} 9.22632798208E11
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="used"} 0.0
opencast_storage{part="OsgiFileSystemAssetStore",kind="total"} 1.021414735872E12
```

### DIFF
--- a/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
+++ b/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
@@ -612,4 +612,9 @@ public class AwsS3AssetStore extends AwsAbstractArchive implements RemoteAssetSt
   void setBucketName(String orgId, String bucketName) {
     orgBucketNameMap.put(orgId, bucketName);
   }
+
+  @Override
+  public String getStorageName() {
+    return "Amazon S3 Asset Store";
+  }
 }
--- a/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
+++ b/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
@@ -615,6 +615,6 @@ public class AwsS3AssetStore extends AwsAbstractArchive implements RemoteAssetSt
 
   @Override
   public String getStorageName() {
-    return "Amazon S3 Asset Store";
+    return "S3 Asset Store";
   }
 }
--- a/modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
+++ b/modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
@@ -129,10 +129,10 @@ public class MetricsExporter {
       .help("Version of Opencast (based on metrics module)")
       .labelNames("part")
       .register();
-  private final Gauge storageAvailable = Gauge.build()
-      .name("opencast_storage_available")
+  private final Gauge storage = Gauge.build()
+      .name("opencast_storage")
       .help("Shows available and used up space of different directories of Opencast.")
-      .labelNames("storage")
+      .labelNames("part", "kind")
       .register();
   private Gauge eventsInAssetManager;
 
@@ -206,10 +206,10 @@ public class MetricsExporter {
     }
 
     for (StorageUsage storageUsageImplementation : storageUsageImplementations) {
-      String prefix = storageUsageImplementation.getClass().getSimpleName() + "_";
-      storageAvailable.labels(prefix + "totalSpace").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
-      storageAvailable.labels(prefix + "usableSpace").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
-      storageAvailable.labels(prefix + "usedSpace").set(storageUsageImplementation.getUsedSpace().orElse(-1L));
+      String part = storageUsageImplementation.getClass().getSimpleName();
+      storage.labels(part, "total").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
+      storage.labels(part, "usable").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
+      storage.labels(part, "used").set(storageUsageImplementation.getUsedSpace().orElse(-1L));
     }
 
     // collect metrics
--- a/modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
+++ b/modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
@@ -28,6 +28,7 @@ import org.opencastproject.serviceregistry.api.ServiceRegistration;
 import org.opencastproject.serviceregistry.api.ServiceRegistry;
 import org.opencastproject.serviceregistry.api.ServiceState;
 import org.opencastproject.serviceregistry.api.SystemLoad;
+import org.opencastproject.storage.StorageUsage;
 import org.opencastproject.util.doc.rest.RestQuery;
 import org.opencastproject.util.doc.rest.RestResponse;
 import org.opencastproject.util.doc.rest.RestService;
@@ -44,8 +45,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.StringWriter;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.SortedSet;
 import java.util.stream.Collectors;
 
 import javax.servlet.http.HttpServletResponse;
@@ -88,6 +91,8 @@ public class MetricsExporter {
   /** The logger */
   private static final Logger logger = LoggerFactory.getLogger(MetricsExporter.class);
 
+  private final List<StorageUsage> storageUsageImplementations = new ArrayList<>();
+
   // Prometheus metrics registry for exposing metrics
   private final CollectorRegistry registry = CollectorRegistry.defaultRegistry;
 
@@ -124,6 +129,11 @@ public class MetricsExporter {
       .name("opencast_version")
       .help("Version of Opencast (based on metrics module)")
       .labelNames("part")
+      .register();
+  private final Gauge storageAvailable = Gauge.build()
+      .name("opencast_storage_available")
+      .help("Shows available and used up space of different directories of Opencast.")
+      .labelNames("storage")
       .register();
   private Gauge eventsInAssetManager;
 
@@ -196,6 +206,13 @@ public class MetricsExporter {
       }
     }
 
+    for (StorageUsage storageUsageImplementation : storageUsageImplementations) {
+      String prefix = storageUsageImplementation.getClass().getSimpleName() + "_";
+      storageAvailable.labels(prefix + "totalSpace").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
+      storageAvailable.labels(prefix + "usableSpace").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
+      storageAvailable.labels(prefix + "usedSpace").set(storageUsageImplementation.getUsedSpace().orElse(-1L));
+    }
+
     // collect metrics
     final StringWriter writer = new StringWriter();
     TextFormat.writeOpenMetrics100(writer, registry.metricFamilySamples());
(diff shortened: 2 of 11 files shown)


modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
```diff
@@ -196,6 +205,13 @@ public Response metrics() throws Exception {
       }
     }
 
+    for (StorageUsage storageUsageImplementation : storageUsageImplementations) {
+      String prefix = storageUsageImplementation.getClass().getSimpleName() + "_";
+      storageAvailable.labels(prefix + "totalSpace").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
+      storageAvailable.labels(prefix + "usableSpace").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
+      storageAvailable.labels(prefix + "usedSpace").set(storageUsageImplementation.getUsedSpace().orElse(-1L));
```
mtneug: These methods are not implemented by all asset stores, e.g. [AWS S3](https://github.com/opencast/opencast/blob/996ab59f5a8158ce32555860b44f3adbca0e7796/modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/AwsAbstractArchive.java#L78-L88). I'm not sure if you can simply change the S3 implementation and return an empty `Option`.

mtneug (changes requested):

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
mtneug: Ah my bad. The S3 implementation is not promoting the `StorageUsage` service.

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
```diff
@@ -125,6 +129,11 @@ public class MetricsExporter {
       .help("Version of Opencast (based on metrics module)")
       .labelNames("part")
       .register();
+  private final Gauge storageAvailable = Gauge.build()
+      .name("opencast_storage_available")
```
lkiesow: This results in:
```
opencast_storage_available{part="OsgiFileSystemAssetStore_usedSpace"} 
```

The “storage available used” sounds a bit weird. If the storage is still available, it can't really be used, can it? How about just using `opencast_storage` instead?

```
opencast_storage{part="OsgiFileSystemAssetStore_usedSpace"} 
```

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
```diff
@@ -196,6 +205,13 @@ public Response metrics() throws Exception {
       }
     }
 
+    for (StorageUsage storageUsageImplementation : storageUsageImplementations) {
+      String prefix = storageUsageImplementation.getClass().getSimpleName() + "_";
+      storageAvailable.labels(prefix + "totalSpace").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
```
lkiesow: Can we use two labels instead? Something like `part` and `kind` maybe?
Right now, you get metrics like this:
```
opencast_storage_available{part="OsgiFileSystemAssetStore_usableSpace"} 9.23058225152E11
```
splitting that into two labels could look something like this:
```
opencast_storage_available{part="OsgiFileSystemAssetStore", kind="usable"} 9.23058225152E11
```

This makes it very easy, for example, to create alerting rules for all storage spaces against filling up by just filtering for `usable` instead of having to create a rule for each and every name + type combination.

lkiesow (changes requested):
I like this 🚀 
Still, some minor nitpicks ;-P

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
marwyg: Thank you for taking a look! :) First I put the StorageUsage everywhere possbile, but ran into "not implemented exceptions" :laughing: So I left with the 3 remaining classes.

But maybe it would be nice to have this S3 metric in the future? Or maybe it has its own metrics and there is no need to implement stuff in Opencast for this?

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
marwyg: Makes absolutly sense, I have overlooked that :+1: It looks like this now:

`opencast_storage{part="WorkspaceImpl",kind="total"} 1.021414735872E12`

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
marwyg: This is super nice, I didn't know it works like this. I changed it and now it looks like this:

`opencast_storage{part="WorkspaceImpl",kind="total"} 1.021414735872E12`

marwyg:
Thank you for the review! The current state looks like this now:

```properties
# TYPE opencast_storage gauge
# HELP opencast_storage Shows available and used up space of different directories of Opencast.
opencast_storage{part="WorkspaceImpl",kind="usable"} 9.22632798208E11
opencast_storage{part="OsgiFileSystemAssetStore",kind="used"} 0.0
opencast_storage{part="WorkspaceImpl",kind="total"} 1.021414735872E12
opencast_storage{part="OsgiFileSystemAssetStore",kind="usable"} 9.22632798208E11
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="total"} 1.021414735872E12
opencast_storage{part="WorkspaceImpl",kind="used"} 0.0
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="usable"} 9.22632798208E11
opencast_storage{part="WorkingFileRepositoryRestEndpoint",kind="used"} 0.0
opencast_storage{part="OsgiFileSystemAssetStore",kind="total"} 1.021414735872E12
```

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
mtneug: AFAIK there are no direct APIs to get the used/available storage in S3. So, I would leave it out.

mtneug:
Do we really want to expose the Java class names as labels?

marwyg:
I thought it would be simpler in the future to add more storage metrics, when someone adds something new and puts the StorageUsage class in the OSGi Component declaration?

On the other hand.. if this person doesn't implement the necessary methods like in the AWS Class for example, the metrics endpoint will throw exceptions. Well okay, maybe I should define explicitly for what we would like to show the metrics and then I also can define the names explicitly instead of using the class names.

I am not 100% sure how to call them, would `Workspace`, `Working File Repository` and `Asset Store` be okay?

lkiesow:
I think just having to implement the `StorageUsage` and not having to modify the metrics endpoint as well is actually quite nice. However, it would actually be quite easy to have both I think.

We could extend the `StorageUsage` interface and add something like:
```java
public String getStoragename();
```
This would need to be implemented by every storage class, but you could just make it return a static string. Something like:
```java
public String getStoragename() {
  return "workspace";
}
```
Then you could just call `getStoragename()` to get the value for `part`. Although we then also might want to rename `part` to `name`?

marwyg:
- I added the suggested methods and therefore changed the labels
- I renamed "part" to "name"
- I added a _bytes suffix
- I changed the help text

The result looks like this now:

```properties
# TYPE opencast_storage_bytes gauge
# HELP opencast_storage_bytes Disk space statistics (total, used, usable) for Opencast storage components.
opencast_storage_bytes{name="File System Asset Store",kind="total"} 1.021414735872E12
opencast_storage_bytes{name="Working File Repository",kind="total"} 1.021414735872E12
opencast_storage_bytes{name="Working File Repository",kind="usable"} 9.18900600832E11
opencast_storage_bytes{name="Workspace",kind="usable"} 9.18900600832E11
opencast_storage_bytes{name="File System Asset Store",kind="usable"} 9.18900600832E11
opencast_storage_bytes{name="Workspace",kind="used"} 0.0
opencast_storage_bytes{name="Working File Repository",kind="used"} 0.0
opencast_storage_bytes{name="File System Asset Store",kind="used"} 0.0
opencast_storage_bytes{name="Workspace",kind="total"} 1.021414735872E12
```

modules/asset-manager-storage-aws/src/main/java/org/opencastproject/assetmanager/aws/s3/AwsS3AssetStore.java
```diff
@@ -612,4 +612,9 @@ void setS3TransferManager(TransferManager s3TransferManager) {
   void setBucketName(String orgId, String bucketName) {
     orgBucketNameMap.put(orgId, bucketName);
   }
+
+  @Override
+  public String getStorageName() {
+    return "Amazon S3 Asset Store";
```
lkiesow: Since so many systems nowadays implement S3, I would omit `Amazon`. That could confuse people:
```suggestion
    return "S3 Asset Store";
```

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
```diff
@@ -125,6 +129,11 @@ public class MetricsExporter {
       .help("Version of Opencast (based on metrics module)")
       .labelNames("part")
       .register();
+  private final Gauge storage = Gauge.build()
+      .name("opencast_storage_bytes")
+      .help("Disk space statistics (total, used, usable) for Opencast storage components.")
```
lkiesow: We don't actually have to change this. But while reading through this again, I thought that people usually talk about `free` space instead of `usable` space. But that's really nitpicking. 
```suggestion
      .help("Disk space statistics (total, used, free) for Opencast storage components.")
```

modules/metrics-exporter/src/main/java/org/opencastproject/metrics/impl/MetricsExporter.java
```diff
@@ -196,6 +205,13 @@ public Response metrics() throws Exception {
       }
     }
 
+    for (StorageUsage storageUsageImplementation : storageUsageImplementations) {
+      String storageName = storageUsageImplementation.getStorageName();
+      storage.labels(storageName, "total").set(storageUsageImplementation.getTotalSpace().orElse(-1L));
+      storage.labels(storageName, "usable").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
```
lkiesow: `usable` vs `free`. See above

```suggestion
      storage.labels(storageName, "free").set(storageUsageImplementation.getUsableSpace().orElse(-1L));
```

lkiesow (commented):
Tested this and I think this looks pretty good.
Two nitpics, but in general I think this is ready to be merged.

marwyg:
As suggested by @lkiesow I changed the label "usable" to "free" (it sound better) and also renamed `Amazon S3 Asset Store` to `S3 Asset Store`.

marwyg:
Because of conflicts with the develop branch, I rebased and updated my PR. The rebase ran successfully without any problems.

lkiesow:
Force-pushed to re-trigger actions

PR #7383, opened 2026-02-10
Add support for uploading captions to YouTube

JulianKniephoff:
This commit extends the YouTube publication service and the `publish-youtube` workflow operation handler to support uploading caption tracks along with the video.

The metadata for the uploaded captions is determined as follows:
- Language: Extracted from the media package element's tags. The service looks for a tag starting with `lang:` and uses the remainder as the BCP-47 language code (e.g., `lang:en`).
- Name: The flavor of the caption track is used as the caption track name on YouTube (e.g., `captions/vtt`).

The `publish-youtube` workflow operation now includes two new configuration options, `caption-flavors` and `caption-tags`, to specify which elements should be uploaded as captions.

### DIFF
--- a/docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/publish-youtube-woh.md
@@ -16,10 +16,12 @@ with both streams inside of it.
 Parameter Table
 ---------------
 
-|configuration keys         |description                                                                   |
-|---------------------------|------------------------------------------------------------------------------|
-|source-flavors             |The flavors to publish to YouTube                                             |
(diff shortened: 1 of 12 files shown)


ypatios (approved):
@JulianKniephoff this looks good, as far as I can tell, clean code. thank you.
Just for confirmation: there's no explicit additional config needed for this feature (apart from the direct config of the WOH, eg "caption-flavors", right?

JulianKniephoff:
Exactly. :) There is no mapping needed here like in #7206 and #7205 because the language encoded in the tags of the subtitle tracks already have the correct format (BCP 47) by convetion: https://docs.opencast.org/develop/admin/#configuration/subtitles/#tags.

PR #7386, opened 2026-02-11
Simplify referring to the various project roots in Maven config

JulianKniephoff:
The main thing this does is removing the `opencast.basedir` property, which we maintained throughout the module hierarchy to be able to refer to the root of the entire project. This is redundant, since Maven already offers such a property:

    maven.multiModuleProjectDirectory

The other small thing is it unifies the usage of `basedir` vs. `project.basedir`, which are the same, and which refer to the root of an individual module.

### DIFF
--- a/assemblies/dist-admin/pom.xml
+++ b/assemblies/dist-admin/pom.xml
@@ -8,7 +8,6 @@
     <relativePath>../pom.xml</relativePath>
   </parent>
   <properties>
-    <opencast.basedir>${project.basedir}/../..</opencast.basedir>
     <enableJobDispatching>true</enableJobDispatching>
   </properties>
   <artifactId>opencast-dist-admin</artifactId>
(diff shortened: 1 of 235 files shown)


lkiesow:
Can you give us a description of what this does?

JulianKniephoff:
@lkiesow I added a description here and in the commit. 🫣

JulianKniephoff:
Unfortunately I have to take this back, actually. Doing some more research I learned that said property is private: https://github.com/apache/maven/issues/8662.

Maven 4 will get `project.rootDirectory` for this, but that's not out, yet: https://github.com/apache/maven/pull/1061, https://maven.apache.org/whatsnewinmaven4.html#declaring-the-root-directory-and-directory-properties

We **could** rely on a plugin for this: https://github.com/jdcasey/directory-maven-plugin, but I'm not going to push this further for now. 🤷‍♀️

PR #7387, opened 2026-02-11
Improve YouTube video retraction

JulianKniephoff:
Previously, the retraction of YouTube videos relied on searching for the video by its episode name. This was unreliable if multiple videos had the same name or if the name had changed.

This commit changes the retraction logic to extract the YouTube video ID directly from the publication's URI stored in the MediaPackage.

Note that during retraction we now attempt to remove the video from the specific playlist referenced in the publication URL (the YouTube watch URL includes both `v` and `list` parameters). This removal is best-effort and failures are ignored; we do not try to remove the video from any other playlists. Since a video can be part of multiple playlists, including ones not managed by Opencast, broad automated removals remain intentionally unsupported.


ypatios (approved):
@JulianKniephoff thanks for this final addition (super important to have robust deletion). It all looks good and clean from my POV. great work!

PR #7396, opened 2026-02-17
Matrix Notification Workflow Operation

### LINKED ISSUE #7365: Enable workflows to send matrix notifications
Similar to the send-email WOH.

lkiesow:
This patch provides an initial version of a Matrix notification workflow operation. You can use this, for example, to notify you as part of the error workflow if the processing of an event has failed.
 
In the future, this can easily be expanded to use metadata to determine where to send notifications or to include in the notification messages. But that's for another patch.
 
IMG20260217184710

This closes #7365.

**How to test this**
 
Get an `access_token` you can use to authenticate as the user you want to send
the notifications from:
```
❯ USER=@someone:matrix.org
❯ PASS=someVerySecretPassword
❯ curl -XPOST \
  -H 'Content-Type: application/json' \
  -d '{"type":"m.login.password", "user":"'"${USER}"'", "password":"'"${PASS}"'"}' \
  "https://matrix.org/_matrix/client/r0/login"
{..., "access_token":"aSecretSessionToken", ...}
```
 
Edit `etc/org.opencastproject.workflow.handler.matrix.MatrixNotificationWorkflowOperationHandler.cfg`
and set home server and token:
```properties
matrix.server.url=https://matrix.org
matrix.access.token=secret_...
```
 
Create a workflow `etc/workflows/matrix.yaml`:
```yaml
---
id: matrix
title: Test Workflow
operations:
  - id: matrix-notify
    description: Notify Matrix about workflow completion
    configurations:
      - room-id: "!<room-identifier>:matrix.org"
      - message: Workflow has been completed successfully
```
 
Run the workflow:
```
❯ curl http://localhost:8080/ingest/addMediaPackage/matrix \
  -u admin:opencast \
  -F title="I 🖤 Opencast"
```


modules/matrix-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/matrix/MatrixNotificationWorkflowOperationHandler.java
```diff
@@ -0,0 +1,266 @@
...
+  @Override
+  public WorkflowOperationResult start(WorkflowInstance workflow, JobContext context)
+          throws WorkflowOperationException {
+
+    var mediapackage = workflow.getMediaPackage();
+    logger.debug("Running Matrix notification workflow operation on media package {}", mediapackage);
+
+    // Required configuration
+    var roomId = getConfig(workflow, OPT_ROOM_ID);
+    var message = getConfig(workflow, OPT_MESSAGE);
```
mtneug, line 204: As noted in #7365, we now have a couple of operations in this direction: `mattermost-notify`, `send-email`, `http-notify`, more?

What I find frustrating is that each has a custom message/body parameter. `send-email` have a complex templating mechanism. `mattermost-notify` has some predefined message variables. `http-notify` seems to be similar to this implementation with no special templating. Can this be unified? At least for `mattermost-notify` and Matrix that are probably closed in terms of purpose.

docs/guides/admin/docs/workflowoperationhandlers/matrix-notification-woh.md
```diff
@@ -0,0 +1,44 @@
...
+|configuration keys | description
+|---------------- --| ----------------------------------------------
+|room-id            | The Matrix room ID to send the notification to
+|message            | The message body to send
+
+Operation Example
+-----------
+
+```yaml
+  - id: matrix-notification
```
mtneug: Can you rename this to `matrix-notify` to use the same naming pattern as `mattermost-notify` and `http-notify`?

mtneug (changes requested):

docs/guides/admin/docs/workflowoperationhandlers/matrix-notification-woh.md
lkiesow: That makes sense. I've changed the name. It's now`matrix-notify'.

modules/matrix-notification-workflowoperation/src/main/java/org/opencastproject/workflow/handler/matrix/MatrixNotificationWorkflowOperationHandler.java
lkiesow: I thought about that, but they are simple enough and there is enough difference, that having different operations still make sense, I think. Therefor I would leave it like this.

However, some of these notification mechanism do support some form of templating. I may make sense in the future to pull the metadata and template handling into a parent class or have a helper class to have a unified approach for that. But that's outside the scope of this pull request.

docs/guides/admin/docs/workflowoperationhandlers/matrix-notification-woh.md
```diff
@@ -0,0 +1,44 @@
...
+
+Operation Example
+-----------
+
+```yaml
+  - id: matrix-notify
+    description: Notify Matrix about workflow completion
+    configurations:
+      - room-id: !room-id!
+      - message: Workflow has been completed successfully
```
gregorydlogan: Quotes are *not* optional at the very least on the `room-id` key.  Spent way too much time figuring that out 😅 

```suggestion
      - room-id: "!room-id!"
      - message: "Workflow has been completed successfully"
```

gregorydlogan (changes requested):
Tested working, aside from the doc complaint

PR #7397, opened 2026-02-17
Add automated forward chain merge workflow for `r/X.x` release branches

### LINKED ISSUE #2397: Automate inter-branch merges
Patches are automatically merged from older to newer branches like this:

```
r/8x  --->  r/9.x  --->   develop
```

These merges are currently done by committers but could be mostly automated.
For this, it would be nice to create a GitHub Actions workflow which:

- Reacts to a merge in a release branch (`r/.*` e.g. `r/8.x`)
- Checks out the next higher release branch or `develop` (e.g. `r/9.x`)
- Merges the older release branch into the newer one (`r/8.x --> r/9.x`)
    - Opens a pull request if there is a merge conflict indicating the conflict to committers
- Runs a build + tests
- Pushes the merge

Ideally, the workflow would also:

- Cancel older running builds on the same branch to avoid unnecessarily many merges
- In case of a merge conflict, assign the conflict to the pull request author and maybe the committer who merged the pull request causing the error.

ferishili:
This PR introduces an automated forward chain merge workflow for release branches which fixes #2397.

[UPDATE]: this PR has been refactored and updated, for more info please follow [this comment](https://github.com/opencast/opencast/pull/7397#issuecomment-4178396988)

### What it does

The workflow automatically performs forward merges when changes are pushed to release branches following the pattern:

```
r/X.x
```

When a push happens on such a branch, the workflow:

1. Detects the source release branch (e.g. `r/1.x`)
2. Determines all higher release branches (`r/2.x`, `r/3.x`, …)
3. Appends `develop` as the final target
4. Sequentially merges forward in a chain:

```
r/1.x → r/2.x → r/3.x → ... → develop
```

Each merge uses a non-fast-forward merge to preserve history clarity.

The workflow can also be triggered manually via `workflow_dispatch`.

---

### Safety and Validation

For each forward step the workflow:

* Verifies module parent versions are consistent
* Stops on merge conflicts
* Creates a PR automatically if a conflict occurs
* Verifies branch integrity after pushing
* Stops the chain if any failure occurs

This prevents silent drift between release lines.

---

### Matrix Notification

If configured, the workflow sends a Matrix notification when:

* A merge conflict occurs
* A validation check fails
* A push fails

To enable Matrix notifications, the following repository secrets must be configured:

* `MATRIX_BASE_URL`:  https://your-matrix-server

* `MATRIX_ROOM_ID`:  The target room ID (e.g. `!abcdef:matrix.org`)

* `MATRIX_ACCESS_TOKEN`: A user access token with permission to post messages

---

### Permissions Required

The repository must allow GitHub Actions to:

* Create and approve pull requests
  (Settings → Actions → General → Workflow permissions)

The workflow uses:

```yaml
permissions:
  contents: write
  pull-requests: write
```

---

### Why this is useful

This ensures that fixes applied to lower release branches are consistently propagated forward, reducing manual work and minimizing the risk of missing merges between maintained versions.

It standardizes forward porting and provides visibility in case of conflicts.

### Points to improve:
- Matrix communication could be optional, and or use the webhook approach instead of user token!


.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,241 @@
...
+          git config user.name "github-actions"
+          git config user.email "github-actions@github.com"
+
+      - name: Determine source branch
+        id: source
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "branch=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
+          else
+            echo "branch=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
```
gregorydlogan: Aren't these two echo lines the same?  Are they supposed to be?

gregorydlogan (commented):
This honestly is probably more rigorous than most RMs when merging forwards, so that's good at least.  I'm not sure I would want to run this for *every* push, but I guess let's see what others want.  I generally batch my work anyway, so I'll end up pushing between 2 and 10 PRs at the same time, which would then be immediately pushed fowards.  If people are merging one PR at a time, this might make the history *very* noisy.

Future concern: This will need to properly handle submodules, once we get there.  If this goes in first then I'll patch it appropriately at that time.

gregorydlogan:
Thinking about this more, and *actually reading* @lkiesow's [comment](https://github.com/opencast/opencast/issues/2397#issuecomment-3910257991), it would be *really* nice for each forward merge to independently run Opencast's tests - the same way a PR or a push does.

Right now what you're doing is a single branch immediately merges all the way up to `develop`, and then pushes.  This works great as long as nothing in that chain breaks at runtime - but that's not always true.  Could we instead have a given branch only merge forward *one* branch (so, 17->18, or 18->19, but not 17->18->19 in one run), and then let the newer branch run the same checks it does on a push and *only then* does the next forward merge happen?

I guess what I'm asking is, can this new forward merge be conditional on the exit status of one or more *other* GHA workflows.  If so, then the algo looks like this:

On push to `r/N.x`:
Run all `r/N.x` tests
if all `r/N.x` tests passing, forward merge to `r/(N+1).x`

Rinse, repeat for each branch forward unto `develop`, which is not in the list of push targets for this workflow.

.github/workflows/auto-forward-merge-chaning.yml
ferishili: uups, of course! will fix it soon

ferishili:
Further changes will follow soon!

ferishili:
The auto forward-merge workflow has now been refactored to include test coverage as well. Please have a look, @gregorydlogan.

It now runs the test-* workflows before pushing, so tests are validated as part of the forward-merge process.

One important requirement to make this workflow fully chainable is creating the `CHAINING_SYNC_TOKEN` repository secret via the GitHub token generation page. The token must have **no expiration date** and include the **repo** and **workflow** scopes.

If this secret is not configured, forward chaining will stop after the next target branch instead of continuing through the full chain.

ferishili:
Samples for Element-Matrix notifications:

#### Default and cleanup errors:

Screenshot 2026-04-02 at 16 47 19

#### Conflicts:

Screenshot 2026-04-02 at 16 47 26

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  target:
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'push'
```
gregorydlogan: Need to filter by repository, otherwise this will do *hilarious* things to people's repo histories :D

```suggestion
      (github.event_name == 'workflow_dispatch' || github.event_name == 'push') &&
      github.repository == 'opencast/opencast'
```

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+        with:
+          fetch-depth: 0
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          base: ${{ needs.target.outputs.branch }}
+          filters: |
+            paella:
+              - '.github/workflows/test-paella-7.yml'
+              - 'modules/engage-paella-player-7/**'
```
gregorydlogan, line 226: These don't exist, and potentially won't exist any more depending on the outcome of #7592

```suggestion
```

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+              - 'mvnw'
+              - '.mvn/**'
+
+  tests-paella:
+    needs: [merge-attempt, detect-changes]
+    if: needs.detect-changes.outputs.paella == 'true'
+    uses: ./.github/workflows/test-paella-7.yml
+    with:
+      ref: ${{ needs.merge-attempt.outputs.temp_ref }}
+
```
gregorydlogan, line 256: Same as above re: paella 7

```suggestion
```

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+  # For this to perform chaining admin must create the repository secret called "CHAINING_SYNC_TOKEN",
+  # generated by GitHub Token Generation Page with [no expiration date, scopes: repo, workflow]
+  push-attempt:
+    env:
+      GH_TOKEN: ${{ secrets.CHAINING_SYNC_TOKEN || secrets.GITHUB_TOKEN }}
+    needs: [ # Everything must be good!
+      target,
+      merge-attempt,
+      detect-changes,
+      tests-paella,
```
gregorydlogan, line 301: Moar Paella.

```suggestion
```

gregorydlogan (changes requested):
Still testing this (since it takes *forever* as it grinds away on all the tests), but looks good in general.  Some minor issues related to Paella 7 here, buuut....

This needs to go into older versions I would think - *maybe* 18, but more likely 19 or 20.  If it just goes into `develop` then it does not catch changes to the older versions, so it would realistically only start helping once we cut an `r/21.x` branch.

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+
+on:
+  push:
+    branches:
+      - 'r/*'
+  workflow_dispatch:
+
+concurrency:
+  group: auto-forward-chain-${{ github.ref }}
+  cancel-in-progress: true
```
gregorydlogan, line 11: Some part of this likely isn't quite working, though I don't know what:

Image

The bottom run is in the middle of the OC tests, the top run just started but long enough ago that I would have expected the bottom run to have been killed by now.

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+        run: git push origin --delete ${{ needs.merge-attempt.outputs.temp_ref }}
+        env:
+          GH_TOKEN: ${{secrets.GITHUB_TOKEN }}
+
+  matrix-notification:
+    runs-on: ubuntu-latest
+    # We only need to listen to the critical jobs
+    needs: [target, merge-attempt, push-attempt, cleanup]
+    # Trigger if merge-attempt failed OR push-attempt (which implies tests failed) OR cleanup failed
+    if: always() && (needs.merge-attempt.result == 'failure' || needs.push-attempt.result == 'failure' || needs.cleanup.result == 'failure')
```
gregorydlogan, line 349: Unsure of this logic.  Why the `always()`?  It might be nice to trigger this on the tests failing as well.  I'm seeing failures in testing unrelated to these changes that don't trigger Matrix notifications.

.github/workflows/auto-forward-merge-chaning.yml
ferishili: Considering this workflow is going to land in 18.x, they should still be there, right?
PS: I will rebase this to 18.x as soon as we decide!

gregorydlogan: Yes, they would be needed in 18.  Probably need a follow-on PR to remove them since the auto forward merge is going to drag them forward :D

.github/workflows/auto-forward-merge-chaning.yml
```diff
@@ -0,0 +1,380 @@
...
+    # Run if merge-attempt created a branch, regardless of push success
+    if: always() && needs.merge-attempt.result == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository to make git branches known to this job
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Delete temporary branch
+        run: git push origin --delete ${{ needs.merge-attempt.outputs.temp_ref }}
```
gregorydlogan, line 340: I'm seeing errors unrelated this, but perhaps a failure of this should be ignored.  I've somehow gotten things into an error state in my testing where this branch *does not exist*, and then I get a Matrix notification that it can't be deleted and I need to clean it up :)

The only caveat to ignoring this that I can see is that if we somehow don't have write permissions to the repository then we'd have orphaned branches buuuut, if you're messing around with repository permissions *while this is running* then you're likely in for a bad time regardless.

PR #7456, opened 2026-03-03
Reindex more things

gregorydlogan:
This PR prompts users reindexing for allow-by-episode-role-id to reindex more things.  As reported in Matrix, only reindexing the ACL gets you events without any metadata.

This targets 18.x since that's where the index migration happens.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests
* [x] ~include a [release note](https://docs.opencast.org/develop/developer/#participate/development-process/#release-notes) if a feature, or supports a feature (ie, backend part of a frontend feature)~

### DIFF
--- a/docs/guides/admin/docs/upgrade.md
+++ b/docs/guides/admin/docs/upgrade.md
@@ -104,6 +104,6 @@ the present files `event_mapping.json`, `series_mapping.json`, `theme_mapping.js
 
 [[#6696](https://github.com/opencast/opencast/pull/6696)] only requires an update to the indexed Elasticsearch/Opensearch
 ACLs if you haven't used this feature before. **If this feature was already enabled, no further actions are required.**
-If it was not enabled before upgrading, you can start the rebuild by calling the `/index/rebuild/AssetManager/ACL` and
-`/index/rebuild/Search` endpoints. These endpoints will reindex the event ACLs in both the AssetManager index and the
+If it was not enabled before upgrading, you can start the rebuild by calling the `/index/rebuild/AssetManager` and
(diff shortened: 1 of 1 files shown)


lkiesow:
What you are reporting here essentially means that calling the `.../ACL` endpoint effectively destroys your index? It feels like just changing the documentation for the upgrade isn't sufficient.

Do we know why this is happening? Can we easily fix this? We probably should if we can?

If this is hard to fix, we might as well remove the ACL endpoint? That's still preferable to having something that corrupts your data if used.

gregorydlogan:
The time I saw it causing corruption is if you drop the index completely, then *only* call to reindex the ACL sub-index.  This is, honestly, expected behaviour in that you're telling the system to reindex the ACLs *and only the ACLs*.

An argument could be made to better protect the user from doing something dumb, but if you have permissions to drop the index completely then there's really only so much we can do.

I'm fine removing the subindex options and just requiring a full AM reindex, but I haven't really thought through the implications of that for real adopters.  There's probably a usecase for only doing sub-sections, which would be why that endpoint was added in the first place.

gregorydlogan:
Any objections to getting this in as is @lkiesow?

Arnei:
@gregorydlogan Should this be brought up at the dev meeting?

gregorydlogan:
Will bring it up, but I'm betting we don't do anything with this now. A full reindex being required for 21 and lots of adopters past this already makes me wonder if we need to address this still.

gregorydlogan:
During discussion at the meeting, I decided to drop this.

PR #7461, opened 2026-03-05
Refactor docker compose service configuration

JulianKniephoff:
This is just a backport of #7334 to 18.x, since I still have to muck around in that as well for what I'm doing right now, and I'm annoyed. 😅

### DIFF
--- a/docs/scripts/devel-dependency-containers/docker-compose-all-sql.yml
+++ b/docs/scripts/devel-dependency-containers/docker-compose-all-sql.yml
@@ -1,6 +1,5 @@
 services:
-  elasticsearch:
-    container_name: opencast-opensearch
+  opensearch:
     image: opencast/opensearch:1
     init: true
     build:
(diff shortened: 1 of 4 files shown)


docs/scripts/devel-dependency-containers/docker-compose-all-sql.yml
```diff
@@ -1,6 +1,5 @@
 services:
-  elasticsearch:
-    container_name: opencast-opensearch
```
lkiesow, line 3: Why drop this? I find it very handy to have since it makes it much easier to clean up if you spun up e.g. the `all-sql` earlier and then restarting using the container compose file with no databases. You don't end up with random names. You can just always clean up the same environment.

JulianKniephoff: The names aren't random, right? There is a deterministic algorithm. I also don't see how not having this would make cleanup any harder, can you elaborate? 🤔

JulianKniephoff: Re. why: I hate quoting myself, but:

> This allows container and volume names to be overridden using the -p option of docker compose for better flexibility. For example, this enables running multiple instances of the same services simultaneously for different projects or testing scenarios.
> —@JulianKniephoff in #7334

😅

PR #7476, opened 2026-03-13
Get listproviders for metadata from External API

KatrinIhler:
A while ago we stopped returning collections for metadata fields with listproviders attached from the External API to improve performance. However, that means that an external application has no way to map a value for one of those fields to its "human-readable value", meaning what the Admin UI would show you, or show a selection of alternative values.

By including what listprovider is attached to that metadata field (an information we already have internally), the application could fetch the collection itself from the listproviders service if it wants to, without impacting performance for the External API.

Since this adds an attribute to the output data, this counts up the API version as well.

Example data:
```
  {
    "identifier": "0a81278e-2865-484c-adcf-c0f6d407ca51",
    "metadata": [
      {
        "flavor": "dublincore/episode-extended",
        "title": "EVENTS.EVENTS.DETAILS.CATALOG.EPISODE",
        "fields": [
          {
            "translatable": false,
            "readOnly": false,
            "id": "faculty",
            "label": "Fachbereich",
            "type": "text",
            "value": "FACULTY1",
            "required": true,
            "listprovider": "FACULTIES" <- new!
          }
        ]
      }
  }
```

### DIFF
--- a/docs/guides/admin/docs/releasenotes/external api
+++ b/docs/guides/admin/docs/releasenotes/external api
@@ -1,0 +1,1 @@
+Since we added an optional listprovider attribute to returned metadata fields, the API version was increased to 1.12.0.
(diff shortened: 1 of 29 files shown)


Arnei (approved):
Works and code looks reasonable. All good :)

PR #7489, opened 2026-03-19
Order by _score and _doc

geichelberger:
This orders the search results by _score and _doc with the least priority. Ordering by _score has only an effect on search results; otherwise, the value is the same for every entry. _doc can be used to stabilize the result order.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests


gregorydlogan (commented):
So, the search index (`SearchServiceIndex`) doesn't actually use this method, instead using the high level client directly ([here](https://github.com/opencast/opencast/blob/15680cf5a252796bf18c1a9ccb4a10f65a46183c/modules/search-service-impl/src/main/java/org/opencastproject/search/impl/SearchServiceIndex.java#L210)).  Fixing that is outside of the scope here, but should we apply the same sorting to the search results?  Or does that matter?

geichelberger:
It would make sense to apply the same sorting to the search service index.

geichelberger:
I fixed it for the search index, too.

PR #7540, opened 2026-04-02
Show user roles without associated users in the user table in the ACL…

### LINKED ISSUE #1361: User ACLs for not registered users in opencast are listed in the event/series access policy view as non-user ACLs
If an event or series has user ACLs assigned for users who are not yet registered in Opencast, the user ACL is listed with the non-user roles instead of the user roles:

Image

Use case:
In our case, ILIAS creates most of the series. ILIAS can assign ACLs to users who are not yet registered in Opencast. Once the Opencast user account is created, the user can log in and access his series and videos directly.

Arnei:
…s in the admin ui. Fixes https://github.com/opencast/admin-interface/issues/1361

If you have user role sanitization off, the admin ui will display roles in the ACL tab in two categories, "users" and "roles and groups". Currently, the definition of a role for the users category is that there should exist a user for the username after the user role prefix, in Opencast or otherwise. E.g. if the role is ROLE_USER_GEORGE, it will be sorted in the users category only if there is a user "george" in Opencast or if "george" can be provided by a user provider.

This patch changes the definition. Now, a role only needs to start with the user role prefix to be sorted into the users category. E.g. the role ROLE_USER_GEORGE would now always show up in the users category, no matter if such a user actually exists.

This should be more in line with expected behaviour, and help with certain use cases where a user does not yet exist (meaning it should help with whatever ILIAS is doing).
The downside is that now strings such as ROLE_USER_GEORGE can appear next to usernames like "Theodore" in the users category, which may be strange to some users.

### How to test this patch

Make sure to set `sanitize=false` in `etc/org.opencastproject.userdirectory.UserIdRoleProvider.cfg`.
If you don't have an ILIAS to test with, you can define custom user roles for non-existing users in the `etc/custom.properties` under `org.opencastproject.security.custom.roles`.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-service/src/main/java/org/opencastproject/adminui/endpoint/AclEndpoint.java
+++ b/modules/admin-service/src/main/java/org/opencastproject/adminui/endpoint/AclEndpoint.java
@@ -316,9 +316,12 @@ public class AclEndpoint {
       jsonRole.put("organization", role.getOrganizationId());
       jsonRole.put("isSanitize", isSanitize());
       if (!isSanitize()) {
+        boolean isUserRole = role.getName().startsWith(getUserRolePrefix());
         User user = userDirectoryService.loadUser(role.getName().replaceFirst(getUserRolePrefix(), ""));
         if (user != null) {
           jsonRole.put("user", generateJsonUser(user));
+        } else if (isUserRole) {
+          jsonRole.put("user", new HashMap<String, Object>());
         }
       }
       jsonRoles.add(jsonRole);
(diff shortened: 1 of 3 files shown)


snoesberger:
User roles without associated users are now displayed correctly alongside the known user roles. However, I experience strange behaviour as soon as an ACL template is active. In this case, if I add a new user to the user ACL, the new user overwrites the second-last user in the list. Please also refer to the video for a better understanding.

https://github.com/user-attachments/assets/b2a5a1a4-c793-402c-a4ce-2648bf3d4123

Arnei:
I'm assuming what you are seeing is not because of this PR, but because of something else and possibly fixed by https://github.com/opencast/admin-interface/pull/1576

KatrinIhler (approved):

PR #7547, opened 2026-04-07
Add possibility to upload audio-only as presentation track

### LINKED ISSUE #7379: Audio-only uploads fail when using `presentation` flavor instead of `presenter`
The recently added audio-only upload feature currently only accepts audio files with the `presenter` flavor.

If an audio-only upload is submitted using the `presentation` flavor, the workflow fails immediately with an error indicating that no resources with the `presenter` flavor can be found.

This behavior is confusing, since `presentation` is a commonly used and valid flavor, and there is no clear indication that audio-only uploads require `presenter` specifically.

jduehring:
This PR adds the necessary operations to be able to upload and process audio-files that were uploaded as a `presentation` track. This aims to fix https://github.com/opencast/opencast/issues/7379 . 

### How to test this patch

Upload a audio file with the `presentation` flavor and see if it is processed correctly. 

**Note:**  There currently appears to be an issue with playback of audio-only files in Paella Player 8, as such streams are not being detected. This is unrelated to this PR and likely needs to be fixed in the paella8 repository. For now, please test using Paella Player 7.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/workflows/partial-preview.yaml
+++ b/etc/workflows/partial-preview.yaml
@@ -58,7 +58,7 @@ operations:
       - time: 50%
 
   - id: cover-image
-    if: NOT (${presenter_source_video} OR ${presentation_source_video})
+    if: NOT (${presenter_source_video} OR ${presentation_source_video}) AND ${presenter_source_audio}
     description: Create a cover image for audio-only files
     configurations:
       - stylesheet: file://${karaf.etc}/branding/coverimage.xsl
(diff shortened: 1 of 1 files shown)


Arnei (commented):
It is unfortunate that this bloats our workflows, but I don't see a better way either.

Can confirm that this works for Paella 7.

What does not work is uploading an audio track for both presenter and presentation, that results in this Paella error: `Error: Audio only video stream must be main audio player. Check the role property at video manifest`. Although I do not know what the use case for that would be, so maybe it is quite alright to not handle that case?

Arnei:
@jduehring poke

jduehring:
@Arnei I had a look at this again. 

> What does not work is uploading an audio track for both presenter and presentation, that results in this Paella error: Error: Audio only video stream must be main audio player. Check the role property at video manifest.

So I could check if both presenter and presentation audio have audio-only uploaded and then continue with only one of them. However this would further bloat the workflows with only a small benefit? 

We could of course also just prevent audio-only uploads for one flavor, but that would defeat the purpose of this PR. 

Imo we could leave it as it is?  I don't really see another good way right now.

Arnei (approved):

PR #7554, opened 2026-04-08
Fix typo in library tag in pom.xml

marwyg:
Just a typo in a pom.xml


mtneug:
Ehm, apparently I introduced this typo in Opencast 9.0. So, how did that work all these years?

gregorydlogan:
Not sure, but it was somehow included in the additional libraries exactly as you tried to set it.  I'm betting it's a transitive dependency of something *else* in there...

PR #7574, opened 2026-04-14
Remove unused(?) waveform operation

jduehring:
The `partial-preview` workflow contains two waveform operations. While the second one is used in the editor, I was unable to find any use for the first.

If anyone knows why it was added, please let me know. Otherwise, I suggest removing it.

### How to test this patch

Just upload a video using the default workflow and see if everything works as expected and the generated waveform still shows up in the editor. 

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/etc/workflows/partial-preview.yaml
+++ b/etc/workflows/partial-preview.yaml
@@ -30,6 +30,19 @@ operations:
       - encoding-profile: mp4-preview.http
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+  # Audio waveform
+
+  # Create the waveform.
+  # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+  - id: waveform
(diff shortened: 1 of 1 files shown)


Arnei (changes requested):
I think you managed to remove the wrong one of the two ^^. `/waveform` is the flavor used by the editor (per default, is configurable).

The other waveform operation should be safe to remove though. At least I cannot think of a place that uses the `/audio+preview` flavor anymore (was used by the admin ui integrated editor I believe). (`preview` flavors are usually used in such places as the media module or the lti tools).

jduehring:
> I think you managed to remove the wrong one of the two ^^. `/waveform` is the flavor used by the editor (per default, is configurable).

Yeah, I think I got confused with the auto generated waveforms when testing. Thanks! Fixed it and removed the correct operation.

Arnei (approved):

PR #7646, opened 2026-05-11
Add admin ui endpoint for users to self update

Arnei:
Adds a new users endpoint to the admin ui. It should allow users to update their own user data, without requiring rights to be able to edit *all* user data.
For now, this only allows users to change their own password, as that is all that the admin ui requires right now.

### How to test this patch

Test this together with the frontend patch https://github.com/opencast/admin-interface/pull/1604 and various users.

### Your pull request should…

* [ ] have a concise title
* [ ] pass automated tests

### DIFF
--- a/modules/admin-service/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
+++ b/modules/admin-service/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
@@ -506,6 +506,49 @@ public class UsersEndpoint {
     }
   }
 
+  @PUT
+  @Path("self")
+  @RestQuery(
+      name = "selfUpdate",
+      description = "Allow the user to update their own user data. Currently only changing password is allowed",
+      returnDescription = "Status ok",
+      restParameters = {
+          @RestParameter(description = "The password.", isRequired = false, name = "password", type = STRING),
+      },
+      responses = {
+          @RestResponse(responseCode = SC_OK, description = "User password has been updated."),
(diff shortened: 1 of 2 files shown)


modules/admin-service/src/main/java/org/opencastproject/adminui/endpoint/UsersEndpoint.java
```diff
@@ -506,6 +506,50 @@ public Response updateUser(@PathParam("username") String username, @FormParam("p
...
+          @RestParameter(description = "The password.", isRequired = false, name = "password", type = STRING),
+      },
+      pathParameters = @RestParameter(name = "username", type = STRING, isRequired = true, description = "The "
+          + "username"),
+      responses = {
+          @RestResponse(responseCode = SC_OK, description = "User password has been updated."),
+          @RestResponse(responseCode = SC_FORBIDDEN, description = "Not enough permissions to change password."),
+          @RestResponse(responseCode = SC_BAD_REQUEST, description = "Invalid data provided.")
+      })
+  public Response selfUpdatePassword(@PathParam("username") String username, @FormParam("password") String password) {
```
gregorydlogan: Why are we passing the username here when we only accept the current user's username later on? Unless the future plans are for the ui to use this endpoint with sudo or admin permission checks?

Though whatever the current admin user modal uses can already do that...

Arnei: Nah you're right, we can drop passing the username. I could not even explain why I thought this was safer.

PR #7670, opened 2026-05-21
Only switch org if orgs are different

KatrinIhler:
As mentioned in the last Dev meeting - it doesn't make sense to throw org switching errors on a single-tenant system where the org header and the current organization are both set to `mh_default_org`.

### DIFF
--- a/modules/kernel/src/main/java/org/opencastproject/kernel/security/RemoteUserAndOrganizationFilter.java
+++ b/modules/kernel/src/main/java/org/opencastproject/kernel/security/RemoteUserAndOrganizationFilter.java
@@ -125,7 +125,7 @@ public class RemoteUserAndOrganizationFilter implements Filter {
 
       // See if there is an organization provided in the request
       String organizationHeader = httpRequest.getHeader(ORGANIZATION_HEADER);
-      if (StringUtils.isNotBlank(organizationHeader)) {
+      if (StringUtils.isNotBlank(organizationHeader) && !organizationHeader.equals(originalOrganization.getId())) {
 
         // Organization switching is only allowed if the request is coming in with the global admin role enabled
         if (!originalUser.hasRole(GLOBAL_ADMIN_ROLE)) {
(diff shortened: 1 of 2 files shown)


mtneug:
How exactly can you reproduce this?

geichelberger:
With this change, the header is no longer validated in every case; code later in the pipeline cannot rely on it. While this may not currently cause any issues, it's always a good practice to be defensive.

That seems more like a bug in the code that's causing the exception to be raised.

KatrinIhler:
> How exactly can you reproduce this?

That's a bit complicated. What you'd need is an Opencast that's also serving files, but configure a different download.url in custom.properties for that file-serving node so Opencast doesn't know it's part of the same cluster. If you then distribute files with `checkAvailability` set to true (e.g. by starting a workflow) you should get stack traces in the log with org switching errors in the header.

KatrinIhler:
> With this change, the header is no longer validated in every case; code later in the pipeline cannot rely on it. While this may not currently cause any issues, it's always a good practice to be defensive.
> 
> That seems more like a bug in the code that's causing the exception to be raised.

This is certainly a multi-part problem and I'm gonna contribute a few more patches for this. But I think this is still a reasonable thing to do. The org header is validated in the sense that we know then that it contains the same org that is already set in the security service, in which case the header has no effect anyway - it is only used in this filter to potentially switch the organization, and that is all it is meant to be used for. Or am I missing something?

mtneug:
> That's a bit complicated. What you'd need is an Opencast that's also serving files, but configure a different download.url in custom.properties for that file-serving node so Opencast doesn't know it's part of the same cluster. If you then distribute files with `checkAvailability` set to true (e.g. by starting a workflow) you should get stack traces in the log with org switching errors in the header.

ok so this is affecting the node that serves the files, i.e. it is the one throwing the exception? We run this configuration only that we serve files directly using nginx. Probably because of this, we have not seen exceptions.

I wonder if this configuration would then simply require the user to add the URL configured in `download.url` to one Opencast tenant. This can be done easily in `etc/org.opencastproject.organization-mh_default_org.cfg`:

```
prop.org.opencastproject.host.my-download-url=<same value as in download.url without the path>
```

In this way, this domain-port pair is marked as to belong to the tenant `mh_default_org`. This also works for single-tenant systems. For multi-tenant systems, this mapping can only be done for a single tenant, e.g. `mh_default_org`, but the org switching via headers should still work.

KatrinIhler:
@mtneug Yeah, that's what I'm gonna recommend to that client as well. Still I felt that for a single-tenant system, that error shouldn't happen since there is no actual org switching.

gregorydlogan:
This fix makes sense to me, though I obviously don't run production clusters with multiple tenants.  Given that there's been no -1s, I'm going to merge this.

PR #7671, opened 2026-05-21
Modify targets of opencast.service unit file

Akumatic:
Currently the unit only checks for network.target. This only indicates that the network stack has been started. It doesn't indicate whether network interfaces are configured or a network connection is established.

When the database and Elasticsearch are running on different servers, you want to be able to communicate with them. systemd offers a target for that case: `network-online.target`.

`After=` only affects ordering. To make sure a network connection is established, either `Wants=` or `Requires=` is needed as well. This commit choses `Wants=` as weak depencency - this way, Opencast can still start if `network-online.target` fails and is independend of that target.

See also https://systemd.io/NETWORK_ONLINE/

### How to test this patch

Update the systemd unit file, reload the daemon via `systemctl daemon-reload` and restart `opencast.service`

### DIFF
--- a/docs/scripts/service/opencast.service
+++ b/docs/scripts/service/opencast.service
@@ -2,8 +2,10 @@
 Description=Opencast
 Documentation=https://docs.opencast.org
 
+Wants=network-online.target
+
 After=local-fs.target
-After=network.target
+After=network-online.target
 After=remote-fs.target
 
 [Service]


Akumatic:
ICLA has now been submitted

gregorydlogan:
@lkiesow IIRC I copied this file into the deb packaging and thus need to update my scripts.  Unclear if you have done the same, but I figured I'd ping you to ask.

gregorydlogan:
I have checked both the RPMs and debs, and we actually source the service file from the tarballs/sources so there's no need to modify those repositories!

As far as I'm concerned this can go in, so merging.

Akumatic:
> As far as I'm concerned this can go in, so merging.

It might also be worth picking the commit for the currently supported versions so that the change takes effect when new packages for these versions are built

PR #6901, opened 2025-07-14
Allow multiple videos for SubtitleTimeShift WOH

Arnei:
Helps with #6895.

The SubtitleTimeshiftWOH only allowed one flavor with exactly one track in it for video source flavors. This changes it so that it allows multiple flavors with multiple tracks in it. This is supposed to make it easier to write workflows where the subtitles have to be shifted by the duration of multiple tracks (e.g. by a themes title slide AND bumper video).

This changes a WOH configuration key which should be mentioned in the release notes? Might also mean that this cannot get into r/18.x?

### How to test this patch

Will require changing your workflow to include this operation. Test the configuration `video-source-flavors` with different flavors and flavor contents.

### Your pull request should…

* [ ] have a concise title
* [ ] [close an accompanying issue](https://docs.opencast.org/develop/developer/#participate/development-process/#automatically-closing-issues-when-a-pr-is-merged) if one exists
* [ ] [be against the correct branch](https://docs.opencast.org/develop/developer/development-process#acceptance-criteria-for-patches-in-different-versions)
* [ ] include migration scripts and documentation, if appropriate
* [ ] pass automated tests
* [ ] have a clean commit history
* [ ] [have proper commit messages (title and body) for all commits](https://medium.com/@steveamaza/e028865e5791)
* [ ] explain why it needs to be merged into the legacy branch, if it is targeting the legacy branch

PR #7697, opened 2026-06-03
Fix missed subtitle-timeshift configuration key

Akumatic:
With #6901, the subtitle-timeshift configuration key `video-source-flavor` changed to `video-source-flavors` in Opencast 19. While one occurance got adjusted in `partial-theming.yaml`, a second one was missed and is still using the old config key.

Since that behavior got introduced in OC 19, it should also target that branch and be picked/merged to newer versions afterwards

### DIFF
--- a/etc/workflows/partial-theming.yaml
+++ b/etc/workflows/partial-theming.yaml
@@ -119,7 +119,7 @@ operations:
     configurations:
       - subtitle-source-flavor: chapters/trimmed
       - target-flavor: chapters/shifted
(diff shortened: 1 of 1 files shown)


Akumatic:
> Hi @Akumatic Thank you for contributing to Opencast. We noticed that you have not yet filed an [Individual Contributor License Agreement](https://www.apereo.org/licensing/agreements/icla). 

ICLA was filed on May 21

Arnei (approved):

PR #7706, opened 2026-06-08
Test used workflow definition element, configuration_panel_json

JamesUoM:
Tests checked the unused `configuration_panel` element not the current `configuration_panel_json` element. This patch addresses that.

### AI Usage
N/A


JamesUoM:
Wrong target

PR #7716, opened 2026-06-11
Improvements for Deletion

KatrinIhler:
How it started: Trying to delete an event.
How it's going: ...

The main problem was that the snapshot in the delete workflow failed bc assets were missing on disk and could not be hardlinked to. Since we want to delete anyway, we don't need to include them in this snapshot imo. 

Then deletion got stuck running the workflow bc the AssetManager had not been properly updated and still thought the engage publication existed, but retraction was skipped bc the mediapackage was no longer in search.

The rest is small docs changes and a duplicate dependency in a pom. **However**, this also renames the delete-workflow to `retract-for-deletion` to avoid confusion bc the delete workflow doesn't actually delete the event. This might be controversial. Thoughts?

### DIFF
--- a/docs/guides/admin/docs/releasenotes/delete-workflow
+++ b/docs/guides/admin/docs/releasenotes/delete-workflow
@@ -1,0 +1,2 @@
+Community delete workflow including default configuration value has been renamed to "retract-for-deletion" to avoid
+confusion.
--- a/docs/guides/admin/docs/workflowoperationhandlers/snapshot-woh.md
+++ b/docs/guides/admin/docs/workflowoperationhandlers/snapshot-woh.md
@@ -18,6 +18,7 @@ Parameter Table
 |source-tags       |text            |Comma separated list of tags. Specifies which media should be the source of a snapshot.|
 |source-flavors    |presenter/source|Comma separated list of flavors. Specifies which media should be the source of a snapshot.|
 
+If neither tags nor flavors are configured, all elements of the media package are included.
 
 Operation Example
 -----------------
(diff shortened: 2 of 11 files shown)


mtneug:
> **However**, this also renames the delete-workflow to `retract-for-deletion` to avoid confusion bc the delete workflow doesn't actually delete the event. This might be controversial. Thoughts?

We had this discussion some time ago in the dev meeting. Not quite sure what the result was. Personally, I don't really care either way, but if we rename this, this can affect external systems. IIRC, the Moodle plugins have a default config value for the name of the "delete" workflow. Well, since some of the default config values need updating anyway, this could also be a good opportunity.

KatrinIhler:
@mtneug The result was that we should get rid of the delete workflow entirely iirc, but that was out of scope for this PR. :sweat_smile: Iirc, no one was specifically against renaming. You're right that this can affect the LMS plugins, which is why I added release notes and didn't aim this at a minor release. This should imo be caution enough, but I'm also not that attached to this particular change - I just encountered two different people that were confused about what that workflow does/doesn't do. :D

Arnei (changes requested):
These changes look good and make sense to me. I am in favour of giving the workflow a more proper name, even if that causes work for external systems.

The only reason I ticked "request changes" is because I am not seeing the release notes note you mentioned :D

KatrinIhler:
@Arnei Good find, forgot to commit them...

Arnei (approved):

lkiesow:
This may actually have broken deletion completely: https://github.com/opencast/opencast/issues/7911

PR #7768, opened 2026-07-18
Drop axios dependency in LTI tools

lkiesow:
axios was used in exactly one file for a handful of GET/POST/DELETE JSON calls plus one multipart upload with a progress callback. By now, native JS easily supports what axios does and we can drop the dependency.

This removes axios and its transitive dependencies (22 packages) from package-lock.json.

### How to test this patch

Mostly covered by the LTI selenium tests, but you can go to `/ltitools/index.html` and tests if the functionality still works. Keep in mind though that some edge cases (e.g. upload with no series) or even error handling for that was never properly implemented.

### AI Usage
Claude Sonnet was used for parts of the work.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/lti/package-lock.json
+++ b/modules/lti/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.1.0",
       "dependencies": {
         "@vitejs/plugin-react": "^6.0.0",
-        "axios": "^1.13.6",
         "bootstrap": "^5.3.8",
         "i18next": "^26.3.6",
         "i18next-browser-languagedetector": "^8.2.1",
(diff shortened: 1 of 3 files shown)


Arnei:
I do not see the benefit of this. As far as I know axios has been a well maintained package for years and there are no signs that that will change any time soon. It's bundle size is also relatively small compared to our other dependencies, so I'm not sure that matters much. Doing this ourselves just seems to increase code complexity (admittedly not by all that much, but still).

lkiesow:
This is not about Axios being a non-well maintained package. It is well maintained. It's just getting rid of dependencies without a lot of benefit. Axios in here comes from a time before `fetch()` was well established. It was the go-to library to make requests manageable. This has changed and using `fetch()` isn't any harder.

The only place left where axios does have a benefit in here is the upload progress bar for which the code indeed becomes slightly complexer. But as you pointed out already: Not by much.

But having axios also costs us. It has regular updates, and we get a dependency update every month. That's a pull request someone needs to spend time on. Worse, people find and the project fixes [security issues on a regular basis](https://app.opencve.io/cve/?vendor=axios). Most are completely irrelevant for us and only affect you if you run axisos in a server. But you don't know and technically, you should go through all of these and check if they affect you so you can either update axis or fix/mitigate the issue. That's even more time.

And that's time you spend on a tool that hasn't changed in years and is unlikely to change soon, so that the very slight increase in complexity doesn't matter at all: No one looks at it anyway.

That's my reasoning for getting rid of axios and why I'm always happy to reduce our dependencies unless we actually benefit from the library. I hope that explains it better.

Arnei:
Thank you for the detailed explanation, I think I understand your position better now. I still can't say that I'm convinced though :D, but I'm also not going to block this.

PR #7772, opened 2026-07-18
Fix TypeScript check in modules/lti and add it to CI

lkiesow:
The CRA→Vite migration (e5d1298c07) removed the Jest toolchain but left src/App.test.tsx behind, which relied on Jest's global `it()`. Since no test runner exists for this module anymore, this patch removes the orphaned file. This also adds a `tsc --noEmit` step to the LTI tools workflow so this kind of breakage is caught in CI.

### How to test this patch

CI should pass.

### AI Usage
Worked with Clause Sonnte.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests


.github/workflows/test-lti-tools.yml
```diff
@@ -25,6 +25,10 @@ jobs:
       working-directory: modules/lti/
       run: npm ci
 
+    - name: type check
+      working-directory: modules/lti/
+      run: npx tsc --noEmit -p tsconfig.json
+
```
Arnei, line 31: If this is important enough to us to fail CI, we might as well fail fast and add it to the build step, similar as we do it in the admin ui and editor.

modules/lti/src/App.test.tsx
```diff
@@ -1,10 +0,0 @@
-import React from 'react';
```
Arnei, line 1: This is probably fine to get rid of because we still have selenium tests which also achieve this. So +1 for removing this instead of reintroducing jest.

Arnei (changes requested):

.github/workflows/test-lti-tools.yml
lkiesow: If we are using TypeScript rather than JavaScript, I think it makes sense to type-check the code.
For the positioning: You mean putting it in as part of the `npm run build` command?

Arnei: Yes, that is what I meant, sorry for being unclear. Like: https://github.com/opencast/admin-interface/blob/f38be3e708d00a4f7f666a8464e65cb3330f49be/package.json#L47

PR #7774, opened 2026-07-19
Elect the job dispatcher automatically via a database lease

### LINKED ISSUE #7367: Failover for Job Dispatching
Automatically pick one node to do job dispatching. No configuration needed. If one node goes down, the next one should take over.

lkiesow:
Job dispatching (deciding which node picks up and runs a queued job) previously required manually enabling `dispatch.interval` on exactly one cluster node and disabling it everywhere else. Getting this wrong either stopped dispatching cluster-wide (if the sole enabled node went down) or caused two nodes to dispatch independently against the same database (if accidentally enabled on more than one), with only JPA optimistic locking on JpaJob as a soft, after-the-fact safeguard against the resulting races, which, in practice, caused issues for the entire cluster under load.

This introduces automatic leader election so admins no longer have to hand-pick a dispatch node:

- New `oc_dispatch_lock` table/entity (DispatchLockJpaImpl) holds a single-row lease: an owner and an expiry timestamp, seeded already expired (epoch) so the first node to look for it can always claim it. The table is created automatically by EclipseLink on next boot, same as every other entity in this persistence unit; no hand-written DB migration is needed.
- New DispatchLeaseManager acquires/renews the lease with a single, portable JPQL bulk UPDATE, comparing the stored expiry against the database's own CURRENT_TIMESTAMP rather than any node's local clock. Race safety comes entirely from the database engine's row-level locking during the update itself, so this works identically on MariaDB, PostgreSQL, and the H2 database used in tests, with no advisory locks and no vendor-specific SQL.
- JobDispatcher now tries to acquire or renew the lease at the start of every dispatch tick and only proceeds if it currently holds it. The lease is released proactively on clean shutdown and on reconfiguration, so a standby node can take over immediately instead of waiting out the full lease timeout. A new `dispatch.lock.timeout` property controls that timeout (default: 3x the dispatch interval).
- Dispatching is now enabled by default (`dispatch.interval=2`) on every node that runs this component, flipping the model from opt-in-one-node to opt-out-per-node: admins only need to touch config to explicitly exclude a node from ever becoming the dispatcher (`dispatch.interval=0`). Running dispatching on multiple admin/all-in-one nodes for HA is now the supported configuration rather than a misconfiguration to avoid.
- Removed the build-time `enableJobDispatching` mechanism (assemblies/*/pom.xml properties, the "job dispatching" Ant target in assemblies/resources/build.xml, and its invocation in assemblies/pom.xml) that used to patch the packaged config file per assembly, since the shipped default now covers it directly.
- Documented the mechanism in the developer architecture guide (job-dispatch.md, previously just a stub) and added an HA section to the multi-server admin installation guide.

Existing installs that already set `dispatch.interval` explicitly (whether 0 or a positive value) keep that behavior unchanged after upgrade; only nodes where the property was never set pick up the new default.

Adds DispatchLeaseManagerTest, covering contested acquisition between multiple candidates, self-renewal, TTL-based failover, explicit release, and releasing when not the current holder. Also adjusts the ComponentContext fixture in ServiceRegistryJpaImplTest, which previously disabled dispatching via `dispatch.interval=0` while manually driving the dispatch runnable directly for deterministic tests - it now uses a large interval so the periodic schedule never fires during the test while still letting the lease manager be created and the lease be won.

This fixes #7367

### How to test this patch

First, test with an all-in-one as usual. Nothing should change there. Then, build a small cluster. Here things get interesting:

First make sure you have a cluster-wide database (I use PostgreSQL for the following examples). Maybe use the dev containers for that:
```
podman-compose -f docker-compose-postgresql.yml up -d
```

Next, build the `allinone` distribution. I used two `allinone` nodes as the cluster:
```
mvnw clean install -Pallinone -DskipTests -Dcheckstyle.skip=true
```

Next, configure the cluster. To make things easy, put the following in `build/prep.sh` and then run the script to configure both nodes with `cd build; sh prep.sh`:

```sh
#!/bin/sh

set -eux

tar xf opencast-dist-*.tar.gz
mv opencast-dist-allinone opencast-dist-allinone-primary
tar xf opencast-dist-*.tar.gz
mv opencast-dist-allinone opencast-dist-allinone-secondary

mkdir -p shared-storage
ln -s ../../shared-storage opencast-dist-allinone-primary/data/opencast
ln -s ../../shared-storage opencast-dist-allinone-secondary/data/opencast

sed -i 's_localhost:8080_localhost:8081_' opencast-dist-allinone-secondary/etc/custom.properties
sed -i 's_port=8080_port=8081_' opencast-dist-allinone-secondary/etc/org.ops4j.pax.web.cfg

for node in allinone-primary allinone-secondary; do
  # configure database
  sed -i 's_^.*db.jdbc.driver=.*$_org.opencastproject.db.jdbc.driver=org.postgresql.Driver_' \
    "opencast-dist-${node}/etc/custom.properties"
  sed -i 's_^.*db.jdbc.url=.*$_org.opencastproject.db.jdbc.url=jdbc:postgresql://127.0.0.1/opencast_' \
    "opencast-dist-${node}/etc/custom.properties"
  sed -i 's_^.*db.jdbc.user=.*$_org.opencastproject.db.jdbc.user=opencast_' \
    "opencast-dist-${node}/etc/custom.properties"
  sed -i 's_^.*db.jdbc.pass=.*$_org.opencastproject.db.jdbc.pass=dbpassword_' \
    "opencast-dist-${node}/etc/custom.properties"

  # configure organization
  sed -i 's_^.*admin.ui.url=.*$_prop.org.opencastproject.admin.ui.url=http://localhost:8080_' \
    "opencast-dist-${node}/etc/org.opencastproject.organization-mh_default_org.cfg"
  sed -i 's_^.*engage.ui.url=.*$_prop.org.opencastproject.engage.ui.url=http://localhost:8081_' \
    "opencast-dist-${node}/etc/org.opencastproject.organization-mh_default_org.cfg"
  sed -i 's_^.*file.repo.url=.*$_prop.org.opencastproject.file.repo.url=${prop.org.opencastproject.admin.ui.url}_' \
    "opencast-dist-${node}/etc/org.opencastproject.organization-mh_default_org.cfg"

  # Enable debug logs for service registry
  sed -i 's/#log4j2.logger.ingest.level = DEBUG/log4j2.logger.ingest.level = DEBUG/' \
    "opencast-dist-${node}/etc/org.ops4j.pax.logging.cfg"
  sed -i 's/#log4j2.logger.ingest.name = org.opencastproject.ingest/log4j2.logger.ingest.name = org.opencastproject.serviceregistry.impl/' \
    "opencast-dist-${node}/etc/org.ops4j.pax.logging.cfg"
done
```

Then in two terminal windows run `./bin/start-opencast` in the primary and secondary folders. You should see the one you start first dispatching jobs in the debug logs if you run `log:tail` in the Karaf consoles. The second one should skip dispatching.

If you turn off the one you start first, the secondary should take over.

### How this looks in action

Screenshot from 2026-07-19 13-10-16

*Secondary is dispatching jobs. Primary is skipping job dispatching.*

—

Screenshot from 2026-07-19 13-10-32

*Primary took over job dispatching after secondary was shut down.*

### AI Usage

Code and docs mostly generated with Claude Sonnet.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

### DIFF
--- a/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/DispatchLeaseManager.java
+++ b/modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/DispatchLeaseManager.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to The Apereo Foundation under one or more contributor license
+ * agreements. See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ *
+ * The Apereo Foundation licenses this file to you under the Educational
+ * Community License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License
+ * at:
+ *
+ *   http://opensource.org/licenses/ecl2.txt
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+
+package org.opencastproject.serviceregistry.impl;
+
+import static org.opencastproject.db.Queries.namedQuery;
+
+import org.opencastproject.db.DBSession;
+import org.opencastproject.serviceregistry.impl.jpa.DispatchLockJpaImpl;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Date;
+
+/**
+ * Elects exactly one job dispatcher across a cluster of candidate nodes by acquiring and renewing a lease stored in
+ * a single, shared database row. Only the node currently holding the lease should perform job dispatching; all
+ * other candidate nodes must sit idle until they acquire it, e.g. after the current holder stops renewing it.
+ *
+ * <p>Acquisition and renewal are both done with a single, portable JPQL bulk update whose {@code WHERE} clause
+ * compares the stored expiry against the database's own {@code CURRENT_TIMESTAMP}, not the calling node's local
+ * clock. This makes the race between competing nodes safe by relying purely on the database engine's row-level
+ * locking during the update, without needing vendor-specific advisory locks, and makes every competing node judge
+ * lease freshness against one shared clock rather than against each other's local clocks.
+ */
+public class DispatchLeaseManager {
+
+  private static final Logger logger = LoggerFactory.getLogger(DispatchLeaseManager.class);
+
+  private final DBSession db;
+  private final String owner;
+  private final long leaseTimeoutSeconds;
+
+  public DispatchLeaseManager(DBSession db, String owner, long leaseTimeoutSeconds) {
+    this.db = db;
+    this.owner = owner;
+    this.leaseTimeoutSeconds = leaseTimeoutSeconds;
+  }
+
+  /**
+   * Attempt to acquire the dispatch lease, or renew it if this node already holds it.
+   *
+   * @return true if this node is (now, or still) the elected dispatcher, false otherwise
+   */
+  public boolean tryAcquireOrRenew() {
+    ensureSeeded();
+
+    Date newExpiry = new Date(System.currentTimeMillis() + leaseTimeoutSeconds * 1000);
+    int updated = db.execTx(namedQuery.update(
+        "DispatchLock.acquire",
+        Pair.of("id", DispatchLockJpaImpl.SINGLETON_ID),
+        Pair.of("owner", owner),
+        Pair.of("leaseExpires", newExpiry)
+    ));
+
+    boolean isLeader = updated == 1;
+    logger.trace("Dispatch lease acquire/renew attempt by '{}': {}", owner, isLeader ? "leader" : "not leader");
+    return isLeader;
+  }
+
+  /**
+   * Release the lease if this node currently holds it, allowing another candidate node to take over immediately
+   * instead of waiting for the lease to expire. Safe to call even if this node isn't the current leader.
+   */
+  public void release() {
+    int updated = db.execTx(namedQuery.update(
+        "DispatchLock.release",
+        Pair.of("id", DispatchLockJpaImpl.SINGLETON_ID),
+        Pair.of("owner", owner)
+    ));
+    if (updated == 1) {
+      logger.debug("Released dispatch lease held by '{}'", owner);
+    }
+  }
+
+  /**
+   * Make sure the singleton lock row exists. Safe to call from multiple nodes concurrently; only one insert wins,
+   * the rest are simply ignored.
+   */
+  private void ensureSeeded() {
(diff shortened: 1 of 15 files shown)


modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/DispatchLeaseManager.java
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Attempt to acquire the dispatch lease, or renew it if this node already holds it.
+   *
+   * @return true if this node is (now, or still) the elected dispatcher, false otherwise
+   */
+  public boolean tryAcquireOrRenew() {
+    ensureSeeded();
+
+    Date newExpiry = new Date(System.currentTimeMillis() + leaseTimeoutSeconds * 1000);
```
gregorydlogan, line 68: This introduces the requirement that the cluster have clocks which are more or less in sync (specifically within `leaseTimeoutSeconds`).  This should probably be documented, or we'll find that one adopter who has a cluster which spans timezones will complain.

Alternatively, the job dispatch markdown explicitly says "The comparison uses the database's own `CURRENT_TIMESTAMP` rather than any node's local clock", can we make the JPQL above something like 
```
@NamedQuery(
        name = "DispatchLock.acquire",
        query = "UPDATE DispatchLock d SET d.owner = :owner, d.leaseExpires = CURRENT_TIMESTAMP + leasyTimeoutSeconds * 1000"
            + "WHERE d.id = :id AND (d.owner = :owner OR d.leaseExpires < CURRENT_TIMESTAMP)"
    ),
```

and avoid even touching *any* node's local clock?  We're interested in the arbiter of truth's opinion (the database), not any node's wacky sense of time, right?

gregorydlogan (changes requested):
LGTM aside from me nitpicking that the docs and implementation should match ;)

modules/serviceregistry/src/main/java/org/opencastproject/serviceregistry/impl/DispatchLeaseManager.java
lkiesow: > This should probably be documented, or we'll find that one adopter who has a cluster which spans timezones will complain.

This should already be timezone agnostic:
- `System.currentTimeMillis()` returns milliseconds since the UTC epoch. It's completely timezone-agnostic — a JVM running with `TZ=America/New_York` and one running `TZ=Asia/Tokyo` will return the same value at the same instant. `java.util.Date` similarly just wraps an epoch-millis long; it has no timezone of its own.
- So two nodes literally in different timezones, but with clocks properly synced (e.g. via NTP), would compute the exact same `newExpiry` instant. Timezone offset alone causes no error.

IIRC, local time drift would also not end up in a big problem, but I need to think through this again later. Also, getting this into the DB would actually be nice, but I remember that there was a problem with H2 and time calculations. But I'll check later and come back to you here.

gregorydlogan: Ok, for some reason I thought `currentTimeMillis` was affected by TZ.  That plus NTP (which, really, if you're not NTPing your servers you have other, far more serious issues) solves 90% of the issue.  Will let you poke at the DB side of things, but imo this is clear to merge prior to the 21 branch.

Issue #7664, opened 2026-05-19
API Tests


KatrinIhler:
Since we're planning major changes in the near future (e.g. the switch to Quarkus), it'd be nice if we would notice immediately if endpoint behavior changed. The university of Osnabrück has a student that we tasked with this job, this issue is mostly for documentation purposes. :)

KatrinIhler:
Notes from our meeting today (in German, I'm too lazy to translate rn)

```
- automatisiert via Github
- develop.opencast.org als Basis erstmal, später eigene Instanz (z.B. Docker von Quarkus)
- wie häufig - hängt von Performance ab 
- Frameworks: 
  - Bruno
    - wie Postman
    - lokal, git-native, nutzbar über GHA
    - einfach zu benutzen
    - Tests in Javascript :/
    - Open-source, kostenlos
  - Python
    - pytest
    - unit tests
    - mehr Code
  - Quarkus
    - läuft kontinuierlich, erkennt Änderungen, macht nur dazu tests → brauchen wir das?
    - komplizierter & langsamer, solange noch kein Quarkus
    - beim Bauen schon inkludieren?
  - → Bruno ausprobieren
  - Bes. wichtig: /api, /search/, alles für CAs
  - GraphQL noch nicht
```

KatrinIhler:
- test /ingest with Bruno
- item is object, list of item is array
- try to break endpoints
- don't redirect to login from api, get 403
- error handling - wrongly documented, inconsistent errors
- inconsistent formatting for same entity, e.g. get & put/post for acls
- relevant: /api, /search, endpoints for CAs: https://docs.opencast.org/develop/developer/#architecture/capture-agent/capture-agent/

gregorydlogan:
Suggestion: Do the CA specific ones *last* (cf `/capture-admin`) since these are under active development.  Get the rest of it going, and then either Elan, or me can add the CA specific tests to the rest of the bunch.

KatrinIhler:
> Suggestion: Do the CA specific ones last (cf /capture-admin) since these are under active development. Get the rest of it going, and then either Elan, or me can add the CA specific tests to the rest of the bunch.

Imo not really necessary, bc writing basic tests is pretty fast and the new API will still be backwards-compatible.

PR #7775, opened 2026-07-19
feat/wip: Introduce initial Bruno API test suite and GitHub Workflow integration

philippschein:
This pull request introduces the first, unfinished iteration of an automated API test suite for Opencast using **Bruno**. It also adds the initial integration of these tests into our GitHub Actions workflows. 

I am developing this as a working student at the University of Osnabrück. The main goal of this initial PR is to establish the base folder structure, add the first set of API requests, and ensure the CI pipeline successfully triggers and runs. Since this is a work in progress, full test coverage will follow in upcoming iterations.

### How to test this patch

**Testing the Bruno collection locally:**
1. Install the Bruno desktop app from [usebruno.com](https://www.usebruno.com/) or install the CLI globally:
```bash
npm install -g @usebruno/cli
```

**Using the Desktop App:**
* Open the Bruno desktop app, select **Open Collection**, and choose the new Bruno folder introduced in this repository.
* Select **Opencast-Develop** as the environment.
* Right-click the test folder and click on **Run**.

**Using the CLI:**
* To run the tests via the command line, navigate to the Bruno folder and execute:
```bash
bru run --env-file Opencast-Develop
```

**Testing the GitHub Workflow:**
1. Check the Actions tab on this pull request to verify that the new workflow installs the Bruno CLI and executes the test suite properly.

### AI Usage

I used an AI assistant to create the initial test structure of the Bruno tests and to create some of the more complicated assertion tests. For this text I translated my notes with AI and used it to draft this pull request description in English.

### Your pull request should…

* [x] have a concise title
* [x] pass automated tests

Related to #7664


mtneug:
Welcome @philippschein 👋 

Thanks for this contribution. I only skimmed through the changes, but this seems very promising. Some general remarks:

* I don't think `docs/bruno_api_tests` is a good location for end-to-end tests. Having this under docs seems unexpected to me. How about just `test/e2e`?
* The file/folder names inconsistently use upper/lower case.
* We already have test media files in the Opencast repository (search for *.mp4, *.mp3, *.zip). I would suggest using those files instead of adding new ones. You can copy them over to your test folder. Git should notice that the content is the same and thus keep the repository small.

test/e2e/environments/opencast_develop.bru
```diff
@@ -0,0 +1,7 @@
+vars {
+  baseUrl: https://stable.opencast.org
```
Arnei: You probably either want develop.opencast.org, or rename the file to opencast_stable.bru

test/e2e/tests/api/api_events/post_event.bru
```diff
@@ -0,0 +1,70 @@
...
+vars:post-response {
+  eventId: res.body.identifier
+}
+
+script:post-response {
+  if (res.getStatus() === 201) {
+    const axios = require('axios');
+    const id = res.getBody().identifier;
+    const base = bru.getEnvVar('baseUrl');
+    const auth = { username: bru.getEnvVar('username'), password: bru.getEnvVar('password') };
```
Arnei: Can you actually get the secret like this, given that in `opencast_develop.bru` it is in `vars.secret` instead of `vars`?

.github/workflows/test-opencast-build.yml
```diff
@@ -171,6 +171,28 @@ jobs:
           jq -r '.workflows.workflow.state' < instances.json | tee status
           grep -q SUCCEEDED status
 
+      - name: run bruno tests
+        if: matrix.java == 21
+        continue-on-error: true
+        working-directory: test/e2e
+        run: |
+          npx -y @usebruno/cli run tests -r \
```
Arnei: Could we cache that installation?

test/e2e/tests/ingest/ingest.bru
```diff
@@ -0,0 +1,30 @@
...
+  mediaPackage: {{mediaPackage_B}}
+}
+
+assert {
+  res.status: eq 200
+  res.headers["content-type"]: contains xml
+}
+
+tests {
+  test("Ingest media package should start workflow and return workflow XML", function() {
```
Arnei: This test is actually going to start a full blown workflow which uses a bunch of processing resources. If all we care about is whether the endpoint returns `INSTANTIATED`, maybe we could run an empty workflow like noop.yaml instead?

test/e2e/tests/api/api_groups/post_group_invalid.bru
```diff
@@ -0,0 +1,27 @@
...
+  name: post_group_invalid
+  type: http
+  seq: 9
+}
+
+post {
+  url: {{baseUrl}}/api/groups
+  body: formUrlEncoded
+  auth: inherit
+}
```
Arnei: A lot of these `_invalid` test files seems to just copy some stuff over from their non-invalid original. Any chance we could reduce code duplication here? Generally I am somewhat worried about some tests being failed to update if their tested endpoint changes.

Arnei (changes requested):
Adding a ReadMe file to the test folder would be nice. Could give a general explanation and how to run the tests.

Speaking of running the tests, the cli command did not work:
```
bru run --env-file Opencast-Develop
```

Should probably be

```
bru run --env-file environments/opencast_develop.bru
```

or if you *actually* are trying to test locally
```
bru run --env-file environments/opencast_develop.bru --env-var baseUrl=http://localhost:8080 -r tests
```

I also tried getting the deskop app to run, but all test requests fail with

> getaddrinfo ENOTFOUND {{baseurl}}


Disclaimer: I did not read all of the tests, nor do I plan to. That's a lot of tests.

test/e2e/opencast_develop.json
```diff
@@ -0,0 +1,26 @@
+{
```
Arnei: Is this also an "environment" file? If so, why do we have two of those in different places?

gregorydlogan:
> Adding a ReadMe file to the test folder would be nice. Could give a general explanation and how to run the tests.

I'm a CLI only guy, and with minor tweaks (mainly ripping the command from the workflow rather than the PR description) I have the CLI side working

> Comprehension question: If I understand this right, all the tests in a folder run in a sequence defined by the meta parameter `seq`. And state, which probably means mainly changes to Opencast, carries between tests? As someone who has mainly dealt in unit tests so far this feels a bit strange. If I want to add a test to say, `api_events`, that means I have either read all the other tests before I can append my new test, and/or make sure my test does not change state in a way that it breaks other tests. That seems like a lot of potential complexity. Anything we can do about that? Is there anthing like `setUp` or `tearDown`? Do we even _want_ to run all of these tests in sequence?

This is definitely a concern, and honestly not something we can really do much about short of nuking the filesystem and database between tests - not reasonable given 21's current slow restart times.  Running multiple passes of the tests in sequence doesn't seem to break anything, but I'm also not confident (also have not read the full set of tests) that run N is looking at copy N's test workflows rather than N-1's.  More complex tests, eg: putting things into maintenance mode will necessitate more caution in the future I suspect.

Looking at what actually happens when I point this at localhost, we can't run this full set of tests as is, for every build of every PR.  We just can't, the test boxes would keel over and die very quickly.  Even if we were to change the workflow to noop, it would hopelessly clutter up the admin ui on public test infra.  We either need a more private instance where we actually don't care, we need to only run certain tests nightly, or we need the tests to clean up after themselves and leave nothing behind.  Given that each test should be more or less self contained, this might remove Arne's issue above about test sequencing.  Each test should assume it's the only one, build whatever it needs, then remove it all.  This increases test complexity, but I don't see any other approach.

Other than that though, I very much approve of having these.  The idea of being able to review a PR for changes, rather than verifying that existing behaviour hasn't broken is very appealing.

